PR #824, opened 2015-02-05
Changed property filtering to use JSON path instead of reflection.

Ceiu:
- JSON property filtering now uses the current JSON context "path"
  to determine whether or not a property should be included rather
  than using reflection on each object.
- Removed a lot of code obsoleted by this change.
- Added JSON filtering to Gutterball's DTOs. Currently only basic
  filtering is being performed on these objects. If HATEOAS
  filtering is necessary in the future, it will need to be added
  to the JsonProvider.

### DIFF
--- a/common/src/main/java/org/candlepin/common/jackson/DynamicPropertyFilter.java
+++ b/common/src/main/java/org/candlepin/common/jackson/DynamicPropertyFilter.java
@@ -15,10 +15,14 @@
 package org.candlepin.common.jackson;
 
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonStreamContext;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.ser.PropertyWriter;
 
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * DynamicPropertyFilter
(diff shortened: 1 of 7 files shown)


mstead:
This looks good for gutterball. Verifying that all is well in CP -- spot checks look Ok. I'm having some issues with getting spec tests to pass, but I believe this is unrelated.

common/src/main/java/org/candlepin/common/jackson/DynamicPropertyFilter.java
```diff
@@ -28,12 +32,27 @@
...
     public boolean isSerializable(Object obj, JsonGenerator jsonGenerator,
         SerializerProvider serializerProvider, PropertyWriter writer) {
-        DynamicFilterData filterData =
-            ResteasyProviderFactory.getContextData(DynamicFilterData.class);
+
+        DynamicFilterData filterData = ResteasyProviderFactory.getContextData(DynamicFilterData.class);
+
         if (filterData != null) {
-            return !filterData.isAttributeExcluded(writer.getName(), obj);
+            this.path.clear();
```
mstead: We can define the 'path' collection in this method so that we didn't have to clear it each time. Any reason that you define path collection as a field?

Ceiu: Saves some reallocation (assuming Java doesn't throw it on the stack) and initialization in a method that is run a lot.

PR #825, opened 2015-02-06
Eliminate Lock Contention on Consumer Last Check-In Time

dgoodwin:
Problems surface in deployments with extremely high load. Because we update the consumer's row every time they authenticate, any request running long can cause other requests to wait longer than mysql's 50 second limit trying to get the lock to update that row.

To work around this I introduced a new table for consumer check-in times. The Consumer's last checkin time is calculated as the max in the table for that consumer.

A job is included to run periodically (every 8 hours by default) and clean out all but the most recent row for each consumer. In a really big deployment, this could be well into the millions. Multiple attempts were made to delete all but the most recent row for each consumer, none would scale to that level and work on all dbs. Instead we query checkins to be deleted in batches and delete them one chunk (30k) at a time until there are none left.

End result should eliminate lock wait timeouts on cp_consumer.lastcheckin because we're now doing fresh inserts in each request, rather than everyone trying to update the same row.

### DIFF
--- a/server/src/main/java/org/candlepin/model/CheckIn.java
+++ b/server/src/main/java/org/candlepin/model/CheckIn.java
@@ -1,0 +1,88 @@
+/**
+ * Copyright (c) 2009 - 2012 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.model;
+
+import org.hibernate.annotations.ForeignKey;
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Index;
+
+import java.util.Date;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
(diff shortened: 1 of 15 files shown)


server/src/main/java/org/candlepin/model/CheckInCurator.java
```diff
@@ -0,0 +1,77 @@
...
+        // them until we stop finding results.
+        q.setMaxResults(DELETE_BATCH_SIZE);
+        List<String> checkInIds = q.list();
+
+        int totalDeleted = 0;
+        int totalQueries = 1;
+        while (checkInIds.size() > 0) {
+            log.debug("Found {} check-in IDs to be deleted.", checkInIds.size());
+            String deleteHql = "DELETE from CheckIn where id in (:deleteIds)";
+            Query deleteQuery = currentSession().createQuery(deleteHql);
```
Ceiu: Minor optimization: Could move this query out of the loop to save some re-parsing time here (much like the select query above).

dgoodwin: Good idea, on it.

server/src/main/java/org/candlepin/model/CheckIn.java
```diff
@@ -0,0 +1,101 @@
...
+ * Checkins are added to this table, with older entries cleaned out after some configurable
+ * period of time. We're really only interested in the latest, but trying to update a
+ * single column was a frequent cause of lock contention in heavily loaded environments.
+ */
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.PROPERTY)
+@Entity
+@Table(name = "cp_consumer_checkin")
+public class CheckIn extends AbstractHibernateObject {
+
```
wottop: Any chance we need to be more specific, i.e. ConsumerCheckIn ?

dgoodwin: Hmm, next up will be something like GuestIdCheckIn, I could go either way, could consider CheckIn just the most common ConsumerCheckIn, or we could be more specific. Think we should go specific?

wottop:
After thinking about it, we can change it later just as easily as now.

ack

PR #826, opened 2015-02-09
Fix broken db upgrade on Oracle.

dgoodwin:
All DBMSes appear to create an index on a primary key, Oracle actually errors
out if you try to create another. (Postgres and MySQL do not)


wottop:
ack

deployed on all 3 db's

PR #827, opened 2015-02-10
Some updates to the GB reports params

mstead:
- changed 'custom' to 'custom_results' to avoid confusion with custom products (asked about it a couple of times)
- added actual report parameters for include/exclude parameters.
- updated description for consumer status report 'status' parameter


Ceiu:
ACK

PR #830, opened 2015-02-17
Allows attribute on SKU to enable or disable content for all associated ...

wottop:
...products

### DIFF
--- a/server/spec/sku_level_enable_override_spec.rb
+++ b/server/spec/sku_level_enable_override_spec.rb
@@ -1,0 +1,80 @@
+require 'spec_helper'
+require 'candlepin_scenarios'
+
+describe 'SKU Level Enable Override' do
+  include CandlepinMethods
+  include CertificateMethods
+
+  before(:each) do
+    @owner = create_owner random_string('test_owner')
+    @org_admin = user_client(@owner, random_string('user'))
+    @env = @org_admin.create_environment(@owner['key'], 'testenv1',
+      "My Test Env 1", "For test systems only.")
+
+    @content1 = create_content
+    @content2 = create_content
+    content_list = @content1.id + "," + @content2.id
+    @product1 = create_product(nil, nil,  {:attributes => { :content_override_active => content_list }})
+    @product2 = create_product(nil, nil,  {:attributes => { :content_override_disabled => content_list }})
+    @providedProduct = create_product
+    @cp.add_content_to_product(@providedProduct['id'], @content1['id'], false)
+
+    @cp.create_subscription(@owner['key'], @product1['id'], 10, [@providedProduct.id])
+    @cp.create_subscription(@owner['key'], @product2['id'], 10, [@providedProduct.id])
+    @cp.refresh_pools(@owner['key'])
+  end
+
+  it 'sku override for enable shows on provided product in entitlement' do
+    consumer = @org_admin.register(random_string('consumer'), :system, nil,
+        {'system.certificate_version' => '3.2'})
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'],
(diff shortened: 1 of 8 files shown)


server/spec/sku_level_enable_override_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    @owner = create_owner random_string('test_owner')
+    @org_admin = user_client(@owner, random_string('user'))
+    @env = @org_admin.create_environment(@owner['key'], 'testenv1',
+      "My Test Env 1", "For test systems only.")
+
+    @content1 = create_content
+    @content2 = create_content
+    content_list = @content1.id + "," + @content2.id
+    @product1 = create_product(nil, nil,  {:attributes => { :content_override_active => content_list }})
+    @product2 = create_product(nil, nil,  {:attributes => { :content_override_disabled => content_list }})
```
dgoodwin: The attribute names don't quite match, active/inactive, enabled/disabled, but active/disabled doesn't seem right.

server/spec/sku_level_enable_override_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    @providedProduct = create_product
+    @cp.add_content_to_product(@providedProduct['id'], @content1['id'], false)
```
dgoodwin: Technically I don't think we're getting any coverage of content_override_disabled, might need another provided product with the content enabled, then assert disabled in the end.

server/src/main/java/org/candlepin/util/X509V3ExtensionUtil.java
```diff
@@ -77,10 +80,12 @@
     private static final Object END_NODE = new Object();
     private static boolean treeDebug = false;
     @Inject
-    public X509V3ExtensionUtil(Configuration config, EntitlementCurator entCurator) {
+    public X509V3ExtensionUtil(Configuration config, EntitlementCurator entCurator,
+            ProductServiceAdapter prodAdapter) {
```
dgoodwin: I fear we should hit the V1 extension util as well, I can't think of a good reason not too.

We really need to get these paths for cert generation refactored into something a little more refined sometime but that can wait.

server/src/main/java/org/candlepin/util/X509V3ExtensionUtil.java
```diff
@@ -407,9 +412,29 @@ private Branding getBranding(Pool pool, String productId) {
             }
             content.setArches(archesList);
 
+            Boolean enabled = pc.getEnabled();
```
dgoodwin: I think there's a performance problem here, we're looking up the SKU product, checking for the attributes, and splitting the strings all within a for loop that runs once per content set. However none of that data ever changes. We should probably check both attributes before entering the loop and build up a set we can check for values in.

server/spec/sku_level_enable_override_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    products.each do |p|
+      if p.id == @providedProduct.id
+        p['content'].size.should == 1
+        p['content'][0]['id'].should == @content1['id']
+        p['content'][0]['enabled'].should be_nil # true
+      end
+    end
+  end
+end
```
dgoodwin: Just realized we're limited to the size of a product attribute which is 255 characters. :( Content IDs I think are usually 4 chars, plus a ',', so the max is 51 content IDs. That might be sufficient, but we do need to check. I'll ask the stakeholders on the bugs.

dgoodwin: dgregor confirms in bugzilla that this should be fine so we're good to proceed.

dgoodwin:
ACK

PR #832, opened 2015-02-20
Guestid lock contention

dgoodwin:
Very similar to the consumer last checkin PR from a few weeks ago, this one aims to reduce lock contention on guest ID updates.

Previously every guest ID update wiped out all rows and recreated them for that host, even if nothing changed. This gave us an updated timestamp we could leverage for who most recently reported a guest ID. It also gave us a pile of lock wait timeouts in heavily loaded environments.

Instead we now insert a timestamp into a new table every time we see a consumer update that contains guest IDs. The max value in this table for a consumer is joined together with the guest IDs for that consumer so we can query who most recently reported the guest. A job is included to periodically destroy old guest checkin timestamps.


dgoodwin:
Will and I have both been through this, calling that sufficient.

PR #833, opened 2015-02-23
1194997: Don't use referencesUniqueColumn true.

dgoodwin:
This option (when set to true) is unsupported on postgres and will cause an
error, *if* you're using liquibase 2.x. The issue does not surface if you are
running liquibase 3.x.

Set to false to match all existing use of addForeignKeyConstraint element.

WARNING: This will not apply clean on an existing db with the old changeset due to checksum change. I believe this is ok only developers have this deployed, who will need to regen db clean or do a little hacking if they hit the issue.


dgoodwin:
Self-ack.

PR #834, opened 2015-03-03
Improved error message for blocked host restricted pool binds.

dgoodwin:
Previous error was causing confusion, hopefully this is better.


wottop:
ack

PR #836, opened 2015-03-05
Fix unmapped guest pools with derived product data.

dgoodwin:
Attributes and provided products were not being properly copied to unmapped
guest pools on creation, resulting in breakage of the main VDC use-case.

Some cleanup will take place in the multi-org branch.


wottop:
your branch name has a typo. not sure if it matters.

dgoodwin:
Shouldn't matter for anything no.

wottop:
ack

PR #837, opened 2015-03-05
1198494: Auto-heal continuously attaches subscriptions to make the syste...

wottop:
...m compliant on a guest machine


dgoodwin:
Looks good. ACK.

PR #840, opened 2015-03-10
Fix NPE deleting consumer without identity cert.

dgoodwin:
This can happen for ubercert consumers in Satellite at the very least, it seems
to be happening very rarely for real systems, and the cause of this is
currently unknown. None the less we should allow them to be deleted.


wottop:
ack

PR #842, opened 2015-03-10
Fix checkstyle errors and class name misspelling

wottop:


awood:
:clap: :rocket: :smiley:

PR #843, opened 2015-03-11
1200142: Support overriding content metadata expiry to 0 in Satellite.

dgoodwin:
This is a temporary solution to a problem in Satellite where clients
should always have metadata expire set to 0. In the absence of some
kind of global override for this value on all content, we'll expose
an API call which overrides all content to 0. Satellite will call
this API whenever a manifest is imported.

A big caveat here, this will have no effect on existing entitlements.
The client sees the local value as a manual edit and thus preserves
it. There is little that can be done about this, even re-binding would
require a client check-in in between the old and new entitlements.


dgoodwin:
This isn't happening. Closing and working on new solution.

PR #844, opened 2015-03-11
1200142: Force content metadata expire to 0 in standalone.

dgoodwin:
Imports only run in standalone, thus it's safe to assume that on import
we can force metadata expire to 0, allowing clients to immediately see
changes to published content.

On upgrade we'll flip all content to 0. This should be safe as these
tables are currently empty in production, thus this should only affect
standalone servers.


wottop:
ack

dgoodwin:
retest this please

PR #845, opened 2015-03-12
Mstead/gb mgmt enabled filtering fixes

mstead:
Added a new column to ComplianceStatus that tracks whether
the consumer has an entitlement that sets management_enabled.

This value is populated when a consumer event is processed.

This approach was a lot easier than having the query determine
it and avoids another subquery. The data does not change once
received and stored, so this is Ok.

Added ability to filter consumer_status by management_enabled.

Added reasons to consumer_status default output and a query parameter to allow disabling them if required.


dgoodwin:
retest this please

dgoodwin:
buildbot: retest this please

dgoodwin:
retest this please

Ceiu:
ACK

PR #846, opened 2015-03-16
Awood/missing gb deps

awood:
Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1201924

### DIFF
--- a/gutterball/deps/el7.txt
+++ b/gutterball/deps/el7.txt
@@ -55,4 +55,5 @@ scannotation
 servlet
 sisu-cglib/cglib
 slf4j/api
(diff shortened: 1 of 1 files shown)


dgoodwin:
Was this EL7 specific and only surfacing because... somebody finally tried it on EL7?

awood:
> Was this EL7 specific and only surfacing because... somebody finally tried it on EL7?

Yes. :anguished: I had been validating the EL7 builds by whether they built or not.  And the way the code was everything compiled but it left the ClassNotFoundException lurking in the darkness.

PR #848, opened 2015-03-16
1201727: Filter entitle_id from reason attributes

alikins:
For reasons related to unmapped guests, filter out
the attribute 'entitlement_id' so deployed clients
don't attempt to match it with local valid entitlement
certs, since it can reference expired entitlements.

Do this after returning from the rules so
StatusReasonMapGenerator can create a meaningful
name.

### DIFF
--- a/server/spec/unmapped_guest_spec.rb
+++ b/server/spec/unmapped_guest_spec.rb
@@ -1,6 +1,15 @@
 require 'spec_helper'
 require 'candlepin_scenarios'
 
+def assert_no_entitlement_id(reasons)
+    reasons.each do |reason|
+        reason.should have_key('attributes')
+        attributes = reason['attributes']
+        attributes.should_not have_key('entitlement_id')
+        attributes.should have_key('name')
+    end
+end
+
 # This spec tests virt limited products in a standalone Candlepin deployment.
 # (which we assume to be testing against)
 describe 'Unmapped Guest Pools' do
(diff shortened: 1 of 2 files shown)


dgoodwin:
So this effectively means you can no longer see why you're yellow in the gui. It shows yellow overall for the system, but all installed products and now entitlements appear normal. CLI will still show you, but that entitlement ID was used to mark the ent as yellow.

What do you think? Acceptable tradeoff? 

I'm still suspecting that the reason shouldn't go out if the ent is expired. You're not really using an unmapped guest pool anymore.

dgoodwin:
Recent email from Dan indicates it should be specific to the system and not any subscription, but I think in this case when it's expired, it shouldn't be sent as a reason anymore, so we should probably trace that down.

alikins:
> So this effectively means you can no longer see why you're yellow in the gui. It shows yellow overall for the system, but all installed products and now entitlements appear normal. CLI will still show you, but that entitlement ID was used to mark the ent as yellow.

Pretty much. The mysubs tab eventually uses reasons.Reasons.get_subscription_reasons_map() to 
build a map of sub/ent id -> [reasons]. ContractSubDetailsWidget() and MySubscriptionsTab uses the keys of that map as the ents to show status for.

But current client code will never populate that for expired or non valid entitlements because of the
KeyError. So this patch will cause unmapped reasons (expired or valid) to never influence the gui.

alikins:
> I'm still suspecting that the reason shouldn't go out if the ent is expired. You're not really using an unmapped guest pool anymore.

Yeah, it's not clear to me why that reason exists for expired ents.

alikins:
> So this effectively means you can no longer see why you're yellow in the gui. It shows yellow overall for the system, but all installed products and now entitlements appear normal. CLI will still show you, but that entitlement ID was used to mark the ent as yellow.

Not sure it's workable, but current clients will only fail on reasons with an entitlement_id that references an entitlement not in valid_entitlement_certs. Stacking reasons however should be okay.
It would be a little odd, but maybe we could present expired unmapped ents as faux 'stacking' reasons?

alikins:
(likely OBSOLETED by sub end date fixes)

dgoodwin:
Closing for now per last comments.

PR #849, opened 2015-03-16
Fix entitlement validity code for temp entitlements.

awood:
Note that the expired unmapped guest entitlement cleaner job is still necessary because expired items are removed at the pool level.


dgoodwin:
Commit message doesn't have a bug # in it, and the message could use some more details, I'm not sure what's going on here.

Branch name indicates bug #1199443, but that does not look related and is actually an issue I was working on with PR #850

PR #850, opened 2015-03-17
1199443: Add entitlement end date override.

dgoodwin:
In the past both pools and entitlements had a start/end date. This was removed
as the data was redundant at the time, however the Entitlement getEndDate() and
setEndDate() methods remained for API compatability, and just returned the
pool's values.

With the addition of 24h unmapped guest entitlements, we actually want to have
entitlements with a different end date than the pool. This was implemented by
stamping the correct date in the cert, but not available anywhere else in the
code causing issues with things like installed product validity date ranges.

Fixed by adding an entitlement end date override. getEndDate() will return this
value if it's set, otherwise resort to the pools. Compliance code was already
properly set up to use the entitlement end date due to the original
implementation. (from getEndDate)

### DIFF
--- a/server/src/main/java/org/candlepin/controller/Entitler.java
+++ b/server/src/main/java/org/candlepin/controller/Entitler.java
@@ -252,20 +252,13 @@ public class Entitler {
                 consumer, "unmapped_guests_only", "true");
         }
 
-        Date now = new Date();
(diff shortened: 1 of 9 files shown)


awood:
:joy: ACK

PR #852, opened 2015-03-17
1201520: Addition of (Temporary) to pool description

wottop:
For use with pools for unmapped guests only flag

### DIFF
--- a/server/src/main/java/org/candlepin/resource/util/CalculatedAttributesUtil.java
+++ b/server/src/main/java/org/candlepin/resource/util/CalculatedAttributesUtil.java
@@ -52,7 +52,9 @@ public class CalculatedAttributesUtil {
 
         PoolComplianceType type = poolTypeRules.getPoolType(p);
         type.translatePoolType(i18n);
-        attrMap.put("compliance_type", type.getPoolType());
+        attrMap.put("compliance_type", type.getPoolType() +
+                ("true".equals(p.getAttributeValue("unmapped_guests_only")) ?
+                " " + i18n.tr("(Temporary)") : ""));
 
         if (c == null) {
             return attrMap;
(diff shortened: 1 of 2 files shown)


dgoodwin:
Need to rev rules.js version.

Each of these changes probably would have made a good individual commit. 

Investigating implications of the change from entitlement_id back to product_id in the response, I'm worried empty ID there may cause problems, and also how it's going to be displayed in the client just changed again, back to what we had before.

dgoodwin:
With no installed products:

$ curl -k -u admin:admin -X GET "https://localhost:8443/candlepin/consumers/40d62f6b-9c19-4364-968b-0c1c06ac4fc0/compliance"
{
  "date" : "2015-03-18T12:27:16.353+0000",
  "compliantUntil" : null,
  "nonCompliantProducts" : [ ],
  "compliantProducts" : { },
  "partiallyCompliantProducts" : { },
  "partialStacks" : { },
  "reasons" : [ {
    "key" : "UNMAPPEDGUEST",
    "message" : "Guest has not been reported on any host and is using a temporary unmapped guest subscription.",
    "attributes" : {
      "product_id" : "",
      "name" : "UNABLE_TO_GET_NAME"
    }
  } ],
  "compliant" : false,
  "status" : "partial"
}%                                        


[root@rhel7 product]# subscription-manager status
+-------------------------------------------+
   System Status Details
+-------------------------------------------+
Overall Status: Insufficient

UNABLE_TO_GET_NAME:
- Guest has not been reported on any host and is using a temporary unmapped guest subscription.


If you do have installed products, in the GUI, we now show yellow state but there's no way to see why. I guess temporary helps but it's tucked up in the details, you still see yellow system, but green products and entitlements. It's a bit tricky to identify why.

I am genuinely not sure what we need here, this has flipped or been attempted to flip about 4x now. We have to settle what and how this gets flagged. 

Based on our first fix discussion, phone call earlier this week, and feedback from Dan, I have to think the most logical interpretation thus far has been that this is not a problem with an installed product, it is a problem with the use of a subscription and/or the overall system.

PR #853, opened 2015-03-19
1199483: Entitlement date override not persisting during entitlement gen...

wottop:
...eration

Affects 24-hour entitlements as they have expiration dates that are not the
same as the pool


alikins:
s/experation/expiration in commit message.

otherwise ACK

PR #855, opened 2015-03-24
1204086: Disallow unmapped guests from attaching to future subscriptions

wottop:


server/src/test/java/org/candlepin/policy/js/entitlement/PreEntitlementRulesTest.java
```diff
@@ -665,6 +666,23 @@ public void unmappedGuestBadDate() {
...
+        consumer.setFact("virt.is_guest", "true");
+        consumer.setCreated(new Date());
+        ValidationResult result = enforcer.preEntitlement(consumer, pool, 1, CallerType.BIND);
+        assertTrue(result.hasErrors());
+        assertEquals(1, result.getErrors().size());
+        assertEquals("virt.guest.cannot.bind.future.unmapped.guest.pool",
+            result.getErrors().get(0).getResourceKey());
+    }
+
+    @Test
```
dgoodwin: +1 for unit testing this.

server/spec/unmapped_guest_spec.rb
```diff
@@ -233,4 +233,19 @@
...
+    pools = @guest1_client.list_pools :consumer => @guest1_client.uuid
+    pools.should have(1).things
+    pools[0].subscriptionId.should == @sub1.id
+
+    # just the future one, created in this method
+    pools = @guest1_client.list_pools :consumer => @guest1_client.uuid, :activeon => Date.today + 500
+    pools.should have(1).things
+    pools[0].subscriptionId.should == future_sub.id
+  end
 end
```
dgoodwin: Given the issues we're having with spec test runtime it's probably time to make sure we're focusing on unit tests much more, and rely on spec tests only for things where we need some knowledge that it's behaving correctly in real world. You've got this unit tested, lets drop the spec test.

PR #857, opened 2015-03-26
Awood/1204311

awood:
If a subscription has derived products and/or derived provided products, those objects should be promoted on a temporary pool.  E.g. the subscription's derived provided products become the provided products of the temporary pool.


awood:
This is failing the "One Sub Pool Per Stack should not regenerate certs on refresh pools when sub pool has not been changed" spec test.  I'll reopen when I get that figured out.

PR #858, opened 2015-03-26
Set Eclipse to use its own target directory.

awood:
When running simultaneous tasks with Buildr and Eclipse, all sorts of
problems pop up.  Buildr creates a class file and then Eclipse gets
bent out of shape about it or vice versa.  Generally the error messages
are around byte-code incompatibilities.

This patch gives Eclipse its own target directory that it can manage as
it sees fit.


wottop:
nice. ack.

PR #860, opened 2015-03-31
Fix SQL in guest ID checkin time upgrade script.

dgoodwin:
Script would fail if you had a host that had reported more than one guest ID.
Discovered much simpler query as all the data we need is technically already in
the cp_consumer_guests table, no need for the join.


dgoodwin:
Tested on both mysql and postgres each with a pre-existing db containing multiple hypervisors each with more than one guest ID reported. End result should be cp_guest_ids_checkin containing one row per hypervisor with updated/created timestamps matching the largest in cp_consumer_guests for that consumer_id.

PR #861, opened 2015-03-31
1207316: Fix host autobind for guest when subscription requires cert v3.

dgoodwin:
We had logic during a direct bind to special case a consumer of type hypervisor
    and allow them to get a V3 certificate. (as virt-who does not create
    hypervisors with a certificate_version fact) However we missed a similar case
    where we filter down the pools available to autobind from if the consumer
    can't handle a V3 certificate, and this location was not special casing
    for type hypervisor.
    
    Moved the cert v3 logic to a common location on Consumer model itself, and
    updated both locations to use it.
    
    Revamped some tests that started failing due to the change in logic, as they were
    incorrectly using mocks for a POJO, resulting in tests that were unnecessarily
    brittle.


wottop:
A couple of hits on checkstyle

wottop:
ack

PR #863, opened 2015-03-31
1207822: EnvironmentResource fix for duplicate content IDs

Ceiu:
- EnvironmentResource will now throw a ConflictException if an API
  user attempts to promote content that has already been promoted.
- Added spec tests to verify the above behavior

### DIFF
--- a/server/spec/environment_spec.rb
+++ b/server/spec/environment_spec.rb
@@ -77,14 +77,60 @@ describe 'Environments' do
     @env['environmentContent'].size.should == 1
   end
 
-  it 'cleans up env content when content is deleted' do
-    content = create_content
-    job = @org_admin.promote_content(@env['id'],
-        [{:contentId => content['id']}])
+  it 'does not allow content to be promoted more than once' do
+    content1 = create_content
+    content2 = create_content
+
+    job = @org_admin.promote_content(@env['id'], [{:contentId => content1['id']}])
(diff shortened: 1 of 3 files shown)


wottop:
Need to ensure that the change of enabled flag does not get rejected based on this new code.

server/spec/environment_spec.rb
```diff
@@ -77,14 +77,60 @@
     @env['environmentContent'].size.should == 1
   end
 
-  it 'cleans up env content when content is deleted' do
+  it 'does not allow content to be promoted more than once' do
```
wottop: Did the previous test get removed? Why?

Ceiu: Nah, just oddities with the diff. The cleanup test got moved down to line 118.

server/spec/environment_spec.rb
```diff
@@ -77,14 +77,60 @@
...
+      @env['id'], [{:contentId => content3['id']}, {:contentId => content4['id']}]
+    )
+    wait_for_job(job['id'], 15)
+
+    @env = @org_admin.get_environment(@env['id'])
+    @env['environmentContent'].size.should == 4
+  end
+
+  it 'cleans up env content when content is deleted' do
+    content1 = create_content
```
wottop: Didn't see it down here

wottop:
ack

PR #864, opened 2015-04-01
1207810: Add missing symlink for gutterball and c3p0.

dgoodwin:
Tested and found to contain the symlink, and allow gb to run on EL7 system experiencing the bug.


awood:
I suggest a new strategy, @dgoodwin: let the Wookiee win.

PR #866, opened 2015-04-01
Awood/1204311

awood:
Promote derived objects to first-class objects on temporary pools

### DIFF
--- a/server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
+++ b/server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
@@ -33,6 +33,7 @@ import org.candlepin.policy.js.ProductCache;
 
 import com.google.inject.Inject;
 
+import org.apache.commons.lang.BooleanUtils;
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -176,17 +177,18 @@ public class PoolRules {
             if (virtQuantity != null) {
                 Product poolProduct = sub.getProduct();
                 // Favor derived products if they are available
+                // Set the pool's product to the sub's derived product
                 if (sub.getDerivedProduct() != null) {
                     poolProduct = sub.getDerivedProduct();
                 }
 
-                Pool derivedPool = helper.createPool(sub, poolProduct.getId(),
+                Pool newPool = helper.createPool(sub, poolProduct.getId(),
                     virtQuantity, virtAttributes);
 
                 // Using derived here because only one derived pool
                 // is created for this subscription
-                derivedPool.setSourceSubscription(new SourceSubscription(sub.getId(), "derived"));
-                pools.add(derivedPool);
+                newPool.setSourceSubscription(new SourceSubscription(sub.getId(), "derived"));
+                pools.add(newPool);
             }
         }
         return pools;
@@ -237,7 +239,7 @@ public class PoolRules {
     /**
      * Refresh pools which have no subscription tied (directly) to them.
      *
-     * @param floatingPools ools with no subscription ID
+     * @param floatingPools pools with no subscription ID
      * @return pool updates
      */
     public List<PoolUpdate> updatePools(List<Pool> floatingPools) {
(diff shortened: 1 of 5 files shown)


dgoodwin:
No commits from wottop, is everything pushed?

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -176,17 +177,29 @@ private long calculateQuantity(Subscription sub) {
...
+                // Set the pool's product to the sub's derived product
                 if (sub.getDerivedProduct() != null) {
                     poolProduct = sub.getDerivedProduct();
                 }
 
-                Pool derivedPool = helper.createPool(sub, poolProduct.getId(),
+                Pool newPool = helper.createPool(sub, poolProduct.getId(),
                     virtQuantity, virtAttributes);
 
+                // Set the pool's provided products to the sub's derived provided products
```
dgoodwin: This is a little strange because these are being set above in the call to createPool. It would probably make more sense to do the same thing we are for poolProduct, and pass in the set of provided products. However the logic looks correct.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -176,17 +177,29 @@ private long calculateQuantity(Subscription sub) {
...
+                Set<ProvidedProduct> providedBySub = new HashSet<ProvidedProduct>();
+                if (sub.getDerivedProduct() != null) {
+                    if (sub.getDerivedProvidedProducts() != null) {
+                        for (Product p : sub.getDerivedProvidedProducts()) {
+                            providedBySub.add(new ProvidedProduct(p.getId(), p.getName(), newPool));
+                        }
+                    }
+                    newPool.setProvidedProducts(providedBySub);
+                }
+
```
dgoodwin: If I comment out this whole new block, all the tests still pass, any ideas?

server/src/test/java/org/candlepin/policy/PoolRulesTest.java
```diff
@@ -767,11 +774,12 @@ public void standaloneVirtLimitSubCreate() {
         assert ("true".equals(unmappedVirtPool.getAttributeValue("virt_only")));
         assert ("true".equals(unmappedVirtPool.getAttributeValue("unmapped_guests_only")));
 
-        assertProvidedProducts(s.getProvidedProducts(),
+        // The derived provided products of the sub should be promoted to provided products
+        // on the unmappedVirtPool
+        assertProvidedProducts(s.getDerivedProvidedProducts(),
```
dgoodwin: The change looks to hit both create and update pools, but the test appears to only be covering create. (and the test passes even if I comment out important parts of the patch.

dgoodwin:
I can't spot why and I'm hoping it's not something to do with the new eclipse workspace stuff, but I cannot make this test fail by commenting out changes, and the update path appears uncovered entirely. Ideally would like to see a test asserting unmapped guest pool with derived product data from createPools, then applying again via updatePools and asserting everything again.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
awood: Looks like we need to add a test then!  The reproducer case is

* deploy with test data
* Run psql -U candlepin candlepin -c "SELECT a.productid, a.id, b.subscriptionid, b.subscriptionsubkey  FROM cp_pool a, cp_pool_source_sub b WHERE a.id = b.pool_id AND a.productid like 'awesomeos-server-basic-%c' ORDER BY a.startdate DESC;"
* Refresh pools
* Run the same query

The results should be identical.  The test data has a lot of extra stuff in it, so I have a stripped down version that narrows it to the affected products.  Holler if you want it.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
awood: Good idea

awood:
Just pushed a new commit that unifies some code in PoolHelper.  I'll work with wottop tomorrow on getting the unit tests sorted.  The real heart of the bug was incorrect logic in getExpectedProvidedProducts.  The fix there should cover the update path.

dgoodwin:
Ok the logic does look correct to me as well, will do some manual testing and await a test fix.

dgoodwin:
Manual testing looks good, on initial test data deploy I see correct unmapped guest pools with just the 37060 provided product. A refresh pools for org admin leaves this unchanged. Also verified the end result after attaching to a guest, everything looks correct now.

Just a unit test fix + update addition I think.

PR #867, opened 2015-04-01
1110427: Hypervisor Resource expanded to allow fact enhancement for hype...

wottop:
...rvisor consumers


dgoodwin:
I think limiting this to updating facts might miss some requirements, if we look at what we have floating around, the hypervisor "system name" keeps popping up and folks want to be able to find them in Sat/Portal UI by the hostname like a regular system, rather than looking at UUIDs.

Could we restructure this more as some kind of "bulk consumer update"? Maybe even accept a list of consumers as the param and pass them to that resource method one at a time somehow? That might not work due to the UUIDs but basically something that would allow us to expand on it to other consumer properties that we can update normally, I could see installed products and capabilities both coming up in future.

dgoodwin:
Keep an eye on security as well.

wottop:
So you want to allow a full consumer update, but with the system UUID as the key? Just reroute to the ConsumerResource methods from the HypervisorResource?

dgoodwin:
That feels like the most forward thinking thing yeah, we already know they want hostnames, many other things could pop up as well. Like update consumer we leave what isn't specified so hopefully the requests don't become huge specifying a bunch of unused things.

Something else we should consider, why don't we fold hypervisor resource functionalityinto your new API, then virt-who only makes one call, and each host would include it's guest IDs. The old API effectively becomes obsolete. If we're going full consumer update I have to admit it doesn't seem to make sense to call the old one when we go modify virt-who to use this.

dgoodwin:
We also hit a problem in hosted where the updates with thousands of hypervisors and guests was taking over 60s thus timing out with the client. That could easily happen here, could we make this async?

mstead:
I like the idea of rolling all the functionality into a single API call that is async. hypervisorCheckIn currently calls updateConsumer IIRC -- no reason we couldn't add facts, installed products, or whatever to that as long as updateConsumer supported it.

mstead:
This leave a lot of flexibility going forward, and wouldn't require changes to get virt-who to store extra "consumery" bits of data later on.

mstead:
Would likely have to change the incoming mapping to something a little broader, but I think it is doable.

wottop:
Updated with async processing of full consumer data.

server/client/ruby/candlepin_api.rb
```diff
@@ -1227,6 +1236,11 @@ def post_text(uri, data=nil)
     return response.body
   end
 
+  def post_text_return_json(uri, data=nil)
```
dgoodwin: What's the benefit or difference for this method that requires a new mechanism for posting json?

server/spec/hypervisor_check_in_spec.rb
```diff
@@ -288,6 +288,20 @@ def check_hypervisor_consumer(consumer, expected_host_uuid, expected_guest_ids)
...
+    virtwho = create_virtwho_client(@user)
+    json ="{\"hypervisors\":" +
+          "[{" +
+          "\"name\" : \"hypervisor_999\"," +
+          "\"hypervisorId\" : {\"hypervisorId\":\"uuid_999\"}," +
+          "\"guestIds\" : [{\"guestId\" : \"guestId_1_999\"}]" +
+          "}]}";
+    job_detail = JSON.parse(virtwho.hypervisor_update(@owner['key'], json))
+    job_detail['state'].should == 'CREATED'
+    job_detail['id'].should start_with('hypervisor_update_')
```
dgoodwin: Should probably test success of the job here.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -0,0 +1,201 @@
...
+                    Consumer knownHost = consumerCurator.getHypervisor(hypervisorId, owner);
+                    if (knownHost == null) {
+                        if (!create) {
+                            throw new Exception("Unable to find hypervisor with id " +
+                                                hypervisorId + " in org " + ownerKey);
+                        }
+                        log.info("Registering new host consumer for hypervisor ID: {}", hypervisorId);
+                        consumerResource.create(hypervisor, principal, null, owner.getKey(), null);
+                    }
+                    else {
```
dgoodwin: I would log to match the create case here.

server/spec/hypervisor_check_in_spec.rb
dgoodwin: Admittedly I know I've been complaining about too much spec testing here, but in this case we've added a new API call that's the way forward, and left the bulk of the testing using the old one. As this stands we have a bunch of tests using the wrong API. I would look to abstract the update method used in the rspec tests, have one class using the old API, and one using the new, but then *share* the tests, no copy paste. Test coverage of the actual virt update functionality is pretty important here.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -0,0 +1,201 @@
...
+            String json = decompress(data);
+            HypervisorList hypervisors = (HypervisorList) Util.fromJson(json, HypervisorList.class);
+            log.info("Hypervisor consumers for create/update: " + hypervisors.getHypervisors().size());
+
+            for (Consumer hypervisor : hypervisors.getHypervisors()) {
+                if (hypervisor.getHypervisorId() != null &&
+                        hypervisor.getHypervisorId().getHypervisorId() != null) {
+                    String hypervisorId = hypervisor.getHypervisorId().getHypervisorId();
+                    hypervisor.setType(new ConsumerType(ConsumerTypeEnum.HYPERVISOR));
+                    Consumer knownHost = consumerCurator.getHypervisor(hypervisorId, owner);
```
dgoodwin: Big performance issues here. We recently worked to fix this in the old API call as the long transactions causes big deadlock and timeout issues. You can see this in the old API call where we lookup hypervisorConsumersMap, guestConsumersMap, and guestHypervisorConsumers. 

You'll see guestHypervisorConsumers eventually passed to consumerResource.performConsumerUpdates(). I did not add it to update consumer resource method but it looks like we might need to here. Not passing this through is likely resulting in one lookup per guest ID as well, so thousands and thousands of DB queries whereas the old API was doing ~4.

I would be very worthwhile to generate a JSON file or something for use with virt-whom and test this with about 2000 hypervisors each with 10-15 guest IDs, just as a reasonable real world scenario. If we want to plan for future I'd bump that up a lot. I'd suggest doing this before making the performance changes so you can see them taking effect.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -0,0 +1,201 @@
...
+                    if (knownHost == null) {
+                        if (!create) {
+                            throw new Exception("Unable to find hypervisor with id " +
+                                                hypervisorId + " in org " + ownerKey);
+                        }
+                        log.info("Registering new host consumer for hypervisor ID: {}", hypervisorId);
+                        consumerResource.create(hypervisor, principal, null, owner.getKey(), null);
+                    }
+                    else {
+                        consumerResource.updateConsumer(knownHost.getUuid(), hypervisor);
```
dgoodwin: A little worried here we're missing functional parity with the old API call, but if we do the above proposal for rspec (running the hypervisor resource spec against both API calls to update), it should tell us.

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -186,6 +190,47 @@ else if (guestIdsUpdated) {
...
+        @PathParam("owner") @Verify(value = Owner.class,
+            require = Access.READ_ONLY,
+            subResource = SubResource.HYPERVISOR) String ownerKey,
+        @QueryParam("create_missing") @DefaultValue("true") boolean createMissing) {
+
+        log.info("Hypervisor update by principal: " + principal);
+        Owner owner = this.getOwner(ownerKey);
+
+        return HypervisorUpdateJob.forOwner(owner, hypervisorJson, createMissing, principal);
+    }
```
dgoodwin: We need some way for virt-who to check which API call it can make, typically I think we're using GET / and whether or not that container something we're interested in. A capability might also make sense but we don't do that yet afaik.

server/spec/hypervisor_check_in_spec.rb
```diff
@@ -288,6 +288,20 @@ def check_hypervisor_consumer(consumer, expected_host_uuid, expected_guest_ids)
...
 
+  # simple end to end test
+  it 'should allow virt-who to report hypervisor data' do
+    virtwho = create_virtwho_client(@user)
+    json ="{\"hypervisors\":" +
+          "[{" +
+          "\"name\" : \"hypervisor_999\"," +
+          "\"hypervisorId\" : {\"hypervisorId\":\"uuid_999\"}," +
+          "\"guestIds\" : [{\"guestId\" : \"guestId_1_999\"}]" +
+          "}]}";
```
dgoodwin: We should test with some facts in there too, the major ones like sockets and such.

dgoodwin:
And of course once we're ready, we should start prepping virt-who to use this thing, that works going to be on us as well.

server/client/ruby/candlepin_api.rb
wottop: The existing post_text only accepts text a s a return. I am returning json, which blew the old method up.

dgoodwin: We should probably just pass in the response type as a method param and have it default to the accept/text. Another optional arg similar to data=nil.

server/spec/hypervisor_check_in_spec.rb
```diff
@@ -295,4 +301,295 @@ def create_virtwho_client(user)
...
   end
+
+  # async version of method calls
+
+  def get_async_host_guest_mapping(hypervisor_id, guest_id_list)
+      guestIds = []
+      guest_id_list.each do |guest|
+          guestIds << {'guestId' => guest}
+      end
+      json = {"hypervisors" => ["name" => hypervisor_id, "hypervisorId" => {"hypervisorId" => hypervisor_id}, "guestIds" => guestIds]}
```
dgoodwin: One of the drivers for this was to report hypervisor hostnames. In the spec tests we're just using the hypervisor ID as the hostname, however this is exactly the same behaviour as the old API call, ideally we should have test coverage on actual hostname reporting that can't be confused with a false positive when no hypervisor hostname is specified.

server/spec/hypervisor_check_in_spec.rb
```diff
@@ -295,4 +301,295 @@ def create_virtwho_client(user)
...
+      job_detail = JSON.parse(virtwho1.hypervisor_update(@owner2['key'], host_mapping))
+    end.should raise_exception(RestClient::ResourceNotFound)
+  end
+
+  it 'should raise bad request exception if mapping was not provided [async]' do
+    virtwho = create_virtwho_client(@user)
+    lambda do
+      virtwho.hypervisor_update(@owner['key'], nil)
+    end.should raise_exception(RestClient::BadRequest)
+  end
```
dgoodwin: Really not great to see the amount of copy paste in here. At the very least could we move the async tests to directly below their non-async counterparts, so at least when something needs an edit, we can see both spots on screen at once.

I still think with a bit more work we could abstract out the API call dependent parts into methods, and sub-class the test suite for async tests that just implement those methods. (i.e. a method for setup that takes common args, and methods to verify the result json that take common args)

If you say it's not feasible a second time I'll believe you. :) It's just a LOT of copy paste.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -205,10 +214,23 @@ public VirtConsumerMap getGuestConsumersMap(Owner owner,
...
 
         Query q = currentSession().createSQLQuery(sql);
-        q.setParameterList("guestids", possibleGuestIds);
-        q.setParameter("ownerid", owner.getId());
-        List<String> consumerUuids = q.list();
+        List<String> consumerUuids = new ArrayList<String>();
+
+        int fromIndex = 0;
+        int toIndex = fromIndex + MAX_IN_QUERY_LENGTH;
+        int count = 0;
```
dgoodwin: count is unused now. 

Looks good otherwise.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -699,4 +745,88 @@ public int getConsumerEntitlementCount(Owner owner, ConsumerType type) {
             .setResultTransformer(Criteria.DISTINCT_ROOT_ENTITY)
             .list();
     }
+
+    /**
+     * InExpressionIgnoringCase
+     *
+     * Allows Criterion for case insensitive comparison to list of values
```
dgoodwin: Cool.

server/src/main/java/org/candlepin/model/Status.java
```diff
@@ -37,7 +37,7 @@
     private boolean standalone;
     private Date timeUTC;
     private String[] managerCapabilities = {"cores", "ram", "instance_multiplier",
-        "derived_product", "cert_v3", "guest_limit", "vcpu"};
+        "derived_product", "cert_v3", "guest_limit", "vcpu", "async_virt_who"};
```
dgoodwin: Lets not box ourselves in on virt-who, that's a specific client, but we're offering an API for whatever client. Suggest "hypervisors_async"?

What do others think about this one? This is a departure from what we'd normally do which is add a new entry in GET / for a supported resource, similar to "guestids" there now. However it's a little weird in this case, not sure what the "rel" name would be as "hypervisors" is already there.

I'm leaning towards using the capability.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -0,0 +1,270 @@
...
+            JobDataMap map = context.getMergedJobDataMap();
+            String ownerKey = map.getString(JobStatus.TARGET_ID);
+            Boolean create = map.getBoolean(CREATE);
+            Principal principal = (Principal) map.get(PRINCIPAL);
+
+            HypervisorUpdateResult result = new HypervisorUpdateResult();
+
+            Owner owner = ownerCurator.lookupByKey(ownerKey);
+            if (owner == null) {
+                context.setResult("Nothing to do. Owner does not exist");
```
dgoodwin: Logging, that should not be happening quietly.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -0,0 +1,270 @@
...
+                return;
+            }
+            byte[] data = (byte[]) map.get(DATA);
+            String json = decompress(data);
+            HypervisorList hypervisors = (HypervisorList) Util.fromJson(json, HypervisorList.class);
+            log.info("Hypervisor consumers for create/update: " + hypervisors.getHypervisors().size());
```
dgoodwin: Lets log the owner key here too for the sake of splunk debugging. "Updating hypervisor consumers for org {}", I might also log a quick note on the amount of data in the update.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -0,0 +1,270 @@
...
+
+    /*
+     * Create a new hypervisor type consumer to represent the incoming hypervisorId
+     */
+    private Consumer createConsumerForHypervisorId(String incHypervisorId,
+            Owner owner, Principal principal) {
+        Consumer consumer = new Consumer();
+        consumer.setName(incHypervisorId);
+        consumer.setType(new ConsumerType(ConsumerTypeEnum.HYPERVISOR));
+        consumer.setFact("uname.machine", "x86_64");
```
dgoodwin: This doesn't look right.

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -186,6 +190,53 @@ else if (guestIdsUpdated) {
...
+     * @httpcode 202
+     * @httpcode 200
+     *
+     */
+    @POST
+    @Consumes(MediaType.TEXT_PLAIN)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Transactional
+    @Path("/{owner}")
+    public JobDetail updateOwnersHypervisors(
```
dgoodwin: Lets mark the old method deprecated, and could you name them somewhat similar so it's a bit more clear that they're the same functionality just a newer version. Something like hypervisorUpdate and hypervisorUpdateAsync perhaps.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
wottop: This was how it was done in the original. If we don't need or want this I will remove from both. Perhaps it is a placeholder or initialization for the facts.

PR #869, opened 2015-04-06
Pool types are now stored in the DB

Ceiu:
- Added a new column, "type," to cp_pool
- The PoolType is saved to the DB when a Pool instance is persisted
  or updated. The value stored is the name of the enum returned
  by the getType() method at the time of persistence
- Added an upgrade task which adds the new column and updates existing
  pools by setting their perceived type


server/src/main/java/org/candlepin/liquibase/PoolTypeUpgradeTask.java
```diff
@@ -0,0 +1,196 @@
...
+            "  WHERE PA1.id IS NULL AND PA2.id IS NULL" +
+            ")"
+        );
+
+        // Commit & restore original autocommit state
+        this.connection.commit();
+        this.connection.setAutoCommit(autocommit);
+    }
+
+}
```
dgoodwin: This looks like just straight SQL statements, could this just be a regular liquibase update without the custom class?

Ceiu: It certainly can -- I hadn't even considered that option.

awood:
add to whitelist

dgoodwin:
Apologies I think I found an issue with the upgrade after merging. I ran it against a real db from some internal testing and ended up with unmapped_guest_only pools showing up as type BONUS. 

http://pastebin.test.redhat.com/274912

PR #870, opened 2015-04-06
1204962: remove throw/catch/eat of runtime exception in bind

wottop:
Submitted as a pull request, but more like experimental theatre.

The criteria is to ensure identical REST output while removing stack-trace logging for non-fatal conditions. This required removing a few thrown exceptions and then adjusting the code stack back to the resource.

All suggestions are welcome.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -474,7 +474,7 @@ public class CandlepinPoolManager implements PoolManager {
             for (ConsumerInstalledProduct cip : consumer.getInstalledProducts()) {
                 fullList.add(cip.getId());
             }
-            log.warn("No entitlements available for products: " + fullList);
+            log.info("No entitlements available for products: " + fullList);
             return null;
         }
 
@@ -525,7 +525,7 @@ public class CandlepinPoolManager implements PoolManager {
             entitleDate, owner, null, possiblePools);
 
         if (bestPools == null) {
-            log.warn("No entitlements for host: " + host.getUuid());
+            log.info("No entitlements for host: " + host.getUuid());
             return null;
         }
 
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -26,6 +26,7 @@ import org.candlepin.common.paging.PageRequest;
 import org.candlepin.config.ConfigProperties;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.ConsumerCurator;
+import org.candlepin.model.ConsumerInstalledProduct;
 import org.candlepin.model.Entitlement;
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -468,8 +469,13 @@ public Pool find(String poolId) {
...
 
         if (bestPools == null) {
-            throw new RuntimeException("No entitlements for products: " +
-                Arrays.toString(productIds));
+            List<String> fullList = new ArrayList<String>();
+            fullList.addAll(Arrays.asList(productIds));
+            for (ConsumerInstalledProduct cip : consumer.getInstalledProducts()) {
+                fullList.add(cip.getId());
+            }
+            log.warn("No entitlements available for products: " + fullList);
```
dgoodwin: Why did the consumer installed products come into play here? I think the list of productIds is already set to what was requested, which would normally be their installed products.

dgoodwin: I might even take that just down to INFO or DEBUG.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1478,20 +1477,9 @@ public Response bind(
...
-            }
-            catch (CertVersionConflictException cvce) {
-                throw cvce;
-            }
-            catch (RuntimeException re) {
-                log.error("Autobind error", re);
-            }
+            AutobindData autobindData = AutobindData.create(consumer).on(entitleDate)
+                    .forProducts(productIds).withPools(fromPools);
+            entitlements = entitler.bindByProducts(autobindData);
```
dgoodwin: Why did CertVersionConflictException and ForbiddenException disappear here?

dgoodwin:
Overall it looks good, returning null makes a lot more sense than runtime exception in the context of a select best pools.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
wottop: The bind is across the installed products plus the ones passed to the method. This will make a complete list for the logging.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: As we catch and throw without even logging, there is no functional reason for this. Is it a placeholder in the code?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
dgoodwin: Ah it switches them out lower, I would have thought that would be the correct list by this point. ACK.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
dgoodwin: Ahh I follow, ty.

PR #871, opened 2015-04-07
Address several gutterball startup/shutdown issues

mstead:
1) Establish QPID connection in a separate thread to avoid delay in bringing up the gutterball report API and stalling deployment of candlepin.
2) Properly close the gutterball context: shut down qpid connection
3) Adjusted the default retry configuration values for connecting to qpid (1h was way too long)


dgoodwin:
Looks good. ACK.

PR #872, opened 2015-04-07
Alikins/1193803 manifest import speed [DO NOT MERGE]

alikins:
Wip/notes/etc on manifest import for reference.

https://bugzilla.redhat.com/show_bug.cgi?id=1193803

Before:

sat6 client timeout importing a manifest with 
employee skus.

The ruby client times out after about 2 minutes, but the
request itself continues for ~5 minutes. 

~2 minutes import Contents and Products (about a second each)
~2 minutes generating new pools and subs

The tomcat thread running this was using 100% of one cpu.
Most of the time spent was in hibernate, doing a huge amount
of object comparisons to determine if everything in the
large transaction is up to date. 
See
http://stackoverflow.com/questions/10143880/hibernate-queries-much-slower-with-flushmode-auto-until-clear-is-called
for one description of this hibernate issue.

Now:

Takes about 12-14 seconds to do the manifest import.

Changes:
- Build list of uniq contents and products to import, then import them.
  Elliminates some unneeded duplication, and skips a lot of create or
  update logic. Saves 10-20 seconds. (This may be obsoleted by the
  COMMIT change)

- After creating new objects, refresh the session so the latest version
  is included in the session cache. This eliminates a lot of the update
  checking from the auto flush, and improves content/product import
  import alot. The product/content import stage goes from ~120 seconds
  to about 1-2 seconds total.

- Grab a reference to the EntityManager for the session and call
  setFlushMode(FlushModeType.COMMIT) on it. This prevents hibernate
  needing to constantly check if every object is up to date, and waits
  until transaction commit to "flush". This cuts the pool/sub creation
  and refresh portion from 2-3 minutes to approximately 10 seconds total.

- Experiemental: Remove the poolCurator.flush() in
  CandlepinPoolManager.processPoolUpdates(). The flush here seems to
  defeat some of the gain of using COMMIT flush mode, and removing it
  seems to work and gains a few seconds on the above (~16 secs to ~12
  secs), but is likely needed for prod or more active systems.

- Experimental: The hashCode for Product was changed to not depend 
  entirely on the product id, as per
  http://docs.jboss.org/hibernate/orm/4.2/manual/en-US/html/ch13.html#transactions-basics-identity
  This seems to worth a little performance when apply over the above
  (maybe a seconds worth in clock time, but more noticable in profile
  data). But I'm unsure if this would break compatability.

- Unaddressed potential improvements:

  - With the import down to ~12 seconds, the time spent creating json
  objects for the event queue become significant. Mostly in
  EventFactory.entityToJson(). Since that is done before the object is
  sent to the message bus and not async, the manifest import process
  blocks on it. For the kind of large subscriptions a employee sku
  manifest creates, that accounts for ~4secs of the ~12 seconds of a
  manifest import.

  - Some of these changes maybe applicable to the general refresh pools
  case as well.

- Unaddressed potential issues

  - Most testing was done with fresh db, and one manifest import.
    Updating an import, or concurrent imports were untested.

- General oddities

  - manifest import, and pool refresh/manager do a lot of slightly
    unusual thins in efforts to fight various symptons. It's hard to
    tell if there are interactions. It may be worth attempt to remove
    all of those fixes, and seeing if there is a better fundamental fix.
    (thinking of things like the poolCurator.refresh() in
    CandlepinPoolManager, and poolCurator.lockAndLoad(), the various
    session refresh()'s I added, etc).

  - On manifest import [POST /owners/{owner_key}/import] the request
    itself is marked @NonTransaction, but the transaction that is used
    results from the call to OwnerResource.findOwners() [via
    auth/principal creation?] which seems odd. Importer.importObjects()
    is @Transactional, but the request has already created a trx by
    then. Not sure I understand which interceptor should be closing it.
    (It does seem to work fine, just seems odd).

  - Creating the Event objects seems to take ~100ms on average, and as
    long as ~500ms for some large objects (like the Subscription
    mentioned above). Thats all created before queueing the Event, so it
    is a blocking process, and adds to the total request time.

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -1257,6 +1257,7 @@ class Candlepin
 
   def create_basic_client(username=nil, password=nil)
     @client = RestClient::Resource.new(@base_url,
+                                       :timeout => -1,
                                        :user => username, :password => password,
                                        :headers => {:accept_language => @lang})
   end
@@ -1267,6 +1268,7 @@ class Candlepin
     @uuid = @identity_certificate.subject.to_s.scan(/\/CN=([^\/=]+)/)[0][0]
 
     @client = RestClient::Resource.new(@base_url,
+                                       :timeout => -1,
                                        :ssl_client_cert => @identity_certificate,
                                        :ssl_client_key => @identity_key,
                                        :headers => {:accept_language => @lang})
@@ -1275,6 +1277,7 @@ class Candlepin
   def create_trusted_consumer_client(uuid)
     @uuid = uuid
     @client = RestClient::Resource.new(@base_url,
+                                       :timeout => -1,
                                        :headers => {"cp-consumer" => uuid,
                                                     :accept_language => @lang}
                                         )
--- a/server/src/main/java/org/candlepin/audit/EventBuilder.java
+++ b/server/src/main/java/org/candlepin/audit/EventBuilder.java
@@ -38,7 +38,7 @@ public class EventBuilder {
     private final EventFactory factory;
 
     private Event event;
-    
+
     private static Logger log = LoggerFactory.getLogger(EventBuilder.class);
 
     public EventBuilder(EventFactory factory, Target target, Type type) {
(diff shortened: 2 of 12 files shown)


dgoodwin:
retest this please

dgoodwin:
Closing due to  performance gains from per-org products.

PR #873, opened 2015-04-08
 Updated the method used to determine pool types

Ceiu:
- Pool.getType now only reports UNMAPPED_GUEST iff the pool has both
  the "pool_derived" and "unmapped_guests_only" attribute.
- Pool type DB migration queries and pool tests have been updated to
  reflect the change to Pool.getType.


dgoodwin:
Looks good thanks!

PR #875, opened 2015-04-10
Add Candlepin sub-projects as dependencies in POM files.

awood:
Previously the code was only adding "Buildr::Artifact" instances to the
POM files, but sub-projects are of the type "JarTask".  What we
really needed to do was filter out instances of "Rake::FileTask".  Note
that Buildr::Artifact is descended from Rake::FileTask so we need to do
an exact match on the class type using instance_of.


dgoodwin:
retest this please

PR #876, opened 2015-04-10
Moved queries back to custom task to address issue with MySQL

Ceiu:
- The update queries have been moved back to a custom upgrade task
  to address an issue MySQL has with executing subqueries on a table
  during an update operation


dgoodwin:
retest this please

dgoodwin:
Looks good, and your original changeset did not make it into the 0.9.49 build so changing this checksum should be ok for everyone but devels. Lets keep an eye out for anyone complaining of changed checksums and let them know they must re-deploy, or hackaround the checksum in their db.

PR #877, opened 2015-04-13
Added MySQL docker container

Ceiu:


dgoodwin:
re-test this please

dgoodwin:
This is really great, very good investment, have been wanting a mysql docker container for awhile.

Couple things, running the container seems to be installing some qpid packages, this should probably be done in the build step instead.

installing Qpid
Package qpid-cpp-server-0.26-13.fc20.x86_64 already installed and latest version
Package qpid-cpp-server-0.26-13.fc20.x86_64 already installed and latest version
Resolving Dependencies
--> Running transaction check
---> Package qpid-cpp-server-store.x86_64 0:0.26-13.fc20 will be installed
--> Processing Dependency: db4 for package: qpid-cpp-server-store-0.26-13.fc20.x86_64
--> Processing Dependency: libdb_cxx-4.8.so()(64bit) for package: qpid-cpp-server-store-0.26-13.fc20.x86_64
---> Package qpid-tools.noarch 0:0.26-13.fc20 will be installed
--> Processing Dependency: python-qpid >= 0.8 for package: qpid-tools-0.26-13.fc20.noarch
--> Processing Dependency: python-qpid-qmf for package: qpid-tools-0.26-13.fc20.noarch
--> Running transaction check
---> Package libdb4.x86_64 0:4.8.30-11.fc20 will be installed
---> Package libdb4-cxx.x86_64 0:4.8.30-11.fc20 will be installed
---> Package python-qpid.noarch 0:0.26-2.fc20 will be installed
--> Processing Dependency: python-qpid-common = 0.26-2.fc20 for package: python-qpid-0.26-2.fc20.noarch
---> Package python-qpid-qmf.x86_64 0:0.26-2.fc20 will be installed
--> Processing Dependency: qpid-qmf(x86-64) = 0.26-2.fc20 for package: python-qpid-qmf-0.26-2.fc20.x86_64
--> Processing Dependency: libqmf2.so.1()(64bit) for package: python-qpid-qmf-0.26-2.fc20.x86_64
--> Running transaction check
---> Package python-qpid-common.noarch 0:0.26-2.fc20 will be installed
---> Package qpid-qmf.x86_64 0:0.26-2.fc20 will be installed
--> Finished Dependency Resolution


I'm not 100% sure it's working for me, if I try to run the spec tests via the commands from the README:

$ docker run -i -P -t --rm candlepin-mysql cp-test -r
...
(in /candlepin, development)
Downloading junit:junit:pom:4.12-beta-2
Downloading http://central.maven.org/maven2/junit/junit/4.12-beta-2/junit-4.12-beta-2.pom
junit-4.12-beta-2.pom: 100% |.............................................................................................|   23.1KB/23.1KB Time: 00:00:00
Downloading pl.pragmatists:JUnitParams:pom:1.0.3
Downloading http://central.maven.org/maven2/pl/pragmatists/JUnitParams/1.0.3/JUnitParams-1.0.3.pom
JUnitParams-1.0.3.pom: 100% |..............................................................................................|    5.6KB/5.6KB Time: 00:00:00
Downloading org.sonatype.oss:oss-parent:pom:9
Downloading http://central.maven.org/maven2/org/sonatype/oss/oss-parent/9/oss-parent-9.pom
oss-parent-9.pom: 100% |...................................................................................................|    6.4KB/6.4KB Time: 00:00:00
Cleaning candlepin
Completed in 7.959s

And it ends there.


This is a bit outside the scope of this but if you have time, could you give a look at the -s option? This used to leave a shell open for you when it was done doing everything else, it doesn't seem to work anymore on any container. It might actually be time for a re-think to be honest, Since this was written we now have docker exec, which we can use to run bash inside an already running container. I almost think -s should be removed, and we should make sure we're setup to properly daemonize so we could just launch the container with the service running and then exec our way in.

Lastly should update the README to note mysql as an option as well.

Ceiu:
I've run the container a few times now and I don't see the qpid stuff installing on run.

Regarding the -s flag, I'm still pretty green when it comes to docker, but I can see a use for the flag. When testing the database migrations, I've been starting a new container, running bash, then running cp-test to set the initial state, checking out the new branch to test, then doing a deploy from the candlepin-dev directory, killing the container after running the migration test. I could also do that by starting the container as a daemon and then exec'ing into it; but it feels like it'd be more work to do that (start, enter, exit, kill vs start+enter, exit+kill).

dgoodwin:
My errors looked to have been caused by a half baked local candlepin-base that was getting picked up and breaking things. Rebuilding base and mysql with --no-cache fixed the issue.

Looks good to me now, just doing some more testing, and will push a patch to include some fixes for the rebuild script.

PR #883, opened 2015-04-30
Added the PopulateHostedDB task and API call

Ceiu:
- The AdminResource now has a new API call, "pophosteddb," for
  populating the database in a hosted environment with product and
  content data
- The new API call runs as an asynchronous job performed by the
  PopulateHostedDBTask class

### DIFF
--- a/server/src/test/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTaskTest.java
+++ b/server/src/test/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTaskTest.java
@@ -18,11 +18,9 @@ import static org.mockito.Matchers.*;
 import static org.mockito.Mockito.*;
 
 import org.candlepin.model.Content;
-import org.candlepin.model.ContentCurator;
 import org.candlepin.model.PoolCurator;
 import org.candlepin.model.Product;
 import org.candlepin.model.ProductContent;
-import org.candlepin.model.ProductCurator;
 import org.candlepin.service.ProductServiceAdapter;
 import org.candlepin.test.DatabaseTestFixture;
 import org.candlepin.test.TestUtil;
@@ -43,7 +41,6 @@ import java.util.Map.Entry;
 public class PopulateHostedDBTaskTest extends DatabaseTestFixture {
 
     @Test
-    @SuppressWarnings("checkstyle:methodlength")
     public void testExecute() throws Exception {
         // Setup
         JobExecutionContext jec = mock(JobExecutionContext.class);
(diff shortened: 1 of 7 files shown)


dgoodwin:
retest this please

server/src/test/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTaskTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Test
+    public void testExecute() throws Exception {
+        // Setup
+        JobExecutionContext jec = mock(JobExecutionContext.class);
+
+        ContentCurator contentCurator = mock(ContentCurator.class);
+        PoolCurator poolCurator = mock(PoolCurator.class);
+        ProductCurator productCurator = mock(ProductCurator.class);
+        ProductServiceAdapter psa = mock(ProductServiceAdapter.class);
```
dgoodwin: If I had to guess I'd imagine this is virtually impossible for us to actually test locally against a real DB. Is that right?

If this is going to go wrong it'll be the usual slew of hibernate detached entity errors and such, it seems like it might be good to make this an actual DB test (extend DatabaseTestFixture), use real curators for saving, but mock the adapters and the product ID list we're going to import. Then we're closely simulating prod and we get some certainty that the job will actually get the data into the db. Would that be possible?

Ceiu:
I've modified the test to no longer mock the product/content curators and instead verify the created products are present in the database after the task is completed.

server/src/test/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTaskTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+        LinkedList<Product> dependentProducts2b = new LinkedList<Product>();
+        dependentProducts2b.add(dp3);
+        dependentProducts2b.add(dp4);
+
+        when(psa.getProductsByIds(productIds)).thenReturn(products);
+        when(psa.getProductsByIds(dependentProductIds1)).thenReturn(dependentProducts1);
+        when(psa.getProductsByIds(dependentProductIds1b)).thenReturn(dependentProducts1b);
+        when(psa.getProductsByIds(dependentProductIds2)).thenReturn(dependentProducts2);
+        when(psa.getProductsByIds(dependentProductIds2b)).thenReturn(dependentProducts2b);
```
dgoodwin: How about a mockGetProductsByIds(String [] pids, Product ... products), would take out a chunk of lines and probably get rid of the need for checkstyle suppression.

dgoodwin: I can't quite spot why there are 5 collections of product IDs, wouldn't it just be one call to the job, one call for all product IDs?

server/src/test/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTaskTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+        HashSet<String> productIds = new HashSet<String>();
+        productIds.add("prod1");
+        productIds.add("prod2");
+        productIds.add("prod3");
+
+        HashSet<String> dependentProductIds1 = new HashSet<String>();
+        dependentProductIds1.add("dprod1");
+        dependentProductIds1.add("dprod2");
+        dependentProductIds1.add("prod3");
```
dgoodwin: What's the distinction between a dprod and a prod? And why do they get intermingled in this one set?

server/src/test/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTaskTest.java
```diff
@@ -0,0 +1,187 @@
...
+        // Test
+        PopulateHostedDBTask task = new PopulateHostedDBTask(
+            psa, this.productCurator, this.contentCurator, poolCuratorSpy
+        );
+
+        task.execute(jec);
+
+        // Verify
+        // verify(productCuratorSpy, times(7)).createOrUpdate(any(Product.class));
+        // verify(contentCuratorSpy, times(12)).createOrUpdate(any(Content.class));
```
dgoodwin: Can remove.

server/src/test/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTaskTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+
+
+/**
+ * PopulateHostedDBTaskTest
+ */
+public class PopulateHostedDBTaskTest extends DatabaseTestFixture {
+
+    @Test
+    @SuppressWarnings("checkstyle:methodlength")
```
dgoodwin: Ideally would like to see this only used if absolutely necessary.

server/src/test/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTaskTest.java
Ceiu: That whole thing is to make sure it doesn't keep looping around asking for the same products repeatedly.
Related to your question above, the five collections are to support the three collections I want, as well as two extra collections that will be used only if it's not filtering products it's already seen.

dgoodwin:
Alright I sincerely hope this works against IT's db, hopefully they can test it in a devel/stage environment for us, we may have to tweak it for them a bit but it looks promising as is.

PR #884, opened 2015-04-30
Fix unmapped guest pool clean on import undo.

dgoodwin:


Ceiu:
Looks good. Spec and unit tests are all passing now.

ACK

PR #885, opened 2015-05-04
Apply permissions while getting Job Status

mstead:
Allow JobStatus to be fetched by a regular user
or consumer if they have the appropriate permission
to do so.

Permissions will be applied such that:
  * a super admin can view any job status
  * a user can view a job status if they have some level of access
    to an org (no matter who started the job)
  * a consumer can only view the status of jobs they have started

### DIFF
--- a/server/spec/job_status_spec.rb
+++ b/server/spec/job_status_spec.rb
@@ -6,6 +6,7 @@ describe 'Job Status' do
 
   before(:each) do
     @owner = create_owner(random_string("test_owner"))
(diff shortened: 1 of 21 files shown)


mstead:
@wottop once this is merged, you will have to make a small adjustment to the way you set up JobDetail for the hypervisor check-in Job.

You'll have to add:
jobDetail.put(JobStatus.OWNER_ID, ownerId);

wottop:
ack

PR #887, opened 2015-05-08
Fix significant subscription problems in multi-tenant deployment when products change.

dgoodwin:
See commit details.


Ceiu:
ACK

Importing manifests for two orgs with "conflicting" product information now results in the proper number of product IDs in both pools and subscriptions both before and after each. No test or rspec errors.

PR #890, opened 2015-05-18
Added a data/migration validation utility

Ceiu:
- Added the "cmv" tool to server/bin. This tool can be used to create
  snapshots for one or more orgs and then validate a deployment against
  a previous snapshot at a later date
- Fixed a few typos and cryptic bits of code in cpc

### DIFF
--- a/server/bin/cmv
+++ b/server/bin/cmv
@@ -5,6 +5,11 @@ require File.expand_path('candlepin_api', File.dirname(__FILE__) + '/../client/r
 require 'optparse'
 require 'set'
 
+$field_filter = {
+    'include' => [],
+    'exclude' => []
+}
+
 
 class CMVCommand
     def initialize(help=nil)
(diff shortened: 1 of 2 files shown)


komidore64:
protip: don't use `and` and `or` in conditional statements or boolean logic. `and` and `or` operate differently from `&&` and `||`. Always stick with `&&` and `||`.

komidore64:
you're doing some unneeded work in all the places where you're checking for `if !some_object.nil?`. Anything other than `nil` or `false` evaluates to `true` in a `if some_object` conditional, so you can just say `if some_object`.

for instance:
```ruby
def check(obj)
  if obj
    true
  else
    false
  end
end

check(42) # => true
check("42") # => true
check(0) # even zero! => true
check('') # => true
check(nil) # => false
check(false) # => false
```

server/bin/cmv
```diff
@@ -0,0 +1,446 @@
...
+            'orgs.pools.derivedProductAttributes' => 'name',
+            'orgs.subscriptions' => 'id',
+        }
+
+        exclude = [
+            'orgs.pools.productAttributes.updated',
+            'orgs.products.attributes.updated',
+            'orgs.products.updated',
+            'orgs.products.productContent.content.updated'
+        ]
```
dgoodwin: Are we going to exclude the things we know will appear different? (href, owner links on products, etc) Ideally I'd like to see it run without args and return that everything is fine. Although I guess that does influence the generic nature of this a bit...

server/bin/cmv
```diff
@@ -0,0 +1,446 @@
+#!/usr/bin/env ruby
```
dgoodwin: Could you add a little documentation here at the top just to clarify what this is for anyone who looks in future, cmv could be a bit ambiguous.

PR #891, opened 2015-05-20
Awood/mead

awood:
Changes required to allow Mead building for Candlepin.

I can give detail instructions for testing in email or IRC.


common/candlepin-common.spec.tmpl
```diff
@@ -3,176 +3,56 @@
...
+## GCJ-compiled .jar.so files.
+%global __jar_repack %{nil}
 
 %global parent_proj candlepin
 
 Name: %{parent_proj}-common
-Summary: Common code for Candlepin and related projects
-License: GPLv2
-Version: 1.0.23
+Version: 1.0.22
```
dgoodwin: This is a little weird, I think we downgraded here.

dgoodwin:
Am I correct that we run buildr pom when we modify dependencies, but normally not required for a regular tag?

If we fail to do so, the build will not compile and we'll have to re-tag to get a working build.

awood:
> Am I correct that we run buildr pom when we modify dependencies, but normally not required for a regular tag?
> 
> If we fail to do so, the build will not compile and we'll have to re-tag to get a working build.

Yeah, you must run `buildr pom` when new dependencies are added.  The only way to prevent the "retagging" problem is to do a `build --test` before you tag which would alert you of the problem.

PR #892, opened 2015-05-21
HATEOAS/JSON filtering enhancements

Ceiu:
- include and exclude may now be used together to create more precise
  filters; using both will default to blacklist mode

- API consumers can now control the filter mode explicitly by
  specifying filtermode=[whitelist|blacklist]

### DIFF
--- a/common/src/main/java/org/candlepin/common/resteasy/interceptor/DynamicFilterInterceptor.java
+++ b/common/src/main/java/org/candlepin/common/resteasy/interceptor/DynamicFilterInterceptor.java
@@ -53,9 +53,16 @@ public class DynamicFilterInterceptor implements PreProcessInterceptor {
         boolean containsExcludes = queryParams.containsKey("exclude");
         boolean containsIncludes = queryParams.containsKey("include");
 
-        // We want the list to be a blacklist by default when neither include nor exclude is
-        // provided, so we don't accidentally filter anything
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/candlepin/common/resteasy/interceptor/DynamicFilterInterceptor.java
```diff
@@ -48,30 +48,38 @@
...
-            return null;
+        boolean containsExcludes = queryParams.containsKey("exclude");
+        boolean containsIncludes = queryParams.containsKey("include");
+
+        DynamicFilterData filterData = new DynamicFilterData();
+
+        if (queryParams.containsKey("filtermode")) {
+            List<String> values = queryParams.get("filtermode");
+            filterData.setWhitelistMode("whitelist".equalsIgnoreCase(values.get(0)));
+        } else {
```
wottop: [checkstyle] /home/wpoteat/candlepin/common/src/main/java/org/candlepin/common/resteasy/interceptor/DynamicFilterInterceptor.java:61:9: '}' should be alone on a line.

Ceiu: Whoops. That one always burns me.

Fixed.

wottop:
ack

PR #894, opened 2015-05-26
Added product name filtering to the status trend report

Ceiu:
- Adding "product_name=<value>" to the status trend report query will
  now allow filtering on consumers who have installed the specified
  product, so long as there's at least one compliance record indicating
  the product's status (compliant, partial or non)


wottop:
ack

PR #895, opened 2015-05-27
1123560: Allow consumer with service level to autobind to pool without

wottop:
Previously this scenario was blocked by rule. Seems that a product
without a support_level attribute should be usable by anyone, not
just those without a service level assignment.


dgoodwin:
Looks correct, ack.

PR #896, opened 2015-06-01
Build task to check for usage of libraries with vulnerabilities.

awood:
Uses the OWASP Dependency Checker project.

Run with `buildr dependency:check`


wottop:
ack. nice improvements.

wottop:
re-ack

PR #897, opened 2015-06-03
Added additional filtering to the consumer status report

Ceiu:
Gutterball's consumer status report can now be filtered on (subscription) sku, subscription name or product name. The filtering works much like those provided by the status trend report in that only one is allowed in a given request.

These filter additions required a minor(-ish) change to the object model. Specifically, the ComplianceStatus class required some updates to some of its ElementCollection fields and the JSON converters it used.

Finally, the model changes I opted to use required the addition of an ID field in all three gb_*compprod_snap tables. As such, this commit includes a liquibase update for the tables, and a migration script to populate the ID columns.

The generic filtering API was not added, as it would best fit in common and would be a significant change in itself.


gutterball/src/main/java/org/candlepin/gutterball/liquibase/ComplianceStatusProductId.java
```diff
@@ -0,0 +1,201 @@
...
+        this.logger = logger;
+
+        this.preparedStatements = new HashMap<String, PreparedStatement>();
+    }
+
+    /**
+     *
+     *
+     *
+     */
```
mstead: Kill the empty comments or fill it out. Up to yourself.

gutterball/src/main/java/org/candlepin/gutterball/liquibase/ComplianceStatusProductId.java
```diff
@@ -0,0 +1,201 @@
...
+     * @return
+     *  a 32-character UUID
+     */
+    protected String generateUUID() {
+        return UUID.randomUUID().toString().replace("-", "");
+    }
+
+
+
+
```
mstead: Being picky, but wouldn't pass checkstyle with this whitespace  in here.

gutterball/src/main/java/org/candlepin/gutterball/report/ConsumerStatusReport.java
```diff
@@ -151,14 +173,60 @@ protected ReportResult execute(MultivaluedMap<String, String> queryParams, PageR
...
+                consumerIds,
+                ownerFilters,
+                statusFilters,
+                null,
+                null,
+                null,
+                attributeFilters,
+                pageRequest
+            );
+        }
```
mstead: Would it make sense to create specific methods in ComplianceSnapshotCurator for the calls to getSnapshotIterator above? Would probably be easier to understand what the method is doing. Looks like passing 'null' for certain parameters has a reason... perhaps we can make that more aparent.

gutterball/src/main/resources/logback.xml
```diff
@@ -9,8 +9,8 @@
         </encoder>
     </appender>
 
-    <logger name="org.candlepin.gutterball" level="INFO"/>
-    <logger name="org.candlepin.common" level="INFO"/>
+    <logger name="org.candlepin.gutterball" level="DEBUG"/>
+    <logger name="org.candlepin.common" level="DEBUG"/>
```
mstead: Should probably bump this back to INFO?

gutterball/src/main/java/org/candlepin/gutterball/report/ConsumerStatusReport.java
Ceiu: Potentially. This is kind of a stylistic thing derived from the status trend report where the curator/report doesn't actually care about which filters are provided, but combining some of them doesn't make sense logically, so they were disabled at the API level.

gutterball/src/test/resources/logback-test.xml
```diff
@@ -10,7 +10,13 @@
...
   <logger name="liquibase" level="ERROR"/>
 
-  <logger name="org.candlepin.gutterball" level="INFO"/>
+<!--
+  <logger name="org.hibernate.SQL" level="DEBUG"/>
+  <logger name="org.hibernate.type" level="TRACE"/>
+-->
+
+
+  <logger name="org.candlepin.gutterball" level="DEBUG"/>
```
mstead: I'm torn here... Should this be set to INFO?

gutterball/src/main/resources/db/changelog/2015-06-02-10-26-compliance-status-products.xml
```diff
@@ -0,0 +1,59 @@
...
+
+    <changeSet id="20150602102655-7" author="crog">
+        <addPrimaryKey
+            columnNames="id"
+            constraintName="gb_noncompprod_snap_pk"
+            tableName="gb_noncompprod_snap"
+        />
+    </changeSet>
+</databaseChangeLog>
+<!-- vim: set expandtab sts=4 sw=4 ai: -->
```
mstead: Might be worth adding indexes on the columns that we are filtering on.

Ceiu: Primary keys themselves function as indexes, and there are already pre-existing indexes on the other two columns.

gutterball/src/test/resources/logback-test.xml
Ceiu: In test, probably not unless it's generating a lot of noise. I don't really see it hurting anything either way, so I'm okay with whichever.

gutterball/src/main/resources/db/changelog/2015-06-02-10-26-compliance-status-products.xml
mstead: ACK.

mstead: > Primary keys themselves function as indexes, and there are already pre-existing indexes on the other two columns.

ACK

mstead:
:+1:  I'll put this feature through its paces today.

PR #898, opened 2015-06-03
Added valid checksums to the last guest update table changeset

Ceiu:
- Added a pair of validCheckSums to the changeset for populating the
  last guest update table. These represent the current, valid, checksum
  and a checksum for the version that was merged on Feb. 25, 2015.


wottop:
ack

PR #899, opened 2015-06-03
1227930: Deleting invalid content Ids causes error 500

Ceiu:
- Fixed a bug in EnvironmentResource that was causing a 500 error
  when an invalid content ID was provided
- The demotion logic has been slightly modified such that an invalid
  content ID will now cause a 404 to be thrown without performing
  any demotions

### DIFF
--- a/server/src/main/java/org/candlepin/resource/EnvironmentResource.java
+++ b/server/src/main/java/org/candlepin/resource/EnvironmentResource.java
@@ -42,6 +42,7 @@ import org.xnap.commons.i18n.I18n;
 import java.util.HashSet;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import javax.ws.rs.Consumes;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -235,14 +236,28 @@ public JobDetail demoteContent(
         @QueryParam("lazy_regen") @DefaultValue("true") Boolean lazyRegen) {
 
         Environment e = lookupEnvironment(envId);
-        Set<String> demotedContentIds = new HashSet<String>();
+
+        HashMap<String, EnvironmentContent> demotedContent = new HashMap<String, EnvironmentContent>();
```
wottop: Should be 'Map<String, EnvironmentContent>'

server/src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -235,14 +236,28 @@ public JobDetail demoteContent(
...
+        }
+
+        // We've validated everything here, so we're clear to start deleting...
+        for (EnvironmentContent envContent : demotedContent.values()) {
+            this.envContentCurator.delete(envContent);
         }
 
+        // Impl note: Unfortunately, we have to make an additional set here, as the keySet isn't
+        // serializable. Attempting to use it causes exceptions.
+        HashSet<String> demotedContentIds = new HashSet<String>(demotedContent.keySet());
```
wottop: demotedContentIds should be a Set

wottop:
ack

PR #900, opened 2015-06-04
5x Bind Improvement for Large SKUs

dgoodwin:
Eliminate a db lookup for the exact same SKU product being run twice per content set when generating entitlement certificates, leading to substantial performance improvements.

Also includes a more minor fix to stop filtering the content twice. This only saves a second or so but it was the next highest hit and completely unnecessary.

Using test SKU with ~2600 content sets, bind time was 10.8 seconds. After these changes we're down to 2.2s.

The latter patch is getting a bit messy because the V1 / V3 code split is messy and really needs a refactor. Going to attempt to do this on 2.0 / per-org products branch and limit changes to this 0.9.x stream, so went with the messy but fast approach here.


mstead:
Looks good and works well. I agree that the cert gen code could use a refactor. It has always been a little hard to follow IMO. Good stuff.

There appears to be a test failure in this PR. Need to get that fixed up and it'll be good to go.

dgoodwin:
My bad, removed a line that looked useless after my last test run. Ready now.

PR #901, opened 2015-06-04
Products and Content are now org-specific

Ceiu:
Product and Content are now stored in a per-org/owner manner. This change impacts several areas of the codebase, starting from the object model, to the way products/manifests are imported to how product/content data is queried.

Additionally, several other changes were also made during this process:
- Many model objects now reference Products and Content directly instead of requiring an additional lookup through a service adapter. This allows several other model objects to also be referenced directly on their logical parents. Primarily, product and pool attributes now exist entirely on products (as pools just use a product for their base info).

- Subscriptions have been almost entirely removed. Use of the Subscription and the SubscriptionServiceAdapter/SubscriptionCurator should be limited entirely to a small portion of the import process.

- Use of the service adapters as a whole should be minimized, instead favouring the use of curators.

- ProductCache has been removed entirely

- Changes to the API have been minimized, but some calls now require an owner/org in addition to their pre-existing requirements.

- The refresh pools operation is a no-op in non-hosted environments

- Some code which may now be dead has been commented out and left in place, at least until it can be reviewed and verified to no longer be needed. Additionally, there are a handful of TODO tasks which may require more attention before merging.


wottop:
Spec tests: if the create product is going to change across the board anyway, why not make owner key a variable instead of a parameter? Are there any calls to the orginal createProduct left?

wottop:
refresh_pools_spec: Big  section of commented tests. Has the TODO been settled?

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -49,44 +54,164 @@ public ProductCurator() {
...
+        if (!dest.getProductContent().equals(src.getProductContent())) {
+            dest.getProductContent().clear();
+            for (ProductContent pc : src.getProductContent()) {
+                dest.addProductContent(new ProductContent(dest, pc.getContent(), pc.getEnabled()));
+            }
+        }
+
+        if (!dest.getDependentProductIds().equals(src.getDependentProductIds())) {
+            dest.getDependentProductIds().clear();
+            dest.setDependentProductIds(src.getDependentProductIds());
```
alikins: Why copy/set this if they are already the same?

(also, no test coverage for that stanza)

wottop:
Service package:
Remove commented section of DefaultEntitlementCertServiceAdapter
DefaultProductServiceAdapter createProduct: remove todo [it works, right?]

dgoodwin:
Alright, let's do this.

dgoodwin:
**Build Log**
*last 10 lines*

 ```
[...truncated 122 lines...]
Build step 'Execute shell' marked build as failure
Archiving artifacts
Sending e-mails to: chainsaw@redhat.com
IRC notifier plugin: Sending notification to: #candlepin
Started calculate disk usage of build
Finished Calculation of disk usage of build in 0 seconds
Started calculate disk usage of workspace
Finished Calculation of disk usage of workspace in 0 seconds
Setting status of 63c37c426690526b7665eab52f43272404946dad to FAILURE with url http://rhsm-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/candlepin-docker-unittest-pullrequest/465/ and message: 'Build finished.'
```

PR #902, opened 2015-06-04
Official RHEL RPM Based Candlepin Containers

dgoodwin:
This PR adds 4 containers for Candlepin on RHEL. The two -base containers are RHEL 6 and 7, primed and ready for rpm installation. but candlepin is not actually yet installed. These use internal RH urls for builds so no actual RHEL content is in here, but it should be fairly trivial to update the FROM lines to point to CentOS and build. These are intended to be used by our automation to verify that rpms are installing and launching correctly. During "docker run" a YUM_REPO environment variable must be passed to the startup script specifying the location to install candlepin from.

The candlepin-rhel6 and candlepin-rhel7 layer on top of the base variants and actually install Candlepin from internal RPM builds. Likewise, these URLs in the candlepin.repo files could be modified to point to public Candlepin repos.

cpdb and cpsetup are temporarily copied into the containers until I can get modifications into a proper tagged build of candlepin, at which point these duplicates can be removed.

See the README.mkd for usage and more details on each.


docker/rebuild.sh
```diff
@@ -10,22 +10,34 @@
...
+
+#cd ../oracle
+#docker build -t candlepin-oracle .
+#docker tag -f candlepin-oracle docker.usersys.redhat.com/candlepin/candlepin-oracle
+#docker push docker.usersys.redhat.com/candlepin/candlepin-oracle
+
+#cd ../mysql
+#docker build -t candlepin-mysql .
+#docker tag -f candlepin-mysql docker.usersys.redhat.com/candlepin/candlepin-mysql
+#docker push docker.usersys.redhat.com/candlepin/candlepin-mysql
```
Ceiu: Commented out/migrated code left in. I'm guessing this isn't intentional.

dgoodwin: True!! Will fix.

PR #903, opened 2015-06-05
Awood/more mead

awood:
Adds several components missing from the Maven builds such as translations and the version.

### DIFF
--- a/gutterball/pom.xml
+++ b/gutterball/pom.xml
@@ -14,6 +14,7 @@
   <version>2.0.0</version>
   <packaging>war</packaging>
   <properties>
+    <release>1</release>
     <org.apache.qpid-qpid-common.version>0.30</org.apache.qpid-qpid-common.version>
     <org.apache.qpid-qpid-client.version>0.30</org.apache.qpid-qpid-client.version>
     <org.apache.mina-mina-core.version>1.0.1</org.apache.mina-mina-core.version>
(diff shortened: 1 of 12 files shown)


gutterball/pom.xml
```diff
@@ -821,6 +822,14 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.candlepin</groupId>
+      <artifactId>candlepin-common</artifactId>
+      <version>2.0.0</version>
+      <classifier>complete</classifier>
```
dgoodwin: Does every common update require a manual version update in gutterball/candlepin?

pom.xml
```diff
@@ -8,7 +8,8 @@
 
     <groupId>org.candlepin</groupId>
     <artifactId>candlepin-parent</artifactId>
-    <version>1.1</version>
+    <!-- This is the canonical version -->
+    <version>2.0</version>
```
dgoodwin: What is this version? Not following the canonical.

gutterball/pom.xml
awood: No, the `pom` plugin will take care of that.  After every update to common, you will need to run `buildr pom` but we should be doing that frequently anyway.

pom.xml
awood: The version of "candlepin-parent".  It's not a version we use or a version that really matters much, but before the canonical location was in `buildfile`.  Now it's in `pom.xml`.

gutterball/pom.xml
dgoodwin: Works for me.

dgoodwin:
Everything looks good to me Al, shall we merge?

PR #905, opened 2015-06-10
The populate hosted DB task no longer runs on standalone configurations

Ceiu:


server/src/main/java/org/candlepin/resource/AdminResource.java
```diff
@@ -119,6 +123,13 @@ public String initialize() {
     @Path("pophosteddb")
     public JobDetail populatedHostedDB() {
         // TODO: Remove this method once we no longer need the task.
+
+        // Impl note: We don't need to bother doing this
+        if (config.getBoolean(ConfigProperties.STANDALONE)) {
+            log.warn("Ignoring populate DB request due to standalone config.");
```
mstead: Message here is a little confusing depending on how you read it. I like something more along the lines as you've stated in the task itself: "Aborting populate DB task in standalone environment"

Perhaps something like: "Ignoring populate DB request in standalone environment".

mstead:
Other than that, this PR seems fine to me.

PR #908, opened 2015-06-12
Added per-org/product pool refreshing and by-product owner lookups

Ceiu:
- Added a new/fixed an old API call to ProductResource to lookup owners
  for a given product or list thereof
- The refresh pools operation can now be triggered on a per-org basis by
  specifying changed products
- Added new methods to candlepin_api.rb: get_owners_with_products and
  refresh_pools_for_orgs_with_product
- Updated and added spec tests


wottop:
good tests. ack.

PR #909, opened 2015-06-12
Fixed linking issues in Jenkins smoke-test script

Ceiu:
- Docker containers are no longer linked by container ID, as this was
  not supported in Docker 1.3. Instead, the DB container has a unique
  name generated for it, which is used to link to the container under
  test


dgoodwin:
ACK thanks

PR #914, opened 2015-06-17
Do not calculate compliance status if there are no installed products.

dgoodwin:
Big performance optimization for distributors which can have many entitlements
but no installed products. Looking up the consumer would then result in a huge
chunk of serialization to the compliance rules to calculate data we then did
nothing with.

No installed products = skip adding data to installed products.


wottop:
ack

PR #916, opened 2015-06-17
 Added filter options to Consumer/Entitlement resource ent lookups

mstead:
Affected API calls:
  GET /consumers/:uuid/entitlements
  GET /entitlements

Caller can now specify an 'attribute' query parameter to filter
results by pool/product attributes.

Caller can now specify a 'matches' query parameter to filter
entitlements by an entitlement pool's:
- product name/id
- contract number
- order number
- provided product name/id

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -743,8 +743,30 @@ class Candlepin
     path << "per_page=#{params[:per_page]}&" if params[:per_page]
     path << "order=#{params[:order]}&" if params[:order]
     path << "sort_by=#{params[:sort_by]}&" if params[:sort_by]
+
+    attr_filters = params[:attr_filters] || []
+    attr_filters.each do | attr_name, attr_value |
+      path << "attribute=#{attr_name}:#{attr_value}&"
+    end
+    path << "matches=#{params[:matches]}" if params[:matches]
     results = get(path)
     return results
+  end
+
+  # NOTE: Purely for testing via the entitlement resource.
+  #       Very similar to list_entitlements above (consumer resource)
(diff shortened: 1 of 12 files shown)


server/client/ruby/candlepin_api.rb
```diff
@@ -743,10 +743,32 @@ def list_entitlements(params={})
...
 
+  # NOTE: Purely for testing via the entitlement resource.
+  #       Very similar to list_entitlements above (consumer resource)
+  def list_ents_via_entitlements_resource(params={})
+    path = "/entitlements?"
+    path << "consumer=#{params[:consumer_uuid]}&" if params[:consumer_uuid]
+
+    attr_filters = params[:attr_filters] || []
+    attr_filters.each do | attr_name, attr_value |
+      path << "attribute=#{attr_name}:#{attr_value}&"
```
wottop: Is a colon a sufficient delimiter? Can an attribute value also have a colon?
Double colon maybe?

wottop:
I assume that the attribute value is match only. No wildcards?

server/client/ruby/candlepin_api.rb
wottop: I see this 
String[] parts = this.paramValue.split(":", 2); // Maximum of two parts
Guess it shouls work even if the value has a colon. Carry on.

wottop:
Just the one question. Looks good.

mstead:
This filtering works the same as the old Pool attribute filtering in the OwnerResource.

It supports a bunch of different formats. Wildcards are supported. It also handles '![value]' --  attribute=support_level:!Standard -- which means != or doesn't exist.

wottop:
ok, good. 

ack

PR #919, opened 2015-06-22
Added additional locking functionality to curators

Ceiu:
- Added the lock method to AbstractHibernateCurator, which
  will set the locking mode on the provided entity for the
  remainder of the transaction
- Fixed an issue with adding content to products while
  looking up that product on MySQL
- Updated candlepin_api.rb to use the newer org-specific
  call to retrieve product certificates; import_products
  and various spec tests have also been updated accordingly

### DIFF
--- a/server/bin/import_products.rb
+++ b/server/bin/import_products.rb
@@ -79,7 +79,7 @@ def create_owner(cp, new_owner)
   cp.create_activation_key(owner['key'], "awesome_os_pool")
 end
 
-thread_pool = Pool.new(4)
+thread_pool = Pool.new(5)
 data['owners'].each do |new_owner|
     thread_pool.schedule do
         create_owner(cp, new_owner)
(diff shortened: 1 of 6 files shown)


wottop:
spec failure against mysql: should auto-heal when guest is migrated to another host

     Failure/Error: ents['pool']['id'].should_not == original_virt_limit_pool
       expected not: == "8a8d09d64e207363014e209a5d791456"
                got:    "8a8d09d64e207363014e209a5d791456"
     # ./server/spec/virt_spec.rb:220:in `block (2 levels) in <top (required)>'

Ceiu:
After quite a bit of testing, I was able to trigger that failure both in this branch and in master, running in our Docker container and in a VM w/MySQL. Combined with the fact the test in question (virt_spec.rb:  'should auto-heal when guest is migrated to another host') doesn't ever call the resource containing the locking functionality, I'm confident that this error doesn't relate to this PR.

wottop:
ack

PR #920, opened 2015-06-22
Reverted automatic permission elevation

Ceiu:
- Scripts no longer auto-elevate to root if not already run as root, as Docker has problems with the way this was done
- rebuild.sh no longer needs to be run from its containing directory


dgoodwin:
Really good to get this out of there.

PR #924, opened 2015-06-25
Several improvements to the Docker image building scripts

Ceiu:
- RHEL images no longer include the cpdb or cpsetup files, as these
  are available via RPM
- rebuild.sh now exits with a non-zero value if any image failed to
  build successfully
- rebuild.sh now outputs error messages when images fail to build
- Added additional documentation for the USE_CACHE variable/-c option


dgoodwin:
Perfect, thanks!

PR #928, opened 2015-06-30
Removed classes surrounding the PoolTypeRules functionality

Ceiu:
- The PoolComplianceType, PoolComplianceTypeRules and respective test
  classes have been removed. PoolComplianceType has been merged into
  Pool as a public enum class.
- A pool's compliance type is no longer determined by Javascript-based
  rules. Instead, this is determined by the new Pool.getComplianceType()
  method.
- CalculatedAttributesUtilTest no longer calculates the quantity nor
  increment values


wottop:
Just for my education, why don't we need suggested quantity and increment?

Ceiu:
The running theory is that they aren't used, and it's another trip to the JS rules. By eliminating those and reworking the compliance types, we save two serialization round-trips through the rules.

wottop:
I guess we will find out soon if they are used when they are gone.

wottop:
ack

PR #931, opened 2015-07-07
Fix cross-compile issue with ConcurrentHashMap.

awood:
The Java 1.7 ConcurrentHashMap#keySet() returns a Set<K> while the Java
1.8 version returns ConcurrentHashMap.KeySetView<K>.  Since Java allows an
overridden method to return a sub-type of the parent method's return
type, no compilation are raised.  When we run the code on Java 1.7,
however, it immediately fails since the ConcurrentHashMap.KeySetView
class is not available.

One solution is to compile against a 1.7 bootstrap classpath, but since
1.7 is not packaged for recent versions of Fedora, we will work around
the issue by upcasting ConcurrentHashMaps to just Maps.

See https://gist.github.com/AlainODea/1375759b8720a3f9f094 for a
detailed explanation.

### DIFF
--- a/buildfile
+++ b/buildfile
@@ -188,6 +188,7 @@ define "candlepin" do
 
   compile.options.target = '1.6'
   compile.options.source = '1.6'
+  compile.options.lint = 'all'
 
   # path_to() (and it's alias _()) simply provides the absolute path to
   # a directory relative to the project.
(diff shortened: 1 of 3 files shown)


awood:
We're going to hit these types of problems until we can figure out a way to compile against a 1.6 `rt.jar`, but I'm not sure how to do that given that 1.6 isn't even available in F22 and JRE components aren't in the Maven central repo.

wottop:
ack

PR #932, opened 2015-07-08
Updated @Verify to use owner when checking products

Ceiu:
- When using @Verify with both an owner and a product, the
  lookup on the product will be owner-specific
- Added @Verify to most requests in ProductResource and
  OwnerProductResource
- ProductResource.getProductCertificate now throws an exception along
  with the other org-agnostic write operations

Note that this change introduces its own set of oddities. Particularly, to ensure the proper permission is checked, a @Verify must be present on both an owner (that occurs in the argument list before the product) and the product; and due to the ordering, a user must meet both verify criteria to perform the action. As such, the @Verify tags may need to be adjusted to have the proper explicit permissions for this update.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/OwnerProductResource.java
+++ b/server/src/main/java/org/candlepin/resource/OwnerProductResource.java
@@ -14,6 +14,7 @@
  */
 package org.candlepin.resource;
 
+import org.candlepin.auth.interceptor.Verify;
 import org.candlepin.common.auth.SecurityHole;
 import org.candlepin.common.exceptions.BadRequestException;
 import org.candlepin.common.exceptions.NotFoundException;
@@ -118,8 +119,9 @@ public class OwnerProductResource {
      */
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    public List<Product> list(@PathParam("owner_key") String ownerKey,
-                              @QueryParam("product") List<String> productIds) {
+    public List<Product> list(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @QueryParam("product") List<String> productIds) {
 
         Owner owner = this.getOwnerByKey(ownerKey);
 
@@ -159,8 +161,9 @@ public class OwnerProductResource {
     @Path("/{product_id}")
     @Produces(MediaType.APPLICATION_JSON)
     @SecurityHole
-    public Product getProduct(@PathParam("owner_key") String ownerKey,
-                              @PathParam("product_id") String productId) {
+    public Product getProduct(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId) {
 
         Owner owner = this.getOwnerByKey(ownerKey);
         Product product = productCurator.lookupById(owner, productId);
@@ -185,8 +188,10 @@ public class OwnerProductResource {
     @Path("/{product_id}/certificate")
     @Produces(MediaType.APPLICATION_JSON)
     @SecurityHole
-    public ProductCertificate getProductCertificate(@PathParam("owner_key") String ownerKey,
-                                                    @PathParam("product_id") String productId) {
+    public ProductCertificate getProductCertificate(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId) {
+
         Product product = this.getProduct(ownerKey, productId);
         return this.productCertCurator.getCertForProduct(product);
     }
@@ -202,7 +207,10 @@ public class OwnerProductResource {
      */
     @POST
     @Produces(MediaType.APPLICATION_JSON)
-    public Product createProduct(@PathParam("owner_key") String ownerKey, Product product) {
+    public Product createProduct(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        Product product) {
+
         Owner owner = this.getOwnerByKey(ownerKey);
 
         product.setOwner(owner);
@@ -221,8 +229,8 @@ public class OwnerProductResource {
     @Path("/{product_id}")
     @Produces(MediaType.APPLICATION_JSON)
     public Product updateProduct(
-        @PathParam("owner_key") String ownerKey,
-        @PathParam("product_id") String productId,
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId,
         Product product) {
 
         Product toUpdate = this.getProduct(ownerKey, productId);
@@ -322,9 +330,10 @@ public class OwnerProductResource {
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/{product_id}/batch_content")
-    public Product addBatchContent(@PathParam("owner_key") String ownerKey,
-                                   @PathParam("product_id") String productId,
-                                   Map<String, Boolean> contentMap) {
+    public Product addBatchContent(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId,
+        Map<String, Boolean> contentMap) {
 
         Product product = this.getProduct(ownerKey, productId);
         List<ProductContent> productContent = new LinkedList<ProductContent>();
@@ -351,10 +360,11 @@ public class OwnerProductResource {
     @POST
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/{product_id}/content/{content_id}")
-    public Product addContent(@PathParam("owner_key") String ownerKey,
-                              @PathParam("product_id") String productId,
-                              @PathParam("content_id") String contentId,
-                              @QueryParam("enabled") Boolean enabled) {
+    public Product addContent(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId,
+        @PathParam("content_id") String contentId,
+        @QueryParam("enabled") Boolean enabled) {
 
         Product product = this.getProduct(ownerKey, productId);
         Content content = this.getContent(product.getOwner(), contentId);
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -167,9 +167,15 @@ public Product getProduct(@PathParam("product_id") String productId) {
...
-        return this.productCertCurator.getCertForProduct(product);
+    public ProductCertificate getProductCertificate(
+        @PathParam("product_id") String productId) {
+
+        throw new BadRequestException(this.i18n.tr(
+            "Organization-agnostic product write operations are not supported."
+        ));
+
+        // Product product = this.findProduct(productId);
+        // return this.productCertCurator.getCertForProduct(product);
```
dgoodwin: Dead code to be removed?

server/src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -286,10 +300,11 @@ public void deleteProduct(@PathParam("product_id") String productId) {
     @GET
     @Path("/{product_id}/statistics")
     @Produces(MediaType.APPLICATION_JSON)
-    public List<Statistic> getProductStats(@PathParam("product_id") String productId,
-                            @QueryParam("from") String from,
-                            @QueryParam("to") String to,
-                            @QueryParam("days") String days) {
+    public List<Statistic> getProductStats(
+        @Verify(Product.class) @PathParam("product_id") String productId,
```
dgoodwin: A method without a @Verify param and without @SecurityHole becomes super-admin only, and I think we want to preserve that particularly in these global resource classes, introducing @Product verify might actually be opening this up to callers who shouldn't be able to add content to a product or get product stats, I think what we want here is to remove @Verify and take this right back to super admin only. (which is what Katello would operate as when it makes these calls)

server/src/main/java/org/candlepin/resource/ProductResource.java
Ceiu: Yeah, forgot to remove it.

server/src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -306,11 +321,12 @@ public void deleteProduct(@PathParam("product_id") String productId) {
     @GET
     @Path("/{product_id}/statistics/{vtype}")
     @Produces(MediaType.APPLICATION_JSON)
-    public List<Statistic> getProductStats(@PathParam("product_id") String productId,
-                            @PathParam("vtype") String valueType,
-                            @QueryParam("from") String from,
-                            @QueryParam("to") String to,
-                            @QueryParam("days") String days) {
+    public List<Statistic> getProductStats(
+        @Verify(Product.class) @PathParam("product_id") String productId,
```
dgoodwin: Probably should drop this one too.

server/src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -342,20 +358,20 @@ public void deleteProduct(@PathParam("product_id") String productId) {
     @GET
     @Path("/owners")
     @Produces(MediaType.APPLICATION_JSON)
-    public List<Owner> getProductOwners(@QueryParam("product") String[] productId) {
-        List<String> ids = Arrays.asList(productId);
+    public List<Owner> getProductOwners(
+        @Verify(Product.class) @QueryParam("product") List<String> productIds) {
```
dgoodwin: Same.

server/src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -363,15 +379,14 @@ public void deleteProduct(@PathParam("product_id") String productId) {
     @Path("/subscriptions")
     @Produces(MediaType.APPLICATION_JSON)
     public JobDetail[] refreshPoolsForProduct(
-        @QueryParam("product") String[] productId,
+        @Verify(Product.class) @QueryParam("product") List<String> productIds,
```
dgoodwin: Same.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -118,8 +119,9 @@ protected Owner getOwnerByKey(String key) {
      */
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    public List<Product> list(@PathParam("owner_key") String ownerKey,
-                              @QueryParam("product") List<String> productIds) {
+    public List<Product> list(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @QueryParam("product") List<String> productIds) {
```
dgoodwin: This is a GET, so you'd need read access on the owner and the product to list the products in an owner, that seems right.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -185,8 +188,10 @@ public Product getProduct(@PathParam("owner_key") String ownerKey,
     @Path("/{product_id}/certificate")
     @Produces(MediaType.APPLICATION_JSON)
     @SecurityHole
-    public ProductCertificate getProductCertificate(@PathParam("owner_key") String ownerKey,
-                                                    @PathParam("product_id") String productId) {
+    public ProductCertificate getProductCertificate(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId) {
```
dgoodwin: This one I almost think should go back to super admin only if it was in 1.0, which I suspect it was. This doesn't feel like something regular users or consumers should be calling, and I could tsee them having read access to the owner and product.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -202,7 +207,10 @@ public ProductCertificate getProductCertificate(@PathParam("owner_key") String o
      */
     @POST
     @Produces(MediaType.APPLICATION_JSON)
-    public Product createProduct(@PathParam("owner_key") String ownerKey, Product product) {
+    public Product createProduct(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        Product product) {
```
dgoodwin: This one worries me a bit as well. An org admin might have write perms for an Owner, should they be able to create products? I think product creation might have been super admin only, forcing normal users to come in via a manifest import, or Katello UI which operates as a super admin when calling candlepin.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -221,8 +229,8 @@ public Product createProduct(@PathParam("owner_key") String ownerKey, Product pr
     @Path("/{product_id}")
     @Produces(MediaType.APPLICATION_JSON)
     public Product updateProduct(
-        @PathParam("owner_key") String ownerKey,
-        @PathParam("product_id") String productId,
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId,
```
dgoodwin: Similar to create, probably should be super admin only as far as I can think ahead right now.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -322,9 +330,10 @@ protected Content getContent(Owner owner, String contentId) {
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/{product_id}/batch_content")
-    public Product addBatchContent(@PathParam("owner_key") String ownerKey,
-                                   @PathParam("product_id") String productId,
-                                   Map<String, Boolean> contentMap) {
+    public Product addBatchContent(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId,
+        Map<String, Boolean> contentMap) {
```
dgoodwin: If we're in agreement on create/update, same for this one.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -351,10 +360,11 @@ public Product addBatchContent(@PathParam("owner_key") String ownerKey,
     @POST
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/{product_id}/content/{content_id}")
-    public Product addContent(@PathParam("owner_key") String ownerKey,
-                              @PathParam("product_id") String productId,
-                              @PathParam("content_id") String contentId,
-                              @QueryParam("enabled") Boolean enabled) {
+    public Product addContent(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId,
```
dgoodwin: If we're in agreement on create/update, same for this one.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -378,9 +388,10 @@ public Product addContent(@PathParam("owner_key") String ownerKey,
     @DELETE
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/{product_id}/content/{content_id}")
-    public void removeContent(@PathParam("owner_key") String ownerKey,
-                              @PathParam("product_id") String productId,
-                              @PathParam("content_id") String contentId) {
+    public void removeContent(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId,
+        @PathParam("content_id") String contentId) {
```
dgoodwin: If we're in agreement on create/update, same for this one.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -425,11 +437,13 @@ public void deleteProduct(@PathParam("owner_key") String ownerKey,
...
-                                           @PathParam("product_id") String productId,
-                                           @QueryParam("from") String from,
-                                           @QueryParam("to") String to,
-                                           @QueryParam("days") String days) {
+    public List<Statistic> getProductStats(
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId,
+        @QueryParam("from") String from,
+        @QueryParam("to") String to,
+        @QueryParam("days") String days) {
```
dgoodwin: Super admin only I fear. :( Probably not even used, but potential DOS there.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -470,8 +485,8 @@ public void deleteProduct(@PathParam("owner_key") String ownerKey,
     @Path("/{product_id}/subscriptions")
     @Produces(MediaType.APPLICATION_JSON)
     public JobDetail refreshPoolsForProduct(
-        @PathParam("owner_key") String ownerKey,
-        @PathParam("product_id") String productId,
+        @Verify(Owner.class) @PathParam("owner_key") String ownerKey,
+        @Verify(Product.class) @PathParam("product_id") String productId,
         @QueryParam("lazy_regen") @DefaultValue("true") Boolean lazyRegen) {
```
dgoodwin: Super admin only. :(

dgoodwin:
A bit concerned this is opening up API calls that should not be, but it might need some testing to know for sure, based on what I remember of permissions, this probably does open some API calls to any org admin that they didn't have before, and probably shouldn't have at all. Pretty big potential for abuse particularly in prod if this is correct.

PR #935, opened 2015-07-14
Fixed duplicate results when using Entitlement matches filter

mstead:
Since we joining on content, we need to make sure we have a distinct root
entity when querying with the matches filter.

## Testing

Attach one of the stackable ram subsciptions from the test data. I used ram-2gb-stackable -- "RAM Limiting Package (2GB stackable)"

Then run the curl command below, and expect a singe result.

```bash
curl -k -u admin:admin "https://localhost:8443/candlepin/consumers/:consumer_uuid/entitlements?matches=*ram*"
```

### DIFF
--- a/server/spec/entitlement_spec.rb
+++ b/server/spec/entitlement_spec.rb
@@ -22,12 +22,24 @@ describe 'Entitlements' do
                                     :attributes => { 'instance_multiplier' => 2,
                                         'multi-entitlement' => 'yes' })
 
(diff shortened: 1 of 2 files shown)


wottop:
Fix grammar in commit comment

wottop:
other than the above, ack

mstead:
Grammar fixed in commit message.

wottop:
ack

PR #936, opened 2015-07-14
Added preconditions to custom change tasks

Ceiu:
- Added preconditions to custom change tasks, and any reliant task,
  to prevent them from running during offline/sql output mode

This update is primarily to help IT with their internal workflow that requires examining/processing the SQL statements to be run. This won't provide a 100% accurate representation (as the custom tasks tend to run quite a few statements), but it does allow updateSQL to run.


dgoodwin:
Seems to deploy clean even on top of a db that already had these changes applied, no checksum errors, ACK.

dgoodwin:
Feel free to backport to 0.9.51 hotfix branch and build for rhel!

PR #937, opened 2015-07-15
Fix derived product import with 0.9 manifests.

dgoodwin:
We implemented some fixes to let old manifests import the main product ID from
a legacy manifest, but derived product ID was missed. Because this is the key
piece of data we check to trigger derived product behaviour, we ended up with
no derived product data importing or taking effect.

Implement the same hack required for product ID.


wottop:
ack

PR #938, opened 2015-07-16
Fix InsufficientPermission error when checking HypervisorCheckIn Job

rnovacek:
When virt-who created new HypervisorCheckIn job using new async API, it
couldn't check state of that job because it was not allowed. This commit
fixes the issue and virt-who can check state of job it created.


dgoodwin:

cnsnyder:
ACK

PR #939, opened 2015-07-16
Fix EL7 docker rpm container test.

dgoodwin:
Addition of set -e introduced problems in the startup.sh script, as the cpsetup
command does error out on EL7 because the service it not yet running, thus we
can't check status and initialize. Previously this would fail and continue in
the script, starting supervisor.

Instead we'll keep the set -e, but add a new cpsetup option that skips service
stop/restart/init, and eliminates the failed command during startup.

Also added wget calls to fetch the latest cpsetup and cpdb for this base
container used for rpm installation tests. This isn't used by the more official
tagged containers with a build in them, only for the base image that doesn't
yet have cp installed.

### DIFF
--- a/docker/candlepin-rhel6-base/startup.sh
+++ b/docker/candlepin-rhel6-base/startup.sh
@@ -27,7 +27,13 @@ fi
 # TODO: use env variables to check which database we're linked to, for now
 # we'll just assume postgres.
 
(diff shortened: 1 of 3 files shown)


Ceiu:
ACK
The cpsetup issue should be resolved once this branch is merged to master.

PR #940, opened 2015-07-16
Activation key auto bind across pools

wottop:
Ensures that multiple consumers can use the same key to register.
The key is auto attach and has all the pools in the group associated.
Test uses the consumers to exhaust the pools and then one more key
 usage to ensure successful registration.


wottop:
This PR is just a test of a scenario, not added functionality.

Ceiu:
ACK

Test works in isolation, in a group and in the entire suite without side effects.

PR #942, opened 2015-07-17
Updated preconditions to be compatible with Liquibase 2.x

Ceiu:
Added a dummy precondition to fix an issue with older versions of liquibase which require at least one precondition present.

### DIFF
--- a/server/src/main/resources/db/changelog/20140408160212-add-pool-source-sub-table.xml
+++ b/server/src/main/resources/db/changelog/20140408160212-add-pool-source-sub-table.xml
@@ -87,7 +87,9 @@
     </changeSet>
 
     <changeSet id="20140408160212-3" author="ckozak">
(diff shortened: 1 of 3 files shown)


server/src/main/resources/db/changelog/20140408160212-add-pool-source-sub-table.xml
```diff
@@ -87,7 +87,9 @@
     </changeSet>
 
     <changeSet id="20140408160212-3" author="ckozak">
-        <preConditions onSqlOutput="FAIL" onFail="CONTINUE"/>
+        <preConditions onSqlOutput="FAIL" onFail="CONTINUE">
+            <tableExists schemaName="candlepin" tableName="cp_pool"/>
```
dgoodwin: Is that schemaName correct/required? Don't think we use schema's anywhere do we?

PR #943, opened 2015-07-20
Fixes for the docker rpm smoke tests on jenkins.

dgoodwin:
3 seconds isn't always enough for postgres to come up, causing the candlepin container to fail fast if it's got the candlepin rpm pre-installed. (a discrepancy fixed in the job itself on jenkins)

Added some more info for debugging purposes.

Fixes failures if run on Python 2.7.9+ where SSL certs are checked by default.

### DIFF
--- a/docker/jenkins-scripts/smoke-test.sh
+++ b/docker/jenkins-scripts/smoke-test.sh
@@ -62,7 +62,7 @@ if len(sys.argv) < 3:
 image_name = sys.argv[1]
 cp_repo_url = sys.argv[2]
 max_wait_time = 600
-max_log_lines = 50
+max_log_lines = "all"
 
 server_container_id = None
 db_container_id = None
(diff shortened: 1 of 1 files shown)


docker/jenkins-scripts/smoke-test.sh
```diff
@@ -78,64 +89,77 @@ try:
...
-                run_command("docker logs --tail=%s %s" % (max_log_lines, server_container_id), False, True)
+            regex = re.compile(".*:(\\d+)\\s*\\Z")
+            output = run_command("docker port %s 8443/tcp" % server_container_id)
+            match = regex.match(output[0])
+
+            if match:
+                port = match.group(1)
+
+                # Wait for it to start...
+                print "Containers started successfully using port %s. Waiting for Candlepin to start..." % port
```
Ceiu: Looks like your verbage change got clobbered here. If you care enough to change it back, should also correct the port message to be less awkward.

Ceiu:
ACK

Runs successfully against rhel6-base, rhel6 and rhel7-base on python 2.7.10. Fails to run against rhel7 without adding --skip-service to the startup.sh (which is outside the scope of this commit).

PR #945, opened 2015-07-21
Fix docker rpm smoke-test script AttributeError.

dgoodwin:
ssl module is already imported and exists on the older Python we're targetting.
The actual error surfaces as an AttributeError.


Ceiu:
ACK

PR #949, opened 2015-07-24
1243146: Suggested quantity should be calculated for pool listings

wottop:


Ceiu:
ACK
Suggested quantity shows up as expected

PR #950, opened 2015-07-28
Crog/docker build script update

Ceiu:
Several improvements/changes to the Docker containers and the scripting to build them. In particular, the non-RHEL containers are now based on CentOS, and the tagging/publishing process has been improved to use the version/tag from git when Candlepin isn't provided via rpm.


Ceiu:
This needs more work

PR #952, opened 2015-07-29
1247242: CPM will attempt to retry entitlements under certain conditions

Ceiu:
- When the CandlepinPoolManager triggers an EntitlementRefusedException while
  entitling by products due to entitlements not being available, it will
  retry the operation. This works around a synchronization issue that can
  occur when a pool is selected for consumption, but is locked and consumed
  by another request/thread before it can be consumed.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -794,7 +794,10 @@ public class CandlepinPoolManager implements PoolManager {
                 String key = e.getResult().getErrors().get(0).getResourceKey();
 
                 if (key.equals("rulefailed.no.entitlements.available") && retries-- > 0) {
-                    log.info("No entitlements available while attempting to entitle; retrying");
+                    log.info(
+                        "Entitlements exhausted between select best pools and bind operations; retrying"
+                    );
+
                     continue;
                 }
 
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -92,6 +92,8 @@ public class CandlepinPoolManager implements PoolManager {
     private PoolCurator poolCurator;
     private static Logger log = LoggerFactory.getLogger(CandlepinPoolManager.class);
 
+    private static final int MAX_ENTITLE_RETRIES = 3;
+
     private EventSink sink;
     private EventFactory eventFactory;
     private Configuration config;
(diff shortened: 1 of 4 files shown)


Ceiu:
I opted for the retry approach rather than attempting to change how the locking is done due to the potential difficulties in picking a locking target (would likely need to be the owner for this particular operation/issue), and the repercussions of adding more locking to the model at this stage.
Admittedly this really only side-steps the issue, rather than fixing it entirely, but I feel without a proper change/design to the locking model, adding more locks will just make things worse.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -761,28 +763,44 @@ public Pool find(String poolId) {
...
+            catch (EntitlementRefusedException e) {
+                String key = e.getResult().getErrors().get(0).getResourceKey();
 
-        // now make the entitlements
-        for (PoolQuantity entry : bestPools) {
-            entitlements.add(addOrUpdateEntitlement(consumer, entry.getPool(),
-                null, entry.getQuantity(), false, CallerType.BIND));
-        }
+                if (key.equals("rulefailed.no.entitlements.available") && retries-- > 0) {
+                    log.info("No entitlements available while attempting to entitle; retrying");
```
dgoodwin: Could we change this to something more specific:

"Entitlements exhausted between select best pools and bind operations, retrying"

If we hit this, I believe that's what it has to mean as we check quantity during select best pools.

dgoodwin:
Good fix IMO, just a little nervous about bringing a threaded test into spec tests we already run threaded.

How long does it take to run that one test? Does everything  behave ok when you run all tests in parallel? Typically we don't introduce concurrent requests into tests, but rather scripts in bin/ that we have floating around for various problems like this that have occurred.

How do you feel about adding it to the suite?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
Ceiu: Done.

Ceiu:
The new spec test specifically for this issue only takes a second or two to run on my machine. Running the entire suite in parallel also does not appear to have issues.

Most of the work is direct through the api, so I could probably pull it out without much difficulty if that's preferred.

Ceiu:
Remind me to squash the commits before this is merged.

dgoodwin:
Couple seconds and well behaved in full suite and I'm happy.

Something that occurred to me last night. How does this behave if the autobind that will eventually fail, has successfully grabbed a few entitlements. i.e. key has pools 1 and 2, 1 has ample quantity but 2 is going to run out. Does the retry correctly see the consumer now has additional entitlements granted from the first attempt?

dgoodwin:
**Build Log**
*last 10 lines*

 ```
[...truncated 11904 lines...]
Build step 'Execute shell' marked build as failure
Archiving artifacts
Sending e-mails to: chainsaw@redhat.com
IRC notifier plugin: Sending notification to: #candlepin
Started calculate disk usage of build
Finished Calculation of disk usage of build in 0 seconds
Started calculate disk usage of workspace
Finished Calculation of disk usage of workspace in 0 seconds
Setting status of 4b2b16870847b0610057dda38335120a7a0ad421 to FAILURE with url http://rhsm-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/candlepin-docker-unittest-pullrequest/525/ and message: 'Build finished.'
```

dgoodwin:
Alright I think we're good to go, this is a nice non-invasive solution.

PR #954, opened 2015-07-30
Reduce database chatter in GET /status.

dgoodwin:
This looks like it might be the largest chunk of processing our app does in
prod. Client makes ample use of this, but IT also uses it as a health check.

First step drops querying of rules version and source, instead we set these on
the provider singleton whenever we recompile rules. (i.e. they changed) The two
places they can change (rules upload and import) are updated to immediately
recompile rather than wait for someone to attempt to use them.

Ideally we'd have dropped all db connectivity from the class, but unfortunately
we need to know if the db loses it's connection, so IT can drop the node from
rotation automatically. Instead, we'll reduce the query dramatically and just
query the updated timestamp.

A nice future enhancement would be to catch db errors (all of them somewhere)
and set some kind of flag once we see any.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/StatusResource.java
+++ b/server/src/main/java/org/candlepin/resource/StatusResource.java
@@ -20,8 +20,12 @@ import org.candlepin.common.util.VersionUtil;
 import org.candlepin.config.ConfigProperties;
 import org.candlepin.model.RulesCurator;
 import org.candlepin.model.Status;
+import org.candlepin.policy.js.JsRunnerProvider;
 
 import com.google.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Map;
 
@@ -35,6 +39,7 @@ import javax.ws.rs.core.MediaType;
  */
 @Path("/status")
 public class StatusResource {
+    private static Logger log = LoggerFactory.getLogger(StatusResource.class);
 
     /**
      * The current version of candlepin
@@ -49,10 +54,11 @@ public class StatusResource {
     private boolean standalone = true;
 
     private RulesCurator rulesCurator;
+    private JsRunnerProvider jsProvider;
 
     @Inject
     public StatusResource(RulesCurator rulesCurator,
-                          Configuration config) {
+                          Configuration config, JsRunnerProvider jsProvider) {
         this.rulesCurator = rulesCurator;
 
         Map<String, String> map = VersionUtil.getVersionMap();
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -89,17 +96,23 @@ public StatusResource(RulesCurator rulesCurator,
...
-                rulesCurator.listAll();
-            }
+            rulesCurator.getUpdatedFromDB();
         }
         catch (Exception e) {
+            log.error("Error checking database connection", e);
             good = false;
         }
+
         Status status = new Status(good, version, release, standalone,
```
Ceiu: Would it make sense to change this method around to store a status instance and simply return it unless (a) it's null (b) it's older than X seconds or (c) a "force refresh" flag is provided? That would probably cut down some extra processing time a bit, assuming users didn't just always apply the force refresh flag.

server/src/main/java/org/candlepin/sync/RulesImporter.java
```diff
@@ -53,6 +56,9 @@ public void importObject(Reader reader) throws IOException {
                 existingRules.getVersion() + " new version: " + newRules.getVersion());
             curator.update(newRules);
             sink.emitRulesModified(existingRules, newRules);
+
+            // Trigger a recompile of the rules:
+            jsProvider.get();
```
Ceiu: Could we just make compileRules public and call that directly? .get appears to create some context and a new runner. If we're only interested in a recompile, we could save ourselves a bit of work there.

server/src/main/java/org/candlepin/resource/StatusResource.java
dgoodwin: Current discussions with IT are leaning towards serving a static representation to clients who make this API call, so only running this for IT status checking. For that, we probably need it to be live, though I'm working on getting them to call it less often. Hopefully we can find a way to drop the db query from here entirely, ideally I want something catching exceptions and setting a flag that status will return. If we could do that and just get to this tiny bit of pure java, I'd have to think we're pretty fast. This call as it stands here is 2ms on my machine max.

server/src/main/java/org/candlepin/sync/RulesImporter.java
dgoodwin: Yeah makes sense, I'll adjust.

Ceiu:
ACK
Spec and unit tests are looking good with the most recent commit; DB queries are down to 1 per request and the status API request is returning almost instantly.

PR #955, opened 2015-07-30
Dgoodwin/hornetq logging

dgoodwin:
Just a couple logging improvements I want to get into stage to get a better understanding of where hornetq might be falling over. Neither should result in any significant traffic to the logs.


wottop:
ack

PR #957, opened 2015-08-05
Update required Qpid packages.

awood:
Based on information discovered in
http://projects.theforeman.org/issues/11285

QPid requires a package to be installed to persist items to disk.  The old implementation was "qpid-cpp-server-store" and the newer one is "qpid-cpp-server-linearstore".  You need one of them or else everything vanishes once the service is restarted.


wottop:
ack

PR #958, opened 2015-08-07
Turn off compiler linting.  Too noisy.

awood:
Eliminates the numerous warnings Buildr has been printing during compilation.

### DIFF
--- a/buildfile
+++ b/buildfile
@@ -188,7 +188,8 @@ define "candlepin" do
 
   compile.options.target = '1.6'
   compile.options.source = '1.6'
-  compile.options.lint = 'all'
+  # Enable to turn on compiler linting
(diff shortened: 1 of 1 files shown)


vritant:
@awood shouldn't we heed to the warnings though, or silence the ones we don't want to individually, if we know what we are doing?

awood:
Yes, ideally we would.  Unfortunately, the code has many, many warnings right now and I haven't had a chance to address them.  They show up in Eclipse, so I felt like having them output during the build was extraneous.

We should all probably get in the habit of adding some `SuppressWarnings` annotations as we move through the code.  Most of the warnings are around casting objects in the Hibernate layer IIRC.

Ceiu:
Joyful ACK

PR #959, opened 2015-08-10
List owner entitlements

vritant:
* fix refresh pool spec tests
* refactor entitlement curator to allow searching for entitlements by owner, consumer, or all.
* clean up refactor entitlement
* add junit tests for resources and entitlement curator
* update owner/{key}/entitlements API to paginate and search by product, attributes, etc
* add spec tests for update of owner entitlements API ( not done yet )


dgoodwin:
ok to test

vritant:
@dgoodwin I have a concern with this change. this API aleady existed, and there was no regeneration of dirty entitlements when the API was called. while the API change is backwards compatible, the default behavior has changed to regenerate all the dirty entitlements. this could be potentially expensive for the entire owner organization.

I could change default behavior to NOT regenerate by default, but I don't thing that is desired. thoughts?

Ceiu:
retest this please

dgoodwin:
@vritant I would think this API should not regenerate dirty entitlements, it's not a client API call to fetch certs so regenerating them here will probably do something unintentional and break more systems holding the old certs.

vritant:
@dgoodwin removed regem
@Ceiu done

vritant:
@Ceiu oops, spoke too soon. forgot the junit tests

vritant:
retest this please

vritant:
retest this please

wottop:
ack

PR #962, opened 2015-08-12
More Docker build script updates

Ceiu:
- All container tags are now pushed, rather than just specific versions
- Updated default Docker registry


dgoodwin:
Could you nab the references in the README as well.

Ceiu:
README updated

PR #964, opened 2015-08-13
Docker build script updates

Ceiu:
- Version checking now uses /bin/sh as the base command, rather than cd
- Added a linter option to cp-test


Ceiu:
retest this please

PR #966, opened 2015-08-14
Docker build script updates

Ceiu:
- Oracle's build script no longer attempts to use docker's tty mode
- Added the -r/--rpm option to the Jenkins smoke test script to
  specify the repo URL (2nd parameter) is a direct reference to
  an RPM
- Added the -f (fast) option to run rspec tests in parallel
- The smoke test now rounds the time remaining instead of truncating

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -66,6 +66,7 @@ usage() {
         -d  deploy a live candlepin
         -t  populate Candlepin database with test data (implies -d)
         -r  run rspec test suite (implies -d)
+        -f  run rspec tests in parallel (implies -r)
         -u  run unit test suite
         -l  run the linter against the code
         -s  run a bash shell when done
@@ -75,7 +76,7 @@ usage() {
 HELP
 }
 
-while getopts ":dtrulsc:p:v" opt; do
+while getopts ":dtrfulsc:p:v" opt; do
     case $opt in
         d  ) DEPLOY="1";;
         t  )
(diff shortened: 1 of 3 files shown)


Ceiu:
To test the individual changes here
- Ensure the candlepin-base, candlepin-oracle and one of candlepin-rhel6-base or candlepin-rhel7-base are built. These can be done individually by invoking the build.sh scripts in the respective directories, or by running the rebuild.sh script in the docker directory. When building individually, candlepin-base must be built before candlepin-oracle.
- While building the candlepin-oracle container, ensure that it builds successfully with and without a tty. See the following link for details on running without tty: http://unix.stackexchange.com/questions/68574/invoke-a-command-script-disconnected-from-the-controlling-terminal
- Once built, invoke any of the candlepin-* containers with an explicit cp-test command to verify the -f option is recognized and runs the tests in parallel: `docker run --rm -Pti candlepin/candlepin-oracle cp-test -f`   This can be run with any of the DB-oriented built in the preceding steps.
- Finally, the smoke test can be run by specifying both a container and an RPM location that would be accessible from within the container. We can make use of an RPM @awood put together for testing: `./jenkins-scripts/smoke-test.sh --rpm candlepin/candlepin-rhel6-base http://file.rdu.redhat.com/~awood/candlepin-0.9.51.4-1.noarch.rpm`

If all goes well, everything should complete successfully and without errors/warnings.

Ceiu:
Disregard the step involving the -f flag -- this has been removed.

PR #967, opened 2015-08-18
Associating product content repeatedly from rspec can cause lock fail…

wottop:
…ures

Each call to the API for adding content starts a new thread that will attempt a
pessimistic lock on the product. If they happen quickly enough, the lock fails. Now
all content gets added in one call for the test data load.


Ceiu:
ACK

PR #969, opened 2015-08-19
Push migration detection from virt-who checkin to system checkin.

dgoodwin:
The lookup for the host to have most recently reported every guest UUID in a
virt-who checkin was showing up as a major performance bottleneck, exposing the
fact that there are virt-who deployments of far larger size than anything we
anticipated.

When reporting thousands of guest IDs into orgs with tens of thousands already
there, the performance very rapidly degrades beyond 60 second timeouts and
continues to scale up linearly as new guest IDs are added.

The check was done during virt-who checkins to detect migration, revoke
entitlements restricted to the previous host, and auto-heal if necessary.

We realized however that (a) most guest IDs never become registered consumers,
and (b) this behavior is causing bugs in revoking entitlements that the guest
systems are still attempting to use.

Instead, we push the logic to system check-in time when fetching entitlements,
certs, or serials, similar to when we regenerate dirty entitlements. This
eliminates all the wasted computation on guest IDs that never become consumers,
and the query becomes a small search for a guest UUID rather than thousands all
at once.

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -746,6 +746,7 @@ class Candlepin
     path << "per_page=#{params[:per_page]}&" if params[:per_page]
     path << "order=#{params[:order]}&" if params[:order]
     path << "sort_by=#{params[:sort_by]}&" if params[:sort_by]
+    path << "regen=false&" if params[:regen]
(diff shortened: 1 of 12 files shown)


wottop:
ack

PR #970, opened 2015-08-19
Upgrade Quartz to latest 2.2.1 (from 2013).

dgoodwin:


wottop:
TestBot: retest this please

wottop:
You added a not-null column and then didn't seed it. I started up with master then ran the deploy/update.

Updating Database
Liquibase update Failed: Migration failed for change set db/changelog/20150818110722-upgrade-to-quartz-2-dot-2.xml::20150818110722-1::dgoodwin:
     Reason: liquibase.exception.DatabaseException: Error executing SQL ALTER TABLE public.qrtz_fired_triggers ADD sched_time BIGINT NOT NULL: ERROR: column "sched_time" contains null values

dgoodwin:
My bad, I can't seem to trigger this with a straight master + testdata deploy, then deploy on branch without -g. I will add a default ASAP.

dgoodwin:
This might be a little tough to reproduce, entries in that table are transient, only existing when a trigger is being fired, not sure if that means there are rows when a job is running, or just before it actually launches. 

Quratz wants this field to be a bigint, not a datetime or anything we can work with easily. Following this response for the off chance you can get data in this table when doing an upgrade.

http://stackoverflow.com/questions/28211747/upgrade-to-quartz-2-2-jdbc-jobstore-new-column-sched-time

To test I deployed master and added a row of junk with:

insert into qrtz_fired_triggers(entry_id, trigger_name, trigger_group, instance_name, fired_time, priority, state_ values('a', 'a', 'a', 'a', 4000, 4000, 'a');

Then deploy this branch without -g to upgrade the db and ensure it's ok. I did this for both postgres and mysql, we have no pre-existing oracle db's so did not test this and disabled the changeset for those db's.

dgoodwin:
retest this please

dgoodwin:
Lets try this.

dgoodwin:
Retest this please.

Ceiu:
ACK
No issues deploying after priming the DB.

PR #971, opened 2015-08-19
Log execution time of all jobs by default.

dgoodwin:
Override logExecutionTime in the few jobs that run every 5 seconds to keep those quieter, everything else we log how long it ran for.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/tasks/KingpinJob.java
+++ b/server/src/main/java/org/candlepin/pinsetter/tasks/KingpinJob.java
@@ -62,6 +62,8 @@ public abstract class KingpinJob implements Job {
     @Override
     public void execute(JobExecutionContext context) throws JobExecutionException {
 
+        long startTime = System.currentTimeMillis();
+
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/candlepin/pinsetter/tasks/KingpinJob.java
```diff
@@ -62,16 +62,23 @@
...
 
+        String jobUuid = context.getJobDetail().getKey().getName();
+        long startTime = System.currentTimeMillis();
+
         // Store the job's unique ID in log4j's thread local MDC, which will automatically
         // add it to all log entries executed for this job.
         try {
             MDC.put("requestType", "job");
-            MDC.put("requestUuid", context.getJobDetail().getKey().getName());
+            MDC.put("requestUuid", jobUuid);
```
wottop: By moving this outside of the try/catch, you have made 15 unit tests NPE

dgoodwin:
Oie my bad. Updated now.

Ceiu:
ACK
Not a huge fan of the units missing, but it's consistent with other, similar output.

PR #972, opened 2015-08-19
Forced refresh on the add/delete rules operations

wottop:
The cached rules data needs to be forced past the date checking to ensure
proper version use.
The spec test also required a sleep so that the rpm rules and the test update
rules do not have the same time in mysql.


Ceiu:
ACK

PR #973, opened 2015-08-28
Storage band implementation

mstead:
Implemented storage band counting. Each entitlement
will equate to 1TB of coverage and marketing products
will use the multiplier such that sub quantity purchased
* multiplier will yeild pool quantity (total capacity in TB).

For example:
pool_quanity = sub_qty_bought * (storage_band_attr * multiplier)
             = 1 * (1 * 256)
             = 256

The resulting pool will have a quantity of 256, which can
cover 256TB of storage.

The rules will calculate a system's coverage based on a
band.storage.usage fact.

Consider a system is using 128TB:
- a quantity of 128 entitlements will cover the system (green).
- a quantity < 128 entitlements will partially cover the
  system (yellow)
- no entitlements (red)

Marketing Product (requirements)
---------------------------------
The marketing product should define a multiplier that is equal
to the total capacity in TBs. It must be stackable and must
define the storage_band attribute (should be set to 1).

- stacking_id: "product_stack_id"
- mutli-entitlement: "yes"
- storage_band: 1

Consumer Requirements
-------------------------
The consumer must have the band.storage.usage fact set to
the total TB that it is using.

NOTE: If this fact is not set, the rules will assume a value
      of 1TB.

- band.storage.usage: 128

### DIFF
--- a/server/bin/import_products.rb
+++ b/server/bin/import_products.rb
@@ -255,6 +255,12 @@ def create_mkt_product(cp, owner, product)
     return
   end
 
+  small_quantity = SMALL_SUB_QUANTITY
+  large_quantity = LARGE_SUB_QUANTITY
+  if product.has_key?('quantity')
+    small_quantity = large_quantity = product['quantity']
+  end
+
   provided_products = product['provided_products'] || []
   derived_product_id = product['derived_product_id']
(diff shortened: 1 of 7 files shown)


wottop:
ack

PR #974, opened 2015-08-28
Revert "Wrap all REST API requests in a transaction."

dgoodwin:
This feature was found to cause a massive memory leak which would surface under
heavy load. The 52a51af8 commit specifically is the point where the memory
starts to degrade. Before this change used PS old gen would peak at about
200mb, after this change it will climb up to 2.6gb.

May be possible to re-introduce this functionality once the interceptor is
fixed to not leak memory.

Kept one portion of the reverted commits to not modify a live hibernate object
in UserResource when listing roles. (as this was generally dangerous)

This reverts commit 78581503e7f9dad531ef047668099a15d365d3bf.
This reverts commit 37c0207a91050a5c6f9dce439417b787dc2177fd.
This reverts commit 3628df5902413c57dea30aca6bc3c4cd569d0fc1.
This reverts commit 52a51af8c0696ba304521a8ed21f2d7f647ca109.


dgoodwin:
retest this please

Ceiu:
ACK.

Used the loadtest written during performance testing. When running a single instance against the local deployment on my VM, this branch finishes in just under 4 minutes with the tenured generation utilizing 170mb and 1.7GB total allocated. Current master, in the same conditions, deadlocks around the 11m mark after utilizing 2GB total memory, with 1.3GB of that allocated to the tenured generation.

Further improvements can likely be made with some c3p0 configuration, though this is outside the scope of this commit.

PR #975, opened 2015-08-28
Revert consumer/hypervisor last check-in tables.

dgoodwin:
This was intended to help reduce lock wait timeouts by removing contention on
cp_consumer.lastcheckin. However we later discovered the performance impact was
just too great as the size of the tables grew and unfortunately the work has to
be reverted.

This was only fixing a symptom and not the underlying cause, so that will have to be dealt 
with instead. Recent changes may help reduce the long transactions that caused this in 
any case.

Upgrade script is included to revert the values back to the max in the db at the time.

Quartz will also have to be updated to remove refs to the jobs that are being
removed. In IT's case this will need to be done manually as their quartz tables
are in a separate database.

NOTE: if you attempt to upgrade a pre-existing mysql db with this, it will fail. I'm treating the mysql dbms as if it were IT here, as that is our only pre-existing mysql deployment. I don't want them to run liquibase and get SQL they can't run (because their quartz tables are in another db). So upgrade is only provided for postgresql, otherwise it assumes you're doing a fresh db deploy and don't have quartz jobs laying around.


Ceiu:
ACK

PR #978, opened 2015-09-09
1194776: Gutterball equivalent of cpdb to handle upgrade etc.

nguyenfilip:


awood:
candlepin-pull-request-bot, add to whitelist

common/dbtool/dbtool
```diff
@@ -0,0 +1,399 @@
...
+            choices=["Postgres", "Oracle", "Mysql"],
+            dest="database_type",
+            help="Type of the database to run against. Valid values are: Postgres, Mysql, Oracle. Defaults to Postgres")
+    
+    parser.add_option("--product",
+            dest="product",
+            type="choice",
+            choices=["candlepin","gutterball"],
+            help="Product. Valid values are: candlepin, gutterball.")
+
```
Ceiu: Depending on whether or not this is going to replace cpdb entirely, we may want to default this to "candlepin" so it can be a drop-in replacement for cpdb without breaking any existing tooling that may use it.

nguyenfilip: That is a good point. I was afraid that in some use cases e.g. when Candlepin and Gutterball is installed on the same machine, the user could accidently drop-create Candlepin db

mstead:
retest this please

common/dbtool/dbtool
mstead: While @Ceiu makes a good point, I think @nguyenfilip concern may be valid. Personally I think we should prevent them from making that mistake and force them to specify the target product.

I suppose though that the db creds would have to be the same for it to happen, but...

mstead: They will have to update any scripts to use the new tool regardless.

nguyenfilip: Ah the db creds. Thats true. I changed my mind, --product=candlepin default should be ok in that case :-)

PR #979, opened 2015-09-09
1259248: Fixed issue with refresh pools removing ueber certs

Ceiu:
- Refresh pools, and anything making use of the function, will no
  longer remove any generate ueber certs
- Partial information for ueber certs or consumers will no longer
  cause 500 errors when generating or retrieving ueber certs

### DIFF
--- a/server/src/main/java/org/candlepin/resource/OwnerResource.java
+++ b/server/src/main/java/org/candlepin/resource/OwnerResource.java
@@ -1475,22 +1475,23 @@ public class OwnerResource {
         Owner o = findOwner(ownerKey);
 
         if (o == null) {
-            throw new NotFoundException(i18n.tr(
-                "owner with key: {0} was not found.", ownerKey));
-        }
-
-        Consumer ueberConsumer = consumerCurator.findByName(o, Consumer.UEBER_CERT_CONSUMER);
-
-        // ueber cert has already been generated - re-generate it now
-        if (ueberConsumer != null) {
-            List<Entitlement> ueberEntitlement
-                = entitlementCurator.listByConsumer(ueberConsumer);
-            // Immediately revoke and regenerate ueber certificates:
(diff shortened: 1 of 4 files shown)


Ceiu:
This can be tested by following the steps provided by @mstead on the original BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1259248

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1475,22 +1475,20 @@ public EntitlementCertificate createUeberCertificate(@Context Principal principa
...
-                = entitlementCurator.listByConsumer(ueberConsumer);
-            // Immediately revoke and regenerate ueber certificates:
-            poolManager.regenerateCertificatesOf(ueberEntitlement.get(0), true, false);
-            return entitlementCertCurator.listForConsumer(ueberConsumer).get(0);
+            throw new NotFoundException(i18n.tr("owner with key: {0} was not found.", ownerKey));
         }
 
         try {
+            Consumer ueberConsumer = consumerCurator.findByName(o, Consumer.UEBER_CERT_CONSUMER);
+            List<Entitlement> ueberEntitlements = entitlementCurator.listByConsumer(ueberConsumer);
```
vritant: @Ceiu This line should be moved inside the if code block in the safety of the null check for ueberConsumer ( and not rely on listByConsumer to be safe ) 

PR looks good otherwise, will merge once this is tweaked.

Ceiu: I've made the change, but for what it's worth, it's entirely unnecessary. The contract on listByConsumer is such that we know it always returns a list and has no input restrictions. The code flow doesn't change at all here, and we end up making it slightly more ugly/convoluted for no additional stability or security.

vritant: @Ceiu I understand, Thanks for humouring me :+1:

PR #981, opened 2015-09-10
Don't lose reports when multiple virt-whos exist for the same owner

vritant:


vritant:
@candlepin-pull-request-bot you insult me.

cnsnyder:
If possible it would be great to have a test to show two of these jobs (with the same owner) will not run at the same time. Other than that everything looks good. Nothing in my manual testing has failed and the changes look good.

cnsnyder:
Looks good, ACK

PR #985, opened 2015-09-15
Persist reporter id for Hypervisor checkins

vritant:
While an owner can have multiple hypervisor - guest mapping reporters, Assumption is that at a given time, only one reporter reports hypervisor mappings for a particular hypervisor of an owner. This PR ensures that we:
* persist if reporter id is available in the request, else ignore.
* update if reporter id changed for the hypervisor
* warn if a hypervisor's reporter changed.
* dont warn if a reporter id was not set in the first place ( to avoid excessive warnings when this version first hits production )

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -112,10 +112,13 @@ class Candlepin
     return consumers
   end
 
-  def hypervisor_update(owner, json_data, create_missing=nil)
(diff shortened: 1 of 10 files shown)


vritant:

Ceiu:
add to whitelist

cnsnyder:
Looks good to me, ACK

PR #986, opened 2015-09-15
Removed un-needed method from HypervisorResource

vritant:


cnsnyder:
@vritant Looking through the history, I don't see anything to worry about. Removing this method seems to make sense. ACK

PR #987, opened 2015-09-16
1263474: Standalone candlepin now returns the expected error message …

cnsnyder:
…and code

When we removed the DefaultSubscriptionServiceAdapter, the activateSubscription method of its replacement (ImportSubscriptionServiceAdapter) became a no op. This causes a success response code to be sent to subscription manager with no content in the body. As subscription manager at this time has no default message to display this resulted in no feedback being displayed to the user.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1263474


Ceiu:
ACK

PR #988, opened 2015-09-17
 1256926: Fixed Pool match filter query

mstead:
When using match parameters while listing available pools
hibernate would return a partially initialized list of
pool objects when matching on a collection of sub objects.

For example, if the match filter matched on 1 of 4 provided
products, only the matched provided product would be loaded
in the Pool's collection. Only a hibernate refresh would
bring the remainder in after the initial lookup.

This patch changes the query to use correlated subqueries
instead of left outer joins, which works around the partial
collection issue.

I tried implementing a unit test for the curator but
wasn't able to reproduce. I have a hunch that HSQLDB
may be dealing with joins a little differently than
postgres. A spec test easily reproduced this issue.

https://bugzilla.redhat.com/show_bug.cgi?id=1256926

### DIFF
--- a/server/bin/performance-dataloader.rb
+++ b/server/bin/performance-dataloader.rb
@@ -80,7 +80,7 @@ def create_pools(cp, owner_key, count, attributes, quantity)
   time = Benchmark.realtime do
     (1..(count)).each do |i|
       mkt_prod_id = random_string('mkt-prod')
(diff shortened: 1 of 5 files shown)


mstead:
retest this please

mstead:
retest this please

awood:
Life with Hibernate: :scream:

PR #989, opened 2015-09-21
Updated cpc to work with changes to Ruby's public_instance_methods me…

Ceiu:
…thod

- CandlepinCLI.callable_name? now properly converts stringable objects
  to strings before attempting to use them as strings. This should fix
  an issue on versions of Ruby on which public_instance_methods returns
  Symbol objects instead of method name Strings.

### DIFF
--- a/server/client/ruby/cpc
+++ b/server/client/ruby/cpc
@@ -26,8 +26,8 @@ class CandlepinCLI < Candlepin
 
   private
 
-  def self.callable_name?(method_name)
-    !method_name.end_with?('=')
+  def self.callable_name?(method)
+    !method.to_s.end_with?('=')
   end
 
 end


Ceiu:
retest this please

server/client/ruby/cpc
```diff
@@ -26,8 +26,9 @@ class CandlepinCLI < Candlepin
 
   private
 
-  def self.callable_name?(method_name)
-    !method_name.end_with?('=')
+  def self.callable_name?(method)
+    method_name = method.to_s
+    return !['?', '='].include?(method_name[-1])
```
awood: You should probably remove the "?".  Methods ending in a question mark are callable.  They just, by convention, return booleans.  You could probably just get away with

```
!method_name.to_s.end_with?('=')
```

PR #991, opened 2015-09-23
Cdk consumer

wottop:
DO NOT MERGE.

Using PR to collaborate and comment


server/spec/consumer_resource_cdk_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    @cdk_product = create_product("cdk_product",
+                                  "CDK Product",
+                                  {:attributes => { :expired_after => "60"}})
+    @p_product1 = create_product("p_product_1",
+                                  "Provided Product 1")
+    @p_product2 = create_product("p_product",
+                                  "Provided Product 2")
+    @consumer = consumer_client(@user, @consumername, :system, 'cdk_user', facts= {:dev_sku => "cdk_product"})
+    installed = [
+        {'productId' => @cdk_product.id, 'productName' => @cdk_product.name},
```
mstead: This product should not be an installed product -- it is the dev_sku marking product that will be looked up.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -186,6 +211,22 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
                 data.setConsumer(consumer);
             }
         }
+        if (consumer.isCdk()) {
+            if (config.getBoolean(ConfigProperties.STANDALONE) ||
+                    poolCurator.listByOwner(consumer.getOwner(), new Date()).size() == 0) {
```
mstead: Here we look up all pools just to call getSize(). It would probably be more efficient to add a method to the curator to ask if it has an active subscription. Then we avoid pulling all the pool data back from the database.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -205,6 +246,65 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+    private Pool getCdkPool(Consumer consumer, String sku) {
+        List<Pool> pools = poolCurator.listByConsumer(consumer);
+        for (Pool p : pools) {
+            if (p.hasAttribute(Pool.DEVELOPMENT_POOL_ATTRIBUTE) &&
+                p.hasAttribute(Pool.REQUIRES_CONSUMER_ATTRIBUTE) &&
+                p.getAttributeValue(Pool.REQUIRES_CONSUMER_ATTRIBUTE).equals(consumer.getUuid())) {
+                return p;
+            }
+        }
+        return null;
```
mstead: Why look up all the consumer's pools just to pull out the single dev pool -- seems like wasted resources. Let's do a lookup for that pool via the curator so that we don't have to uselessly pull back all the pool data for the consumer. There should only ever be one dev pool per consumer, right?

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -205,6 +246,65 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+    }
+
+    protected Pool assembleDevPool(Consumer consumer, String sku) {
+        // all good. create a pool for the CDK consumer
+        Set<Product> providedProducts = new HashSet<Product>();
+        Date now = new Date();
+
+        Product prod = getCdkInstalledProduct(consumer, sku);
```
mstead: This is not an installed product, it is the SKU. By reusing this method here, we get an error about missing installed product if it isn't found.

server/src/main/java/org/candlepin/service/ProductServiceAdapter.java
```diff
@@ -39,6 +39,13 @@
     Product getProductById(Owner owner, String id);
 
     /**
+     * Query a specific product by its string ID.
+     * @param id product id
+     * @return specific product
+     */
+    Product getProductById(String id);
+
```
mstead: Not sure about this one-off method on the adapter. The original goal was to remove the adapter. I wonder if it would make sense to perhaps create a new method that is targeted at the dev sku. That way IT could isolate the lookup if required and trim down the adapter to that method later on, instead of completely removing it.

The fact that this feature doesn't work 'Out Of The Box' without a backing subscription source is bothersome. The 'shared product' concept here makes it difficult I know, since candlepin doesn't have that any more. Perhaps out of the box the feature targets an owner, but production targets the 'shared' product (ignores the owner arg during lookup). I'll have to think about this some more. Please do the same.

server/src/main/java/org/candlepin/service/impl/DefaultProductServiceAdapter.java
```diff
@@ -67,6 +67,12 @@ public Product getProductById(Owner owner, String id) {
     }
 
     @Override
+    public Product getProductById(String id) {
+        // this is a no-op unless we are running in the portal
+        return null;
+    }
+
```
mstead: See comment on the interface method.

mstead:
Since this might be reused for other platforms other than CDK, perhaps we should remove references to CDK and instead refer to them as devX or developerX. Thoughts?

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -254,6 +254,35 @@ public Pool find(Serializable id) {
...
+            date = new Date();
+        }
+        Criteria crit = createSecureCriteria();
+        crit.add(Restrictions.eq("activeSubscription", Boolean.TRUE));
+        crit.add(Restrictions.eq("owner", o));
+        crit.add(Restrictions.le("startDate", date));
+        crit.add(Restrictions.ge("endDate", date));
+        crit.setProjection(Projections.rowCount());
+
+        long count = (Long) crit.list().get(0);
```
mstead: I've always done this as follows, however, I'm not sure if there is a difference in using uniqueResult vs list().get(0). Might be something to investigate -- I'm not 100% sure which is better TBH.

```java
long count = (Long) crit.uniqueResult();
```

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -205,6 +248,68 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+        String prodThenString = prod.getAttributeValue("expired_after");
+        if (prodThenString != null && Long.parseLong(prodThenString) < maxDevLifeDays) {
+            interval = Long.parseLong(prodThenString);
+        }
+        return 1000 * 60 * 60 * 24 * interval;
+    }
+
+    private Pool getDevPool(Consumer consumer, String sku) {
+        PoolFilterBuilder poolFilters = new PoolFilterBuilder();
+        poolFilters.addAttributeFilter(Pool.REQUIRES_CONSUMER_ATTRIBUTE, consumer.getUuid());
```
mstead: Perhaps we should filter on dev pool type here in case we ever use the consumer specific restriction again.

server/src/main/java/org/candlepin/service/ProductServiceAdapter.java
mstead: Any thoughts on this?

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -205,6 +248,68 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+        if (StringUtils.isEmpty(prod.getAttributeValue("support_level"))) {
+            // if there is no SLA, apply the default
+            prod.setAttribute("support_level", this.DEFAULT_DEV_SLA);
+            prod = productCurator.createOrUpdate(prod);
+        }
+
+        for (ConsumerInstalledProduct ip : consumer.getInstalledProducts()) {
+            providedProducts.add(retrieveNamedDevProduct(consumer, ip.getProductId()));
+        }
+
```
mstead: Looking at lines 280 - 289:
We are doind N + 1 product lookups here via the adapter. Can we make use of the ProductAdapter.getProductsByIds(owner, ids) call so that we only hit the adapter once? It would likely create some complexity in determining what products were found, but would save on lookups.

In the error messages, we might also want to distinguish b/w not finding the SKU vs installed products.

wottop: It would require some post processing to determine what is missing. Should this call [in this specific scenario] just have a null owner?

mstead: I'd say pass the owner in. It'll get ignored in prod.

server/src/main/resources/rules/rules.js
```diff
@@ -59,6 +59,7 @@ var ARCH_ATTRIBUTE = "arch";
 var RAM_ATTRIBUTE = "ram";
 var INSTANCE_ATTRIBUTE = "instance_multiplier";
 var REQUIRES_HOST_ATTRIBUTE = "requires_host";
+var REQUIRES_CONSUMER_ATTRIBUTE = "requires_consumer";
```
mstead: Need to bump the rules version.

server/src/main/java/org/candlepin/controller/Entitler.java
mstead: BUMP - Was this looked at? Is my thought here correct?

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -205,6 +250,102 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+                else {
+                    message.append(i18n.tr("The installed product ''{0}'' for this development " +
+                            "unit does not exist in the repository. ",
+                            prodIds.get(i)));
+                }
+            }
+        }
+        throw new ForbiddenException(message.toString());
+    }
+
```
mstead: This could produce a very large message, containing some duplication with "The installed product {0} for this development unit does not exist in the repository." for each product. Would using 'repository' here be a little confusing?

Also, IMO we should throw a separate exception for unknown SKU vs unknown installed products.

Why not just loop through the products and remove the product_id from the list of prodIds if found. Any left over are products that were not found. The check that list to see if the sku was found. if not, throw an exception. If the SKU is not found, autobind failed.

Then if the sku was found and there are products remaining in the list, throw an exception with a single message and perhaps comma separated list of installed product ids.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -205,6 +250,103 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+        devProductIds.add(sku);
+        for (ConsumerInstalledProduct ip : consumer.getInstalledProducts()) {
+            devProductIds.add(ip.getProductId());
+        }
+        List<Product> prods = productAdapter.getProductsByIds(consumer.getOwner(), devProductIds);
+        if (prods.size() != devProductIds.size()) {
+            // determine exception message
+            produceMissingDevProductException(devProductIds, prods);
+        }
```
mstead: I think it would be better to rename the produceMissingDevProductException method to verifyDevProducts and move the 'if (prods.size() != devProductIds.size())' check inside it.

PR #995, opened 2015-09-30
Vritant/subservice

vritant:


vritant:
@Ceiu ready for review

PR #997, opened 2015-10-05
Updated cp-test to run the linter from the top-level CP directory

Ceiu:
- cp-test now runs checkstyle from $CP_HOME rather than the individual
  project directory
- Re-added the note about the insecure registry


cnsnyder:
Works for me, ACK

PR #1000, opened 2015-10-16
Vritant/fr translations 1167490

vritant:
* fix translations with unescaped single quotes
* add buildr tasks to validate and fix if needed


alikins:
Assuming these changed will be (or have already been) pushed up to zanata, ACK.

vritant:
@alikins I will do that before I update the BZ, thanks!

PR #1001, opened 2015-10-20
Vritant/hosted tests

vritant:
@Ceiu @mstead Am requesting multiple eyes on this. please take a look :)
simply run deploy -ha to run tests in hosted mode, this will take care of setting hosted mode and overriding customized module in candlepin.conf,  

more details : http://www.candlepinproject.org/docs/candlepin/mode_agnostic_spec_testing.html

On a high level, the changes are:
* updated createPools api in owner resource to enrich & create bonus pools ( went ahead with this within this PR cause its closely tied to most of our tests )
* fixed a few misc. bugs in candlepin which were required to pass the tests 
  * NPE issue in Entitlement.toString
  * Pool.java serialization issue
  * Product.java lazy loading issue
  * issue where we weren't overriding product with derived product on bonus pools upon pool refresh ( PoolRules.java )
* added hosted test resources that are exlcuded from candlepin.war unless specified.
* moved some resolver utility methods from OwnerResource to ResolverUtil so could be shared with HostedTestResource
* added a ruby client for that resource - hostedtest_api
* updated import_products script to call create pools instead of create subscription, and no-longer call refresh pools
* added methods in candlepin_scenarios with elaborate comments  to use hostedtest_api in a mode-agnostic way - this way tests dont have to worry which mode we are running candlepin in.
* deprecated updateSubscriptions
* re-wired updatePools to call updateSubscriptions for now, with a todo to tackle that in a separate PR - the number of tests that use this is less, and wanted to do the least possible refactoring at this point.
* ensured all tests are either pending or ignored in hosted mode.
* ensured standalone tests aren't broken because of changes.


vritant:

wottop:
Spec tests run in both modes serially. Hosted will not do parallel. I hear that is expected.

server/client/ruby/candlepin_api.rb
```diff
@@ -1359,11 +1389,14 @@ def put(uri, data=nil)
     return JSON.parse(response.body) unless response.body.empty?
   end
 
-  def delete(uri, data=nil)
+  def delete(uri, data=nil, dont_parse=false)
     puts ("DELETE #{uri}") if @verbose
     client = get_client(uri, Net::HTTP::Delete, :delete)
     client.options[:payload] = data.to_json if not data.nil?
     response = client[URI.escape(uri)].delete(:content_type => :json, :accepts => :json)
+    if dont_parse
```
Ceiu: What situation came up that we needed this block?

server/spec/import_spec.rb
```diff
@@ -35,7 +35,7 @@
 
   it 'creates pools' do
     pools = @import_owner_client.list_pools({:owner => @import_owner['id']})
-    pools.length.should == 6
+    pools.length.should == 5
```
Ceiu: What caused these to change?

server/spec/refresh_pools_spec.rb
```diff
@@ -217,10 +214,14 @@
     product2 = create_product(name, name, :owner => owner2['key'])
 
     # migrate the subscription to another owner.
+    sub = get_hostedtest_subscription(pool.subscriptionId)
     sub["owner"] = owner2
-    @cp.update_subscription(sub)
-    @cp.list_subscriptions(owner1["key"]).length.should == 0
-    @cp.list_subscriptions(owner2["key"]).length.should == 1
+    update_hostedtest_subscription(sub)
+    #TODO - removing sleeps gives concurrent modification errrors - investigate
```
Ceiu: Concurrent modification where? If it's in CP, I would argue it needs to be fixed rather than hiding it via sleeps -- actual clients won't be so kind.

server/spec/refresh_pools_spec.rb
```diff
@@ -242,14 +242,13 @@
...
-    @cp.update_subscription(sub)
-    @cp.list_subscriptions(owner1["key"]).length.should == 0
-    @cp.list_subscriptions(owner2["key"]).length.should == 1
+    update_hostedtest_subscription(sub)
 
     # Refresh the second owner so that the pools are updated.
-    @cp.refresh_pools(owner2["key"])
-
+    @cp.refresh_pools(owner2["key"], true)
+    sleep 1
```
Ceiu: Is this sleep a consequence of the previous TODO tag?

server/src/main/java/org/candlepin/hostedtest/HostedTestSubscriptionServiceAdapter.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public boolean hasUnacceptedSubscriptionTerms(Owner owner) {
+        return false;
+    }
+
+    @Override
+    public void sendActivationEmail(String subscriptionId) {
+    }
```
Ceiu: Not necessarily a requirement, but for methods like this, I generally like to add a comment that states the method was intentionally left empty so it's clear that it wasn't omitted/unimplemented by mistake.

server/src/main/java/org/candlepin/hostedtest/HostedTestSubscriptionServiceAdapter.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public boolean canActivateSubscription(Consumer consumer) {
+        return false;
+    }
+
+    @Override
+    public void activateSubscription(Consumer consumer, String email, String emailLocale) {
+    }
```
Ceiu: Same here

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -104,10 +104,16 @@
...
+    /*
+     * hibernate persists empty set as null, and tries to fetch
+     * dependentProductIds upon a fetch when we lazy load. to fix this, we eager
+     * fetch.
+     */
     @ElementCollection
     @CollectionTable(name = "cpo_product_dependent_products",
                      joinColumns = @JoinColumn(name = "product_uuid"))
     @Column(name = "element")
+    @LazyCollection(LazyCollectionOption.FALSE)
```
Ceiu: I'm curious what the scenario is here that caused a problem. In all of the constructors and in the setter, we set the collection to an empty hash set, so there aren't many cases where this can be null.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1217,6 +1045,62 @@ public JobDetail refreshPools(
...
+    @Path("{owner_key}/pools")
+    public void updatePool(@PathParam("owner_key") @Verify(Owner.class) String ownerKey,
+            Pool pool) {
+
+        if (pool.getType() != PoolType.NORMAL) {
+            throw new BadRequestException(i18n.tr("Cannot update bonus pools, as they are auto generated"));
+        }
+
+        // convert to subscription so we can use updateSubscription
+        Subscription subscription = new Subscription(pool.getOwner(), pool.getProduct(),
```
Ceiu: We should probably skip the conversion to a subscription object here and instead just update the pool directly.

server/bin/deploy
```diff
@@ -178,10 +178,11 @@ init
 
 DBUSER="candlepin"
 
-while getopts ":fgtolmav" opt; do
+while getopts ":fghtolmav" opt; do
```
Ceiu: Commenting here due to a shortcoming in Github

Don't forget to update the help output with the new option details.

server/client/ruby/candlepin_api.rb
vritant: @Ceiu When the response is a boolean, we cant parse it to json.

server/src/main/java/org/candlepin/model/Product.java
vritant: per my comment above I don't think it is that were setting this as null, but hibernate is persisting it as null.

I think we are hitting: https://forum.hibernate.org/viewtopic.php?f=1&t=1011750
( but for a Set and not a list. )
Ive tried and contemplated various approaches ( detaching object from
hibernate, creating a new session in the getSubscription call, etc)
with no success.


using @OrderColumn doesnt work either cause we have a set of strings.

server/src/main/java/org/candlepin/hostedtest/HostedTestSubscriptionServiceAdapter.java
vritant: fixed

server/src/main/java/org/candlepin/hostedtest/HostedTestSubscriptionServiceAdapter.java
vritant: fixed

server/bin/deploy
vritant: fixed

server/src/main/java/org/candlepin/resource/OwnerResource.java
vritant: @Ceiu I agree, hence the TODO on the API. To do so will mean a few more ripple effect changes ( like the changes incurred as a part of changing createPools ) , and given the fact this is a test only API used in far fewer places than createPools ( only when tests need to modify upstream subscription source ), I thought we should tackle this later.

server/src/main/java/org/candlepin/model/Product.java
Ceiu: That link looks to be referring to storing nulls at the end of the list and not getting them back, and getting an empty list out when it only contained null. I think we're hitting a different problem entirely.

vritant: @Ceiu what do you recommend we do?

server/spec/import_spec.rb
vritant: So this was cause I removed the virt_limit attribute from the product3 in StandardExporter , because in hosted mode the importer spec fails otherwise.
on a second thought, ignoring the import spec completely in hosted mode, and reverting this change to  pools.length.should == 6

server/spec/refresh_pools_spec.rb
vritant: @Ceiu yes, in candlepin . 
I'd recommend tackling this in a separate PR. I might be wrong , but I think this is already broken in master.


org=] INFO  org.candlepin.resource.OwnerResource - Cleaning up owner: Owner [id: ff8080815111ba3d015111bd259a023d, key: initial-owner-wi0Hxy90]
2015-11-16 14:19:48,852 [thread=QuartzScheduler_Worker-11] [job=refresh_pools_f5d4845b-bd06-41a8-840e-445f372b0325, org=] ERROR org.candlepin.pinsetter.tasks.RefreshPoolsJob - RefreshPoolsJob encountered a problem.
org.candlepin.common.exceptions.ConcurrentModificationException: Request failed due to concurrent modification, please re-try.
        at org.candlepin.model.AbstractHibernateCurator.flush(AbstractHibernateCurator.java:386) ~[AbstractHibernateCurator.class:na]
        at org.candlepin.model.PoolCurator.delete(PoolCurator.java:559) ~[PoolCurator.class:na]
        at com.google.inject.persist.jpa.JpaLocalTxnInterceptor.invoke(JpaLocalTxnInterceptor.java:58) ~[guice-persist-3.0.jar:na]
        at org.candlepin.controller.CandlepinPoolManager.deletePool(CandlepinPoolManager.java:1590) ~[CandlepinPoolManager.class:na]
        at com.google.inject.persist.jpa.JpaLocalTxnInterceptor.invoke(JpaLocalTxnInterceptor.java:58) ~[guice-persist-3.0.jar:na]
        at org.candlepin.controller.CandlepinPoolManager.removeAndDeletePoolsOnOtherOwners(CandlepinPoolManager.java:563) ~[CandlepinPoolManager.class:na]
        at org.candlepin.controller.CandlepinPoolManager.refreshPoolsForSubscription(CandlepinPoolManager.java:502) ~[CandlepinPoolManager.class:na]
        at com.google.inject.persist.jpa.JpaLocalTxnInterceptor.invoke(JpaLocalTxnInterceptor.java:66) ~[guice-persist-3.0.jar:na]
        at org.candlepin.controller.CandlepinPoolManager.refreshPoolsWithRegeneration(CandlepinPoolManager.java:192) ~[CandlepinPoolManager.class:na]

server/spec/refresh_pools_spec.rb
vritant: not really. candlepin is just too slow.

vritant:

vritant:

server/spec/refresh_pools_spec.rb
vritant: changed the spec test to use the async api.

server/spec/refresh_pools_spec.rb
vritant: changed the spec test to use the async api.

mstead:
retest this please

server/spec/candlepin_scenarios.rb
```diff
@@ -486,13 +605,8 @@ def create_candlepin_export_update
...
     end_date = Date.new(2025, 5, 29)
-    @cp.create_subscription(@owner['key'], product1.id, 12, [], '', '12345', '6789', nil, end_date)
-    @cp.create_subscription(@owner['key'], product2.id, 14, [], '', '12345', '6789', nil, end_date)
-
-    @cp.refresh_pools(@owner['key'])
-
-    pool1 = @cp.list_pools(:owner => @owner.id, :product => product1.id)[0]
-    pool2 = @cp.list_pools(:owner => @owner.id, :product => product2.id)[0]
+    pool1 = create_pool_and_subscription(@owner['key'], product1.id, 12, [], '', '12345', '6789', nil, end_date)
+    pool2 = create_pool_and_subscription(@owner['key'], product2.id, 14, [], '', '12345', '6789', nil, end_date)
```
mstead: Refresh pools getting called twice here? Could use the skip flag on the first.

server/spec/cert_revocation_spec.rb
```diff
@@ -12,10 +12,8 @@
     @monitoring_prod = create_product random_string('monitoring')
 
     #entitle owner for the virt and monitoring products.
-    @cp.create_subscription(@owner['key'], @monitoring_prod.id, 6)
-    @cp.create_subscription(@owner['key'], @virt_prod.id, 3)
-
-    @cp.refresh_pools(@owner['key'])
+    create_pool_and_subscription(@owner['key'], @monitoring_prod.id, 6)
+    create_pool_and_subscription(@owner['key'], @virt_prod.id, 3)
```
mstead: Refresh pools getting called twice here? Could use the skip flag on the first.

server/spec/consumer_resource_activation_key_spec.rb
```diff
@@ -141,10 +137,9 @@
   it 'should allow a consumer to register with an activation key with an auto-attach across pools' do
     # create extra product/pool to show selectivity
     prod1 = create_product(random_string('product1'), random_string('product1'))
-    subs1 = @cp.create_subscription(@owner['key'], prod1.id, 1)
-    subs2 = @cp.create_subscription(@owner['key'], prod1.id, 1)
-    subs3 = @cp.create_subscription(@owner['key'], prod1.id, 2)
-    @cp.refresh_pools(@owner['key'])
+    create_pool_and_subscription(@owner['key'], prod1.id, 1)
+    create_pool_and_subscription(@owner['key'], prod1.id, 1)
+    create_pool_and_subscription(@owner['key'], prod1.id, 2)
```
mstead: Refresh pools getting called three times here? Could use the skip flag on the first two.

server/spec/consumer_resource_host_guest_spec.rb
```diff
@@ -222,11 +222,9 @@
     provided_product = create_product(random_string('product'),
       random_string('product'),
       {:owner => @owner1['key']})
-
     
-    @cp.create_subscription(@owner1['key'], vip_product.id, 10, [provided_product.id])
-    @cp.create_subscription(@owner1['key'], std_product.id, 10, [provided_product.id])
-    @cp.refresh_pools(@owner1['key'])
+    create_pool_and_subscription(@owner1['key'], vip_product.id, 10, [provided_product.id])
+    create_pool_and_subscription(@owner1['key'], std_product.id, 10, [provided_product.id])
```
mstead: Refresh pools getting called twice here? Could use the skip flag on the first.

server/spec/candlepin_scenarios.rb
vritant: not in this case. if we use the skip flag , we do not return the pool reference.

PR #1005, opened 2015-10-23
1269045: Gutterball DB Growth Fix

mstead:
https://bugzilla.redhat.com/show_bug.cgi?id=1269045

Also, this PR is lacking test coverage due to time constraints. I don't like this, but I'll try and get something in while waiting for feedback from the review.

**Pull Request Details**
Changed event new/oldentity column type to mediumtext
so that large objects will not be used by the database.

Using Lobs made hibernate use large object storage which
caused disk space cleanup issues with postgres and
would leave the large objects around even after an event
record was deleted.

Using mediumtext in hibernate and in liquibase allows storing
strings of 16MB (size limited text column for postgres), and should be
plenty of space to store our event JSON.

Events that are not handled will not be stored. When handling events, the JSON for Event.new/oldEntity will only be stored until the event has been fully processed, at which point it is nullified.

I've put in a quick and easy event cleaning task. We can turn this into a framework later, or replace with quartz. Just wanted something in place to do some automatic cleanup.

I also did a column rename as it was causing issues in mysql.

**Some Things To Test**
1. Please make sure to test the DB upgrade with events in the DB.
2. Make sure that existing events are cleaned out. Note though that there is no automatic cleanup for large objects -- this may have to be a manual step as vacuum_lo tool is part of another package which may not be installed. May have to provide steps to clean this up.


gutterball/src/main/java/org/candlepin/gutterball/model/Event.java
```diff
@@ -310,4 +306,9 @@ public void setStatus(Status status) {
         this.status = status;
     }
 
+    public void clearJsonFields() {
```
Ceiu: Would it make sense to automatically call this method via on-persist/commit hook, depending on the event state?

gutterball/src/test/java/org/candlepin/gutterball/junit/GutterballLiquibaseResource.java
```diff
@@ -158,5 +158,8 @@ public void truncateLiquibaseSchema() {
     public void createLiquibaseSchema() {
         exec("CREATE SCHEMA LIQUIBASE");
         database.setLiquibaseSchemaName("LIQUIBASE");
+
+        // HSQLDB does not support mediumtext, so we fake it here.
+        exec("CREATE TYPE MEDIUMTEXT AS VARCHAR(100000)");
```
Ceiu: Interesting tidbit I learned while looking up if HSQLDB has a limit on varchar: You can use K, M or G in the length specifier, like VARCHAR(10k)
http://hsqldb.org/doc/guide/sqlgeneral-chapt.html#sgc_char_types

gutterball/src/main/java/org/candlepin/gutterball/model/Event.java
mstead: Good idea. I can look into that.

Ceiu:
Since there's so much to test, I'm going to add comments as I go.

To prime the DB for testing, I ran the entire rspec suite with amqp enabled so I'd have plenty of data sitting around.

To start with:
- The deploy with events present went fine. Only one event remained after the liquibase run, and it was a CREATED event in the RECEIVED state.
- On Postgre, the newentity/oldentity fields now appear to be TEXT fields rather than MEDIUMTEXT. Not sure if that in itself is an issue, but the concerning bit is that the fields are still populated with a number. I'm hoping that's just holdover from the old version of the table.

mstead:
> To start with:
> 
>     The deploy with events present went fine. Only one event remained after the liquibase run, and it was a CREATED event in the RECEIVED state.
>     On Postgre, the newentity/oldentity fields now appear to be TEXT fields rather than MEDIUMTEXT. Not sure if that in itself is an issue, but the concerning bit is that the fields are still populated with a number. I'm hoping that's just holdover from the old version of the table.
This is the event from above. postgres does not have a 'mediumtext' column. Liquibase translates it to a TEXT field for postgres. Hibernate will put a size constraint on it for postgres.

Ceiu:
MySQL/MariaDB takes a bit of a hit, since the entity fields were originally LONGTEXT that get downsized to a MEDIUMTEXT, 

- Table is updated properly on initial deploy
- Events are removed on deploy successfully
- Event entities are cleared once the event is processed
- Events are cleared as they age out and the cleanup task is run

ACK

Ceiu:
ACK

Existing events are purged on deploy.

mstead:
Ready to merge once tests pass and prove OK.

vritant:
Error reported by jenkins:
```
   [junit] SLF4J: Failed toString() invocation on an object of type [org.candlepin.model.Entitlement]
    [junit] java.lang.NullPointerException
    [junit] 	at org.candlepin.model.Entitlement.toString(Entitlement.java:274)
```

this is fixed in #1001    7bf1b55f7268ef4a23edaf1cd311c68ead0be370

PR #1007, opened 2015-10-27
Create new event for Entitlement Expiration

wottop:
Appends message with information that the expiration occurred when the
guest did not have a host mapping


cnsnyder:
Looks good, ACK

PR #1011, opened 2015-11-02
Audit filtering

nguyenfilip:
Adding filtering of Audit Events. See javadoc of the following: class EventFilter, ConfigProperties.AUDIT_FILTER*


nguyenfilip:
Ah we probably do not want to pull this into master

PR #1015, opened 2015-11-11
1274074: Fix deadlock during highly concurrent entitlement jobs.

awood:
This PR should eliminate the possibility of deadlock on cp_consumer during a bind by pool.  It can, however, cause lock wait timeouts especially in hosted mode.  If several bind by pool jobs are all created at once, the compliance checking portion of CandlepinPoolManager.getStatus can easily result in some of the jobs not being able to acquire a lock in time.

The additional commits add a mechanism to throttle the number of concurrent Entitler jobs and to dispense with compliance checking on distributors.

### DIFF
--- a/buildfile
+++ b/buildfile
@@ -192,8 +192,6 @@ define "candlepin" do
 
   compile.options.target = '1.6'
   compile.options.source = '1.6'
-  # Enable to turn on compiler linting
-  # compile.options.lint = 'all'
 
   # path_to() (and it's alias _()) simply provides the absolute path to
   # a directory relative to the project.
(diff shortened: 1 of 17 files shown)


awood:
jenkins, retest this please

nguyenfilip:
For the first commit "Fix deadlock during highly concurrent...." , we dont really have a local reproducer. I will verify (on Postgres) that the consumer gets locked on the database level, when lockAndLoad()

nguyenfilip:
"Fix deadlock during highly concurrent" verified by looking up locks on DB level: 
select id from cp_consumer where id =$1 for update

nguyenfilip:
Verified "Limit the number of binds per consumer that can run concurrently". I set the throttling to 1 and created 2 consumers. The logic does allow only 1 Entitler job per consumer.

nguyenfilip:
Verified "Do not run compliance rules on distributors."

PR #1016, opened 2015-11-16
Added correction to the output of some getters in Product; few misc bug fixes

Ceiu:
- getAttribute, getProductContent and getDependentProductIds will now
  set their collections if it is null at the time it is called
- Product.toString no longer causes NPEs if it is called before the
  product has an owner set
- Added additional logging to ProductCurator
- CandlepinPoolManager now sets the owner of a product even when it
  already exists

Not entirely sure if this will prove useful long-term. I feel like I recall some code that explicitly looked for a null value to perform certain actions. If so, and if the old code proves to be a better solution, this commit may need to be reverted/discarded.


vritant:
@Ceiu I had already tried LazyCollectionOption.EXTRA. I had spend 2 days on this issue and spend a lot of Alex's hours too, in the end we were happy with eager fetch. I'd like to suggest we leave it.

PR #1017, opened 2015-11-17
1198178: Adds api method to remove all entitlements for a consumer by…

cnsnyder:
… pool id

### DIFF
--- a/server/spec/unbind_spec.rb
+++ b/server/spec/unbind_spec.rb
@@ -25,6 +25,27 @@ describe 'Unbind' do
     consumer.list_entitlements.should be_empty
   end
 
+  it 'should remove entitlements by pool id without touching other entitlements' do
+    consumer = consumer_client(@user, 'consumer')
+
+    testing = create_product(nil, random_string('testing'))
+    @cp.create_subscription(@owner['key'], testing.id, 4)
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1743,6 +1743,35 @@ public void unbindBySerial(
...
+     * <p>
+     * By Pool Id
+     *
+     * @httpcode 403
+     * @httpcode 404
+     * @httpcode 200
+     */
+    @DELETE
+    @Produces(MediaType.WILDCARD)
+    @Path("/{consumer_uuid}/entitlementsByPoolId/{pool_id}")
```
wottop: I think I would go with "/{consumer_uuid}/entitlements/pool/{pool_id}"

server/spec/unbind_spec.rb
```diff
@@ -80,6 +101,4 @@
     # None of the serials should be revoked
     serials.each { |serial| @cp.get_serial(serial).revoked.should be_false }
   end
-end
-
-
+end
```
wottop: Needs newline at end of file

wottop:
Tests pass on my machine

PR #1018, opened 2015-11-17
Updated the per-org products migration for performance with lots of data

Ceiu:
- The per-org products upgrade task now verifies that it actually
  migrated a given product before using it in further queries
- All of the constraints and foreign keys are added as a separate step
  in the 2.0 migration, performed after all the data has been migrated
- When running on MySQL or MariaDB, the foreign key checks are disabled
  during the migration


mstead:
retest this please

mstead:
retest this please

mstead:
retest this please

PR #1022, opened 2015-11-19
Updated CandlepinPoolManager to properly update owners during refresh

Ceiu:
- When refreshing products and content, the resolved owner will now
  always be set
- Product resolution will now throw an exception if the product is
  not resolved properly
- Improved warnings when multiple versions of the same product or
  content are found on a single subscription


vritant:
Time is of the essence for this PR, so TODOs are reasonable.

PR #1023, opened 2015-11-19
Updated property definition in Gutterball Event changeset

Ceiu:
- Updated the entitydata.type property definition in the event entity
  changeset to now use "mediumtext" for mysql and hsqldb, but "text"
  for everything else. This fixes several unit test failures stemming
  from various side-effects of not having a default type.


wottop:
ack

PR #1024, opened 2015-11-19
Added a delay in the guest entitlement spec tests

Ceiu:
- Added an additional delay in the one_sub_pool_per_stack_spec test
  'should remove guest entitlement when guest is migrated' to help
  alleviate some periodic failures when testing on MySQL

### DIFF
--- a/server/spec/one_sub_pool_per_stack_spec.rb
+++ b/server/spec/one_sub_pool_per_stack_spec.rb
@@ -342,6 +342,9 @@ describe 'One Sub Pool Per Stack' do
     # Simulate migration
     @host2_client.update_consumer({:guestIds => [{'guestId' => @guest_uuid}]})
 
+    # Need to wait a moment here for MySQL to catch up
+    sleep 2
+
     # Guest entitlement should now be revoked.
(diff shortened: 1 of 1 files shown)


server/spec/one_sub_pool_per_stack_spec.rb
```diff
@@ -342,6 +342,9 @@
     # Simulate migration
     @host2_client.update_consumer({:guestIds => [{'guestId' => @guest_uuid}]})
 
+    # Need to wait a moment here for MySQL to catch up
+    sleep 2
```
barnabycourt: This seems error prone if the system is under a lot of load. Is there an alternative method to watch for completion of the update_consumer?

Ceiu: Nothing is jumping out at me from looking through this. There's nothing (intentionally) async going on here, and as such, there's nothing to really check. The lack of a return from that particular request doesn't help things any, either.

The issue itself stems from the two requests coming from different clients, and only occurs with MySQL. If the second request (list_entitlements) comes from the host2_client, the problem goes away there as well.

barnabycourt:
ACK

PR #1026, opened 2015-11-20
Ignore milliseconds for unbind by pool test

cnsnyder:
This PR adds a new ruby utility method 'remove_matching!' that recursively removes all strings contained within a given object while maintaining the order of the elements.
This new function is then used to remove all datetime like values from the broken unbind test so that we can pass on mysql.

NOTE: The reason we need to remove these values to begin with is we lose the millisecond value of the object after it has been committed to and retrieved from a mysql database.

### DIFF
--- a/server/client/ruby/util.rb
+++ b/server/client/ruby/util.rb
@@ -1,38 +1,19 @@
+require 'date'
+
 # A file for utility functions and classes that are miscellaneous
-def remove_matching!(object, pattern)
-    # Recursively removes all string items matching a given regexp,
-    # deleting key's where appropriate
+def fix_times!(object, list_of_keys_to_check)
+    # Recursively fixes times for the list of keys provided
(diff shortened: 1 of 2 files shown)


dgoodwin:

awood:
That regex is a real mind-bender.  I'm a little confused why your reference JSON has milliseconds in it since that JSON looks to have come back from Candlepin too.  Can you give an example of the mismatch?

nguyenfilip:
I will do some inline notes related to naming.

However I also have a broader question here that challanges the need of this PR. Since we want to grant cross-database support, our REST methods should return JSONs without miliseconds in the first place. I dont think there is any reason to have miliseconds in database anyway. Would it be reasonable reasonable to modify our Entitlement/Pool curator so that it stores the entities into database without the miliseconds? 

If we do that, maybe we won't have to implement such utility function.

PR #1027, opened 2015-11-23
Fixed a bug in the version detection and improve the Oracle container

Ceiu:
- The container version detection no longer attempts to run in tty mode
- The Oracle container no longer has a custom version of the build_container
  script and, instead, uses a custom.sh which is now invoked by the global
  build_container.sh script if present


wottop:
ack

PR #1028, opened 2015-11-23
1283401:  500 status (or an empty reply) when getting all consumers

vritant:
* simplest solution for this BZ is to simply enforce that the user mentions at least one search critera
* currently none of the consumers call this API without any search criteria ( please confirm )
* This PR does NOT address the problem where using a single criteria can match hundreds of thousands of consumers, but I don't think enforcing a default page limit is a good idea, as it may break existing consumers.
* if this implementation plan is acceptable, we could change [python-rhsm](https://github.com/candlepin/python-rhsm/blob/5426aca93722f562cae7d6245999d6e874d80f04/src/rhsm/connection.py#L1013) to not let owner be optional.

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -825,11 +825,7 @@ class Candlepin
   def list_consumers(args={})
     query = "/consumers?"
 
-    # Ideally, call this method with an owner parameter. Only
-    # super admins will be able to call without and list all
-    # consumers.
-    query = "/owners/#{args[:owner]}/consumers?" if args[:owner]
-
+    query << "owner=#{args[:owner]}&" if args[:owner]
     query << "username=#{args[:username]}&" if args[:username]
     query << "type=#{args[:type]}&" if args[:type]
     query << "page=#{args[:page]}&" if args[:page]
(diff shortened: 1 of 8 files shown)


cnsnyder:
I think we should look more into what the reasoning for this curl request is before accepting this solution. I'm not sure what the utility of getting all consumers via the aforementioned API is.
Would pagination solve this issue? What consumers would it break? Could we look at streaming (from the db to cp to the consumer) to reduce the memory consumption?

As far as I can tell, there are not many requests in production to GET /consumers/ (only 4 in 21+ million). I'm not sure if blocking this API by requiring at least one search criteria really solves the issue or if it will simply result in those who want this information making more requests to get it (I doubt that though).

vritant:
* @cnsnyder good point, I should have asked Shayne, thanks for asking him! I had looked at all current versions of candlepin consumers, and did NOT find a call that searched for consumers without any criteria, so I assumed this was not a real world scenario. based on Shayne's response to your question in the BZ, it seems okay to restrict a search without any criteria.
* There is already pagination on this API, but we don't *enforce* any page limit in candlepin REST APIs. I think if there is an issue reported on the clients in the future, we should rely on the client to use the pagination parameters. thoughts?
* the intent for this solution was to do the minimum we can, it seems safe to restrict a criteria-less search, we shouldn't need to list ALL the consumers in a candlepin.

cnsnyder:
After hearing your thoughts and Shayne's, I am happy with this as a solution to BZ 1283401. Asking clients who are having difficulty to use pagination parameters seems reasonable. Given the above, ACK.

PR #1031, opened 2015-11-30
Limit the number of binds per consumer that can run concurrently.

awood:
I plan on submitting this as a PR to master as well.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/tasks/EntitlerJob.java
+++ b/server/src/main/java/org/candlepin/pinsetter/tasks/EntitlerJob.java
@@ -14,12 +14,14 @@
  */
 package org.candlepin.pinsetter.tasks;
 
-import static org.quartz.JobBuilder.*;
+import static org.quartz.JobBuilder.newJob;
 
+import org.candlepin.common.config.Configuration;
 import org.candlepin.controller.Entitler;
 import org.candlepin.model.Consumer;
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/candlepin/pinsetter/tasks/EntitlerJob.java
```diff
@@ -82,9 +89,33 @@ public static JobDetail bindByPool(String poolId, Consumer consumer, Integer qty
         JobDetail detail = newJob(EntitlerJob.class)
             .withIdentity("bind_by_pool_" + Util.generateUUID())
             .requestRecovery(false) // do not recover the job upon restarts
+            .storeDurably()
```
mstead: Any particular reason why you had to set this? Just curious.

awood: It's required for trigger-less jobs.  You get an error if it is not set.

mstead:
retest this please

mstead:
To make sure that I understand this PR, you are basically checking how many EntitlerJobs are running for an Owner, and if the threshold is reached, queue the job to be executed later?

mstead:
@awood If my understanding of the PR above is correct, then this looks Ok to me.

Couple of things:
* please add some detail about the fix in the commit message
* Spec test failures seem to be due to Jenkins.

mstead:
ACK.

PR #1034, opened 2015-12-02
Updated artifact collection to ensure it is always performed

Ceiu:
Moved the artifact collection to the cleanup step to ensure it is always performed and added additional checks to ensure the cleanup isn't performed multiple times, nor that absent artifacts don't halt the collection process.


wottop:
ack

PR #1036, opened 2015-12-03
1287609: Fixing race condition

nguyenfilip:
Making DynamicPropertyFilter thread safe and adding a few unit tests for it since they were missing. Because this class wasn't thread safe a user could hit serialization server side error when invoking 
```
consumers/<UUID>/entitlements?exclude=certificates.key&exclude=certificates.cert
```


nguyenfilip:
Just to note. I was considering using ThreadLocal, but I don't think the list initialization is the biggest bottleneck in the class anyway. So I've chosen the simplest solution.

wottop:
ack. please make branches off of master for future pull requests. e.g. nguyenfilip/race

PR #1038, opened 2015-12-07
translation typos BZ1190814

vritant:
fixed typos as suggested in the BZ both in message IDs and message Strings, with the exception of "Too many content sets for certificate {0}. A newer client may be available to address this problem. See knowledge database https://access.redhat.com/knowledge/node/129003 for more information."
here, the change 'kbase' to 'knowledge base' could change the translation as many translations have literally enunciated "kbase" in their respective alphabet.

### DIFF
--- a/common/po/as.po
+++ b/common/po/as.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2015-12-03 12:24-0500\n"
+"POT-Creation-Date: 2015-12-07 00:44-0500\n"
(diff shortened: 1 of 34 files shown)


common/po/as.po
```diff
@@ -185,14 +183,14 @@ msgstr ""
 #: gutterball/src/main/java/org/candlepin/gutterball/report/StatusTrendReport.java:129
 msgid ""
 "Filter on subscriptions which have management enabled set to this value "
-"(boolean)"
+"(Boolean)"
```
Ceiu: This change should coincide with the change in the code. If the lookup is case sensitive (map lookups are, typically), this string won't be found.

Ceiu:
@candlepin-pull-request-bot add to whitelist

Ceiu:
ACK

PR #1039, opened 2015-12-07
(0.9.51) Disabling spec tests that are intended for standalone mode

mstead:
Some spec tests were failing in 0.9.51 in hosted mode that were intended to test standalone functionality. Only run these tests in standalone mode.


vritant:
NACK. please consider #1041

PR #1042, opened 2015-12-08
Port Developer Subscriptions from master to 0.9.51

mstead:
This patch backports the developer subscription feature from the
master branch. This is a net new commit since the code in master
was added in a bunch of different commits and the merge was not
very smooth.

In hosted mode, a developer subscription will be granted to any
consumer during autobind that will cover all of the consumer's
installed products. An owner must have and active entitlement and
a match must be made on the dev_sku specified by the consumer
in order for the entitlement to be granted. The developer
entitlement comes from a special pool that is created
for that consumer only and will be deleted along with the
entitlement on manual removal or expiry. Multiple runs of the
auto-attach process that detect a change in coverage, will cause
the entitlement/pool to be removed and a new one created. The
duration of the entitlement/pool will be the consumer registration
date + expires_after duration attribute on the dev_sku product.

### DIFF
--- a/server/bin/test_data.json
+++ b/server/bin/test_data.json
@@ -1756,6 +1756,20 @@
             "provided_products": [
                 "900"
             ]
+        },
+        {
+            "name": "Dev Product",
+            "id": "dev-sku-product",
+            "type": "MKT",
+            "skip_subs": "true",
+            "attributes": {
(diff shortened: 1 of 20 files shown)


wottop:
ack

PR #1045, opened 2015-12-11
Adds the UTC timezone to the date serialization test.

cnsnyder:
This PR updates the test for date serialization to include the now expected UTC timezone.


wottop:
ack

PR #1046, opened 2015-12-11
Test milliseconds or not

wottop:
The test was ammeded to show that the incoming JSON will get properly parsed
whether it contains milliseconds in its time format or not.


Ceiu:
ACK

PR #1047, opened 2015-12-14
 Implementing some CDK improvements

mstead:
* Fixed dev pool lookups
* Delete dev pool on each auto attach and recreate it.
* Calculate dev pool end-date based on the consumer's
  registration time.


server/bin/test_data.json
```diff
@@ -1983,6 +1983,7 @@
             "name": "Development SKU Product",
             "id": "dev-sku-product",
             "type": "MKT",
+            "skip_pools": "true",
```
wottop: I am not familiar with this attribute. Is it part of the CDK stuff?

mstead: Nope. It was put in for the derived pool (VDC) stuff. It just creates the Product and skips sub/pool creation. This puts the test data in a state that is more like the CDK setup.

wottop:
ack

PR #1051, opened 2015-12-18
1292877: Do not delete dev pool before we are done with it

mstead:
Since we were deleting the dev pool before the derived products
were loaded (lazy), we couldn't look them up when addressing
modified entitlements.

Moved the dev pool deletion to the end of the remove entitlement
process to be sure we were done with it.


wottop:
buildbot: retest this please

wottop:
ack

PR #1053, opened 2016-01-08
 	Updated the DataCleanupTask and its scheduling mechanism

Ceiu:
- The DataCleanupTask now uses cron-like scheduling in place of the
  interval with units scheduling
- Events and compliance snapshots are now both cleaned up as part
  of the cleanup task
- Several tests added


gutterball/src/main/java/org/candlepin/gutterball/curator/ComplianceSnapshotCurator.java
```diff
@@ -1149,5 +1157,52 @@ private int compareDatesByDate(Date date1, Date date2) {
...
+     * Deletes any compliances older than the specified duration in minutes.
+     *
+     * @param minutes
+     *  The number of minutes to use for cleaning up compliances
+     *
+     * @return
+     *  the number of compliance snapshots deleted during the cleanup operation
+     */
+    @Transactional
+    public int cleanupCompliances(int minutes) {
```
wottop: Does making this transactional mean that if there are a thousand compliances and one will not delete, none will?

Ceiu:
Should. At least, that's the intent.
I should probably look into how @Transactional interacts with .flush so I can better answer your question.

gutterball/src/main/java/org/candlepin/gutterball/curator/ComplianceSnapshotCurator.java
Ceiu: Should. At least, that's the intent.
I should probably look into @Transactional and how it interacts with .flush so I can better answer your question.

wottop:
I am good with this whenever you get your quetions answered.

Ceiu:
I made some changes after chatting with @nguyenfilip. Currently setting up to profile to make sure memory usage stays reasonable.

Ceiu:
Alright, this should be the last version. Commits have been squashed, so I probably broke any checked out branches. Delete and try again, I suppose.

wottop:
ack

PR #1054, opened 2016-01-08
1294044, 1294049 : Allow Subscription Manager Gui register process to work with…

wottop:
… Dev client

Stop the return of all available service levels and just return "". This way only
one dry run is attempted. The dev pool is created at dry run and available to use
on the bind step.

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -892,7 +892,7 @@ class Candlepin
     delete("/consumers/#{consumer_id}/entitlements/pool/#{pool_id}")
   end
 
-  def autobind_dryrun(consumer_id, service_level=nil)
+  def autobind_dryrun(consumer_id=nil, service_level=nil)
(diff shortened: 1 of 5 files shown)


wottop:
retest this please

wottop:
Going for a rebase.

wottop:
buildbot: restest this please

cnsnyder:
@wottop This does not work for me as a fix to bz 1294044. I can reproduce the behavior described in the bz in master but this PR does not produce the expected results as posted in the original BZ. Are there additional changes necessary on the subscription-manager side to get the gui working with the return value of "" from this PR?

cnsnyder:
This PR also fixes BZ 1294049 (Both of these bugs had the same root cause).

This works for me. ACK.

cnsnyder:
@wottop Could you please edit the commit message to include both bz numbers?

wottop:
It does on two lines

PR #1055, opened 2016-01-08
1295452: Look up dev sku by consumer only

mstead:
Pulled from 0.9.51 into master.

Since there should only ever be one dev pool per
consumer, look up the existing pool by dev_pool and
requires_consumer attributes, ignoring the dev_sku
fact. Otherwise, a change in the dev_sku fact will
result in multiple dev pools.


wottop:
ack

PR #1058, opened 2016-01-15
Several improvements made to the CP2 migration task

Ceiu:
- Created a new abstract classes, LiquibaseCustomTask and
  LiquibaseCustomTaskWrapper, which contains common functionality
  many custom tasks utilize
- Updated the per-org products and pool-type upgrade tasks to use
  the new abstract superclass
- Began splitting the per-org products update task into two tasks:
  one to do the initial migration of data without affecting existing
  data, and another to do destructive changes to shared tables
- The first part of the per-org products task can be run repeatedly
- Log output has been further improved
- Migration now prefetches content and products to cut down on
  the number of per-org queries
- Logging has been updated to be slightly more friendly
- The setNull patch for Oracle has been improved to only use the
  VARCHAR type when the database actually appears to be Oracle.
- The per-org products migration now uses bulk-inserts where applicable to
  cut down on the amount of individual queries being run
- Fixed several logging mismatches from the reversion away from the
  migrate + update prototype
- Unique checks are now also temporarily disabled during migration on
  MySQL


wottop:
Looks good. Runs against all 3 DB's effectively. Normally, I'd go for the squash, but this is a big checkin.

ACK

PR #1059, opened 2016-01-15
Implementation of revokeEntitlements(List)

nguyenfilip:
Goal of this  change is to speed up Import Manifest in Standalone Mode

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -177,13 +177,16 @@ public class CandlepinPoolManager implements PoolManager {
         // remove everything that isn't actually active
         subIds.removeAll(deletedSubs);
         // delete pools whose subscription disappeared:
-
+        List<Pool> poolsToDelete = new ArrayList<Pool>();
         for (Pool p : poolCurator.getPoolsFromBadSubs(owner, subIds)) {
             if ((p.getType() == PoolType.NORMAL || p.getType() == PoolType.BONUS) &&
                 (ueberPoolId == null || !ueberPoolId.equals(p.getId()))) {
-                deletePool(p);
+                poolsToDelete.add(p);
             }
         }
+
+        deletePools(poolsToDelete);
+
 
         // TODO: break this call into smaller pieces.  There may be lots of floating pools
         List<Pool> floatingPools = poolCurator.getOwnersFloatingPools(owner);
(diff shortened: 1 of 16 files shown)


nguyenfilip:
Rebasing this on @mstead fix for unit test failures.

mstead:
retest this please

mstead:
retest this please

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -177,13 +177,16 @@ void refreshPoolsWithRegeneration(Owner owner, boolean lazy) {
...
+        List<Pool> poolsToDelete = new ArrayList<Pool>();
         for (Pool p : poolCurator.getPoolsFromBadSubs(owner, subIds)) {
             if ((p.getType() == PoolType.NORMAL || p.getType() == PoolType.BONUS) &&
                 (ueberPoolId == null || !ueberPoolId.equals(p.getId()))) {
-                deletePool(p);
+                poolsToDelete.add(p);
             }
         }
+        if (!poolsToDelete.isEmpty()) {
+            deletePools(poolsToDelete);
```
mstead: Delete pools has an empty do nothing check -- up to yourself, but we could remove the if here.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -364,6 +481,33 @@ public Entitlement replicate(Entitlement ent) {
...
+            ent.getCertificates().clear();
+            getEntityManager().remove(ent);
+        }
+
+        // Maintain runtime consistency.
+        for (Entitlement ent : entitlements) {
+            ent.getCertificates().clear();
+            // Pool.entitlements is EXTRA Lazy. Before removing anything
+            // from that collection, its necessary to initialize it
+            Hibernate.initialize(ent.getPool().getEntitlements());
```
mstead: Just for my information, is this something that is commonly done? Is it just the nature of how our models are set up?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1250,6 +1377,30 @@ public void deletePool(Pool pool) {
...
+
+        List<Entitlement> entitlementsToRevoke = new ArrayList<Entitlement>();
+
+        for (Pool p : pools) {
+            entitlementsToRevoke.addAll(p.getEntitlements());
+        }
+
+        if (!pools.isEmpty()) {
+            revokeEntitlements(entitlementsToRevoke);
+            poolCurator.batchDelete(pools);
```
mstead: Again up to you, but the method calls are safe without the !empty check here.

nguyenfilip:
Updated docs for listBySourceEntitlements

server/src/main/java/org/candlepin/model/EntitlementCurator.java
nguyenfilip: I think EXTRA lazy is just buggy when combined with CascadeType.CREATE (yes CREATE, its not a typo :-) 

I created a unit test that shows this weird behavior (EntitlementCuratorTest.removeConsumerWithEntitlements) you can run it and debug it yourself. Without calling Hibernate.initialize it will fail with nasty SQL exceptions. The reason is, I think, that the if you remove the entitlement from the extra lazy collection, Hibernate will create it again due to CREATE cascading.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: amended

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: amended

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1201,34 +1204,158 @@ void removeEntitlement(Entitlement entitlement,
...
+
+        log.info("Batch revoking entitlements: " + entsToRevoke.size());
+
+        entsToRevoke =  new ArrayList<Entitlement>(entsToRevoke);
+
+        List<Pool> poolsToDelete = poolCurator.listBySourceEntitlements(entsToRevoke);
+
+        for (Pool pool : poolsToDelete) {
+            entsToRevoke.addAll(pool.getEntitlements());
+        }
```
awood: This is a little confusing to me.  We're gathering up all the sibling entitlements here and revoking them too?

mstead:
Code looks ok to me. Hot spots for me are around modified products/entitlements.

Moving on to more manual testing of the functionality.

mstead:
There is a patch available for fixing the spec tests for this PR. It was fixed in other branches but wasn't back ported. I'll try and pull that in with another commit.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1201,34 +1204,158 @@ void removeEntitlement(Entitlement entitlement,
...
+            batch.add(entitlement);
+
+            // We work in batches of maximum size 1000.
+            if (i % 1000 == 0) {
+                Set<Entitlement> modifiedEnts = entitlementCurator.batchListModifying(batch);
+                this.regenerateCertificatesOf(modifiedEnts, true);
+                batch.clear();
+            }
+        }
+        log.info("Modifier entitlements done.");
```
awood: Should this go down a few lines below the `if(!batch.isEmpty())`?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1201,34 +1204,158 @@ void removeEntitlement(Entitlement entitlement,
...
+
+        if (!batch.isEmpty()) {
+            this.regenerateCertificatesOf(entitlementCurator.batchListModifying(batch), true);
+        }
+
+        Set<Consumer> distinctConsumers = entitlementCurator.getDistinctConsumers(entsToRevoke);
+
+        log.info("Recomputing status for " + distinctConsumers.size() + " consumers.");
```
awood: Maybe

```
log.info("Recomputing status for {} consumers.", distinctConsumers.size());
```

Not a big difference here since toString() on the size is going to be really fast, but I just like to keep everything consistent.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1201,34 +1204,158 @@ void removeEntitlement(Entitlement entitlement,
...
+
+        log.info("Starting batch delete of pools and entitlements");
+        poolCurator.batchDelete(poolsToDelete);
+        entitlementCurator.batchDelete(entsToRevoke);
+        entitlementCurator.flush();
+        log.info("All deletes flushed successfully");
+
+        List<Entitlement> stackingEntitlements = filterStackingEntitlements(entsToRevoke);
+
+        for (Entitlement ent : stackingEntitlements) {
```
awood: It seems like we are looping through entsToRevoke (or a subset thereof) three times: the first loop to change the pool consumed values, the second to search for stacking entitlements, and the the third to loop through the list of stacking entitlements we just received.  Can we consolidate?

```
for (Entitlement e : entsToRevoke)
  //get pool
  //adjust pool quantity
  if (isStackingEntitlement(e)) {
    //do stacking stuff or append to stackingEntitlements list
  }
}

// do stacking stuff here if not possible in previous loop
```

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1201,34 +1204,158 @@ void removeEntitlement(Entitlement entitlement,
...
     @Transactional
     public int removeAllEntitlements(Consumer consumer) {
-        int count = 0;
-        for (Entitlement e : entitlementCurator.listByConsumer(consumer)) {
-            removeEntitlement(e, false);
-            count++;
-        }
-        return count;
+        List<Entitlement> entsToDelete = entitlementCurator.listByConsumer(consumer);
+        revokeEntitlements(entsToDelete);
```
awood: This method is `removeAllEntitlements` while the one above is `revokeAllEntitlements`.  Before, `revokeAllEntitlements` had that extra compliance rules run but now they have the same implementation.  That seems incorrect to me.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -202,17 +226,68 @@ private Criteria createModifiesDateFilteringCriteria(Consumer consumer, Date sta
...
+        }
+        // Retrieve all products at once from the adapter
+        List<Product> overlappingProducts =
+                productAdapter.getProductsByIds(pidEnts.getAllProductIds());
+        for (Product overlappingProduct : overlappingProducts) {
+            boolean modifies = false;
+            for (Entitlement inputEntitlement : entitlements) {
+                if (overlappingProduct.modifies(inputEntitlement.getProductId())) {
+                    modifies = true;
+                    break;
```
awood: Why are we breaking here when we are checking modifies later in the loop?  Won't that lead to `modifying.addAll` using the *next* item in the list?

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -202,17 +226,68 @@ private Criteria createModifiesDateFilteringCriteria(Consumer consumer, Date sta
...
+                }
+                Iterator<ProvidedProduct> providedProductOfInput = inputEntitlement
+                        .getPool().getProvidedProducts()
+                        .iterator();
+                // No need to continue checking once we have found a modified
+                // product
+                while (!modifies && providedProductOfInput.hasNext()) {
+                    if (overlappingProduct.modifies(providedProductOfInput.next().getProductId())) {
+                        modifies = true;
```
awood: Seems like a `break` would make sense here though.

mstead:
retest this please

mstead:
@awood  makes some good points in his review. I think we need to have a chat with @nguyenfilip about some of the changes he made to make sure that we aren't missing something. I personally don't feel confident merging this without his input.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: You are refering to listBySourceEntitlements(X)? That method is not looking for 'sibling' entitlements of X. I think the idea has been (and I kept it) that the method looks for entitlements that have been created as a consequence of X. So for example bonus pools that have been created as a consequence of entitlements X.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: You are suggesting to use log.isInfoEnabled()?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: the signature of the method is removeEntitlement(Entitlement e, boolean regenModified). The regenModified attribute when set to false used to skip regeneration and compliance status computation. The regenModified flag seems like a performance optimization, not anything functional.

I can introduce the flag again, but I thought it is not necessary because batch revoke is designed to be fast.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
nguyenfilip: This seems incorrect to me as well. I think this needs a bit more test coverage.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
awood: > You are suggesting to use log.isInfoEnabled()?

No, log.info (log.*level* generally) with SLF4J can take `{}` as place holders and will then populate them with subsequent arguments.  The place holder syntax is more concise and and log.info has a guard within it to immediately return if the log.level is higher than info.  Then end result is that toString() doesn't get called on the subsequent arguments.

So 

```
log.info("Value is " + someObject);
```

will result in a call to `someObject.toString()` before log.info is even called, since the arguments need to be resolved before they can be sent to the method.

```
log.info("Value is {}", someObject);
```

defers the cost of `toString()` until after the logger checks the level.  It's the equivalent of

```
if (log.isInfoEnabled()) {
  log.info("Value is " + someObject);
}
```

nguyenfilip:
retest this please

PR #1060, opened 2016-01-18
Awood/client refactor

awood:
Ruby bindings for the Candlepin REST API.

You will need to run a `bundle install` before attempting to use or test the bindings.

Unit tests are provided for the bindings and can be run from the `server/client/ruby/test` directory with `./test_candlepin.rb`.  (If you are using an older version of OpenSSL, the tests will print some annoying warning messages.  Those can be suppressed by running `ruby -W0 ./test_candlepin.rb`).

One unit test is currently failing ("updates an owner pool") due to an issue within Candlepin itself.

I am a little unhappy with the length of the `candelpin.rb` file.  We could split up the bindings into one resource per file and source each resource's bindings in `candlepin.rb`, but given how interrelated some of the different resources are, I don't know that having multiple source files makes life any easier.  Any thoughts on the organization are appreciated.


Ceiu:
Github isn't letting me add this as a line note, but on line 16 in test_candlepin.rb, would it make sense ot change :be_2xx to :be_successful? The latter fits the naming scheme of the other http-status-oriented matchers a bit better.

Ceiu:
It looks like all/a lot of tests moved to test_candlepin.rb ?  How will that impact things like being able to run a group of related tests from the CLI?

Ceiu:
Note to anyone reviewing/merging this branch: The jenkins rspec tests will never complete successfully, as they're configured to use the old client/spec tests.

awood:
> It looks like all/a lot of tests moved to test_candlepin.rb ? How will that impact things like being able to run a group of related tests from the CLI?

`test_candlepin.rb` is just a unit test suite for the bindings themselves.  The spec tests will remain more or less as they are.

awood:
> Github isn't letting me add this as a line note, but on line 16 in test_candlepin.rb, would it make sense ot change :be_2xx to :be_successful? The latter fits the naming scheme of the other http-status-oriented matchers a bit better.

Good point.  I'll put that on my todo list.

Ceiu:

vritant:
>I am a little unhappy with the length of the candelpin.rb file. We could split up the bindings into one resource per file and source each resource's bindings in candlepin.rb, but given how interrelated some of the different resources are, I don't know that having multiple source files makes life any easier. Any thoughts on the organization are appreciated.

@awood given this is a very thin client, I prefer the existing pattern, a single file for all resources. it's easier to traverse and explore. I am going to add [POSTMAN](https://chrome.google.com/webstore/detail/postman/fhbjgbiflinjbdggehcddcbncdddomop?hl=en) client collections for candlepin soon, and will probably follow the same pattern, and add a single collection for all APIs instead of a collection per resource.

awood:
This is ready to merge on my end.  @Ceiu @vritant

vritant:
- [ ]  Does it make sense to move the helper classes Message, JSONClient, Candlepin::Util, Candlepin::API,etc ( basically everything but the resource classes)  to their respective files?

- [ ]  JSONRequestHeaderFilter: some common code could be moved to a separate method:
```
unless opts && opts[:header] && opts[:header]['Content-Type']
       jsonify(opts)
      @header_filter.replace = true
end
```
and 
```
    # We need to strip off any leading slash on a relative URL
    u = HTTPClient::Util.urify(uri)
    if @base_url && u.scheme.nil? && u.host.nil?
      uri = uri[1..-1] if uri[0] == "/"
    end
```
- [ ] We should move / copy documentation from the APImodule to candlepinproject.org

- [ ]  raise ArgumentError.new("Bind by pool or by product but not by both") 
* Candlepin already does this and many more validation checks to the argument for the bind api. I propose we let candlepin handle those

- [ ]  we should document migration path for the ruby client and the spec tests on candlepinproject.org

- [ ]  select_from: what is the advantage of yielding if we are not processing anything after the yield? for the example in the comments, we could simply do
```
x = select_from(h, :hi)
x[:send_off] = h[:goodbye].upcase
```

- [ ] validate_keys the failure message could be misleading if block is passed, don't you think? perhaps we could add a block argument for the failure message?

- [ ] We could move the custom matchers from the test file to a utility that spec tests can use.

- [ ] We could rename candlepin.rb to candlepin_api_new.rb, or something else to indicate this is the newer client?

- [ ] debug! < this could be a debug toggle instead of ensuring its true always, what do you think?

- [ ] debug is an instance variable on the client, which means we no longer have a global debug. The newer debug is very convenient, but perhaps we could keep the option of a global debug with another static variable in addition to this? When my test_candlepin.rb failed, I had to debug! on all the clients to find out where the problem was, and wished there was an global debug

- [ ] We could create a rubocop task for buildr and add it to the check_all task

- [ ] We could create a buildr task for running test_candlepin.rb and add it to check_all task

- [ ] in a unit test context does not work for me. what am I missing?

- [ ] Please consider overriding failure messages in custom matchers so they are legible, something like::
```
RSpec::Matchers.define :be_success do
  match do |res|
    (200..206).cover?(res.status_code)
  end
  failure_message_for_should do |res|
    "URL-> #{res.header.request_method} #{res.header.request_uri}\n\
Candlepin Request Id-> #{res.headers['x-candlepin-request-uuid']}\n\
Result status-> #{res.status_code}\n#{res.reason}-> #{res.body}"
  end
end
```

- [ ] Only half of all apis are tested. do we really need to test all of them? for reference, in [this list](http://pastebin.com/LhYtCe7W) , the ones marked with "<<<" are not tested

vritant:
@awood I will create the buildr tasks, documentation for module API and migration path and removing validation from bind API in a separate PR.

PR #1061, opened 2016-01-19
1297047: Candlepin allows import of expired manifest

vritant:
* Return an ImportRecord upon import success.
* The status of the record marks if:
  * There were no active subscriptions in the manifest
  * There were one or more inactive subscriptions in the manifest

### DIFF
--- a/server/spec/import_warning_spec.rb
+++ b/server/spec/import_warning_spec.rb
@@ -1,0 +1,87 @@
+require 'spec_helper'
+require 'candlepin_scenarios'
+require 'json'
+require 'zip/zip'
+
+describe 'Import Warning', :serial => true do
+
+  include CandlepinMethods
+  include VirtHelper
+
+  before(:all) do
+    @cp = Candlepin.new('admin', 'admin')
+    pending("candlepin running in hosted mode") if is_hosted?
+
+    @cp_export = StandardExporter.new
+    @cp_export.create_candlepin_export()
+    @cp_export_file = @cp_export.export_filename
+
+    @import_owner = @cp.create_owner(random_string("test_owner"))
+    @cp.import(@import_owner['key'], @cp_export_file)
+
+    @exporters = [@cp_export]
+  end
+
+  after(:all) do
+    @cp.delete_owner(@import_owner['key'])
+    @exporters.each do |e|
+      e.cleanup()
+    end
+  end
+
+  it 'warns about inactive subscriptions' do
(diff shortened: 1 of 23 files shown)


wottop:
The bug says it should fail if expired subs. I would at least expect it to fail if all are expired.
If this is not the case based on discussions, please update the bug.

vritant:
@wottop done.

server/src/main/java/org/candlepin/audit/Eventful.java
```diff
@@ -0,0 +1,27 @@
...
+ * Marker Interface for objects which may produce events
+ */
+public interface Eventful {
+
+    /**
+     * Gets the id of an object
+     * @return The id of the object
+     */
+    Object getId();
+}
```
wottop: Why introduce another interface here? It is using the information you already have in AbstractHibernateObject. Semantics?

vritant: @wottop to let us send out events for DTOs

server/spec/import_warning_spec.rb
```diff
@@ -0,0 +1,87 @@
+require 'spec_helper'
+require 'candlepin_scenarios'
+require 'json'
+require 'zip/zip'
+
+describe 'Import', :serial => true do
```
wottop: This needs to be 'Import Warning'

wottop:
Added manifest with expired entitlement to the bug. It runs correctly. Fix the note above and I'll ack it.

wottop:
I attached a manifest to the bug that has only an expired cert. I imported it and it behaved as expected.

It also wiped out all the test data pools. If we import a fully expired manifest, do we want to leave the existing pools alone?

vritant:
@wottop I assumed that was the expected behavior as we have an [existing spec test](https://github.com/candlepin/candlepin/blob/master/server/spec/import_update_spec.rb#L45) to ensure that behavior.

wottop:
I'll accept that. it does seem like a consequence that a user will complain about. It will wipe out all subs and entitlements for no net gain. I guess they can do that now if they import the wrong file.

You still need to get the describe tag in import_warning_spec.

server/spec/import_warning_spec.rb
vritant: @wottop fixed, sorry for the delay!

wottop:
ack

PR #1062, opened 2016-01-19
Fixed the rules import spec tests on MySQL (0.9.51)

Ceiu:
- Added a delay to the rules upload function in the spec tests to
  work around a time resolution issue on databases like MySQL which
  don't store/use milliseconds


Ceiu:

PR #1064, opened 2016-01-19
cp-test no longer sends -h to older Candlepin deploy scripts

Ceiu:
If the current deploy script does not appear to have a -h flag, it will not be passed into the deploy script.

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -174,7 +174,7 @@ if [ "$DEPLOY" == "1" ]; then
 
     DEPLOY_FLAGS="-g"
 
-    if [ "$RSPEC" == "1" ] && [ "$HOSTED" == "1" ]; then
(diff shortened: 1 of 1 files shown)


docker/candlepin-base/cp-test
```diff
@@ -174,7 +174,7 @@ if [ "$DEPLOY" == "1" ]; then
 
     DEPLOY_FLAGS="-g"
 
-    if [ "$RSPEC" == "1" ] && [ "$HOSTED" == "1" ]; then
+    if [ "$RSPEC" == "1" ] && [ "$HOSTED" == "1" ] && (bin/deploy -? | grep -q -- -h); then
```
alikins: I'd probably quote the '-h' just to emphasis it's not an arg to grep. And maybe '-?' as well, just because unquoted punctuation in shell scripts looks suspicious (usage here isn't a problem though).

Ceiu: Double or single quotes?

PR #1073, opened 2016-01-26
Vritant/batch attach

vritant:


Ceiu:

nguyenfilip:
@vritant Are you planning to submit technical reference documentation on cp.org for batch attach, while the details are still fresh in your mind? Additionally to batch attach, there is a technical pattern 'Delayed compliance check' (not sure how to name it). I am referring to the ConsumerComplianceJob.scheduleStatusCheck. That pattern could also be documented separately as part of this PR.

vritant:
@nguyenfilip makes sense, I shall work on it. Also, I think we should merge this after #1093 , what do you think? I could use that code here, at least a little bit. Please don't hold off on reviewing this PR though, the only changes should be merge conflicts and a single use of revoke entitlements, and may be that commit can be reviewed separately? does that make sense?

vritant:
@nguyenfilip created candlepin/candlepinproject.org#74 for docs

nguyenfilip:
@vritant We will see what to merge first, we may need to update our PRs to use the same common methods.

vritant:
closed candlepin/candlepinproject.org#74 , replaced with candlepin/candlepinproject.org#77

server/src/test/java/org/candlepin/controller/PoolManagerFunctionalTest.java
```diff
@@ -207,7 +209,7 @@ public void testQuantityCheck() throws Exception {
 
         // The cert should specify 5 monitoring entitlements, taking a 6th should fail:
         assertEquals(null, poolManager.entitleByProducts(data));
-
+        monitoringPool = poolCurator.find(monitoringPool.getId());
```
nguyenfilip: Just curious, why you added this line?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1785,42 +1964,62 @@ void handleSelfCertificate(Consumer consumer, Pool pool,
...
-            return newEntitlement;
+        public Map<String, Entitlement> handleEntitlement(Consumer consumer,
+                Map<String, PoolQuantity> poolQuantities,
+                Map<String, Entitlement> entitlements) {
+            Map<String, Entitlement> result = new HashMap<String, Entitlement>();
+            for (Entry<String, PoolQuantity> entry : poolQuantities.entrySet()) {
+                Entitlement newEntitlement = new Entitlement(entry.getValue().getPool(), consumer, entry
+                        .getValue().getQuantity());
+                result.put(entry.getKey(), newEntitlement);
+                consumer.addEntitlement(newEntitlement);
```
nguyenfilip: This is the reason why revoke doesn't work in PoolManagerFunctionalTest.testRevocation() (the "TODO fix me" that you've left there) without EntityManager.clear().  The instance newEntitlement is being added to consumer.entitlements collection before being persisted. Because we implement equals/hashcode by @Id, this is fundamentally flawed (and actually disalowed even on Java level, see note in https://docs.oracle.com/javase/8/docs/api/java/util/Set.html). The revokeEntitlements in the PoolManagerFunctionalTest.testRevocation() method will try to remove the newEntitlement (revoked one) from the collection but will fail because the entitlement will have @Id already assigned and thus equals/hashCode value changes.

@mstead @vritant Guys I will send you shortly a longer explanation of this problem, because I feel its something that we do in our code base regulary (including me) and its adding so much pain to work with Hibernate

server/src/test/java/org/candlepin/controller/PoolManagerFunctionalTest.java
```diff
@@ -241,7 +251,7 @@ public void testRegenerateEntitlementCertificatesWithSingleEntitlement()
         throws Exception {
         AutobindData data = AutobindData.create(childVirtSystem).on(new Date())
                 .forProducts(new String [] {provisioning.getId()});
-        this.entitlementCurator.refresh(poolManager.entitleByProducts(data).get(0));
```
nguyenfilip: Why removing refresh on this line? 

The regenerateECAndAssertNotSameCertificates in this test uses the entiteledCertificate that you produced [1] (I saw the same id in the debugging session). By not refreshing, the method regenerateECAndAssertNotSameCertificates will see only the memory image of the created cert, not what has been actualy stored into database.

In case you think refresh really not needed here, then we should remove call to '.get()' as it is superficial without the refresh.

[1]

```
    private Set<EntitlementCertificate> collectEntitlementCertIds(
        Consumer consumer) {
        Set<EntitlementCertificate> ids = Util.newSet();
        for (Entitlement entitlement : consumer.getEntitlements()) {
            for (EntitlementCertificate ec : entitlement.getCertificates()) {
                ids.add(ec);
            }
        }
        return ids;
    }
```

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
vritant: :+1:

server/src/test/java/org/candlepin/controller/PoolManagerFunctionalTest.java
```diff
@@ -599,12 +609,132 @@ public void testDevPoolRemovalAtUnbind() throws EntitlementRefusedException {
...
+        Pool pool2 = createPool(owner, p, 10L, TestUtil.createDate(2000, 3, 2),
+                TestUtil.createDate(2050, 3, 2));
+        poolCurator.create(pool2);
+
+        Map<String, Integer> poolQuantities = new HashMap<String, Integer>();
+        poolQuantities.put(pool1.getId(), 1);
+        poolQuantities.put(pool2.getId(), 1);
+
+        List<Entitlement> results = poolManager.entitleByPools(devSystem, poolQuantities);
+        // TODO fix me.This should not be necessary
```
nguyenfilip: The same as above, hopefuly after the implementation is fixed, this will not be necessary

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1302,78 +1396,127 @@ public Entitlement adjustEntitlementQuantity(Consumer consumer,
...
+                            i18n.tr("Argument mismatch in entitlement update, number of pools: {}," +
+                            " number of entitlements: {}",
+                            entitlements.keySet().size(), poolQuantityMap.keySet().size()));
+        }
         else {
             handler = new UpdateHandler();
         }
 
-        // Grab an exclusive lock on the consumer to prevent deadlock.
-        consumer = consumerCurator.lockAndLoad(consumer);
```
nguyenfilip: Why is it no longer necessary to lock consumer?

vritant: @nguyenfilip yup, since that was the only update on the consumer :) much lower risk of lock waits now :) this was actually the incentive for computing compliance asynchronously.

nguyenfilip:
@vritant I reviewed the major parts including async compliance status and this PR looks good to me. The only thing I would consider is to fix the NewHandler and associated tests (TODOs in the code) as noted inline.

server/src/test/java/org/candlepin/controller/PoolManagerFunctionalTest.java
vritant: @nguyenfilip needed to add this when I was clearing the cache at some point of time. will be cleaned up in the merged PR.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
vritant: @nguyenfilip This worked brilliantly. thanks for the help! fixed in the merged PR.

server/src/test/java/org/candlepin/controller/PoolManagerFunctionalTest.java
vritant: @nguyenfilip  added all the refreshes back in.

server/src/test/java/org/candlepin/controller/PoolManagerFunctionalTest.java
vritant: @nguyenfilip removed it in the merged PR. This was yet another clean up to be removed after I was no longer clearing in the batch bind impl.

vritant:
replaced by #1123

PR #1075, opened 2016-01-27
1301812: Fixed an issue with event generation for stack derived pools (0.9.49)

Ceiu:
- Changes made to a stack derived pool will no longer cause null
  pointer exceptions during manifest import or pool refresh


Ceiu:
Note that the Jenkins spec tests will not complete without another patch to fix a variety of issues surrounding deploying within a Docker environment. These tests pass locally.

PR #1076, opened 2016-01-27
1301812: Fixed an issue with event generation for stack derived pools (0.9.51)

Ceiu:
- Changes made to a stack derived pool will no longer cause null
  pointer exceptions during manifest import or pool refresh


mstead:
0.9.51.x has been frozen for production - CDK support. Closing this PR since we won't likely be releasing from this branch again.

PR #1081, opened 2016-01-28
Fix server/bin scripts that have been broken since candlepin 2.0

vritant:
* fix scripts that were using older client versions
* bin scripts should now use HostedTest resources
* I have only fixed scripts that were using create_subscription. there may be other scripts still broken, no idea what is used and what is stale.
*  found that spec tests are still being written using create_subscription. Throw an error if that happens.
* move hosted test methods from candlepin scenarios to hostedtest_api so scripts can use them.

### DIFF
--- a/server/bin/bind-performance.rb
+++ b/server/bin/bind-performance.rb
@@ -8,9 +8,12 @@
 # Can be run and re-run against a typical dev deployment without any arguments.
 
 require  "../client/ruby/candlepin_api"
+require  "../client/ruby/hostedtest_api"
 require 'pp'
 
 require 'benchmark'
+
+include HostedTest
 
 ADMIN_USERNAME = "admin"
 ADMIN_PASSWORD = "admin"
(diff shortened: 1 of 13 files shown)


server/bin/bind-performance.rb
```diff
@@ -26,9 +29,9 @@ def time_rand from = 0.0, to = Time.now
   Time.at(from + rand * (to.to_f - from.to_f))
 end
 
-cp = Candlepin.new(ADMIN_USERNAME, ADMIN_PASSWORD, nil, nil, HOST, PORT)
+@cp = Candlepin.new(ADMIN_USERNAME, ADMIN_PASSWORD, nil, nil, HOST, PORT)
```
Ceiu: Why change this to an instance variable? Seems like it shouldn't work, and definitely doesn't make sense semantically.

vritant: @Ceiu the hosted test api needs @cp

vritant: @Ceiu also, I tested all the scripts, and this one worked. only one of them did not work I think, and that was broken for un-related reasons, I figured its stale code at this point not worth fixing.

PR #1083, opened 2016-01-29
1302801: Surface pools by subscription id

wottop:


server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -946,6 +946,31 @@ public Owner updateOwner(@PathParam("owner_key") @Verify(Owner.class) String key
...
+     * Gets pools for a Subscription
+     *
+     * @return a list of Pool objects
+     * @httpcode 400
+     * @httpcode 200
+     */
+    @GET
+    @Path("{owner_key}/subscriptions/{subscription_id}/pools")
+    @Produces(MediaType.APPLICATION_JSON)
+    public List<Pool> getPoolsForSubscription(@PathParam("owner_key") @Verify(Owner.class) String key,
```
vritant: Just curious if you considered using the existing API /owners/#{owner_key}/pools ?
we are going to introduce a similar API to fetch a list of pools of an owner given the parent pool Id, hence I ask. Not suggesting a change, just curious.

vritant:
@wottop every thing looks good, but please consider:
* moving the spec tests to owner_resource_spec.rb
* adding a new PoolCuratorTest unit test

PR #1086, opened 2016-02-02
1300705: Fix perf issue finding dev pools

mstead:
Auto-attach for a developer consumer was taking approx
2 minutes due to the query used to check if a development
pool existed for the consumer.

This patch improves the query time by adding an owner filter
to the Pools query.

**Testing Note**
This bug was only reproducible with a lot of pools already existing. I've tested this already against a large data set and since it takes a while to get set up, I'd suggest just doing a functionality check to verify this PR.


wottop:
ack

PR #1089, opened 2016-02-02
Add special case code to handle empty CRLs.

awood:
Empty CRLs are missing the revokedCertificates sequence which creates a
difficult situation for the streaming code since it has to consume part
of the stream in order to determine if the list is empty or not.  If the
list isn't empty, then we've consumed part of the ASN1 that we actually
needed.  This state is especially problematic for the
X509CRLStreamWriter because it relies on X509CRLEntryStream to leave the
stream in a known state.

### DIFF
--- a/server/src/main/java/org/candlepin/util/X509CRLStreamWriter.java
+++ b/server/src/main/java/org/candlepin/util/X509CRLStreamWriter.java
@@ -20,7 +20,9 @@ import static org.candlepin.util.DERUtil.*;
 import org.apache.commons.io.IOUtils;
 import org.bouncycastle.asn1.ASN1EncodableVector;
 import org.bouncycastle.asn1.ASN1InputStream;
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/candlepin/util/X509CRLStreamWriter.java
```diff
@@ -225,7 +240,7 @@ public synchronized X509CRLStreamWriter preScan(InputStream crlToChange, CRLEntr
             asn1In = new ASN1InputStream(crlToChange);
             while ((o = asn1In.readObject()) != null) {
                 if (o instanceof DERSequence) {
-                    // Now we are at the signatureAlgorithm
+                    // Now we are at the signatureAlgorith
```
wottop: Probably meant to leave the 'm' in

wottop:
looks good. ack.

PR #1090, opened 2016-02-03
BZ1304417 Mass Pool lock

nguyenfilip:
Defensive locking of pools during revocation of entitlements.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -1228,6 +1228,13 @@ public class CandlepinPoolManager implements PoolManager {
             return;
         }
 
+        List<Pool> poolsToLock = new ArrayList<Pool>();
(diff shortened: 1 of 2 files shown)


nguyenfilip:
@vritant Would you like to take a look at yet another version of locking? What do you think about this?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1238,6 +1245,8 @@ public void revokeEntitlements(List<Entitlement> entsToRevoke) {
             entsToRevoke.addAll(pool.getEntitlements());
         }
 
+        poolCurator.lock(poolsToLock);
```
vritant: Did you mean to do poolCurator.lock(poolsToDelete); ??

nguyenfilip: yes :+1:

PR #1091, opened 2016-02-03
Allow X509CRLStreamWriter to change AuthorityKeyIdentifier extension.

awood:
The AuthorityKeyIdentifier is an extension that identifies the public
key of the key pair used to sign the CRL.  If users change keys, in the
old implementation the AuthorityKeyIdentifier would not be updated to
reflect the new key pair.  This patch corrects that oversight.

### DIFF
--- a/server/src/main/java/org/candlepin/util/CrlFileUtil.java
+++ b/server/src/main/java/org/candlepin/util/CrlFileUtil.java
@@ -204,7 +204,7 @@ public class CrlFileUtil {
             // Note: This will break if we ever stop using RSA keys
             PrivateKey key = this.pkiReader.getCaKey();
             X509CRLStreamWriter writer = new X509CRLStreamWriter(
-                input, (RSAPrivateKey) key);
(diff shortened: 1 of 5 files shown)


nguyenfilip:
@awood @Ceiu Small question that is slightly related to this PR. Are we ok with the fact that source level is JDK 1.8? e.g. X509CRLStreamWriter uses intValueExact() (JDK 8 only). I believe the answer is yes because there are already commits from Oct 2015 that do use JDK8 only features.

Ceiu:
I believe that's an oversight and more a consequence of developing on F21+ than anything else. It's been a while since I checked, but I think we need to support back to 1.6.

nguyenfilip:
@awood other than the JDK note, this PR looks good to me. I tried [1] to check the modified certs and  they turn out to be correctly replacing the extension. I guess the JDK 8 source level issue will be fixed in a different PR anyway right?

[1] openssl crl -inform DER -in XXXX.crl -text -noout

PR #1093, opened 2016-02-05
Forward port of Batch Revoke Entitlements to Master

nguyenfilip:


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1547,71 +1551,70 @@ public void regenerateCertificatesOf(Owner owner, String productId, boolean lazy
...
-        // entitlement and clean them up as well
+        List<Pool> poolsToDelete = poolCurator.listBySourceEntitlements(entsToRevoke);
+        
+        List<Pool> poolsToLock = new ArrayList<Pool>();
+        for (Entitlement ent: entsToRevoke) {
+            poolsToLock.add(ent.getPool());
+            
+            // If we are deleting a developer entitlement, be sure to delete the
+            // associated pool as well.
+            if (ent.getPool() != null && ent.getPool().isDevelopmentPool())
```
nguyenfilip: @mstead I have been forward porting revoke entitlements and bumped revocation of development pool entitlement. Could you just quickly look if this handling of revocation of development pools is correct? 

Btw why is it necessary to delete the pool during the revocation?

mstead: > @mstead I have been forward porting revoke entitlements and bumped revocation of development pool entitlement. Could you just quickly look if this handling of revocation of development pools is correct?
> 
> Btw why is it necessary to delete the pool during the revocation?

There should only ever be one dev pool/entitlement combination per consumer. The life of the pool == the life of the entitlement. If the ent is deleted, the pool must go with it.

A big part of this is that we don't want the pool hanging around if it will never be used again -- especially since this is a pool per consumer who uses this functionality.

vritant:
Please merge #1102 for tests to pass

vritant:
#1102 is now merged in to master, please rebase master.

vritant:
@wottop assigned this to myself cause there was no asignee, but I did hear you talking about this, so please let me know if you are already reviewing, in which case I will go to something else.

wottop:
@vritant I looked at one of the other ports of the same code. Probably better to have someone different look at this one. Go ahead.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1547,71 +1551,71 @@ public void regenerateCertificatesOf(Owner owner, String productId, boolean lazy
...
+
+
+        for (Pool pool : poolsToDelete) {
+            entsToRevoke.addAll(pool.getEntitlements());
         }
 
-        // The quantity is calculated at fetch time. We update it here
-        // To reflect what we just removed from the db.
-        pool.setConsumed(pool.getConsumed() - entitlement.getQuantity());
+        poolCurator.lock(poolsToDelete);
```
vritant: Do you think its better to make a single lock call instead of:
poolCurator.lock(poolsToLock); and poolCurator.lock(poolsToDelete); or am I missing something?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1547,71 +1551,71 @@ public void regenerateCertificatesOf(Owner owner, String productId, boolean lazy
...
+        log.info("All deletes flushed successfully");
+
+        List<Entitlement> stackingEntitlements = filterStackingEntitlements(entsToRevoke);
 
-            String stackId = pool.getProduct().getAttributeValue("stacking_id");
+        for (Entitlement ent : stackingEntitlements) {
+            Pool pool = ent.getPool();
+            Consumer consumer = ent.getConsumer();
+            String stackId = pool.getProductAttributeValue("stacking_id");
             Pool stackedSubPool = poolCurator.getSubPoolForStackId(consumer, stackId);
```
vritant: consider using the [batch impl](https://github.com/candlepin/candlepin/pull/1073/files#diff-0cac811ddb60a663206629eab27f0c12R641) of this from #1073

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1631,63 +1636,93 @@ void removeEntitlement(Entitlement entitlement, boolean regenModified) {
...
-        if (regenModified) {
-            // Find all of the entitlements that modified the original entitlement,
-            // and regenerate those to remove the content sets.
-            // Lazy regeneration is ok here.
-            this.regenerateCertificatesOf(entitlementCurator.listModifying(entitlement), true);
+            // We work in batches of maximum size 1000.
+            if (i % 1000 == 0) {
+                Set<Entitlement> modifiedEnts = entitlementCurator.batchListModifying(batch);
+                this.regenerateCertificatesOf(modifiedEnts, true);
+                batch.clear();
```
vritant: Am curious why we are doing this in a batch since all we are doing is marking the entitlement as dirty here ( cause lazy is true), is the reason we are doing this in a batch is that some databases have a limitation of in clause to 1000 ? if yes I need to put similar checks in #1073 . please let me know!

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1631,63 +1636,93 @@ void removeEntitlement(Entitlement entitlement, boolean regenModified) {
...
         }
+        consumerCurator.flush();
 
-        sink.queueEvent(event);
+        log.info("All statuses recomputed.");
+
+        // for each deleted entitlement, create an event
+        for (Entitlement entitlement : entsToRevoke) {
+            Consumer consumer = entitlement.getConsumer();
+            Event event = eventFactory.entitlementDeleted(entitlement);
```
vritant: the event object may not be sent out if the expired event is sent out. please consider creating this in the else clause?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1631,63 +1636,93 @@ void removeEntitlement(Entitlement entitlement, boolean regenModified) {
...
-        for (Entitlement e : entitlementCurator.listByConsumer(consumer)) {
-            removeEntitlement(e, false);
-            count++;
-        }
-        // Rerun compliance after removing all entitlements
-        complianceRules.getStatus(consumer);
-        return count;
+        List<Entitlement> entsToDelete = entitlementCurator.listByConsumer(consumer);
+        revokeEntitlements(entsToDelete);
+        return entsToDelete.size();
```
vritant: Why is it not needed to not compute consumer compliance here any more?
Unless there is a valid reason, perhaps we could kick off an asynchronous consumer compliance job from #1073  ?

Looks like there were two APIs to remove all entitlements of a consumer, one with compliance check and other without, so this was an explicit decision at some point, hence the question

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -274,17 +298,67 @@ private Criteria createModifiesDateFilteringCriteria(Consumer consumer, Date sta
...
 
+    /**
+     * A version of list Modifying that finds Entitlements that modify
+     * input entitlements.
+     * When dealing with large amount of entitlements for which it is necessary
+     * to determine their modifier products.
+     * @param entitlement
+     * @return Entitlements that are being modified by the input entitlements
+     */
+    public Set<Entitlement> batchListModifying(List<Entitlement> entitlements) {
```
vritant: Looks like I implemented [something very similar](https://github.com/candlepin/candlepin/pull/1073/files#diff-8864d89395315f6a5f8456d0e504fd91R365) for batch attach in #1073, with two differences:
1. I had the liberty that all the entitlements being passed to my method were of the same consumer, which is not the case here.
2. Second difference I think is an issue here. Both Master branch and batch attach explicitly exclude the entitlements being passed as input to this method from the result, which I think this implementation does not.

That being said, based on which branch is merged first, I think there there are two sets of methods in our implementation and we should remove that from one of the branches.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -337,6 +411,48 @@ private void addToMap(Map<String, Set<Entitlement>> map, Entitlement e) {
...
+                consumers, earliestStartDate, latestEndDate)
+                .list();
+
+        return new ProductEntitlements(overlapEnts);
+    }
+
+    private Date findLatestEndDate(List<Entitlement> entitlements) {
+        Date max = null;
+        for (Entitlement e : entitlements) {
+            if (max == null || max.before(e.getEndDate())) {
```
vritant: This Question is to both you and me ( cause I implemented the same thing in #1073 ) So I think we should ask @mstead .
@mstead by any chance do entitlements have null end dates that imply never ending entitlements? I ask cause I know our upstream source of information did do that in a lot of places.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -776,4 +838,17 @@ public Pool findDevPool(Consumer consumer) {
         criteria.setMaxResults(1).uniqueResult();
         return (Pool) criteria.uniqueResult();
     }
+
+    public void lock(List<Pool> poolsToLock) {
+        if (poolsToLock.isEmpty()) {
```
vritant: I have a [similar method](https://github.com/candlepin/candlepin/pull/1073/files#diff-0cac811ddb60a663206629eab27f0c12R511) in #1073 , but there I need to return the pools as well. perhaps we could simply use that method instead of this?

server/src/main/java/org/candlepin/model/ProductEntitlements.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * This data class is a container for Entitlements mapped by their provided
+ * Products. The reason to have this class is to be able to easily answer query:
+ * Given a product, which entitlements provide it?
+ * @author fnguyen
+ *
+ */
+public class ProductEntitlements {
+    private Map<String, Set<Entitlement>> entsByProductIds = new HashMap<String, Set<Entitlement>>();
```
vritant: All this logic is there in #1073 too, but you have encapsulated it well. At this point I feel we should have collaborated in our PRs from the start :(

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: You are right! The history is that I moved method listBySourceEntitlements to a different line and now the collection poolsToLock doesn't bring any benefit.

nguyenfilip: Yes I think single call will be better! I was moving listBySourceEntitlements and now the calls to 'lock' are somehow randomly placed in the beginning of the method.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: Answer to "why it is not needed any more"

The current code actually computes compliance every time at the end of revokeEntitlements method.

Answer to: "Looks like there were two APIs to remove all entitlements of a consumer, one with compliance check and other without"

I think the one without check was there again for performance reasons. It was there so that its possible to recompute compliance after all entitlements are revoked. This is however exactly what the current batch revoke does.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -366,7 +366,7 @@ private void cleanupAndDelete(Owner owner, boolean revokeCerts) {
             }
             else {
                 // otherwise just remove them without touching the CRL
-                poolManager.removeAllEntitlements(c);
+                poolManager.revokeAllEntitlements(c);
```
nguyenfilip: As @vritant correctly pointed out (@mstead this is also in the email I sent yesterday), this may not be very good decision. By doing this I effectively created code that ignores `revokeCerts` flag in method 
```java
 public void deleteOwner(@PathParam("owner_key") String ownerKey,
        @QueryParam("revoke") @DefaultValue("true") boolean revoke) 
```

After my change the method looks like this:
```java
     for (Consumer c : consumers) {
            log.info("Removing all entitlements for consumer: " + c);

            if (revokeCerts) {
                poolManager.revokeAllEntitlements(c);
            }
            else {
                // otherwise just remove them without touching the CRL
                poolManager.removeAllEntitlements(c);
            }
        }
```

So even in case when a REST client sets this flag to false, then code in this PR will regenerate certificate regardless. 

I should either
- clean up the resource method and remove the flag alltogether (which is API change)
- introduce the flag into my code, which is very easy thing to do (@vritant I think this is the best route to take right now, what do you think?)

vritant: @nguyenfilip I'm in favor of the second approach, simply cause if there are some use cases where we do not need to compute compliance ( or delay it ), then there might be some advantage in skipping it.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: No the reason is not _in clause_ limitation. The reason is to optimize underlying batchListModifying method. The method batchListModifying starts by searching overlapping product ids:
```java
ProductEntitlements pidEnts = getOverlappingForModifying(entitlements);
```
The getOverlappingForModifying might return high number of entitlements. More then actually necessary. The rest of the batchListModifying method filters pidEnts. So this batch size (1000) actually gives us a way to configure the following trade off: 

If the _batch size_ is too big, the batchListModifying starts by finding too many entitlements and then consumes a lot of memory. On the other hand if _batch size_ is too small, we end up with too many calls to batchListModifying which is better on memory but worse on time performance.

However, thinking about this a bit more, I agree the batch size of 1000 is kinda arbitrary. It worked well on the problematic Import dataset. But again I agree its really arbitrary constant :-P

nguyenfilip:
retest this please

nguyenfilip:
retest this please

server/src/main/java/org/candlepin/resource/OwnerResource.java
nguyenfilip: I added the flag back

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: The original code behaved the same way. 

```java
  Event event = eventFactory.entitlementDeleted(entitlement);
        if (!entitlement.isValid() &&
                entitlement.getPool().isUnmappedGuestPool() &&
                consumerCurator.getHost(consumer.getFact("virt.uuid"), consumer.getOwner()) == null) {
            event = eventFactory.entitlementExpired(entitlement);
            event.setMessageText(event.getMessageText() + ": " +
                i18n.tr("Unmapped guest entitlement expired without establishing a host/guest mapping."));
        }
```
Do you think it is a bug even in master?

server/src/main/java/org/candlepin/model/EntitlementCurator.java
mstead: > This Question is to both you and me ( cause I implemented the same thing in #1073 ) So I think we should ask @mstead .
> @mstead by any chance do entitlements have null end dates that imply never ending entitlements? I ask cause I know our upstream source of information did do that in a lot of places.

I believe that entitlements should always have a start/end date. Start/End dates of entitlements actually come from the Pool they were created from and those dates have a NotNull constraint.

vritant:
replaced by #1123

PR #1094, opened 2016-02-05
1302051: subscription type field for list --consumed

wottop:
Adds the compliance type to outgoing pool json. It is displayed on
Subscription Manager CLI and GUI


nguyenfilip:
ack.

In the future I think we should rename/adjust addCalculatedAttributes method and specifically CalculatedAttributesUtil class. It contains some parameters that are unused. I will put that on backlog.

PR #1096, opened 2016-02-10
Candlepin Data Generator

Ceiu:
First version of the data generator, which generates owners, users, roles, content, products and pools. 

There may need to be additional pool generators added to cover more of the uses cases we have for Candlepin, but for now, many of the basic cases are handled.

The biggest current flaw with the data generator is that it does not validate anything with the server. It is assumed it be running against an empty database and that the POST/PUT requests succeed. This may need to change in the future.


wottop:
ack

PR #1098, opened 2016-02-11
Updated the auto configuration to respect a user-specified database name

Ceiu:
- Using the auto configuration with a custom database name will now
  properly generate configuration with the specified DB name

### DIFF
--- a/server/bin/deploy
+++ b/server/bin/deploy
@@ -108,7 +108,7 @@ create_var_cache_candlepin() {
 }
 
 autoconf() {
-  buildr -s -e $1 erb $2 $3
+  buildr -s -e $1 app_db_name=$APP_DB_NAME erb $2 $3
 
   if [ "$?" -ne "0" ]; then
     err_msg "ERROR: candlepin.conf generation failed!"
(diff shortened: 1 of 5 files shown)


awood:
Buildbot, retest this please

awood:
Looks like this needs to get rebased to master to pick up some spec test fixes

Ceiu:
Ahh, yeah. Rebased, hopefully that takes care of it.

server/bin/deploy
```diff
@@ -108,7 +108,7 @@ create_var_cache_candlepin() {
 }
 
 autoconf() {
-  buildr -s -e $1 erb $2 $3
+  buildr -s -e $1 app_db_name=$APP_DB_NAME erb $2 $3
```
awood: Where is `APP_DB_NAME` set to a default?

Ceiu: At the top (line 3). Same applies to the Gutterball script.

awood: > At the top (line 3). Same applies to the Gutterball script.

Ah!  Didn't realize it was a pre-existing variable.

PR #1099, opened 2016-02-12
Fixed incorrect pathing with the teepipe

Ceiu:
- teepipe now lives in /tmp and is referenced as such by all commands,
  fixing issues with running unit and spec tests


cnsnyder:
Looks good to me. I am presently building and running the containers, will merge if those run as expected.

cnsnyder:
Works for me. Merging. ACK.

PR #1100, opened 2016-02-12
Add missing RPM requires.

awood:
The spec file was missing a requires on the JDBC driver which `cpdb` needs.


awood:
buildbot, retest this please

wottop:
ack

PR #1103, opened 2016-02-15
Specify 'utf-8' as txt encoding for source files.

alikins:
'builder gettext:extract" uses xgettext to extract string
catalogs, and sans '--from-code=utf-8', it assumes source
files are encoded as plain ASCII. If a file includes (valid)
utf-8 that is not also plain ASCII, xgettext fails with an
error like:

xgettext: Non-ASCII comment at or before
gutterball/src/main/java/org/candlepin/gutterball/util/cron/CronSchedule.java:39.
          Please specify the source encoding through --from-code.

### DIFF
--- a/tasks/gettext.rake
+++ b/tasks/gettext.rake
@@ -72,7 +72,7 @@ module Gettext
             java_files << Dir[File.join(src, '**', '*.java')]
           end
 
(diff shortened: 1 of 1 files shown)


vritant:
please merge #1102 for tests to pass

vritant:
#1102 is now merged in to master, please rebase master.

vritant:

PR #1106, opened 2016-02-17
Add Vagrant support for dev env with postgres

barnabycourt:


dgoodwin:

awood:
@barnabycourt Cool, but would you mind dropping a brief readme in the vagrant directory describing how to use all this?  Just a paragraph or two is fine.

vagrant/roles/candlepin-root/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+    - postgresql-jdbc
+    - postgresql-server
+    - postgresql
+    - zlib
+    - zlib-devel
+    - liquibase
+    - python-psycopg2
+    - libxml2-python
+    - wget
+- name: Add remote debugging config to comcat
```
cnsnyder: s/comcat/tomcat/g ?

cnsnyder:
Looks good, and works well for me.

@barnabycourt One question/thought for future improvement, Should we include a script of some sort that will install dependencies (such as Ansible) and do other configuration necessary (such as making the directories to your source checkout +x) so that one could clone this repo and in one command get a running deployment vm?

The above question is, to me, not a blocker to the merge of this improvement. I will however wait for your response before merging.

barnabycourt:
I don't know of a good way to check for the NFS configuration that would not be too invasive. Perhaps a simple script could be written to check the installation of ansible, directory permissions, and vagrant version but those requirements are currently spelled out in the README. 

Some of the complexities such as script would have to solve would be determining the NFS server user so that I think that script would be a good addon later  but would prefer not to do it now.

cnsnyder:
ACK, merging.

PR #1107, opened 2016-02-18
Forward port of Batch Revoke to 0.9.54

nguyenfilip:
@barnabycourt  @awood This is the PR for forward port of batch revoke to 0.9.54

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -1245,6 +1245,13 @@ public class CandlepinPoolManager implements PoolManager {
             return;
         }
 
+        List<Pool> poolsToLock = new ArrayList<Pool>();
+        for (Entitlement ent: entsToRevoke) {
+            poolsToLock.add(ent.getPool());
+        }
+
+        poolCurator.lock(poolsToLock);
+
         log.info("Batch revoking entitlements: " + entsToRevoke.size());
 
         entsToRevoke =  new ArrayList<Entitlement>(entsToRevoke);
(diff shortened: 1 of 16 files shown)


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1227,36 +1232,163 @@ void removeEntitlement(Entitlement entitlement,
...
+
+        poolCurator.lock(poolsToDelete);
+
+        for (Entitlement ent : entsToRevoke) {
+            //We need to trigger lazy load of provided products
+            //to have access to those products later in this method.
+            ent.getPool().getProvidedProducts().size();
+            Pool pool = ent.getPool();
+            pool.setConsumed(pool.getConsumed() - ent.getQuantity());
+
```
vritant: @nguyenfilip reminder, need to talk to you about missing
pool.getEntitlements().remove(ent);

nguyenfilip: This is a great thing for quite a long discussion. Right now I don't think there is a problem with missing entitlements.remove.

However...

In this branch, the code that you mention is in EntitlementCurator.batchDelete. I don't think it should be there (it causes performance issues when Pool has a lot of entitlements).

If we never initialize pool.entitlements during revocation, then we won't run into any issues. If we do initialize pool.entitlements at some point and then do not call pool.getEntitlements().remove(ent); then the ent will stay in the database. 

I dont think the old code to which you refer to ever worked, because entitlements are EXTRA lazy, the actual remove did not initialize the collection and didn't remove anything from database. Later if you initialized the collection, the ent would stay in the database regardless.

PR #1109, opened 2016-02-18
Fix apicrawler

awood:
The API crawler had several problems: resource classes not listed in
RootResource were not getting scanned and overloaded methods were not
being documented correctly.

This patch uses Scannotation to search for classes providing a REST
resource instead of relying on listing in the RootResource.

This patch should produce a "candlepin_methods.json" that no longer causes the Jekyll builder to throw warnings.


wottop:
ack

PR #1113, opened 2016-02-23
Fix ImporterTest, clean up test files

vritant:


awood:
buildbot, ok to test

server/src/test/java/org/candlepin/sync/ImporterTest.java
```diff
@@ -592,7 +592,7 @@ public void testReturnsSubscriptionsFromManifest() throws IOException, ImporterE
 
         Product prod = new Product("prodId", "prodTest", null);
         prod.setDependentProductIds(null);
-        File prodFile = new File("product.json");
+        File prodFile = new File("target/test/resources/upstream/product.json");
```
awood: Instead of loading these directly from the file path, I think the preferred way to do it is via the classpath.  E.g.

```java
ImporterTest.class.getClassLoader().getResourceAsStream("upstream/product.json");
```

vritant:
@awood done

PR #1115, opened 2016-02-24
Remove statistics from candlepin

vritant:


awood:
build bot, add to whitelist

awood:
build bot, ok to test

server/src/main/java/org/candlepin/resource/RootResource.java
```diff
@@ -71,7 +71,6 @@
         addResource(PoolResource.class);
         addResource(ProductResource.class);
         addResource(RulesResource.class);
-        addResource(StatisticResource.class, "statistics/generate");
```
awood: You can delete the `addResource` method that takes an object and string as well, since it looks like this resource was the only one using it.

awood:
Aside from the above comment, looks fantastic.  Happy to see these go.

vritant:
@awood done

PR #1117, opened 2016-02-25
Warn developers when they deploy hosted mode to not spec test in para…

vritant:
…llel mode


wottop:
It's a start. It would be nice to scuttle the test when it's run, but this will do for now.

PR #1121, opened 2016-03-01
Correct error in SKU search SQL seen in MySQL.

awood:
The '1=1' stipulation is to reject searches performed on non-marketing
product SKUs.  Originally the stipulation was "true='true'" which worked
on PostgreSQL but not on MySQL.


wottop:
ack

PR #1124, opened 2016-03-02
1311664: Ignore empty guest IDs and hypervisor IDs

cnsnyder:


server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -108,7 +110,7 @@ public HypervisorResource(ConsumerResource consumerResource,
     @Deprecated
     @Transactional
     public HypervisorCheckInResult hypervisorUpdate(
-        Map<String, List<GuestId>> hostGuestMap, @Context Principal principal,
+        Map<String, ArrayList<GuestId>> hostGuestMap, @Context Principal principal,
```
wottop: Why the change to ArrayList? List still works.

cnsnyder: I've changed to ArrayList as the List class does not allow removal of items in the list using an iterator (as done below).

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -123,22 +125,43 @@ public HypervisorCheckInResult hypervisorUpdate(
...
         VirtConsumerMap hypervisorConsumersMap =
                 consumerCurator.getHostConsumersMap(owner, hostGuestMap.keySet());
 
+        int emptyGuestIdCount = 0;
+        int emptyHypervisorIdCount = 0;
         Set<String> allGuestIds = new HashSet<String>();
-        for (Entry<String, List<GuestId>> hostEntry : hostGuestMap.entrySet()) {
-            for (GuestId gid : hostEntry.getValue()) {
-                allGuestIds.add(gid.getGuestId());
+        Set<GuestId> toRemove = new HashSet<GuestId>();
```
wottop: Doesn't get used anywhere in the diff. Still needed?

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -123,22 +125,43 @@ public HypervisorCheckInResult hypervisorUpdate(
...
+        if (emptyGuestIdCount > 0) {
+            log.warn("Found " + emptyGuestIdCount + " empty guest id(s). Ignoring.");
+        }
         // Maps virt guest ID to registered consumer for guest, if one exists:
         VirtConsumerMap guestConsumersMap = consumerCurator.getGuestConsumersMap(
                 owner, allGuestIds);
 
         HypervisorCheckInResult result = new HypervisorCheckInResult();
-        for (Entry<String, List<GuestId>> hostEntry : hostGuestMap.entrySet()) {
+        for (Entry<String, ArrayList<GuestId>> hostEntry : hostGuestMap.entrySet()) {
```
wottop: Same LIst/ArrayList comment as before

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -123,22 +125,42 @@ public HypervisorCheckInResult hypervisorUpdate(
...
                 consumerCurator.getHostConsumersMap(owner, hostGuestMap.keySet());
 
+        int emptyGuestIdCount = 0;
+        int emptyHypervisorIdCount = 0;
         Set<String> allGuestIds = new HashSet<String>();
-        for (Entry<String, List<GuestId>> hostEntry : hostGuestMap.entrySet()) {
-            for (GuestId gid : hostEntry.getValue()) {
-                allGuestIds.add(gid.getGuestId());
+        Iterator<Entry<String, ArrayList<GuestId>>> hostEntries = hostGuestMap.entrySet().iterator();
+        while (hostEntries.hasNext()) {
```
awood: I would prefer to see these as for loops.  I find that to be more idiomatic to Java.

```java
for (Iterator<Entry<String, ArrayList<GuestId>>> it = hostGuestMap.entrySet().iterator(); it.hasNext(); ) {

}
```

Also I didn't notice anywhere that you were actually using the Entry object's name so why not just do `hostGuestMap.values()`?

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -123,22 +125,42 @@ public HypervisorCheckInResult hypervisorUpdate(
...
+                    allGuestIds.add(guestId);
+                }
+                else {
+                    emptyGuestIdCount++;
+                    guestIds.remove();
+                }
             }
         }
+        if (emptyGuestIdCount > 0) {
+            log.warn("Found " + emptyGuestIdCount + " empty guest id(s). Ignoring.");
```
awood: Avoid string concats in log statements.

Use

```java
log.warn("Found {} empty guest id(s). Ignoring.", emptyGuestIdCount);
```

See [here](http://www.slf4j.org/manual.html)

server/src/main/java/org/candlepin/resource/HypervisorResource.java
awood: List worked for me in Java 1.8.  Were there compatibility issues in 1.6?

```java

@Test
public void testRemove() {
    List<String> l = new ArrayList<String>();
    l.add("Hello");
    l.add("World");
    l.add("Foo");
    assertEquals(3, l.size());

    for (Iterator<String> x = l.iterator(); x.hasNext();) {
        String s = x.next();
        if ("Foo".equals(s)) {
            x.remove();
        }
    }

    assertEquals(2, l.size());
    assertFalse(l.contains("Foo"));
}
```

server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
```diff
@@ -310,4 +312,82 @@ public void checkInCreatesNoNewConsumerWhenCreateIsFalse() throws Exception {
...
+             any(Set.class))).
+             thenReturn(new VirtConsumerMap());
+        when(consumerCurator.getGuestConsumersMap(any(Owner.class), any(Set.class))).
+            thenReturn(new VirtConsumerMap());
+
+        when(ownerCurator.lookupByKey(eq(owner.getKey()))).thenReturn(owner);
+        when(principal.canAccess(eq(owner), eq(SubResource.CONSUMERS), eq(Access.CREATE))).
+            thenReturn(true);
+        when(consumerTypeCurator.lookupByLabel(
+         eq(ConsumerTypeEnum.HYPERVISOR.getLabel()))).thenReturn(hypervisorType);
```
awood: Indentation is off here.

server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
```diff
@@ -310,4 +312,82 @@ public void checkInCreatesNoNewConsumerWhenCreateIsFalse() throws Exception {
...
+        when(idCertService.generateIdentityCert(any(Consumer.class)))
+            .thenReturn(new IdentityCertificate());
+
+        HypervisorCheckInResult result = null;
+        try {
+            result = hypervisorResource.hypervisorUpdate(hostGuestMap,
+            principal, owner.getKey(), true);
+        }
+        catch (PersistenceException e) {
+            fail("Expected no exception");
```
awood: This isn't necessary.  The test method already throws and exception, and if an exception is thrown, JUnit marks the test as an "error" (which is actually what the result [should be](http://stackoverflow.com/questions/3425995/whats-the-difference-between-failure-and-error-in-junit) marked as)

server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
```diff
@@ -310,4 +312,82 @@ public void checkInCreatesNoNewConsumerWhenCreateIsFalse() throws Exception {
...
+        }
+        assertTrue(result != null);
```
awood: Use assertNotNull()

server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
```diff
@@ -310,4 +312,82 @@ public void checkInCreatesNoNewConsumerWhenCreateIsFalse() throws Exception {
...
+        assertTrue(result.getCreated().size() == 1);
```
awood: `assertEquals(1, result.getCreated().size())`

Using the appropriate assert yields better messaging from JUnit.

server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
```diff
@@ -310,4 +312,82 @@ public void checkInCreatesNoNewConsumerWhenCreateIsFalse() throws Exception {
...
+
+    @Test
+    public void ensureEmptyHypervisorIdsAreIgnored() throws Exception {
+        Owner owner = new Owner("admin");
+
+        Map<String, ArrayList<GuestId>> hostGuestMap = new HashMap<String, ArrayList<GuestId>>();
+        hostGuestMap.put("", new ArrayList(Arrays.asList(new GuestId("GUEST_A"),
+                                         new GuestId("GUEST_B"))));
+        hostGuestMap.put("HYPERVISOR_A", new ArrayList(Arrays.asList(new GuestId("GUEST_C"),
+                                                     new GuestId("GUEST_D"))));
```
awood: The indentation is inconsistent here.  Just add one additional indention on line continuations instead of spacing all the way over past the containing parenthesis.

server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
```diff
@@ -310,4 +312,82 @@ public void checkInCreatesNoNewConsumerWhenCreateIsFalse() throws Exception {
...
+        try {
+            result = hypervisorResource.hypervisorUpdate(hostGuestMap,
+            principal, owner.getKey(), true);
+        }
+        catch (PersistenceException e) {
+            fail("Expected no exception");
+        }
+        assertTrue(result != null);
+        assertTrue(result.getCreated().size() == 1);
+        assertTrue(!result.getCreated().toArray(new Consumer[1])[0].getHypervisorId().equals(""));
```
awood: Why convert to an array?  Does `result.getCreated().get(0)` not work?

And here again, use `assertEquals("", ...)` so you get a message like `Expected "" but was "Hello world"` instead of JUnit just throwing an `AssertionError` with zero context.

server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
```diff
@@ -310,4 +312,82 @@ public void checkInCreatesNoNewConsumerWhenCreateIsFalse() throws Exception {
...
+            thenReturn(true);
+        when(consumerTypeCurator.lookupByLabel(
+            eq(ConsumerTypeEnum.HYPERVISOR.getLabel()))).thenReturn(hypervisorType);
+        when(idCertService.generateIdentityCert(any(Consumer.class)))
+            .thenReturn(new IdentityCertificate());
+
+        HypervisorCheckInResult result = null;
+        try {
+            result = hypervisorResource.hypervisorUpdate(hostGuestMap,
+           principal, owner.getKey(), true);
```
awood: Indentation out of whack here too

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -106,6 +107,49 @@ public static boolean isSchedulable(JobCurator jobCurator, JobStatus status) {
...
+                hypervisor.getHypervisorId().getHypervisorId() != null) {
+                if (hypervisor.getHypervisorId().getHypervisorId().equals("")) {
+                    // Ignore hypervisors that have a hypervisor ID of ""
+                    hypervisorsIterator.remove();
+                    emptyHypervisorIdCount++;
+                    continue;
+                }
+                incomingHosts.put(hypervisor.getHypervisorId().getHypervisorId(), hypervisor);
+                hosts.add(hypervisor.getHypervisorId().getHypervisorId());
+                if (hypervisor.getGuestIds() != null && !hypervisor.getGuestIds().isEmpty()) {
```
awood: Do we really need to check if it's empty?  Won't an empty list just bomb out of the loop immediately?

awood:
Once this is approved, if the reviewer lets me know, I will squash these commits.

nguyenfilip:
@awood Looks good to me. Can you squash those commits?

PR #1126, opened 2016-03-02
1286697: Fix issue with cascaded deletes on pools.

awood:


awood:
@nguyenfilip Do we need to remove the `LazyCollection(LazyCollectionOption.EXTRA)` annotation that you mentioned in the bug?  Or is the annotation fine but the bug only manifests when it is removed?

vritant:
Note: removeEntitlement method is being replaced in the batch revoke PR.

nguyenfilip:
@awood It only manifests if it is removed. I don't think we should remove that annotation right now.

Ceiu:

awood:
@nguyenfilip Does this issue still need addressing with the batch revoke stuff merged in now?  I added `orphanRemoval` but `ConsumerResourceVirtEntitlementTest` fails now because presumably I'm not doing a flush somewhere.

PR #1130, opened 2016-03-03
1310173: Backport logrotate config entry

wottop:


awood:
Please change the commit message here since this particular PR is not a backport.

PR #1134, opened 2016-03-04
Added Versioning to Products and Content

Ceiu:
Added entity versioning to product and content, allowing the curators to converge identical entities used by multiple orgs and diverge when orgs change their version.

- Changed all new tables from cpo_* to cp2_*
- Removed the owner column from cp2_products and cp2_content
- Added the new join tables "cp2_owner_products" and "cp2_owner_content"
- Added collection management utility methods to both Product and Content
  for the new join table
- OwnerProductCurator now reuses owner lookup results when fetching
  products and content
- Content now contains a mapping to its various owners
- Updated lookup queries in the product and content curators
- Removed old migration task files
- Updated Product.equals and Product.hashCode
- Several classes updated to work with the changes to the product curator
  and product
- Removed duplicated code from the product curator and candlepin pool
  manager
- smoke-test.sh renamed to smoke-test.py
- Updated Product.equals to no longer use .equals on any collections
  managed by Hibernate, due to an issue with Hibernate's collections
  not respecting the contract on .equals
- Added four new spec tests for testing product versioning
- ProductCurator no longer generates a UUID for copies of products
  when updating from a shared instance
- product_versioning_spec now contains a test for each version-related
  update and creation case
- Updated content addition and removal to properly update the product
  as it changes
- Deleting a product from one org no longer deletes it from all orgs
- The "created" time for a Hibernate entity is no longer updated
  automatically if it has already been set when the object is persisted
- Product collections are no longer copied and duplicated if they are
  already identical
- Added a new utility method, Util.collectionsAreEqual for comparing
  two collections without using their equals method
- Made a few corrections to the hash code and equals calculations
  in Product and Content
- Removed a bunch of dead code in Content, ContentCurator and
  OwnerContentResource
- Content no longer converts null values to empty strings for
  values which are nullable
- ContentCurator and ProductCurator now have a removeContent/Product
  method which wraps all the book keeping associated with versioning
- Content now has a "locked" flag, which can be used to prevent API-level
  changes to the content
- Products that are locked no longer allow attaching or removing content
  from them
- Any product or content received from an upstream source -- either
  a manifest or adapter -- will now be locked; preventing it from being
  modified through the API
- Added two new tests to product versioning spec tests and one to
  the content versioning spec tests
- Modified how the VerifyAuthorizationFilter determines the owner
  from the query/arguments
- Product.getHref now attempts to use the Owner from the Resteasy
  context for building the URL. If the context does not have an
  owner, no URL is provided
- Product and Content no longer have an updated_upstream field, instead
  using an entity_version field to store the last known hashcode for a
  given entity. This field is automatically updated whenever either entity
  is persisted or updated.
- ProductCurator and ContentCurator now use the entity version field
  when finding existing versions of a given entity
- CandlepinPoolManager now caches the output of creating or updating
  entities and skips additional lookups when resolving references
- Added an index on the entity_version field and disallowed null values
- The clone and merge operations for Product and Content will now
  create copies of mutable values (such as Date instances) instead
  of just using the shared reference directly
- Removed the final modifier from some methods in the AbstractHibernateCurator
- Improved duplicate entity warnings during refresh and import
- Product and content creation are now inserted in bulk rather than
  individually during the migration task
- Insert count mismatches is now a migration failure instead of just
  a warning during migration
- Added comments on the various clone methods and copy constructors
  indicating where some of the extra copy work is being performed


server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -519,9 +776,16 @@ public void setDependentProductIds(Set<String> dependentProductIds) {
...
     public String getHref() {
         // If we don't have an owner here, we're in a bit of trouble.
-        return (this.getOwner() != null && this.getOwner().getKey() != null && this.getId() != null) ?
-            "/owners/" + this.getOwner().getKey() + "/products/" + this.getId() :
-            "";
+        // PER-ORG PRODUCT VERSIONING TODO:
+        // Do we have a way to determine the current owner from context? We need a way to reference
+        // a specific product (uuid) but with owner permissions...? Should this just change to UUIDs
+        // everywhere? This is messy.
+
```
mstead: IMO getHref() should point to this exact product -- using the uuid as you do below -- and should not be based on the context of the current Owner. Permissions on GET /product/:uuid  would then have to be modified to check the Principal's Owner in order to view it via the ProductResource.

mstead: If this is a 'global' product, I guess you behave as you do today?

Ceiu: The product and content UUIDs are to be completely internalized, as with the converging and diverging products, it's entirely possible that a RHID for a given org could point to one UUID one day, and another the next.

server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationTask.java
```diff
@@ -0,0 +1,455 @@
...
+     *  a PreparedStatement instance representing the bulk insert operation
+     */
+    protected PreparedStatement generateBulkInsertStatement(String table, int rows, String... cols)
+        throws DatabaseException, SQLException {
+
+        if (rows > 0) {
+            StringBuilder builder, rowbuilder;
+
+            if (!this.database.getDatabaseProductName().matches(".*(?i:oracle).*")) {
+                rowbuilder = new StringBuilder(2 + cols.length << 1);
```
awood: It's a little odd to see the shift operator here.  Is the goal to double `cols.length` and add 2?  If so, I feel like "2 + cols.length * 2" is going to be a lot clearer to future maintainers and the performance cost is probably negligible given that this code only runs once.

server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationTask.java
```diff
@@ -0,0 +1,455 @@
...
+                rowbuilder.deleteCharAt(rowbuilder.length() - 1).append(") ");
+
+                builder = new StringBuilder(30 + (20 * cols.length + table.length()) * rows);
+                builder.append("INSERT ALL ");
+
+                for (int i = 0; i < rows; ++i) {
+                    builder.append(rowbuilder);
+                }
+
+                builder.append("SELECT 1 FROM DUAL");
```
awood: Why do we need to select from dual here?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -362,24 +361,23 @@ private void updateContentRefs(Owner owner, Product product) {
     public Set<Content> getChangedContent(Owner owner, Collection<Content> content) {
         Set<Content> changed = Util.newSet();
 
-        log.debug("Syncing {} incoming content.", content.size());
+        log.debug("Syncing {} incoming content", content.size());
+        for (Content incoming : content) {
+            Content existing = this.contentCurator.lookupById(owner, incoming.getId());
```
mstead: Instead of doing N lookups here, would it be better to somehow hit the DB once to get all the existing incoming?

mstead: Would require more processing and id mapping and stuff, but might save some time.

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -89,16 +107,15 @@
     @Column
     private Long multiplier;
 
-    // NOTE: we need a product "type" so we can tell what class of
-    // product we are...
+    // TODO: we need a product "type" so we can tell what class of
+    // product we are... (02/15/16: is this still true?)
```
mstead: Product type is set as an attribute I believe.

server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationTask.java
```diff
@@ -0,0 +1,455 @@
...
+            int count = result.getInt(1);
+            result.close();
+
+            ResultSet orgids = this.executeQuery("SELECT id, account FROM cp_owner");
+            for (int index = 1; orgids.next(); ++index) {
+                String orgid = orgids.getString(1);
+                String account = orgids.getString(2);
+
+                this.logger.info(
+                    "Migrating data for org %s (%s) (%d of %d)", account, orgid, index, count
```
awood: Does this work?  Does slf4j support strfmt type templating?

server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationTask.java
Ceiu: Oracle shenanigans. The syntax requires that you do finish an INSERT ALL with a SELECT of some kind, and "SELECT 1" by itself isn't (always) valid there, so we use an internal table that should always exist.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
Ceiu: Yeah. Pretty sure I have a TODO somewhere around this line stating we should look into that eventually.

server/src/main/java/org/candlepin/model/Product.java
Ceiu: Okay, so that TODO/NOTE is safe to remove now?

server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationTask.java
Ceiu: We have our own wrapper which I modified a while back to accept printf-style inputs.

server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationTask.java
Ceiu: But... my nanoseconds when compiled without optimizations!

I suppose I can change this.

server/src/main/java/org/candlepin/model/Product.java
mstead: > Okay, so that TODO/NOTE is safe to remove now?

Ya, I believe so.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
mstead: > Yeah. Pretty sure I have a TODO somewhere around this line stating we should look into that eventually.
> 
> Edit: Yup. Line 346

:+1:  Missed that.

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -334,15 +273,346 @@ else if (boolAttrs != null && boolAttrs.contains(attr.getName())) {
...
+            .setParameter("1", updated.getUuid())
+            .setParameter("2", current.getUuid())
+            .setParameterList("3", ownerIds)
+            .executeUpdate();
+
+        log.debug("{} activation keys updated", akCount);
+
+        // Installed products
+        sql = "UPDATE cp2_installed_products SET product_uuid = ?1 " +
+            "WHERE product_uuid = ?2 AND consumer_id IN (SELECT id FROM cp_consumer WHERE owner_id IN (?3))";
```
mstead: I realize you have a note above regarding possibly reaching the IN clause limitation, but, I'm particularly concerned about the IN clause here since the sub select is targeting consumer ids in multiple owners. The result could easy blow that limit if I'm reading the query right.

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -334,15 +273,346 @@ else if (boolAttrs != null && boolAttrs.contains(attr.getName())) {
...
+            .setParameter("1", updated.getUuid())
+            .setParameter("2", current.getUuid())
+            .setParameterList("3", ownerIds)
+            .executeUpdate();
+
+        log.debug("{} pools updated", pdpCount);
+
+        sql = "UPDATE cp2_pool_provided_products SET product_uuid = ?1 " +
+            "WHERE product_uuid = ?2 AND pool_id IN (SELECT id FROM cp_pool WHERE owner_id IN (?3))";
+
```
mstead: Same IN concerns for selecting pool ids -- bonus pools can drive the pool counts up rather high as we've seen in some of the customer cases we've seen as of late.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -339,8 +313,29 @@ private DeveloperProducts getDevProductMap(Consumer consumer, String sku) {
...
+            if (sku.equals(product.getId()) &&
+                StringUtils.isEmpty(product.getAttributeValue("support_level"))) {
+
+                // if there is no SLA, apply the default
+                product.setAttribute("support_level", this.DEFAULT_DEV_SLA);
+            }
+
+            // TODO: Is there any guarantee we'll have the product locally by this point? If we have
+            // to create it, this blows up. May be a use case for bringing back the createOrUpdate
+            // method.
```
mstead: The products here may/may not exist. The marketing product (sku) on a dev pool would never come in as part of a refresh pools. Those products are intended to be 'global' and never associated with an actual subscription. There is also potential for a provided product to not exist as well, if it is net new and hasn't been brought in via a refresh yet.

i **think** we need createOrUpdate here.

server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationTask.java
```diff
@@ -0,0 +1,455 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+
+
+/**
+ * The PerOrgProductsMigrationTask performs the post-db upgrade data migration to the cp2_* tables.
+ */
```
mstead: Could probably use a description of what this is actually going to do. For example, what is it going to do and why?

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -133,7 +153,7 @@ public Product(String productId, String name, Owner owner) {
     public Product(String productId, String name, Owner owner, Long multiplier) {
         setId(productId);
         setName(name);
-        setOwner(owner);
+        setOwners(new HashSet<Owner>());
```
mstead: We are passing an owner here, but we are never adding it to the collection.

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -334,15 +273,346 @@ else if (boolAttrs != null && boolAttrs.contains(attr.getName())) {
...
+        int pppCount = session.createSQLQuery(sql)
+            .setParameter("1", updated.getUuid())
+            .setParameter("2", current.getUuid())
+            .setParameterList("3", ownerIds)
+            .executeUpdate();
+
+        log.debug("{} provided products updated", pppCount);
+
+        sql = "UPDATE cp2_pool_dprovided_products SET product_uuid = ?1 " +
+            "WHERE product_uuid = ?2 AND pool_id IN (SELECT id FROM cp_pool WHERE owner_id IN (?3))";
```
wottop: pool name typo?

server/src/main/java/org/candlepin/model/Product.java
Ceiu: Good catch; Saved me some pain on a few unit tests as well.

Ceiu:
To whomever finds courage to merge this behemoth into master, ping me before you do so I can squash the commit log down to something reasonable.

server/spec/product_versioning_spec.rb
```diff
@@ -0,0 +1,248 @@
...
+
+    prods = @cp.list_products_by_owner(owner2["key"])
+    prods.size.should == 1
+    prods[0]["uuid"].should == prod4["uuid"]
+    prods[0]["uuid"].should_not == prod1["uuid"]
+    prods[0]["uuid"].should_not == prod2["uuid"]
+  end
+
+  # TODO ? :
+  # - Updating a shared product used by another resource properly links the new product
```
mstead: Is this something you meant to do?

Ceiu: Kind of. It's something that's implicitly tested with the content changes, so I don't believe it's entirely dire to get it in. If we're going for 100% coverage though, I probably should.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -326,8 +325,8 @@ private void addProductContentToMap(Map<String, Content> contentMap, Product pro
 
             if (existing != null && !content.equals(existing)) {
                 log.warn(
-                    "Multiple versions of the same content found on a single subscription: {}.{}",
-                    (content.getOwner() != null ? content.getOwner().getId() : null), content.getId()
+                    "Multiple versions of the same content found on a single subscription: {}",
+                    content.getId()
```
mstead: I know this is old, but I've often wondered what the effect of this is on the overall process. Might be worth adding a little more context to the warning. Thoughts?

Ceiu: Not sure I follow. What do you have in mind?

mstead: We log a warning that multiple versions of the same content were found, but there's no info on what that means to the overall refresh process. Does one content replace the other? Are both versions still going to be available? If there is not going to be an impact to the resulting subscriptions, then perhaps we should change the log level to debug?

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -397,58 +560,143 @@ public boolean hasContent(String contentId) {
...
-        return 31;
+        catch (LazyInitializationException e) {
+            // One of the above collections (likely the first) has not been initialized and we're
+            // not able to fetch them. We still need a hashCode, and the caller is likely to run
+            // into this exception in the very near future, but we still need to generate
+            // something here. We'll treat it as if they were empty (and not added).
+
+            // Note that this only applies to lazy collections, so the product attributes are safe
+            // to check. ...for now.
+        }
```
mstead: @crog REMINDER: If the collections are no longer lazy, then we should remove this guard since we are relying on hascode for versioning.

mstead:
Just a couple of questions left. Sorry if the answers are in teh code changes. :)

1) Should there be an index on the version columns seeing as we are using them in the lookups? I'm no DB expert but I thought I'd ask.

2) Is there potential for a product (or content) to have duplicate version records. I could be missing something, but I'm not sure I see anything that would prevent this (if it is even possible). I wonder about the cases where multiple refreshes occur around the same time. Could it be possible that the stars align and a new product version is created at the same time resulting in duplicates?

Ceiu:
Yeah, should definitely index that column with as often as we'll be using it.

As far as duplicate records, I'm not sure I fully understand, but I'll take a stab at the answer anyway: It's possible, initially, for duplicate product/content records to exist after a migration, where all of the version strings will be null. However, over time as products are updated, the version field will be updated and the various records will start merging.

The concerning bit here is what happens after a migration where every product is null. As it's designed, the first few updates for an owner will pull in the majority of the products and compare all of them. It might be worth it to figure out something to do with the migration to either give each entity a version to avoid pulling the entire DB in on the first update.

mstead:
Is there a process that will clean up any product versions that are no longer being referenced by any owner?

Ceiu:
No explicit job/process. If my design isn't flawed, it shouldn't be necessary though. When a product/content is used by a single owner, updates to it are made in-place without rolling a new instance. Deletions on entities for a single owner work as they do in current master -- the entity is deleted at the Hibernate layer. So, ideally those two things should mean that eventually, deletions made to products should eventually remove the instance. I think the only way we end up with dangling references is during some race condition type stuff. But, with transaction blocks, hopefully that doesn't/can't happen.

mstead:
@Ceiu Other than the issues we've talked about, I think that concludes my review and it gets an ACK from me once that is done.

I'm not sure where others stand though.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -105,7 +105,7 @@ public Entitler(PoolManager pm, ConsumerCurator cc, I18n i18n, EventFactory evtF
 
     public List<Entitlement> bindByPool(String poolId, Consumer consumer,
         Integer quantity) {
-        log.info("Looking up pool to bind: " + poolId);
+        log.info("Looking up pool to bind: {}", poolId);
```
awood: :+1:

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -339,8 +313,45 @@ private DeveloperProducts getDevProductMap(Consumer consumer, String sku) {
...
-        return new DeveloperProducts(sku, prods);
+
+        Owner owner = consumer.getOwner();
+        Set<Product> products = new HashSet<Product>();
+
+        for (Product product : productAdapter.getProductsByIds(consumer.getOwner(), devProductIds)) {
+            for (ProductContent pc : product.getProductContent()) {
+                pc.getContent().addOwner(owner);
+
+                // VERSIONING TODO: We may need to hit the content curator here. Hopefully the update on
```
awood: Is this still an open task?

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -339,8 +313,45 @@ private DeveloperProducts getDevProductMap(Consumer consumer, String sku) {
...
+                // This should only happen if we try to update a product that hasn't yet been
+                // created. We'll do that here.
+                log.debug("Product doesn't yet exist locally; creating it now. {}", product);
+                product = productCurator.createProduct(product, consumer.getOwner());
+            }
+
+            products.add(product);
+        }
+
+
```
awood: Extraneous line here

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -446,4 +451,42 @@ private String getConcurrentModificationMessage() {
         return i18n.tr("Request failed due to concurrent modification, please re-try.");
     }
 
+    /**
+     * Performs a direct SQL update or delete operation with a collection by breaking the collection
+     * into chunks and repeatedly performing the update.
+     * <p/>
+     * The parameter receiving the collection chunks must be the last parameter in the query and
```
awood: Comment trails off here.

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -148,6 +172,103 @@ public Content(String id) {
...
+     * state of any null collections and any identifier fields.
+     *
+     * @param source
+     *  The Content instance to copy
+     */
+    protected Content(Content source) {
+        this.setUuid(source.getUuid());
+        this.setId(source.getId());
+
+        this.setCreated(source.getCreated());
```
awood: Since `Date` is not immutable in Java, will this create a problem where two objects are referencing the same Date object?  Should we make a defensive copy here?  `setCreated(new Date(source.getCreated().getTime()))`

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -148,6 +172,103 @@ public Content(String id) {
...
+        this.setLabel(source.getLabel());
+        this.setOwners(source.getOwners());
+        this.setName(source.getName());
+        this.setVendor(source.getVendor());
+        this.setContentUrl(source.getContentUrl());
+        this.setRequiredTags(source.getRequiredTags());
+        this.setReleaseVer(source.getReleaseVer());
+        this.setGpgUrl(source.getGpgUrl());
+        this.setMetadataExpire(source.getMetadataExpire());
+        this.setModifiedProductIds(source.getModifiedProductIds());
```
awood: Do we need a defensive copy here too?

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -148,6 +172,103 @@ public Content(String id) {
...
+        if (source.getArches() != null) {
+            this.setArches(source.getArches());
+        }
+
+        return this;
+    }
+
+    @Override
+    public Object clone() {
+        return new Content(this);
```
awood: `clone` is pretty tricky.  I'm looking at Josh Bloch's *Effective Java* and he's got a [long section](http://jtechies.blogspot.com/2012/07/item-11-override-clone-judiciously.html) on it with a bunch of warnings and comments.  Ultimately he advises just skipping `clone` if at all possible and either using a copy constructor which you wrote or a copy factory `public static Content newInstance(Content toCopy)`.

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -200,12 +321,66 @@ public void setId(String id) {
...
+     *
+     * @param owners
+     *  A collection of owners to be associated with this content
+     *
+     * @return
+     *  A reference to this content
+     */
+    public Content setOwners(Collection<Owner> owners) {
+        this.owners.clear();
+        if (owners != null) {
```
awood: Shouldn't we perform the null check before trying to call `clear()`?

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -122,9 +137,18 @@
     @Size(max = 255)
     private String arches;
 
+    @XmlTransient
+    @Column(name = "entity_version")
+    private Integer entityVersion;
+
+    @XmlTransient
+    @Column
+    private Integer locked;
```
awood: Why is `locked` not a boolean?

server/src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -64,50 +100,359 @@ public Content lookupByUuid(String uuid) {
...
+            .setParameterList("1", ownerIds)
+            .list();
+
+        sql = "UPDATE cp2_product_content SET content_uuid = ?1 " +
+            "WHERE content_uuid = ?2 AND product_uuid IN (?3)";
+
+        int pcCount = this.safeSQLUpdateWithCollection(sql, ids, updated.getUuid(), current.getUuid());
+        log.debug("{} product-content relations updated", pcCount);
+
+
```
awood: Extra whitespace here

server/src/main/java/org/candlepin/controller/Entitler.java
Ceiu: Potentially. Currently all the unit and spec tests seem to be okay with it, but I want to leave it in until we give it a bit more hands-on testing via IT.

server/src/main/java/org/candlepin/model/Content.java
Ceiu: Yeah. Looks like this is one of the collections that is modified and set externally in a number of places, so copying it is probably the way to go.

server/src/main/java/org/candlepin/model/Content.java
Ceiu: The part I don't like about either of those options directly is it violates the dependency injection design we're attempting to adhere to. Clone is nice since you don't need to know what specific instance you're cloning to get a valid clone, whereas with the static factory method and the copy constructor, you have to know, at design time, what instance you'll be working with.

I know in our case that's currently not a problem, but I'd rather avoid the problem the first go around instead of planting landmines for others.

I'll update the clone method to behave properly.

server/src/main/java/org/candlepin/model/Content.java
Ceiu: The first line clears the owners collection on the object, the second line checks the input collection.

server/src/main/java/org/candlepin/model/Content.java
Ceiu: SQL issues. When creating it as a boolean, Oracle complains. Letting liquibase use the pseudo-type for boolean created issues with the migration queries and typing. Dropping it to an integer was the only way I saw to make everything happy without adding a bunch of backend-specific branches everywhere.

server/src/main/java/org/candlepin/model/Content.java
Ceiu: Good catch

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -258,22 +365,78 @@ public void setName(String name) {
...
      */
-    public void setOwner(Owner owner) {
-        this.owner = owner;
+    public Product setOwners(Collection<Owner> owners) {
+        if (this.owners == null) {
+            this.owners = new HashSet<Owner>();
+        }
+
+        this.owners.clear();
+        if (owners != null) {
```
awood: Is this dealing with a field named `owners` and a parameter named `owners`?  That's kind of confusing if so.  How about changing the parameter name to "otherOwners" or something.

server/src/main/java/org/candlepin/resteasy/filter/VerifyAuthorizationFilter.java
```diff
@@ -169,41 +183,74 @@ protected Owner findOwner(Map<Verify, Object> argMap, Principal principal, Acces
             }
 
             if (Owner.class.isAssignableFrom(verifyType)) {
-                Access requiredAccess = defaultAccess;
```
awood: I might have missed this earlier, but why aren't we checking access level?

Ceiu: Once findOwnerFromParams returns, the next thing we do after putting the owner in the context is call hasAccess, which does those checks as well. If the access check context differs (I don't believe it does), then that may have been removed in error.

server/src/main/java/org/candlepin/model/Content.java
awood: Did you try using `@Type(type = "org.hibernate.type.NumericBooleanType")`?  I don't have a problem with it being an integer in the database, but we can instruct it in how to convert the integers back and forth to booleans.  See http://docs.jboss.org/hibernate/orm/4.2/manual/en-US/html_single/#types-value-basic

Ceiu: I didn't even get to the Hibernate level. The issues with juggling were happening at the jdbc level in the migration task -- I'd have to know which DB I was talking to and manually cast/set it appropriately.

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -148,6 +173,131 @@ public Content(String id) {
...
+        catch (CloneNotSupportedException e) {
+            // This should never happen.
+            throw new RuntimeException("Clone not supported", e);
+        }
+
+        copy.owners = null;
+        copy.modifiedProductIds = null;
+
+        if (this.getOwners() != null) {
+            copy.setOwners(this.getOwners());
```
awood: Won't this leave the clone with the same reference to the original's set of owners?

Ceiu: We've already discussed this in person, but for any other readers, setOwners copies the elements from the given collection into its own collection, so both instances will point to the same owner instances, but won't share a collection instance.

Ceiu:
Reminder: Ping me before the final decision to merge this so I can squash the commits down.

mstead:
retest this please

PR #1140, opened 2016-03-08
Vritant/client refactor

vritant:


tasks/rubocop.rake
```diff
@@ -0,0 +1,23 @@
...
+
+    desc 'Run RuboCop on select files'
+      require 'rubocop/rake_task'
+      RuboCop::RakeTask.new(:rubocop) do |task|
+      task.verbose = true
+      task.patterns =
+        ['server/client/ruby/candlepin.rb',
+         'server/spec/*.rb',
+         'server/client/ruby/test/*.rb']
+      puts "Running Rubocop on #{task.patterns}"
```
awood: Use `info()` as it is part of buildr's logging/output framework

tasks/rubocop.rake
```diff
@@ -0,0 +1,23 @@
...
+      task.formatters = ['fuubar']
```
awood: Nice.  I did not know this existed!

tasks/rubocop.rake
```diff
@@ -0,0 +1,23 @@
+module Rubocop
+  module ProjectExtension
+    include Extension
+
+    desc 'Run RuboCop on select files'
+      require 'rubocop/rake_task'
+      RuboCop::RakeTask.new(:rubocop) do |task|
+      task.verbose = true
```
awood: This block needs to be indented.  Also, I'm not seeing an `end` for this block.

tasks/rubocop.rake
```diff
@@ -0,0 +1,23 @@
...
+      task.patterns =
```
awood: I would make these patterns project specific.  Something like `File.join(project.base_dir, '**/*.rb')` as a default.  Go ahead and create a `Config` class so that we can set options in `buildfile`.  See `msgfmt.rake`, `gettext.rake`, or `liquibase.rake` for examples of setting up configuration objects for tasks.

vritant:
@awood done

PR #1141, opened 2016-03-08
Fix translations - un-escaped single quotes

vritant:
Automatically fixed translations using buildr validate_translation:fix


awood:
Cool.  But are these changes just going to get clobbered when we merge down from Zanata again?

vritant:
@awood  updated the documentation to add the validation step everytime: candlepin/candlepinproject.org#81
also, @Ceiu and I are collaborating so that we can run buildr check_all with every PR, which should prevent this from happening again.

vritant:
@awood does that unblock the merge?

awood:
> @awood does that unblock the merge?

Yes!

PR #1146, opened 2016-03-15
Request scope caching of cp_rules.updated 

nguyenfilip:
In our configuration, Google Guice creates a new instance of JsRunner for every Inject annotation. it means that  JsRunner is being constructed ~36 times during every request. In each of these constructions, the provider JsRunnerProvider is checking to the database if cp_rules.updated changed. This brings significant overhead. 

This PR introduces request scoped caching of cp_rules.updated value. The JsRunnerProvider now uses request scoped bean JsRunnerRequestCache in order to cache the value of cp_rules.updated for the duration of the request.

We cannot cache the value longer. The reason is that in the clustered environment, any cluster node of Candlepin can updated the cp_rules table. We don't have any clustered in-memory cache, hence synchronization on database level seems to be inevitable. 

There is an alternative route to implement this request scoped cache. That is configuring JsRunner to be request scoped and thus not constructed for every Injection annotation. I rather not went that route, to keep consistent programming model in Candlepin, where every injected object is a new instance.

### DIFF
--- a/server/src/main/java/org/candlepin/guice/CandlepinModule.java
+++ b/server/src/main/java/org/candlepin/guice/CandlepinModule.java
@@ -14,6 +14,13 @@
  */
 package org.candlepin.guice;
 
+import java.util.Properties;
+
+import javax.inject.Provider;
+import javax.validation.MessageInterpolator;
+import javax.validation.Validation;
+import javax.validation.ValidatorFactory;
+
 import org.candlepin.audit.AMQPBusPublisher;
 import org.candlepin.audit.EventSink;
 import org.candlepin.audit.EventSinkImpl;
@@ -125,6 +132,12 @@ import org.candlepin.util.DateSource;
 import org.candlepin.util.DateSourceImpl;
 import org.candlepin.util.ExpiryDateFunction;
 import org.candlepin.util.X509ExtensionUtil;
+import org.hibernate.cfg.beanvalidation.BeanValidationEventListener;
+import org.hibernate.validator.HibernateValidator;
+import org.hibernate.validator.HibernateValidatorConfiguration;
+import org.quartz.JobListener;
+import org.quartz.spi.JobFactory;
+import org.xnap.commons.i18n.I18n;
 
 import com.google.common.base.Function;
 import com.google.inject.AbstractModule;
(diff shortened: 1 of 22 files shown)


nguyenfilip:
retest this please

nguyenfilip:
@barnabycourt @vritant @mstead This is ready for review

nguyenfilip:
Ah, spec tests failed, I didn't expect that. Ok needs a bit more work then.

nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
Instead of annotation RequestScoped now I use CandlepinSingletonScoped. It has 'Singleton' in it's name but it's acutally request scoped.

The RequestScoped doesn't work for Quartz jobs!

barnabycourt:
If it is request scoped and not singleton scoped, should a new CandlepinRequestScoped scope be created to avoid future confusion?

server/src/main/java/org/candlepin/pinsetter/core/GuiceJobFactory.java
```diff
@@ -37,11 +37,11 @@
 
     private static Logger log = LoggerFactory.getLogger(GuiceJobFactory.class);
     private Injector injector;
-    private CandlepinSingletonScope candlepinSingletonScope;
+    private CandlepinRequestScope candlepinSingletonScope;
```
barnabycourt: This should probably rename the local variable as well

server/src/main/java/org/candlepin/guice/CandlepinModule.java
```diff
@@ -161,9 +160,9 @@ public CandlepinModule(Configuration config) {
     @Override
     public void configure() {
         // Bindings for our custom scope
-        CandlepinSingletonScope singletonScope = new CandlepinSingletonScope();
-        bindScope(CandlepinSingletonScoped.class, singletonScope);
-        bind(CandlepinSingletonScope.class).toInstance(singletonScope);
+        CandlepinRequestScope singletonScope = new CandlepinRequestScope();
```
barnabycourt: rename the local to requestScope instead of singletonScope

server/src/main/java/org/candlepin/servlet/filter/CandlepinScopeFilter.java
```diff
@@ -47,10 +47,10 @@
 
     private static Logger log = LoggerFactory.getLogger(CandlepinScopeFilter.class);
 
-    private final CandlepinSingletonScope singletonScope;
+    private final CandlepinRequestScope singletonScope;
```
barnabycourt: mixing singleton & request here

server/src/test/java/org/candlepin/TestingModules.java
```diff
@@ -236,9 +238,12 @@ public void configure() {
             // context listener
             bind(Configuration.class).toInstance(config);
 
-            CandlepinSingletonScope singletonScope = new CandlepinSingletonScope();
-            bindScope(CandlepinSingletonScoped.class, singletonScope);
-            bind(CandlepinSingletonScope.class).toInstance(singletonScope);
+            CandlepinRequestScope singletonScope = new CandlepinRequestScope();
```
barnabycourt: mixing singleton and request scope here

nguyenfilip:
Renamed CandlepinSingletonScope to CandlepinRequestScope. No implementation changes. It was just confusing naming. The comments on the CandlepinSingletonScope have been clearly stating its per request/per unit of work scope.

nguyenfilip:
retest this please

mstead:
ACK

PR #1147, opened 2016-03-16
Added JsonIgnore to reduce amount of data serialized

vritant:


vritant:

barnabycourt:
retest this please

vritant:
@barnabycourt has some issues, am looking into it now.

vritant:
@nguyenfilip  This is clear for merge!!!!

vritant:
@nguyenfilip I mean its ready for review, merge is up to you :grinning:

server/src/main/java/org/candlepin/model/Pool.java
```diff
@@ -244,10 +245,12 @@ public String getDescription() {
     @ManyToOne
     @JoinColumn(name = "product_uuid", nullable = false)
     @NotNull
+    @JsonIgnore
     private Product product;
 
     @ManyToOne
     @JoinColumn(name = "derived_product_uuid")
+    @JsonIgnore
```
nguyenfilip: Is it necessary to set @ JsonIgnore / JsonProperty on all: field, getter, setter?  From docs for Jackson 2.3 (our version) [1] I understand that because we already use JsonIgnore and JsonProperty annotations on both getter and setter, the JsonIgnore on the field is noop.

I tried listPools and updatePool locally without JsonIgnore on the field and it seems to work as expected. 

If there is no reason to have JsonIgnore on the field, I would remove, to limit confusion. If there is an explicit reason to have it on all three places, maybe a comment is in order here.

[1] http://fasterxml.github.io/jackson-annotations/javadoc/2.3.0/com/fasterxml/jackson/annotation/JsonIgnore.html

nguyenfilip: This comment also applies to other places where JsonIgnore has been introduced on a field.

PR #1150, opened 2016-03-21
Set nextUpdate correctly when modifying an empty CRL.

awood:
Due to the way the streaming process works, when X509CRLStreamWriter is
asked to modify an empty CRL, it uses a different method than it would
normally.  This different method was failing to update the nextUpdate
field on the generated CRL.


wottop:
ack

PR #1152, opened 2016-03-22
Added a serialVersionUID to the SystemPrincipal class

Ceiu:
- Added a serialVersionUID to SystemPrincipal to address a deserialization
  issue that would occur if a Candlepin deployment had scheduled tasks
  with that principal and was then upgraded to a newer version of CP where
  the SystemPrincipal had been updated


server/src/main/java/org/candlepin/auth/SystemPrincipal.java
```diff
@@ -18,6 +18,12 @@
  * SystemPrincipal
  */
 public class SystemPrincipal extends Principal {
+    /**
+     * Set the serialVersionUID to the value of the previous version auto-generated by Java, as the
+     * last known change won't affect functionality here. We just have to hope the auto-generated
+     * previous version is always generated the same way.
```
awood: [Serializable](http://docs.oracle.com/javase/7/docs/api/java/io/Serializable.html) says "... since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations".  I think we should be okay, but let's keep our fingers crossed. :worried: 

We should probably go through and add serialVersionUIDs to all the other serializable classes while we are thinking about it.

PR #1153, opened 2016-03-22
Allow configurable delta for nextUpdate on the CRL.

awood:
Although it's not the original intention, the nextUpdate value on a CRL
is frequently used to mean the expiration date for the CRL.  Previously,
Candlepin was setting nextUpdate to be one day away.  This patch makes
the number of days away configurable so that users can provide a little
more breathing room in case CRL generation fails or is delayed.

See http://security.stackexchange.com/a/55784


awood:
Fixes #1179

PR #1155, opened 2016-03-23
1320711: Fixed class cast exception in EntitleByProductsJob

mstead:
The job was being created with a List<String> for target pools,
but on execution, it was being casted to a HashSet.

Always cast to a Collection here, as that's what is passed
by the job creation's static method.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/tasks/EntitleByProductsJob.java
+++ b/server/src/main/java/org/candlepin/pinsetter/tasks/EntitleByProductsJob.java
@@ -34,7 +34,6 @@ import org.slf4j.LoggerFactory;
 
 import java.util.Collection;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
 
 /**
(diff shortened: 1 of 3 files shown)


awood:
Looks good to me.  Once Jenkins finishes it's thing I will merge

server/src/main/java/org/candlepin/pinsetter/tasks/EntitleByProductsJob.java
```diff
@@ -60,7 +59,7 @@ public void toExecute(JobExecutionContext ctx) throws JobExecutionException {
             String uuid = (String) map.get(JobStatus.TARGET_ID);
             Date entitleDate = (Date) map.get("entitle_date");
             String[] prodIds = (String[]) map.get("product_ids");
-            HashSet<String> fromPools = (HashSet<String>) map.get("from_pools");
+            Collection<String> fromPools = (Collection<String>) map.get("from_pools");
```
jmrodri: +1 this needs to match the ```JobDetail bindByProducts(String[] prodIds, Consumer consumer, Date entitleDate, Collection<String> fromPools)```

PR #1156, opened 2016-03-23
Awood/hanging indent

awood:
This fun little PR corrects line continuation indentation throughout our code base.  I've set Checkstyle to only allow one level of indentation on a continuation.  Annoyingly though, Checkstyle doesn't really do the right thing when it comes to nested method chains.  In those circumstances, I tagged the method `@SuppressWarnings("checkstyle:indentation")`.  My goal is to keep this particular check in enforcing mode for a week or so while everyone gets their editors configured accordingly.  Once everyone is set up, we can drop the check and the `SuppressWarnings` and just rely on everyone's editors to do the right thing for the most part.

### DIFF
--- a/server/src/main/java/org/candlepin/model/PoolCurator.java
+++ b/server/src/main/java/org/candlepin/model/PoolCurator.java
@@ -192,8 +192,7 @@ public class PoolCurator extends AbstractHibernateCurator<Pool> {
      * @return list of EntitlementPools
      */
     @Transactional
-    public List<Pool> listByOwnerAndProduct(Owner owner,
-            String productId) {
+    public List<Pool> listByOwnerAndProduct(Owner owner, String productId) {
         return listAvailableEntitlementPools(null, owner, productId, null, false);
     }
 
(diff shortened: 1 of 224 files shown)


server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -496,13 +494,11 @@ public Pool create(Pool entity) {
...
-            currentSession().createCriteria(Entitlement.class)
-                .setProjection(Projections.rowCount())
-                .createCriteria("pool")
-                    .createCriteria("sourceEntitlement")
-                        .add(Restrictions.idEq(entitlementId))
-                .list().get(0);
+        Integer result = (Integer) currentSession().createCriteria(Entitlement.class)
+            .setProjection(Projections.rowCount())
+            .createCriteria("pool")
+            .createCriteria("sourceEntitlement").add(Restrictions.idEq(entitlementId))
```
wottop: shouldn't the .add call be on a separate line? Its the same call level.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -251,15 +252,14 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
                 }
                 // update reporter id if it changed
                 if (jobReporterId != null && reportedOnConsumer != null &&
-                        reportedOnConsumer.getHypervisorId() != null &&
-                        (reportedOnConsumer.getHypervisorId().getReporterId() == null || !jobReporterId
-                                .contentEquals(reportedOnConsumer.getHypervisorId().getReporterId()))) {
+                    reportedOnConsumer.getHypervisorId() != null &&
+                    (reportedOnConsumer.getHypervisorId().getReporterId() == null || !jobReporterId
+                            .contentEquals(reportedOnConsumer.getHypervisorId().getReporterId()))) {
```
wottop: Still a double tab, maybe just one?

awood:
@wottop fixed those issues you found

wottop:
ack. will merge when tests are done.

awood:
buildbot, retest this please

wottop:
ack

PR #1157, opened 2016-03-28
Provide explicit serialVersionUIDs.

awood:
@crog This PR can be tacked onto #1152

### DIFF
--- a/common/src/main/java/org/candlepin/common/util/RpmVersionComparator.java
+++ b/common/src/main/java/org/candlepin/common/util/RpmVersionComparator.java
@@ -27,6 +27,7 @@ import java.util.Comparator;
  * check the test.
  */
 public class RpmVersionComparator implements Comparator<String>, Serializable {
+    private static final long serialVersionUID = 4036652892664704660L;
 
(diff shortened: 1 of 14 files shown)


mstead:
@awood it might be worth adding a little but of context to the commit message as to why this was needed.

Ceiu:
Were these UIDs randomly generated, or are they the values currently being used by Java (disregarding the compiler implementation details mentioned on the other PR)?

awood:
> Were these UIDs randomly generated, or are they the values currently being used by Java (disregarding the compiler implementation details mentioned on the other PR)?

They're compiler generated ones courtesy of Eclipse.

mstead:
What was the reason this was needed again?

Ceiu:
@mstead We have a case going from 0.9x to 2 where a task class changes slightly (just enough for the serialVersionUID to change), but the functionality doesn't. But due to the way serialization works, Java throws an exception all the same because the version UIDs don't match.

mstead:
Ah right. I recall that now. Thanks.

PR #1158, opened 2016-03-29
Fixed Optimistic Lock issue with ConsumerComplianceJob

mstead:
If a consumer was updated or deleted before the job was run
an OptimisticLockException was thrown and the job would fail.
Losing a compliance can be problematic as the consumer
status would be out of date.

The OLE would occur if the consumer was updated/deleted after
it was looked up and before it was locked.

This PR locks the consumer isntance on lookup, eliminating the
gap between fetch and lock.

If the Consumer does not exist, the job will now finish
successfully with an appropriate result message stating that
the targeted consumer could no longer be found.

### DIFF
--- a/server/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/server/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -59,6 +59,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 
 import javax.persistence.LockModeType;
+import javax.persistence.NoResultException;
 
 /**
  * ConsumerCurator
@@ -295,6 +296,27 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
         getEntityManager().lock(c, LockModeType.PESSIMISTIC_WRITE);
         return c;
 
+    }
+
+    /**
+     * Find a consumer by uuid and immediately lock it.
+     *
+     * @param consumerUuid the uuid of the target consumer.
+     *
+     * @return the Consumer matching the given uuid, null if the consumer was not found.
+     */
+    public Consumer lockAndLoadByUuid(String consumerUuid) {
+        String jpql = "select c from Consumer c where c.uuid = :consumerUuid";
+
+        try {
+            return getEntityManager().createQuery(jpql, Consumer.class)
+                .setParameter("consumerUuid", consumerUuid)
+                .setLockMode(LockModeType.PESSIMISTIC_WRITE)
+                .getSingleResult();
+        }
+        catch (NoResultException nre) {
+            return null;
+        }
     }
 
     @Transactional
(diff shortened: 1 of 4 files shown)


Ceiu:
Couple runs through the spec tests didn't turn any OLEs up with this branch.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -297,6 +298,28 @@ public Consumer lockAndLoad(Consumer c) {
...
     }
 
+    /**
+     * Find a consumer by uuid and immediately lock it.
+     *
+     * @param consumerUuid the uuid of the target consumer.
+     *
+     * @return the Consumer matching the given uuid, null if the consumer was not found.
+     */
+    @Transactional
```
vritant: does not make sense to @Transactional-ize the method that's sole purpose is to lock an entity.

server/src/test/java/org/candlepin/model/ConsumerCuratorTest.java
```diff
@@ -72,13 +72,12 @@ public void setUp() {
...
 
     @Test
-    @SuppressWarnings("unchecked")
     public void normalCreate() {
         Consumer consumer = new Consumer("testConsumer", "testUser", owner, ct);
         consumerCurator.create(consumer);
 
-        List<Product> results = entityManager().createQuery(
-                "select c from Consumer as c").getResultList();
+        List<Consumer> results = entityManager().createQuery(
```
vritant: :laughing:

vritant:
@mstead  Good for merge after:
- [ ]  checkstyle is fixed
- [ ]  @Transactional is removed from the lockAndLoad method

server/src/main/java/org/candlepin/model/ConsumerCurator.java
mstead: > does not make sense to @Transactional-ize the method that's sole purpose is to lock an entity.

Why not, as this would be the first time the entity is looked up. It is different than the lockAndLoad method as the consumer was expected to already have been looked up in Larger transaction. Perhaps I'm missing something.

server/src/test/java/org/candlepin/model/ConsumerCuratorTest.java
mstead: I know right!?  lol

server/src/main/java/org/candlepin/model/ConsumerCurator.java
vritant: @mstead Perhaps the same principle applies as there though?
If the calling method is not have a @Transactional , this method might:

1. open the transaction
1. fetch and lock the consumer
1. close the transaction - and thereby lose the lock

mstead: Ya, makes sense.

mstead:
@vritant All issues addressed. Let's wait until the spec tests run before merging.

Thanks for the review! Really liked the check list!! Very cool!

server/src/test/java/org/candlepin/model/ConsumerCuratorTest.java
jmrodri: Glad that wasn't my code. WHEW!

vritant:
@mstead merged, thanks for fixing my mess :)

PR #1159, opened 2016-03-29
1321630: Breaks up queries into smaller ones that will not exceed sta…

cnsnyder:
…ck depth on hypervisor checkin

### DIFF
--- a/server/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/server/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -27,16 +27,25 @@ import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 
 import org.hibernate.Criteria;
+import org.hibernate.EntityMode;
+import org.hibernate.HibernateException;
+import org.hibernate.Hibernate;
 import org.hibernate.Query;
 import org.hibernate.ReplicationMode;
+import org.hibernate.criterion.CriteriaQuery;
 import org.hibernate.criterion.Criterion;
 import org.hibernate.criterion.Disjunction;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
+import org.hibernate.engine.spi.TypedValue;
+import org.hibernate.internal.util.StringHelper;
+import org.hibernate.type.CompositeType;
+import org.hibernate.type.Type;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
@@ -60,6 +69,7 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
 
     private static final int MAX_FACT_STR_LENGTH = 255;
     private static final int NAME_LENGTH = 250;
+    private static final int MAX_IN_QUERY_LENGTH = 500;
     private static Logger log = LoggerFactory.getLogger(ConsumerCurator.class);
 
     public ConsumerCurator() {
@@ -205,9 +215,22 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
         // each guest ID.
 
         Query q = currentSession().createSQLQuery(sql);
-        q.setParameterList("guestids", possibleGuestIds);
-        q.setParameter("ownerid", owner.getId());
-        List<String> consumerUuids = q.list();
+        List<String> consumerUuids = new ArrayList<String>();
+
+        int fromIndex = 0;
+        int toIndex = fromIndex + MAX_IN_QUERY_LENGTH;
+
+        while (fromIndex < possibleGuestIds.size()) {
+            if (toIndex > possibleGuestIds.size()) {
+                toIndex = possibleGuestIds.size();
+            }
+            List<String> subList = possibleGuestIds.subList(fromIndex, toIndex);
+            q.setParameterList("guestids", subList);
+            q.setParameter("ownerid", owner.getId());
+            consumerUuids.addAll(q.list());
+            fromIndex = toIndex;
+            toIndex += MAX_IN_QUERY_LENGTH;
+        }
 
         if (consumerUuids == null || consumerUuids.size() == 0) {
             return guestConsumersMap;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -724,4 +756,86 @@ public int getConsumerEntitlementCount(Owner owner, ConsumerType type) {
...
             .list();
     }
+    /**
+     * InExpressionIgnoringCase
+     *
+     * Allows Criterion for case insensitive comparison to list of values
+     *
+     * @author wpoteat
+     */
+    public class InExpressionIgnoringCase implements Criterion {
```
Ceiu: What is the case where we want to implement this in code rather than advising people setup their databases (or set it up for them) with a case-insensitive collation?

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -724,4 +756,86 @@ public int getConsumerEntitlementCount(Owner owner, ConsumerType type) {
...
+
+        private final String propertyName;
+        private final Object[] values;
+
+        public InExpressionIgnoringCase(final String propertyName, final Object[] values) {
+            this.propertyName = propertyName;
+            this.values = values;
+        }
+
+        public String toSqlString(final Criteria criteria, final CriteriaQuery criteriaQuery)
```
Ceiu: When building large strings like this, use a proper StringBuilder. Java 1.7+ will somewhat do this for you, but it does so by creating several instances unnecessarily, when you could do it quicker/more efficiently by handling it yourself.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -724,4 +756,86 @@ public int getConsumerEntitlementCount(Owner owner, ConsumerType type) {
...
+                String cols = " ( " + StringHelper.join(" = lower(?) and ",
+                        wrappedLowerColumns) + "= lower(?) ) ";
+                cols = this.values.length > 0 ? StringHelper.repeat(cols + "or ",
+                        this.values.length - 1) + cols : "";
+                cols = " ( " + cols + " ) ";
+                return cols;
+            }
+        }
+
+        public TypedValue[] getTypedValues(final Criteria criteria, final CriteriaQuery criteriaQuery)
```
Ceiu: This isn't used anywhere in the PR. Is this something required by the API?

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -724,4 +756,86 @@ public int getConsumerEntitlementCount(Owner owner, ConsumerType type) {
...
+            throws HibernateException {
+            final ArrayList<TypedValue> list = new ArrayList<TypedValue>();
```
Ceiu: Good call on the use of ArrayList here -- converts to an array really quickly. You may want to try giving it a default size, though, to avoid worrying about several reallocations and copies as the contents grow.

Side note: Why are all the variables declared as final here?

Ceiu: Good call on the use of ArrayList here -- converts to an array really quickly. You may want to try giving it a default size, though, to avoid worrying about several reallocations and copies as the contents grow.

Side note: Why are all the variables declared as final here?

server/src/main/java/org/candlepin/model/ConsumerCurator.java
cnsnyder: Yes. It is required by Criterion.

alikins:
[non ack/nacking observation]

The commit that introduce the code involved here (85f8ae7a4c5bebb24802732b34e9b5ac40fa12e8):
```
    Batch lookups done during hypervisor checkins.
    
    When virt-who checks in we were doing one query per hypervisor to see if it has
    a consumer already registered, then two queries per each guest ID on that
    hypervisor to lookup it's guest consumer, and it's most recent host consumer.
    
    Instead we lookup all this data once for every hypervisor ID and guest ID in
    the checkin. The data is then passed around via a new DTO and used instead of
    going straight to the db for every entity.
```

I'm kind of wondering why we changed that in the first place.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -491,21 +514,30 @@ public Consumer getHost(String guestId, Owner owner) {
...
-            .add(Restrictions.eq("gconsumer.owner", owner))
-            .addOrder(Order.desc("checkins.updated"))
-            .setProjection(Projections.property("consumer"));
+        List<Consumer> hypervisors = new ArrayList<Consumer>();
+        int fromIndex = 0;
+        int toIndex = fromIndex + MAX_IN_QUERY_LENGTH;
 
-        // Note: may contain duplicates but is sorted so they appear later:
-        List<Consumer> hypervisors = crit.add(guestIdCrit).list();
+        Object[] ids = Util.getPossibleUuids(guestIds.toArray(new String [guestIds.size()])).toArray();
```
alikins: Can we ditch this entirely? In particular, the Util.getPossibleUuid bits that call Util.transformUuid.

iirc, that code was to handle some now old/unsupported version of vmware that generated bogus uuids. 

I believe it artificially inflates the number of guest uuids we search for. (double? worse?). 

I think even if we attempt this, I think some large portion of the uuid space may be eliminated by
being an invalid uuid. (ie, if they don't have a 4 for the first digit of the third partition ala https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29)

alikins:
[shiny thing that seems potentially related]
postgres recursive queries
http://www.postgresql.org/docs/9.1/static/queries-with.html

alikins:
[more useless commentary that is non ack/nacking...]

For sometime that is not now...

About 2/3rds of ConsumerCurator is guest/host map related. Time to extract that code to it's own classes?

Ceiu:
ACK
Code quality changes are being omitted, as this is a backport of code that has since been removed master.

Test results:
```
0.9.49-HOTFIX:
[root@6d4503399ae9 server]# ./client/ruby/cpc create_owner test
[root@6d4503399ae9 server]# curl -k -u admin:admin -T test_vodaphone.json -X POST "https://localhost:8443/candlepin/hypervisors?owner=test&create_missing=true"
{
  "displayMessage" : "Runtime Error ERROR: stack depth limit exceeded\n  Hint: Increase the configuration parameter \"max_stack_depth\" (currently 2048kB), after ensuring the platform's stack depth limit is adequate. at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse:2,102",
  "requestUuid" : "0e84e02d-594e-480e-a7ed-d9575eeff053"
}

==================================================

This PR:
[root@6d4503399ae9 server]# ./client/ruby/cpc create_owner test
[root@6d4503399ae9 server]# curl -k -u admin:admin -T test_vodaphone.json -X POST "https://localhost:8443/candlepin/hypervisors?owner=test&create_missing=true"
<very large blob of json results>
  "updated" : [ ],
  "unchanged" : [ ],
  "failedUpdate" : [ ]
}
```

PR #1162, opened 2016-03-30
Added config option for enabling/disabling per-org products

Ceiu:
- Added a new boolean property "candlepin.per_org_products" to candlepin.conf.
  This property can be used to configure whether or not products are treated
  as global (updating a product/content affects every org using it) or
  per-org (each org gets their own instance when they differ). The value is
  set to true by default.
- Added the dot-properties to handle reading from Candlepin configuration
  during our spec tests
- Added some new tests for verifying that toggling per-org products does
  not act unexpectedly
- Adding content to a product that already has the content being added is
  now a no-op

### DIFF
--- a/server/spec/content_versioning_spec.rb
+++ b/server/spec/content_versioning_spec.rb
@@ -1,12 +1,20 @@
 require 'date'
 require 'spec_helper'
 require 'candlepin_scenarios'
+require 'dot_properties'
 
 
 
 describe 'Content Versioning' do
 
   include CandlepinMethods
+
+  before(:all) do
+    props = DotProperties.load('/etc/candlepin/candlepin.conf')
+    per_org = props['candlepin.per_org_products']
+
+    @per_org = [nil, '1', 't', 'true', 'y', 'yes'].include?(per_org)
+  end
 
   it "creates one content instance when shared by multiple orgs" do
     owner1 = create_owner random_string('test_owner')
@@ -43,6 +51,8 @@ describe 'Content Versioning' do
   end
 
   it "creates a new content instance when an org updates a shared instance" do
+    pending("candlepin is not running with per-org products") if !@per_org
+
     owner1 = create_owner random_string('test_owner')
     owner2 = create_owner random_string('test_owner')
     owner3 = create_owner random_string('test_owner')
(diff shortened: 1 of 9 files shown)


barnabycourt:
@Ceiu Will this result in all consumers using a given product being marked as requiring a cert update or just the ones for the owner performing a pool refresh?

Ceiu:
Looks like entitlement cert updating was forgotten entirely in CP2. I'll be posting another PR that addresses that in the near future, as it's turning into a fairly large change which is well outside the scope of this PR.

server/spec/content_versioning_spec.rb
```diff
@@ -43,6 +51,8 @@
   end
 
   it "creates a new content instance when an org updates a shared instance" do
+    pending("candlepin is not running with per-org products") if !@per_org
```
awood: I've updated the RSpec version we're using.  If my PR hits before yours, you may need to switch `pending` to `skip`.  RSpec 3 now treats pending tests as failures and skipped tests as warnings.  RSpec 2 treated pending tests as warnings and didn't have skipped tests.

server/spec/content_versioning_spec.rb
```diff
@@ -78,6 +88,44 @@
...
+    label = "shared content"
+    type = "shared_content_type"
+    vendor = "generous vendor"
+    updated_upstream = Date.today
+
+    content1 = @cp.create_content(owner1["key"], name, id, label, type, vendor)
+    content2 = @cp.create_content(owner2["key"], name, id, label, type, vendor)
+    content3 = @cp.create_content(owner3["key"], name, id, label, type, vendor)
+
+    content1["uuid"].should == content2["uuid"]
```
awood: Also as part of the RSpec upgrade, switch `==` to `eq`.

```
content1["uuid"].should eq(content2["uuid"])
```

Ceiu:
Closing this PR, as this feature only touches a part of what needs to be done to handle the feature properly.

While the option works fine for what's presented here, the API is still hard-coded to require owners and owner-specific calls, which make little sense in the case where per-org products are disabled. The option would be to undeprecate the "old" owner-agnostic product and content resources, but only enable them when per-org products are disabled.

PR #1163, opened 2016-03-30
Products and Content now check the lock state when checking equality

Ceiu:
- Products and Content will now check their lock state when checking
  for equality or generating hash codes. This should prevent an issue
  where an entity could become locked or unlocked by merging or
  diverging from another entity in a different lock state.


Ceiu:
retest this please

wottop:
ack

PR #1165, opened 2016-03-31
Delete purged jobs from job tables in candlepin and qrtz

vritant:
@reviewer : please test the fix before you merge it in, I have not re-produced the issue. @Ceiu was able to help me verify on his setup that the principle of this fix works though, so I am confident.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
+++ b/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
@@ -67,6 +67,7 @@ public class PinsetterKernel {
 
     public static final String CRON_GROUP = "cron group";
     public static final String SINGLE_JOB_GROUP = "async group";
+    public static final String[] DELETED_JOBS = new String[] { "StatisticHistoryTask" };
 
     private static Logger log = LoggerFactory.getLogger(PinsetterKernel.class);
 
@@ -169,6 +170,22 @@ public class PinsetterKernel {
             }
             log.debug("Jobs implemented:" + jobImpls);
             Set<JobKey> jobKeys = scheduler.getJobKeys(jobGroupEquals(CRON_GROUP));
+
+            /*
+             * purge jobs that have been deleted from this version of Candlepin.
+             * This is necessary as we might not even have the Class definition
+             * at classpath, Hence any attempt at fetching the JobDetail by the
+             * Scheduler or JobStatus by the JobCurator will fail.
+             */
+            for (JobKey jobKey : jobKeys) {
+                for (String deletedJob : DELETED_JOBS) {
+                    if (jobKey.getName().contains(deletedJob)) {
+                        scheduler.deleteJob(jobKey);
+                        jobCurator.deleteJobNoStatusReturn(jobKey.getName());
+                        break;
+                    }
+                }
+            }
 
             for (String jobImpl : jobImpls) {
                 if (log.isDebugEnabled()) {
--- a/server/src/main/java/org/candlepin/model/JobCurator.java
+++ b/server/src/main/java/org/candlepin/model/JobCurator.java
@@ -72,6 +72,13 @@ public class JobCurator extends AbstractHibernateCurator<JobStatus> {
         }
     }
 
+    public int deleteJobNoStatusReturn(String jobId) {
+        return this.currentSession().createQuery(
+            "delete from JobStatus where id = :jobid")
+                .setParameter("jobid", jobId)
+                .executeUpdate();
+    }
+
     public int cleanupAllOldJobs(Date deadline) {
         return this.currentSession().createQuery(
             "delete from JobStatus where updated <= :date")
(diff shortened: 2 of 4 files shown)


server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
```diff
@@ -67,6 +67,7 @@
 
     public static final String CRON_GROUP = "cron group";
     public static final String SINGLE_JOB_GROUP = "async group";
+    public static final String[] DELETED_JOBS = new String[] { "StatisticHistoryTask" };
```
alikins: Is there someway we can introspect what the DELETED_JOBS should be instead of hard coding it?
Or possibly making it config?

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
```diff
@@ -170,6 +171,22 @@ else if (!isClustered()) {
             log.debug("Jobs implemented:" + jobImpls);
             Set<JobKey> jobKeys = scheduler.getJobKeys(jobGroupEquals(CRON_GROUP));
 
+            /*
+             * purge jobs that have been deleted from this version of Candlepin.
+             * This is necessary as we might not even have the Class definition
+             * at classpath, Hence any attempt at fetching the JobDetail by the
+             * Scheduler or JobStatus by the JobCurator will fail.
+             */
```
alikins: Could we provide/inject a replacement class for deleted job classes?  

Something like a base 'DeletedTask', and when the job goes away, we make say StatisticsHistoryTask a sub class of DeletedTask. That way the jobkey->task class reference doesn't break, but we could tell it's an existing job that can be ignored/deleted.

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
vritant: @alikins I did consider it, but thought that it was overkill.
* I figured that deleting a job to this extent is a code change, so it should not be an additional challenge to add a string to this array.
* using a config I thought would be an overkill too, given we would / should only change that config only every time we delete a job class.
* even if we add a config, we will hard code the defaults in ConfigProperties, in which case I figured its better to hard code with the code that uses it.

@alikins further thoughts?

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
vritant: @alikins this was an option that was considered before I got into implementation, but the approach that ended up being implemented seemed more cleaner in the long run, basically leaving zombie - noop- job definitions vs. just hard coded class names.

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
```diff
@@ -170,6 +171,22 @@ else if (!isClustered()) {
...
+            for (JobKey jobKey : jobKeys) {
+                for (String deletedJob : DELETED_JOBS) {
+                    if (jobKey.getName().contains(deletedJob)) {
```
wottop: why contains? aren't we going for direct name matches?

vritant: @wottop I could have gone with starts with, but I thought contains is safer.
the job key name / job id is the identifier of the specific job instance, and is usually like 
StatisticHistoryTask-d5d1d055-18bb-4103-a2f8-c08b134a3205

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
mstead: I think that allowing it to be configurable could get us out of a sticky situation where we missed adding to this array and released. Temp fix would be "Oh ya, we forgot that, just add it to the config file". No need to get a new build to fix the issue right away.

I mean, getting ourselves into that state wouldn't happen all that often, but, it would be nice to have something there just in case. Very simple to add.

I think that if we could as @alikins said, find a way to introspect what DELETED_JOBS should be, would be even better as we'd never have to worry about it again.

At the very least, I'd vote for a config option.

Ceiu: Could we catch the SerializationException and just delete it then? I'm a little adverse to such a thing, since one-off bugs could cause jobs to go missing, but maybe that's something that could be done only when a problem is detected?

wottop:
works as directed. ack.

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
vritant: @mstead after discussion with @wottop earlier today, im concerned about adding a config option.
In addition to the above reasons, essentially users of the application will be able to delete jobs that are mentioned in that config whether or not that job was deleted.... TL-DR, that option could be abused as well..

PR #1164, opened 2016-03-30
Fix non deterministic job status spec and make owner deletion an atom…

vritant:
…ic transaction

PR #1166, opened 2016-04-01
Stop lazy migration for guests on hypervisors

wottop:
The migration will be handled on the hypervisors' guests after the check in instead of on next entitlement listing. The entitlement check was originally moved because the process was synchronous for hypervisor reporting and that caused timeouts. Now that it is an asynchronous process, we will move the check back to each hypervisor report. This way the overall system metrics can be  accurate.


vritant:
cause for spec test failure is fixed in #1164, yet to be merged. Apologies for the inconvenience.

cnsnyder:
UpdateConsumer is not called for guest consumers on async hypervisor check in. The logic in UpdateConsumer should likely go in CheckForGuestsUpdate

wottop:
CheckForGuestsUpdate is before the hypervisor is stored. I added the code to the async process.

cnsnyder:
There is one case that doesn't seem to work for me here. If I've reported that a guest exists on two different hypervisors (lets call them A and B and assume we receive the reports in order A, B). Then if we later receive a report indicating that the guest is no longer on hypervisor B. If we subsequently receive a report indicating that the guest is still on hypervisor A (and no other changes have occurred to hypervisor A) then we end up not checking for migration of the guest back to hypervisor A.

cnsnyder:
We need to check for migration when the guests are added in the CheckForGuestsUpdate method.
We will already be iterating over the list of existing/added guest ids.

wottop:
buildbot: retest this please

server/spec/hypervisor_check_in_spec.rb
```diff
@@ -619,4 +619,122 @@ def create_virtwho_client(user)
...
   end
+
+it 'should allow a single guest to be migrated and revoke host-limited ents' do
+    owner = create_owner random_string('test_owner1')
+    user = user_client(owner, random_string("user"))
+    virtwho = create_virtwho_client(user)
+    uuid1 = random_string('system.uuid')
+    guests = [{:guestId => uuid1}]
+
+    host_consumer = user.register(random_string('host1'), :hypervisor, nil,
```
cnsnyder: Any reason this consumer is a "hypervisor" while the one just below it is a "system".

server/spec/hypervisor_check_in_spec.rb
```diff
@@ -619,4 +619,122 @@ def create_virtwho_client(user)
...
+    new_consumer_client = Candlepin.new(nil, nil, new_host_consumer['idCert']['cert'], new_host_consumer['idCert']['key'])
+    guest_client = Candlepin.new(nil, nil, guest_consumer['idCert']['cert'], guest_consumer['idCert']['key'])
+
+    virtwho.hypervisor_check_in(owner['key'], get_host_guest_mapping('hypervisor_id_1', [uuid1]))
+
+    # consumer_client.update_guestids(guests)
+    pools = consumer_client.list_pools :consumer => host_consumer['uuid']
+    pools.length.should == 1
+    consumer_client.consume_pool(pools[0]['id'], {:quantity => 1})
+    new_consumer_client.consume_pool(pools[0]['id'], {:quantity => 1})
```
cnsnyder: Should new_consumer_client consume from the same pool that consumer_client does?

cnsnyder:
Looks fine to me. Will merge after tests pass.

cnsnyder:
buildbot: retest this please

cnsnyder:
@wottop Do not pull. An amend went sideways. Please repush your local branch to this one.

cnsnyder:
@wottop Thanks!

PR #1167, opened 2016-04-04
@JsonIgnore Cdn on a pool, provide a seperate API to fetch a pool's Cdn

vritant:


vritant:
cause for spec test failure is fixed in #1164, yet to be merged. Apologies for the inconvenience.

vritant:
@candlepin/jenkins-bot test this please

wottop:
buildbot: retest this please

server/src/main/java/org/candlepin/resource/PoolResource.java
```diff
@@ -267,6 +268,29 @@ public void deletePool(@PathParam("pool_id") String id) {
...
+        @PathParam("pool_id") @Verify(Pool.class) String id) {
+
+        Pool pool = poolManager.find(id);
+
+        if (pool == null) {
+            throw new NotFoundException(i18n.tr(
+                "Subscription Pool with ID ''{0}'' could not be found.", id));
+        }
+
+        return pool.getCdn();
```
Ceiu: Do we want to throw an exception here if there is no CDN? That seems to be our general response to an entity not existing.

vritant: @Ceiu I thought about it, and I think today we essentially respond with a null on a pool object, and since there are no explicit requirements around this API, nor do we know any existing user , I thought it best to maintain that legacy behavior and not error out.
That being said, one could use the same arguments to argue otherwise ( consistency ).

I dont personally have any preference, as this isnt a feature used yet, cause I it might change if its ever used anyway, so I'll defer to any preference you have. thoughts?

vritant: when I say existing behavior, I mean using null to imply there is no Cdn.

Ceiu:
ACK

PR #1170, opened 2016-04-06
Move eventful to model for dependency on api jar

vritant:


Ceiu:
ACK

Pretty simple change. Will merge once all the tests complete.

PR #1180, opened 2016-04-11
Add RSPEC_FILTER option to cp-test

alikins:
quoting cleanup based on shellcheck suggestions

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -96,22 +96,24 @@ cleanup() {
 
 usage() {
     cat <<HELP
-    usage: cp-test [options]
+usage: cp-test [options]
 
-    OPTIONS:
-        -d  deploy a live candlepin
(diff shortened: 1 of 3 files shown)


docker/candlepin-base/cp-test
```diff
@@ -123,6 +123,7 @@ while getopts ":dtrHulsc:p:v" opt; do
             RSPEC="1"
             DEPLOY="1"
             ;;
+        T  ) RSPEC_FILTER="${OPTARG}";;
```
awood: Doesn't the `T` need to be added to the `getopts` command in the while loop?

wottop:
buildbot: retest this please

Ceiu:
Could we just modify -r to take the parameter or default to 1 if no filters were applied? I'd rather use cp-test -r "filter string" than -rT "filter string"

cnsnyder:
Works well for me. ACK

PR #1181, opened 2016-04-11
[wip/for reference] Add example logstash config for candlepin logs.

alikins:
Some example logstash configs for using candlepin with logstash. 

The configs here worked roughly a year ago, but likely have bitrotted against current versions of candlepin. May be useful for including in documentation.


wottop:
buildbot: retest this please

PR #1182, opened 2016-04-11
Alikins/gen certs bootstrap ca

alikins:
various tweaks to gen-certs, mostly for generating a root ca key/csr/ca cert, and signing
server certs with it (ie, no self-signed certs except for the ca root). Also creates an apache
style cert for use with mod_ssl.

Known issues:
- largely redundant upstream, since sat6 setup stuff does most of this
- kind of assumes a meaning openssl.cnf , which the script doesn't setup
- could go ahead and copy ca cert to /etc/rhsm/ca for dev 
- 300 lines of shell script seems like a totally reasonable way to bootstrap a CA right?

### DIFF
--- a/server/bin/gen-certs
+++ b/server/bin/gen-certs
@@ -2,20 +2,28 @@
 
 CERTS_HOME=/etc/candlepin/certs
 UPSTREAM_CERTS_HOME=$CERTS_HOME/upstream
+
 KEYSTORE_PASSWORD=$CERTS_HOME/keystore-password.txt
 CA_KEY=$CERTS_HOME/candlepin-ca.key
+CA_KEY_PASSWORD=$CERTS_HOME/candlepin-ca-password.txt
+
 CA_REDHAT_CERT=conf/candlepin-redhat-ca.crt
 CA_UPSTREAM_CERT=$UPSTREAM_CERTS_HOME/candlepin-redhat-ca.crt
+CA_CERT_BASE=$CERTS_HOME/candlepin-ca
+CA_CERT=$CA_CERT_BASE.crt
+CA_CERT_CSR=$CA_CERT_BASE.csr
+CA_CERT_P7=$CA_CERT_BASE.p7
+# private?
+CA_KEY=$CA_CERT_BASE.key
 CA_PUB_KEY=$CERTS_HOME/candlepin-ca-pub.key
-CA_CERT=$CERTS_HOME/candlepin-ca.crt
-CA_CERT_P7=$CA_CERT.p7
 # for mod_ssl testing
 SERVER_KEY_PASSWORD=$CERTS_HOME/candlepin-server-password.txt
 SERVER_KEY=$CERTS_HOME/candlepin-server.key
 SERVER_KEY_FULL=$SERVER_KEY.orig
-SERVER_CERT=$CERTS_HOME/candlepin-server.crt
-SERVER_CERT_REQ=$SERVER_CERT.req
-SERVER_CERT_P7=$SERVER_CERT.p7
+SERVER_CERT_BASE=$CERTS_HOME/candlepin-server
+SERVER_CERT=$SERVER_CERT_BASE.crt
+SERVER_CERT_CSR=$SERVER_CERT_BASE.csr
+SERVER_CERT_P7=$SERVER_CERT_BASE.p7
 KEYSTORE=$CERTS_HOME/keystore
 
 if [ -z "$CA_CERT_DAYS" ]; then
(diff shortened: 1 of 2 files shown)


server/bin/gen-certs
```diff
@@ -47,19 +81,153 @@ if [ ! -d $UPSTREAM_CERTS_HOME ]; then
...
 fi
 
-HOSTNAME=${HOSTNAME:-$(hostname)}
 
 if [ -f $CA_KEY ] && [ -f $CA_CERT ] && [ "$FORCECERT" != "1" ]; then
     echo "Certificates are already present."
 else
+    echo
+    echo "Creating CA private key password"
+    sudo su -c "echo $RANDOM > $CA_KEY_PASSWORD"
```
awood: Should these `echo`s be with a `-n`?  I know some programs can interpret the trailing newline as part of the actual password and thus making it impossible to type on the CLI.

server/bin/gen-certs
```diff
@@ -47,19 +81,153 @@ if [ ! -d $UPSTREAM_CERTS_HOME ]; then
...
+        -trustout -signkey $CA_KEY -days 365 -req -in $CA_CERT_CSR -out $CA_CERT
+
+    echo
+    echo "Creating apache server certs"
+    echo "Creating apache server private key password"
+    sudo su -c "echo $RANDOM > $SERVER_KEY_PASSWORD"
+
+    echo
+    echo "Creating apache server cert private key"
+    sudo openssl genrsa -out $SERVER_KEY -passout "file:$SERVER_KEY_PASSWORD" 1024
```
awood: Go big or go home.  4096 bits!

server/bin/gen-certs
```diff
@@ -47,19 +81,153 @@ if [ ! -d $UPSTREAM_CERTS_HOME ]; then
...
     echo "Certificates are already present."
 else
+    echo
+    echo "Creating CA private key password"
+    sudo su -c "echo $RANDOM > $CA_KEY_PASSWORD"
+
+    echo
     echo "Creating CA private key"
-    $SUDO openssl genrsa -out $CA_KEY 1024
+    sudo openssl genrsa -out $CA_KEY -passout "file:$CA_KEY_PASSWORD" 1024
```
awood: More bits plz!

server/bin/gen-certs
```diff
@@ -47,19 +81,153 @@ if [ ! -d $UPSTREAM_CERTS_HOME ]; then
...
+    echo "Create a tomcat server cert signing request based on keystore"
+    sudo keytool -certreq -v -alias "tomcat" -keystore $KEYSTORE -storepass $KEYSTORE_PASS -file $TOMCAT_SERVER_CERT_CSR
+
+    echo
+    echo "Sign the tomcat server cert with the CA cert/key $CA_CERT/$CA_KEY"
+    sudo openssl x509 -req -in $TOMCAT_SERVER_CERT_CSR -CA $CA_CERT -CAkey $CA_KEY -CAcreateserial -out $TOMCAT_SERVER_CERT -days 365
+
+    echo
+    echo "Import the signed tomcat server cert $TOMCAT_SERVER_CERT into keystore $KEYSTORE"
+    #    sudo keytool -importcert -trustcacerts -v -v -v -keystore $KEYSTORE \
```
awood: Did `importcert` not work?  Creating the key and csr in keytool but still having to go through the PKCS12 to JKS dance makes the first part kind of pointless.

wottop:
buildbot: retest this please

alikins:
I should note this pr is intended 'for reference' more so than something that needs to be merged directly. It was the result of some sort I had done when attempting to replicate some deploy problems with more complicated cert setups than just the self signed certs. 

I think it got revived at some point when I was attempting to change consumer id certs to contain more info (org, etc) with the intention that would allow the layer in front of candlepin to direct requests better for sharding (ie, the F5 and ssl termination layer would be able to find the org from the consumer cert and send a request to that orgs shard). But that never saw the light of day.

server/bin/gen-certs
alikins: Could be, for CA_KEY_PASSWORD the script only uses it as input for 'openssl genrsa'.

The docs for that are a little vague, but sounds like openssl at least expects there can be a newline there.

```
       file:pathname
                 the first line of pathname is the password. If the same pathname argument is supplied to -passin
                 and -passout arguments then the first line will be used for the input password and the next line
                 for the output password. pathname need not refer to a regular file: it could for example refer to
                 a device or named pipe.
```

server/bin/gen-certs
alikins: Yeah, should have paramaterized that.

That code also only supports RSA priv keys and not EC keys. For the purposes of testing, suppose it wouldn't hurt to ge several (or at least a big RSA and a EC key for at least the server keys) and create corresponding certs, likely defaulting to the RSA based keys and certs.

server/bin/gen-certs
alikins: Hmm, don't remember. 

There were some shenanigans with keytool that caused problems (maybe early versions of JKS/keytool wouldn't import PEM files?). I don't really remember the details unfortunately.
May be unneeded now.

alikins:
Another potential change here that I didn't include...

Once we have the CA certs, we could go ahead and copy the CA cert pem to /etc/rhsm/ca/.
Saves a step or two for local testing.

The step after that would be to import the CA cert into firefox/etc keyring.

Then it is a little easier to import any generated consumer id  or ent certs into firefox/nss.
My build/install/smoke test scripts for subman will convert the generated certs into pkcs7 and
copy to my home dir, if I decided to import into firefox. Unfortunately, there isn't a great way to
add them to a running firefox from the cli, tends to require a restart.

alikins:
Oh, one other thing.

This assumes /etc/pki/tls/openssl.cnf is a useful and valid openssl.cnf for generating the certs and csrs.
The script will inherit the 'OPENSSL_CONF' from the env/cli to point it elsewhere if one exists.

The script should probably gen one on the fly, but alas. Hate having to gen a config file for that, but because openssl.

And I've lost the tweaked one I was using. I can only assume it was magical and granted wishes.

PR #1188, opened 2016-04-13
Added script to manipulate manifests

vritant:


vritant:

nguyenfilip:
Even though this is handy it is just zip/unzip right now right? Is it worth merging this into our scripts? Maybe if you added a few more things such as a commandline option to upload manifest with the "force=SIGNATURE_CONFLICT&force=MANIFEST_SAME"?

vritant:
@nguyenfilip true, its just a handy script. but when I have tried to do it manually ( unzip, tweak, re-zip ) , I have always been hit by some or the other issue , for example, iirc you have to not write the directories as entries in the archive, and write files only... Ive always had to resort to a ruby script.

The thought of committing this script came out of discussion with @mstead , I thought it could be useful for others too. an optional upload sounds good, lemme work on that.

vritant:
@nguyenfilip done

nguyenfilip:
I think the script is cool and I think its ready to be merged. One question though.

The script unpacks then makes modifications and packs everything back again. Do you think this is the most useful workflow? When I had to do something with manifests I usually had the following workflow:
 - unpack manifest
 - make some manual modifications and inspection
 - possibly zip the manifest again
 - import the manifest to candlepin

So it would make sense to me that the script allows those actions individually.

vritant:
@nguyenfilip let me know if this works for you.. added -p to pause and ask the user to proceed( or quit ) at every step. this way we let people perform the steps individually or in a single shot if they so require.

nguyenfilip:
Absolutely perfect, ack.

PR #1189, opened 2016-04-14
Updated how rspec filters are specified

Ceiu:
- The rspec tests may now be filtered from the cp-test command by adding
  the filter string with the -r option
- Updated usage text for cp-test


alikins:
ack

PR #1190, opened 2016-04-15
Fixed cdn test for pools

vritant:


awood:
I am glad you fixed this, but why is this necessary in the first place?  How did we get broken spec tests in master?

vritant:
@awood 
- i'm to blame for this, I recently added this test, and since it is a non deterministic failure, neither me nor my reviewer caught it.
- when I got to fixing it, I figured it probably fails because the hostedtest service did not have the CDN objects set right.
- I could have fixed hostedtest service, but I realized that isn't needed, as all candlepin needs to do in hosted mode is to export a manifest with cdn label set on it if requested ( already test in export spec ) so I marked pending based on the mode.
- candlepin needs to fetch a cdn for a pool only in standalone mode, this feature ( that is not actually used by any existing consumers ) was designed for standalone use only.

nguyenfilip:
Tried this both in Standalone and Hosted. I think it makes sense that import shouldn't be tested in hosted. Ack.

PR #1192, opened 2016-04-15
Add missing @Consumes annotations.

awood:
Also add warning when a @Consumes is set to the */* media type but
actually reads the request body.  Currently, Candlepin only consumes
JSON so methods should be specific about that instead of using the
wildcard type as a shortcut.


Ceiu:

cnsnyder:
Looks good to me. (Non-{n,}acking comment): As we discussed there are a few places that we specify a @Consumes annotation on GETs which do no bind the body ([example here](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/resource/OwnerResource.java#L848)). Not sure if we should do anything about that.

PR #1194, opened 2016-04-15
unit test thrice to catch non deterministic tests

vritant:
- Unit test as many times as needed using `cp-test -uuuu`
- This helps to catch non-deterministic unit tests
- create buildr lint task to checkstyle, rubocop, and rpmlint
- add jenkins buildr task to do misc. validation
- add -b option to cp-test to execute arbitrary buildr task
- now you can run arbitrary buildr commands like:
  - `cp-test -b test:PoolManagerTest`
  - `cp-test -b jenkins`

if accepted, we will need to :
- [ ] re-build docker images
- [ ] rename checkstyle job and update it to execute `cp-test -b jenkins -l` instead of just `cp-test -l`

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -106,7 +106,8 @@ OPTIONS:
   -r [filter]      run rspec test suite (implies -d); may be filtered by test
                    suite and name
   -H               run rspec tests in "hosted" mode (implies -r and -d)
-  -u               run unit test suite
+  -u               run unit test suite N number of times, where N is number
+                   of times u was specifed in the arguments
   -l               run the linters against the code
   -s               run a bash shell when done
   -b <task>        execute arbitrary buildr task
@@ -243,10 +244,9 @@ fi
 
 if [ ! -z "$ARBITRARY" ]; then
     cd "$CP_HOME"
-    echo "Running buildr $ARBITRARY..."
     tee /var/log/candlepin/buildr.log < /tmp/teepipe &
-
-    buildr "${ARBITRARY}" > /tmp/teepipe 2>&1
+    echo "Running buildr $ARBITRARY..."
+    buildr $ARBITRARY > /tmp/teepipe 2>&1
 fi
 
 cleanup
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -23,6 +23,7 @@ export JAVA_HOME=/usr/lib/jvm/java-$JAVA_VERSION
 export SUPERVISOR=1
 export AUTOCONF=1
 export FORCECERT=1
+export UNITTEST=0
 
 trapex() {
     target="$1"
@@ -51,7 +52,8 @@ collect_artifacts() {
         move_artifact '/var/log/candlepin/audit.log' '/artifacts/'
         move_artifact '/var/log/candlepin/candlepin.log' '/artifacts/'
         move_artifact '/var/log/candlepin/error.log' '/artifacts/'
-        move_artifact '/var/log/candlepin/checkstyle.log' '/artifacts/'
+        move_artifact '/var/log/candlepin/lint.log' '/artifacts/'
+        move_artifact '/var/log/candlepin/buildr.log' '/artifacts/'
         move_artifact '/var/log/candlepin/unit_tests.log' '/artifacts/'
         move_artifact '/var/log/candlepin/rspec.log' '/artifacts/'
     fi
@@ -105,8 +107,9 @@ OPTIONS:
                    suite and name
   -H               run rspec tests in "hosted" mode (implies -r and -d)
   -u               run unit test suite
-  -l               run the linter against the code
+  -l               run the linters against the code
   -s               run a bash shell when done
+  -b <task>        execute arbitrary buildr task
   -c <ref>         git reference to checkout
   -p <project>     subproject to build (defaults to "server")
   -v               enable verbose/debug output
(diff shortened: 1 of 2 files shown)


vritant:

vritant:
@Ceiu not sure, but we talked about doing this a while back, lemme know what you think!

vritant:

Ceiu:
NACK at the moment

Not a huge fan of this. I don't want to start changing stuff intended for one thing (cp-test being our poorly-named container entry point) for something else entirely (Jenkins) -- especially when we're not even sure we'll be keeping that thing around very much longer.

The approach I'd rather see is to leave -l as it is, leave :check_all as it is, and add a -b option to cp-test to pass something directly through to buildr, then we can use that in the tasks directly. The new :jenkins task kind of falls into the above, but is probably fine. Either way, the existence of :jenkins shouldn't cause :check_all to change.

Additionally, running the unit tests multiple times is something that could/should be done within the Jenkins task, not automatically in our containers. I can assure you that when I'm running unit tests manually in the container, I won't want to be running them three times every time.

docker/candlepin-base/cp-test
```diff
@@ -183,18 +187,24 @@ if [ "$LINTER" == "1" ]; then
...
 # TODO: keep track of return code?
 cd "$CP_HOME/$PROJECT"
 
-if [ "$UNITTEST" == "1" ]; then
-    echo "Running unit tests..."
-
-    tee /var/log/candlepin/unit_tests.log < /tmp/teepipe &
-    buildr test > /tmp/teepipe 2>&1
+if [ "$UNITTEST" -gt 0 ]; then
+    echo "Running unit tests $UNITTEST times"
```
Ceiu: Parens on times here (ie: time(s))

docker/candlepin-base/cp-test
```diff
@@ -231,4 +241,12 @@ if [ "$RSPEC" == "1" ]; then
     buildr "${BUILDR_ARGS}" > /tmp/teepipe 2>&1
 fi
 
+if [ ! -z "$ARBITRARY" ]; then
+    cd "$CP_HOME"
+    echo "Running buildr $ARBITRARY..."
+    tee /var/log/candlepin/arbitrary.log < /tmp/teepipe &
```
Ceiu: Can we rename this to "buildr.log" ?

Ceiu:
Looks good. A couple of very minor changes, but that's it.

vritant:
@Ceiu done. please test locally before merging!!

docker/candlepin-base/cp-test
```diff
@@ -231,4 +241,12 @@ if [ "$RSPEC" == "1" ]; then
     buildr "${BUILDR_ARGS}" > /tmp/teepipe 2>&1
 fi
 
+if [ ! -z "$ARBITRARY" ]; then
+    cd "$CP_HOME"
+    echo "Running buildr $ARBITRARY..."
+    tee /var/log/candlepin/buildr.log < /tmp/teepipe &
```
Ceiu: One more request that came up during my manual testing:

We should flip the tee declaration and echo here, then output the "Running buldr ..." to the teepipe so it shows up in the log. Right now it doesn't, so the log doesn't show what command was run.

docker/candlepin-base/cp-test
```diff
@@ -183,18 +187,24 @@ if [ "$LINTER" == "1" ]; then
...
-
-    tee /var/log/candlepin/unit_tests.log < /tmp/teepipe &
-    buildr test > /tmp/teepipe 2>&1
+if [ "$UNITTEST" -gt 0 ]; then
+    echo "Running unit tests $UNITTEST time(s)"
+    # run $UNITTEST time(s) to increase chance of capturing
+    # non deterministic unit test failures.
+    for (( i=1; i<=$UNITTEST; i++ ))
+    do 
+        tee /var/log/candlepin/unit_tests.log < /tmp/teepipe &
```
Ceiu: Sorry, keep finding little things here.

The teepipe doesn't need to be setup every time, so we can probably move this out of the loop. Like the buildr comment, maybe we should change all the echos around so they show up in their respective logs.

docker/candlepin-base/cp-test
```diff
@@ -231,4 +241,12 @@ if [ "$RSPEC" == "1" ]; then
     buildr "${BUILDR_ARGS}" > /tmp/teepipe 2>&1
 fi
 
+if [ ! -z "$ARBITRARY" ]; then
+    cd "$CP_HOME"
+    echo "Running buildr $ARBITRARY..."
+    tee /var/log/candlepin/buildr.log < /tmp/teepipe &
+
+    buildr "${ARBITRARY}" > /tmp/teepipe 2>&1
```
Ceiu: If we change this line to ```buildr $ARBITRARY > ...```, we can issue compound commands via -b. Quick way to test this is with ```cp-test -b "clean test"```. As written now, it quotes the 'clean test' and builder attempts to run it as one task; changing it to not quote $ARBITRARY acts as a more accurate passthrough.

docker/candlepin-base/cp-test
```diff
@@ -105,16 +107,17 @@ OPTIONS:
                    suite and name
   -H               run rspec tests in "hosted" mode (implies -r and -d)
   -u               run unit test suite
```
Ceiu: We should update this to indicate that specifying it multiple times triggers multiple runs

Ceiu:
ACK

Will merge once automated tests finish

PR #1195, opened 2016-04-18
Swagger integration for development purposes 

nguyenfilip:
This is first PR against the development branch. It introduces swagger.json integration for OwnerResource and StatusResource. After deploying Candlepin, the swagger.json is available at: http://localhost:8080/candlepin/swagger.json and the Swagger UI documentation is available at http://localhost:8080/candlepin/docs


nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
retest this please

barnabycourt:
Hi, I was able to load the swagger.json directly an while I was able to load the page at  /docs the API itself was not displayed.

barnabycourt:
Also, Having gone through all the work on adding the annotations in order to make swagger work with our hibernate DAO objects, should we be considering updating the REST endpoints to use DTOs instead so we are not conflating database annotations & logic with serialization logic?

nguyenfilip:
@barnabycourt Most of the work needed to make Swagger work with our serialization is now done. While I believe that adding DTOs is a good move in the long term, I think it will be HUGE effort (need to adopt mapping of DAO entities to DTOs, also will need to add huge amounts of DTOs and test everything). So I think taht adding DTOs will be slow incremental process. That said Swagger can work as implemented in this PR and when we start adding DTOs it will still work for the new REST endpoints that will feature DTOs.

What do you see when you load http://localhost:8080/candlepin/docs? Do you at least get prompt for HTTP Basic AUTH? Or do you see nothing at all?

barnabycourt:
I get no prompt for auth, I see the green swagger banner with the text box to enter a url and the explore button but the page is white underneath the banner. I get no difference if I add the username & pwd in the url string. 

I deployed by checking out your PR and running "vagrant up"

barnabycourt:
In the Chrome log I see the following:
http://localhost:8080/candlepin/docs/lib/jquery-1.8.0.min.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/jquery.slideto.min.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/jquery.wiggle.min.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/js-yaml.min.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/jquery.ba-bbq.min.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/handlebars-2.0.0.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/lodash.min.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/highlight.9.1.0.pack_extended.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/swagger-oauth.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/backbone-min.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/highlight.9.1.0.pack.js Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/candlepin/docs/lib/jsoneditor.min.js Failed to load resource: the server responded with a status of 404 (Not Found)
swagger-ui.min.js:1 Uncaught TypeError: Handlebars.template is not a function
http://localhost:8080/candlepin/docs/lib/marked.js Failed to load resource: the server responded with a status of 404 (Not Found)
localhost/:34 Uncaught ReferenceError: $ is not defined

nguyenfilip:
My mistake:

```
$ grep lib .gitignore 
**/lib/
```

Nonetheless, I will try to look for online hosted versions of those libs if possible.

nguyenfilip:
@barnabycourt  Amended the lib/* files to the first commit "Initial swagger integration...."
@awood is it ok that this commit will introduce ~1.1MB of libs? I am not sure if we have any best practices on this. If it is a problem I can start kicking out the deps and link them from remote CDNs.

barnabycourt:
@nguyenfilip that worked much better. 
what is the GET /owners/{owner_key}/atom endpoint?

server/src/main/java/org/candlepin/guice/CandlepinModule.java
```diff
@@ -148,6 +137,24 @@
...
 import com.google.inject.persist.jpa.JpaPersistModule;
 
+import org.hibernate.cfg.beanvalidation.BeanValidationEventListener;
+import org.hibernate.validator.HibernateValidator;
+import org.hibernate.validator.HibernateValidatorConfiguration;
+import org.quartz.JobListener;
+import org.quartz.spi.JobFactory;
+import org.xnap.commons.i18n.I18n;
+
+import java.util.Properties;
```
awood: Please correct the automatic import ordering in your IDE.  If you are using Eclipse, you can go to Preferences, Java, Code Style, Organize Imports.  Then click the "Import..." button and select the file `project_conf/eclipse/candlepin.importorder`.

Using the same import order as everyone else will reduce superfluous changes in the diffs.

server/src/main/java/org/candlepin/guice/CandlepinModule.java
```diff
@@ -329,6 +338,25 @@ private void configureExporter() {
...
+         * Using this binding, the swagger.(json|xml) will be available
+         * for an authenticated user at context: URL/candlepin/swagger.json
+         */
+        bind(ApiListingResource.class);
+        bind(SwaggerSerializers.class);
+
+        bind(CandlepinSwaggerModelConverter.class);
+
+        BeanConfig beanConfig = new BeanConfig();
+        beanConfig.setSchemes(new String[] { "http" });
```
awood: I see later down that we are requiring basic auth?  If so, we need to server over https.

server/src/main/java/org/candlepin/swagger/CandlepinSwaggerModelConverter.java
```diff
@@ -0,0 +1,836 @@
...
+ * more details of how the serialization works. This converter make sure that
+ * when a model is nested (e.g. NestedOwner) it will include only those fields
+ * that are annotated for HateoasInclude
+ *
+ *
+ * @author fnguyen
+ *
+ */
+public class CandlepinSwaggerModelConverter extends AbstractModelConverter implements ModelConverter {
+    Logger LOGGER = LoggerFactory.getLogger(CandlepinSwaggerModelConverter.class);
```
awood: Usually, the loggers are marked as `private static final`.

server/src/main/java/org/candlepin/swagger/CandlepinSwaggerModelConverter.java
```diff
@@ -0,0 +1,836 @@
...
+            Class<?> cls = (Class<?>) type;
+            if (cls.getName().equals("javax.ws.rs.Response")) {
+                return true;
+            }
+        }
+        else {
+            if (type instanceof com.fasterxml.jackson.core.type.ResolvedType) {
+                com.fasterxml.jackson.core.type.ResolvedType rt =
+                    (com.fasterxml.jackson.core.type.ResolvedType) type;
+                LOGGER.debug("Can't check class " + type + ", " + rt.getRawClass().getName());
```
awood:
```
LOGGER.debug("Can't check class {}, {}", type, rt.getRawClass().getName())
```

server/src/main/java/org/candlepin/swagger/CandlepinSwaggerModelConverter.java
```diff
@@ -0,0 +1,836 @@
...
+         * real type of field is passed to _mapper
+         */
+        if (type instanceof NestedComplexType) {
+            propType = pMapper.constructType(((NestedComplexType) type).getOriginalType());
+        }
+        else {
+            propType = pMapper.constructType(type);
+        }
+
+        LOGGER.debug("resolveProperty " + propType);
```
awood: You can use curly brace expansion here too

server/src/main/java/org/candlepin/swagger/CandlepinSwaggerModelConverter.java
```diff
@@ -0,0 +1,836 @@
...
+        }
+        JAXBAnnotationsHelper.apply(member, property);
+    }
+
+    private Property resolveApiAnnotated(ModelConverterContext context, Property property,
+        Annotation[] annotations, ApiModelProperty mp, JavaType propType) {
+        String or = mp.dataType();
+
+        JavaType innerJavaType = null;
+        LOGGER.debug("overriding datatype from " + propType + " to " + or);
```
awood: Curly brace expansion

server/src/main/webapp/docs/css/typography.css
```diff
@@ -0,0 +1,14 @@
...
+  font-style: normal;
+  font-weight: 400;
+  src: local('Droid Sans'), local('DroidSans'), url('../fonts/DroidSans.ttf') format('truetype');
+}
+/* Google Font's Droid Sans Bold */
+@font-face {
+  font-family: 'Droid Sans';
+  font-style: normal;
+  font-weight: 700;
+  src: local('Droid Sans Bold'), local('DroidSans-Bold'), url('../fonts/DroidSans-Bold.ttf') format('truetype');
```
awood: Let's just use `font-family: sans-serif` in the CSS so we don't have to package our own fonts.  This [font-face](http://www.paulirish.com/2009/bulletproof-font-face-implementation-syntax/) definition won't even work for a lot of browsers.

awood:
Please reformat the commit messages a bit so they will display nicely in `tig` and github: 80 character or less summary line, blank line, and then however much text is necessary for explanation but wrapped at 80 characters per line.  If you use vim, `autocmd FileType gitcommit setlocal textwidth=80`

nguyenfilip:
@barnabycourt The atom feed works now. I just uncommented it. I vividly believe that it didn't work before (Swagger UI not loaded when I included it).

Wrt javadoc I don't think there is a simple way to do it right now, since the Swagger doc is generated at runtime from annotations.

server/src/main/java/org/candlepin/guice/CandlepinModule.java
nguyenfilip: What is the correct order? From the candlepin.importorder it seems that Eclipse is ordering the imports correctly.

org.candlepin and org.hibernate shouldn't be next to each other and com.* should be between them

candlepin.importorder:
{{{
7=javax
6=java
5=ch
4=net
3=org
2=com
1=com.redhat.rhn
0=org.candlepin
}}}

server/src/main/java/org/candlepin/guice/CandlepinModule.java
nguyenfilip: added HTTPS

server/src/main/java/org/candlepin/swagger/CandlepinSwaggerModelConverter.java
nguyenfilip: renamed and fixed all the logging

nguyenfilip:
@awood Fixed everything besides the import order (see comment there)

@barnabycourt Do you want to have the UI deployed separately? I like the fact that its bundled with Candlepin but if you prefer it separately I can do that. It will require a few more things to handle CORS but it should be easy. 

@awood @barnabycourt : And about the Javadoc comments. The only good way to make it work is to is to use swagger-doclet subproject that Alex mentioned, instead of swagger-core. I subjectively feel that swagger-core is more used project with wider adoption. Moreover, adoptiong doclet means that I would have to drop the current implementation of Converters (to respect our serialization rules) and implement similar thing in doclet (if thats possible). 

I lean towards using Annotations even though it is less convenient than javadoc doclets.

server/src/main/java/org/candlepin/guice/CandlepinModule.java
awood: > What is the correct order? From the candlepin.importorder it seems that Eclipse is ordering the imports correctly.

:frowning: Sorry!  Your commit corrected someone else's ordering error and I wasn't paying attention!  We should probably add `io` as an entry in the importorder file though so that the `javax` and `java` packages remain at the bottom.

awood:
> @awood Fixed everything besides the import order (see comment there)

Thanks!  The only thing I still am concerned about is if the application is served over HTTP will it still ask for basic auth?  Serving over HTTP should either require no authentication or not be provided at all since if someone is confused or tricked into sending their credentials over HTTP, there's nothing protecting them.

As for the javadoc, I agree.  It's just too much trouble for a marginal benefit.

nguyenfilip:
@awood I agree. Shouldn't that be Candlepin's default? What do you think about the following change:

https://github.com/candlepin/candlepin/pull/1203

Of course I can create special filter for swagger.xml and candlepin/docs context, if you think 1203 is overkill. That would be a more defensive change :-)

nguyenfilip:
@barnabycourt rebased and squashed some commits. I would still like to see some commits to be separate, e.g. commit 'Candlepin specific changes to ModelConverter' isolates changes that I made during my development (plus checkstyle fixes).

After the tests pass this should be ready for merge.

nguyenfilip:
retest this please

PR #1197, opened 2016-04-18
Fixed quoting for deploy command in cp-test

Ceiu:


Ceiu:
Note for reviewers: The hosted tests will fail until this is merged in and the Docker images are rebuilt.

PR #1202, opened 2016-04-20
Fix guest migration spec test - sleep appropriately for mysql

vritant:


nguyenfilip:
@vritant How does the sleep help with the sorting?

vritant:
@nguyenfilip ConsumerCurator.getHost sorts by updated time. so a sleep of 1 second guarantees that the updated date on the second consumer will not match the updated date on the first consumer, and will be after it.

nguyenfilip:
@vritant Omg :-)

PR #1203, opened 2016-04-21
Authenticate over SSL only

nguyenfilip:
Candlepin shouldn't send WWW-Authenticate when the client communicates using
insecure channel, instead Candlepin should send back 403 Forbidden

### DIFF
--- a/server/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/server/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -46,6 +46,16 @@ public class ConfigProperties {
 
     public static final String CANDLEPIN_URL = "candlepin.url";
 
+    /**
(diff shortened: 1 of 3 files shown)


awood:
Instead of a 403, I'm thinking either a 400 or maybe a 301 or 308?  But otherwise looks good!

nguyenfilip:
I used 400. Thanks!

awood:
@nguyenfilip Looks good.  One final thought: should we allow this check to be disabled via a config value (where the SSL check defaults to on)?  I can see some utility in allowing Candlepin to be deployed over HTTP to allow for easier debugging.  With the SSL check always on, that wouldn't be possible.

nguyenfilip:
@awood Good idea! Added the config option.

awood:
:tada:  Nice work!

PR #1205, opened 2016-04-25
Updated rhel7-base docker container to no longer use dead link for epel

Ceiu:


cnsnyder:
Works for me! Merging.

PR #1206, opened 2016-04-25
Tests to ensure that a master pool of unlimited quantity behaves in t…

wottop:
…he system

### DIFF
--- a/server/spec/pool_unlimited_master_spec.rb
+++ b/server/spec/pool_unlimited_master_spec.rb
@@ -1,0 +1,91 @@
+require 'spec_helper'
+require 'candlepin_scenarios'
+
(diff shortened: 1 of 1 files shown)


Ceiu:
Looks good and works, but I'm lacking context on the problem. For unlimited master pool, does that mean we're still only supporting those with a virt_limit of unlimited, or do we also need to support pools with some kind of unlimited value directly (ie: -1 or some such)?

wottop:
We need to support the quantity coming in from the adapter being unlimited for a master pool. This is independent of it having a virt limit.

Ceiu:
Never mind. My initial question came from a repeated misreading of "consume_pool" as "create_pool."

ACK

PR #1207, opened 2016-04-26
Updated component versions for commons-io and httpclient

Ceiu:
- Candlepin now uses commons-io 1.4 and httpclient 4.3.2


awood:
ACK

PR #1214, opened 2016-04-27
Added a DTO for activation keys

Ceiu:
- Added a DTO for activation keys to better match the API output
  in 0.9.x
- ActivationKeyResource now uses the ActivationKeyData DTO for
  input and output across all its methods


server/src/main/java/org/candlepin/resource/dto/ActivationKeyData.java
```diff
@@ -0,0 +1,325 @@
...
+
+        if (overrides != null) {
+            this.contentOverrides.addAll(overrides);
+        }
+
+        return this;
+    }
+
+    public Collection<AKContentOverride> getContentOverrides() {
+        return this.contentOverrides;
```
nguyenfilip: Just a cosmetic thing here. I think its great to have encapsulated methods (removeContentOverride, setContentOverrides, addContentOverride). Having those methods I would use unmodifiable collection when returning getContentOverrides so that clients cannot accidently modify the contentOverrides  collection directly:

```
    public Collection<AKContentOverride> getContentOverrides() {
        return Collections.unmodifiableCollection(this.contentOverrides);
    }
```  

The same goes for the values() sets that are being returned for fields: pools, products

nguyenfilip:
I verified the chunked streaming encoding for /activation_keys

Other than the cosmetic suggestion above, I think this is OK. Let me know if you want to go with the suggestoin. Otherwise I will merge this

server/src/main/java/org/candlepin/resource/dto/ActivationKeyData.java
Ceiu: Yeah... only reason I didn't was because it would due to laziness, but you're right and we should probably just do it cleanly to begin with.

I'll make the change and throw in some documentation.

server/src/main/java/org/candlepin/resource/dto/ActivationKeyData.java
```diff
@@ -0,0 +1,757 @@
...
+
+/**
+ * DTO representing the activation key data exposed to the API layer.
+ */
+@XmlRootElement
+public class ActivationKeyData {
+
+    /**
+     * DTO class representing activation key pool data.
+     * <p/>
```
nguyenfilip: I think that in javadoc you should not use <p/> but you should use 'broken' markup <p> see [1]. It seems that in other parts of Candlepin javadoc, we tend to use broken markup as well.

[1] http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format

Ceiu: Fixed here and in 14 other files.

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
```diff
@@ -279,9 +302,28 @@ public ActivationKey removeProductIdFromKey(
...
+            .locateMapper(Object.class, MediaType.APPLICATION_JSON_TYPE);
+
+        StreamingOutput output = new StreamingOutput() {
+            @Override
+            public void write(OutputStream stream) throws IOException, WebApplicationException {
+                JsonGenerator generator = mapper.getJsonFactory().createGenerator(stream);
+                generator.writeStartArray();
+
+                for (ActivationKey key : keyList) {
+                    mapper.writeValue(generator, new ActivationKeyData(key));
```
barnabycourt: Do you need to keep creating new ActivationKeyData objects? Why not re-use one and reload it using the fromModel() method?

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
```diff
@@ -279,9 +302,28 @@ public ActivationKey removeProductIdFromKey(
      */
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    public List<ActivationKey> findActivationKey() {
-        List<ActivationKey> keyList = activationKeyCurator.listAll();
-        return keyList;
+    public Response findActivationKey() {
+        // TODO: Replace this with use of a cursor/iterator
```
barnabycourt: Are you going to fill in this TODO?

Ceiu: I can, but it can't really be done until the cursor stuff goes in.

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
Ceiu: I haven't tested it, but my initial reaction is that we don't know exactly when mapper converts an object to JSON for output. If it caches stuff until they're flushed or some such, we could end up clobbering a bunch of data.

barnabycourt: fair point, the generator may cache values until flush gets called

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
barnabycourt: ack

nguyenfilip: @Ceiu The cursor stuff is in #1215 right? Do you want to wait for it to hit the master?

Ceiu: Not really. There will probably be changes to be made there, and the TODO called out here would likely be touched up as part of a bigger patch that addresses several similar calls -- especially in light of some performance details we got today.

PR #1215, opened 2016-04-28
Added cursor support to Candlepin and its curators

Ceiu:
- Moved unboundedInCriterion to CPRestrictions.in
- Updated Swagger annotation for listing products
- Added the CandlepinQuery interface, which provides methods for listing
  entities, scrolling or iterating through entities as well as fetching the
  row count and setting the offsets and ordering.
- Added the ResultIterator interface, and two implementations, which provides
  standard iterator methods, as well as a close method for closing the backing
  cursor
- Removed some extraneous debug logging
- Added some whitespace in a few places for improved readability
- The collection-returning methods in several curators have been updated to
  return CandlepinQuery objects instead of static lists
- Updated various calls to curator methods accordingly
- Updated the unit tests to either mock the CandlepinQuery class
  or use CandlepinQuery.list for processing test results
- Fixed an issue in ExporterTest which would cause failures
  depending on which version.properties file was loaded
- Added some methods for building Page instances from CandlepinQuery objects
- Added the CandlepinQueryInterceptor, which processes CandlepinQuery
  instances and writes the output using a cursor, applying any paging
  specified in the request
- Several classes have been refactored or entirely replaced with the
  addition of the new interceptor
- Added the new TransformedCandlepinQuery decorator class, which
  transparently applies transformation to the results of a CandlepinQuery
- Added cursor support for pagination


Ceiu:
Note for reviewers:
Give some attention to the singleton behavior of JsonProvider. If there's a cleaner way to do what I need to do there, let me know.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -506,6 +495,212 @@ private String getConcurrentModificationMessage() {
...
+            query.setCacheMode(CacheMode.GET);
+        }
+
+        ScrollableResults cursor = query.scroll(ScrollMode.FORWARD_ONLY);
+        int count = 0;
+
+        try {
+            boolean cont = true;
+
+            if (evict) {
```
nguyenfilip: This entire block can be simplified to [1]. Are you concerned about repeated evaluation of "if statement" in the while loop?

[1]
```java
//            if (evict) {
                Session session = this.currentSession();

                while (cont && cursor.next()) {
                    E result = (E) cursor.get(column);

                    cont = processor.process(result);
                    if (evict)
                        session.evict(result);

                    ++count;
                }
//            }
//            else {
//                while (cont && cursor.next()) {
//                    cont = processor.process((E) cursor.get(column));
//                    ++count;
//                }
//            }
```

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -506,6 +495,212 @@ private String getConcurrentModificationMessage() {
...
+     * @param query
+     *  The query through which to scroll
+     *
+     * @param processor
+     *  A ResultProcessor instance to use for processing each result
+     *
+     * @return
+     *  the number of rows processed and sent to the result processor
+     */
+    protected int scroll(Criteria query, ResultProcessor<E> processor) {
```
nguyenfilip: Later, we will also need to overload these to take org.hibernate.Query as a parameter. Not just Criteria. Unfortunatelly the two interfaces have no common ancestor. But both have the scroll capability.

server/src/main/java/org/candlepin/model/ResultIterator.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Override
+    public void remove() {
+        throw new UnsupportedOperationException(
+            "The remove operation is not supported on ResultIterator instances."
+        );
+    }
+
+    public void close() {
```
nguyenfilip: missing  @ Override annotation

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -506,6 +495,212 @@ private String getConcurrentModificationMessage() {
...
+        if (query == null) {
+            throw new IllegalArgumentException("query is null");
+        }
+
+        if (evict) {
+            query.setCacheMode(CacheMode.GET);
+        }
+
+        ScrollableResults cursor = query.scroll(ScrollMode.FORWARD_ONLY);
+        return new ResultIterator(this.currentSession(), cursor, column, evict);
```
nguyenfilip: missing type parameter <E>

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
Ceiu: Yeah, that's why it was broken out. As soon as we have bits of code that evict and some that don't, the compiler won't optimize the loop down (at least, last I checked). Since it's small enough, simple enough and likely to be run tens/hundreds of thousands of times, I felt it's a good place to manually optimize.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
Ceiu: Good catch. I'll add that in.

server/src/main/java/org/candlepin/model/ResultIterator.java
Ceiu: ack

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
Ceiu: ack

server/src/test/java/org/candlepin/test/ResultIteratorTest.java
```diff
@@ -0,0 +1,145 @@
...
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.model;
```
nguyenfilip: Incorrect package, should be 'org.candlepin.test'

server/src/test/java/org/candlepin/model/IterableStreamingOutputTest.java
```diff
@@ -0,0 +1,120 @@
...
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.resteasy;
```
nguyenfilip: Also incorrect package, should be org.candlepin.model

server/src/test/java/org/candlepin/test/ResultIteratorTest.java
```diff
@@ -0,0 +1,145 @@
...
+    public void setup() {
+        this.session = (Session) this.entityManager().getDelegate();
+    }
+
+    @Test
+    public void testHasNextWithElements() {
+        this.ownerCurator.create(TestUtil.createOwner());
+        Query query = this.session.createQuery("SELECT o FROM Owner o");
+
+        ResultIterator<Owner> iterator = new ResultIterator(
```
nguyenfilip: Missing type parameter in the constructor: 

```
new ResultIterator<Owner>
```

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -182,15 +190,17 @@ protected Content fetchContent(Owner owner, String contentId) {
     @ApiOperation(notes = "Retrieves a list of Products", value = "list")
```
nguyenfilip: Because you are using Response, you need to tell  Swagger which entity will be returned. Use response and responseContainer parameters for this:

```java 
 @ApiOperation(response=Product.class,
        responseContainer="List",
    ...
```

nguyenfilip:
Looks good. After the test pass (missing imports) and Swagger doc is fixed, I will merge this.

nguyenfilip:
@Ceiu Seems that only a small checkstyle complaint needs to be cleaned up.

Ceiu:
Will fix the checkstyle issues after rebasing before we merge. This PR is currently waiting until our next branch/tag to be merged in.

server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -264,6 +269,8 @@ public Content updateContent(Content entity, ContentData update, Owner owner,
                     }
                 }
 
+                affectedProducts.close();
```
awood: Does this need to be in a `finally`?

Ceiu: Probably doesn't absolutely need to be, but it doesn't hurt to add it.

Ceiu:
retest this please

server/src/main/java/org/candlepin/model/CandlepinQuery.java
```diff
@@ -0,0 +1,254 @@
...
+     * entire query result into memory before processing it. This method will always pass the first
+     * column of each row to the processor.
+     *
+     * @param processor
+     *  A ResultProcessor instance to use for processing each result
+     *
+     * @return
+     *  the number of rows processed and sent to the result processor
+     */
+    int scroll(ResultProcessor<T> processor);
```
kahowell: Since Java 8 uses `forEach` as the method name with the same semantics, I think we should rename `scroll` to `forEach` for consistency.

server/src/main/java/org/candlepin/model/CandlepinQuery.java
```diff
@@ -0,0 +1,254 @@
...
+     * Executes this query and iterates over the first column of the results. Other columns in
+     * each row are silently discarded.
+     * <p></p>
+     * WARNING: This method must be called from within a transaction, and the iterator must
+     * remain within the bounds of that transaction.
+     *
+     * @return
+     *  an iterator over the first column of the results
+     */
+    ResultIterator<T> iterate();
```
kahowell: Sadly, I like the `iterate` name (verb) better, but I think we should only have `iterator`.

Ceiu: The only reservation I have for keeping it around is that the return type is more explicit than the standard iterator() method returns. If we didn't need to worry about closing the iterator/cursor, it wouldn't be an issue.

server/src/main/java/org/candlepin/model/CandlepinQuery.java
Ceiu: Done.

Ceiu:
This branch is waiting on a merge of #1358 which adds pagination support and cleans up several bits of refactored or dead code.

Ceiu:
Rebasing this on top of the caching commits that made it to master highlighted a major disconnect between the caching bits and the cursor bits. There are many methods which need to be revisited and potentially rewritten to utilize both, rather than one or the other.

nguyenfilip:
ack

PR #1221, opened 2016-05-02
1306628: Added additional logging to environment deletion

Ceiu:
This is a smaller, more immediate patch to a much bigger issue involving auditing on operations that cascade to other model objects. For this PR, only the details in the BZ were addressed due to the massive scope of the underlying issue which will be fixed in another PR to come.


wottop:
ack

PR #1223, opened 2016-05-03
1332270: Ability to quickly associate entitlements with consumers

vritant:


cnsnyder:
Ran tests locally backed by postgre. All spec tests and unit tests passed. Merging.

PR #1225, opened 2016-05-04
1332635: Treat null guest lists as empty guest lists on hypervisor checkin

cnsnyder:
Guest lists reported as null are treated as if they were empty lists.

### DIFF
--- a/server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
+++ b/server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
@@ -353,7 +353,6 @@ public class HypervisorResourceTest {
         HypervisorCheckInResult result = hypervisorResource.hypervisorCheckIn(
             hostGuestMap, principal, owner.getKey(), true);
         assertNotNull(result);
-        System.out.println(result);
         assertEquals(1, result.getCreated().size());
 
         List<Consumer> created = new ArrayList<Consumer>(result.getCreated());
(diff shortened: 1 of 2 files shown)


vritant:
okay to merge if:

 - [ ] a comment is added near the if to cite the bz and some short clarification of why.
 - [ ] a unit test or spec test is added to test that tests that in the same report, one hypervisor is reported as successful, and another fails because of this reason
 - [ ]  there is plan / trello card / PR to implement the same in master, both async and sync apis.

barnabycourt:
NACK
A null value should be treated as an empty list and not an error due to https://github.com/rails/rails/issues/13766 in the katello proxy

cnsnyder:
Here is the bug tracking the fix for this issue in master: https://bugzilla.redhat.com/show_bug.cgi?id=1332637

The meaning of null has been changed in this PR. We no longer want to fail a list that is null. Instead we will treat a null guest list as if it were an empty guest list (meaning that hypervisor is considered to have no guests at the time the report is received).

server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
```diff
@@ -395,4 +395,39 @@ public void ensureEmptyGuestIdsAreIgnored() throws Exception {
...
+        when(consumerTypeCurator.lookupByLabel(eq(ConsumerTypeEnum.HYPERVISOR.getLabel())))
+            .thenReturn(hypervisorType);
+        when(idCertService.generateIdentityCert(any(Consumer.class)))
+            .thenReturn(new IdentityCertificate());
+
+        HypervisorCheckInResult result = hypervisorResource.hypervisorCheckIn(
+            hostGuestMap, principal, owner.getKey(), true);
+        assertNotNull(result);
+        assertNotNull(result.getCreated());
+        System.out.println(result);
```
vritant: :point_up_2:

cnsnyder:
@vritant @barnabycourt Updated! Mind looking it over one more time?

PR #1226, opened 2016-05-04
Batch Revoke Performance

nguyenfilip:
The batch revoke code contains unnecessary initialization of entitlements
collection. This was introduced to maintain consistency to mitigate errors of
caused by Hibernate cascading. However, since Pool.entitlements collection is
not initialized during batch revoke, the cascading is not an issue.

### DIFF
--- a/server/src/main/java/org/candlepin/model/PoolCurator.java
+++ b/server/src/main/java/org/candlepin/model/PoolCurator.java
@@ -28,7 +28,6 @@ import com.google.inject.persist.Transactional;
 import org.hibernate.Criteria;
 import org.hibernate.FetchMode;
 import org.hibernate.Filter;
-import org.hibernate.Hibernate;
 import org.hibernate.LockOptions;
 import org.hibernate.Query;
 import org.hibernate.ReplicationMode;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -682,13 +681,6 @@ public void batchDelete(List<Pool> pools) {
         for (Pool pool : pools) {
             currentSession().delete(pool);
         }
-        for (Pool pool : pools) {
```
barnabycourt: Given the comment below, what is the risk from not clearing the entitlements here?

nguyenfilip: The risk is that if pool.entitlements has been initialized before this PoolCurator.batchDelete is called, we might get ConstraintViolation exceptions on Postgre.

barnabycourt: Is there any way to check if pool.entitlements has been initialized? Worst case, please update the javadoc to include this information. Best case, we could guard against the issue and raise an exception so that testing would catch it if someone used it after entitlements have been initialized.

nguyenfilip:
@barnabycourt I think you are right, the check should go in there, will discuss this as part of #1233 because this change is incorporated in that PR

PR #1227, opened 2016-05-04
1332637: Treat null guest lists as empty guest lists

cnsnyder:
This is the forward port of the fix for bz 1332635. There is really nothing much different here about the fix than there was in candlepin version 0.9.54. The async logic did not have the same flaw as the synchronous logic did.


vritant:
Looks good except for the test failure.

barnabycourt:
1332635 should not be listed in the title as this PR is to master and will never hit the branch that BZ is for.

PR #1229, opened 2016-05-05
Changed the content set id to reduce confusion in testing

wottop:
The value '0' can equal false. Should not use as a content id.


Ceiu:
ACK

PR #1232, opened 2016-05-05
Batch List Modifying

nguyenfilip:
Retrieval of modifier products by a query instead of using iteration over
entitites.

### DIFF
--- a/server/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/server/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -21,7 +21,6 @@ import org.candlepin.common.paging.PageRequest;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 
-import org.apache.commons.collections.CollectionUtils;
 import org.hibernate.Criteria;
 import org.hibernate.ReplicationMode;
 import org.hibernate.criterion.CriteriaSpecification;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -302,39 +271,31 @@ private Criteria createModifiesDateFilteringCriteria(Set<Consumer> consumers, Da
...
+            "              JOIN inPool.product inMktProd " +
+            "              LEFT JOIN inPool.providedProducts inProvidedProd" +
+            "              WHERE eIn in (:ein) AND inConsumer = eOut.consumer AND " +
+            "                inPool.endDate >= outPool.startDate AND " +
+            "                inPool.startDate <= outPool.endDate AND" +
+            "                ( inProvidedProd.id = outModProdId OR inMktProd.id = outModProdId ) " +
+            "             )" +
+            "                       "
+            , Entitlement.class).setParameter("ein", entitlements).getResultList();
+        return new HashSet<Entitlement>(ents);
```
Ceiu: The only place we use this method is in the block above in CPM. We should probably just change the return type to Collection<Entitlement> rather than copying everything to a new set.

Also, with the changes forth-coming in my branch, it might be better to change this to return just the entitlement IDs rather than the objects (or, at least, have a version of this method which does so).

Ceiu:
This PR has been merged into the PR addressing BZ 1331793 (#1233). We can either close this PR, or rebase #1233 when this is merged.

nguyenfilip:
Closing this one, I will add the tests in a separate PR.

PR #1233, opened 2016-05-10
1331793: Improved performace of expired pool cleanup task

Ceiu:
- Added a limit to the number of pools the expired pools query will fetch
  per "block." The cleanup expired pools operation will now fetch and
  cleanup pools in blocks of 2048 (by default) rather than fetching
  every known pool and running out of memory
- Hibernate queries which make use of the IN operator no longer create
  new lists when the source collection is not a List instance
- Several redundant blocks of code were removed in favour of the updated
  in-operator handling
- Added explicit handling of iterable entitlement IDs when regenerating
  certificates; it is now performed with a bulk update rather than
  loading the entire entitlement set one-by-one
- Retrieval of modifier products by a query instead of using iteration over
  overlapping entitlements (which consumes a lot of memory)
- cleanupExpiredPools, and the query behind it, no longer pull
  pools which are derived, as they were ignored by the process
  anyway
- cleanupExpiredPools will no longer delete pools that have an
  entitlement which overrides the end date such that it ends
  after the pool's base end date
- The regeneration of certificates of modifying entitlements
  during entitlement revocation is now performed in bulk on
  the DB
- Entitlement.getDirty is now Entitlement.isDirty
- Added new tests to cover untested methods
  EntitlementCurator.markEntitlementsDirty and
  EntitlementCertificateGenerator.regenerateCertificatesByEntitlementId
- Updated CandlepinPoolManager.regenerateDirtyCertificates to no longer
  copy entitlements to a separate list to iterate

### DIFF
--- a/server/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/server/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -18,10 +18,10 @@ import org.candlepin.common.exceptions.BadRequestException;
 import org.candlepin.common.paging.Page;
 import org.candlepin.common.paging.PageRequest;
 
+import com.google.common.collect.Iterables;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 
-import org.apache.commons.collections.CollectionUtils;
 import org.hibernate.Criteria;
 import org.hibernate.ReplicationMode;
 import org.hibernate.criterion.CriteriaSpecification;
@@ -37,10 +37,14 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
+import javax.persistence.Query;
+
+
 
 /**
  * EntitlementCurator
@@ -242,6 +246,7 @@ public class EntitlementCurator extends AbstractHibernateCurator<Entitlement> {
                 }
             }
         }
+
         return entitledProductIds;
     }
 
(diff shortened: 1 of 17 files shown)


server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -682,13 +722,6 @@ public void batchDelete(List<Pool> pools) {
         for (Pool pool : pools) {
             currentSession().delete(pool);
         }
-        for (Pool pool : pools) {
-            // Maintain runtime consistency. The entitlements for the pool
-            // have been deleted on the database because delete is cascaded on
-            // Pool.entitlements relation
-            Hibernate.initialize(pool.getEntitlements());
```
nguyenfilip: After talking with @barnabycourt  I think it would be better to instead of deleting this to check whether the collection is initialized and if it is then do the clear. This is very defensive, it will shield us from possible ConstraintViolationExceptions



```java
    /**
     * Batch deletes a list of pools.
     * @param pools
     */
    public void batchDelete(List<Pool> pools) {
        for (Pool pool : pools) {
            currentSession().delete(pool);
            // Maintain runtime consistency. The entitlements for the pool
            // have been deleted on the database because delete is cascaded on
            // Pool.entitlements relation
            if (Hibernate.isInitialized(pool.getEntitlements())) {
                pool.getEntitlements().clear();
            }
        }
    }
```

barnabycourt: Could we log a warning that this method should not be used with initialized entitlement collections? If we don't log a warning a developer will never know they did something wrong.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -301,52 +276,57 @@ private Criteria createModifiesDateFilteringCriteria(Set<Consumer> consumers, Da
...
+                "                FROM Entitlement eIn" +
+                "                    JOIN eIn.consumer inConsumer" +
+                "                    JOIN eIn.pool inPool" +
+                "                    JOIN inPool.product inMktProd" +
+                "                    LEFT JOIN inPool.providedProducts inProvidedProd" +
+                "                WHERE eIn in (:ein) AND inConsumer = eOut.consumer AND" +
+                "                    inPool.endDate >= outPool.startDate AND" +
+                "                    inPool.startDate <= outPool.endDate AND" +
+                "                    (inProvidedProd.id = outModProdId OR inMktProd.id = outModProdId)" +
+                "        )";
```
mstead: Might be worth adding a description here of what this query is actually doing. Might be a little hard to follow for someone that isn't that strong with reading SQL queries.

mstead:
Looks ok to me.

PR #1234, opened 2016-05-10
Fixed periodic failures in some unit tests

Ceiu:
- Fixed an issue in the JobCuratorTest which allowed two jobs to have the
  same time for the purposes of ordering
- Fixed an issue in the ConsumerCuratorTest where the update date could
  end up identical unintentionally
- Added a tolerance of three milliseconds to the time check in
  X509CRLStreamWriterTest to account for minor slowdowns or scheduling
  issues during testing

### DIFF
--- a/server/src/main/java/org/candlepin/model/JobCurator.java
+++ b/server/src/main/java/org/candlepin/model/JobCurator.java
@@ -155,6 +155,9 @@ public class JobCurator extends AbstractHibernateCurator<JobStatus> {
     }
 
     public JobStatus getByClassAndTarget(String target, Class<? extends KingpinJob> jobClass) {
+        // FIXME:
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/model/JobCurator.java
```diff
@@ -155,6 +155,9 @@ public long findNumRunningByClassAndTarget(String target, Class<? extends Kingpi
     }
 
     public JobStatus getByClassAndTarget(String target, Class<? extends KingpinJob> jobClass) {
+        // FIXME:
```
barnabycourt: @Ceiu Is this reminder something you were planning to fix in this PR?

Ceiu: Nope. Just a note for whomever dives into the guts of the job system next.

PR #1236, opened 2016-05-11
Clean working copy before checkout

nguyenfilip:
Our script that pulls Github PR branches was using 'git pull' which may leave
untracked files and local changes in the working copy. That may result in errors
during checkouts. It is safer to fetch+reset and then clean the working copy.

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -170,7 +170,11 @@ else
     echo "Using /candlepin."
     CP_HOME="/candlepin"
     cd $CP_HOME
-    git pull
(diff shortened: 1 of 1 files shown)


nguyenfilip:
Example of problems caused by the former behavior see Jenkins job candlepin-pullrequest-unittests  build 1428

nguyenfilip:
@vritant What do you think about this (also see the failure in Jenkins job)?

Ceiu:
This should be good. It's a little nuclear, but since the images should be clean of ancillary bits anyway, this *should* have no impact on the currently unaffected branches, and it should fix the cases where our images are built in semi-outdated states.

ACK

PR #1244, opened 2016-05-15
Check for scheduled compliance jobs in list consumers API

vritant:


server/src/main/java/org/candlepin/model/JobCurator.java
```diff
@@ -72,6 +73,22 @@ public void cancelNoReturn(String jobId) {
...
 
+    public int cancelIfNotRunning(List<String> jobIds) {
+        if (CollectionUtils.isEmpty(jobIds)) {
+            return 0;
+        }
+        String hql = "update JobStatus j " +
+            "set j.state = :canceled " +
+            "where j.id in (:jobids) and j.state in (:created, :pending)";
+        Query query = this.currentSession().createQuery(hql)
+            .setParameterList("jobids", jobIds)
```
Ceiu: We'll need to manually convert this to do a safe list, or not allow large collections.

PR #1246, opened 2016-05-16
1335576: Ensure unlimited host dependent pool has finite virt-limit quantity

wottop:
Ensure that unlimited master pool with finite virt_limit means that each host entitlement spawns a pool with finite quantity.


Ceiu:
ACK

PR #1247, opened 2016-05-16
1336054: Overconsumption Fix

nguyenfilip:
Because of repeatable read isolation semantics, we sometimes
got stale data even after locking an entity. This caused
overconsumption.


awood:
Looks good to me.  I will merge once all the tests pass.  :clap: :tada: for figuring this one out.

PR #1250, opened 2016-05-17
Updated query backing the PoolCurator.listExpiredPools method

Ceiu:
- PoolCurator.findExpiredPools now uses left joins and null checks
  when comparing entitlements and attributes, fixing an issue where
  pools without entitlements or attributes would not be found
- CandlepinPoolManager.cleanupExpiredPools now uses a transaction for
  each block of pools rather than the entire cleanup operation
- Added AbstractHibernateCurator.evictAll for performing bulk eviction
  on entity collections

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -499,33 +499,53 @@ public class CandlepinPoolManager implements PoolManager {
         regenerateCertificatesByEntIds(updatedMasterPools, lazy);
     }
 
-    @Transactional
+    /**
+     * Deletes all known expired pools. The deletion of expired pools also triggers entitlement
+     * revocation and consumer compliance recalculation.
+     * <p></p>
+     * This method will delete pools in blocks, using a new transaction for each block unless a
+     * transaction was already started before this method is called.
+     */
     public void cleanupExpiredPools() {
         int count = 0;
         boolean loop = false;
 
+        log.debug("Beginning cleanup expired pools job");
+
         do {
-            List<Pool> pools = poolCurator.listExpiredPools(PoolCurator.EXPIRED_POOL_BLOCK_SIZE);
-            count += pools.size();
+            // This call is run within a new transaction if we're not already in a transaction
+            int blockSize = this.cleanupExpiredPoolsImpl();
+            count += blockSize;
 
-            // TODO: Do we still need this? This bit is incredibly slow.
-            for (Iterator<Pool> pi = pools.iterator(); pi.hasNext();) {
-                Pool pool = pi.next();
-                log.info("Cleaning up expired pool: {} (expired: {})", pool.getId(), pool.getEndDate());
-            }
-
-            // Delete the block of pools
-            this.deletePools(pools);
-
-            // Cleanup memory to try to avoid flooding the heap
-            this.poolCurator.clear();
-
-            loop = pools.size() >= PoolCurator.EXPIRED_POOL_BLOCK_SIZE;
+            loop = blockSize >= PoolCurator.EXPIRED_POOL_BLOCK_SIZE;
         } while (loop);
 
         if (count > 0) {
             log.info("Cleaned up {} expired pools", count);
         }
+    }
+
+    /**
+     * Performs the cleanup of a block of expired pools.
+     *
+     * @return
+     *  the number of expired pools deleted as a result of this method
+     */
+    @Transactional
+    protected int cleanupExpiredPoolsImpl() {
+        List<Pool> pools = poolCurator.listExpiredPools(PoolCurator.EXPIRED_POOL_BLOCK_SIZE);
+
+        for (Pool pool : pools) {
+            log.info("Cleaning up expired pool: {} (expired: {})", pool.getId(), pool.getEndDate());
+        }
+
+        // Delete the block of pools & flush the results to tell Hibernate to evict the objects
+        // (we hope). Even if it doesn't, and even if the transaction completion is going to
+        // flush the objects anyway, it should not hurt and is an explicit call.
+        this.deletePools(pools);
+        this.poolCurator.flush();
+
+        return pools.size();
     }
 
     private boolean isExpired(Subscription subscription) {
@@ -1631,11 +1651,13 @@ public class CandlepinPoolManager implements PoolManager {
             //to have access to those products later in this method.
             ent.getPool().getProvidedProducts().size();
             Pool pool = ent.getPool();
-            pool.setConsumed(pool.getConsumed() - ent.getQuantity());
+            int entQuantity = ent.getQuantity() != null ? ent.getQuantity() : 0;
+
+            pool.setConsumed(pool.getConsumed() - entQuantity);
             pool.getEntitlements().remove(ent);
             Consumer consumer = ent.getConsumer();
             if (consumer.getType().isManifest()) {
-                pool.setExported(pool.getExported() - ent.getQuantity());
+                pool.setExported(pool.getExported() - entQuantity);
             }
         }
 
@@ -1839,7 +1861,7 @@ public class CandlepinPoolManager implements PoolManager {
 
         for (Pool p : pools) {
             if (log.isDebugEnabled()) {
-                log.debug("Deletion of pool {} will cause revocation of the following entitlements: {}.",
+                log.debug("Deletion of pool {} will revoke the following entitlements: {}",
                     p.getId(), getEntIds(p.getEntitlements()));
             }
             entitlementsToRevoke.addAll(p.getEntitlements());
--- a/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
+++ b/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
@@ -568,6 +568,20 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
         return object;
     }
 
+    /**
+     * Evicts all of the given entities from the level-one cache.
+     *
+     * @param collection
+     *  An iterable collection of entities to evict from the session
+     */
+    public void evictAll(Iterable<E> collection) {
+        Session session = this.currentSession();
+
+        for (E entity : collection) {
+            session.evict(entity);
+        }
+    }
+
     public List<E> takeSubList(PageRequest pageRequest, List<E> results) {
         int fromIndex = (pageRequest.getPage() - 1) * pageRequest.getPerPage();
         if (fromIndex >= results.size()) {
--- a/server/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/server/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -43,6 +43,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import javax.persistence.EntityManager;
 import javax.persistence.Query;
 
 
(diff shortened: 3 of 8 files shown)


vritant:
un-assigning myself cause I couldnt find time to finish testing this, and ill be out for a few days.
I do have one investigation to suggest. what is the current deployment's database max_allowed_packet size? is the current batch size appropriate in accordance with that?
I suggest that cause the job continues to fail for me locally for that reason.

vritant:
@Ceiu rebase please?

mstead:
retest this please

nguyenfilip:
This PR should fix https://bugzilla.redhat.com/show_bug.cgi?id=1344765

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -482,6 +482,20 @@ public E evict(E object) {
         return object;
     }
 
+    /**
```
vritant: im assuming you meant to leave this method in in case its used in the future.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -363,16 +363,59 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
+     */
     @Transactional
     public void delete(Entitlement entity) {
         Entitlement toDelete = find(entity.getId());
-        log.debug("Deleting entitlement: {}", toDelete);
-        log.debug("certs.size = {}", toDelete.getCertificates().size());
 
-        for (EntitlementCertificate cert : toDelete.getCertificates()) {
-            currentSession().delete(cert);
+        if (toDelete != null) {
```
vritant: curious, why a null check here? Im concerned if we are masking a problem. why would we call this method with a non existent entitlement?

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -363,16 +363,59 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
+            this.deleteImpl(toDelete);
+
+            // Maintain runtime consistency.
+            entity.getCertificates().clear();
+            entity.getConsumer().getEntitlements().remove(entity);
+            entity.getPool().getEntitlements().remove(entity);
```
vritant: This needs to have a similar check like #1272

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -363,16 +363,59 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
+     *  The collection of entitlement entities to delete
+     */
+    public void batchDelete(Collection<Entitlement> entitlements) {
+        for (Entitlement entitlement : entitlements) {
+            this.deleteImpl(entitlement);
+
+            // Maintain runtime consistency.
+            entitlement.getCertificates().clear();
+            entitlement.getConsumer().getEntitlements().remove(entitlement);
+            entitlement.getPool().getEntitlements().remove(entitlement);
```
vritant: NACK. this undoes #1272

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -230,15 +232,22 @@ public Pool find(Serializable id) {
...
-        Criteria criteria = this.createSecureCriteria("pool")
-            .createAlias("pool.entitlements", "entitlement")
-            .createAlias("pool.attributes", "attributes")
-            .add(Restrictions.lt("pool.endDate", now))
-            .add(Restrictions.ne("attributes.name", "derived_pool"))
-            .add(Restrictions.or(
-                Restrictions.isNull("entitlement.endDateOverride"),
-                Restrictions.lt("entitlement.endDateOverride", now)
-            ));
+        DetachedCriteria attribCheck = DetachedCriteria.forClass(Pool.class, "attribPool")
```
vritant: :+1:

server/src/main/java/org/candlepin/model/EntitlementCurator.java
Ceiu: ...? That's for 0.9.49. Also, significantly newer. :P

vritant:
@Ceiu in addition to addressing the above comments, please:

- [ ] add 1344765 to the commit message
- [ ] add a unit test that verifies a pool is purged even if it does not have any entitlement ( which was the case in the bug )

server/src/main/java/org/candlepin/model/EntitlementCurator.java
vritant: @Ceiu similar fix was made for master  in #1262 , which this undoes.
tbh I should have asked @nguyenfilip  to add a comment here so a future dev does not delete it inadvertently. could you add it in this PR please? if not, I can create a separate PR .

server/src/main/java/org/candlepin/model/EntitlementCurator.java
Ceiu: The entitlement entity given to the method can still be not-null, but the resolution fails. If there's nothing to delete, then we fail out with null pointer exceptions instead of doing nothing.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
Ceiu: Specific line, please

Ceiu:
There's already a unit test for verifying pools without entitlements are deleted. See: https://github.com/candlepin/candlepin/pull/1250/files#diff-410bec408e3b2222674ea8c61351d35bR735

server/src/main/java/org/candlepin/model/EntitlementCurator.java
vritant: if (Hibernate.isInitialized(ent.getPool().getEntitlements())) {
                ent.getPool().getEntitlements().remove(ent);
            }

vritant:
@Ceiu 

2016-06-16 13:43:47,340 [thread=QuartzScheduler_Worker-2] [job=ExpiredPoolsJob-18667bdc-7165-4cb3-8a72-16aff9bfdb30, org=] ERROR org.candlepin.pinsetter.core.PinsetterJobListener - Job [ExpiredPoolsJob-18667bdc-7165-4cb3-8a72-16aff9bfdb30] failed.
org.quartz.JobExecutionException: javax.persistence.TransactionRequiredException: no transaction is in progress
	at org.candlepin.pinsetter.tasks.KingpinJob.refireCheck(KingpinJob.java:134) ~[KingpinJob.class:na]
	at org.candlepin.pinsetter.tasks.KingpinJob.execute(KingpinJob.java:104) ~[KingpinJob.class:na]
	at org.quartz.core.JobRunShell.run(JobRunShell.java:202) ~[quartz-2.2.1.jar:na]
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573) [quartz-2.2.1.jar:na]
Caused by: javax.persistence.TransactionRequiredException: no transaction is in progress
	at org.hibernate.ejb.AbstractEntityManagerImpl.flush(AbstractEntityManagerImpl.java:993) ~[hibernate-entitymanager-4.2.5.Final.jar:4.2.5.Final]
	at org.candlepin.model.AbstractHibernateCurator.flush(AbstractHibernateCurator.java:395) ~[AbstractHibernateCurator.class:na]
	at org.candlepin.controller.CandlepinPoolManager.cleanupExpiredPoolsImpl(CandlepinPoolManager.java:621) ~[CandlepinPoolManager.class:na]
	at org.candlepin.controller.CandlepinPoolManager.cleanupExpiredPools(CandlepinPoolManager.java:593) ~[CandlepinPoolManager.class:na]
	at org.candlepin.pinsetter.tasks.ExpiredPoolsJob.toExecute(ExpiredPoolsJob.java:49) ~[ExpiredPoolsJob.class:na]
	at org.candlepin.pinsetter.tasks.KingpinJob.execute(KingpinJob.java:84) ~[KingpinJob.class:na]
	... 2 common frames omitted

vritant:
Im currently un assigning myself from this review cause I think I am taking too long, and dont feel comfortable merging this in yet.
I should note that none of the issues I'm raising appear to be because of changes in this PR, this PR is simply focused on using joins.
That being said, here are my concerns:

1. consistent OOM errors:
when we serialize the pools to publish events, we run out of memory and almost all of it is consumed by pool.getProvidedProductDtos - im currently investigating more.
2. LockWaitTimeOut - I suspect this is not related to this job. I got this error only once and most likely it was cause the past transaction of the same job, but an earlier failed instance, did not relinquish the pool locks.
3. transaction failure from the above comment: This is the reason im not comfortable yet. This error happens at the end of the job, when we flush, but I have not yet been able to reach that point in execution because of the OOM errors that happen before that point.
4. what is the appropriate EXPIRED_POOL_BLOCK_SIZE ? I wanted to test that the current value of 2048 pools does not exceed the max_allowed_packet limit .  I have seen this failure locally , and would like to test using the max_allowed_packet size that is actually used today.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -575,35 +575,55 @@ void refreshPoolsForMasterPool(Pool pool, boolean updateStackDerived, boolean la
...
         }
     }
 
+    /**
+     * Performs the cleanup of a block of expired pools.
+     *
+     * @return
+     *  the number of expired pools deleted as a result of this method
+     */
+    @Transactional
```
nguyenfilip: Since cleanupExpiredPoolsImpl is private, the Transactional attribute doesn't have any effect. Thats why @vritant  saw TransactionRequiredException

Either make the method public or dont flush. There is actually no need to flush since deletePools is gonna flush.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -575,35 +575,55 @@ void refreshPoolsForMasterPool(Pool pool, boolean updateStackDerived, boolean la
...
+    private int cleanupExpiredPoolsImpl() {
+        List<Pool> pools = poolCurator.listExpiredPools(PoolCurator.EXPIRED_POOL_BLOCK_SIZE);
+
+        for (Pool pool : pools) {
+            log.info("Cleaning up expired pool: {} (expired: {})", pool.getId(), pool.getEndDate());
+        }
+
+        // Delete the block of pools & flush the results to tell Hibernate to evict the objects
```
nguyenfilip: Flush won't evict anything. If you need to evict you will need to call EntityManager.clear() OR evict pools individually.

nguyenfilip:
@vritant You were correct. The code as is won't work (will always rollback transaction - see my comment)

@Ceiu Given the problem with Transactional annotation on private method and flush throwing exception, I feel like we need some spec tests around ExpiredPoolsJob. Can't we just add a hidden resource method that invokes ExpiredPoolsJob for testing and write a Spec test around that? I can do that, just let me know if you think its a good idea.

nguyenfilip:
OOM blocks me to test this against IT's DB Dump as well.

I will go ahead and write a spec test for this, to verify it works functionally on small dataset. However to ack this passes against the DB dump (which means solving the OOM).

@vritant I know you work on the OOM, but I can take that over if you agree (since it is related to this PR). Let me know.

server/src/test/java/org/candlepin/controller/PoolManagerFunctionalTest.java
```diff
@@ -746,5 +731,127 @@ public void testBatchBindZeroQuantity() throws EntitlementRefusedException {
...
+        Date expiredEnd = new Date(ct - 3600000);
+
+        Owner owner = this.createOwner();
+        Product product1 = this.createProduct(owner, "test-product-1", "Test Product 1");
+        Product product2 = this.createProduct(owner, "test-product-2", "Test Product 2");
+        Pool activePool = this.createPool(owner, product1, 1L, activeStart, activeEnd);
+        Pool expiredPool = this.createPool(owner, product2, 1L, expiredStart, expiredEnd);
+
+        try {
+            this.beginTransaction();
```
nguyenfilip: Why drive transactions yourself in the test?

nguyenfilip:
@Ceiu When I started on BZ 1344765 (and before I remembered that you have this PR up) I wrote a few unit tests for listExpiredPools specifically. If you want them 
```
git checkout ace95927cca976bf4b852cfc276068b9e0176f7b -- server/src/test/java/org/candlepin/model/PoolCuratorExpiredPoolsTest.java
```

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: My opinion is that you should call clear() after each 'block' of expired pools

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
mstead: > Either make the method public or dont flush. There is actually no need to flush since deletePools is gonna flush.

For what it is worth, the method can be protected as well IIRC.

vritant: I changed it to protected but left the flush in, its not called in deletePools and does not hurt.


@nguyenfilip question though. given that we just deleted the objects. will hibernate evict these objects from the cache upon flush?

vritant:
@nguyenfilip  w.r.t. ace95927cca976bf4b852cfc276068b9e0176f7b lets add the test in a separate PR?

also, I was able to successfully test that changing to lazy fetch prevents the OOM error.
I was also able to verify that the batch size of 2k works given current max_allowed_packet used.
per our discussion earlier, we are waiting for @Ceiu 's DTO work to go into this, so this marking this PR as in progress

server/src/test/java/org/candlepin/controller/PoolManagerFunctionalTest.java
vritant: @nguyenfilip I suspect this was because @Transaction was being ignored, cause the method was private. this should have been a red flag to me. any who, fixed now.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
nguyenfilip: Ah, yes you are right, since they have been deleted they will be detached.

nguyenfilip:
@vritant yeah adding ace9592 in separate PR is fine.

vritant:
@nguyenfilip please look at #1282 . I'm done testing this, this is good to merge on my end.

PR #1251, opened 2016-05-17
addOwner in Product and Content now replace equal owners on call

Ceiu:
- .addOwner in Product and Content will now update an
  owner-product/content relation when called with an owner that is
  already mapped to the given product or content


Ceiu:
After further reflection, I've decided I don't like this fix. It conflicts with the verbage of addOwner (should add, not replace), and doesn't provide a complete fix if a product or content is provided with more than just a single, transient owner.

I'll have a new PR shortly with a more robust fix.

PR #1252, opened 2016-05-18
CandlepinPoolManager now removes extraneous product owners during refresh

Ceiu:
- During pool refresh, any owners other than the owner for which pools
  are being refreshed are removed from incoming products and content.
  This should address an issue where we could receive transient owners
  on an entity, and currently does not impact equality checks.


mstead:
ACK

PR #1253, opened 2016-05-18
1322156: Update derived pool quantity in hosted environment

wottop:
When virt_limit changes, a non host_limited derived pool needs
to be updated to reflect it. The host_limited scenario was already
accounted for.

### DIFF
--- a/server/spec/candlepin_scenarios.rb
+++ b/server/spec/candlepin_scenarios.rb
@@ -485,6 +485,29 @@ class StandardExporter < Exporter
         :attributes => { :arch => "x86_64", 'stacking_id' => 'stack-dc' }
     })
 
+    @products[:product_virt_limit] = create_product('product_virt_limit', random_string(), {
+        :attributes => { :arch => "x86_64", :virt_limit => "7"}
+    })
+    @products[:product_virt_limit_2] = create_product('product_virt_limit_2', random_string(), {
+        :attributes => { :arch => "x86_64", :virt_limit => "7"}
+    })
+    @products[:derived_product_virt_limit] = create_product('derived_product_virt_limit', random_string())
+
+    @products[:product_virt_limit_stack] = create_product('product_virt_limit_stack', random_string(), {
+        :attributes => { :arch => "x86_64", :virt_limit => "10", :stacking_id => "stack-44"}
+    })
+
+    @products[:product_virt_limit_stack_2] = create_product('product_virt_limit_stack_2', random_string(), {
+        :attributes => { :arch => "x86_64", :virt_limit => "10", :stacking_id => "stack-44"}
+    })
+
+    @products[:product_virt_limit_hosted] = create_product('product_virt_limit_hosted', random_string(), {
+        :attributes => { :arch => "x86_64", :virt_limit => "15", :host_limited => "true", :stacking_id => "stack-25"}
+    })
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -148,32 +148,48 @@ private Pool createBonusPool(Pool masterPool, List<Pool> existingPools) {
...
-            virtAttributes.put("pool_derived", "true");
-            virtAttributes.put("physical_only", "false");
-            if (hostLimited || config.getBoolean(ConfigProperties.STANDALONE)) {
-                virtAttributes.put("unmapped_guests_only", "true");
-            }
-            // Make sure the virt pool does not have a virt_limit,
-            // otherwise this will recurse infinitely
-            virtAttributes.put("virt_limit", "0");
+        if (attributes.containsKey("virt_limit")  && virtQuantity != null) {
+            if (!hasBonusPool(existingPools)) {
```
vritant: It is better to :

```java
 Pool  bonusPool = getBonusPool(existingPools);

 if(bonusPool == null) {
    //create bonus pool
 } else {
   //update pool if needed
 }
```
this way we:

- iterate through pools only once
- get to clean up the hasBonusPool method

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -148,32 +148,48 @@ private Pool createBonusPool(Pool masterPool, List<Pool> existingPools) {
...
+                Pool bonus = getBonusPool(existingPools);
+                long virtLong = -1L;
+                try {
+                    virtLong = Long.parseLong(virtQuantity);
+                }
+                catch (NumberFormatException pe) {
+                    // it is unlimited, leave it at -1L
+                }
+                if (virtLong != bonus.getQuantity()) {
+                    bonus.setQuantity(virtLong);
```
vritant: don't need this line since the manager.setPoolQuantity covers this bit.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -148,32 +148,48 @@ private Pool createBonusPool(Pool masterPool, List<Pool> existingPools) {
...
+                    poolManager.setPoolQuantity(bonus, virtLong);
+                    return bonus;
```
vritant: first, since this method now updates pool as well, we should rename it to `createOrUpdateBonusPool` from `createBonusPool`

also, why return the bonus pool if we are updating it? the `poolManager.setPoolQuantity` method merges the change, if that is the concern.

returning the pool here calls `poolManager.createPool` , which:
* logs that the update as "new pool"
* calls poolCurator.create(pool) ( which would probably cause an error )
* emit a pool create event instead of pool update event

I think we should simply not return in the else case and simply emit an update event, thoughts? am I missing something?

vritant:
@wottop may I request a spec test? there is an example in `import_warning_spec.rb` where we can extract, manipulate and re-zip a manifest, that should help a lot.
if yes, please consider moving the common bits to `candlepin_scenarios` so other spec files can share that logic, since this is turning out to be a common use case now..

vritant:
@wottop , I think this needs to be backported as well.

vritant:
remove default mark

wottop:
Closing until we determine what the fate of changing SKU attributes after deployment.

PR #1254, opened 2016-05-20
Tests of Batch List Modifying

nguyenfilip:
Functional tests for batch list modifying query.


Ceiu:
Why did you opt to create a new class for the modifier data?

nguyenfilip:
@Ceiu Just because PoolManagerFunctionalTest is big and it was hard scrolling and adding things there. But if you think its better to have everything there I can just move it

Ceiu:
That's what I figured. Either is fine -- the only thing we're saving by having it in one class is the curator declarations; but it's test code so it's not critical that we save nanoseconds here and there.

ACK

PR #1257, opened 2016-05-25
1338908: The logging message for auto-attach should be more descriptive

wottop:


nguyenfilip:
Ack.

```
2016-05-31 12:39:22,874 [thread=http-bio-8443-exec-4] [req=259c223c-86e7-4871-9900-0d73627de050, org=admin] INFO  org.candlepin.policy.     js.autobind.AutobindRules - Consumer is compliant and does not require more entitlements.
```

PR #1259, opened 2016-05-26
1338950: Lock owner during an import

vritant:


vritant:
closing against #1274

PR #1262, opened 2016-05-27
1337906: Stack derived pool cleanup

nguyenfilip:
In some code paths, it may be possible that deletion of entitlements
might not succeed (DELETE statement is not issued to the database).
The underlying issue here is with collection Pool.entitlements. Because
this collection has cascade set to PERSIST, one must be very careful
when deleting entitlement to also remove it from the collection.
Otherwise, Hibernate will un-schedule the deletion.


vritant:

vritant:
@nguyenfilip why test the import scenario also with this PR?
asking cause I thought the spec test in other PRs was sufficient?

nguyenfilip:
@vritant The problem manifested itself only in refresh pools code path. It seemed to me that import manifest is the best way to realistically trigger refresh pools in CP 2.0.

PR #1264, opened 2016-05-31
Case insensitive username consumers permission

jturel:
I don't have a working Java environment to see if my test passes but nevertheless I'm not sure there was a reason to not have this comparison use equalsIgnoreCase.

My expectation is that I will get a 200 rather than 4xx for issuing this request where the consumer in question has the 'username' attribute of (for example) 'jTuReL', and I pass 'jturel' as the value of the cp-user header.

`curl -H 'cp-user: jturel' -H 'cp-lookup-permissions: true' "$CP_DEV/owners/consumers/aaaaaaa-9a0b-4937-ad72-658fc2cbb98d"`

### DIFF
--- a/server/src/main/java/org/candlepin/auth/permissions/UsernameConsumersPermission.java
+++ b/server/src/main/java/org/candlepin/auth/permissions/UsernameConsumersPermission.java
@@ -53,7 +53,7 @@ public class UsernameConsumersPermission implements Permission, Serializable {
         // Implied full access to the relevant Consumers:
         if (target.getClass().equals(Consumer.class)) {
             return ((Consumer) target).getOwner().getKey().equals(owner.getKey()) &&
-                ((Consumer) target).getUsername().equals(user.getUsername()) &&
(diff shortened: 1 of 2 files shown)


barnabycourt:
@jturel Is the intent here to change Candlepin usernames being cases sensitive to making them case insensitive?

jturel:
@barnabycourt No, I do not wish to affect the entire application - just this isolated comparison in the UsernameConsumersPermission. Since you asked the question it makes me wonder if Candlepin was designed for "JTUREL" and "jturel" to be two distinct users. Is that the case?

nguyenfilip:
@jturel I think to make this work, we would need a bit more involved fix. The username is also used in method getCriteriaRestrictions which is there to determine what the user is allowed to retrieve from the database. It seems that Candlepin has been designed to exact match the username.

I can investigate further, but first, why do you need this to be case insensitive?

nguyenfilip:
jturel said: "we are gonna address this from the Adapter side, due to the fact that this change is basically incompatible with candlepin in that the design expects case-sensitive usernames"

PR #1267, opened 2016-06-06
1338846: Fixing property filter race condition

nguyenfilip:
Making DynamicPropertyFilter thread safe and adding a few unit tests for
it since they were missing. Because this class wasn't thread safe a user
could hit serialization server side errors because requests were
colliding on the List that is being modified inside the filter.

### DIFF
--- a/common/src/main/java/org/candlepin/common/jackson/DynamicPropertyFilter.java
+++ b/common/src/main/java/org/candlepin/common/jackson/DynamicPropertyFilter.java
@@ -32,27 +32,25 @@ import java.util.List;
  */
 public class DynamicPropertyFilter extends CheckableBeanPropertyFilter {
 
-    private List<String> path = new ArrayList<String>(10);
-
     public boolean isSerializable(Object obj, JsonGenerator jsonGenerator,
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/candlepin/common/jackson/DynamicPropertyFilter.java
```diff
@@ -32,27 +32,26 @@
...
     public boolean isSerializable(Object obj, JsonGenerator jsonGenerator,
         SerializerProvider serializerProvider, PropertyWriter writer) {
 
         DynamicFilterData filterData = ResteasyProviderFactory.getContextData(DynamicFilterData.class);
 
         if (filterData != null) {
-            this.path.clear();
-            this.path.add(0, writer.getName());
+            List<String> path = new ArrayList<String>(10);
+            path.clear();
```
vritant: clearing an empty list is not necessary.

nguyenfilip: Fixed

PR #1268, opened 2016-06-08
1330245: delete duplicate su from logrotate conf

vritant:


vritant:
note this is related to 1344083, in addition to the above bug, the the PR was written and merged in before the BZ could be created.

PR #1270, opened 2016-06-09
1344469: logrotate permission denied error

vritant:
This is the forward port for #1269


barnabycourt:
@vritant have you verified that the permissions still work properly after the first rotation and the files being created are things like candlepin.log-2016-06-10.gz?

vritant:
@barnabycourt http://paste.fedoraproject.org/379005/91469014/

PR #1273, opened 2016-06-14
1345918: Unique product content violation

nguyenfilip:
Upon concurrent ProductResource.removeContent requests we were getting
unique constraint violation errors for table "cp_product_content_pkey".

The reason for this is that productContent collection is a
ElementCollection and when such collection is declared as List, then
Hibernate has no way of tracking changes to this collection and then
updating database correctly (List may contain duplicates). Hibernate's
strategy in this case is to (upon every merge) delete everything from
the collection table and then insert in memory entities from that
collection back. In concurrent situations this resulted same rows
being inserted duplicitly.

Besides this solution it's also possible to use OrderColumn annotation,
but that would require schema changes.

### DIFF
--- a/server/src/main/java/org/candlepin/model/Product.java
+++ b/server/src/main/java/org/candlepin/model/Product.java
@@ -117,7 +117,7 @@ public class Product extends AbstractHibernateObject implements SharedEntity, Li
     @CollectionTable(name = "cp2_product_content", joinColumns = @JoinColumn(name = "product_uuid"))
     @Column(name = "element")
     @LazyCollection(LazyCollectionOption.EXTRA) // allows .size() without loading all data
-    private List<ProductContent> productContent;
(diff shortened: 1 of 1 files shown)


barnabycourt:

barnabycourt:
@nguyenfilip looks like there is still one rspec failure on mysql:
```
Failures:

  1) Modifier Entitlement is regenerated when modified product subscription disappears
     Failure/Error: content_repo_type(modifier_cert, @modifier_content_1.id).should be_nil

       expected: nil
            got: "yum"
     # ./server/spec/modifier_spec.rb:159:in `block (2 levels) in <top (required)>'
```

vritant:
@barnabycourt that failure is unrelated. I added a [trello card](https://trello.com/c/BFNh8hyi/92-non-deterministic-test-failure) for it!

PR #1274, opened 2016-06-14
1338950: prevent successful concurrent imports

vritant:
in 0.9.54, there is a rare chance ( not reproducable ) that two concurrent reports can result in two records in the cp_export_metadata table for the same "type" . Given the current code, this prevents any future manifest activity ( delete / import / refresh ) because our query expects there to be a unique owner and type  ( user / system ) combination for every record.
we are adding a constraint that prevents this situation in the reported branch. in candlepin 2.0, we shall lock the owner and more.
more details in the bz!

### DIFF
--- a/server/src/main/resources/db/changelog/20160614113932-add-owner-and-type-constraint-export-metadata.xml
+++ b/server/src/main/resources/db/changelog/20160614113932-add-owner-and-type-constraint-export-metadata.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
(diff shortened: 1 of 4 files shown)


vritant:
@candlepin-pull-request-bot re-test this please

mstead:
Would be nice for commit message / PR description to have some context as to what the change does.

vritant:
@mstead done

Ceiu:
conditional ack:
the import functionality as a whole needs to be revisited in the future, as there are other issues and dead code lurking about.

vritant:
@Ceiu created https://trello.com/c/tztzRYLg/315-fix-import-metadata-types

PR #1276, opened 2016-06-16
1345765 - Concurrent content demotions cause 400

nguyenfilip:
Under concurrent demotion requests, we were getting StaleStateException
from hibernate indicating that a request is trying to delete already
deleted instance. Also, the demotion of content was using one
transaction per content demotion, so request could demote some content,
then throw exception and never get into the part of the code that is
responsible for regeneration of environment certs.

In this commit I introduce RdbmsExceptionTraslator that can help to
determine a reason for a Hibernate Exception. Also I use batch
operation for database operation (demotion of content) to mitigate a
possibility of partial demotion.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/EnvironmentResource.java
+++ b/server/src/main/java/org/candlepin/resource/EnvironmentResource.java
@@ -32,6 +32,7 @@ import org.candlepin.model.EnvironmentContent;
 import org.candlepin.model.EnvironmentContentCurator;
 import org.candlepin.model.EnvironmentCurator;
 import org.candlepin.pinsetter.tasks.RegenEnvEntitlementCertsJob;
+import org.candlepin.util.RdbmsExceptionTranslator;
 import org.candlepin.util.Util;
 
 import com.google.inject.Inject;
@@ -43,12 +44,14 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xnap.commons.i18n.I18n;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import javax.persistence.RollbackException;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -284,9 +290,20 @@ public JobDetail demoteContent(
             demotedContent.put(contentId, envContent);
         }
 
-        // We've validated everything here, so we're clear to start deleting...
-        for (EnvironmentContent envContent : demotedContent.values()) {
-            this.envContentCurator.delete(envContent);
+        try {
+            envContentCurator.bulkDelete(new ArrayList<EnvironmentContent>(demotedContent.values()));
```
nguyenfilip: bulkDelete deletes all the content in one transaction. The original code starts and commits a transaction for every delete. Thats a significant difference

server/src/main/java/org/candlepin/util/RdbmsExceptionTranslator.java
```diff
@@ -0,0 +1,57 @@
...
+ * concurrent requests) than to utilize pessimistic locking.
+ *
+ * @author fnguyen
+ */
+public class RdbmsExceptionTranslator {
+    private static Logger log = LoggerFactory.getLogger(RdbmsExceptionTranslator.class);
+
+    /**
+     * Decides if the sqlException was thrown because of the deleted record
+     * not being in the database anymore
```
nguyenfilip: Yeah correct. Will rewrite the javadoc now

nguyenfilip:
This PR currently returns 200 in case of duplicit demotion. But after second thought I think this is not a good idea. The current code throws an error back by design.

nguyenfilip:
@mstead I am sticking with the original plan of just returning consistent error to the client and making sure no partial demotions are possible. I will test this a bit more before removing 'in-progress' to make sure that Transactional works as intended. You made me a bit nervous showing that the original fix didn't have that!

nguyenfilip:
@mstead tested this by using breakpoints and concurrently issuing demotion of the content:

```
curl -H "Content-type: application/json" -k -u admin:admin -X DELETE 'https://localhost:8443/candlepin/environments/eid/content?content=13' &
curl -H "Content-type: application/json" -k -u admin:admin -X DELETE 'https://localhost:8443/candlepin/environments/eid/content?content=10&content=13&content=11' &
```

I put breakpoint for both deletes to line 

```
envContentCurator.bulkDeleteTransactional
```

Then I let the first DELETE to go through. Then I observed the second delete, it tried to demote all the content in one transaction. 

Wrt implementation I added 

```
    @Transactional
    public void bulkDeleteTransactional(List<E> entities) {
```

I didn't want to distrub the current (nontransactional) batch method that is already used somewhere else.

PR #1278, opened 2016-06-21
Product Versioning and DTO Transition

Ceiu:
- Updated ProductManager and ContentManager such that CRUD operations
  on products and contents performed by the managers no longer require
  clones of the entities. Callers of these operations have been
  updated accordingly.
- Updated the ProductManager tests to use an actual database rather
  than mocking database interactions
- Fixed a bug in Product.hashCode which was causing hashes to be
  generated incorrectly
- Added new utility methods for entity creation to DatabaseTestFixture
- A few typo and stylistic corrections
- Updated ProductManager.removeProductContent to create a copy of a
  product before removing content from it as a temporary workaround
  for an issue with Hibernate's management of element collections
- Updated ProductManager tests to not use reference checks when
  verifying whether or not the correct object is being returned
- Added the new CandlepinCriteria class to be used with the new fluent
  query abstractions
- Product.owners and all its accessors and mutators have been removed.
  Owner-product mapping should now be managed with the
  OwnerProductCurator.
- Added the new OwnerProductCurator for managing owner-product
  mappings
- Reverted usage of a second session during product update
- Product.productContent is now mapped directly as a M:M relationship
  rather than managed automatically with Hibernate's EntityCollection
  relationship
- ProductContent is no longer an embedded object
- Usage of CascadeType.DELETE_ORPHAN has been removed from Product
  and, instead, replaced with JPA's removeOrphans option on the
  join annotation
- Several classes updated accordingly to work with the above changes
- Many methods from ProductCurator and OwnerCurator have moved to
  OwnerProductCurator, as it makes the most sense there
- Updated several tests and test util to use the new OwnerProductCurator
- Added the addContentToProduct method for, as the name suggests, adding
  content to a product in a way that won't violate product or content
  versioning and without impacting every owner using the product
- Removed and deprecated some content managing methods on the Product
  entity itself, as these may cause hidden errors when modifying shared
  products
- Removed some unnecessary @Transactional annotations
- Reverted the change to ProductCurator which converted it to a full
  entity from an embedded object, as this ended up having no positive
  impact and was creating new, unnecessary management work
- Removed the owner parameter from the constructor of Product and Content
- Content.owners and all its accessors and mutators have been removed.
  Owner-content mapping should now be managed with the 
  OwnerContentCurator.

===

Thinks to look for while reviewing:
- Consistency in the new additions. I'm pretty sure I kept them consistent, but they changed a couple times to deal with problems that kept coming up. 
- Potential bugs introduced by the transition away from externally exposed entity objects
- Consistency with adapters either using the new DTOs. They should all be using either the new DTOs or the old model objects -- never a combination of the two. Note that not all entities have been replaced with DTOs. This is largely due to time constraints and should be done in the future (along with a proper hydrator/populator class for converting from DTOs back to entities).


server/src/test/java/org/candlepin/controller/EntitlerTest.java
```diff
@@ -28,15 +28,17 @@
...
 import org.candlepin.model.Entitlement;
 import org.candlepin.model.EntitlementCurator;
 import org.candlepin.model.Owner;
+import org.candlepin.model.OwnerProductCurator;
 import org.candlepin.model.Pool;
 import org.candlepin.model.PoolAttribute;
 import org.candlepin.model.PoolCurator;
 import org.candlepin.model.Product;
 import org.candlepin.model.ProductCurator;
+import org.candlepin.model.dto.ProductAttributeData;;
```
nguyenfilip: double semicolon on lines 40,41

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -0,0 +1,400 @@
...
+            .setParameter("owner_id", owner.getId())
+            .setParameter("product_uuid", product.getUuid())
+            .getSingleResult();
+
+        return count > 0;
+    }
+
+    @Transactional
+    public boolean mapProductToOwner(Product product, Owner owner) {
+        if (!this.isProductMappedToOwner(product, owner)) {
```
nguyenfilip: Why is it necessary to check? I was looking at the usage of this method and it seems that just catching constraintViolation exception here would make more sense and would improve performance.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -0,0 +1,400 @@
...
+            this.create(new OwnerProduct(owner, product));
+
+            return true;
+        }
+
+        return false;
+    }
+
+    @Transactional
+    public int mapProductToOwners(Product product, Owner... owners) {
```
nguyenfilip: What are the plans for the use of this method? I don't see any references in the code.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -0,0 +1,400 @@
...
+            .setParameter("owner_id", owner.getId())
+            .executeUpdate();
+    }
+
+    /**
+     * Updates the product references currently pointing to the original product to instead point to
+     * the updated product for the specified owners.
+     * <p/></p>
+     * <strong>Warning:</strong> Hibernate does not gracefully handle situations where the data
+     * backing an entity changes via direct SQL or other outside influence. While, logically, a
```
nguyenfilip: I suspect that situations where you saw refresh not working was when you loaded data and refreshed it later in the same  transaction. In that case the repeatable read isolation on MySQL will give you the same data as loaded earlier in the transaction. If that's the case its not Hibernate's fault but its RDBMS transactional semantics. Do you remember the exact problem you saw?

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -0,0 +1,400 @@
...
+     * @param updated
+     *  The product other objects should reference
+     *
+     * @param owners
+     *  A collection of owners for which to apply the reference changes
+     *
+     * @return
+     *  a reference to the updated product
+     */
+    public Product updateOwnerProductReferences(Product current, Product updated, Collection<Owner> owners) {
```
nguyenfilip: Why is the third parameter a collection? The only usage of updateOwnerProductReferences that I see will always pass only single owner.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
Ceiu: This happened during unit testing and on PostgreSQL. If an entity contained another embedded or nested entity that was updated/removed by one of the reference modifiers ([update|remove][Product|Content]References), Hibernate would either throw an exception (remove) or revert back to the original state that had long been overwritten and flushed away (update). The only way I could get it to stop behaving this way is by evicting the entity entirely and requerying it.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
Ceiu: Oh yeah. Before I wrote it, I had a number of places where I had a loop over an owners collection in place that I intended to go back and change to use this method.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
Ceiu: Legacy bits, currently. In the first iteration, all of the CRUD operations in the product/content managers took collections of owners, so some of the ancillary methods reflect that.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
Ceiu: Catching the exception looks to be off the table. Even caught, Hibernate marks the session as rollback only, so we can't commit the change.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
nguyenfilip: Ok thats a different thing that I had in mind.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
nguyenfilip: Yeah, my oversight. Working with any exceptions thrown during flush is completely wrong idea in any JPA code.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -283,90 +290,99 @@ private Owner refreshOwner(Owner owner) {
...
+        //     }
 
-            for (Product product : sub.getDerivedProvidedProducts()) {
-                this.updateContentRefs(content, owner, product);
-            }
-        }
+        //     for (Product product : sub.getDerivedProvidedProducts()) {
+        //         this.updateContentRefs(content, owner, product);
+        //     }
+        // }
```
mstead: Should remove this code if it is no longer needed (instead of commenting it out)

Ceiu: Yeah. There's a bunch of stuff that's pending removal once it's reviewed and determined to be obsolete/useless. When you see a block like this, it basically means that I was pretty sure removing it is the right thing to do, but not sure enough to delete it entirely. They should be more heavily scrutinized than other parts.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -283,90 +290,99 @@ private Owner refreshOwner(Owner owner) {
...
-            // Check that this product isn't linking to the same content multiple times...
-            if (mapped.containsKey(content.getId())) {
-                log.warn(
-                    "Multiple references to the same content found on a single product; " +
-                    "discarding duplicate: {} => {}, {}",
-                    product, content, mapped.get(content.getId())
-                );
+                // Check that this product isn't linking to the same content multiple times...
+                // TODO: This probably isn't necessary anymore, since our product DTO is more strict
+                // than our entities
```
mstead: Anything left to do here?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -453,33 +463,33 @@ else if (!incoming.equals(existing)) {
...
+        //         pset.add(this.updateProductRef(products, owner, product));
+        //     }
+        //     sub.setProvidedProducts(pset);
+
+        //     pset.clear();
+        //     for (ProductData product : sub.getDerivedProvidedProducts()) {
+        //         pset.add(this.updateProductRef(products, owner, product));
+        //     }
+        //     sub.setDerivedProvidedProducts(pset);
+        // }
```
mstead: Remove the commented out code if it is no longer needed.

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -166,7 +166,7 @@ private ConfigProperties() {
     public static final String PINSETTER_MAX_RETRIES = "pinsetter.retries.max";
     public static final int PINSETTER_MAX_RETRIES_DEFAULT = 10;
 
-    private static final String[] DEFAULT_TASK_LIST = new String[]{
+    public static final String[] DEFAULT_TASK_LIST = new String[]{
```
vritant: Why change the scope to public?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
mstead: :+1:

server/src/main/java/org/candlepin/config/ConfigProperties.java
Ceiu: At one point in this branch, I was using it in the test config to work around an issue in PinsetterKernel (which now has a comment in it explaining what I think is a better resolution). It eventually got reverted, but I didn't think opening up our default config was a problem so the scope never got changed back.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -0,0 +1,400 @@
...
+        // which prevents it from updating collections backed by a join table.
+        // As an added bonus, it's quicker, but we'll have to be mindful of the memory vs backend
+        // state divergence.
+
+        // Impl note:
+        // We have a restriction in removeProduct which should prevent a product from being removed
+        // from an owner if it is being used by a pool. As such, we shouldn't need to manually clean
+        // the pool tables here.
+
+        // TODO:
```
nguyenfilip: As I understand this is not necessary. Because cp_installed_products reflects what consumer does have installed on the machine. The fact that we remove an available product from an owner doesn't change that. Or is there anything more to this besides what is mentioned in the TODO?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
Ceiu: Ahh, yeah. I've removed this comment, since we're still getting the data from the adapter which may or may not be providing us with subclass that isn't as strict as our base DTO.

server/src/main/java/org/candlepin/model/OwnerProduct.java
```diff
@@ -0,0 +1,118 @@
...
+    public OwnerProduct() {
+        // Intentionally left empty
+    }
+
+    public OwnerProduct(Owner owner, Product product) {
+        this.setOwner(owner);
+        this.setProduct(product);
+    }
+
+    // public OwnerProduct(String ownerId, String productUuid) {
```
nguyenfilip: Commented code can be removed

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
Ceiu: Good point. I've removed the commented out block below for the installed products and the TODO block called out here. If there are any other unlinked RHID references I'm forgetting, we should probably add checks for those (or bring the TODO comment back).

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
nguyenfilip: Just a small note about this. I think this behavior is caused by cascading. The same goes for OwnerContentCurator.testRemoveOwnerContentReferences where you have to evict and then find instead of just refresh. Refresh there will first try to cascade the refresh to collection which causes exceptions because the elements no longer exist in the database.

Yeah I dont think there is anything to be done about this. Eitehr clear() or evict() seems the most bulletproof response in these situations...

server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -75,45 +116,56 @@ public ContentManager(ContentCurator contentCurator, ProductCurator productCurat
...
-        Content existing = this.contentCurator.lookupById(owner, entity.getId());
+        if (owner == null) {
+            throw new IllegalArgumentException("owner is null");
+        }
+
+        log.debug("Creating new content for org: {}, {}", entity, owner);
+
+        Content existing = this.ownerContentCurator.getContentById(owner, entity.getId());
+
+        // TODO: FIXME:
```
nguyenfilip: My understanding is that this PR fixes this so this TODO can be removed?

Ceiu: This TODO and the matching one in ProductManager are referring to the this bug that's triggered if the create method is called with an entity that already exists. I suppose a quick way to address the TODO is to mark the createContent(entity, owner) method as protected so it can't be called directly.

server/src/main/java/org/candlepin/model/dto/ProductData.java
```diff
@@ -0,0 +1,1099 @@
...
+import java.util.List;
+import java.util.Set;
+
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.XmlTransient;
+
+
+
+/**
+ * DTO representing the product data exposed to the API and adapter layers.
```
nguyenfilip: Instead of heandwriting the sample json, you can use Swagger annotations on fields. See example in Status.java

Ceiu: That comment was mostly for my benefit while I wrote up the DTOs, but the Swagger docs would be a good addition to that. I'll get all the DTOs updated accordingly.

server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -75,45 +116,56 @@ public ContentManager(ContentCurator contentCurator, ProductCurator productCurat
...
             throw new IllegalStateException("Content has already been created");
         }
 
         // Check if we have an alternate version we can use instead.
-        List<Content> alternateVersions = this.contentCurator.getContentByVersion(
-            entity.getId(), entity.hashCode()
-        );
+        List<Content> alternateVersions = this.contentCurator
+            .getContentByVersion(entity.getId(), entity.hashCode())
+            .list();
```
mstead: Any reason you are returning criteria instead of just returning the results from the curator?

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -226,51 +196,23 @@ protected Content(Content source) {
...
-        }
-
-        if (source.getArches() != null) {
-            this.setArches(source.getArches());
-        }
+        // Impl note:
+        // In most cases, our collection setters copy the contents of the input collections to their
+        // own internal collections, so we don't need to worry about our two instances sharing a
+        // collection. The exception here is the modifiedProductIds, which uses the collection
+        // directly, so we'll need to make a defensive copy.
```
mstead: Doesn't look like this is still valid.

server/src/main/java/org/candlepin/controller/ContentManager.java
Ceiu: Ahh, this is a small bonus feature of more improvements I'm planning to make to the curators once 2.0 is out and stable. That's actually a CandlepinCriteria, which will eventually let us pull cursors, lists and make minor tweaks to the criteria before it's run (limits, sorting, etc.). This part was migrated in before the final decision to just use DTOs, as at one point I had to change the session for that query. The session changing portion got dropped, but I left the rest in place.

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -939,7 +1077,174 @@ public Product setLocked(boolean locked) {
...
+            // something here. We'll treat it as if they were empty (and not added).
+
+            // This typically only occurs when we're initially pulling the entity down from a normal
+            // lookup. Hibernate stores the entity in a hashmap, which triggers a call to this
+            // method before Hibernate has fully hydrated it and before it's assigned it to an
+            // entity manager or session. As such, as soon as we try to use one of the above
+            // collections, things explode.
+
+            // Note that this only applies to lazy collections, so the product attributes are safe
+            // to check for now.
```
mstead: Would it be worth adding a debug statement here? Kind of makes me nervous.

server/src/main/java/org/candlepin/model/dto/ProductContentData.java
```diff
@@ -0,0 +1,238 @@
...
+     * Initializes a new ProductContentData instance using the data contained by the given entity.
+     *
+     * @param entity
+     *  The source entity from which to copy data
+     */
+    public ProductContentData(ProductContent entity) {
+        if (entity != null) {
+            this.populate(entity);
+        }
+    }
```
mstead: Curious as to why the above constructors are old with the null object values. Seems like it would be an error condition for those sorts of methods.

Ceiu: Not sure what you mean by "old with the null object." Is that supposed to be "okay with?" I'm going to answer assuming that's the case

Initially they did throw an exception much like the populate method itself. At some point I decided that if you intended to create a new DTO, you were probably intending to create it with whatever data you happened to have available. In the event of a null, that would be no data.

I suppose, writing it out, it doesn't make much sense. I can revert that change so the DTO constructors also throw an illegal argument exception in the event of a null entity or source DTO.

server/src/main/java/org/candlepin/model/Product.java
Ceiu: That check's been in master for a while now -- probably just an indentation change. The only time the exception is thrown is during hydration before Hibernate gets it back. So, I can add some debug output here, but it'll be mostly noise.

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
```diff
@@ -73,24 +74,25 @@
...
+        PoolManager poolManager, ServiceLevelValidator serviceLevelValidator,
+        ActivationKeyRules activationKeyRules, ProductCurator productCurator,
+        OwnerProductCurator ownerProductCurator) {
 
         this.activationKeyCurator = activationKeyCurator;
         this.i18n = i18n;
         this.poolManager = poolManager;
         this.serviceLevelValidator = serviceLevelValidator;
         this.activationKeyRules = activationKeyRules;
         this.productCurator = productCurator;
```
mstead: ProductCurator no longer used in this class -- should remove it.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -212,6 +216,7 @@ public ConsumerResource(ConsumerCurator consumerCurator,
         this.consumerCurator = consumerCurator;
         this.consumerTypeCurator = consumerTypeCurator;
         this.productCurator = productCurator;
```
mstead: ProductCurator no longer used -- should remove it.

server/src/main/java/org/candlepin/model/dto/ProductContentData.java
mstead: Ya, meant Ok with. :)

:+1:  to the changes you've proposed.

server/src/main/java/org/candlepin/model/Product.java
mstead: Ok, I'm fine with leaving out the debug if you don't think it would be worth it.

PR #1279, opened 2016-06-21
1348291: Corrected placeholder index for translated "locked" message

Ceiu:
- The "product/content is locked" error message has been corrected to use
  the correct index for the product ID placeholder.


server/src/main/java/org/candlepin/audit/EventAdapterImpl.java
```diff
@@ -106,13 +106,14 @@ public void addMessageText(List<Event> events) {
         for (Event event : events) {
             String eventType = (event.getTarget().name() + event.getType().name());
             String message = MESSAGES.get(eventType);
+
             if (message == null) {
+                // TODO: Should this be translated here if we're translating it below?
```
barnabycourt: Please resolve this TODO

Ceiu: Done.

PR #1280, opened 2016-06-23
1345766: Concurrent content promotion cause 400

nguyenfilip:
Similar to 1345765. This method was designed in the past to throw 409 in
case when a content has already been promoted to an environment. This
wasn't working sometimes under concurrent access. Also the method allows
to pass in a set of content ids to promote but it is possible that
method just partially promotes a subset of those ids.

Adding transactionality and exception handling.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/EnvironmentResource.java
+++ b/server/src/main/java/org/candlepin/resource/EnvironmentResource.java
@@ -214,7 +214,7 @@ public class EnvironmentResource {
 
         Environment env = lookupEnvironment(envId);
 
-        Set<String> contentIds = new HashSet<String>();
+
         // Make sure this content has not already been promoted within this environment
         // Impl note:
         // We have to do this in a separate loop or we'll end up with an undefined state, should
@@ -237,8 +237,10 @@ public class EnvironmentResource {
             }
         }
 
+        Set<String> contentIds = new HashSet<String>();
+
         try {
-            batchCreate(contentToPromote, env, contentIds);
+            contentIds = batchCreate(contentToPromote, env);
         }
         catch (PersistenceException pe) {
             if (rdbmsExceptionTranslator.isConstraintViolationDuplicateEntry(pe)) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -323,6 +328,29 @@ public JobDetail demoteContent(
...
     }
 
+    /**
+     * To make promotion transactional
+     * @param contentToPromote
+     * @param env
+     * @param contentIds
+     */
+    @Transactional
+    public void batchCreate(List<org.candlepin.model.dto.EnvironmentContent> contentToPromote,
```
Ceiu: We could have this method return the collection rather than accept it as input to be modified.

server/src/main/java/org/candlepin/util/RdbmsExceptionTranslator.java
```diff
@@ -54,4 +56,31 @@ public boolean isUpdateHadNoEffectException(RollbackException sqlException) {
...
+                return true;
+            }
+
+            //Postgres error code DUPLICATE OBJECT
+            //https://www.postgresql.org/docs/8.3/static/errcodes-appendix.html
+            if (cve.getSQLState() != null &&
+                cve.getSQLState().equals("23505")) {
+                return true;
+            }
+        }
```
Ceiu: Missing the definition for Oracle.

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -136,7 +136,6 @@
     private Integer entityVersion;
 
     @XmlTransient
-    @Column
```
awood: Do we not need this for JPA to map correctly?

nguyenfilip: Thanks @awood, this was absolute blunder. This shouldn't be in this PR at all.

server/src/main/java/org/candlepin/util/RdbmsExceptionTranslator.java
nguyenfilip: I don't have an Oracle box to test this. And AFAIK we do not run on Oracle right now. Can I just leave a TODO here?

server/src/main/java/org/candlepin/resource/EnvironmentResource.java
nguyenfilip: Fixed, thanks.

PR #1282, opened 2016-06-30
fix deletePools - delete derived pools

vritant:
unless its a data error, I think delete pools wrongfully assumes that **all** derived pools will be deleted when we delete master pools.


vritant:
While I think this fix is good, there are some more pools not being deleted and I'd like to investigate more.

vritant:
points from @mstead 
1. we need to delete stack derived pools
2. we need to revoke ents for derived pools were deleting.

vritant:
3. need to list where deletePools is called and see if its okay to do this

mstead:
> points from @mstead
> 1. we need to delete stack derived pools
> 2. we need to revoke ents for derived pools were deleting.

Delete stack derived pools in some cases. i.e when there is no other entitlement contributing to the stack.

While I agree with the issue when deleting expired pools, we need to make sure that we are deleting the right things.

vritant:
- [ ] Delete stack derived pools in some cases. i.e when there is no other entitlement contributing to the stack.
- [ ] revoke ents for derived pools
- [ ] verify this logic should be in deletePools. what code paths call this method?

vritant:
closed this against #1288

PR #1283, opened 2016-06-30
Added utility constructors to some of the new DTOs

Ceiu:
- Added utility constructors to ProductData and ContentData for creating
  DTOs initialized with some values

### DIFF
--- a/server/src/main/java/org/candlepin/model/Content.java
+++ b/server/src/main/java/org/candlepin/model/Content.java
@@ -120,7 +120,7 @@ public class Content extends AbstractHibernateObject implements SharedEntity, Cl
     @CollectionTable(name = "cp2_content_modified_products", joinColumns = @JoinColumn(name = "content_uuid"))
     @Column(name = "element")
     @Size(max = 255)
-    private Set<String> modifiedProductIds = new HashSet<String>();
+    private Set<String> modifiedProductIds;
 
     @Column(nullable = true)
     @Size(max = 255)
@@ -135,22 +135,11 @@ public class Content extends AbstractHibernateObject implements SharedEntity, Cl
     @Type(type = "org.hibernate.type.NumericBooleanType")
     private boolean locked;
 
-
-    public Content(String name, String id, String label, String type, String vendor, String contentUrl,
-        String gpgUrl, String arches) {
-
-        setName(name);
-        setId(id);
-        setLabel(label);
-        setType(type);
-        setVendor(vendor);
-        setContentUrl(contentUrl);
-        setGpgUrl(gpgUrl);
-        setArches(arches);
-    }
-
+    /**
+     * Default constructor
+     */
     public Content() {
-        // Intentionally left empty
+        this.modifiedProductIds = new HashSet<String>();
     }
 
     /**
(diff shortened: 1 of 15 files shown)


Ceiu:
As embarrassing as it is to admit this, I found a few of bugs, inconsistencies and whatnot in the DTO while adding some tests for ContentData and ProductData.

Going to spend some time getting those fleshed out so this isn't a problem going forward.

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -160,9 +149,30 @@ public Content() {
...
+    public Content(String id, String name, String type, String label, String vendor) {
+        this(id);
+
+        this.setName(name);
+        this.setType(type);
+        this.setLabel(label);
+        this.setVendor(vendor);
+    }
+
+    public Content(String name, String id, String label, String type, String vendor, String contentUrl,
```
vritant: @Ceiu  we should be consistent with the order of the arguments. I like the above order, but as long as we are consistent, I would be fine.

server/src/main/java/org/candlepin/model/dto/CandlepinDTO.java
```diff
@@ -28,8 +29,7 @@
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonFilter("DefaultFilter")
-public abstract class CandlepinDTO implements Cloneable {
-
+public abstract class CandlepinDTO implements Cloneable, Serializable {
```
vritant: :+1:

server/src/main/java/org/candlepin/model/Content.java
Ceiu: Agreed. I'd actually like to remove the other constructor entirely, before it starts getting used in more places than just ueber content creation.

server/src/main/java/org/candlepin/model/dto/CandlepinDTO.java
Ceiu: Don't let me forget to add serialVersionId constants to the DTO classes.

server/src/main/java/org/candlepin/model/dto/ContentData.java
```diff
@@ -111,6 +111,38 @@ public ContentData() {
...
+     * @param type
+     *  The type of the content to be represented by this DTO
+     *
+     * @param label
+     *  The label of the content to be represented by this DTO
+     *
+     * @param vendor
+     *  The vendor of the content to be represented by this DTO
+     */
+    public ContentData(String id, String name, String type, String label, String vendor) {
```
vritant: based on how order of the constructors in Content changes, it would be nice if this was consistent with that too

Ceiu: Keep in mind the constructor in Content that starts with name isn't new -- that's legacy that could be removed or adjusted accordingly. If we're keeping it around, the argument order specified by the method called out here would be my preference (id, name, etc....), as it goes from most to least importance.

vritant:
looks good so far, please add new commits so it will be quicker to review, I can always wait for a squash to merge it in.

server/src/main/java/org/candlepin/model/dto/ContentData.java
vritant: my preference too.

server/src/main/java/org/candlepin/model/dto/CandlepinDTO.java
vritant: :alarm_clock: Tomorrow at 11:00am

PR #1285, opened 2016-07-11
1350771: ExclusiveLock held by Candlepin

nguyenfilip:
Removing foreign key constraints on checkin tables to mitigate a
posisibility of lock waits during GET requests.

When a long running request acquires X lock on a consumer, then any
other requests by the consumer (even GET) will get blocked. This is
because we update consumer checkin tables that have foreign key to
consumer table. This behavior is specific to Postgres 9.2, newer
Postgres has a new feature that is less strict to updates to foreign
tables.

### DIFF
--- a/server/src/main/java/org/candlepin/model/CheckIn.java
+++ b/server/src/main/java/org/candlepin/model/CheckIn.java
@@ -14,7 +14,6 @@
  */
 package org.candlepin.model;
 
-import org.hibernate.annotations.ForeignKey;
 import org.hibernate.annotations.GenericGenerator;
 import org.hibernate.annotations.Index;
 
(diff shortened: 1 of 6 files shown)


nguyenfilip:
This PR depends on #1289

nguyenfilip:
The testing of this branch will need to be done locally. The Jenkins jobs doesn't work for 0.9.49. 

I have run the Spec tests locally for this on Postgres. Passed for me.

nguyenfilip:
retest this please

nguyenfilip:
@vritant @mstead Just FYI, the failures in this PR are known issue for 0.9.49 branch, this must be tested locally. As stated in the previous comment, feel free to review this

nguyenfilip:
Ah this should also be accompanied by adding indexes instead of those FKs

vritant:
@nguyenfilip is there a fix coming up / already done for other versions?

nguyenfilip:
@vritant this particular fix is specific to 0.9.49, the code changed in newer versions of Candlepin

PR #1293, opened 2016-07-14
Fix non deterministic test failure in modifier spec

vritant:
Dont trigger async compliance during batch revoke. the async job may even begin executing before the current job has commit-ed its transaction. in this case, the entitlement was not marked as dirty at fetch time ( because the consumer compliance job marks it so ) and we had an inconsistent state.

We cannot simply remove the job right now. the only remaining use is in the bind operation, where we need to lock the consumer if we use the sync implementation. The appropriate fix for that, coming soon-ish in a separate PR is to move compliance information to a separate table, *without* foreign key on the consumer.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -1698,22 +1698,17 @@ public class CandlepinPoolManager implements PoolManager {
             return;
         }
 
-        log.info("Scheduling Compliance status for {} consumers.", consumerSortedEntitlements.size());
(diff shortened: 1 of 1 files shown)


nguyenfilip:
retest this please

nguyenfilip:
Only one of the spec test combinations failed and that is due to Jenkins problem:

```
ERROR: Build step failed with exception
/home/jenkins/workspace/candlepin-pullrequest-spectests-hosted-mysql does not exist.
	at org.apache.tools.ant.types.AbstractFileSet.getDirectoryScanner(AbstractFileSet.java:483)
```

PR #1295, opened 2016-07-15
Optimized hash code calculation to use cached values for some collections

Ceiu:
- When calculating the hash code of a product or content instance, the hash
  of collections of immutable values is cached and reused
- Methods that mutate the collections which have cached hashes will also
  clear any stored hash
- The Candlepin 2.0 migration task now precalculates the modified
  products and dependent products hash for each content and product that
  is migrated
- Added tests for equality checks and hash code generation for products
  and content


Ceiu:
Was a quick workaround for an issue IT is having, but it seems the direction we want to go is hacking around hashCode in general, not optimizing it. Will look for other options and post a different PR.

PR #1297, opened 2016-07-18
Refactored hashCode implementations in certain high-traffic entities

Ceiu:
- The existing hashCode implementations for Product and Content entities
  (as well as related join- and sub-objects) has been moved to the new
  getEntityVersion method
- The new hashCode implementation for Product and Content now use the
  Red Hat ID for the entity
- Improved equality and hashcode testing in Product and Content

### DIFF
--- a/server/src/main/java/org/candlepin/model/Content.java
+++ b/server/src/main/java/org/candlepin/model/Content.java
@@ -528,6 +528,20 @@ public class Content extends AbstractHibernateObject implements SharedEntity, Cl
 
     @Override
     public int hashCode() {
+        HashCodeBuilder builder = new HashCodeBuilder(7, 17)
+            .append(this.id);
+
+        return builder.toHashCode();
+    }
+
+    /**
+     * Calculates and returns a version hash for this entity. This method operates much like the
+     * hashCode method, except that it is more accurate and should have fewer collisions.
+     *
+     * @return
+     *  a version hash for this entity
+     */
+    public int getEntityVersion() {
         // This must always be a subset of equals
         HashCodeBuilder builder = new HashCodeBuilder(37, 7)
             .append(this.id)
@@ -541,14 +555,23 @@ public class Content extends AbstractHibernateObject implements SharedEntity, Cl
             .append(this.gpgUrl)
             .append(this.metadataExpire)
             .append(this.arches)
-            .append(this.locked)
-            .append(this.modifiedProductIds);
+            .append(this.locked);
 
         // Impl note:
-        // Because we handle the collections specially in .equals, we have to do the same special
-        // treatment here to ensure our output doesn't give us wonky results when compared to the
-        // output of .equals
+        // We need to be certain that the hash code is calculated in a way that's order
+        // independent and not subject to Hibernate's poor hashCode implementation on proxy
+        // collections. This calculation follows that defined by the Set.hashCode method.
+        int accumulator = 0;
 
+        if (!this.modifiedProductIds.isEmpty()) {
+            for (String pid : this.modifiedProductIds) {
+                accumulator += (pid != null ? pid.hashCode() : 0);
+            }
+
+            builder.append(accumulator);
+        }
+
+        // Return
         return builder.toHashCode();
     }
 
@@ -636,6 +659,6 @@ public class Content extends AbstractHibernateObject implements SharedEntity, Cl
     @PrePersist
     @PreUpdate
     public void updateEntityVersion() {
-        this.entityVersion = this.hashCode();
+        this.entityVersion = this.getEntityVersion();
     }
 }
--- a/server/src/main/java/org/candlepin/model/Product.java
+++ b/server/src/main/java/org/candlepin/model/Product.java
@@ -20,7 +20,6 @@ import org.candlepin.model.dto.ProductData;
 import org.candlepin.service.UniqueIdGenerator;
 import org.candlepin.util.Util;
 
-import org.hibernate.LazyInitializationException;
 import org.hibernate.annotations.BatchSize;
 import org.hibernate.annotations.Cascade;
 import org.hibernate.annotations.CascadeType;
@@ -1124,6 +1123,20 @@ public class Product extends AbstractHibernateObject implements SharedEntity, Li
 
     @Override
     public int hashCode() {
+        HashCodeBuilder builder = new HashCodeBuilder(7, 17)
+            .append(this.id);
+
+        return builder.toHashCode();
+    }
+
+    /**
+     * Calculates and returns a version hash for this entity. This method operates much like the
+     * hashCode method, except that it is more accurate and should have fewer collisions.
+     *
+     * @return
+     *  a version hash for this entity
+     */
+    public int getEntityVersion() {
         // This must always be a subset of equals
         HashCodeBuilder builder = new HashCodeBuilder(37, 7)
             .append(this.id)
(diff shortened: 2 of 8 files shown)


server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -1124,53 +1123,62 @@ public boolean equals(Object obj) {
 
     @Override
     public int hashCode() {
+        int hash = 7;
```
barnabycourt: please use HCB

barnabycourt:
One general thought for the branch, for calculating entity_version, java does not make any guarantee of stability of the hashCode calculation other than within a single running instance of an application. Every time the application changes the hashCode calculated for a given object may change. Given that, some adjustments may be needed if we need a stable hash for the entity_version. 

An alternative would be to not use the entity_version at all and endure the pain of comparison across all versions. It depends on how much divergence we could run into for a given content or product.

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -528,6 +528,21 @@ public boolean equals(Object other) {
 
     @Override
     public int hashCode() {
+        int hash = 7;
+
+        hash = 17 * hash + (this.id != null ? this.id.hashCode() : 0);
```
barnabycourt: Java native hashCode() has no stability guarantee

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -1124,53 +1123,62 @@ public boolean equals(Object obj) {
...
+    /**
+     * Calculates and returns a version hash for this entity. This method operates much like the
+     * hashCode method, except that it is more accurate and should have fewer collisions.
+     *
+     * @return
+     *  a version hash for this entity
+     */
+    public int getEntityVersion() {
         // This must always be a subset of equals
         HashCodeBuilder builder = new HashCodeBuilder(37, 7)
```
barnabycourt: HashCodeBuilder is not guaranteed stable across multiple runs of an application.

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -528,6 +528,21 @@ public boolean equals(Object other) {
...
+    }
+
+    /**
+     * Calculates and returns a version hash for this entity. This method operates much like the
+     * hashCode method, except that it is more accurate and should have fewer collisions.
+     *
+     * @return
+     *  a version hash for this entity
+     */
+    public int getEntityVersion() {
```
barnabycourt: Java native hashCode() has no stability guarantee between applications runs: https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()

server/src/main/java/org/candlepin/model/Content.java
barnabycourt: why not just proxy the hashCode of the id?

server/src/main/java/org/candlepin/model/ProductAttribute.java
```diff
@@ -163,6 +163,17 @@ public int hashCode() {
...
 
     /**
+     * Calculates and returns a version hash for this entity. This method operates much like the
+     * hashCode method, except that it is more accurate and should have fewer collisions.
+     *
+     * @return
+     *  a version hash for this entity
+     */
+    public int getEntityVersion() {
+        return this.hashCode();
```
barnabycourt: Java native hashCode() has no stability guarantee between applications runs: https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()

barnabycourt:
One other general thought. If the entity_version needs to be stable between application runs then perhaps something like md5 or sha1 should be used.

server/src/main/java/org/candlepin/model/ProductAttribute.java
Ceiu: This is a general statement that doesn't apply to every implementation. In the cases cited here, we're retrieving the hash codes for strings, integers, booleans and entity objects which themselves define a hashCode implementation which is derived from the afore mentioned hash code implementations.

Our hash codes will be consistent from instance to instance, so long as the Java implementation itself remains consistent.

server/src/main/java/org/candlepin/model/Content.java
Ceiu: Could, I suppose. Was just setting up for what I feel is inevitable change on that method.

server/src/main/java/org/candlepin/model/Product.java
Ceiu: We only use the .hashCode method on objects which have explicitly defined an implementation which is consistent between instances. We also have tests which verify this.

server/src/main/java/org/candlepin/model/ProductAttribute.java
barnabycourt: @Ceiu The java implementation hash no guarantee of consistency. See https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()

> Whenever it is invoked on the same object more than once during an execution of a Java application, the hashCode method must consistently return the same integer, provided no information used in equals comparisons on the object is modified. This integer need not remain consistent from one execution of an application to another execution of the same application.

Ceiu: Yes, I understand that, but that's the /general contract on the method definition/. The actual objects we use in our calculations have *concrete implementations* which are defined and documented or implemented ourselves.

https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#hashCode()
https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#hashCode()
https://docs.oracle.com/javase/7/docs/api/java/lang/Long.html#hashCode()
https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#hashCode()

mstead:
I can see both of your points here. Using an SHA256 or something does address @barnabycourt 's concerns, but is it worth the effort. Chances of the hashcode implementation calculations changing are kinda slim on the objects we are using to compute the hash.

If we leave it as is, I don't think the impact of a new version being created IF something changed would be an issue and it would stabilize again over time.

Personally, I think that using something like SHA256 would be a better solution, but I'm also Ok with leaving it as is. It'll likely be faster if left as is though.

IF we did go with the SHA256 approach, we could make use of the Hasher and HashableStringGenerators candlepin classes to do it fairly quickly and easily.

I'm not sure if this comment helps any, but that's where I stand.

server/src/main/java/org/candlepin/model/ProductContent.java
```diff
@@ -144,6 +144,22 @@ public int hashCode() {
     }
 
     /**
+     * Calculates and returns a version hash for this entity. This method operates much like the
+     * hashCode method, except that it is more accurate and should have fewer collisions.
```
nguyenfilip: Why does it have fewer collisions than HashCodeBuilder?

server/src/main/java/org/candlepin/model/ProductContent.java
```diff
@@ -144,6 +144,22 @@ public int hashCode() {
...
+     *
+     * @return
+     *  a version hash for this entity
+     */
+    public int getEntityVersion() {
+        int hash = 17;
+
+        hash = 7 * hash + (this.content != null ? this.content.getEntityVersion() : 0);
```
nguyenfilip: Why not use HashCodeBuilder?

Ceiu: I'm calling content.getEntityVersion() rather than content.getHashCode() as HCB would do.

nguyenfilip: Do you think this might work?
```
        HashCodeBuilder hcb = new HashCodeBuilder(7, 17);
        hcb.append(enabled);
        if (this.content  != null)
            hcb.append(this.content.getEntityVersion());
```

nguyenfilip:
Other than the HashCodeBuilder comment I think this is good to merge

server/src/main/java/org/candlepin/model/ProductContent.java
Ceiu: I imagine it would, but now you're hashing a hash just to use the HCB. You still have to check for null, so the only logic it saves us is the check on enabled.

PR #1300, opened 2016-07-21
1350455: Improve performance of case insensitive search

nguyenfilip:
Changing strategy for case insensitive searches. In the past we
used ignoreCase() and lower() to perform case insensitive search.
That strategy was performing poorly in the absence of an database
index. Adding such index was not possible in database agnostic way
so the new strategy is to use columns suffixed '_lower'. These
columns have index created and they contain only lower case data.

### DIFF
--- a/server/src/main/java/org/candlepin/model/Consumer.java
+++ b/server/src/main/java/org/candlepin/model/Consumer.java
@@ -169,6 +169,13 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
     @Cascade({org.hibernate.annotations.CascadeType.ALL})
     private Map<String, String> facts;
 
+    @ElementCollection
+    @CollectionTable(name = "cp_consumer_facts_lower", joinColumns = @JoinColumn(name = "cp_consumer_id"))
+    @MapKeyColumn(name = "mapkey")
+    @Column(name = "element")
+    @Cascade({org.hibernate.annotations.CascadeType.ALL})
+    private Map<String, String> factsLower;
+
     @OneToOne(cascade = CascadeType.ALL)
     private KeyPair keyPair;
 
@@ -220,6 +227,7 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
         this.owner = owner;
         this.type = type;
         this.facts = new HashMap<String, String>();
+        this.factsLower = new HashMap<String, String>();
         this.installedProducts = new HashSet<ConsumerInstalledProduct>();
         this.guestIds = new ArrayList<GuestId>();
         this.autoheal = true;
@@ -375,6 +383,21 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
      */
     public void setFacts(Map<String, String> factsIn) {
         facts = factsIn;
+        if (factsIn == null) {
+            factsLower = null;
+        }
+        else {
+            factsLower = new HashMap<String, String>();
+            for (Entry<String, String> f : factsIn.entrySet()) {
+                String val = f.getValue();
+                if (val != null) {
+                    val = val.toLowerCase();
+                }
+
+                factsLower.put(f.getKey(), val);
+            }
+        }
+
     }
 
     /**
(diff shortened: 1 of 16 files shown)


nguyenfilip:
Attaching a mapping of the changes to unit tests that cover them:

Modified method | Test
------------ | -------------
ConsumerCurator.findByVirtUuid | ConsumerCuratorTest.caseInsensitiveVirtUuidMatching
ConsumerCurator.getGuestConsumersMap |  ConsumerCuratorTest.getGuestConsumerMapCaseInsensitive
ConsumerCurator.getHost | ConsumerCuratorTest.caseInsensitiveGetHost
ConsumerCurator.getHypervisor | ConsumerCuratorTest.testGetHypervisorCaseInsensitive
ConsumerCurator.getHypervisorIdRestriction | ConsumerCuratorTest.testGetHypervisorsBulkCaseInsensitive
ConsumerCurator.searchOwnerConsumers |  ConsumerCuratorSearchTest.testSearchOwnerConsumersInsensitiveValue  ConsumerCuratorSearchTest.testSearchHypervisorIdsCaseInsensitive
GuestIdCurator.findByConsumerAndId | GuestIdCuratorTest.findByConsumerAndIdCaseInsensitive
GuestIdCurator.findByGuestIdAndOrg |  GuestIdCuratorTest.findByGuestIdAndOrgCaseInSensitive

nguyenfilip:
I have not created _lower column for HypervisorId.hypervisorId. Thats because the current code already converts it to lower case 

```
    public void setHypervisorId(String hypervisorId) {
        // Hypervisor uuid is case insensitive, we need to force it to lower
        // case in order to enforce the unique hypervisorId per org constraint
        if (hypervisorId != null) {
            hypervisorId = hypervisorId.toLowerCase();
        }
        this.hypervisorId = hypervisorId;
    }
```

nguyenfilip:
To test Liquibase migration, first deploy Candlepin without changes in this PR. Then use the following data.sql file:
```sql
insert into cp_owner (id, displayname, account) values ('o','o','o');
insert into cp_consumer (uuid,id, name, owner_id, type_id) values ('c','c','c', 'o',1000);

insert into cp_consumer_guests (id, guest_id, consumer_id) values ('g1', 'aAaA','c');
insert into cp_consumer_guests (id, guest_id, consumer_id) values ('g2', 'bbbb','c');
insert into cp_consumer_guests (id, guest_id, consumer_id) values ('g3', 'CCCC','c');

insert into cp_consumer_facts(cp_consumer_id,element,mapkey) values ('c','aAaA', 'k1');
insert into cp_consumer_facts(cp_consumer_id,element,mapkey) values ('c','bbb', 'k2');
insert into cp_consumer_facts(cp_consumer_id,element,mapkey) values ('c','CCCC', 'k3');
```

Mysql:
```bash
mysql -u candlepin candlepin < data.sql
```

Postgres:
```bash
psql -U candlepin candlepin < data.sql
```

Then checkout this PR, run liquibase update. Then  just have to verify the data were converted to _lower columns. 

Mysql:
```bash
echo "select * from cp_consumer_facts_lower"  | mysql -u candlepin candlepin

echo "select guest_id, guest_id_lower from cp_consumer_guests"  | mysql -u candlepin candlepin
```

Postgres: 
```bash
echo "select * from cp_consumer_facts_lower"  | psql -U candlepin candlepin

echo "select guest_id, guest_id_lower from cp_consumer_guests"  | psql -U candlepin candlepin
```

nguyenfilip:
Also I think its good to note that there is a difference between the former usage of lower() function in native queries e.g.:

```sql
"lower(cp_consumer_facts.element) in (:guestids) " +
```

and usage of ignoreCase() on HIbernate query. Usage of lower() function directly in the query assumes that the parameter (:guestids) is already lowercase! However, ignoreCase() takes care of that.

nguyenfilip:
I think this PR removes all usages of both ignoreCase() and lower() function. By full-text search it is possible to find one more usage of 'lower(' but that is in class InExpressionIgnoringCase that is not being used in the code.

vritant:
@nguyenfilip 
1. just like HypervisorId.hypervisorId , ContentOverride.name is also ensured to be always lower case. ( look at ContentOverrideCurator.sanitize() ) . so we dont need the lower column in this case as well. I would also recommend you add a comment in both these cases with something like "there are other bits of code that depend on this being lower case, so please dont remove this "

2. How about we remove InExpressionIgnoringCase? if you think thats okay, please remove that in this PR, or I can create another PR too, let me know

vritant:
@nguyenfilip PR looks good otherwise

nguyenfilip:
Removed InExpressionIgnoringCase and removed lower column for ContentOverride.name

PR #1304, opened 2016-07-25
Added attribute constants for pool and product attributes

Ceiu:
- Removed most, if not all, magic strings for known attribute
  values and replaced them with constants
- Optimized the message translators

### DIFF
--- a/server/src/main/java/org/candlepin/model/Product.java
+++ b/server/src/main/java/org/candlepin/model/Product.java
@@ -77,10 +77,78 @@ import javax.xml.bind.annotation.XmlTransient;
 @Entity
 @Table(name = "cp2_products")
 public class Product extends AbstractHibernateObject implements SharedEntity, Linkable, Cloneable {
-    public static final String UEBER_PRODUCT_POSTFIX = "_ueber_product";
+    /**
+     * Commonly used/recognized product attributes
+     */
+    public static final class Attributes {
+        /** Attribute to specify the architecture on which a given product can be installed/run; may be set
+         *  to the value "ALL" to specify all architectures */
+        public static final String ARCHITECTURE = "arch";
 
-    public static final String CONTENT_OVERRIDE_ENABLED_ATTRIB = "content_override_enabled";
-    public static final String CONTENT_OVERRIDE_DISABLED_ATTRIB = "content_override_disabled";
+        /** Attribute for specifying the type of branding to apply to a marketing product (SKU) */
+        public static final String BRANDING_TYPE = "brand_type";
+
+        /** Attribute for enabling content overrides */
+        public static final String CONTENT_OVERRIDE_ENABLED = "content_override_enabled";
+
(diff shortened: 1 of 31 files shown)


Ceiu:
Reviewers: Please give some attention to the documentation associated with the new attributes/keys. Some are missing entirely (as I have no idea what they represent), or may be varying degrees of incorrect. If you happen to know what they should be, throw that in any comment and I'll get it updated.

mstead:
I know the [Product Attribute](http://www.candlepinproject.org/docs/candlepin/product_attributes.html) doc isn't complete, but perhaps some of the product attribute descriptions can be pulled from here to make the comments a little more informative.

Ceiu:
@mstead that's where I pulled what little information I have for those attributes. Unfortunately, the doc states "passed to the certificate" for most of the attributes that weren't already filled in. Maybe as part of the "update the docs" we could update these as well.

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -77,10 +77,75 @@
 @Entity
 @Table(name = "cp2_products")
 public class Product extends AbstractHibernateObject implements SharedEntity, Linkable, Cloneable {
-    public static final String UEBER_PRODUCT_POSTFIX = "_ueber_product";
+    /**
+     * Commonly used/recognized product attributes
+     */
+    public static final class Attributes {
+        /** Attribute to specify the architecture on which a given product can be installed/run */
+        public static final String ARCHITECTURE = "arch";
```
mstead: Might be worth noting that a value of "ALL" can be used to specify all archs? Unless we want to put that in the docs instead.

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -77,10 +77,75 @@
...
+        /** Attribute specifying the number of cores that can be covered by an entitlement using the SKU */
+        public static final String CORES = "cores";
+
+        /** Attribute specifying whether or not derived pools created for a given product will be
+         *  host-limited */
+        public static final String HOST_LIMITED = "host_limited";
+
+        /** Attribute used to specify the instance multiplier for a pool */
+        public static final String INSTANCE_MULTIPLIER = "instance_multiplier";
+
```
mstead: Perhaps add: Triggers instance based subscriptions. Multiplies size of the pool, but physical binds will consume entitlements in multiples of this value.

Ceiu: Done

server/src/main/java/org/candlepin/model/Product.java
Ceiu: Done

mstead:
The code looks great. I think that whoever does the documentation story should also revisit the comments on the attribute classes to update them to match the new docs.

I'll update the trello card to reflect this.

PR #1305, opened 2016-07-26
Added rest API to trigger cronjobs

vritant:


Ceiu:
Is there a reason we shouldn't just make this a standard part of the API for admins?

vritant:
@Ceiu before implementation I was concerned that this could be abused or that we would have to support something risky that we did not need. but given that we only allow an explicit list of jobs to run, I think that is a good idea.
but we agreed on the design on the condition that this API would not be shipped, so i'd like to make sure @barnabycourt is okay with it.

barnabycourt:
@Ceiu @vritant Given the use of the default tasks list to limit to explicit jobs I think that this is generally useful for admins. If it is to be made generally available I'm not so sure about /cronjob as the API endpoint. Perhaps something under /jobs?

Ceiu:
/jobs or /admin would be my vote

barnabycourt:
how about layering on top of the scheduler endpoint: /jobs/scheduler/<job_name>?

vritant:
@barnabycourt @Ceiu  the need for a different resource was driven by the requirement that we exclude it from shipping. now that we are no longer bound by that, added the resource :
jobs/cron/#{task}

server/src/main/java/org/candlepin/resource/JobResource.java
```diff
@@ -174,6 +181,42 @@ public SchedulerStatus setSchedulerStatus(boolean running) {
...
+                        .forName(className);
+                return pk.scheduleSingleJob(taskClass, Util.generateUUID());
+            }
+            else {
+                throw new BadRequestException("Not a cron job:" + task);
+            }
+        }
+        catch (ClassNotFoundException e) {
+            throw new IseException(i18n.tr("Error trying to schedule {}: {}", className,
+                    e.getMessage()));
```
Ceiu: Indentation here (and elsewhere) too deep. use 4 characters instead of 8.

server/src/main/java/org/candlepin/resource/JobResource.java
```diff
@@ -174,6 +181,42 @@ public SchedulerStatus setSchedulerStatus(boolean running) {
         return getSchedulerStatus();
     }
 
+    @POST
+    @Path("cron/{task}")
```
Ceiu: Would this make more sense (externally?) to call put this at jobs/schedule/:task?

server/src/main/java/org/candlepin/resource/JobResource.java
```diff
@@ -174,6 +181,42 @@ public SchedulerStatus setSchedulerStatus(boolean running) {
...
+                if (className.equalsIgnoreCase(cronJob)) {
+                    isCron = true;
+                    // helps ignore case
+                    className = cronJob;
+                }
+            }
+            if (isCron) {
+                Class<? extends KingpinJob> taskClass = (Class<? extends KingpinJob>) Class
+                    .forName(className);
+                return pk.scheduleSingleJob(taskClass, Util.generateUUID());
```
Ceiu: You can clean this up a bit and save yourself from line wrapping by declaring the class generically, then casting in the schedule line:

```Class taskClass = Class.forName(className);
return pk.scheduleSingleJob((Class<? extends KingpinJob>) taskClass, Util.generateUUID());```

File: server/src/main/java/org/candlepin/resource/JobResource.java [hunk 4]
```diff
@@ -174,6 +181,42 @@ public SchedulerStatus setSchedulerStatus(boolean running) {
...
+            }
+            else {
+                throw new BadRequestException("Not a cron job:" + task);
```
Ceiu: We're translating the exception messages below, but not this one. Is that intentional?

server/src/main/java/org/candlepin/resource/JobResource.java
vritant: @Ceiu how about "schedule_cron"? I see merit in clarifying this is for cron jobs only.

server/src/main/java/org/candlepin/resource/JobResource.java
vritant: if would be a tiny bit misleading though if it errors out. if there is an exception thrown, i'd rather it be in the Class.forName line. thoughts?

server/src/main/java/org/candlepin/resource/JobResource.java
vritant: no, missed that when I moved the class. will fix with other fixes after above things are sorted out.

server/src/main/java/org/candlepin/resource/JobResource.java
Ceiu: I'm not even sure I understand the difference without sifting through each job and looking for DEFAULT_SCHEDULE.

server/src/main/java/org/candlepin/resource/JobResource.java
Ceiu: ClassCastException in either case, one line away from the declaration. You never actually use the class outside of this line, so if you're going to start line wrapping for the sake of exception accuracy, you may as well drop the extraneous declaration and do it all in one line:

```
return pk.scheduleSingleJob((Class<? extends KingpinJob>) Class.forName(className),
    Util.generateUUID());
```

Ceiu:
Changes look good. Will merge once Jenkins passes.

server/src/main/java/org/candlepin/resource/JobResource.java
```diff
@@ -174,6 +181,45 @@ public SchedulerStatus setSchedulerStatus(boolean running) {
...
+                    // helps ignore case
+                    className = permissibleJob;
+                }
+            }
+            if (allowed) {
+                Class taskClass = Class.forName(className);
+                return pk.scheduleSingleJob((Class<? extends KingpinJob>) taskClass, Util.generateUUID());
+            }
+            else {
+                throw new BadRequestException(i18n.tr("Not a permissible job: {0}", task));
```
Ceiu: In the next steps for this overarching task, we should revisit this error message and focus a bit more on the "why" it's not allowed to be triggered manually. Might require more work than just an else-branch though.

Ceiu:
ACK

```
(16:06:34) crog@cl-vm-1:~/devel/candlepin/server(vritant/cron-tests) $ curl -k -u admin:admin "https://localhost:8443/candlepin/jobs/schedule/ExpiredPoolsJob"
{
  "displayMessage" : "Runtime Error No resource method found for GET, return 405 with Allow header at org.jboss.resteasy.core.registry.SegmentNode.match:375",
  "requestUuid" : "9c158d57-e7f2-4f35-a403-4b6dce585fe9"
}(16:07:42) crog@cl-vm-1:~/devel/candlepin/server(vritant/cron-tests) $ curl -k -u admin:admin -X POST "https://localhost:8443/candlepin/jobs/schedule/ExpiredPoolsJob"
{
  "id" : "a6382ff0-446c-42f9-b3d8-67b227a89cfd",
  "state" : "CREATED",
  "startTime" : null,
  "finishTime" : null,
  "result" : null,
  "principalName" : "System",
  "targetType" : null,
  "targetId" : null,
  "ownerId" : null,
  "resultData" : null,
  "statusPath" : "/jobs/a6382ff0-446c-42f9-b3d8-67b227a89cfd",
  "done" : false,
  "group" : "async group",
  "created" : "2016-07-29T20:08:01+0000",
  "updated" : "2016-07-29T20:08:01+0000"
}(16:08:01) crog@cl-vm-1:~/devel/candlepin/server(vritant/cron-tests) $ less /var/log/candlepin/candlepin.log 
(16:08:19) crog@cl-vm-1:~/devel/candlepin/server(vritant/cron-tests) $ curl -k -u admin:admin -X POST "https://localhost:8443/candlepin/jobs/schedule/RefreshPoolsJob"
{
  "displayMessage" : "Not a permissible job: RefreshPoolsJob",
  "requestUuid" : "adee9a1f-3034-43d3-92ab-dc4760e019ca"
}
```

PR #1306, opened 2016-07-27
Undo async compliance in batch bind

vritant:


vritant:

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1504,17 +1504,15 @@ else if (!poolQuantities.keySet().equals(entitlements.keySet())) {
...
-        catch (PinsetterException e) {
-            log.error("ConsumerComplianceJob schedule failed", e.getMessage());
+        /*
+         * If the consumer is not a distributor, check consumer's new compliance
+         * status and save. the getStatus call does that internally, so we only
+         * need to check for the update.
+         */
+        complianceRules.getStatus(consumer, null, false, false);
+        if (!isDistributor) {
+            consumerCurator.update(consumer);
```
nguyenfilip: Why is it necessary to call update()? The getStatus() already calls updateWithOptionalFlush internally and it does so only when status changed (which is the correct behavior I think)

nguyenfilip:
@vritant Just to have a bit more context. What is the main reason for taking this out? As I understand, async bind increased performance of bind significantly.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
vritant: @nguyenfilip this PR mostly reverts code to its prior state, so this was pre-existing behaviour.
I *think* the intent was that in getStatus, the update is only if complianceChanged or entStatusChanged, while this way its unconditionally changed.

that being said, I dont see what other changes could be made in this method itself, but this is a common method used in a lot of places so I found it safer to keep it this way.
thoughts?

PR #1310, opened 2016-08-01
1359814: Import manifest fail with excess entitlement (0.9.49)

nguyenfilip:
Fixes in processPoolUpdate which is part of import code path.
Defensively checking to determine if the pool has already been deleted.
In such cases, we need to skip processing updates for it.


nguyenfilip:
Currently running spec tests on my VM

nguyenfilip:
On Postgresql this passes for me. On MySQL there is a failure (ueber_cert_spec) that seems to be unrealted

vritant:
Closing this PR since we do not want this fix in 0.9.49. Please put up a PR for version 9.54 instead.

PR #1311, opened 2016-08-01
1361537: Import manifest fail with excess entitlement (2.0)

nguyenfilip:
Fixes in processPoolUpdate which is part of import code path.
Defensively checking to determine if the pool has already been deleted.
In such cases, we need to skip processing updates for it.


wottop:
I see that your spec test covers the case of updating a pool for a virt change.
I would like a test (unit) that shows processPoolUpdates never hits the merge if the pool no longer exists.

nguyenfilip:
@wottop makes sense!

wottop:
ack

PR #1314, opened 2016-08-03
1363647: Can't deploy on PostgreSQL 9.5.3

nguyenfilip:
Database datatype smallint was causing issues when
deploying on Postgresql


wottop:
I have run this against master. I don't see the issue you describe.

I am running postgresql-9.5.3-1.fc24.x86_64 on Fedora 24

nguyenfilip:
The problem was in my candlepin.conf. It sets hbm2ddl.auto to validate, which forces Hibernate to check the data type
```
 #Old Satellite config
jpa.config.hibernate.hbm2ddl.auto=validate
```

PR #1307, opened 2016-07-28
REST Swagger Checkstyle check

dsarman:
Adds custom Checkstyle check that looks at all classes with "Path" and without "SupressSwaggerCheck" annotation, and checks if their methods have the proper Swagger annotation ("ApiOperation").
Also integrated the check into buildfile and fixed all Checkstyle errors

PR #1319, opened 2016-08-05
Removed useless dependency from buildfile.

dsarman:
Fixup of #1307.
Removed useless dependency from candlepin common project in buildfile
which caused that pom.xml files after "buildr eclipse" were show as
modified in git.


Ceiu:
ACK

PR #1323, opened 2016-08-10
1344100: servicelevels returned by candlepin should exclude values from expired pools

dsarman:
Changed corresponding HQL query to take into account pools endDate, and also pools entitlements endDateOverride.

### DIFF
--- a/server/spec/consumer_resource_spec.rb
+++ b/server/spec/consumer_resource_spec.rb
@@ -671,6 +671,24 @@ describe 'Consumer Resource' do
 
   end
 
+  it 'should not list service levels from expired pools' do
+    product = create_product(random_string('product'),
+                              random_string('product'),
+                              {:attributes => {:support_level => 'Expired'},
+                               :owner => @owner1['key']})
+    create_pool_and_subscription(@owner1['key'], product.id,
+                                 1, [], '', '', '', Date.today - 2, Date.today - 1)
+
+    user_cp = user_client(@owner1, random_string('billy'))
+    consumer = user_cp.register(random_string('system'), :system, nil,
+                                {}, nil, nil, [], [])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+    consumer = @cp.get_consumer(consumer['uuid'])
+
+    service_levels = consumer_client.list_owner_service_levels(@owner1['key'])
+    service_levels.length.should eq(0)
+  end
+
   it 'should allow a consumer dry run an autosubscribe based on service level' do
     product1 = create_product(random_string('product'),
       random_string('product'),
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -648,8 +648,10 @@ public void lock(List<Pool> poolsToLock) {
             "FROM Pool AS Pool " +
             "  INNER JOIN Pool.product AS Product " +
             "  INNER JOIN Product.attributes AS Attribute " +
+            "  LEFT JOIN Pool.entitlements AS Entitlement " +
```
nguyenfilip: With large amount of entitlements, I think this might be very slow. What do you think about using EXISTS clause instead of the left join?

What I mean is adding, into the where clause, an OR that will check whether Entitlement exists such that it entitles the Pool, and at the same time has endDateOverride >= current_date()

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -648,8 +648,10 @@ public void lock(List<Pool> poolsToLock) {
             "FROM Pool AS Pool " +
             "  INNER JOIN Pool.product AS Product " +
             "  INNER JOIN Product.attributes AS Attribute " +
```
nguyenfilip: We usually name the aliases with small case -"attribute" instead of "Attribute"

server/src/main/java/org/candlepin/model/PoolCurator.java
nguyenfilip: Also, @Ceiu mentioned that Hibernate may not send the DISTINCT down to the database (but maybe that is issue only when using Criteria). That would be another good reason not to use JOIN's here at all, but replace them with EXIST subqueries in the WHERE clause.

Ceiu: This looks to have been addressed already, but for clarity, the DISTINCT issue applies to both HQL and Criteria, but only in the case where we're using ScrollableResults.

server/spec/consumer_resource_spec.rb
```diff
@@ -671,6 +671,24 @@ def verify_installed_pids(consumer, product_ids)
 
   end
 
+  it 'should not list service levels from expired pools' do
+    product = create_product(random_string('product'),
+                              random_string('product'),
+                              {:attributes => {:support_level => 'Expired'},
+                               :owner => @owner1['key']})
+    create_pool_and_subscription(@owner1['key'], product.id,
+                                 1, [], '', '', '', Date.today, Date.today - 1)
```
wottop: You have a range from today though yesterday. Let's use something valid like Date.today-2, Date.today-1.

wottop:
ack

PR #1327, opened 2016-08-19
Capability list change detection wrong

wottop:
In addition, the new list would be merged in either case, change or not.

### DIFF
--- a/server/src/test/java/org/candlepin/resource/ConsumerResourceUpdateTest.java
+++ b/server/src/test/java/org/candlepin/resource/ConsumerResourceUpdateTest.java
@@ -815,6 +815,34 @@ public class ConsumerResourceUpdateTest {
     }
 
     @Test
+    public void consumerChangeDetection() {
+        Consumer existing = getFakeConsumer();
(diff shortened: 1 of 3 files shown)


server/src/test/java/org/candlepin/resource/ConsumerResourceUpdateTest.java
```diff
@@ -815,6 +815,33 @@ public void consumerCapabilityUpdate() {
...
+        existing.setCapabilities(caps1);
+
+        Consumer toUpdate = getFakeConsumer();
+        toUpdate.setCapabilities(caps1);
+        assertFalse(resource.performConsumerUpdates(existing, toUpdate, null));
+
+        toUpdate.setCapabilities(caps2);
+        assertTrue(resource.performConsumerUpdates(existing, toUpdate, null));
+
+        toUpdate.setCapabilities(null);
```
nguyenfilip: Consumer.setCapabilities checks that null is never set into the Consumer.capabilities field. I think intention of this part of the test was to test this path: 
```java
 if (update.getCapabilities() != null) {
```
it doesn't do that.

PR #1331, opened 2016-08-19
Upgrade Hibernate to 5.1.1

nguyenfilip:
This upgrade is needed to support EHCache. Reviewer please also verify this works with JDK 1.7 - I checked it, but just to be sure.

### DIFF
--- a/server/src/main/java/org/candlepin/model/JobCurator.java
+++ b/server/src/main/java/org/candlepin/model/JobCurator.java
@@ -24,6 +24,7 @@ import org.candlepin.pinsetter.core.model.JobStatus.TargetType;
 import org.candlepin.pinsetter.tasks.KingpinJob;
 
 import com.google.inject.Inject;
+import com.google.inject.persist.Transactional;
 
 import org.hibernate.Criteria;
 import org.hibernate.Query;
@@ -50,6 +51,7 @@ public class JobCurator extends AbstractHibernateCurator<JobStatus> {
         this.config = config;
     }
 
+    @Transactional
     public JobStatus cancel(String jobId) {
         this.cancelNoReturn(jobId);
         JobStatus result = this.find(jobId);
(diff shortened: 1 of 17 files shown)


nguyenfilip:
Investigation is needed to figure out why the Transactional annotations are necessary for tests to pass.

nguyenfilip:
retest this please

nguyenfilip:
I cannot get the unit tests to fail the same way again. Now everything looks happy even without Transactional annotations

mstead:
Out of curiosity, what changed that now requires us to add the new @Transactional annotations?

server/src/main/java/org/candlepin/model/JobCurator.java
```diff
@@ -59,6 +61,7 @@ public JobStatus cancel(String jobId) {
         return result;
     }
 
+    @Transactional
```
nguyenfilip: @mstead I dont know exact answer, but the reason we need to add Transactional is that new Hibernate doesn't allow running update HQL outside of a transaction. If you think about it it is more appropriate behavior anyway.

nguyenfilip:
To review with JDK 1.6. 
install Oracle Java 1.6
```
 cd into dir where you want your java to be installed 
 wget http://file.rdu.redhat.com/fnguyen/jdk-6u45-linux-x64.bin
 chmod +x jdk-6u45-linux-x64.bin
./jdk-6u45-linux-x64.bin
```
Modify JAVA_HOME for tomcat:
```
sudo vi /etc/tomcat/tomcat.conf
export JAVA_HOME=/home/fnguyen/down/jdk1.6.0_45
```

make sure you build with 1.6 as well - export JAVA_HOME before you run buildr or deploy scripts...

Before building candlepin, remove checks project (Swagger checkstyle plugin) 
```
rm -rf checks/
``
Deploy on MySQL (because Postgres JDBC driver requires Java 1.8 and working around that would  be hassle again)

mstead:
Tested against 1.7 and 1.8.

PR #1332, opened 2016-08-19
Cache the host on a guest on the first lookup

wottop:
Improves performace where the host was retrieved repeatedly directly
from the database each time.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -104,7 +104,7 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
-  activesupport
+  activesupport (~> 4.2)
   builder
   buildr (= 1.4.24)
   buildr-findBugs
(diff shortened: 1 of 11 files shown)


mstead:
I don't think I like the way that the host data is set on the consumer here. The curator is now responsible for setting the correct data on the Consumer in a separate call. Pulling a Consumer from the database and calling the accessor methods will not reflect the correct state of the consumer.

For example, if I look up a consumer by id, and then call hasHost it will return 'true', but it could be a host consumer itself. Or, the host could be null even if it has a host.

IMO the caching should be somewhere else.

wottop:
Perhaps the boolean hasHost should be hasCheckedForHost [and perhaps inverse] If host is null and hasHost is true, it will do the lookup to the data base. If it gets null back, then it flips has host to false. That keeps it from re-calling the DB when there is no host.

Does that help here?

wottop:
Closing for now. Potentially better option on new branch.

PR #1333, opened 2016-08-24
Fix potential double entitlement issue by locking earlier in the transaction (0.9.54)

cnsnyder:
If there are more than one guest on the same hypervisor that checkin or autobind at near the same time it is possible for the hypervisor to end up with more than one of an entitlement necessary to cover the guests.

I was able to replicate this by autobinding multiple guests at the same time. When done with 2 guests the hypervisor ends up with 2 of the same entitlement.

The solution to this is to lock the hypervisor consumer earlier in the transaction to ensure only one autobind modifies the entitlements of the hypervisor at any time.

The performance implications seem to be minimal with a failure rate of 7% when auto attaching 500 guests (that are on the same hypervisor) at the same time. (For reference the failure rate is around 10% in my tests on 0.9.54-HOTFIX currently).

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -545,6 +545,7 @@ public class CandlepinPoolManager implements PoolManager {
     public List<Entitlement> entitleByProductsForHost(Consumer guest, Consumer host,
             Date entitleDate, Collection<String> possiblePools)
         throws EntitlementRefusedException {
+        host = consumerCurator.lockAndLoad(host);
         List<Entitlement> entitlements = new LinkedList<Entitlement>();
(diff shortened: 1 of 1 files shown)


nguyenfilip:
retest this please

nguyenfilip:
@cnsnyder  Since its against 0.9.54, is there an associated upstream/downstream BZ?

nguyenfilip:
@cnsnyder Why backporting  17ee59c ? If its just for debugging purposes during development, I think we should not backport it to 0.9.X

nguyenfilip:
retest this please

nguyenfilip:
retest this please

cnsnyder:
@nguyenfilip It was just for debugging purposes. I can remove it.

wottop:
acked on 100 concurrent guest heals

PR #1334, opened 2016-08-29
Cache host consumers on curator

wottop:
Saving repeated lookups to database. Only caches over one REST query.

### DIFF
--- a/server/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/server/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -71,6 +71,8 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
     private static final int MAX_IN_QUERY_LENGTH = 500;
     private static Logger log = LoggerFactory.getLogger(ConsumerCurator.class);
 
+    private Map<String, Consumer> cachedHosts = new HashMap<String, Consumer>();
+
     public ConsumerCurator() {
         super(Consumer.class);
     }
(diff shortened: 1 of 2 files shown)


barnabycourt:
@wottop Hi, One suggestion about the use of two maps. Why not use one map and check if the cachedHosts map contains the key? That way you would only need one additional map for the functionality.

wottop:
If we look up the consumer and the answer is null, then the get from the table is null.  We will look up the consumer every time because of it. The second table says we looked it up.

kahowell:
@wottop, are you planning on changing this if/when #1336 gets merged? Or is this something that we can't/shouldn't use JCache for?

barnabycourt:
@kahowell this is not necessarily a good fit for jcache as it is only cached within the request context. We would never want this to be used across multiple requests

mstead:
Please add some unit tests.

mstead:
Perhaps add some comments in the class as to why this works, why this is necessary, and the primary case this is used.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -512,7 +519,9 @@ public Consumer getHost(String guestId, Owner owner) {
             .addOrder(Order.desc("updated"))
             .setMaxResults(1)
             .setProjection(Projections.property("consumer"));
-        return (Consumer) crit.add(guestIdCrit).uniqueResult();
+        Consumer host = (Consumer) crit.add(guestIdCrit).uniqueResult();
```
Ceiu: Can we move the .add to the block above where we're adding all the other restrictions? 

```
         Criteria crit = currentSession()
             .createCriteria(GuestId.class)
             .createAlias("consumer", "gconsumer")
             .add(Restrictions.eq("gconsumer.owner", owner))
             .add(guestIdCrit)
             .addOrder(Order.desc("updated"))
             .setMaxResults(1)
             .setProjection(Projections.property("consumer"));

        Consumer host = (Consumer) crit.uniqueResult();
```

Ceiu:
ACK. Auto attaching a guest in master hits this query 38 times; auto-attaching the same consumer with this branch brings that down to three times.

Just need a unit test to verify the cache is used (even though it's fairly obvious by looking at it).

barnabycourt:
@Ceiu Hmm... why would it still be called 3 times? 
1 for the Guest
1 for the Host
Where is the 3rd call coming from?

PR #1336, opened 2016-08-31
Integrate EHCache as 2LC and JCache

nguyenfilip:
Integration of EHCache as 2LC and JCache. This PR depends on #1331 

This is only integration of the caches, together with a possibility to configure statistics over JMX in candlepin.conf:
```
cache.jmx.statistics=true
jpa.config.hibernate.generate_statistics=true
jpa.config.hibernate.use_structured_entries=true
jpa.config.net.sf.ehcache.configurationResourceName=ehcache-stats.xml
```
This PR doesn't introduce any caching.

### DIFF
--- a/buildfile
+++ b/buildfile
@@ -90,6 +90,11 @@ JAVAX = ['org.hibernate.javax.persistence:hibernate-jpa-2.1-api:jar:1.0.0.Final'
          'javax.transaction:jta:jar:1.1']
 ANTLR = ['antlr:antlr:jar:2.7.7']
 
+EHCACHE = ['org.hibernate:hibernate-ehcache:jar:5.1.1.Final', 
+           'net.sf.ehcache:ehcache:jar:2.10.1', 
+           'org.ehcache:jcache:jar:1.0.0', 'javax.cache:cache-api:jar:1.0.0',
+           'net.sf.ehcache:management-ehcache-v2:jar:2.10.1']
+
 HIBERNATE = [group('hibernate-core', 'hibernate-entitymanager', 'hibernate-c3p0',
                    :under => 'org.hibernate',
                    :version => '5.1.1.Final'),
@@ -97,6 +102,7 @@ HIBERNATE = [group('hibernate-core', 'hibernate-entitymanager', 'hibernate-c3p0'
              'org.hibernate:hibernate-tools:jar:3.2.4.GA',
              'org.hibernate:hibernate-validator:jar:4.3.1.Final',
              ANTLR,
+             EHCACHE,
              'asm:asm:jar:3.0',
              'cglib:cglib:jar:2.2',
              'org.javassist:javassist:jar:3.20.0-GA',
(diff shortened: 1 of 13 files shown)


nguyenfilip:
The best way to test this PR is to browse cache with JConsole: jconsole localhost:3322 

Also, eye-check the dependencies in buildfile that I had to introduce.

To test HibernateStatisticsCurator you can inject it e.g. into StatusResource and then just play with it. To be honest there is no plan so far to use HibernateStatisticsCurator so if you feel its not a best idea to have this 'dead code' in Candlepin, let me know and I will remove it.

nguyenfilip:
Just a few links to EHCache docs:
http://www.ehcache.org/generated/2.10.2/html/ehc-all/
http://www.ehcache.org/apidocs/2.10.2/index.html

mstead:
Looks good!

Regarding the StatisticsCurator, we could include a call to if in debug mode or have a config options for it or something. I think it would be useful to have if we ever wanted IT to provide the info to us. No rebuild required.

Consider this PR ACKED but I won't merge it yet since it still has the in-progess label.

nguyenfilip:
@mstead squashed and rebased against master. Now, the files changed view is much cleaner.

I am still not sure how to use StatisticsCurator. We can use think of something later when we get more experienced with the 2LC

barnabycourt:
@nguyenfilip How are the HibernateStatus and the HibernateStatisticsCurator being used in this pull request? If they are there for future functionality then perhaps that part could be left off until there is function using it. If it is being used somewhere and I am just missing it then please disregard.

nguyenfilip:
@barnabycourt it is for the future. We might need it at some point to collect statistics, but I am unsure where/when we want to do that right now. I will pull out these bits to a feature branch for future reference.

nguyenfilip:
@mstead @barnabycourt removed the in-code statistics support (will keep it in branch fnguyen/ehcache-hibernate-stats). Note that, obviously, all the JMX statistics support stays in

nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
retest this please

barnabycourt:
@nguyenfilip If I read this properly, the overFlowToDkisk configuration paramater was deprecated as of EHCache 2.6. Since we are using 2.10, should some of these other parameters be used instead? http://www.ehcache.org/documentation/2.7/configuration/fast-restart.html

nguyenfilip:
@barnabycourt per [1], I removed the setting completely. The disk store shouldn't be used unless explicitly defined.

[1] http://www.ehcache.org/generated/2.10.2/html/ehc-all/#page/Ehcache_Documentation_Set%2Fco-tiers_configuring_disk_store.html%23

nguyenfilip:
/etc/tomcat.conf

```
JMX_CONF="-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.port=3322   -Dcom.sun.management.jmxremote.ssl=false   -Dcom.sun    .management.jmxremote.authenticate=false "

JAVA_OPTS=" $JMX_CONF "
```

barnabycourt:
@nguyenfilip We will need a secondary PR to update the developer docs for candlepinproject.org on how to set up & use the JMX monitoring of the cache.

mstead:
ACK.  @barnabycourt if you didn't find anything else, I'll merge this in.

PR #1337, opened 2016-09-01
2LC for Content and Product

nguyenfilip:
@mstead This PR introduces 2LC for Content, Product and also short (5 seconds cache) for rules query

### DIFF
--- a/server/src/main/java/org/candlepin/cache/CandlepinCacheRegions.java
+++ b/server/src/main/java/org/candlepin/cache/CandlepinCacheRegions.java
@@ -1,0 +1,27 @@
+/**
+ * Copyright (c) 2009 - 2012 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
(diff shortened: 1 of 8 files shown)


nguyenfilip:
retest this please

nguyenfilip:
retest this please

mstead:
Caching seems to be working fine when accessed through the ProductResource. However, I tested with manifest import and the cache hits/misses do not seem to line up.

I imported a manifest on a clean DB with no test data. The manifest had 4 different products. During the refresh pools routine, I step through the code that handles the incoming products while inspecting the cache stats. On initial import, I do not see any hit/miss count increase at all (remain at 0). Next I re-import with the MANIFEST_SAME override and do the same again. I see 2 misses in the cache and that's it... Despite product lookups happening 4 times in the code with 4 different products.

I'm looking into this further.

mstead:
ACK - Additional work will be done in seperate PRs to address the issues found.

PR #1339, opened 2016-09-07
Ruby Client - Create Pool bug

nguyenfilip:
Provided products must be specified with productId not id.

providedProducts nad derivedProvidedProducts are specified to use ProvidedProduct class!
```
  @JsonProperty("providedProducts")
  @JsonProperty("derivedProvidedProducts")
```

The ruby client mentions :derived_products but that collection is not on Pool at all.


wottop:
ack

PR #1340, opened 2016-09-07
Product.locked is never updated

nguyenfilip:
ProductManager.applyProductChanges should update value of locked column for
existing product, based on the incoming product.

If (incorrectly) the value of locked isn't updated, then Products will be
always marked dirty even though they have never been changed (possible performance
problem)

### DIFF
--- a/server/src/main/java/org/candlepin/controller/ProductManager.java
+++ b/server/src/main/java/org/candlepin/controller/ProductManager.java
@@ -511,6 +511,10 @@ public class ProductManager {
             entity.setDependentProductIds(update.getDependentProductIds());
         }
 
+        if (update.isLocked() != null) {
+            entity.setLocked(update.isLocked());
+        }
+
         return entity;
     }
 
--- a/server/client/ruby/hostedtest_api.rb
+++ b/server/client/ruby/hostedtest_api.rb
@@ -104,6 +104,43 @@ module HostedTest
     end
   end
 
+  def add_batch_content_to_product(owner_key, product_id, content_ids, enabled=true)
+    if is_hosted?
+      data = {} 
+      content_ids.each do |id| 
+        data[id] = enabled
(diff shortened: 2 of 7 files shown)


nguyenfilip:
@Ceiu can you please review this one?

Ceiu:
ACK

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -1290,10 +1290,6 @@ public boolean isChangedBy(ProductData dto) {
             return true;
         }
 
-        if (dto.isLocked() != null && !dto.isLocked().equals(this.locked)) {
```
Ceiu: We still want this check in place

server/src/main/java/org/candlepin/controller/ProductManager.java
```diff
@@ -511,6 +511,8 @@ private Product applyProductChanges(Product entity, ProductData update, Owner ow
             entity.setDependentProductIds(update.getDependentProductIds());
         }
 
+        entity.setLocked(update.isLocked());
```
Ceiu: Just remembered, we need a null check on update.isLocked, in the event there's no change to the value.

```
if (update.isLocked() != null) {
    entity.setLocked(update.isLocked());
}
```

nguyenfilip:
@Ceiu The inital plan to fix tests hasnt worked out. In the end I had to go with more hybrid approach. There are some tests (client_v1_size_spec) that legitimately require to modify product content in hosted mode throughout the test. So my initial plan to create content/product BEFORE the test starts wouldn't work here.

I kept a Trello card on a backlog to change this to more pure solution. The purest solution is to create/modify product/content in memory and have Candlepin pull it during refresh pools. But thats not how HostedTestSubscription* classes were initially designed.

PR #1341, opened 2016-09-12
Lock the hypervisor consumer earlier to avoid over entitlement (2.x)

cnsnyder:
This is the same fix as in #1333 except against master.


wottop:
I am good with this. nguyenfilip, do you have anything else?

nguyenfilip:
No, looks good!

PR #1343, opened 2016-09-14
1370508: Custom products in Activation key are not subscribed when service_level is set 

pjonastik:


wottop:
can you go ahead an merge the commits?

pjonastik:
@wottop I've merged commits.

server/spec/activation_key_spec.rb
```diff
@@ -25,6 +25,68 @@
...
     @activation_key['id'].should_not be_nil
   end
+ 
+  #SLA = Service Level Agreement
+  it 'should assign custom product without SLA set even when SLA is set on activation key' do
+    custom_product = create_product('testp1','testp1', {:custom => true})
+    sub_custom_product = @cp.create_subscription(@owner['key'], custom_product['id'], 30)
+
+    # creating in order to be able set SLA on activation key
+    premium_product = create_product('testp2', 'testp2', {:attributes => {:support_level => 'Premium'}})
```
pjonastik: There are two products: with SLA set = premium_product and without SLA set = custom_product. 
In test you can see that only custom_product is assigned to activation_key. 
Premium_product exists (as comment above explain) because of setting SLA on activation key. It's not possible set SLA on activation key to 'Premium' value without creation of subscription of premium_product.

wottop: Good point

wottop:
ack

PR #1344, opened 2016-09-14
Revert of cp_consumer_facts_lower

nguyenfilip:
Intent of this PR is to revert part of #1300.

Initially in 62d36c29640c50ea14779d22f6b5559c6f78bfd9 I have changed
implementation of case insensitive search to improve performance while querying.
However cp_consumer_facts table usually contains a lot of data and the migration
step thus takes a lot of time. This commit reverts the changes for the
cp_consumer_facts_lower that were introduced in the above mentioned commit

### DIFF
--- a/server/src/main/java/org/candlepin/model/Consumer.java
+++ b/server/src/main/java/org/candlepin/model/Consumer.java
@@ -170,13 +170,6 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
     @Cascade({org.hibernate.annotations.CascadeType.ALL})
     private Map<String, String> facts;
 
-    @ElementCollection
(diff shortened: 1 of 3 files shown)


vritant:
@nguyenfilip what do you think of following steps:
1. add a _lower table, but use the usual facts table for search *if*  the _lower table is not populated for that consumer.
2. populate the _lower table for future check ins ( or / and ) create an async task that populates cp_consumer_facts_lower for consumers

vritant:
after talking to @nguyenfilip , there is a chance this PR wont be merged, this is just in case.

vritant:
@barnabycourt this looks good and can be merged if needed.

PR #1345, opened 2016-09-21
Refactored pool and product attribute storage and access

Ceiu:
- PoolAttribute and ProductAttribute have been removed in favor of
  storing attributes as a key/value map on Pool and Product,
  respectively
- The lowercase columns of attribute values has been removed in
  favor of a better primary key for the attribute tables
- Several methods have been removed entirely as they no longer
  made sense with the removal of the attribute model objects
- ProductData has been updated to use attribute maps
- Attribute validation has been refactored out to the PropertyValidator,
  which handles attribute validation in a more abstract method than
  what was previously done in the product curator
- The query backing the PoolCurator.listAvailableEntitlementPools method
  has been completely rewritten to use as few subqueries as possible,
  which should have a notable positive impact on performance
- The attribute setters in Pool, Product and ProductData have been
  updated to allow null and empty values to exist for an attribute
- Updated the rules to use the new attribute structure
- Removed unnecessary value indexes from the attribute tables
- Added a new property validator attribute key and value length
  checks
- Fixed a bug in CandlepinPoolManager that generated unnecessary
  warnings during refresh
- Added the new CandlepinAttributeDeserializer to handle processing
  pool and product attributes provided in either map/dict style, or
  the old collection-of-objects style
- Added additional JSON tests for the above deserializer


Ceiu:
Before this is merged, one feature we lose by changing to this branch is that we can no longer search on attribute keys with wildcards. Values still accept them, but the keys themselves do not.

server/src/main/java/org/candlepin/model/PoolFilterBuilder.java
```diff
@@ -218,56 +227,105 @@ private DetachedCriteria createPoolAttributeCriteria(String attribute, List<Stri
...
 
         // We don't want to match Product Attributes that have been overridden
-        DetachedCriteria overridden = DetachedCriteria.forClass(PoolAttribute.class, "PoolAttrI")
-            // If we're using wildcards in the name, we should block exact matches
-            .add(Restrictions.eqProperty("PoolAttrI.name", "ProdAttrI.name"))
-            .add(Restrictions.eqProperty("PoolI.id", "PoolAttrI.pool.id"))
-            .setProjection(Projections.property("PoolAttrI.pool.id"));
-        subquery.add(Subqueries.notExists(overridden));
+
+        // TODO: FIXME: This is broken, the entire filter builder needs to be rewritten to properly
```
barnabycourt: Does this need to be fixed?

awood:
This code looks like it's going from 

```
[{
  'name' => 'ownerFooFactor',
  'value' => '9000'
}]
```

to

```
{ 'ownerFooFactor' => '9001' }
```

I am pretty sure that's going to break subscription-manager.  Subscription-manager builds the first construct into a map internally, but it's definitely going to be looking for a list of maps with keys `name` and `value`.  See [jsonwrapper.py](https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/jsonwrapper.py#L55)

Ceiu:
@awood I suppose we could write a serializer to also generate the old object JSON

server/src/main/java/org/candlepin/model/PoolFilterBuilder.java
Ceiu: Technically no, it's no longer used. But it should be fixed/removed as part of the overarching cleaning that needs to happen in PoolCurator and EntitlementCurator.

awood:
> @awood I suppose we could write a serializer to also generate the old object JSON

Yeah, I think we're going to have to do that or else clients will just break when they introspect product attributes.

barnabycourt:
I added the in progress flag until json serializer has been added to support the old format

server/src/main/java/org/candlepin/model/Pool.java
```diff
@@ -544,107 +593,186 @@ public void setOrderNumber(String orderNumber) {
...
-     * false otherwise.
+     * @return
+     *  the value set for the given attribute, or null if the attribute is not set
      */
-    public boolean attributeEquals(String key, String expectedValue) {
-        String val = getAttributeValue(key);
-        if (val != null && val.equals(expectedValue))  {
-            return true;
+    @XmlTransient
+    public String getAttribute(String key) {
```
awood: Agreed.  It pains me to see things so wordy, but given Attributes have both a name and value, it's probably the right call.  Unless of course `getAttribute` returned an Attribute object which I don't think it really should.

Ceiu: The reason "value" was dropped was due to redundancy. Attributes are nothing but key/value pairs, and if you already have the key (name), there's nothing to fetch other than the value.

Adding it back in doesn't really pain me any, aside from some minor inconsistencies we'll have:
- setAttribute(key, value)
- getAttributeValue(key)
- hasAttribute(key)
- removeAttribute(key)
- clearAttributes()

barnabycourt:
Adding back the in progress as I'd like to get a performance run on this branch before merging.

server/src/main/resources/rules/rules.js
```diff
@@ -188,22 +201,51 @@ function createPool(pool) {
...
+            }
+        }
+
+        return false;
+    };
+
+    pool.hasAttribute = function(attrName) {
+        // TODO: This isn't accurate. We consider an attribute to be present if it's set, even if
+        // it has a null value. We should update this method accordingly.
+        // return this.attributeIsPresent(this.attributes, attrName) || this.attributeIsPresent(this.productAttributes, attrName);
```
Ceiu: Nah. The behavior is still incorrect here, and the comment just shows what the correct behavior should be if/when we get around to updating the rules to adhere to how we treat attributes in Candlepin. The rules treat null values the same as "not present," and they merge pool and product attributes in a prioritized method, depending on which you call. Since fixing all of that is a task in itself, I opted to just comment on the broken behavior and leave it.

nguyenfilip:
retest this please

Ceiu:
With the addition of the serializer for ensuring the old format is used when writing attributes, all of the rules changes can be reverted.

Additionally, there needs to be significant cleanup made to the pool lookup code before this is merged in. Merging it in as-is will result in a massive amount of dead code being left in the code base in a non-obvious way.

Ceiu:
retest this please

Ceiu:
retest this please

barnabycourt:
@vritant how did this branch work out under the 1 hour test as compared to master?

barnabycourt:
I'd want to know why some of the calls are running a half sec to 5 seconds slower as compared to master from Dec 1

Ceiu:
Not sure if this was discussed already, but it looks like a couple of things were at play there:
(1) primarily, the performance test was running against the wrong branch
(2) apparently there's some variability in a few of the tests which can cause a swing of +/- 5s overall

It looks like it's around the same performance as the Dec 1 master.

Ceiu:
Rebased (again) against the latest master.

vritant:
@awood @barnabycourt @wottop is this ready for merge?

awood:
@vritant ✔️ from me

vritant:
remove default mark

vritant:
retest this please

barnabycourt:
@vritant, @wottop  What is your take? Are you happy with where this is?

wottop:
good from my end.

Ceiu:
Rebased against master... again.

Ceiu:
retest this please

Ceiu:
retest this please

Ceiu:
retest this please

vritant:
I am merging this PR even though I did not review it, because I think this has a problem of too many reviewers but no one wants to be trigger happy :) 
I apologize if that is a mistake, I would love to fix anything that goes wrong if that is so.

PR #1346, opened 2016-09-26
Ensure debug symbols make it into compiled code.

awood:
It looks like the class files that our in our official builds do not
have debug symbols in them.  No debug symbols saves some space but also
means that Support can't install a Satellite instance and then attach to
a running Candlepin with jdwp.

I have a suspicion that MEAD sets debug=false in an environment
configuration somewhere because the debug symbols are added when I
compile locally.  Setting the values explicitly should ensure the
symbols are left in.

### DIFF
--- a/buildfile
+++ b/buildfile
@@ -226,6 +226,7 @@ define "candlepin" do
   project.version = pom_version(path_to('pom.xml'))
   manifest["Copyright"] = "Red Hat, Inc. #{Date.today.strftime('%Y')}"
 
+  compile.using(:javac, :debug => true)
   compile.options.target = '1.6'
   compile.options.source = '1.6'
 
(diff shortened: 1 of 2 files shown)


awood:
candlepin-bot, retest this please

Ceiu:
ACK

kahowell:
@awood. Same for me. `buildr clean test=no package` fails for me with

    ...
    Packaging candlepin
    Packaging candlepin-2.0.jar
    Buildr aborted!
    Errno::ENOENT : No such file or directory - /home/khowell/candlepin/candlepin/target/classes

Previous commits via git bisect built fine w/:

    Packaging candlepin
    Packaging candlepin-2.0.zip
    Packaging gutterball-2.0.5.war
    Packaging candlepin-api-2.0.19.jar
    Packaging candlepin-certgen-2.0.19.jar
    Packaging candlepin-2.0.19.war
    Completed in 36.887s

awood:
Ah.  That's because you're building from the very top level and the buildfile now thinks that the toplevel project is something that can be compiled.  I will fix it.

PR #1347, opened 2016-09-28
Disable autobind at Owner level (candlepin-0.9.54)

mstead:
**Reviewers please test the scenarios listed [here](https://docs.google.com/document/d/1i0SJB2zX6QWXXa_YMEC1AIktth_44Ml5tBYLykTQWnI/edit)!**

Added autobindDisabled field/column to Owner
* Added the field to the Owner object
* Added utility method to consider null value when
  asking if autobind is disabled for the Owner.
* Added a change set that adds the new column to
  the cp_owner table.

Add ability to update Owner.autobindDisable via API
* PUT owners/:key will now update the autobindDisable
  field on the targeted Owner.

Throw exception on autobind if disabled on owner
* Allows for determining precise impact points and
  allows for customizing the resulting behavior.
* In most cases this results in a BadRequestException
  thrown by the caller.

Block hypervisor checkins when autobind is disabled.
* Done so that migrations will not occur and entitlements
  will remain unchanged.
* One edge case here is when a migration happened before
  autobind was disabled. When a guest checks in in this
  state, all host related entitlements WILL be removed,
  but an attempt is not made to autoheal the guest.

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -275,6 +275,10 @@ class Candlepin
     end
   end
 
+  def heal_owner(owner_key)
+    post "owners/#{owner_key}/entitlements"
+  end
+
   def create_user(login, password, superadmin=false)
     user = {
       'username' => login,
@@ -895,12 +899,16 @@ class Candlepin
     return get("/owner/#{owner_key}/activation_keys")
   end
 
-  def create_activation_key(owner_key, name, service_level=nil)
+  def create_activation_key(owner_key, name, service_level=nil, autobind=false)
     key = {
       :name => name,
     }
     if service_level
       key['serviceLevel'] = service_level
+    end
+
+    if autobind
+      key['autoAttach'] = true
     end
     return post("/owners/#{owner_key}/activation_keys", key)
   end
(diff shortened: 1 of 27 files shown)


Ceiu:
ACK

My only question is whether or not the exception needs to be a checked exception. The work for the current callers has already been done, but any calls using those branches going forward will have to be aware of the disabled autobind feature, for better or worse.

mstead:
> My only question is whether or not the exception needs to be a checked exception. The work for the current callers has already been done, but any calls using those branches going forward will have to be aware of the disabled autobind feature, for better or worse.

@Ceiu This was something that I was thinking about as well. When implementing the feature, I made it a checked exception so that I could better identify the code that was going to be affected. I didn't end up changing it to an unchecked.

I could make the exception extend BadRequestException so that the handling at the Resource layer is transparent. Thoughts?

Ceiu:
I would prefer the exception /not/ extend BadRequestException since it originates from within the controller, away from the request handler, and I feel there should be a very clear and distinct line between the resources and the backend.

There's something about the exception itself that irks me, but I feel like doing it differently to avoid it would be more work and more brittle without a major rewriting of these parts.

PR #1348, opened 2016-09-28
Removed in-place updating for product and content

Ceiu:
- Product and Content entities are now treated as immutable objects
  by their respective managers. Every chage made to either of these
  types of entities will now result in new entities being created,
  even in instances when fewer than two owners are using the entity.
- Updated unit and spec tests accordingly, and added missing tests
  for verifying the new behavior


Ceiu:
retest this please

wottop:
ack

PR #1350, opened 2016-09-30
Added an argument for specifying the Java version in cp-test

Ceiu:
- Added the -j argument to cp-test for specifying a Java
  version to use for building and running Candlepin
- The candlepin-base container now automatically installs
  Java 1.6, 1.7 and 1.8, as well as the current version
  (which happens to be 1.8 at the moment)


wottop:
ack

PR #1351, opened 2016-10-04
1380766: Correct improper calculation of extensions delta.

awood:
The CRL extensions are converted to DER internally.  Previously the code
was accounting for the difference between the length of the byte arrays
and the change in the length in the number of bytes used to express the
length of the sequence.  Since the extensions were already converted to
DER, the change in the number of length bytes was included in the raw
difference already.

This patch removes the double-counting of bytes that was occurring
previously.


wottop:
ack

PR #1352, opened 2016-10-04
Disable autobind at Owner level (2.0)

mstead:
**Reviewers please test the scenarios listed [here](https://docs.google.com/document/d/1i0SJB2zX6QWXXa_YMEC1AIktth_44Ml5tBYLykTQWnI/edit)!**
    
    Added autobindDisabled field/column to Owner
    * Added the field to the Owner object
    * Added utility method to consider null value when
      asking if autobind is disabled for the Owner.
    * Added a change set that adds the new column to
      the cp_owner table.

    Add ability to update Owner.autobindDisable via API
    * PUT owners/:key will now update the autobindDisable
      field on the targeted Owner.

    Throw exception on autobind if disabled on owner
    * Allows for determining precise impact points and
      allows for customizing the resulting behavior.
    * In most cases this results in a BadRequestException
      thrown by the caller.

    Block hypervisor checkins when autobind is disabled.
    * Done so that migrations will not occur and entitlements
      will remain unchanged.
    * Applies to both sync and async APIs.
    * One edge case here is when a migration happened before
      autobind was disabled. When a guest checks in in this
      state, all host related entitlements WILL be removed,
      but an attempt is not made to autoheal the guest.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -38,6 +38,7 @@ import org.candlepin.common.paging.Page;
 import org.candlepin.common.paging.PageRequest;
 import org.candlepin.common.paging.Paginate;
 import org.candlepin.config.ConfigProperties;
+import org.candlepin.controller.AutobindDisabledForOwnerException;
 import org.candlepin.controller.Entitler;
 import org.candlepin.controller.PoolManager;
 import org.candlepin.model.CdnCurator;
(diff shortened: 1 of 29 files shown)


Ceiu:
ACK

Set autobindDisabled via API -- VERIFIED
CLI autobind fails when disabled for owner on register and manual attach -- VERIFIED
rhsmcertd autoheal fails when autobind is disabled  -- VERIFIED
Hypervisor Checkins -- VERIFIED


Will wait for @wottop to merge, since I didn't notice him jump on this as well until after I started re-running the checks on it.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1490,9 +1502,15 @@ else if (hasPoolQuantities) {
             entitlements = entitler.bindByPoolQuantity(consumer, poolIdString, quantity);
         }
         else {
-            AutobindData autobindData = AutobindData.create(consumer).on(entitleDate)
-                .forProducts(productIds).withPools(fromPools);
-            entitlements = entitler.bindByProducts(autobindData);
+            try {
+                AutobindData autobindData = AutobindData.create(consumer).on(entitleDate)
+                        .forProducts(productIds).withPools(fromPools);
```
wottop: indent

wottop:
What was the answer to the change of specific exceptions to generic exception?

server/src/main/java/org/candlepin/resource/ConsumerResource.java
mstead: Fixed.

mstead:
@wottop Style issue is fixed. Please merge when you get a chance.

wottop:
ack

PR #1353, opened 2016-10-04
Manifest import: refactoring

pjonastik:
I found some long methods and duplicate code so I did refactoring work around import manifest to make it more readable.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/OwnerResource.java
+++ b/server/src/main/java/org/candlepin/resource/OwnerResource.java
@@ -1054,52 +1054,16 @@ public class OwnerResource {
         @PathParam("owner_key") @Verify(Owner.class) String ownerKey,
         @QueryParam("force") String[] overrideConflicts, MultipartInput input) {
 
-        if (overrideConflicts.length == 1) {
-            /*
-             * For backward compatibility, look for force=true and if found,
-             * treat it just like what it used to mean, ignore an old manifest
-             * creation date.
-             */
-            if (overrideConflicts[0].equalsIgnoreCase("true")) {
-                overrideConflicts = new String [] { "MANIFEST_OLD" };
-            }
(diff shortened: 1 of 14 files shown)


server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1054,50 +1054,14 @@ public void importManifest(
...
-                if (entry.startsWith("filename")) {
-                    filename = entry.substring(entry.indexOf("=") + 2, entry.length() - 1);
-                    break;
-                }
-            }
-            File archive = part.getBody(new GenericType<File>() {
-            });
+            filename = extractFileNameFromHeaders(part);
+            File archive = part.getBody(new GenericType<File>() {});
             log.info("Importing archive " + archive.getAbsolutePath() +
```
Ceiu: Since you're doing a pretty significant refactor of this area, can you also fix the log lines to use substitution instead of concatenation?

```
log.info("Importing archive: {}", archive.getAbsolutePath());
```

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1125,10 +1089,102 @@ else if (overrideConflicts[0].equalsIgnoreCase("false")) {
...
+                overrideConflicts = new String [] {};
+            }
+        }
+        return overrideConflicts;
+    }
+
+    private void logConflictOverridesIfEnabled(String[] overrideConflicts) {
+        if (log.isDebugEnabled()) {
+            for (String s : overrideConflicts) {
+                log.debug("Forcing conflict if encountered: " + s);
```
Ceiu: Log string concatenation

Ceiu:
ACK; test failures are due to the 0.9.49 branch not supporting some of our Jenkins tasks.

Once the last log line is updated, this is good to go.

PR #1355, opened 2016-10-06
Disable vagrant port forward

barnabycourt:
Add the .vagrant directory to the .gitignore & disable the localhost port-forwarding now that the hostmanager plugin will automatically update your hosts file with candlepin.example.com


cnsnyder:
:+1:

PR #1356, opened 2016-10-06
1371009:  Need clearer error message when register with system name e…

pjonastik:
…xceeding max characters.

-adding unit test
-fix buildfile - repo of org.xnap.commons.i18n was moved

### DIFF
--- a/server/src/test/java/org/candlepin/resource/test/ConsumerResourceTest.java
+++ b/server/src/test/java/org/candlepin/resource/test/ConsumerResourceTest.java
@@ -75,7 +75,9 @@ import org.candlepin.test.TestUtil;
 import org.candlepin.util.ServiceLevelValidator;
 
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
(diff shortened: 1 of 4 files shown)


server/src/test/java/org/candlepin/resource/test/ConsumerResourceTest.java
```diff
@@ -521,6 +527,51 @@ public void testNullPerson() {
...
+        when(ctc.lookupByLabel(eq("system"))).thenReturn(cType);
+        when(up.canAccess(eq(o), eq(SubResource.CONSUMERS), eq(Access.CREATE))).
+            thenReturn(true);
+
+        consumerResource.create(c, up, null, ownerKey, null);
+    }
+
+    private String generateNameLongerThan255() {
+        String name255 =
+            "qwert12345qwert12345qwert12345qwert12345qwert12345" +
```
Ceiu: In the future, for values which are derived from a defined constant like this we should be generating a string rather than hard-coding it.

Ceiu:
ACK

Note that the tests failed in this branch strictly due to the age of the branch. Manually running the tests passed.

PR #1357, opened 2016-10-06
Fix double entitlement issue by lock host during guest autobind

cnsnyder:


cnsnyder:
@reviewer Ignore the missing rpmlint tests in jenkins

wottop:
ack

PR #1358, opened 2016-10-07
Added cursor support for pagination

Ceiu:
- Added the CandlepinQueryInterceptor, which processes CandlepinQuery
  instances and writes the output using a cursor, applying any paging
  specified in the request
- Several classes have been refactored or entirely replaced with the
  addition of the new interceptor
- Added the new TransformedCandlepinQuery decorator class, which
  transparently applies transformation to the results of a CandlepinQuery


Ceiu:
Note for reviewers: This PR is to merge into the cursor branch, not master.

PR #1359, opened 2016-10-12
Add debug symbols to Java projects and not the top project.

awood:
Addresses issue with #1346


kahowell:
FWIW, I also noticed that using at top level:

    compile.using(:debug => true)

builds/deploys properly (notice absence of `:javac`). (Not sure if the debug symbols appear in compiled classes... did not check).

awood:
buildbot, restest this please

PR #1360, opened 2016-10-13
Product entity cached by reference

nguyenfilip:
# Reference caching of Product

Reason for these changes is to improve performance of code paths that use
Pool.providedProducts and Pool.derivedProvidedProducts. The basic idea is that
Product entities are cached in key-value cache by reference. Thus, all the
threads of Candlepin share the same instances of Product entities. Furthermore,
upon first load, the Product instances that are about to be placed in the cached
are fully hydrated (all lazy loaded collections, such as Content, are loaded).
Underlying assumption, that allows us to do this change, is that Product
entities are fully immutable - so there is no chance that cached Product might
be come stale.

Most of the references to getProvidedProducts and getDerivedProvidedProducts
has been removed from the code. The lookups to provided products should be
instead made through ProductCurator. The methods on ProductCurator ensure
caching of fully hydrated Product objects.

The getters still remain on the Pool. This is necessary for some situations. For
example when a Pool is 'fabricated' using
CandlepinPoolManager.convertToMasterPool, the getters are the only reasonable way
to retrieve fabricated provided products.

Attributes Pool.providedProducts and Pool.derivedProvidedProducts are still
mapped as Hibernate collection. Reason for that is, that we need those
collection when creating the association between Pool and provided products. One
can argue this is more of a technical problem. In ideal situation, in the
future, we might want to remove these Hibernate colletion mappings and we might
want to create the links between Pool and provided products using an
intermediate entity between Pool or Product. Or some other mechanism that will
allow us to remove the mapped collections.

New serialization interceptor PoolSerializer is now used for all serializations
that require Pool to have provided products (REST calls, JS rules, HornetQ,
AMQP, etc.) The strategy to serialize provided products has two steps. First,
before a Pool is serialized, the serializer will fill in transient properties
that contain provided products. Then, in second step, the serialization
continues as before and the transient fields are serialized.

Methods provides() and providesDerived() that were on the Pool are now replaced
by ProductCurator.provides() and ProductCurator.providesDerived(). The new methods
on the curator are using direct database lookups, instead of traversing
collections.


nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
There is a random fail during rspec-postgresql and I think thats just a test failure. Created #1364 for that one

nguyenfilip:
retest this please

nguyenfilip:
When running the performance tests, there is one tweak that changes (highly possibly improves) the performance: disabling 2nd level cache. 

To disable 2LC, the easiest way is to comment out all the Cacheable and Cache annotations. E.g.: 

```java
\\@Cacheable(true)
\\@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
public class Product extends AbstractHibernateObject implements SharedEntity, Linkable, Cloneable {
  
```

```java
    //@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
    private List<ProductContent> productContent;
```

mstead:
@nguyenfilip This looks pretty good. The only thing that stood out to me is how the curator is bleeding into the Subscription DTO and the Pool.  Honestly I can't decide if it is a bad thing considering what it is doing. Matter of preference perhaps.

I still haven't put it through its paces. Working on that now.

nguyenfilip:
@mstead Yes I agree that curator in those places is not optimal. Its possible to refactor it out though. 

From some point of view it does make sense however. Today we don't have curator there, but still the data access happens by Hibernate  doing its lazy load 'magic' at runtime.

nguyenfilip:
@mstead One can argue that having curator there is just 'uncovering' what really happens during serialization etc. I am ok keeping it the way it is in the pr. But if you lean towards changing this, let me know and I will refactor. I don't mind.

server/src/main/resources/ehcache-stats.xml
```diff
@@ -1,15 +1,19 @@
...
-    maxBytesLocalHeap="300M"
+    maxEntriesLocalHeap="100000"
     >
 
+	<sizeOfPolicy maxDepth="100" maxDepthExceededBehavior="abort"/>
+
     <defaultCache 
+    	maxEntriesLocalHeap="100000"
         eternal="false" 
         timeToIdleSeconds="86400"
```
barnabycourt: Is a 24 hour expire time proper? Is there a way to make this configurable from the candlepin.conf file?

server/src/main/resources/ehcache.xml
```diff
@@ -1,15 +1,19 @@
...
-    maxBytesLocalHeap="300M"
+    maxEntriesLocalHeap="100000"
     >
 
+	<sizeOfPolicy maxDepth="100" maxDepthExceededBehavior="abort"/>
+
     <defaultCache 
+    	maxEntriesLocalHeap="100000"
         eternal="false" 
         timeToIdleSeconds="86400"
```
barnabycourt: Is a 24 hour expire time proper? Is there a way to make this configurable from the candlepin.conf file?

server/src/main/resources/ehcache.xml
```diff
@@ -1,15 +1,19 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <ehcache
-    maxBytesLocalHeap="300M"
+    maxEntriesLocalHeap="100000"
```
barnabycourt: Is there a way to make this ocnfigurable from the candlepin.conf file?

nguyenfilip:
retest this please

server/src/main/resources/ehcache-stats.xml
nguyenfilip: The documentation states its possible [link](http://www.ehcache.org/generated/2.10.2/html/ehc-all/#page/Ehcache_Documentation_Set%2Fco-cfgbasics_dynamically_changing_cache_config.html%23) but it doesn't work for me. Its possibly because we are using JCache provider.

What does work for me is that at deploy time its possible to modify 
```
/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/ehcache.xml
```

Would be good enough to document this on cp.org?  If not I will investigate more how to externalize the ehcache.xml or why dynamic cache config doesn't have effect at runtime.

server/src/main/resources/ehcache.xml
nguyenfilip: same answer as above

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -84,10 +140,107 @@ public Product lookupByName(Owner owner, String name) {
             .uniqueResult();
     }
 
-    public List<Product> listAllByUuids(Collection<? extends Serializable> uuids) {
-        return this.listByCriteria(
-            this.createSecureCriteria().add(Restrictions.in("uuid", uuids))
-        );
+    public Set<Product> getPoolDerivedProvidedProductsCached(String poolId) {
```
nguyenfilip: Makes sense, I have just commited the changes

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -84,10 +140,107 @@ public Product lookupByName(Owner owner, String name) {
...
+        Set<String> uuids = getDerivedPoolProvidedProducts(poolId);
+        return getProductsByUuidCached(uuids);
+    }
+
+    public Set<Product> getPoolProvidedProductsCached(String poolId) {
```
nguyenfilip: fixed

server/src/main/resources/ehcache.xml
nguyenfilip: answer above

nguyenfilip:
@barnabycourt I measured the memory consumption with fabricated Products. Caching 3,300 products (thats the approx number of content currently in production), 3000 content sets and 69,300 attributes resulted in est. 30MB overhead (visible through yourkit) and I think that extrapolating this is fairly accurate. However, the memory consumption is very hard to even notice.

I have switched the Product reference cache to eternal

PR #1362, opened 2016-10-14
Awood/new liquibase fix

awood:
The new version of Liquibase (3.5.3) that I'm building for Fedora failed when trying to deploy with the cert column as a blob instead of a bytea.  Other cert columns that we've had were bytea so this PR corrects the inconsistency.

When this gets merged, we need to send out a note to everyone informing them to drop and rebuild their database since the modification in the column element in the changeset alters the checksum on the changeset and makes Liquibase angry (since Liquibase now things alterations have occurred outside of its knowledge).


Ceiu:
retest this please

PR #1367, opened 2016-10-25
1388201: Remove unique constraint on content keys; replace with index.

awood:


Ceiu:
We actually don't need this constraint change in master, as the cp2_content table does not have the constraint on content. We can merge this in, but it won't make a difference either way, I believe.

Ceiu:
Actually, since this change has been merged in on the older branches, it's probably best to make the change here as well for consistency.

PR #1371, opened 2016-10-26
1314042: bug caused by import and undo manifest run in parallel

pjonastik:
- we don't want support asynch import and undo manifest on these version, but candlepin shouldn't fail in this scenario.
- I wrote reproducer called 'server/bin/concurrency-import-and-undo-manifest.rb', which run importManifest (with updated manifest on brandings) and undoManifest in parralel and generated these errors:
  1) deadlock on cp_export_metadata,
  2) RefrehPoolJob broken
- I fix these problems by locking owner in import, undo manifest and also in RefreshPoolJob.
- this changes broke tests so I changed these as well and made them little bit readable.
- Note: when I experimented with reproducer I've never got same error like in bugzilla, but this should be covered by this fix too.
- Note2: Checkstyle plugin in eclipse showed me old checkstyle indentation error (I18nProvider), so I fixed it as well.

### DIFF
--- a/server/src/main/java/org/candlepin/model/OwnerCurator.java
+++ b/server/src/main/java/org/candlepin/model/OwnerCurator.java
@@ -27,6 +27,8 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 
+import javax.persistence.LockModeType;
+
 /**
  * OwnerCurator
  */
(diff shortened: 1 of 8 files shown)


pjonastik:
retest this please

server/src/main/java/org/candlepin/model/OwnerCurator.java
```diff
@@ -66,6 +68,16 @@ public Owner lookupByKey(String key) {
             createSecureCriteria().add(Restrictions.in("key", keys)));
     }
 
+    @Transactional
+    //TODO: added annotation only for test purpose, working on solution
```
barnabycourt: Is there more work you are planning for this TODO?

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -327,6 +327,7 @@ public void deleteOwner(@PathParam("owner_key") String ownerKey,
         Event e = eventFactory.ownerDeleted(owner);
 
         cleanupAndDelete(owner, revoke);
+        //above revoke is useless, revokeAllEntitlements is called anyway
```
barnabycourt: if the revoke is useless then please mark the query parameter as deprecated

barnabycourt: Also, if it is not used, please remove it from the cleanupAndDelete() method

barnabycourt:
retest this please

server/src/main/java/org/candlepin/resource/OwnerResource.java
pjonastik: reworked

PR #1375, opened 2016-11-01
Disable the Hibernate 2nd level cache and remove annotations

barnabycourt:
Disable the hibernate 2nd level cache as it is competing for memory with the JSR-107 cache and does not appear to be providing significant speedup.

### DIFF
--- a/server/src/main/resources/META-INF/persistence.xml
+++ b/server/src/main/resources/META-INF/persistence.xml
@@ -19,8 +19,6 @@
             <property name="hibernate.c3p0.min_size" value="5" />
             <property name="hibernate.c3p0.max_size" value="20" />
             <property name="hibernate.c3p0.timeout" value="300" />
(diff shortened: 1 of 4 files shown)


server/src/main/resources/META-INF/persistence.xml
```diff
@@ -19,8 +19,8 @@
             <property name="hibernate.c3p0.min_size" value="5" />
             <property name="hibernate.c3p0.max_size" value="20" />
             <property name="hibernate.c3p0.timeout" value="300" />
-            <property name="hibernate.cache.use_second_level_cache" value="true" />
-            <property name="hibernate.cache.use_query_cache" value="true" />
+            <property name="hibernate.cache.use_second_level_cache" value="false" />
+            <property name="hibernate.cache.use_query_cache" value="false" />
```
mstead: I think these settings default to false. I'd vote to remove them.

mstead:
Other than the comment on the config file, it looks good to me.

PR #1377, opened 2016-11-03
Qpid Spec Test Support

nguyenfilip:
New -q switch for deploy script that will recreate queue 'allmsg' which is a
special queue for testing purposes. -a switch will also cause autoconf
to enable AMQP in candlepin config.

Docker image candlepin-base now installs qpid and starts it using supervisord.

cp-test script has new -q switch that uses guterrball configure-qpid.sh
script, to setup qpid (generate SSL keys, create event exchange) etc.

I also had to modify configure-qpid.sh to be aware of supervisorctl

You can now write spec tests that assert Qpid queue. There is a new
CandlepinQpid helper class.

### DIFF
--- a/server/bin/deploy
+++ b/server/bin/deploy
@@ -133,6 +133,7 @@ usage: deploy [options]
 OPTIONS:
   -f          force cert regeneration
   -g          generate database
+  -q          Qpid enabled. Also deploys 'allmsg' queue used by spec tests
   -t          import test data
   -T          import minimal test data, some owners, users, and roles
   -H          include test resources for hosted mode
@@ -186,11 +187,12 @@ init
 
 DBUSER="candlepin"
 
-while getopts ":fgtTHolmavd:" opt; do
+while getopts ":fqgtTHolmavd:" opt; do
     case $opt in
         f  ) FORCECERT="1" ;;
         g  ) GENDB="1";;
         H  ) HOSTEDTEST="hostedtest";;
+        q  ) QPID="qpid";;
         t  ) TESTDATA="1";;
         T  ) TESTDATA="MIN";;
         o  ) USE_ORACLE="1";;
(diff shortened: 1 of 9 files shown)


nguyenfilip:
retest this please

nguyenfilip:
To install the deps from new Gemfile you need at least the following dnf package qpid-proton-c-devel

nguyenfilip:
retest this please

nguyenfilip:
@mstead would you mind being a second reviewer on this? Since you wrote most of the Qpid stuff, I would like your perspective.

mstead:
> @mstead would you mind being a second reviewer on this? Since you wrote most of the Qpid stuff, I would like your perspective.

Sure! I'll take a look as soon as I can.

server/bin/deploy
```diff
@@ -186,11 +186,12 @@ init
 
 DBUSER="candlepin"
 
-while getopts ":fgtTHolmavd:" opt; do
+while getopts ":fqgtTHolmavd:" opt; do
     case $opt in
         f  ) FORCECERT="1" ;;
         g  ) GENDB="1";;
         H  ) HOSTEDTEST="hostedtest";;
+        q  ) QPID="qpid";;
```
dsarman: This new option should be also documented in the help string.

server/bin/deploy
```diff
@@ -286,6 +287,22 @@ else
...
 
+
+if [ "$QPID" == "qpid" ]; then
+    info_msg "Enable Qpid"
+    QPID="qpid=yes"
+
+    #Setup testing qpid queue allmsg, this will be needed by Qpid spec tests
+    QPID_KEYS_DIR=$PROJECT_DIR/../gutterball/bin/qpid/keys
+    # Delete the allmsg first, just to make the script indempotent
+    qpid-config -b amqps://localhost:5671 --ssl-certificate $QPID_KEYS_DIR/qpid_ca.crt --ssl-key $QPID_KEYS_DIR/qpid_ca.key --force del queue allmsg || true
```
dsarman: This commands output should be silenced, because it prints confusing error message when the queue does not exist (which is not an error).

server/bin/deploy
nguyenfilip: done

server/bin/deploy
nguyenfilip: redirected to /dev/null

server/bin/deploy
```diff
@@ -286,6 +288,23 @@ else
...
+    QPID_KEYS_DIR=$PROJECT_DIR/../gutterball/bin/qpid/keys
+    # Delete the allmsg first, just to make the script indempotent. The error messages produced during
+    # deletion are silenced
+    qpid-config -b amqps://localhost:5671 --ssl-certificate $QPID_KEYS_DIR/qpid_ca.crt --ssl-key $QPID_KEYS_DIR/qpid_ca.key --force del queue allmsg 2 > /dev/null || true
+    qpid-config -b amqps://localhost:5671 --ssl-certificate $QPID_KEYS_DIR/qpid_ca.crt --ssl-key $QPID_KEYS_DIR/qpid_ca.key add queue allmsg --durable
+    qpid-config -b amqps://localhost:5671 --ssl-certificate $QPID_KEYS_DIR/qpid_ca.crt --ssl-key $QPID_KEYS_DIR/qpid_ca.key bind event allmsg "#"
+else
+    QPID=""
+fi
+
```
mstead: In the case where we have already deployed with qpid active, but then want to deploy without, will the allmsg queue have any impact on the new deployment? We've seen in the past that the gutterball queue was choking candlepin over time.

I'm wondering if there is a need to delete the allmsg queue if it is not in use.

nguyenfilip: @mstead good point. What I can do is to try delete the queue every time we use deploy script. In case it fails just silently ignore that failure. What do you think? Another approach might be to create the queue non-durable. Such queue would be deleted upon qpid restartup

I am fine with both approaches

nguyenfilip: fixed

nguyenfilip:
@mstead @dsarman this one should be ready to go

PR #1379, opened 2016-11-04
Qpid Reconnect to Broker

nguyenfilip:


server/src/main/java/org/candlepin/audit/QpidConnection.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * This indempotent method will establish connection to qpid broker. Per JMS standard
+     * it must recreate all JMS objects such as Connection, TopicSession, TopicPublisher.
+     * @throws Exception errors during connecting to the Broker
+     */
+    public void connect() throws Exception {
+        synchronized (connectionLock) {
+            if (connectionStatus == STATUS.CONNECTED) {
+                log.info("Already connected to Qpid Broker, ignoring this connect request");
```
dsarman: Are we really ignoring the request? The method seems to go on normally except for the log message.

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -336,8 +336,8 @@ private ConfigProperties() {
             this.put(AMQP_TRUSTSTORE, "/etc/candlepin/certs/amqp/candlepin.truststore");
             this.put(AMQP_TRUSTSTORE_PASSWORD, "password");
 
-            this.put(AMQP_CONNECTION_RETRY_INTERVAL, "10"); // Every 10 seconds
-            this.put(AMQP_CONNECTION_RETRY_ATTEMPTS, "12"); // Try for 2 mins (10s * 12)
+            this.put(AMQP_CONNECTION_RETRY_INTERVAL, "10"); // Every 5 seconds
```
dsarman: Is the value 10 correct? According to comment it should be 5.

server/src/main/java/org/candlepin/audit/QpidConnection.java
nguyenfilip: Nice catch. Will try to add some test for this

nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
retest this please

mstead:
One of the issues that I've seen in the past (that I don't **think** we had a solution for) is that when a time out occurs and all connection retries have happened, an exception is thrown by Qpid and there is no way to get the connection back, short of a tomcat restart. Does this PR address this as well?

mstead:
> @nguyenfilip One of the issues that I've seen in the past (that I don't think we had a solution for) is that when a time out occurs and all connection retries have happened, an exception is thrown by Qpid and there is no way to get the connection back, short of a tomcat restart. Does this PR address this as well?

I think this might be handled in another one of your PRs for QPID.

server/src/main/java/org/candlepin/audit/QpidConnection.java
nguyenfilip: I will remove this check all together. I found that its better to just let the reconnect to go through. The connectionStatus variable value is not reliable in concurrent environment.

server/src/main/java/org/candlepin/audit/QpidConnection.java
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Connection to a Qpid Broker. It can be in two states: CONNECTED or DOWN.
+ * The class has ability to reconnect to the broker.
+ *
+ * This class also orchestrates all the initial configuration of the
+ * Qpid connection
+ * @author fnguyen
+ *
+ */
+public class QpidConnection {
```
mstead: Perhaps javadoc the public methods for this class.

nguyenfilip: thanks, added

mstead: :+1:

server/src/main/java/org/candlepin/audit/QpidConnection.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+    }
+
+    public Connection newConnection() throws JMSException {
+        log.debug("creating connection");
+        Connection conn = null;
+        conn = connectionFactory.createConnection();
+        conn.start();
+        return conn;
```
nguyenfilip: I don't think this is a good idea. I would rather have false positive (connectionStatus = CONNECTED) then having DOWN incorrectly.

This is kind of tricky to get right, multiple threads might try to reconnect and I would rather set connectionStatus = DOWN only when I am absolutely sure the send failed.

nguyenfilip:
@mstead I think this can go in if you don't have more concerns

server/src/main/java/org/candlepin/audit/QpidConnection.java
mstead: Ok, makes sense.

mstead:
> @mstead I think this can go in if you don't have more concerns

Looks good to me.

PR #1389, opened 2016-11-14
Broker Status

nguyenfilip:
Qpid can be in three states that Candlepin needs to distinguish. First state is
'Qpid down' which is when connection to the Qpid is not available. Second state
is 'Flow Stopped'. That situation occurs when one of the queues bound to the
event exchange is flow stopped (see Flow Control in Qpid docs). Last is normal
happy 'connected' state in which Qpid is available.


nguyenfilip:
retest this please

mstead:
retest this please

nguyenfilip:
@mstead rebased, ready for review

PR #1391, opened 2016-11-15
Golden Ticket

wottop:
Content access certificates generated for consumers when org has indicator
The entitlment and signature blocks are stored by org and environment
The certificates contain the path for org/environment in the encoded cert but
  all of the possible content in the entitlment block.

I will keep looking at adding unit tests where they seem appropriate.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/OwnerManager.java
+++ b/server/src/main/java/org/candlepin/controller/OwnerManager.java
@@ -17,6 +17,7 @@ package org.candlepin.controller;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.ConsumerCurator;
 import org.candlepin.model.Content;
+import org.candlepin.model.ContentAccessCertificateCurator;
 import org.candlepin.model.Environment;
 import org.candlepin.model.EnvironmentCurator;
 import org.candlepin.model.ExporterMetadata;
@@ -26,6 +27,7 @@ import org.candlepin.model.ImportRecordCurator;
 import org.candlepin.model.Owner;
 import org.candlepin.model.OwnerContentCurator;
 import org.candlepin.model.OwnerCurator;
+import org.candlepin.model.OwnerEnvContentAccessCurator;
 import org.candlepin.model.OwnerProductCurator;
 import org.candlepin.model.PermissionBlueprint;
 import org.candlepin.model.PermissionBlueprintCurator;
@@ -33,6 +35,7 @@ import org.candlepin.model.Pool;
 import org.candlepin.model.Product;
 import org.candlepin.model.activationkeys.ActivationKey;
 import org.candlepin.model.activationkeys.ActivationKeyCurator;
+import org.candlepin.service.ContentAccessCertServiceAdapter;
 
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
@@ -61,6 +64,9 @@ public class OwnerManager {
     @Inject private OwnerContentCurator ownerContentCurator;
     @Inject private ContentManager contentManager;
     @Inject private OwnerCurator ownerCurator;
+    @Inject private ContentAccessCertServiceAdapter contentAccessCertService;
+    @Inject private ContentAccessCertificateCurator contentAccessCertCurator;
+    @Inject private OwnerEnvContentAccessCurator ownerEnvContentAccessCurator;
 
     @Transactional
     public void cleanupAndDelete(Owner owner, boolean revokeCerts) {
@@ -151,4 +157,25 @@ public class OwnerManager {
         ownerCurator.delete(owner);
     }
 
+    public void determineContentAccessCertState(Owner owner) {
+        if (!owner.isContentAccessModeDirty()) {
+            return;
+        }
+        if (owner.contentAccessMode()
+            .equals(ContentAccessCertServiceAdapter.DEFAULT_CONTENT_ACCESS_MODE)) {
+            contentAccessCertCurator.deleteForOwner(owner);
+        }
+        owner.setContentAccessModeDirty(false);
+        ownerCurator.merge(owner);
+    }
+
+    @Transactional
+    public void refreshOwnerForContentAccess(Owner owner) {
+        // we need to update the owner's consumers if the content access mode has changed
+        owner = ownerCurator.findAndLock(owner.getKey());
+        this.determineContentAccessCertState(owner);
+        // removed cached versions of content access cert data
+        ownerEnvContentAccessCurator.removeAllForOwner(owner.getId());
+        ownerCurator.flush();
+    }
 }
--- a/server/src/main/java/org/candlepin/controller/Refresher.java
+++ b/server/src/main/java/org/candlepin/controller/Refresher.java
@@ -40,6 +40,7 @@ public class Refresher {
 
     private CandlepinPoolManager poolManager;
     private SubscriptionServiceAdapter subAdapter;
+    private OwnerManager ownerManager;
     private boolean lazy;
     private UnitOfWork uow;
     private static Logger log = LoggerFactory.getLogger(Refresher.class);
@@ -48,9 +49,10 @@ public class Refresher {
     private Set<Product> products = Util.newSet();
 
     Refresher(CandlepinPoolManager poolManager, SubscriptionServiceAdapter subAdapter,
-        boolean lazy) {
+        OwnerManager ownerManager, boolean lazy) {
         this.poolManager = poolManager;
         this.subAdapter = subAdapter;
+        this.ownerManager = ownerManager;
         this.lazy = lazy;
     }
 
@@ -150,6 +152,7 @@ public class Refresher {
 
         for (Owner owner : this.owners.values()) {
             poolManager.refreshPoolsWithRegeneration(subAdapter, owner, lazy);
+            ownerManager.refreshOwnerForContentAccess(owner);
         }
     }
 
--- a/server/src/main/java/org/candlepin/model/CertificateSerialCurator.java
+++ b/server/src/main/java/org/candlepin/model/CertificateSerialCurator.java
@@ -29,6 +29,7 @@ import org.hibernate.criterion.Restrictions;
 import org.hibernate.criterion.Subqueries;
 
 import java.util.Collection;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
@@ -151,5 +152,27 @@ public class CertificateSerialCurator extends AbstractHibernateCurator<Certifica
      */
     public Collection<CertificateSerial> saveOrUpdateAll(Map<String, CertificateSerial> serialMap) {
         return this.saveOrUpdateAll(serialMap.values(), false, false);
+    }
+
+    @SuppressWarnings("unchecked")
+    public List<Long> listEntitlementSerialIds(Consumer c) {
+        List<Long> resultList = null;
+        String hql = "SELECT s.id" +
+            "    FROM EntitlementCertificate ec" +
+            "     JOIN ec.entitlement e" +
+            "     JOIN e.consumer c" +
+            "     JOIN ec.serial s" +
+            "     JOIN e.pool p" +
+            "    WHERE" +
+            "       c.id=:consumerId" +
+            "    AND" +
+            "       p.endDate >= :nowDate";
+        javax.persistence.Query query = this.getEntityManager().createQuery(hql);
+
+        resultList = (List<Long>) query
+            .setParameter("consumerId", c.getId())
+            .setParameter("nowDate", new Date())
+            .getResultList();
+        return resultList;
     }
 }
--- a/server/src/main/java/org/candlepin/model/Consumer.java
+++ b/server/src/main/java/org/candlepin/model/Consumer.java
@@ -143,6 +143,10 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
     @JoinColumn(name = "consumer_idcert_id")
     private IdentityCertificate idCert;
 
+    @OneToOne (fetch = FetchType.LAZY)
+    @JoinColumn(name = "cont_acc_cert_id")
+    private ContentAccessCertificate contentAccessCert;
+
     @ManyToOne
     @JoinColumn(nullable = false)
     @ForeignKey(name = "fk_consumer_consumer_type")
@@ -212,6 +216,11 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
     // WARNING: can't initialize to a default value here, we need to be able to see
     // if it was specified on an incoming update, so it must be null if no value came in.
     private Boolean autoheal;
+
+    // This is normally used from the owner setting. If this consumer is manifest then it
+    // can have a different setting as long as it exists in the owner's mode list.
+    @Column(name = "content_access_mode")
+    private String contentAccessMode;
 
     /**
      * Length of field is required by hypersonic in the unit tests only
@@ -287,6 +296,16 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
 
     public void setIdCert(IdentityCertificate idCert) {
         this.idCert = idCert;
+    }
+
+    @HateoasArrayExclude
+    @XmlTransient
+    public ContentAccessCertificate getContentAccessCert() {
+        return contentAccessCert;
+    }
+
+    public void setContentAccessCert(ContentAccessCertificate contentAccessCert) {
+        this.contentAccessCert = contentAccessCert;
     }
 
     /**
@@ -727,4 +746,12 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
     public boolean isDev() {
         return !StringUtils.isEmpty(getFact("dev_sku"));
     }
+    public String getContentAccessMode() {
+        return this.contentAccessMode;
+    }
+
+    public void setContentAccessMode(String contentAccessMode) {
+        this.contentAccessMode = contentAccessMode;
+    }
+
 }
--- a/server/src/main/java/org/candlepin/model/ContentAccessCertificate.java
+++ b/server/src/main/java/org/candlepin/model/ContentAccessCertificate.java
@@ -12,7 +12,6 @@
  * granted to use or replicate Red Hat trademarks that are incorporated
  * in this software or its documentation.
  */
-
 package org.candlepin.model;
 
 import com.fasterxml.jackson.annotation.JsonFilter;
@@ -33,18 +32,14 @@ import javax.xml.bind.annotation.XmlRootElement;
 import javax.xml.bind.annotation.XmlTransient;
 
 /**
- * Represents certificate used to identify a consumer
+ * Represents certificate used to entitle a consumer
  */
-@XmlRootElement
+@XmlRootElement(name = "cert")
 @XmlAccessorType(XmlAccessType.PROPERTY)
 @Entity
-@Table(name = IdentityCertificate.DB_TABLE)
-@JsonFilter("IdentityCertificateFilter")
-public class IdentityCertificate extends AbstractCertificate implements Certificate {
-
-    /** Name of the table backing this object in the database */
-    public static final String DB_TABLE = "cp_id_cert";
-
+@Table(name = "cp_cont_access_cert")
+@JsonFilter("ContentAccessCertificateFilter")
+public class ContentAccessCertificate extends AbstractCertificate implements Certificate {
     @Id
     @GeneratedValue(generator = "system-uuid")
     @GenericGenerator(name = "system-uuid", strategy = "uuid")
@@ -66,18 +61,13 @@ public class IdentityCertificate extends AbstractCertificate implements Certific
         this.serial = serialNumber;
     }
 
+    @Override
     public String getId() {
         return id;
     }
 
     public void setId(String id) {
         this.id = id;
-    }
-
-    public void update(IdentityCertificate other) {
-        this.setKey(other.getKey());
-        this.setCert(other.getCert());
-        this.setSerial(other.getSerial());
     }
 
     @XmlTransient
@@ -89,4 +79,25 @@ public class IdentityCertificate extends AbstractCertificate implements Certific
         this.consumer = consumer;
     }
 
+    @Override
+    public int hashCode() {
+        return this.id == null ? 0 : id.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || (getClass() != obj.getClass())) {
+            return false;
+        }
+        ContentAccessCertificate other = (ContentAccessCertificate) obj;
+        if (id == other.id &&
+            other.getConsumer().getId().equals(this.getConsumer().getId())) {
+
+            return true;
+        }
+        return false;
+    }
 }
--- a/server/src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
+++ b/server/src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
@@ -1,0 +1,80 @@
+/**
+ * Copyright (c) 2009 - 2012 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.model;
+
+import com.google.common.collect.Iterables;
+import com.google.inject.Inject;
+import com.google.inject.persist.Transactional;
+
+import org.hibernate.criterion.Restrictions;
+
+import java.util.List;
+
+import javax.persistence.Query;
+
+
+
+/**
+ * ContentAccessCertificateCurator
+ */
+public class ContentAccessCertificateCurator extends AbstractHibernateCurator<ContentAccessCertificate> {
+
+    @Inject
+    public ContentAccessCertificateCurator() {
+        super(ContentAccessCertificate.class);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Transactional
+    public ContentAccessCertificate getForConsumer(Consumer c) {
+        return (ContentAccessCertificate) currentSession().createCriteria(ContentAccessCertificate.class)
+            .add(Restrictions.eq("consumer", c))
+            .uniqueResult();
+    }
+
+    /**
+     * Delete unneeded content access certs.
+     *
+     * @return the number of rows deleted.
+     */
+    @Transactional
+    public int deleteForOwner(Owner owner) {
+        // So we must get ids for this owner, and then delete them
+        @SuppressWarnings("unchecked")
+
+        String hql = " SELECT cac.id " +
+            "    FROM Consumer c" +
+            "       JOIN c.owner o" +
+            "       JOIN c.contentAccessCert cac" +
+            "    WHERE" +
+            "       o.key=:ownerkey";
+        Query query = this.getEntityManager().createQuery(hql);
+        List<String> certsToDelete = query.setParameter("ownerkey", owner.getKey()).getResultList();
+
+        hql = "DELETE from ContentAccessCertificate WHERE id IN (:certsToDelete)";
+        String hql2 = "UPDATE Consumer set contentAccessCert = null WHERE " +
+            "contentAccessCert.id IN (:certsToDelete)";
+        query = this.getEntityManager().createQuery(hql);
+        Query query2 = this.getEntityManager().createQuery(hql2);
+        int removed = 0;
+        for (List<String> block : Iterables.partition(certsToDelete,
+            AbstractHibernateCurator.IN_OPERATOR_BLOCK_SIZE)) {
+            String param = block.toString();
+            query2.setParameter("certsToDelete", block).executeUpdate();
+            removed += query.setParameter("certsToDelete", block).executeUpdate();
+        }
+        return removed;
+    }
+}
--- a/server/src/main/java/org/candlepin/model/ImportUpstreamConsumer.java
+++ b/server/src/main/java/org/candlepin/model/ImportUpstreamConsumer.java
@@ -76,13 +76,17 @@ public class ImportUpstreamConsumer extends AbstractHibernateObject {
     @NotNull
     private String ownerId;
 
-    @Column(length = 255, name = "prefix_url_web")
+    @Column(name = "prefix_url_web")
     @Size(max = 255)
     private String prefixUrlWeb;
 
-    @Column(length = 255, name = "prefix_url_api")
+    @Column(name = "prefix_url_api")
     @Size(max = 255)
     private String prefixUrlApi;
+
+    @Column(name = "content_access_mode")
+    @Size(max = 255)
+    private String contentAccessMode;
 
     public ImportUpstreamConsumer() {
         // needed for Hibernate
@@ -95,6 +99,7 @@ public class ImportUpstreamConsumer extends AbstractHibernateObject {
         this.setType(uc.getType());
         this.setWebUrl(uc.getWebUrl());
         this.setApiUrl(uc.getApiUrl());
+        this.setContentAccessMode(uc.getContentAccessMode());
     }
 
     /**
@@ -200,5 +205,21 @@ public class ImportUpstreamConsumer extends AbstractHibernateObject {
      */
     public void setApiUrl(String url) {
         prefixUrlApi = url;
+    }
+
+    /**
+    *
+    * @return the Content Access Mode
+    */
+    public String getContentAccessMode() {
+        return this.contentAccessMode;
+    }
+
+    /**
+     *
+     * @param contentAccessMode
+     */
+    public void setContentAccessMode(String contentAccessMode) {
+        this.contentAccessMode = contentAccessMode;
     }
 }
--- a/server/src/main/java/org/candlepin/model/Owner.java
+++ b/server/src/main/java/org/candlepin/model/Owner.java
@@ -17,9 +17,12 @@ package org.candlepin.model;
 import org.candlepin.common.jackson.HateoasInclude;
 import org.candlepin.model.activationkeys.ActivationKey;
 import org.candlepin.resteasy.InfoProperty;
+import org.candlepin.service.ContentAccessCertServiceAdapter;
 
 import com.fasterxml.jackson.annotation.JsonFilter;
 
+import org.apache.commons.lang.ArrayUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.hibernate.annotations.GenericGenerator;
 
 import java.io.Serializable;
@@ -122,6 +125,27 @@ public class Owner extends AbstractHibernateObject implements Serializable,
      */
     @Column(name = "autobind_disabled")
     private Boolean autobindDisabled;
+
+    /**
+     * Determines the behavior of the content access.
+     *
+     */
+    @Column(name = "content_access_mode")
+    private String contentAccessMode;
+
+    /**
+     * Determines the allowable modes of the content access.
+     *
+     */
+    @Column(name = "content_access_mode_list")
+    private String contentAccessModeList;
+
+    /**
+     * Has the content access mode changed on owner update.
+     *
+     */
+    @Column(name = "content_access_mode_dirty")
+    private boolean contentAccessModeDirty;
 
     /**
      * Default constructor
@@ -302,6 +326,10 @@ public class Owner extends AbstractHibernateObject implements Serializable,
             !this.autobindDisabled.equals(other.autobindDisabled)) {
             return false;
         }
+        if ((this.contentAccessMode == null) ? (other.contentAccessMode != null) :
+            !this.contentAccessMode.equals(other.contentAccessMode)) {
+            return false;
+        }
         return true;
     }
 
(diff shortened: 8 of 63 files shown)


wottop:
buildbot: retest this please

server/src/main/java/org/candlepin/model/OwnerEnvContentAccessCurator.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public OwnerEnvContentAccessCurator() {
+        super(OwnerEnvContentAccess.class);
+    }
+
+    @Transactional
+    public OwnerEnvContentAccess getContentAccess(String ownerId, String environmentId) {
+        OwnerEnvContentAccess result = null;
+        if (environmentId != null) {
+            result = (OwnerEnvContentAccess) this.createSecureCriteria()
```
barnabycourt: Please use JPQL

server/src/main/java/org/candlepin/model/OwnerEnvContentAccessCurator.java
```diff
@@ -0,0 +1,63 @@
...
+            .add(Restrictions.eq("owner.id", ownerId))
+            .add(Restrictions.eq("environment.id", environmentId))
+            .uniqueResult();
+        }
+        else {
+            result = (OwnerEnvContentAccess) this.createSecureCriteria()
```
barnabycourt: Please use JPQL

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1256,9 +1266,54 @@ public void deleteConsumer(
                 returnCerts.add(cert);
             }
         }
+        // we want to insert the content access cert to this list
+        if (consumer.getContentAccessCert() != null) {
+            try {
+                returnCerts.add(contentAccessCertService.getCertificate(consumer));
+            }
+            catch (IOException ioe) {
+                throw new BadRequestException(i18n.tr("Cannot retrieve content access certificate."));
```
barnabycourt: If wrapping an exception, please embed the original exception.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1256,9 +1266,54 @@ public void deleteConsumer(
...
+            }
+            catch (GeneralSecurityException gse) {
+                throw new BadRequestException(i18n.tr("Cannot retrieve content access certificate."));
```
barnabycourt: If wrapping an exception, please embed the original exception.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1256,9 +1266,54 @@ public void deleteConsumer(
...
 
+    @ApiOperation(notes = "Retrieves the body of the Content Access Certificate for the Consumer",
+        value = "getContentAccessBody")
+    @ApiResponses({ @ApiResponse(code = 404, message = "") })
+    @GET
+    @Path("{consumer_uuid}/content")
+    @Produces(MediaType.APPLICATION_JSON)
+    public ContentAccessListing getContentAccessBody(
+        @PathParam("consumer_uuid") @Verify(Consumer.class) String consumerUuid,
+        @QueryParam("if-modified-since") Date since) {
```
barnabycourt: the "since" value is not being used.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1256,9 +1266,54 @@ public void deleteConsumer(
...
+
+        log.debug("Getting content access certificate for consumer: {}", consumerUuid);
+        Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
+        ContentAccessListing result = new ContentAccessListing();
+
+        // if there have been changes since the last cert generation
```
barnabycourt: This If needs to be filled in.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1256,9 +1266,54 @@ public void deleteConsumer(
...
+            ContentAccessCertificate cac = contentAccessCertService.getCertificate(consumer);
+
+            String cert = cac.getCert();
+            String json = cert.substring(cert.indexOf("-----BEGIN ENTITLEMENT DATA-----\n"));
+            json = json.substring(0, json.indexOf("-----END ENTITLEMENT DATA-----") + 30);
+            result.setContentListing(cac.getSerial().getId(), json);
+            result.setLastUpdate(cac.getUpdated());
+        }
+        catch (IOException ioe) {
+            throw new BadRequestException(i18n.tr("Cannot retrieve content access certificate body."));
```
barnabycourt: Original exception should be embedded in the new exception.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1256,9 +1266,54 @@ public void deleteConsumer(
...
+        }
+        catch (GeneralSecurityException e) {
+            throw new BadRequestException(i18n.tr("Cannot retrieve content access certificate body."));
```
barnabycourt: Original exception should be embedded in the new exception.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1337,6 +1392,10 @@ public File exportCertificates(
             .listForConsumer(consumer)) {
             allCerts.add(new CertificateSerialDto(cert.getSerial().getId()));
         }
+        // add content access cert if needed
```
barnabycourt: If all this method really cares about is a list of the serials, a JPA union query could get that information in a more efficient manner.

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
```diff
@@ -289,6 +298,12 @@ public ContentData updateContent(@PathParam("owner_key") String ownerKey,
         }
 
         existing = this.contentManager.updateContent(content, owner, true);
+        try {
+            ownerManager.refreshOwnerForContentAccess(owner);
+        }
+        catch (Exception e) {
+            log.error("Unable to handle content access certs.", e.getStackTrace().toString());
```
barnabycourt: This should re-throw, eating the exception will cause confusion.

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
```diff
@@ -289,6 +298,12 @@ public ContentData updateContent(@PathParam("owner_key") String ownerKey,
         }
 
         existing = this.contentManager.updateContent(content, owner, true);
+        try {
+            ownerManager.refreshOwnerForContentAccess(owner);
```
barnabycourt: Doing this during every content update could lead to pretty ugly performance implications if there are a lot of consumers. What is the performance of this if an owner has 40k consumers?

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
```diff
@@ -307,5 +322,12 @@ public void remove(@PathParam("owner_key") String ownerKey,
         }
 
         this.contentManager.removeContent(content, owner, true);
+        try {
+            ownerManager.refreshOwnerForContentAccess(owner);
```
barnabycourt: What is the performance of this if an owner has 40k consumers?

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
```diff
@@ -307,5 +322,12 @@ public void remove(@PathParam("owner_key") String ownerKey,
...
+        }
+        catch (Exception e) {
+            log.error("Unable to handle content access certs.", e.getStackTrace().toString());
```
barnabycourt: Embed the caught exception in the thrown exception.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -897,12 +897,47 @@ public Owner updateOwner(@PathParam("owner_key") @Verify(Owner.class) String key
...
+        }
+
         ownerCurator.merge(toUpdate);
+
+        if (toUpdate.isContentAccessModeDirty()) {
+            try {
+                ownerManager.refreshOwnerForContentAccess(toUpdate);
+            }
+            catch (Exception e) {
+                throw new BadRequestException(
```
barnabycourt: Embed the caught exception in the newly thrown exception

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -897,12 +897,47 @@ public Owner updateOwner(@PathParam("owner_key") @Verify(Owner.class) String key
...
         }
 
-        // Update the autobindDisabled field if the incoming value is null.
+        // Update the autobindDisabled field if the incoming value is not null.
         if (owner.getAutobindDisabled() != null) {
             toUpdate.setAutobindDisabled(owner.getAutobindDisabled());
         }
 
+        // Update the contentAccess field if the incoming value is not null.
+        if (owner.getContentAccessMode() != null) {
```
barnabycourt: What permissions are needed in order to update this? If the content access list can be updated via the REST API that should only be possible in hosted mode.  An owner should not be able to update this value for themselves.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -897,12 +897,47 @@ public Owner updateOwner(@PathParam("owner_key") @Verify(Owner.class) String key
...
+                throw new BadRequestException(
+                    i18n.tr("The content access mode is not allowed for this owner."));
+            }
+        }
+
         ownerCurator.merge(toUpdate);
+
+        if (toUpdate.isContentAccessModeDirty()) {
+            try {
+                ownerManager.refreshOwnerForContentAccess(toUpdate);
```
barnabycourt: How long does this take for an owner with large numbers of subscriptions & large numbers of consumers?

server/src/main/java/org/candlepin/service/impl/DefaultContentAccessCertServiceAdapter.java
```diff
@@ -0,0 +1,300 @@
...
+        String contentPrefix = getContentPrefix(owner, environment);
+        Product container = new Product();
+        Entitlement emptyEnt = new Entitlement();
+        Pool emptyPool = new Pool();
+        Product skuProduct = new Product();
+        Consumer emptyConsumer = new Consumer();
+
+        containerSet.add(container);
+        container.setId("content_access");
+        container.setName("Golden Ticket");
```
barnabycourt: perhaps "content access certificate" instead of Golden Ticket. As much as I love the name, exposing it to customers is likely to be confusing.

server/src/main/java/org/candlepin/service/impl/DefaultContentAccessCertServiceAdapter.java
```diff
@@ -0,0 +1,300 @@
...
+            container.addContent(c, false);
+        }
+
+        emptyConsumer.setEnvironment(environment);
+        emptyEnt.setPool(emptyPool);
+        emptyEnt.setConsumer(emptyConsumer);
+        emptyPool.setProduct(skuProduct);
+        emptyPool.setStartDate(new Date());
+        emptyPool.setEndDate(new Date());
+        skuProduct.setName("Golden Ticket");
```
barnabycourt: perhaps "content access certificate" instead of Golden Ticket. As much as I love the name, exposing it to customers is likely to be confusing.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -139,6 +139,10 @@
     @JoinColumn(name = "consumer_idcert_id")
     private IdentityCertificate idCert;
 
+    @OneToOne
+    @JoinColumn(name = "consumer_cont_acc_cert_id")
+    private ContentAccessCertificate contentAccessCert;
```
wottop: I'll make this a lazy load

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1255,9 +1265,54 @@ public void deleteConsumer(
...
 
+    @ApiOperation(notes = "Retrieves the body of the Content Access Certificate for the Consumer",
+        value = "getContentAccessBody")
+    @ApiResponses({ @ApiResponse(code = 404, message = "") })
+    @GET
+    @Path("{consumer_uuid}/content")
+    @Produces(MediaType.APPLICATION_JSON)
+    public ContentAccessListing getContentAccessBody(
+        @PathParam("consumer_uuid") @Verify(Consumer.class) String consumerUuid,
+        @QueryParam("if-modified-since") Date since) {
```
kahowell: I found in smoketesting work in candlepin/python-rhsm#191, that by default, it appears to use `java.util.Date.parse` (which accepts a very specific format of dates - and *not* ISO format); it would be great if this could just accept ISO dates, so that I can simply use `isoformat()` on a `datetime` object (or even just pass back in the string I get from the `/content` endpoint) instead of [this](https://github.com/candlepin/python-rhsm/blob/8e148184d99a271784c6503e2cc10527c1bafa16/src/rhsm/connection.py#L1097). Methinks this could be accomplished with a `StringParameterUnmarshaller` for `Date`? Also a quick note: there is some variance allowed in ISO-8601, in Python you get this variant:

    2016-11-17T17:00:00.123456-04:00

(ideally we should accept any ISO-8601 variant).

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1255,9 +1265,54 @@ public void deleteConsumer(
...
+        Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
+        ContentAccessListing result = new ContentAccessListing();
+
+        // if there have been changes since the last cert generation
+        try {
+            ContentAccessCertificate cac = contentAccessCertService.getCertificate(consumer);
+
+            String cert = cac.getCert();
+            String json = cert.substring(cert.indexOf("-----BEGIN ENTITLEMENT DATA-----\n"));
+            json = json.substring(0, json.indexOf("-----END ENTITLEMENT DATA-----") + 30);
```
kahowell: `json` does not include signature yet... I left a [`FIXME`](https://github.com/candlepin/subscription-manager/blob/0790a62837a8e1c4c0944fb2e1e85e6bc8121303/src/subscription_manager/cache.py#L601) for the corresponding subscription manager change once this is fixed.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
kahowell: Regarding RFC 2616 as the format, I'd be inclined to agree if we were specifying the date in a header a la RFC 2616, but since we're exposing as a query param, I lean more towards making it consistent w/ other dates (esp. dates in responses). I would not make a fuss if we went with the RFC 2616 timestamp though. :smile:

server/src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -150,6 +151,12 @@ public void run() {
 
         for (Owner owner : this.owners.values()) {
             poolManager.refreshPoolsWithRegeneration(subAdapter, owner, lazy);
+            try {
+                ownerManager.refreshOwnerForContentAccess(owner);
+            }
+            catch (Exception e) {
+                log.error("Unable to handle content access certificates", e.getStackTrace().toString());
```
wottop: The refresher runs async most of the time. In those cases it cannot.

server/src/main/java/org/candlepin/model/ContentAccessCertificate.java
```diff
@@ -0,0 +1,103 @@
...
+    @GenericGenerator(name = "system-uuid", strategy = "uuid")
+    @Column(length = 32)
+    @NotNull
+    private String id;
+
+    @OneToOne
+    private CertificateSerial serial;
+
+    @OneToOne(mappedBy = "idCert", fetch = FetchType.LAZY)
+    private Consumer consumer;
```
wottop: adding index

server/src/main/java/org/candlepin/model/ContentAccessListing.java
```diff
@@ -0,0 +1,43 @@
...
+package org.candlepin.model;
+
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * ContentAccessListing class dto for return data for content access API
+ */
+public class ContentAccessListing {
```
wottop: so moved

server/src/main/java/org/candlepin/model/Owner.java
```diff
@@ -442,4 +467,55 @@ public Boolean getAutobindDisabled() {
...
+
+    /**
+     * Utility method that defaults to 'entitlement'.
+     *  This covers legacy owners
+     *
+     * @return access mode.
+     */
+    @XmlTransient
+    public String contentAccessMode() {
+        return getContentAccessMode() == null ? "entitlement" : getContentAccessMode();
```
wottop: added

server/src/main/java/org/candlepin/model/OwnerEnvContentAccess.java
```diff
@@ -0,0 +1,122 @@
...
+    private Owner owner;
+
+    @ManyToOne
+    @ForeignKey(name = "fk_owner_env_cont_acc_env")
+    @JoinColumn(nullable = true)
+    private Environment environment;
+
+    @NotNull
+    @Column(name = "content_json")
+    private String contentJson;
```
wottop: String is 2^31-1 characters.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: done

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: done

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: implemented

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: removed

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: done

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: done

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
wottop: It only removes the cached content block in this call. The number of consumers does not change the timing for new content.

Working on the deletion of CACerts on the consumers via db call.

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
wottop: Done

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
wottop: as per previous comment.

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
wottop: done

server/src/main/java/org/candlepin/resource/OwnerResource.java
wottop: as per previous

server/src/main/java/org/candlepin/service/impl/DefaultContentAccessCertServiceAdapter.java
wottop: done

server/src/main/java/org/candlepin/service/impl/DefaultContentAccessCertServiceAdapter.java
wottop: done

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: coming soon

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: Altered spec test to use RFC 2616 [aka date.httpdate in json]
It takes it now.

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
wottop: this is done also

server/src/main/java/org/candlepin/model/OwnerEnvContentAccessCurator.java
wottop: done

server/src/main/java/org/candlepin/model/OwnerEnvContentAccessCurator.java
wottop: done

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: implemented

server/src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Inject
+    public ContentAccessCertificateCurator() {
+        super(ContentAccessCertificate.class);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Transactional
+    public ContentAccessCertificate getForConsumer(Consumer c) {
+        return (ContentAccessCertificate) currentSession().createCriteria(ContentAccessCertificate.class)
```
wottop: Done

server/src/main/java/org/candlepin/controller/OwnerManager.java
```diff
@@ -151,4 +160,24 @@ public void cleanupAndDelete(Owner owner, boolean revokeCerts) {
         ownerCurator.delete(owner);
     }
 
+    public void determineContentAccessCertState(Owner owner) throws GeneralSecurityException, IOException {
```
awood: Just curious: what throws GeneralSecurityException and/or IOException here?

server/src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -40,6 +43,7 @@
 
     private CandlepinPoolManager poolManager;
     private SubscriptionServiceAdapter subAdapter;
+    @Inject private OwnerManager ownerManager;
```
awood: Is the constructor for this class not injected?  I'm a little wary of field injection since it relies on reflection tricks and stuff (and is ultimately less transparent than constructor injection which is how most of the rest of our stuff is).

server/src/main/java/org/candlepin/model/CertificateSerialCurator.java
```diff
@@ -152,4 +153,26 @@ private Criterion getRevokedCriteria() {
...
+        String hql = "SELECT s.id" +
+            "    FROM EntitlementCertificate ec" +
+            "     JOIN ec.entitlement e" +
+            "     JOIN e.consumer c" +
+            "     JOIN ec.serial s" +
+            "     JOIN e.pool p" +
+            "    WHERE" +
+            "       c.id=:consumerId" +
+            "    AND" +
+            "       p.endDate >= :nowDate";
```
awood: Might want to think about timezones here.  `new Date()` will be in the server's current timezone, I believe.  Not sure what endDate will be stored in (UTC?).  Might be worth some unit tests if you haven't already.

server/src/main/java/org/candlepin/model/ImportUpstreamConsumer.java
```diff
@@ -84,6 +84,10 @@
     @Size(max = 255)
     private String prefixUrlApi;
 
+    @Column(length = 255, name = "content_access_mode")
```
awood: `length` here doesn't really do anything.  That property is only used, IIRC, for DDL generation which we don't do since we have Liquibase.

server/src/main/java/org/candlepin/model/Owner.java
```diff
@@ -302,6 +326,10 @@ public boolean equals(Object obj) {
             !this.autobindDisabled.equals(other.autobindDisabled)) {
             return false;
         }
+        if ((this.contentAccessMode == null) ? (other.contentAccessMode != null) :
```
awood: This is a little too complicated for my tiny mind.  A ternary expression inside an if?  I didn't even know you could do that.  Can we simply this some?  Maybe

```java
if (contentAccessMode == null) {
  return other.contentAccessMode == null;
} else {
  return contentAccessMode.equals(other.contentAccessMode);
}
```

server/src/main/java/org/candlepin/model/OwnerEnvContentAccess.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Id
+    @GeneratedValue(generator = "system-uuid")
+    @GenericGenerator(name = "system-uuid", strategy = "uuid")
+    @Column(length = 32)
+    @NotNull
+    private String id;
+
+    @ManyToOne
+    @ForeignKey(name = "fk_owner_env_cont_acc_owner")
```
awood: Pretty sure [`@ForeignKey`](https://docs.jboss.org/hibernate/jpa/2.1/api/javax/persistence/ForeignKey.html) is only used for schema generation which we don't do.

server/src/main/java/org/candlepin/model/UpstreamConsumer.java
```diff
@@ -88,6 +88,10 @@
     @Size(max = 255)
     private String prefixUrlApi;
 
+    @Column(length = 255, name = "content_access_mode")
```
awood: I think you can drop `length` here as well.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -930,6 +959,23 @@ public boolean performConsumerUpdates(Consumer updated, Consumer toUpdate,
             }
         }
 
+        if (updated.getContentAccessMode() != null &&
+            !updated.getContentAccessMode().equals(toUpdate.getContentAccessMode()) &&
+            toUpdate.getType().isManifest()) {
+            if (!StringUtils.isBlank(updated.getContentAccessMode()) &&
```
awood: Saw this `!StringUtils.isBlank...` predicate earlier.  Let's extract that bit out into a common method.  Makes the code easier to read and cuts down on redundancy.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1255,9 +1301,68 @@ public void deleteConsumer(
...
+        @PathParam("consumer_uuid") @Verify(Consumer.class) String consumerUuid,
+        @QueryParam("if-modified-since") Date since) {
+
+        log.debug("Getting content access certificate for consumer: {}", consumerUuid);
+        Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
+        if (StringUtils.isEmpty(consumer.getOwner().getContentAccessMode())) {
+            throw new BadRequestException(i18n.tr("No content access mode assigned"));
+        }
+        if (!contentAccessCertService.hasCertChangedSince(consumer, since)) {
+            throw new NotModifiedException("Not modified since date supplied.");
```
awood: I'm not 100% sure we want to throw an exception here since this isn't really an exceptional case.  In fact, we expect it to happen frequently.  Instead we can just use RestEasy's response builder to build a response with a 304.  Using exceptions for control flow can result in poor performance and some other issues as described by [item 57 in Effective Java](https://www.safaribooksonline.com/library/view/effective-java-2nd/9780137150021/ch09.html)

server/src/main/java/org/candlepin/service/impl/DefaultContentAccessCertServiceAdapter.java
```diff
@@ -0,0 +1,321 @@
...
+
+    public boolean hasCertChangedSince(Consumer consumer, Date date) {
+        if (date == null) {
+            return true;
+        }
+        Environment env = consumer.getEnvironment();
+        Owner owner = consumer.getOwner();
+        OwnerEnvContentAccess oeca = ownerEnvContentAccessCurator.getContentAccess(
+            owner.getId(), env == null ? null : env.getId());
+        if (oeca == null || consumer.getContentAccessCert() == null ||
```
awood:
```
return oeca == null || consumer.getContentAccessCert() == null || oeca.getUpdated().getTime() > date.getTime())
```

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: Its not exactly the same. in one case it checks the existing in one part and the incoming in the other.

awood:
Looks good to me.  Ready for merge pending performance results.

barnabycourt:
On the owner resource, what it the purpose of having the contentAccessModeDirty included in the JSON?

barnabycourt:
Having contentAccessModeList and contentAccessModeListAsArray in the Owner json is redundant. Is there a reason to not include the array form of the list as the value for contentAccessModeList?

barnabycourt:
One other naming question, /conusmers/<id>/content_access does not seem clear to me. Maybe content_definition?

awood:
It would be nice, I think (and members of the hosted team agree), if golden ticket certificates were marked as such.  In other words, a golden ticket is a golden ticket because it *has* a specific characteristic (e.g. True under some new OID) not because it lacks a characteristic inherent to a normal certificate.  I'm not sure if this is the case or not, but I think it would be a nice add.

server/src/main/java/org/candlepin/model/Owner.java
```diff
@@ -442,4 +470,69 @@ public Boolean getAutobindDisabled() {
...
+    }
+
+    public void setContentAccessModeList(String contentAccessModeList) {
+        this.contentAccessModeList = contentAccessModeList;
+    }
+
+    @XmlTransient
+    public boolean hasContentAccessMode(String mode) {
+        if (StringUtils.isEmpty(mode)) {
+            return true;
```
Ceiu: Is this a valid case? Does null/empty represent an actual state? Maybe this only seems weird to me because of the verbiage of the method name.

server/src/main/java/org/candlepin/model/Owner.java
Ceiu: How about just the ternary expression?

```return contentAccessMode == null ? other.contentAccessMode == null : contentAccessMode.equals(other.contentAccessMode);```

File: server/src/main/java/org/candlepin/resource/ConsumerResource.java [hunk 39]
```diff
@@ -1339,11 +1436,16 @@ public File exportCertificates(
...
 
         List<CertificateSerialDto> allCerts = new LinkedList<CertificateSerialDto>();
-        for (EntitlementCertificate cert : entCertService
-            .listForConsumer(consumer)) {
-            allCerts.add(new CertificateSerialDto(cert.getSerial().getId()));
+        for (Long id : entCertService.listEntitlementSerialIds(consumer)) {
+            allCerts.add(new CertificateSerialDto(id));
+        }
+        // add content access cert if needed
+        if (!StringUtils.isEmpty(consumer.getOwner().getContentAccessMode())) {
```
Ceiu: These chained calls may end up being performance pain points later, as I believe Hibernate will lazily load the Owner, and then pull the access mode. Nothing to change quite yet, but something to keep in mind for the future.

server/src/main/java/org/candlepin/model/Owner.java
```diff
@@ -442,4 +470,69 @@ public Boolean getAutobindDisabled() {
...
+
+    /**
+     * Utility method that defaults to 'entitlement'.
+     *  This covers legacy owners
+     *
+     * @return access mode.
+     */
+    @XmlTransient
+    public String contentAccessMode() {
+        return StringUtils.isEmpty(getContentAccessMode()) ?
```
Ceiu: Should we just always do this?

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -894,12 +894,36 @@ public Owner updateOwner(@PathParam("owner_key") @Verify(Owner.class) String key
             }
         }
 
-        // Update the autobindDisabled field if the incoming value is null.
+        // Update the autobindDisabled field if the incoming value is not null.
         if (owner.getAutobindDisabled() != null) {
             toUpdate.setAutobindDisabled(owner.getAutobindDisabled());
         }
 
+        if (owner.getContentAccessMode() != null && config.getBoolean(ConfigProperties.STANDALONE)) {
```
Ceiu: This is another part where the difference between null and an empty string will cause problems.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -894,12 +894,36 @@ public Owner updateOwner(@PathParam("owner_key") @Verify(Owner.class) String key
...
+            throw new BadRequestException(
+                i18n.tr("The owner content access mode cannot be set directly in standalone mode."));
+        }
+        // Update the contentAccess field if the incoming value is not null.
+        if (owner.getContentAccessMode() != null) {
```
Ceiu: And here

wottop:
buildbot: retest this please

barnabycourt:
@wottop I don't see a capability added to the list for content-access-mode

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1262,9 +1298,72 @@ public void deleteConsumer(
...
 
+    @ApiOperation(notes = "Retrieves the body of the Content Access Certificate for the Consumer",
+        value = "getContentAccessBody")
+    @ApiResponses({ @ApiResponse(code = 404, message = ""), @ApiResponse(code = 304, message = "") })
+    @GET
+    @Path("{consumer_uuid}/accessible_content")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getContentAccessBody(
+        @PathParam("consumer_uuid") @Verify(Consumer.class) String consumerUuid,
+        @QueryParam("if-modified-since") Date since) {
```
awood: We should use the "If-Modified-Since" header on the request.  That's what the HTTP spec expects: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html (see section 14.25).  RESTEasy/JAX-RS supports this with the `@HeaderParam` annotation.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1262,9 +1298,72 @@ public void deleteConsumer(
...
+        log.debug("Getting content access certificate for consumer: {}", consumerUuid);
+        Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
+        if (consumer.getOwner().contentAccessMode()
+            .equals(ContentAccessCertServiceAdapter.DEFAULT_CONTENT_ACCESS_MODE)) {
+            throw new BadRequestException(i18n.tr("No content access mode assigned"));
+        }
+        if (!contentAccessCertService.hasCertChangedSince(consumer, since)) {
+            return Response.status(Response.Status.NOT_MODIFIED)
+                .entity("Not modified since date supplied.").build();
+            //            throw new NotModifiedException("Not modified since date supplied.");
```
awood: We can remove this commented line

server/src/main/java/org/candlepin/service/impl/DefaultContentAccessCertServiceAdapter.java
```diff
@@ -0,0 +1,318 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * DefaultEntitlementCertServiceAdapter
+ */
+public class DefaultContentAccessCertServiceAdapter implements ContentAccessCertServiceAdapter {
```
awood: I'd like to see more unity around the interfaces and implementations.  Right now it looks like we have 

```
EntitlementCertServiceAdapter
BaseEntitlementCertServiceAdapter implements EntitlementCertServiceAdapter
ContentAccessCertServiceAdapter completely unrelated to the previous two
DefaultEntitlementCertServiceAdapter extends BaseEntitlementCertServiceAdapter
DefaultContentAccessCertServiceAdapter implements ContentAccessCertServiceAdapter
```

I'd like to see them all in the same family tree at least.  Even better would be if we could move common bits up to a BaseCertServiceAdapter that handles things like writing the JSON blob and basic certificate generation.  Then we have a ContentAccessCertServiceAdapter interface and an EntitlementCertServiceAdapter interface and the implementations both extend the abstract class and implement the appropriate interface.

server/src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
```diff
@@ -323,6 +324,9 @@ private String getContentPrefix(Entitlement ent, boolean useContentPrefix)
         String contentPrefix, Map<String, EnvironmentContent> promotedContent) {
         Set<X509ExtensionWrapper> result = v3extensionUtil.getExtensions(ent,
             contentPrefix, promotedContent);
+        X509ExtensionWrapper typeExtension = new X509ExtensionWrapper(OIDUtil.REDHAT_OID + "." +
```
awood: @barnabycourt do we want to take the opportunity in this class to start baking in the org key into the certificate DN?

server/src/main/java/org/candlepin/resource/ConsumerResource.java
barnabycourt: agreed

server/src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
barnabycourt: It has already been baked into the content-access cert DN, aer you referring to the identity cert DN?

barnabycourt:
@wottop if I register a consumer, and then immediately call get to /consumers/{id}/certificates/serials, the list is empty, however, if I call get on /consumers/{consumer_id}/certificates before I call the get on serials it works fine

barnabycourt:
@wottop For the content-access cert, the headers for CN & O are mapped to the consumer id and owner uuid instead of the consumer uuid and the owner key

awood:
> @awood It has already been baked into the content-access cert DN, aer you referring to the identity cert DN?

I was referring to baking into a normal entitlement certificate DN.  Since in the future, the CDN would probably like to know what org a consumer is in without having to look it up somewhere.

PR #1392, opened 2016-11-16
Re-aligning file location with package def

mstead:
Package and file location didn't line up. Was making eclipse wig out!


Ceiu:
ACK

PR #1393, opened 2016-11-16
Fix issue locking Owner on Oracle

mstead:
For some reason or other, against Oracle,
setting the LockMode on the query does not
seem to work.

As a workaround, look up the Owner by key,
and do a refresh with a LockMode.

### DIFF
--- a/server/src/main/java/org/candlepin/model/OwnerCurator.java
+++ b/server/src/main/java/org/candlepin/model/OwnerCurator.java
@@ -26,6 +26,8 @@ import org.hibernate.criterion.Property;
 import org.hibernate.criterion.Restrictions;
 import org.hibernate.criterion.Subqueries;
 import org.hibernate.sql.JoinType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Collection;
 import java.util.Date;
@@ -41,6 +43,7 @@ import javax.persistence.LockModeType;
 public class OwnerCurator extends AbstractHibernateCurator<Owner> {
 
     @Inject private CandlepinQueryFactory cpQueryFactory;
+    private static Logger log = LoggerFactory.getLogger(OwnerCurator.class);
 
     public OwnerCurator() {
         super(Owner.class);
@@ -67,22 +70,15 @@ public class OwnerCurator extends AbstractHibernateCurator<Owner> {
      */
     @Transactional
     public Owner findAndLock(String ownerKey) {
-        List<Owner> result = getEntityManager().createQuery("select o from Owner o where key = :ownerKey",
+        List<Owner> result = getEntityManager().createQuery("select o from Owner o where o.key = :ownerKey",
             Owner.class)
             .setParameter("ownerKey", ownerKey)
-            .setMaxResults(1)
             .setLockMode(LockModeType.PESSIMISTIC_WRITE)
             .getResultList();
         if (result == null || result.isEmpty()) {
             return null;
         }
-
-        // For some reason or other, setting LockMode on the initial query does
-        // not seem to lock the Owner on Oracle. Refreshing the entity after
-        // looking it up seems to do the trick.
-        Owner owner = result.get(0);
-        getEntityManager().refresh(owner, LockModeType.PESSIMISTIC_WRITE);
-        return owner;
+        return result.get(0);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


mstead:
@nguyenfilip could you take a look at this when you get a minute? I'm not sure I like how I got this to work. Wondering if there is a better way to do it.  Thanks.

vritant:
@mstead looks like the issue might be in [lockAndLoadBatch](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java#L747) as well ? 

perhaps we should re-use the [similar method](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/model/OwnerCurator.java#L88) in OwnerCurator.

mstead:
This fixes Oracle, but now breaks MySQL. :(

Will take another look.

mstead:
@vritant that other method was unused so I removed it.

Seems like Oracle and MySQL have subtle differences when locking -- at least when using JPA.

When we lock Pools we use hibernate directly:
```java
currentSession().refresh(pool, LockOptions.UPGRADE);
```

I'm starting to wonder if we should deviate when locking Owner. I'm trying to follow the suggestion to use JPA API and avoid hibernate classes directly.

Regardless, the latest patch should work for all DBs. Setting the lock mode duing the lookup supports Postgres and MySql. The lock mode on the refresh supports Oracle. This feels really dirty, but I'm not sure of another way to solve it, other than doing it the same way as we do for Pools.

nguyenfilip:
@mstead sure, will take a look tomorrow. So the changes in this PR doesn't lock the owner on Oracle? Is there a failing spec test or how did you uncover the issue?

nguyenfilip:
@mstead also, what are the connection infromation for the Oracle instance that you use?

mstead:
> @mstead sure, will take a look tomorrow. So the changes in this PR doesn't lock the owner on Oracle? Is there a failing spec test or how did you uncover the issue?

The current commit for this PR works on the 3 DBs (as shown in the passing tests for this PR). Looking at the code changes, the lock on the lookup works for Postgres/MySQL but doesn't on Oracle. I added the refresh(owner, lock) to make it work for Oracle as well --- BUT this feels wrong to me.

This PR is working, but I'm not sure if it is the correct way.

There is a spec test that surfaces this issue -- ueber_cert_spec:'should handle concurrent requests to generate cert for an owner'

mstead:
> @mstead also, what are the connection infromation for the Oracle instance that you use?

I've been running the spec test using our docker containers locally. Install docker and run the following to run the spec tests against your local checkout:

```bash
cd FULL_PATH_TO_YOUR_CANDLEPIN_CHECKOUT/docker/candlepin-base
sudo ./build.sh

cd FULL_PATH_TO_YOUR_CANDLEPIN_CHECKOUT/docker/candlepin-oracle
sudo ./build.sh

cd FULL_PATH_TO_YOUR_CANDLEPIN_CHECKOUT

sudo docker run --privileged -v FULL_PATH_TO_YOUR_CANDLEPIN_CHECKOUT:/candlepin-dev -Pti --rm candlepin/candlepin-oracle cp-test -r ueber_cert_spec
```
**NOTE**:  Use of --privileged may/may not be required.

server/src/main/java/org/candlepin/model/OwnerCurator.java
```diff
@@ -76,16 +76,11 @@ public Owner findAndLock(String ownerKey) {
...
 
-    /**
-     * Refreshes the target Owner and locks it.
-     *
-     * @param owner the target owner.
-     * @return the refreshed and locked Owner.
-     */
-    public Owner lockAndLoad(Owner owner) {
+        // For some reason or other, setting LockMode on the initial query does
+        // not seem to lock the Ower on Oracle. Refreshing the entity after
```
nguyenfilip: typo "Ower"

vritant:
LGTM, I can merge if @nguyenfilip finds it okay too.
okay because:
 * there is precedence of refreshing after locking in candlepin already, in [AbstractHibernateCurator.lockAndLoadBatch](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java#L747) .

* refreshing one entity here is not that big an issue performance wise, especially since its an owner.

but looks like the unused method is still there in code. please either delete it or use it for the refresh, id be okay with either.

nguyenfilip:
The problem here is caused by follow on locking feature. This was dragged into candlepin with the Hibernate upgrade. 

https://www.mail-archive.com/hibernate-dev@lists.jboss.org/msg14520.html

I think the best fix here is to disable the feature. Will try that momentary

server/src/main/java/org/candlepin/model/OwnerCurator.java
mstead: Fixed

nguyenfilip:
retest this please

PR #1395, opened 2016-11-18
Fixed a crash when paging certain requests

Ceiu:
- Fixed a crash that could occur when paging requests for products or content
  by owner while not sorting by a owner id or content/product uuid
- Updated OwnerContentCurator to use the CandlepinQuery for its collection-based
  queries

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -536,24 +536,36 @@ class Candlepin
     post("/consumers/#{uuid}/entitlements")
   end
 
-  def list_products(product_ids=nil)
-    method = "/products?"
+  def list_products(product_ids=nil, params={})
+    path = "/products?"
     if product_ids
       product_ids.each { |id|
-        method << "&product=" << id
+        path << "product=#{id}&"
       }
     end
-    get(method)
+
+    # add paging params
+    path << "page=#{params[:page]}&" if params[:page]
+    path << "per_page=#{params[:per_page]}&" if params[:per_page]
+    path << "sort_by=#{params[:sort_by]}&" if params[:sort_by]
+
+    get(path)
   end
 
-  def list_products_by_owner(owner_key, product_ids=nil)
-    method = "/owners/#{owner_key}/products?"
+  def list_products_by_owner(owner_key, product_ids=nil, params={})
+    path = "/owners/#{owner_key}/products?"
     if product_ids
       product_ids.each { |id|
-        method << "&product=" << id
+        path << "product=#{id}&"
       }
     end
-    get(method)
+
+    # add paging params
+    path << "page=#{params[:page]}&" if params[:page]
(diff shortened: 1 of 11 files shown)


vritant:
@Ceiu thanks for the spec tests!

- [ ] i'd rather we remove the unused getOwnersByProduct and getOwnersByContent rather than fix them, especially given :point_down: 
- [ ] there is a subtle **non**-issue here which we need to consider. we fetch *all* the records of the join table in the first query, even if paged... now in this case that is not that big of a deal because we cannot possibly  be dealing with a lot of product and content records per owner ( assuming the above methods are purged, cause otherwise there could be a lot more of owners ), but if we start using this pattern for other entities, we should probably find an alternative. 
- [ ] the comments are great, but id recommend moving that to candlepinproject.org and referring that document in the code, instead of referring other comments in comments. looks like its used enough to be a quick doc with this code as an example perhaps?

Ceiu:
I suppose we could remove the currently unused methods, but I'd prefer to keep them around if it's not causing problems. When we removed the owner and products from products and owners (respectively), we lost any other means of fetching those collections directly or intuitively. When the need eventually arises, having the method available so whichever poor soul ends up needing it doesn't have to figure out why they're missing and what they need to do to get them has value to me; especially since it's literally just the inverse of the getProductsByOwner, and took all of 30s to write, due to the copy/paste nature.

The pulling of the full list of records (which is only UUIDs in both cases) is unfortunate, but required in this case. Hibernate really ties our hands with how we can work around this and, as far as performance goes, this is what was already happening anyway -- I've just made it explicit and usable. The alternative I considered was writing another CandlepinQuery implementation that can work with HQL- or JPQL-based queries, but both do not support adding sorting options on the fly, which means I would need to do some shenanigans with QL parsing.

Honestly, I'm not a huge fan of putting implementation details in a separate wiki which lives in a separate repo. It's good for design bits and "here's how you do x," but not for "this particular bit of code was written in this way for this reason" type things. That is, unless you're looking for a "odd things Hibernate does that you should try to avoid" type of document.

Ceiu:
Almost forgot: As far as this particular problem goes, we'll only need to worry about it in specific cases like this one where, (a) we query on a join object because (b) we are trying to select an entity filtering by the entities on the other side of the join and (c) we don't have a path from one side of the join to the other within the model because (d) many-to-many relationships are expensive to maintain at the model level.

I'm fairly certain this will be a one-off case unique to these two entities.

vritant:
I still think we should have remove the unused methods, cause in the reverse direction you can pull more than 100000 ids from the database. anywho, merging this cause you are passionate about it, usually I prefer removing unused code cause otherwise over time that practice leads to a lot more code to maintain, and we can always retrieve old code from git history.
 A wierd things hibernate doc sounds really good.. if not for candlepinproject , perhaps for your personal blog? given the battles you have with it every day, that sounds like perfect for you, just a suggestion.

PR #1396, opened 2016-11-21
Fixed a bug that would treat unchanged entities as updated

Ceiu:
- Fixed a bug that would count existing products or contents as
  updated during import or refresh, even if the operation did not
  result in a net change to the entity

### DIFF
--- a/server/src/main/java/org/candlepin/controller/ContentManager.java
+++ b/server/src/main/java/org/candlepin/controller/ContentManager.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2009 - 2012 Red Hat, Inc.
+ * Copyright (c) 2009 - 2016 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -315,21 +315,23 @@ public class ContentManager {
      *  A mapping of Red Hat content ID to content entities representing the imported content
      */
     @SuppressWarnings("checkstyle:methodlength")
-    public Map<String, Content> importContent(Owner owner, Map<String, ContentData> contentData,
+    public ImportResult<Content> importContent(Owner owner, Map<String, ContentData> contentData,
         Set<String> importedProductIds) {
 
         if (owner == null) {
             throw new IllegalArgumentException("owner is null");
         }
 
+        ImportResult<Content> importResult = new ImportResult<Content>();
+
         if (contentData == null || contentData.isEmpty()) {
             // Nothing to import
-            return new HashMap<String, Content>();
+            return importResult;
         }
 
-        Map<String, Content> importedContent = new HashMap<String, Content>();
-        Map<String, Content> createdContent = new HashMap<String, Content>();
-        Map<String, Content> updatedContent = new HashMap<String, Content>();
+        Map<String, Content> skippedContent = importResult.getSkippedEntities();
+        Map<String, Content> createdContent = importResult.getCreatedEntities();
+        Map<String, Content> updatedContent = importResult.getUpdatedEntities();
 
         Map<String, Integer> contentVersions = new HashMap<String, Integer>();
         Map<String, Content> sourceContent = new HashMap<String, Content>();
@@ -342,7 +344,7 @@ public class ContentManager {
 
             if (!content.isChangedBy(update)) {
                 // This content won't be changing, so we'll just pretend it's not being imported at all
-                importedContent.put(content.getId(), content);
+                skippedContent.put(content.getId(), content);
                 continue;
             }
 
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -324,12 +324,15 @@ public Content updateContent(ContentData update, Owner owner, boolean regenerate
 
         if (contentData == null || contentData.isEmpty()) {
             // Nothing to import
-            return new HashMap<String, Content>();
+            new ImportResult<Content>(Collections.<String, Content>emptyMap(),
```
nguyenfilip: Creation of empty ImportResult is repeated quite often. What about creating non-parametric constructor that will construct ImportResult with empty maps?

server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -324,12 +324,15 @@ public Content updateContent(ContentData update, Owner owner, boolean regenerate
...
+                Collections.<String, Content>emptyMap(), Collections.<String, Content>emptyMap());
         }
 
-        Map<String, Content> importedContent = new HashMap<String, Content>();
+        Map<String, Content> skippedContent = new HashMap<String, Content>();
         Map<String, Content> createdContent = new HashMap<String, Content>();
         Map<String, Content> updatedContent = new HashMap<String, Content>();
+        ImportResult<Content> importResult = new ImportResult<Content>(skippedContent, createdContent,
```
nguyenfilip: ImportResult is created here and used just to be returned at the end of the method. What about construct it just at line "return importResult"?

server/src/main/java/org/candlepin/controller/ImportResult.java
```diff
@@ -0,0 +1,126 @@
...
+
+        if (skippedEntities == null) {
+            throw new IllegalArgumentException("skippedEntities is null");
+        }
+
+        if (skippedEntities == null) {
+            throw new IllegalArgumentException("skippedEntities is null");
+        }
+
+        this.skippedEntities = skippedEntities;
```
nguyenfilip: Wouldn't it be better to make the class immutable? That way, the getImportedEntities() wouldn't have to do the 'caching' work importedEntities

server/src/main/java/org/candlepin/controller/ImportResult.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Retrieves a map containing the imported entities. The entities will be mapped by their Red
+     * Hat ID.
+     *
+     * @return
+     *  A map containing all imported entities
+     */
+    public Map<String, E> getImportedEntities() {
+        if (this.importedEntities == null) {
```
nguyenfilip: If this class was immutable, we wouldn't have to do this mind-bending thing

server/src/main/java/org/candlepin/controller/ImportResult.java
nguyenfilip: By immutable I mean to defensively copy the map in to the members of this class

server/src/main/java/org/candlepin/controller/ImportResult.java
```diff
@@ -0,0 +1,126 @@
...
+     *  if any of the provided maps are null
+     */
+    public ImportResult(Map<String, E> skippedEntities, Map<String, E> createdEntities,
+        Map<String, E> updatedEntities) {
+
+        if (skippedEntities == null) {
+            throw new IllegalArgumentException("skippedEntities is null");
+        }
+
+        if (skippedEntities == null) {
```
Ceiu: done.

server/src/main/java/org/candlepin/controller/ImportResult.java
```diff
@@ -0,0 +1,126 @@
...
+            throw new IllegalArgumentException("skippedEntities is null");
+        }
+
+        if (skippedEntities == null) {
```
Ceiu: done.

server/src/main/java/org/candlepin/controller/ImportResult.java
Ceiu: The first version actually was immutable, but after hitting another bug and reevaluating the usage of the maps, I opted to make it nothing more than a wrapper object around the maps since the import method isn't retaining them for state, and protecting them against modification doesn't (currently) buy us anything. Also, copying these maps unnecessarily is pretty expensive.

server/src/main/java/org/candlepin/controller/ImportResult.java
Ceiu: We'll still be doing this coalescing at some point (see the code removed from the end of the original importProducts/Content), this method just turns it into a lazy version of it.

server/src/main/java/org/candlepin/controller/ContentManager.java
Ceiu: Kinda mutually exclusive with the above, but if we opt against the previous bit, we can go this route.

nguyenfilip:
ack

PR #1398, opened 2016-11-23
1314042: bug caused by import and undo manifest run in parallel

pjonastik:
I wrote spec for this bug, which run (sync) importManifest (with updated manifest on brandings) and undoManifest in parralel.
    
When I experimented with mentioned spec I've never got same error like in bugzilla. It is becase we don't store subsrciption in DB anymore, but BZ should be covered by this fix.
I got different errors and deadlocks, so I locked owner at begining of both methods.
    
Fix's broken some unit tests so I changed those as well and made them little bit readable.

Note: For clarity, I moved ImportUpdateExporter from import_update_single_pool_spec to others exporters in candlepin_scenarios

### DIFF
--- a/server/spec/candlepin_scenarios.rb
+++ b/server/spec/candlepin_scenarios.rb
@@ -39,17 +39,16 @@ module CandlepinMethods
   end
 
   def create_product(id=nil, name=nil, params={})
-
     # If owner given in params, use it, if not, try to find @owner, if neither
     # is set error out.
     # NOTE: this is the owner key being passed in as a string
     if params[:owner]
-      owner = params[:owner]
+      owner_key = params[:owner]
     elsif @owner
-      owner = @owner['key']
+      owner_key = @owner['key']
     end
-    if ! owner
-      raise "Must call create_product with owner param or set @owner in spec suite."
+    if ! owner_key
+      raise "Must call create_product with owner (key) param or set @owner in spec suite."
     end
 
     # For purposes of testing, you can omit id and name to create with
@@ -61,7 +60,7 @@ module CandlepinMethods
     #Product IDs are 32 characters or less
     id = id[0..31]
 
-    product = @cp.create_product(owner, id, name, params)
+    product = @cp.create_product(owner_key, id, name, params)
     @created_products <<  product
     return product
   end
@@ -312,7 +311,6 @@ class Exporter
 
     @candlepin_client = consumer_client(owner_client, random_string('test_client'),
         "candlepin", user['username'])
-
   end
 
   def do_export(client, dest_dir, opts={}, uuid=nil)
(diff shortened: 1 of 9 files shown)


pjonastik:
Questions for reviewer: 
- What about add next spec example with import (async)? 
- What about add another spec examples with differrent manifest configuration (more/less subscriptions) ?

wottop:
Please correct the failure here:

Import and Undo Manifest in parallel - import start first should (sync) import and undo manifest in parallel successfully
     Failure/Error: fails.should be == 0

       expected: == 0
            got:    1
     # ./server/spec/import_and_undo_exports_in_parallel_spec.rb:149:in `block (2 levels) in <top (required)>'

pjonastik:
@wottop I am woriking on that. I only forgot put it label 'in progress' again.

pjonastik:
@wottop, done.

Ceiu:
Looks good, and seems to work. I'm curious what that lock does to performance, however. You should ping @vritant and ask him how to run the performance test job against this branch a few times to see if that lock is going to be a problem for performance.

As far as the questions you posed:
- If there is a concern that differing manifest sizes can change the result of the test, it may be worth the effort to parameterize the exporter/manifest generation and add a few more tests which just fiddle with the numbers
- I don't fully understand the question, but I'm assuming it has to do with the async import work. I'm fairly certain the owner lock will also block the async job all the same, so it shouldn't be any different here. A test which performs the import asynchronously wouldn't be a bad addition here to verify it.

pjonastik:
@Ceiu: I got performance regression over this branch. With regards to regression, my question is: do we really want to better fix for DEPRECATED  (sync) import operation? In addition, I am going to run another performance test job, when newer performence (master) info appears.

I admit my previous question is not very clear. I will try clarify it. It's true that owner lock will also block the async job all the same and also I know that wouldn't be bad addition. I wanted to add async import because I thnik this code should be cover as well, but my concern is about spec's execution time, which is not small (approximately 10sec). Async and sync import operations are locked in same point in code path, so right now creation of another  async-import-spec-test wouldn't bring anything and test execution would last longer. On the other hand, that spec test will catch possible errors, when code changes in the future.

Ceiu:
How much of a regression? If it's only a few hundred milliseconds, that's no big deal, but if we're doubling the runtime... that may be a problem -- even for a deprecated operation.

I would opt for a longer, more complete spec test over one that runs faster; especially since we're already pushing 30-40 minutes for a full spec test run anyway.

vritant:
@Ceiu I am happy with the performance results of this branch. Peter has run his branch 5 times, and the results are in a separate doc called "peters results" if you would like to take a look.
that being said, our tests dont talk to import export APIs . we have seperate jmeter tests for that if needed.

Ceiu:
ACKing based on @vritant's comment

PR #1400, opened 2016-11-28
Startup Check

nguyenfilip:
Candlepin will quickly fail at startup if the Qpid Broker is not running and healthy.

### DIFF
--- a/server/src/main/java/org/candlepin/audit/QpidQmf.java
+++ b/server/src/main/java/org/candlepin/audit/QpidQmf.java
@@ -233,7 +233,7 @@ public class QpidQmf {
      * @return Fully qualified names of queues
      * @throws JMSException Error connecting
      */
-    public Set<String> getExchangeBoundQueueNames(String exchangeName) throws JMSException {
+    private Set<String> getExchangeBoundQueueNames(String exchangeName) throws JMSException {
         List<Map<String, Object>> mm = runQuery("_schema_id", singularMap("_class_name", "binding"));
 
         Set<String> result = new HashSet<String>();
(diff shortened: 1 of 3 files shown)


nguyenfilip:
retest this please

server/src/main/java/org/candlepin/audit/QpidQmf.java
```diff
@@ -0,0 +1,330 @@
...
+        log.debug("Getting info about queue {}", queueName);
+        List<Map<String, Object>> mm =  runQuery("_object_id", singularMap("_object_name", queueName));
+        if (mm.size() == 0) {
+            throw new RuntimeException("Couldn't find a queue in Qpid: " + queueName);
+        }
+        if (mm.size() > 1) {
+            throw new RuntimeException("Found unexpected amount of information about queue: " + queueName);
+        }
+        return mm.get(0);
+    }
```
mstead: This should be private unless there are plans to use it outside the class.

server/src/main/java/org/candlepin/audit/QpidQmf.java
```diff
@@ -0,0 +1,330 @@
...
+
+    /**
+     * Map with one entry, key and value
+     * @param key
+     * @param value
+     * @return Map with one entry
+     */
+    public static Map<Object, Object> singularMap(Object key, Object value) {
```
mstead: This should be private unless there are plans to use it outside the class.

server/src/main/java/org/candlepin/audit/QpidQmf.java
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    /**
+     * Finds fully qualified names of queues that are bound to an exchange
+     *
+     * @param exchangeName
+     * @return Fully qualified names of queues
+     * @throws JMSException Error connecting
+     */
+    public Set<String> getExchangeBoundQueueNames(String exchangeName) throws JMSException {
```
nguyenfilip: ack

server/src/main/java/org/candlepin/audit/QpidQmf.java
nguyenfilip: ack

server/src/main/java/org/candlepin/audit/QpidQmf.java
nguyenfilip: ack

PR #1403, opened 2016-11-29
1399820: Load exported count lazily due to nested call from entitlements

barnabycourt:


mstead:
@barnabycourt I believe the version is incorrect in the referenced bug.

mstead:
I'll merge once that is addressed.

vritant:
This bug had a follow up second fix. is that needed here too?
we use hibernate 5 in master, hence the question.

vritant:
Nacking this fix on its own based on performance results. a performance test of just get consumers had a negligible improvement in results ( 14069 vs 14847 in master )

barnabycourt:
@vritant How many entitlements or referenced pools were available in the account used for testing? A much better fix for this issue generally would be to stop dynamically calculating exported count & consumed count dynamically

mstead:
We should port the 0.9.54 FieldHandler fix into master so that Lazy loading with '@Formula' is actually performed lazily.

PR #1404, opened 2016-11-30
Refactoring spec tests and spec utility

pjonastik:
Found useless parameter owner in method user_client_with_perms


awood:
requestbot, add to whitelist

PR #1405, opened 2016-11-30
1399356: No transaction in syncCRLWithDB caused immediate rollback

awood:
Futhermore, use EntityManager's flush instead of Session's flush so that
a missing transaction will cause a TransactionRequiredException to be
thrown instead of Session's flush which just fails silently.


Ceiu:
ACK from me for @awood's commit

PR #1407, opened 2016-12-01
Updating autobind disabled strings (0.9.54)

mstead:
Autobind is generally referred to as auto-attach.
Re-aligning client facing messages.


wottop:
visi-ack

PR #1410, opened 2016-12-02
1390214: Internal Server Error During Guest Migration

nguyenfilip:
The bug was caused by calling Arrays.asList(null) - which throws a nullpointer
exception.


nguyenfilip:
retest this please

wottop:
ack

PR #1412, opened 2016-12-05
Fail Fast

nguyenfilip:
New -q switch for deploy script that will recreate queue 'allmsg' which is a
special queue for testing purposes. -a switch will also cause autoconf
to enable AMQP in candlepin config.

Docker image candlepin-base now installs qpid and starts it using supervisord.

cp-test script has new -q switch that uses guterrball configure-qpid.sh
script, to setup qpid (generate SSL keys, create event exchange) etc.

I also had to modify configure-qpid.sh to be aware of supervisorctl

You can now write spec tests that assert Qpid queue. There is a new
CandlepinQpid helper class.

When connection to Qpid goes down, we must reestablish all the JMS Objects. This
behavior is dictated by JMS specification. Also, the timeout of Qpid JMS
failover had to be lowered to be able to easily, and quickly, detect broker
connection failure.

Qpid can be in three states that Candlepin needs to distinguish. First state is
'Qpid down' which is when connection to the Qpid is not available. Second state
is 'Flow Stopped'. That situation occurs when one of the queues bound to the
event exchange is flow stopped (see Flow Control in Qpid docs). Last is normal
happy 'connected' state in which Qpid is available. These states are detected
usign QMF (Qpid Management Framework)

Candlepin can enter Suspend Mode. In this mode Candlepin will only
respond to Status Resource requests and pause all Pinsetter Jobs. It is
expected that as soon as Candlepin enters Suspend Mode, no more new
requests or jobs are started. It is still possible that some of the
ongoing requests and jobs succeed.
When Candlepin is in Suspend Mode, it returns an error that is uniform
with the rest of the Candlepin. Client than can easily use /status/mode resource
to learn more details about the reason of suspend mode.

Candlepin will quickly fail at startup if the Qpid Broker is not running and
healthy.

Candlepin will now automatically transition to/from Suspend Mode. This is
achieved by class SuspendModeTransitioner that technically uses
ScheduledExecutorService thead pool (single threaded). The transitioner
periodically runs checks for Qpid connectivity but also exposes public method
that can be used to attempt transition. The transition ends up being a no-op if
no transition is needed or ends up transitioning Candlepin from/to suspend mode.

The periocity of the transitioner is using growing delay between executions. The
idea behind it is, that after the Qpid has been down for extensive period of time,
there is a high chance it won't come up soon. A config property is in place to
set maximum bound to the delay.

There is also new config property to turn off Suspend Mode.


nguyenfilip:
@mstead @dsarman I rebased and squashed all the commits in fail branch. For the reference, there is still a backup of the old development branch (fnguyen/fail-backup). 

Before merging this into master, it is important to realize that this branch requires everybody to install qpid-proton-devel RPM. This is a development requirement even if the developer  doesn't want to deploy with Qpid! The reason for this is that Qpid Spec tests require ruby Qpid bindings (to receive qpid messages during the test). But the only way to include ruby bindings with Buildr is to put them into the Gemfile (gem 'qpid_proton'). This Gem requires the RPM so everybody must install it for the Buildr to work. Before merging into master, I would like to send a day-ahead e-mail notice. And also update the development requirements cp.org page

With regards to what should be tested as part of the PR review.  I think its very important to test that the configuration switch for disabling suspend mode, works. It is an important feature that allows us to quickly disable the feature in case of trouble. Its not possible to spec test this switch, because our spec tests are not designed as to start/stop/reconfigure Candlepin.

nguyenfilip:
@mstead hm I think even better solution to qpid-proton-devel is to add the detection of that RPM directly to the script. Will do that

nguyenfilip:
@mstead ok the idea with the script won't work. Will need to document and let the team know

nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
The Qpid spec tests are not enabled for PRs against master. See the test results manually in the job:
job/candlepin-pullrequest-spectests-qpid/

nguyenfilip:
retest this please

nguyenfilip:
retest this please

nguyenfilip:
retest this please

barnabycourt:
please update the vagrant env setup to install qpid-proton-c so the vagrant env spins up cleanly

barnabycourt:
Configuring with qpid, if I stop the qpid server I see the message in the candlepin log "Entering new mode SUSPEND for reason QPID_DOWN" but the /candlepin/status endpoint still returns a 200 OK with no indication that things are in a bad state

nguyenfilip:
@barnabycourt Ok, I will add the info about status to candlepin/status. I will keep returning 200 (on /status)

mstead:
> Configuring with qpid, if I stop the qpid server I see the message in the candlepin log "Entering new mode SUSPEND for reason QPID_DOWN" but the /candlepin/status endpoint still returns a 200 OK with no indication that things are in a bad state.
> 
> If I check the "/candlepin/status/mode" endpoint everything the error state is displayed there however, our users are conditioned to look at the /candlepin/status endpoint

This is true for errors displayed by subscription manager as well.

```bash
$ sudo subscription-manager register --user admin --pass admin --org admin
Registering to: bluestar:8443/candlepin
Candlepin is in Suspend mode, please check /status/mode resource to get more details
```
As a user, I don't think I'd know what to do here.

nguyenfilip:
@mstead I think this is a different point than that raised by Barnaby.  Barnaby talks about /suspend mode. You are pointing out that Subman should present better message to the client? I can change the error message, but I think ultimately the subman should be changed to display good error when Candlepin retruns "SERVICE UNAVAILABLE"

barnabycourt:
@nguyenfilip if we are in suspend mode then I would expect that the request to get /status and the /status/mode would return a 503 Service Unavailable http code.

nguyenfilip:
@barnabycourt I can return 503 on /status and /status/mode. But to me that is confusing. I always thought that the HTTP code should say whether the specific HTTP request was successfull. /status and /status mode should, in my opinion, return 200 because it indicates that RETRIEVAL of the status and the suspend mode is successfull

mstead:
> @barnabycourt I can return 503 on /status and /status/mode. But to me that is confusing. I always thought that the HTTP code should say whether the specific HTTP request was successfull. /status and /status mode should, in my opinion, return 200 because it indicates that RETRIEVAL of the status and the suspend mode is successfull

I agree with @nguyenfilip here. 200 on the status call would indicate that the request `for status` succeeded.

docker/candlepin-base/cp-test
```diff
@@ -254,6 +255,15 @@ if [ "$UNITTEST" -gt 0 ]; then
     done
 fi
 
+if [ "$QPID" == "1" ]; then
+    echo "Setting up Qpid"
+    pushd . 
+    cd ../gutterball/bin/qpid
+    ./configure-qpid.sh
```
mstead: Since the -q is attempting to configure QPID, perhaps we should do a complete clean first, and then a configure.

```bash
# Do a cleanup
./configure-qpid.sh -c
# Reconfigure
./configure-qpid.sh
```

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
```diff
@@ -0,0 +1,232 @@
...
+    private void schedule() {
+        int delay = initialDelay * (int) Math.pow(delayGrowth, failedAttempts);
+
+        if (maxDelay != -1 && delay > maxDelay) {
+            log.debug("Maximum delay {} reached", maxDelay);
+            delay = maxDelay;
+        }
+
+        log.debug("Next Transitioner check will run after {} seconds", delay);
+        if (failedAttempts > 3) {
```
barnabycourt: Why wait until the 3rd attempt to log a message?

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
```diff
@@ -0,0 +1,232 @@
...
+     * grows as the failedAttempts grow. The formula is:
+     *
+     *  delay = initialDelay * (delayGrowth ^ failedAttempts)
+     *
+     *  Maximal delay is configurable, -1 means there is no bound
+     *
+     *  When delayGrowth is set to 1, the delay wont grow.
+     */
+    private void schedule() {
+        int delay = initialDelay * (int) Math.pow(delayGrowth, failedAttempts);
```
barnabycourt: math.pow returns a double, casting this to an int runs the risk of overflowing given the exponential growth based on the failed attempts.

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -339,9 +372,15 @@ private ConfigProperties() {
             this.put(AMQP_KEYSTORE_PASSWORD, "password");
             this.put(AMQP_TRUSTSTORE, "/etc/candlepin/certs/amqp/candlepin.truststore");
             this.put(AMQP_TRUSTSTORE_PASSWORD, "password");
+            this.put(QPID_STARTUP_CHECK_ENABLED, "true");
+            this.put(SUSPEND_MODE_ENABLED, "true");
+            this.put(QPID_QMF_RECEIVE_TIMEOUT, "5000");
+            this.put(QPID_MODE_TANSITIONER_DELAY_GROWTH, "2");
+            this.put(QPID_MODE_TRANSITIONER_INITIAL_DELAY, "10");
+            this.put(QPID_MODE_TRANSITIONER_MAX_DELAY, "1200"); // 1200 seconds = 20 minutes
```
barnabycourt: 20 minutes between checks is an awful long time. 5 minutes might be more appropriate.

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
```diff
@@ -0,0 +1,232 @@
...
+            delay = maxDelay;
+        }
+
+        log.debug("Next Transitioner check will run after {} seconds", delay);
+        if (failedAttempts > 3) {
+            log.info("SuspendModeTransitioner failed to reconnect to the Qpid Broker " +
+                "{} times, backing off with next reconnect {} seconds", failedAttempts,
+                delay);
+        }
+        execService.schedule(this, delay, TimeUnit.SECONDS);
```
mstead: Both @barnabycourt and I have noticed that the transitioner backs off way too quickly. To me it seems like we should apply the next calculated back off value to the delay after X failed connection attempts.

@nguyenfilip @barnabycourt  Thoughts?

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
```diff
@@ -0,0 +1,232 @@
...
+
+        if (maxDelay != -1 && delay > maxDelay) {
+            log.debug("Maximum delay {} reached", maxDelay);
+            delay = maxDelay;
+        }
+
+        log.debug("Next Transitioner check will run after {} seconds", delay);
+        if (failedAttempts > 3) {
+            log.info("SuspendModeTransitioner failed to reconnect to the Qpid Broker " +
+                "{} times, backing off with next reconnect {} seconds", failedAttempts,
```
barnabycourt: Is it useful to print the # of times that a reconnect has been attempted?

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
```diff
@@ -0,0 +1,232 @@
...
+                switch (status) {
+                    case CONNECTED:
+                        log.info("Connection to qpid is restored! Reconnecting Qpid and" +
+                            " Entering NORMAL mode");
+                        failedAttempts = 0;
+                        modeManager.enterMode(Mode.NORMAL, Reason.QPID_UP);
+                        break;
+                    case FLOW_STOPPED:
+                    case DOWN:
+                        failedAttempts++;
```
barnabycourt: There is no upper bound on failedAttempts, with poor configuration and a very negligent admin this will overflow and start counting down negative numbers.

barnabycourt:
I validated both the flow-stop and disconnect errors and in both cases this did an excellent job of reconnecting.

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
nguyenfilip: true. Fixed so that logs appear whenever there were failed attempts

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
nguyenfilip: Changed the logic so that the backoff is linear now. 

delay = initialDelay + (delayGrowth * failedAttempts)

Where default values are:
initalDelay = 10 seconds
delayGrowth = 10 seconds

maxDelay changed to 5 minutes

nguyenfilip: True, changed everything that could overflow to BigInteger

server/src/main/java/org/candlepin/config/ConfigProperties.java
nguyenfilip: changed to 5 min

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
nguyenfilip: changed to linear, see comments above

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
nguyenfilip: I kind of like it. Its easy to then search in the logs for e.g. the first time the transitioner failed to reconnect etc.

server/src/main/java/org/candlepin/controller/SuspendModeTransitioner.java
nguyenfilip: Its BigInteger now

nguyenfilip:
Latest qpid spec test job run job/candlepin-pullrequest-spectests-qpid/70

Note that after this PR is merged to master, the Job can be enabled to run automatically as pat of the PR checks

nguyenfilip:
@mstead @barnabycourt GET /status  now returns 503 and standard Candlepin ERROR message when in suspend mode:

```
{
  "displayMessage" : "Candlepin is in Suspend mode, please check /status/mode resource to get more details",
  "requestUuid" : "3efc228c-b147-4143-b7a9-30e775154e30"
}
```
Do you think this is reasonable? Or was the idea that GET /status will return 503 and normal response? 

GET /status/mode returns 200 and the mode information:

```
{
  "mode" : "SUSPEND",
  "reason" : "QPID_DOWN",
  "changeTime" : "2016-12-15T08:42:53+0000"
}
```

nguyenfilip:
@barnabycourt fixed and tested the vagrant image

mstead:
Verified that scheduled jobs are paused when qpid goes down, and resume when it comes back up.

:+1:

mstead:
> @mstead @barnabycourt GET /status now returns 503 and standard Candlepin ERROR message when in suspend mode:
> 
> {
>   "displayMessage" : "Candlepin is in Suspend mode, please check /status/mode resource to get more details",
>   "requestUuid" : "3efc228c-b147-4143-b7a9-30e775154e30"
> }
> 
> Do you think this is reasonable? Or was the idea that GET /status will return 503 and normal response?
> 
> GET /status/mode returns 200 and the mode information:
> 
> {
>   "mode" : "SUSPEND",
>   "reason" : "QPID_DOWN",
>   "changeTime" : "2016-12-15T08:42:53+0000"
> }

Seems fine to me.

nguyenfilip:
retest this please

nguyenfilip:
retest this please

mstead:
@nguyenfilip Please translate any exception messages that may make it to the client. While clients should be handling the exceptions and customizing the messages, until they are updated, they may surface. Subscription manager is prime candidate for this.

mstead:
My review for this PR is complete.

**To address:**
1) Address the translations

**Thoughts/Comments**
I'm still wondering if the back-off on failure functionality is really required. We are polling qpid status every 10 secs (configurable) when qpid is up. On failure, why do we need to back off and wait longer than this? Is there something about checking status when QPID is down, or in flow-stop, that is concerning? Logging can be turned down to DEBUG if it is excessive. Is checking the status taxing on qpid?

nguyenfilip:
@mstead translation added to two occurences of SuspendException

nguyenfilip:
@barnabycourt caching bits added. I chose to put the clearing of the cache into the SuspendModeTransitioner. Alternatively it could have been added to ModeManager

barnabycourt:
test this please

barnabycourt:
@nguyenfilip The unit tests were not updated for the addition of the CacheManager for the SuspendModeTransitioner constructor

nguyenfilip:
@barnabycourt fixed

PR #1413, opened 2016-12-07
The Product and Content managers now clean up orphaned entities

Ceiu:
- Product and Content entities orphaned by update operations are now
  properly cleaned up at the end of the update operation
- Unit test within suites extending the DatabaseTestFixture are now automatically
  run within a transaction
- DatabaseTestFixture no longer throws IllegalStateExceptions if attempting
  to start a transaction when one is already active, or when committing or
  rolling back when a transaction is not active
- Some unit tests have been updated in accordance with the above changes
- Updated the async job call in candlepin_api to also abort on jobs in
  the "failed" state
- SubscriptionReconciler has been moved to the Importer, as it only needs
  to be performed during import
- Optimized product version lookups
- Added an additional spec test for verifying product retrevials

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -402,8 +402,10 @@ class Candlepin
     post("/jobs/scheduler", status)
   end
 
-  def trigger_job(job)
-    post("/jobs/schedule/#{job}")
+  def trigger_job(job, async=false)
+    return async_call(!async) do
+      post("/jobs/schedule/#{job}")
+    end
   end
 
   def create_consumer_type(type_label, manifest=false)
@@ -624,6 +626,10 @@ class Candlepin
     get("/owners/#{owner_key}/content/#{content_id}")
   end
 
+  def get_content_by_uuid(content_uuid)
+    get("/content/#{content_uuid}")
+  end
+
   def delete_content(owner_key, content_id)
     delete("/owners/#{owner_key}/content/#{content_id}")
   end
@@ -759,6 +765,10 @@ class Candlepin
 
   def get_product(owner_key, product_id)
     get("/owners/#{owner_key}/products/#{product_id}")
+  end
+
+  def get_product_by_uuid(product_uuid)
+    get("/products/#{product_uuid}")
   end
 
   def delete_product(owner_key, product_id)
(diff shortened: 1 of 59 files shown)


Ceiu:
I think this is ready to go. The only area of concern is the product resolution code in CandlepinPoolManager.refreshPoolsForMasterPool may be in the wrong place. However, since Hibernate tends to do its writes whenever it feels like it, pushing it further into the path may not be viable or possible.

barnabycourt:
When this scales out, what happens if there are two manifest imports in separate organizations going on at the same time. 
Starting State:
Org A has 1x  Premium
Org B has 1x  Standard 

Org A & Org B start 2 concurrent imports:
Org A imports manifest with 1x RHEL Standard
Org B imports a manifest with 1x RHEL Premium

During the import is there a risk that one or both of the existing products (Premium,  Standard) are marked as deleted and cleaned up by one import while the 2nd import tries to associate with that product? 

More importantly, at production scale where there may be hundreds or thousands of pool refreshes going on at a given time, what are the risks of having the orphan cleanup inline with the refresh. 

If the above is not a concern then excellent. If it is, then how can we work around it.

Ceiu:
@barnabycourt I'm assuming that scenario is a problem, and one we won't be able to solve without locking on the products (and content) directly. More worrying, is I believe it exists in master today with single updates at the API level.

Ceiu:
Turns out we never converge on orphaned products and content, as the version lookup uses an inner join on OwnerProducts, preventing orphans from being retrieved during that step. For extra insurance against this, I've moved the orphan cleanup to a separate job which runs once/week (every Sunday at 3am by default), and fetches the soon-to-be-deleted orphans with a pessimistic write lock before deleting them.

Hopefully with all of these pieces in place, we should be safe from any convergence shenanigans.

Ceiu:
retest this please

Ceiu:
retest this please

Ceiu:
There's currently a issue where running the full test suite causes something (likely the DB) to deadlock, causing every spec test to time out waiting to get the Candlepin status.

Ceiu:
Deadlock issue has been resolved (for real, this time). Should be good to go.

barnabycourt:
Looks good, I'd like a 2nd review from someone else who has been in this code a lot lately.

mstead:
@Ceiu Needs a rebase

mstead:
@barnabycourt, @Ceiu  This seem good to me. 

Once jenkins finishes up we can merge.

PR #1414, opened 2016-12-09
1403160: update product without id in the body

vritant:


Ceiu:
What is this addressing? I'd argue that the user sending an ID-less product should correct their input or we should update the API to not care about the product ID on the endpoint. That aside, it looks like it won't actually break anything.

Ceiu:
ACK

vritant:
@Ceiu agreed that this appears un-ideal, but untill 2.0.19 we accepted this input, and from that perspective its a regression.

PR #1416, opened 2016-12-09
CandlepinQueryInterceptor now uses a separate session for cursors

Ceiu:
- CandlepinQueryInterceptor now uses its own session for streaming the
  query results
- Added additional spec tests for testing streaming certain entities


Ceiu:
This works around the issue of our resultsets closing when we execute explicit secondary queries to perform lazy loading or cache resolution.

vritant:
doesnt seem to work.

PR #1417, opened 2016-12-12
1394401: List of provided products empty on Bonus Pools

nguyenfilip:
The problem is in method PoolHelper.copyProvidedProducts. The method is used to
copy the original pool to a bonus pool. The method
expects the original pool (the one being copied) to be stored in the database.
It is trying to find its provided products in the database. That fails, because
the original pool is not yet there.

Adding more defensive code into that method. First I will check whether the
pool is already persisted (Pool.id != null). When it is not persisted, I will
copy the provided products directly from the original instance.

### DIFF
--- a/server/src/main/java/org/candlepin/policy/js/pool/PoolHelper.java
+++ b/server/src/main/java/org/candlepin/policy/js/pool/PoolHelper.java
@@ -157,21 +157,56 @@ public class PoolHelper {
     }
 
     /**
-     * Copies the provided products from a source pool to a derived pool.
-     * It assumes that source Pool is already in database and that the provided Products or
-     * derived provided Products are linked with the Pool in the database!
+     * Copies the provided products from a source pool to a bonus pool. The
+     * logic is not completely straightforward.
+     *
+     * The source has so called 'main product' (source.getProduct()) and also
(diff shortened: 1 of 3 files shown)


nguyenfilip:
retest this please

nguyenfilip:
Note for myself: need to add more Javadoc about why derived provided products are sent to the bonus pool

nguyenfilip:
retest this please

server/src/main/java/org/candlepin/policy/js/pool/PoolHelper.java
```diff
@@ -157,21 +157,56 @@ private PoolHelper() {
...
+     * logic is not completely straightforward.
+     *
+     * The source has so called 'main product' (source.getProduct()) and also
+     * derived product source.getDerivedProduct(). It also has 'main provided products'
+     * source.getProvidedProducts().
+     *
+     * During the copy, the following takes place.
+     *
+     * If the source has derived product, then the destination will receive source's
+     * main product as main product (destination.getProduct). Also, it will receive
```
wottop: still needs the line above.

nguyenfilip: fixed

wottop:
ack

PR #1419, opened 2016-12-13
1378593: concurrent content creates

vritant:


barnabycourt:
@vritant Wouldn't this still have risk of collision when Candlepin is being run in cluster?

vritant:
Good point, Ive updated the code. if this makes sense, I can fix master as well, as it was a backport.

barnabycourt:
Only thing that needs to be double checked is that the system will work well with IDs that are uuid based as opposed to integer based (even if they are represented as strings). If everything checks out on this then please forward port to master as well.

server/spec/content_resource_spec.rb
```diff
@@ -56,4 +56,19 @@
     # Verify the serial has changed (the entitlement has been regenerated)
     original_serial.should_not == new_serial
   end
+
+  it 'should handle concurrent content creates' do
+    threads = []
+    for i in 0..300
```
awood:
```ruby
300.times do
  threads << Thread.new do 
    @cp.create_content("cname", nil, random_string("clabel"), "ctype", "cvendor", {}, true)
  end
...
```

PR #1420, opened 2016-12-13
DetachedCandlepinQuery now better restores Criteria state

Ceiu:
- The DetachedCandlepinQuery object now properly/better restores Criteria
  state after calls to the .getRowCount method


Ceiu:
This should address the issue where certain queries weren't returning the proper results when paging was applied. If there are still other issues, I have an idea for a more in-depth modification.

Ceiu:
@vritant I've changed up the code a bit, so if you've looked at this at all, give the code another glance.

vritant:
@Ceiu would have liked some tests with this, but not going to block a merge for weeks for that reason, Ive tested manually and am satisfied.

PR #1423, opened 2016-12-15
1372467: Activation keys and service levels

wottop:
Test to ensure that an activation key can utilize pools with differing service
 levels to attach for a consumer


Ceiu:
ACK

PR #1426, opened 2016-12-16
1352956: update checkin_time when Hypervisor is checked in via virt-who

vritant:


vritant:
closing cause we dont need it in this branch

PR #1429, opened 2016-12-20
Batch Revoke Performance Fix (v3)

mstead:
Removed the useage of ProductCachedSerializationModule in
EventFactory while serializing pools since we use the
HATEOS JSON version of Pool when serializing for Events.

Removing this serializer produces the exact same JSON
data. No need to hydrate provided products as it is
never, and has never, been included.

On revoke entitlement, do not remove entitlements
from their pool's collection. Since the collection
is Extra lazy, hibernate will query the ents each
time an entitlement is removed from a Pool's collection.

By doing this, we are no longer maintaining runtime
consistency b/w Pools and Entitlements. The biggest
concern with doing this is that we may run into a
hibernate unscheduling problem. For example a missed
delete. To minimize this risk, this PR removes the cascades
on the Pool.entitlements collection.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -1717,7 +1717,6 @@ public class CandlepinPoolManager implements PoolManager {
             int entQuantity = ent.getQuantity() != null ? ent.getQuantity() : 0;
 
             pool.setConsumed(pool.getConsumed() - entQuantity);
-            pool.getEntitlements().remove(ent);
             Consumer consumer = ent.getConsumer();
             if (consumer.getType().isManifest()) {
                 pool.setExported(pool.getExported() - entQuantity);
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1717,7 +1717,6 @@ public void revokeEntitlements(List<Entitlement> entsToRevoke, Set<String> alrea
             int entQuantity = ent.getQuantity() != null ? ent.getQuantity() : 0;
 
             pool.setConsumed(pool.getConsumed() - entQuantity);
-            pool.getEntitlements().remove(ent);
```
Ceiu: If we're going to drop the runtime consistency from this operation anyway, it might be faster to just craft three update queries for removing the entitlements and update the pool/consumer consumed/exported values.

Ceiu: After further consideration, we've opted not to continue with explicit deletion queries here, as the lack of runtime consistency combined with the complexity of the revoke/refresh operation itself potentially opens the door to hard-to-debug consistency and querying issues further down the path. We'll consider revisiting direct queries again in the future if this area continues to be a performance bottleneck.

Ceiu:
ACK

PR #1430, opened 2016-12-20
1371009: clearer error message (for 0.9.54)

pjonastik:
1371009:Need clearer error message when register with system name exceeding max characters.

-adding unit test

**Change from previous fix**: I cherry-picked commit from cp-0.9.23 and then changed generating of consumer name w.r.t @crog comment in /candlepin/candlepin/pull/1356. Another difference is that changing buildfile was not necessery, because buildfile has already corrected version of org.xnap.commons.i18n.


pjonastik:
On branch candlepin-0.9.54 we've already had failing spec test in virt_spec.rb:
  *Standalone Virt-Limit Subscriptions should revoke guest entitlements when migration happens

Ceiu:
ACK

PR #1431, opened 2016-12-20
Correction for unit tests

wottop:
Failures due to creating expired pools during December.


awood:
Jenkins, retest this please

PR #1432, opened 2016-12-21
1371009:  clearer error message (for master)

pjonastik:
Need clearer error message when register with system name exceeding max characters.

-adding unit test

**Change from previous fix:** I cherry-picked commit from cp-0.9.23 and then changed generating of consumer name w.r.t @crog comment in /candlepin/candlepin/pull/1356. Another difference is that changing buildfile was not necessery, because buildfile has already corrected version of org.xnap.commons.i18n.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -596,11 +596,18 @@ public class ConsumerResource {
      */
     private void checkConsumerName(Consumer consumer) {
         // for now this applies to both types consumer
-        if (consumer.getName() != null &&
-            consumer.getName().indexOf('#') == 0) {
-            // this is a bouncycastle restriction
-            throw new BadRequestException(
-                i18n.tr("System name cannot begin with # character"));
+        if (consumer.getName() != null) {
+            if (consumer.getName().indexOf('#') == 0) {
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -596,11 +596,18 @@ else if (update.getFact("distributor_version") !=  null) {
...
-            throw new BadRequestException(
-                i18n.tr("System name cannot begin with # character"));
+        if (consumer.getName() != null) {
+            if (consumer.getName().indexOf('#') == 0) {
+                // this is a bouncycastle restriction
+                throw new BadRequestException(
+                    i18n.tr("System name cannot begin with # character"));
+            }
+
+            int max = Consumer.MAX_LENGTH_OF_CONSUMER_NAME;
```
Ceiu: Just a note, but it's not necessary to declare the variable locally to access the constant.

Ceiu:
ACK

server/src/main/java/org/candlepin/resource/ConsumerResource.java
pjonastik: @Ceiu : I know. Checkstyle was shouting on me, because of length of line, when local variable was inlined.

PR #1434, opened 2016-12-29
Fix guest migration and hypervisor async job

vritant:


barnabycourt:
@vritant Is there a functional test that could be written to catch this in addition to the unit test?

vritant:
@barnabycourt done

vritant:
Closing this as it is no longer relevant since we reverted to 0.9.54 behaviour.
the one useful bit here is added in #1462

PR #1435, opened 2016-12-29
Fix NPE when selectBestPools returns no pools and debug logging is on

vritant:


awood:
Looks fine to me, but a small request: would you mind changing the commit message to "Fix NPE when selectBestPools returns no pools and debug logging is on" so that it will all fit within the 70 characters that GitHub kind of expects in its interface?

vritant:
@awood done

awood:
Looks like the unit test failures are legit.  (At least I see them on my machine in this branch and not in master).  Should just be a matter of correcting the test assertions in 
org.candlepin.controller.PoolManagerFunctionalTest and org.candlepin.policy.AutobindRulesTest.  E.g.:

```
    [junit] Testcase: instanceAutobindForPhysical8SocketNotEnoughEven took 0.057 sec
    [junit] 	FAILED
    [junit] expected:<null> but was:<[]>
```

barnabycourt:
@vritant If this is targetting 0.9.54, then it needs a  BZ for tracking in releasing an errata.

PR #1438, opened 2017-01-10
1411778: log on service level mismatch between guest and host

vritant:


server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -219,7 +219,10 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
             if (host != null && (force || host.isAutoheal())) {
                 log.info("Attempting to heal host machine with UUID \"{}\" for guest with UUID \"{}\"",
                     host.getUuid(), consumer.getUuid());
-
+                if (!StringUtils.equals(host.getServiceLevel(), consumer.getServiceLevel())) {
+                    log.warn("Host's service level \"{}\" does not match that of the guest's \"{}\"",
```
barnabycourt: If this is running at warning and may be output even if the msg on 220 that includes the UUIDs is not output (since it is at info log level) then the UUIDs should be included in the warning msg

Ceiu:
ACK

PR #1439, opened 2017-01-10
1405125: Remove null byte from end of any fact value.

awood:
This is primarily intended to address ppc64 virt uuids that end with a
null byte, but in principle we do not want any facts to end in null
bytes or even whitespace.


Ceiu:
ACK

It looks like the only way we can set facts for a given consumer today is via API calls that deserialize consumers, which would trigger the filtering mechanism. The only remaining concern would be if a null char/byte makes its way into the middle of the string.

PR #1442, opened 2017-01-11
1371009: clearer error message (fixed typo) for cp-0.9.54

pjonastik:
Fixing typo in error message.
Previous message: Name of the consumer should be shorter than 255 characters.
Fixed message : Name of the consumer should be shorter than 256 characters.


Ceiu:
ACK
Spec failure is a periodic failure that's been fixed in master. Ignoring.

PR #1446, opened 2017-01-17
docker tag no longer supports -f tag

Lorquas:
docker tag no longer supports the -f flag, this is causing an error on the push part of this script, breaking the jenkins job.


Ceiu:
ACK

PR #1447, opened 2017-01-19
1414426: Ability to get future pools by owner

wottop:
Either future only or active and future together.

To test this function you will need to use the two new query parameters 'add_future' and 'only_future' in conjunction with the 'activeon' parameter.

'activeon' determines what 'now' is for the other parameters. When not passed in, the current date will be used. If add_future is set to true, then you will get back any pool whose starting date is after the 'activeon' date in addition to those that the method previously returned. If only_future is set to true, then you will only get back the future date pools.

You may not use both add_future and only_future at the same time. It will return a bad request exception.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/PoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/PoolManager.java
@@ -181,7 +181,6 @@ public interface PoolManager {
      * @param productId only entitlements which provide this product are included.
      * @param activeOn Indicates to return only pools valid on this date.
      *        Set to null for no date filtering.
-     * @param activeOnly if true, only active entitlements are included.
      * @param includeWarnings When filtering by consumer, include pools that
      *        triggered a rule warning. (errors will still be excluded)
      * @param filterBuilder builds and applies all filters when looking up pools.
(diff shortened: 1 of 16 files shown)


barnabycourt:
Please include test instructions in the PR description.

server/src/main/java/org/candlepin/controller/PoolManager.java
```diff
@@ -190,7 +190,8 @@ Entitlement adjustEntitlementQuantity(Consumer consumer, Entitlement entitlement
      */
     Page<List<Pool>> listAvailableEntitlementPools(Consumer consumer, ActivationKey key,
         Owner owner, String productId, String subscriptionId, Date activeOn, boolean activeOnly,
-        boolean includeWarnings, PoolFilterBuilder filterBuilder, PageRequest pageRequest);
+        boolean includeWarnings, PoolFilterBuilder filterBuilder, PageRequest pageRequest,
```
barnabycourt: if activeOnly is never set to False, then this parameter should be stripped out from the entire stack.

wottop:
buildbot: retest this please

barnabycourt:
If add_future and only_future are exclusionary, would it make sense to use an enumerated value?
entitlements_included=[current, current_and_future, future_only]?

wottop:
It got pushed into 51 already with the separate variables.

PR #1448, opened 2017-01-19
Modify server.xml to keep tomcat-native from biting developers.

awood:
The AprLifecycleListener attempts to load the Apache Portable Runtime
(APR).  The APR is a native library used to speed up certain operations.
In our case, loading the APR causes an issue because the APR requires
additional attributes in the Connector element (e.g. SSLCertificateFile)
since the APR uses OpenSSL and PEM files instead of Java and JKS files.
See https://tomcat.apache.org/tomcat-8.0-doc/apr.html

In Fedora, the tomcat-native package installs the APR and the tomcat
package pulls in tomcat-native as a suggested dependency.  Rather than
break developers every time they update Tomcat, the easier approach is
just to disable the AprLifecycleListener.

### DIFF
--- a/server/bin/update-server-xml.py
+++ b/server/bin/update-server-xml.py
@@ -29,7 +29,7 @@ class AbstractBaseEditor(object):
         self.doc = doc
 
     @abc.abstractproperty
-    def insertion_xpath(self):
+    def parent_xpath(self):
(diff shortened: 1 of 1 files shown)


Ceiu:
ACK

PR #1450, opened 2017-01-24
1412113: autobind fails when binding 2 pools together with same stackid

vritant:
To test, deploy master, checkout this branch and run:
buildr rspec:autobind_spec.rb:"succeeds"

the same test passes in this branch.

### DIFF
--- a/server/spec/autobind_spec.rb
+++ b/server/spec/autobind_spec.rb
@@ -1,0 +1,104 @@
+require 'spec_helper'
+require 'candlepin_scenarios'
+
+describe 'Autobind On Owner' do
+
+  include CandlepinMethods
+
+  it 'succeeds when requesting bind of muliple pools with same stack id' do
+  owner_key = random_string('test_owner')
+  owner = create_owner owner_key
+
+  # create 4 products with the same stack id and sockets.
+  prod = create_product('taylorid', 'taylor swift', {:owner => owner_key, :version => "6.1",
+    :attributes => {
+      "sockets" => "2",
+      "vcpu" => "4",
+      "warning_period" => "30",
+      "brand_type" => "OS",
+      :stacking_id => "ouch"}})
+  prod1 = create_product(nil, nil, {:owner => owner_key, :attributes => {
+    :stacking_id => "ouch",
+    "virt_limit" => 1,
+    "sockets" => 1,
+    "instance_multiplier" => 2,
+    "multi-entitlement" => "yes",
+    "host_limited" => "true"}})
+
+  prod2 = create_product(nil, nil, {:owner => owner_key, :attributes => {
+    :stacking_id => "ouch",
+    "virt_limit" => 1,
+    "sockets" => 1,
+    "instance_multiplier" => 2,
+    "multi-entitlement" => "yes",
+    "host_limited" => "true"}})
+  prod3 = create_product(nil, nil, {:owner => owner_key, :attributes => {
+    :stacking_id => "ouch",
+    "virt_limit" => 1,
+    "sockets" => 1,
+    "instance_multiplier" => 2,
+    "multi-entitlement" => "yes",
+    "host_limited" => "true"}})
+
+  # create 4 pools, all must provide product "prod" . none of them
+  # should provide enough sockets to heal the host on it's own
+  create_pool_and_subscription(owner['key'], prod['id'], 10)
+  create_pool_and_subscription(owner['key'], prod1['id'], 20, [prod['id']])
+  create_pool_and_subscription(owner['key'], prod2['id'], 20, [prod['id']])
+  create_pool_and_subscription(owner['key'], prod3['id'], 20, [prod['id']])
+
+  # create a guest with "prod" as an installed product
+  guest_uuid =  random_string('guest')
+  guest_facts = {
+    "virt.is_guest"=>"true",
+    "virt.uuid"=>"myGuestId",
+    "cpu.cpu_socket(s)"=>"1",
+    "virt.host_type"=>"kvm",
+    "system.certificate_version"=>"3.2"
+  }
+  guest = @cp.register('guest.bind.com',:system, guest_uuid, guest_facts, 'admin',
+    owner_key, [], [{"productId" => prod.id, "productName" => "taylor swift"}])
+
+  # create a hypervisor that needs 40 sockets and report the guest with it
+  hypervisor_facts = {
+    "virt.is_guest"=>"false",
+    "cpu.cpu(s)"=>"4",
+    "cpu.cpu_socket(s)"=>"40"
+  }
+  hypervisor_guests = [{"guestId"=>"myGuestId"}]
+  hypervisor_uuid = random_string("hypervisor")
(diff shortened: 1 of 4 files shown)


server/spec/autobind_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+  @cp.update_consumer({:uuid => hypervisor_uuid, :facts => hypervisor_facts})
+
+  @cp.consume_product(nil, {:uuid => guest_uuid})
+  @cp.list_entitlements({:uuid => hypervisor_uuid}).length.should == 3
+  @cp.list_entitlements({:uuid => guest_uuid}).length.should == 1
+  @cp.list_owner_pools(owner_key).length.should == 8
+
+  end
+
+end
```
mstead: The autobind_disabled_for_owner tests targeted a particular portion of autobind. If you feel strongly about moving them into an umbrella suite, then please rename the tests that were moved to show that they are testing autobind disabled for owner functionality.

vritant: I dont feel about this strongly. ill move it to another file.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2082,12 +2082,59 @@ void handleBonusPools(Owner owner, Map<String, PoolQuantity> pools,
...
+            // others must be operated individually
+            for (Entry<String, Entitlement> entry: nonUniqueEnts.entrySet()) {
+                Entitlement ent = entry.getValue();
+                Map<String, Entitlement> entMap = new HashMap<String, Entitlement>();
+                entMap.put(entry.getKey(), ent);
+                Set<String> stackIdSet = new HashSet<String>();
+                if (ent.getPool().isStacked()) {
+                    stackIdSet.add(ent.getPool().getStackId());
+                }
+                handlePostEntitlement(manager, consumer, entMap, stackIdSet);
```
mstead: Is there a way that we could group the ents in a stack and operate on the groupings instead of going one by one?

mstead: Doesn't necessarily mean that we have to, but, just a thought.

vritant: @mstead not without updating underlying code, which would need to work on an entitlement one by one to check if a sub pool anyway.

mstead: Just seems like if we have a group of ents of the same stack, then we should be able to update the subpool in one call. I'm probably making this seem a little easier than it actually is.

Might be nice to have some tests around that filtering code, if it isn't already covered by other tests.

server/src/test/java/org/candlepin/controller/PoolManagerTest.java
```diff
@@ -866,27 +868,64 @@ public void testCreateAndEnrichPoolForPool() {
...
+            assertEquals((Integer) 1, count);
+        }
+    }
 
-        assertEquals(2, total);
-        verify(entitlementCurator, never()).listModifying(any(Entitlement.class));
-        //TODO assert batch revokes have been called
+    @Test
+    public void testRevokeAllEntitlements() {
     }
```
mstead: Empty test.

vritant:
@mstead should pass now!

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2082,12 +2082,61 @@ void handleBonusPools(Owner owner, Map<String, PoolQuantity> pools,
...
+                    String stackId = entitlement.getPool().getStackId();
+                    Integer count = stackIdCounts.containsKey(stackId) ?
+                        stackIdCounts.get(stackId) : 0;
+                    stackIdCounts.put(stackId, count + 1);
                 }
             }
+            Set<String> uniqueStackIds = new HashSet<String>();
+            Map<String, Entitlement> uniqueEnts = new HashMap<String, Entitlement>();
+            Map<String, Entitlement> nonUniqueEnts = new HashMap<String, Entitlement>();
+            for (Entry<String, Entitlement> entry : entitlements.entrySet()) {
```
Ceiu: In the future, we should look for ways to avoid stepping through a collection more than once (or three times, as is the case here). Here, we could either use a map of lists to pull the counts out later, or assume everything is going into the uniqueEnts set until we find a collision, where both would be moved to the nonUniqueEnts.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2082,12 +2082,61 @@ void handleBonusPools(Owner owner, Map<String, PoolQuantity> pools,
...
+            }
+            // ents with unique stack id can be operated in a batch
+            if (!uniqueEnts.isEmpty()) {
+                handlePostEntitlement(manager, consumer, uniqueEnts, uniqueStackIds);
+            }
+            // others must be operated individually
+            for (Entry<String, Entitlement> entry: nonUniqueEnts.entrySet()) {
+                Entitlement ent = entry.getValue();
+                Map<String, Entitlement> entMap = new HashMap<String, Entitlement>();
+                entMap.put(entry.getKey(), ent);
```
Ceiu: Tip: you can use Collections.<String, Entitlement>singletonMap(entry.getKey(), ent)) to save some boiler plate and tiny bits of memory and overhead here. singletonMap returns an immutable (!) map of the given type with the mapping passed in. The immutable nature means it can't always be used, but as a result there's no bucket management so everything is slightly faster and easier on resources.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
Ceiu: What's a typical size of entitlements that'll be handled here? We should expect 250ms+ of connection overhead just looking up the sub pools (and then entitlements again (!)) in this call. Also, that lookup has somehow evaded my gaze up to this point and isn't a CandlepinQuery, so we can't take advantage of any of the scrolling bits it provides (yet).

It'd be nice if we could start batching up some of these DB operations so we can cut down on the overhead when we're stepping through collections like this. In fact, as I drill down further into this, I would argue the whole thing needs a proper evaluation. We're taking entitlements, fetching stack IDs, looking up sub pools, then looking up the entitlements again from the stack IDs. There's a lot of redundancy here.

vritant:
@barnabycourt  is this an ASAP candidate?

mstead:
@vritant @Ceiu Should we address Rog's comments before this is merged, or are we happy with it?

vritant:
@mstead we need to discuss., I've been meaning to discuss this, but was occupied otherwise. lets discuss this the first chance we get.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
vritant: @barnabycourt I will discuss this with  @Ceiu  online.

PR #1451, opened 2017-01-24
Added additional warnings to the per-org products migration task

Ceiu:
- The per-org products migration task now prints warnings when it
  detects a pool or subscription that references a product that
  does not exist

### DIFF
--- a/server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationTask.java
+++ b/server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationTask.java
@@ -168,6 +168,81 @@ public class PerOrgProductsMigrationTask extends LiquibaseCustomTask {
         }
     }
 
+
+    /**
+     * Checks for any pools or subscriptions which contain bad data (typically products which do
+     * not exist).
+     *
+     * @param orgid
+     *  The id of the owner/organization for which to migrate product data
+     */
+    protected void checkForMalformedPoolsAndSubscriptions(String orgid) throws DatabaseException,
+        SQLException {
+
+        ResultSet badProductRefs = this.executeQuery(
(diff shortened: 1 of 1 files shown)


Ceiu:
*Testing:*
- Deploy candlepin-0.9.54 with test data
- In the database, change one or more of the following:
  - Change productid or derivedproductid in cp_pool to a bad value (e.g. 'deadproduct')
  - Add or change a row in cp_pool_products to reference a non-existent product
  - Change product_id or derivedproduct_id in cp_subscription to a bad value
  - Add or change a row in cp_subscription_products to reference a non-existent product
  - Add or change a row in cp_sub_derivedprods to reference a non-existent product
- Checkout this branch (crog/bad_pool_product_ref_warning)
- Deploy in verbose mode (./bin/deploy -v 2>&1 | tee log.txt)
- Verify that an error message was printed once for each pool or subscription modified

*Notes:*
This change only alerts users to a bad database state; it does nothing to resolve the issue itself. The migration will still continue without error and the pools will still exist post-migration, but depending on where the issue existed (cp_pool or cp_pool_products), the pool may act like it was deleted, or it may act like the bad product reference was deleted. It's slightly inconsistent, but it's repeatable, repairable and relatively safe.

barnabycourt:
@Ceiu Does this execute after the product data has been loaded in hosted mode?

Ceiu:
This would be during the actual migration.

barnabycourt:
@Ceiu Do we only skip the pools with a bad primary product or will they end up dead if one of the provided or derived provided products is skipped as well?

Ceiu:
As of right now, neither are skipped, but the former case will act like the pool itself is deleted, while the latter will be migrated without the bad references.

barnabycourt:
ack, migrating without including the bad references for provided & derivedProvided products is a good call.

PR #1455, opened 2017-01-26
Pinhead counting feature

pjonastik:
- This feature should be similar like listConsumer API, so it should return equal count of consumers.

- Feature was firstly implemented SQL, then with JPQL and at the end with hibernate criteria.
  Perfromance test didn't show significant difference between these implementation. Hibernate criteria
  was used because secureCriteria are used in listConsumer method and to use JPQL we would have to make more effort.

- Unit test added and refactored in ConsumerCuratorTest to test DB layer and OwnerResourceTest to test business layer.
  Some older portion of code in OwnerResourceTest was refactored as well.
  Notice methods which only passthrough collections.  These are only helpers for better readability and less verbosity.
      Consider these 2 approaches:

```java
        1)
            List<String> typeLabels = null;
            List<String> skus = null;
            List<String> ids = null;
            List<String> contracts = null;
            int count = consumerCurator.countConsumers(owner.getKey(), typeLabels, 
                                 skus, ids, contracts);
```
```java    
        2)
             int count = consumerCurator.countConsumers(owner.getKey(), typeLabels(null),
                                 skus(null), ids(null), contracts(null));
```
   
 - Spec test added. Refactored.

### DIFF
--- a/server/spec/owner_resource_spec.rb
+++ b/server/spec/owner_resource_spec.rb
@@ -870,3 +870,150 @@ describe 'Owner Resource Future Pool Tests' do
   end
 
 end
+
+describe 'Owner Resource counting feature' do
+
+  include CandlepinMethods
+
+  before(:each) do
+    @owner = create_owner(random_string("test_owner"))
+    @owner_cp = user_client(@owner, random_string('user_name'))
+  end
+
+  it 'should count all consumers of given owner' do
+    @owner_cp.register('consumer_name1')
+    @owner_cp.register('consumer_name2')
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'])
+    expect(Integer(json_count)).to be == 2
+
+    other_owner = create_owner(random_string("test_owner"))
+    other_cp = user_client(other_owner, random_string('bill'))
+    other_cp.register('consumer_name3')
+
+    json_count = @cp.count_owner_consumers(other_owner['key'])
+    expect(Integer(json_count)).to be == 1
+  end
+
+  it 'should count only consumers specified by type label' do
+    @owner_cp.register('consumer_name1', type=:system)
+    @owner_cp.register('consumer_name2', type=:hypervisor)
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], type=[:system])
+    expect(Integer(json_count)).to be == 1
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], type=[:system, :hypervisor])
+    expect(Integer(json_count)).to be == 2
+  end
+
+  it 'should count consumers only with specific skus' do
+    c = @owner_cp.register('consumer_name')
+    sku1 = create_product_and_bint_it_to_consumer_return_sku(c)
+    sku2 = create_product_and_bint_it_to_consumer_return_sku(c)
+    sku3 = create_consumer_with_binding_to_new_product_return_sku
+    expect(sku1).not_to be(sku2)
+    expect(sku2).not_to be(sku3)
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], ["not-existing-sku"])
+    expect(Integer(json_count)).to be == 0
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [sku1])
+    expect(Integer(json_count)).to be == 1
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [sku1, sku2])
+    expect(Integer(json_count)).to be == 1
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [sku1, sku2, sku3])
+    expect(Integer(json_count)).to be == 2
+  end
+
+  it 'should count consumers only with specific subscriptionIds' do
+    c = @owner_cp.register('consumer_name')
+    subId1 = create_product_and_bint_it_to_consumer_return_subId(c)
+    subId2 = create_product_and_bint_it_to_consumer_return_subId(c)
+    subId3 = create_consumer_with_binding_to_new_product_return_subId
+    expect(subId1).not_to be(subId2)
+    expect(subId2).not_to be(subId3)
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [], ["not-existing-subId"])
+    expect(Integer(json_count)).to be == 0
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [], [subId1])
+    expect(Integer(json_count)).to be == 1
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [], [subId1, subId2])
+    expect(Integer(json_count)).to be == 1
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [], [subId1, subId2, subId3])
+    expect(Integer(json_count)).to be == 2
+  end
+
+  it 'should count consumer only with specific contract numbers' do
+    c = @owner_cp.register('consumer_name')
+    cn1 = create_product_and_bint_it_to_consumer_return_contractNr(c)
+    cn2 = create_product_and_bint_it_to_consumer_return_contractNr(c)
+    cn3 = create_consumer_with_binding_to_new_product_return_contractNr
+    expect(cn1).not_to be(cn2)
+    expect(cn2).not_to be(cn3)
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [], [], ["not-exisitng-cn"])
+    expect(Integer(json_count)).to be == 0
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [], [], [cn1])
+    expect(Integer(json_count)).to be == 1
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [], [], [cn1, cn2])
+    expect(Integer(json_count)).to be == 1
+
+    json_count = @owner_cp.count_owner_consumers(@owner['key'], [], [], [], [cn1, cn2, cn3])
+    expect(Integer(json_count)).to be == 2
+  end
+
+  def create_consumer_with_binding_to_new_product_return_sku(product_type='MKT')
+    c = @owner_cp.register('consumer_name')
+    create_product_and_bint_it_to_consumer_return_sku(c, product_type)
+  end
+
+  def create_product_and_bint_it_to_consumer_return_sku(consumer, product_type='MKT')
+    params = {:attributes => {'type' => product_type}}
+    p = create_product(@owner['key'], params)
+
+    pool = create_pool_and_subscription(@owner['key'], p.id, 1)
+    @owner_cp.consume_pool(pool.id, params={:uuid => consumer.uuid, :quantity => 1})
+    p.id #sku
+  end
+
+  def create_consumer_with_binding_to_new_product_return_subId
+    c = @owner_cp.register('consumer_name')
+
+    create_product_and_bint_it_to_consumer_return_subId(c)
+  end
+
+  def create_product_and_bint_it_to_consumer_return_subId(consumer)
+    p = create_product(@owner['key'])
+    pool = create_pool_and_subscription(@owner['key'], p.id, 1)
+    @owner_cp.consume_pool(pool.id, params={:uuid => consumer.uuid, :quantity => 1})
+    pool.subscriptionId
+  end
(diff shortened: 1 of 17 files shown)


server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -799,6 +800,86 @@ public Consumer verifyAndLookupConsumerWithEntitlements(String consumerUuid) {
         return listByCriteria(crit, pageRequest);
     }
 
+    public int countConsumers(String ownerKey,
+        Set<String> typeLabels, List<String> skus, List<String> subscriptionIds, List<String> contracts) {
```
Ceiu: Try not to use specific collections here unless it's absolutely required by the implementation.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -799,6 +800,86 @@ public Consumer verifyAndLookupConsumerWithEntitlements(String consumerUuid) {
...
+     */
+    private Criteria createCriteria(String ownerKey, Set<String> typeLabels,
+        List<String> skus, List<String> subscriptionIds,
+        List<String> contracts) {
+
+        Criteria crit = super.createSecureCriteria("c");
+        crit.createAlias("c.owner", "o")
+            .add(Restrictions.eq("o.key", ownerKey));
+
+        if (hasSomeNotNullString(typeLabels)) {
```
Ceiu: The hasSomeNotNullString is interesting. With sets it's basically a free operation, but with lists, it requires a linear traversal of the entire list. In the general case, it'll end up being a deoptimization for not a lot of gain. In the case of our query, a null value in the IN collection doesn't break anything (so long as the collection itself isn't null or empty).

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -799,6 +800,86 @@ public Consumer verifyAndLookupConsumerWithEntitlements(String consumerUuid) {
...
+
+        boolean hasSkus = hasSomeNotNullString(skus);
+        boolean hasSubscriptionIds = hasSomeNotNullString(subscriptionIds);
+        boolean hasContracts = hasSomeNotNullString(contracts);
+        if (hasSkus || hasSubscriptionIds || hasContracts) {
+            crit.createAlias("c.entitlements", "e").createAlias("e.pool", "po");
+        }
+
+        if (hasSkus) {
+            crit.createAlias("po.product", "pr").createAlias("pr.attributes", "pa")
```
Ceiu: Indentation here

```
crit.createAlias("po.product", "pr")
    .createAlias("pr.attributes", "pa")
    .add(Restrictions.in("pr.id", skus))
    .add(Restrictions.eq("pa.name", "type"))
    .add(Restrictions.eq("pa.value", "MKT"));
```

server/spec/owner_resource_spec.rb
```diff
@@ -823,3 +823,154 @@
...
+  it 'should count consumer only with specific contract numbers' do
+    c = @owner_cp.register('consumer_name')
+    cn1 = create_product_and_bint_it_to_consumer_return_contractNr(c)
+    cn2 = create_product_and_bint_it_to_consumer_return_contractNr(c)
+    cn3 = create_consumer_with_binding_to_new_product_return_contractNr
+    expect(cn1).not_to be(cn2)
+    expect(cn2).not_to be(cn3)
+
+    result = @owner_cp.count_owner_consumers(
+              @owner['key'], [], [], [],["not-exisitng-cn"])
```
Ceiu: A lot of these could fit on one line, no?

pjonastik: Yeah. Good catch. Done.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
pjonastik: In case, when we have correct values plus null values everything is good, but in case when collections contain only null values then we are bad. Without removing nulls from collections, test named : 'blankCriteriaShouldCountAllExistingConsumers' won't pass.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
pjonastik: Done.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
pjonastik: Done.

server/src/test/java/org/candlepin/test/DatabaseTestFixture.java
```diff
@@ -303,6 +304,42 @@ protected Pool createPool(Owner owner, Product product, Long quantity, Date star
         return poolCurator.create(pool);
     }
 
+    protected Pool createPool(Owner owner, Product product, Long quantity, Date startDate, Date endDate,
```
pjonastik: Done.

server/src/test/java/org/candlepin/test/DatabaseTestFixture.java
```diff
@@ -507,4 +544,114 @@ private static void insertValidationEventListeners(Injector inj) {
...
+
+    protected List<String> subscriptionIds(List<String> subscriptionIds) {
+        return subscriptionIds;
+    }
+
+    protected List<String> contracts(List<String> contracts) {
+        return contracts;
+    }
+
+    protected String ownerKey(String key) {
```
pjonastik: I was rethinking this and it should really not be in superclass, but what if I had moved these to ConsumerCuratorTest, would have you still disagreed? Consider count of lines when every unit test must have declared null variables, it's more verbose.

server/src/test/java/org/candlepin/test/DatabaseTestFixture.java
```diff
@@ -507,4 +544,114 @@ private static void insertValidationEventListeners(Injector inj) {
...
+    protected Product createMktProductAndBindItToConsumer(Owner owner, Consumer c) {
+        String id = String.valueOf(TestUtil.randomInt());
+        Product product = createTransientMarketingProduct(id);
+        product = createProduct(product, owner);
+        Pool pool = createPool(owner, product);
+        createEntitlement(owner, c, pool, null);
+        return product;
+    }
+
+    protected Product createTransientMarketingProduct(String productId) {
```
pjonastik: Ok, I will move all utilities back to ConsumerCuratorTest for now, but in case of createTransientMarketingProduct, I can imagine to refactor class ConsumerCuratorSearchTest by using this particular method.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -689,6 +687,36 @@ public void deleteLogLevel(@PathParam("owner_key") String ownerKey) {
...
+        @QueryParam("type") Set<String> typeLabels,
+        @QueryParam("sku") List<String> skus,
+        @QueryParam("subscription_id") List<String> subscriptionIds,
+        @QueryParam("contract") List<String> contracts) {
+
+        findOwner(ownerKey);
+        findAndValidateTypeLabels(typeLabels);
+
+        int count = consumerCurator.countConsumers(ownerKey, typeLabels, skus, subscriptionIds, contracts);
+        return "{ \"consumersCount\" :  " + count + "}";
```
pjonastik: Done.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -689,6 +687,36 @@ public void deleteLogLevel(@PathParam("owner_key") String ownerKey) {
         return page.getPageData();
     }
 
+    private List<ConsumerType> findAndValidateTypeLabels(
```
pjonastik: Agree. I did it, but in ConsumerResource was same code for validation consumerTypeLabels, so I created ConsumerTypeValidator class and moved logic there, but I am still not satisfied with this because I think this should be also injected by guice instead creating new instance in constructor, what do you think?

server/src/main/java/org/candlepin/resource/util/ConsumerTypeValidator.java
```diff
@@ -0,0 +1,72 @@
...
+    public List<ConsumerType> findAndValidateTypeLabels(Set<String> labels) {
+        if (labels != null && !labels.isEmpty()) {
+            List<ConsumerType> types = consumerTypeCurator.lookupByLabels(labels);
+            validate(types, labels);
+            return types;
+        }
+        return null;
+    }
+
+    private void validate(List<ConsumerType> types, Set<String> labels) {
```
Ceiu: We can combine all of these into one method -- there's no reason to break these up, as they're never used elsewhere and don't really simplify the operation at all.

server/src/main/java/org/candlepin/resource/util/ConsumerTypeValidator.java
```diff
@@ -0,0 +1,72 @@
...
+        // Since the type labels are unique, our sizes must match.
+        if (labels.size() != types.size()) {
+            List<String> invalidLabels = findInvalidLabels(labels, types);
+            throw new BadRequestException(i18n.tr("No such unit type(s): {0}",
```
Ceiu: Not entirely comfortable with this. Returning pass-through UI exceptions from a non-UI component. This is probably fine for now, but will definitely be a candidate for change in the future.

server/src/main/java/org/candlepin/resource/OwnerResource.java
Ceiu: The abstraction to a separate class is a decent first step, but definitely add it to the DI framework if we're going to be using it this way.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: I believe this is only an issue with hsqldb, Oracle, MySQL/MariaDB and PostgreSQL all have no issues with null values in a collection used with the IN operator.

pjonastik: I removed collectionsWithNullValuesShouldCountAllExistingConsumers. Hsqldb works, test was wrong. And I 've also changed hasSomeNotNullString method to isNotNullOrEmpty.

server/src/main/java/org/candlepin/resource/util/ConsumerTypeValidator.java
pjonastik: I like coding this way even private methods are used only once. Because you can easily read the few lines in the most top method and immediately understand method without digging into details. This has another benefit and that is the methods are short and don't do much. With this technique you can easily discover methods which are not cohesive and shouldn't belong to class. 
But if you would like to see methods together I have no problem to do it.

server/src/main/java/org/candlepin/resource/OwnerResource.java
pjonastik: Done.

server/src/main/java/org/candlepin/resource/util/ConsumerTypeValidator.java
pjonastik: Yeah I know. This require more work, but at least we pulled out to one layer higher. I left it as it was for now.

server/src/main/java/org/candlepin/resource/util/ConsumerTypeValidator.java
Ceiu: You can leave it as is if you'd prefer.

server/src/main/java/org/candlepin/resource/util/ConsumerTypeValidator.java
Ceiu: ack

server/src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: If the new method does what the name implies, you may be able to replace it with the CollectionUtils.isEmpty method from Apache Commons, or even the "collection != null && !collection.isEmpty()" one-liner.

Ceiu:
Need a rebase to fix the conflicts, but we may be ready to go once that's done.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
pjonastik: Changed with CollectionUtils.isEmpty.

pjonastik:
Confilcts resolved.

PR #1456, opened 2017-01-27
1416823: Abstracted out property validation to new validator framework (0.9.49)

Ceiu:
- Added the new PropertyValidator class and its subclasses
  AttributeValidator and FactValidator, which provide implementations
  for validating attributes and facts, respectively
- Updated the ProductCurator and ConsumerCurator to use the new
  PropertyValidator instances in place of their old private validation
  methods
- ConsumerValidator no longer massages inbound fact data, but will
  now, instead, throw PropertyValidationExceptions when it encounters
  malformed or otherwise unmanagable data
- ConsumerResource now provides the massaging and filtering
  functionality previously performed in the ConsumerCurator

### DIFF
--- a/server/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/server/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -118,6 +118,11 @@ public class ConfigProperties {
     };
 
     public static final String SYNC_WORK_DIR = "candlepin.sync.work_dir";
+
+    /**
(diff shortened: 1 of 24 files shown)


mstead:
This looks pretty good overall. I like the change.  I'd like to see FactValidator have its own set of tests, instead of being buried in other test classes.

Ceiu:
@mstead Done.

mstead:
retest this please

PR #1461, opened 2017-02-01
Made some optimizations to ConsumerCurator

Ceiu:
- A handful of methods in ConsumerCurator have been updated to return
  CandlepinQuery objects rather than returning Pages or Lists
- Affected resource methods have been updated to pass CandlepinQuery
  objects received from the ConsumerCurator from their endpoints


wottop:
ack

PR #1464, opened 2017-02-03
1419193: Fixed an issue with bulk insert statements during migration

Ceiu:
- Fixed an issue with the cp2 per-org products migration that would
  cause it to generate prepared statements with too many parameters
  if run against a deployment with a large enough number of products
  or content


Ceiu:
**Testing:**
- Checkout a pre-2.0 Candlepin branch
- Create an owner and any number of pools; 1 is acceptable here, but more works, too.
- Create a large number (10k+) of products and content
- Attach the content to the products in any desired distribution
- Attach the products to the pools in any desired distribution
- **backup your database**
- Checkout current CP2 master and deploy. The deployment should fail with an obscure "connection has closed" message.
- Restore the database or clear the databasechangeloglock entry by hand
- Checkout this branch and deploy. The deployment should succeed.

This should work on both MySQL and Postgres... and probably Oracle.

mstead:
ack

PR #1468, opened 2017-02-10
Reverted the use of cursors in the heal entire org job

Ceiu:
- Reverted the use of cursors in the heal entire org job when
  pulling consumers for a given org.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/tasks/HealEntireOrgJob.java
+++ b/server/src/main/java/org/candlepin/pinsetter/tasks/HealEntireOrgJob.java
@@ -75,9 +75,7 @@ public class HealEntireOrgJob extends UniqueByEntityJob {
 
             Date entitleDate = (Date) map.get("entitle_date");
 
-            // TODO:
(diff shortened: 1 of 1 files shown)


vritant:
@Ceiu  some context please? what guidelines should we use to decide when to use cursors and when not to? a comment in the code would be nice to prevent someone from adding the cursor back in the future.

Ceiu:
@vritant Generally speaking, we should be using them when (a) the query has the potential to return a massive amount of data which would take a noticeable amount of time to pull into memory (assuming we can fit it all in memory to begin with) and (b) the operation will not be doing another query on the same connection for each row.

In the case of b, we can usually get by with inserts, updates and deletes (i.e. things that don't tell the DB to fetch more data), but with Hibernate even that could be risky if it decides to do some internal state updating after it does whatever write operation you requested.

In this particular case, I know for a fact we're doing a ton of other queries for each consumer there, so there's no way this will work at all; and if it does, I'll be very surprised. I need to check whether or not we're explicitly testing this job, because I strongly feel like this should be failing hard with the typical "resultset is closed" error.

PR #1470, opened 2017-02-10
Moved transaction around autobind to after retry logic

Ceiu:
- Moved the transaction for CandlepinPoolManager.entitleByProducts
  to after the retry logic, which will allow many paths to release
  pool locks between attempts when retrying

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -1039,35 +1039,12 @@ public class CandlepinPoolManager implements PoolManager {
     // NOTE: after calling this method both entitlement pool and consumer
     // parameters will most certainly be stale. beware!
     @Override
-    @Transactional
-    public List<Entitlement> entitleByProducts(AutobindData data)
-        throws EntitlementRefusedException {
-        Consumer consumer = data.getConsumer();
-        String[] productIds = data.getProductIds();
-        Collection<String> fromPools = data.getPossiblePools();
-        Date entitleDate = data.getOnDate();
-        Owner owner = consumer.getOwner();
-
+    public List<Entitlement> entitleByProducts(AutobindData data) throws EntitlementRefusedException {
         int retries = MAX_ENTITLE_RETRIES;
 
         while (true) {
             try {
-                List<PoolQuantity> bestPools = new ArrayList<PoolQuantity>();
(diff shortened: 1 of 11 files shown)


Ceiu:
This is a minor adjustment to what should probably be a bigger refactor job to make sure we're applying this behavior to every relevant code path, and not just specific bind requests.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1097,6 +1074,45 @@ public Pool find(String poolId) {
...
+     * @Transactional annotation to function.
+     *
+     * @param data
+     *  The autobind data to use for entitling a consumer
+     *
+     * @return
+     *  a list of entitlements created as for this autobind operation
+     */
+    @Transactional
+    protected List<Entitlement> entitleByProductsImpl(AutobindData data) throws EntitlementRefusedException {
```
barnabycourt: If this should never be called directly, shouldn't it be private?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1097,6 +1074,45 @@ public Pool find(String poolId) {
...
+        Consumer consumer = data.getConsumer();
+        String[] productIds = data.getProductIds();
+        Collection<String> fromPools = data.getPossiblePools();
+        Date entitleDate = data.getOnDate();
+        Owner owner = consumer.getOwner();
+
+        List<PoolQuantity> bestPools = new ArrayList<PoolQuantity>();
+        // fromPools will be empty if the dev pool was already created.
+        if (consumer != null && consumer.isDev() && !fromPools.isEmpty()) {
```
barnabycourt: Under what condition will the consumer be null?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
Ceiu: The Transactional annotation doesn't work with private methods, and I'm not sure we want to introduce the boilerplate it does to do the same work with private methods.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
Ceiu: Autobind data not having a consumer? Not sure, the block is almost entirely as it was before I moved it.

PR #1475, opened 2017-02-15
Convert calculated pool quantities to columns

wottop:
Adds columns and migrates data on db update.
Programmatically increments and decrements pool
quantities based on entitling.

### DIFF
--- a/server/spec/entitlement_resource_spec.rb
+++ b/server/spec/entitlement_resource_spec.rb
@@ -193,4 +193,99 @@ describe 'Entitlement Resource' do
     end.should raise_exception(RestClient::Forbidden)
   end
 
+  it 'should handle concurrent requests to pool and maintain quanities' do
+    owner_client = user_client(@owner, random_string('owner'))
+    prod = create_product(random_string('product'), random_string('product'),
+      {:attributes => { :'multi-entitlement' => 'yes'},
+       :owner => @owner['key']})
+    pool = create_pool_and_subscription(@owner['key'], prod.id, 50)
+
+    t1 = Thread.new{register_and_consume(pool, "system", 5)}
+    t2 = Thread.new{register_and_consume(pool, "candlepin", 7)}
+    t3 = Thread.new{register_and_consume(pool, "system", 6)}
+    t4 = Thread.new{register_and_consume(pool, "candlepin", 11)}
+    t1.join
+    t2.join
+    t3.join
+    t4.join
+
+    consumed_pool = owner_client.get_pool(pool.id)
+    consumed_pool['consumed'].should == 29
+    consumed_pool['exported'].should == 18
+
+  end
+
+  it 'should not allow over consumption in pool' do
+    owner_client = user_client(@owner, random_string('owner'))
+    prod = create_product(random_string('product'), random_string('product'),
+      {:attributes => { :'multi-entitlement' => 'yes'},
+       :owner => @owner['key']})
+    pool = create_pool_and_subscription(@owner['key'], prod.id, 3)
+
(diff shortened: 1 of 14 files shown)


barnabycourt:
@wottop The pool quantity will need to be modified on entitlement deletion as well as creation.

wottop:
CandlepinPoolManager revokeEntitlements

wottop:
buildbot: retest this please

mstead:
@wottop I'm not sure if it exists yet or not, but I'd like to see a spec test that concurrently binds to a pool and ensures the correct quantities result. Something like a pool with quantity 10 and 5 or 6 threads binding.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -996,4 +996,22 @@ public boolean exists(Pool pool) {
...
+
+    public void calculateExportedForOwnersPools(Owner owner) {
+        String stmt = "update Pool p set p.exported = coalesce(" +
+            "(select sum(ent.quantity) FROM Entitlement ent, Consumer cons, ConsumerType ctype " +
+            "where ent.pool.id = p.id and ent.consumer.id = cons.id and cons.type.id = ctype.id " +
+            "and ctype.manifest = 'Y'),0) where p.owner = :owner";
+        Query q = currentSession().createQuery(stmt);
+        q.setParameter("owner", owner);
+        q.executeUpdate();
+    }
```
mstead: Should get some coverage around these methods in PoolCuratorTest

mstead:
retest this please

server/spec/entitlement_resource_spec.rb
```diff
@@ -193,4 +193,36 @@
...
+    sleep 1
+    t2 = Thread.new{register_and_consume(pool, "candlepin", 7)}
+    sleep 1
+    t3 = Thread.new{register_and_consume(pool, "system", 6)}
+    sleep 1
+    t4 = Thread.new{register_and_consume(pool, "candlepin", 11)}
+    t1.join
+    t2.join
+    t3.join
+    t4.join
```
mstead: Are the sleeps necessary here? Thread.join will make the main thread wait for all the binds to complete.

mstead:
@wottop I'd also be curious as to how this would behave in a multi-threaded case where a pool's available quantity has been exhausted. Is there a possibility of a negative quantity?

vritant:
remove default mark

barnabycourt:
@wottop Looks like there is an error in the migration for mysql that could result in a null value for the quantity_consumed & exported

barnabycourt:
@mstead have all your questions been answered?

mstead:
> @wottop I'd also be curious as to how this would behave in a multi-threaded case where a pool's available quantity has been exhausted. Is there a possibility of a negative quantity?
> 
> Create a pool and make enough concurrent binds to exhaust the pools quantity. Might be tricky to get the test right.

@wottop I'd like to see this in there if we can as I think it is a very important test case.

wottop:
the test 'should end at zero quantity consumed when all consumers are unregistered' was my attempt at that.

mstead:
> the test 'should end at zero quantity consumed when all consumers are unregistered' was my attempt at that.

Yep. Good point. How about the case where we could possibly over consume?

wottop:
'should not allow over consumption in pool' test

mstead:
> 'should not allow over consumption in pool' test

Ugh... I suck. Thanks for pointing that out.

Looks good to me then.

barnabycourt:
Performance test systems were having problems when this last test run was done. I ran them manually and they passed.

PR #1476, opened 2017-02-16
POC for moving the pool lock to the end of entitlement creation.

barnabycourt:


barnabycourt:
This will deadlock due to a shared-row-lock created on the pool during other actions within the transaction.

PR #1477, opened 2017-02-16
Added the test_migration script

Ceiu:
- Added the new test_migration.sh script, which tests deploying
  a given version of Candlepin (specified via gitref), then
  migrates to another version without clearing the database
- Updated the deploy script to wait for Candlepin to respond rather
  than a blind sleep

### DIFF
--- a/server/bin/import_test_data.rb
+++ b/server/bin/import_test_data.rb
@@ -8,6 +8,14 @@ require 'date'
 require 'json'
 require 'pp'
 
+# Impl note:
+# We use print instead of puts throughout this script, as a majority of the operations are
+# performed in parallel threads. Internally, puts prints the message and an automatic line break as
+# two separate, non-atomic operations. This allows time for the threads to print their messages to
+# the same line before the line break is written. We work around this here by using print with a
+# line break in the message itself, reducing/eliminating the possibility for two+ threads to write
+# a message to the same line.
+
 
 SMALL_SUB_QUANTITY = 5
 LARGE_SUB_QUANTITY = 10
@@ -25,7 +33,7 @@ data = {'products'=> [], 'content'=> [], 'owners'=> [], 'users'=> [], 'roles'=> 
 @sourceSubId = 0
 
 filenames.each do |filename|
-  puts filename
+  # puts filename
   product_data_buf = File.read(filename)
   product_data = JSON(product_data_buf, {})
   data['products'] = data.fetch('products') + product_data['products'] unless product_data['products'].nil?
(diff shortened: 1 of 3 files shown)


Ceiu:
This is mostly complete, but I still want to add some extra bits to automatically find the server directory (or, specifically, the deploy script) and change directories appropriately. That aside, it's largely finished. However, it's probably mostly garbage and poorly written, so feel free to grill me on conventions or generally just things I've done poorly.

Ceiu:
I've added the automatic search and cd for the Candlepin directory.

server/bin/import_test_data.rb
```diff
@@ -189,12 +189,14 @@ def create_product(cp, owner, product)
...
-  puts "product name: " + name + " version: " + version + \
-       " arch: " + arch + " type: " + type
+  product_ret = cp.create_product(owner['name'], id, name, {
+    :multiplier => multiplier,
+    :attributes => attrs,
+    :dependentProductIds => dependent_products,
+    :relies_on => relies_on
+  })
+
+  print "product name: " + name + " version: " + version + " arch: " + arch + " type: " + type + "\n"
```
awood: Use interpolation here.  It will read better.

`print "product name: #{name} version: #{version} arch: #{arch} type: #{type}\n"`

server/bin/test_migration.sh
```diff
@@ -0,0 +1,183 @@
...
+}
+
+collect_artifacts() {
+    # If the caller mounted a volume at /artifacts, copy server logs out:
+    if [ -d "/artifacts" ]; then
+        echo "Collecting artifacts..."
+
+        # It's entirely possible for these to not exist, so we'll copy them if we can, but if we
+        # fail, we shouldn't abort
+        move_artifact '/var/log/candlepin/access.log' '/artifacts/'
```
awood: I would DRY this up a little?

```
for x in "access.log" "audit.log" "candlepin.log" #etc
do
   move_artifact "/var/log/candlepin/${x}" "/artifacts"
done
```

server/bin/test_migration.sh
```diff
@@ -0,0 +1,183 @@
...
+        cd $CPDIR
+        CP_HOME=$(pwd)
+
+        echo "Using Candlepin instance found at: $CP_HOME"
+        break
+    fi
+done
+
+if [ -z "$CP_HOME" ]; then
+    echo "ERROR: Unable to locate Candlepin deploy script. Please run this script from the Candlepin server directory"
```
awood: `err_msg`

server/bin/import_test_data.rb
```diff
@@ -37,12 +37,13 @@
...
 
+print "\nCreating owners\n"
 def create_owner(cp, new_owner)
   owner_name =  new_owner['name']
   displayName = new_owner['displayName']
 
-  puts "owner: #{owner_name}"
-  puts "\t displayName: #{displayName}"
+  print "owner: #{owner_name}\n"
+  print "\t displayName: #{displayName}\n"
```
Ceiu: puts isn't atomic and triggers two print operations (message and new line), where print is (at least, as far as the printing bits go). This is why while importing, we'll occasionally see two threads print on the same line with a new line between them, rather than printing on their own lines.

I had fixed this once a couple years back, but it got lost/reverted somewhere along the way.

awood: Ah!  Good to know.  Mind putting a comment to that effect?

Ceiu: Done

PR #1478, opened 2017-02-16
Uber certificate rewrite

mstead:
# Description
A rewrite of this functionality was done to
remove the complexity from the generation
process and supporting data model used to
create the certificate.

In the new approach, uber certificates are
stored in their own table, and no longer
require 'dummy' consumers/pools/constent/product
stored in the DB in order to get a certificate
for an Owner.

We still only allow one uber cert per owner
but the cert is now directly related to the
Owner record and has no ties to any other
entity.

As part of this patch, all legacy uber cert
functionality has been removed and replaced
with the new.

# Testing Basic Candlepin Uber Cert Functionality
**Candlepin Testing Notes**
- Exercise the GET /owners/:key/uebercert API
- Exercise the POST /owners/:key/uebercert API
  * Ensure that a new cert is generated each time.

# Testing The Basic Use Case On A Satellite Instance

**NOTES ON TESTING BASIC CASE FOR SATELLITE:**
We will be using the uber certificate to browse content of an owner with the browser. When testing this out, we will generate the uber certs on a local candlepin (we'll set up the same CA as a Satellite) and we'll test them out on a different instance of Satellite running 0.9.54. This is the easiest way to do this until Sat folk get a deployable version of Sat6 with CP 2.0
 
It took a bit of fiddling around to get this to work, so I hope I didn't miss anything. If something in these test steps does not work, let me know and I can help out.

**Setting Up Satellite**
- Stand up a satellite with forklift (doesn't matter what version)
- Create two Orgs (make note of the org/owner keys)
- For each Org
  - Select the org in the upper left dropdown
  - Generate a manifest using stage, and import it into the Org
  - Navigate to 'Content -> Red Hat Repositories' and enable one of the repos.
  - Navigate to 'Content -> Sync Plans' and create a new plan
    * Click the Add tab, select a product and click the 'Add Selected' button.
  - Click the 'Run Sync Plan' button to sync content to your satellite.
  - Navigate to 'Content -> Sync Status' and wait for the content to fully sync

**Setting up local candlepin**
- SSH into the satellite and SCP the CA and KEY used by candlepin over to your local candlepin server (same destination as the location on the satellite).
  * /etc/pki/katello/private/katello-default-ca.key
  * /etc/pki/katello/certs/katello-default-ca-stripped.crt
- Assign the 'tomcat' group to katello-default-ca.key
- Add the following to the candlepin conf and restart tomcat making sure that candlepin starts without error.
  * candlepin.ca_key=/etc/pki/katello/private/katello-default-ca.key
  * candlepin.ca_cert=/etc/pki/katello/certs/katello-default-ca-stripped.crt
- deploy candlepin with this patch
- Create two Owners in you local candlepin (ensure the owner_keys match those of the Orgs labels you created)

**Creating and Testing The Certs**
 - On your local CP, generate an uber cert for each owner:
  - ./client/ruby/cpc generate_ueber_cert <YOUR_OWNER_KEY> <FILE_NAME>
- Create a cert to be used in your browser for each uber cert:
  - Create a key.pem and cert.pem from the downloaded ueber certificate
  - Create a pfx file that can be imported to the browser. 
	openssl pkcs12 -keypbe PBE-SHA1-3DES -certpbe PBE-SHA1-3DES -export -in cert.pem -inkey key.pem -out [NAME].pfx -name [NAME]
- Browse to "https://192.168.121.8/pulp/repos/:YOUR_OWNER_KEY/" (You should get a security error)
- Import the cert (pfx file) (for the owner YOUR_OWNER_KEY) into your browser of choice.
- Try accessing the above URL again. This time you should see a dir listing for the repo.
- Try accessing the above URL but with the other owner key. You should get another security error.
- Import the key for the other owner, and try again. Make sure that you can see the content.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -336,16 +336,11 @@ public class CandlepinPoolManager implements PoolManager {
             this.refreshPoolsForMasterPool(pool, false, lazy, updatedProducts);
         }
 
-        Pool ueberPool = this.findUeberPool(owner);
-        String ueberPoolId = ueberPool != null ? ueberPool.getId() : null;
-
         // delete pools whose subscription disappeared:
         log.debug("Deleting pools for absent subscriptions...");
         List<Pool> poolsToDelete = new ArrayList<Pool>();
         for (Pool pool : poolCurator.getPoolsFromBadSubs(owner, subscriptionMap.keySet())) {
-            if (pool.getSourceSubscription() != null && !pool.getType().isDerivedType() &&
-                (ueberPoolId == null || !ueberPoolId.equals(pool.getId()))) {
-
+            if (pool.getSourceSubscription() != null && !pool.getType().isDerivedType()) {
                 poolsToDelete.add(pool);
             }
         }
@@ -1424,23 +1419,9 @@ public class CandlepinPoolManager implements PoolManager {
     public List<Entitlement> entitleByPools(Consumer consumer, Map<String, Integer> poolQuantities)
         throws EntitlementRefusedException {
         if (MapUtils.isNotEmpty(poolQuantities)) {
-            return addOrUpdateEntitlements(consumer, poolQuantities, null, false, CallerType.BIND);
+            return addOrUpdateEntitlements(consumer, poolQuantities, null, CallerType.BIND);
         }
         return new ArrayList<Entitlement>();
-    }
-
-    @Override
-    public Entitlement ueberCertEntitlement(Consumer consumer, Pool pool)
-        throws EntitlementRefusedException {
-        Map<String, Integer> poolQuantities = new HashMap<String, Integer>();
-        poolQuantities.put(pool.getId(), 1);
-        List<Entitlement> result = addOrUpdateEntitlements(consumer, poolQuantities, null, true,
-            CallerType.UNKNOWN);
-        if (CollectionUtils.isNotEmpty(result)) {
-            // always going to be one entitlement for a single pool
-            return result.get(0);
-        }
-        return null;
     }
 
     @Override
@@ -1457,8 +1438,8 @@ public class CandlepinPoolManager implements PoolManager {
         Map<String, Entitlement> entitlements = new HashMap<String, Entitlement>();
         entitlements.put(entitlement.getPool().getId(), entitlement);
 
-        List<Entitlement> result = addOrUpdateEntitlements(
-            consumer, poolQuantities, entitlements, true, CallerType.UNKNOWN);
+        List<Entitlement> result = addOrUpdateEntitlements(consumer, poolQuantities, entitlements,
+            CallerType.UNKNOWN);
 
         if (CollectionUtils.isNotEmpty(result)) {
             // always going to be one entitlement for a single pool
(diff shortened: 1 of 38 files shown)


mstead:
There's some duplication in the certificate class, but I plan on cleaning that up along with the certificate generation class realignment story as the same thing appears to be in a few cert classes right now. I decided to leave it in there as there's already enough going on in this PR.

Ceiu:
Testing instructions and Candlepin operations check out. Will sift through the code shortly.

mstead:
NOTE: This was also tested to ensure satellite's capsule syncs function correctly -- all seems fine.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1586,10 +1567,10 @@ else if (!poolQuantities.keySet().equals(entitlements.keySet())) {
         }
 
         handler.handlePostEntitlement(this, consumer, entitlements);
-        handler.handleSelfCertificates(consumer, poolQuantities, entitlements, generateUeberCert);
+        handler.handleSelfCertificates(consumer, poolQuantities, entitlements);
```
Ceiu: With this no longer having the option to generate ueber certs, is there ever a point where we're at risk of not updating a ueber cert when we need to here? Or are they so far out of band now that it's not a concern?

mstead: They are too far out of band to care about.

Any old ueber certs should get cleaned up as part of this PR and any old certs should be ignored IF they happen to escape the cleanup included with this PR. Any new requests to get/generate uber certs will check the new table and generate in there.

Essentially, once this version of CP is deployed, ueber cert gen will start from scratch and be generated the new way. Today, any time satellite wants the ueber cert, it will attempt a GET and default to generate if it doesn't exist, so no worries from that side as well.

server/src/main/java/org/candlepin/model/UeberCertificateCurator.java
```diff
@@ -0,0 +1,68 @@
...
+     * Delete the UeberCertificate related to the specified owner.
+     *
+     * @param owner the target owner
+     * @return true if a certificate was delted for the Owner, false otherwise.
+     */
+    public boolean deleteForOwner(Owner owner) {
+        String deleteForOwner = "delete from UeberCertificate uc where uc.owner = :owner";
+        Query query = getEntityManager().createQuery(deleteForOwner);
+        query.setParameter("owner", owner);
+        return query.executeUpdate() > 0 ? true : false;
```
Ceiu: You can just return ```query.executeUpdate() > 0``` here.

server/src/main/resources/db/changelog/20170214115606-uber-cert-rewrite.xml
```diff
@@ -0,0 +1,203 @@
...
+        <comment>Cleanup all ueber data so that the new style ueber certs can be used.</comment>
+        <sql>${delete_ueber_certs_rewrite}</sql>
+        <sql>${delete_pool_source_sub_rewrite}</sql>
+        <sql>${delete_ueber_ents_rewrite}</sql>
+        <sql>${delete_ueber_pools_rewrite}</sql>
+        <sql>${delete_owner_content_rewrite}</sql>
+        <sql>${delete_ueber_content_rewrite}</sql>
+        <sql>${delete_owner_products_rewrite}</sql>
+        <sql>${delete_ueber_products_rewrite}</sql>
+        <sql>${delete_ueber_consumers_rewrite}</sql>
```
Ceiu: This is an interesting way of doing this. Is there a reason you opted to do these queries as properties rather than changesets (or completely different changelog files for each db)?

mstead: No reason other than allowing me to use a single changeset for multiple DBs -- seemed less verbose. Honestly -- I hate the property approach for this.

Let me look at trying something else.

server/src/main/java/org/candlepin/model/UeberCertificateCurator.java
mstead: @Ceiu  LOL - Wow -- Can't believe I did that.

Ceiu:
ACK

PR #1479, opened 2017-02-16
[WIP] docker-refactor

Lorquas:
This is a complete refactor of how docker works in this repo.  This will break all existing jenkins automation and should be merged in at the same time as changes to [candlepin-jobs](https://github.com/candlepin/candlepin-jobs) .  For  details on this change, see docker/README.mkd

I'm opening this PR to open things for review.  All existing testing is expected to fail.

I've limited the scope of this to replace existing jenkins docker tests.  There are a few areas we can expand this scope to (like updating rhel base images, other prebuilt databases, etc), and I'm open to discussion on what we should add in this PR.


Lorquas:
Note: the registry in docker/.env will have to be updated before merge to master and after manual testing

docker/candlepin-base/setup-supervisord.sh
```diff
@@ -2,6 +2,8 @@
```
Ceiu: Do we still need supervisord?

docker/oracle/old/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM centos:7
```
Ceiu: Do we still need this file?

docker/oracle/old/setup-supervisord.sh
```diff
@@ -0,0 +1,33 @@
+#! /bin/bash
```
Ceiu: Do we still need this file?

vritant:
remove default mark

docker/candlepin-base/setup-supervisord.sh
Lorquas: Yes; the deploy script and the test setup in cp-test relies on several services to be started/running:  tomcat, qpid, sshd (nice to have, not strictly needed for unit test running).  The problem is that in a container systemd itself hasn't started.  This leads us down two paths:

1)  We go ahead and start it the same way you would on a normal machine startup using `/usr/sbin/init` ;  this however requires a 'super' privileged mode like this: `docker run -it -e "container=docker" --privileged=true -d  --security-opt seccomp:unconfined --cap-add=SYS_ADMIN -v /sys/fs/cgroup:/sys/fs/cgroup:ro --name $NAME $IMAGE bash -c "/usr/sbin/init"` .  This only complicates our nice new simplified docker-compose setup, and introduces all the fun problems you get by running docker in  privileged mode.

2)  We could manually call the daemons and fork in our deploy script, which would require a little more logic, but supervisord already does this for us.  The question here being should we re-implement things or find a better tool.

I found myself going down a bit of a rabbit hole with the above so I decided limit the scope of this PR to its original goals and punt on that work for later.

Lorquas:
@Ceiu   woops, forgot to remove the old oracle folder, deleted that
@vritant not sure what you mean?

docker/candlepin-base/setup-supervisord.sh
Ceiu: Nah, that's fine. I thought we were only using supervisord for spinning up the databases; but if we're using it for other things, then we can just leave it. No reason to reinvent the wheel quite yet.

Lorquas:
waiting for https://github.com/candlepin/candlepin-jobs/pull/12  to be merged before working on merging this  (after we decide to do that of course)

cnsnyder:
The commit I've just added has a few updates to this PR.
After debugging with JC we've updated the cp-test script to ensure setup_database is called.
In addition the database used when -m is passed to the test script is now a version of mariadb closer to that of stage and prod. Finally, the configuration of mariadb has been updated to more closely match that of stage/prod (specifically the transaction isolation level is now READ-COMMITTED).

The change of transaction isolation level may or may not break spec tests in a number of our branches.
The investigation and work necessary to fix them will be done after this PR is merged (and we are using the new images to test the various branches etc).

Ceiu:
Needs a rebase

barnabycourt:
retest this please

kahowell:
35c26aa..bf88557: rebased against master

kahowell:
bf88557..9abecd4: rebase against master

kahowell:
All looks good to me, but want to have @Lorquas validate my fixup: 6977932.

Lorquas:
@kahowell LGTM

cnsnyder:
Will merge once the rebase is complete (and once the candlepin-jobs PR is ready as well)

kahowell:
6977932..fcae768: rebase against master

PR #1480, opened 2017-02-17
Fixed an issue with ContentCurator.getContentByProducts not fetching data

Ceiu:
- Worked around an issue in ContentCurator where the getContentByProducts
  method would always return an empty collection

### DIFF
--- a/server/src/main/java/org/candlepin/model/ContentCurator.java
+++ b/server/src/main/java/org/candlepin/model/ContentCurator.java
@@ -14,18 +14,20 @@
  */
 package org.candlepin.model;
 
+import com.google.common.collect.Iterables;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 
 import org.hibernate.Session;
+import org.hibernate.SQLQuery;
 import org.hibernate.criterion.DetachedCriteria;
-import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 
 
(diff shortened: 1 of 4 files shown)


Ceiu:
**NOTE**: This is a temporary-ish fix for the (silent) issue Hibernate has with querying against embedded entities. Once a better fix is introduced, this commit should be reverted.

**Testing Instructions**:
- Deploy current master with test data
- Import a manifest that contains updates for an existing content set which is used by an existing product that is not being updated in the manifest *and* references a second content set which is *not* referenced in the manifest (I can provide such a manifest on request). This should fail with an exception mentioning a content entity not existing.
- Deploy this branch
- Attempt to import the above manifest again. It should work this time.

server/src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -82,16 +86,73 @@ public Content lookupByUuid(String uuid) {
...
+                    if (i != 0) {
+                        builder.append(" OR ");
+                    }
+
+                    builder.append("product_uuid IN (?").append(++i).append(')');
+                }
+
+                builder.append(')');
+
+                SQLQuery query = session.createSQLQuery(builder.toString());
```
mstead: Due to complex nature of building up this query, lets add a quick debug statement showing the resulting query.

Ceiu: Done.

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -185,7 +185,7 @@
     @ElementCollection
     @BatchSize(size = 32)
     @CollectionTable(name = "cp2_product_content", joinColumns = @JoinColumn(name = "product_uuid"))
-    @Column(name = "element")
+    @Column(name = "content_uuid")
     @LazyCollection(LazyCollectionOption.EXTRA) // allows .size() without loading all data
```
mstead: Out of curiosity, what was the purpose of changing this? It seems out of band from this PR. Just looking for some context.

Ceiu: I don't think it has any actual value or affect on anything in the code (I could be wrong though, I'm far from an expert on the annotation side of things). However, the "element" column is just wrong -- it doesn't exist in the cp2_product_content table and is likely just leftover from copy/pasting the annotations from productAttributes above.

mstead: Cool, thanks. 'element' does seem strange. I'd say the @CollectionTable's join column trumps it or something.

mstead:
ACK - Looks good and seems to work. Waiting for tests to complete before merging.

PR #1481, opened 2017-02-17
POC for making all certificate generation lazy

barnabycourt:
Make all entitlement certificate generation lazy and don't return the certificates as part of the entitlement object.


vritant:
remove default mark

PR #1482, opened 2017-02-20
Changed ProductContent to a full entity

Ceiu:
- ProductContent is no longer an @Embeddable entity and is a full entity
- Updated product accordingly

### DIFF
--- a/server/src/main/resources/db/changelog/20170220102027-add-id-column-for-productcontent.xml
+++ b/server/src/main/resources/db/changelog/20170220102027-add-id-column-for-productcontent.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+    <changeSet id="20170220102027-1" author="crog">
+        <!-- On MySQL/MariaDB, we need to drop *all* foreign keys before we can drop our primary key -->
+        <dropAllForeignKeyConstraints baseTableName="cp2_product_content"/>
+
+        <dropPrimaryKey tableName="cp2_product_content" constraintName="cp2_product_content_pk"/>
+
+        <!-- Liquibase won't let us specify the positional offset, despite the docs saying we can. -->
+        <addColumn tableName="cp2_product_content">
+            <column name="id" type="varchar(32)"/>
+        </addColumn>
+
+        <addPrimaryKey tableName="cp2_product_content"
+            columnNames="id"
+            constraintName="cp2_product_content_pk"
+        />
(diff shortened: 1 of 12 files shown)


Ceiu:
This reverts the changes made in #1480. The same test steps should be performed to ensure the bug is not reintroduced by this commit.

**Testing Instructions**:
- Deploy current master with test data
- Import a manifest that contains updates for an existing content set which is used by an existing product that is not being updated in the manifest *and* references a second content set which is *not* referenced in the manifest (I can provide such a manifest on request). This should fail with an exception mentioning a content entity not existing.
- Deploy this branch
- Attempt to import the above manifest again. It should work this time.

Ceiu:
retest this please

vritant:
retest this please

vritant:
retest this please

vritant:
retest this please

vritant:
retest this please

vritant:
retest this please

vritant:
retest this please

vritant:
@Ceiu could this be a legit failure? the changelog fails to apply against IT's db

Ceiu:
@vritant I'm not sure how data could be being truncated for a column that doesn't yet exist.

Ceiu:
retest this please

Ceiu:
Rebased against master

Ceiu:
retest this please

Ceiu:
retest this please

Ceiu:
Rebased against master

Ceiu:
Current test suite failures are unrelated to the code.

mstead:
retest this please

Ceiu:
Rebased against master

server/src/main/resources/db/changelog/20170220102027-add-id-column-for-productcontent.xml
```diff
@@ -0,0 +1,50 @@
...
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+    <changeSet id="20170220102027-1" dbms="mysql" author="crog">
+        <!-- On MySQL/MariaDB, we need to drop *all* foreign keys before we can drop our primary key -->
+        <dropAllForeignKeyConstraints baseTableName="cp2_product_content"/>
```
awood: I would like to see all of these SQL commands in one single `changeSet`.  I say that because it feels like all these statements should be within a single transaction.  The Liquibase docs say:

```
Liquibase attempts to execute each changeSet in a transaction that is committed at the end, or rolled back if there is an error. Some databases will auto-commit statements which interferes with this transaction setup and could lead to an unexpected database state. Therefore, it is usually best to have just one change per changeSet unless there is a group of non-auto-committing changes that you want applied as a transaction such as inserting data.
```

And I think in this particular case, there's a case to be made that we want these changes applied as a transaction so we don't end up with no constraints on a table or something

Ceiu: Done.

PR #1487, opened 2017-02-24
Add Correlation ID to header for external tracking

wottop:
Added to logging for specific REST session
Stored on async jobs spawned by REST call

### DIFF
--- a/common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
+++ b/common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
@@ -14,11 +14,13 @@
  */
 package org.candlepin.common.filter;
 
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
 import java.io.IOException;
+import java.util.Enumeration;
 import java.util.UUID;
 
 import javax.inject.Singleton;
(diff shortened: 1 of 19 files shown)


barnabycourt:
I'm seeing some strangeness in the background/startup processes:
2017-02-28 16:21:38,812 [thread=localhost-startStop-1] [=, org=, csid=] INFO  org.candlepin.audit.HornetqContextListener - Hornetq server started

Is there a way to drop the  [=, org=, csid=]  entirely?

wottop:
I don't know. I only added the last of those. The other name/values were already getting generated. I can look if you like.

server/src/main/java/org/candlepin/pinsetter/tasks/ImportJob.java
```diff
@@ -124,6 +126,8 @@ public static JobDetail scheduleImport(Owner owner, String storedFileId, String
         map.put(STORED_FILE_ID, storedFileId);
         map.put(UPLOADED_FILE_NAME, uploadedFileName);
         map.put(CONFLICT_OVERRIDES, overrides.asStringArray());
+        map.put(JobStatus.CORRELATION_ID, MDC.get("csid"));
+        log.info("CSID: " + map.get(JobStatus.CORRELATION_ID));
```
barnabycourt: Is this log statement necessary?

common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
```diff
@@ -72,6 +75,25 @@ public void doFilter(ServletRequest request, ServletResponse response,
...
+            while (headerNames.hasMoreElements()) {
+                String name = (String) headerNames.nextElement();
+                if ("X-Correlation-ID".equalsIgnoreCase(name)) {
+                    correlationId = ((HttpServletRequest) request).getHeader(name);
+                }
+            }
+
+            if (StringUtils.isAlphanumeric(correlationId) &&
+                correlationId.length() <= CORRELATION_ID_LENGTH) {
+                MDC.put("csid", correlationId);
```
barnabycourt: Please use a constant for "csid"

PR #1498, opened 2017-03-03
1397434: silently drop duplicate facts

vritant:


barnabycourt:
@vritant Is a migration needed to remove existing duplicate facts?

vritant:
@barnabycourt I think there should not be any duplicate facts in existing mysql databases, as the insert should have failed.

PR #1502, opened 2017-03-06
Remove pool source consumer from API

wottop:


vritant:
* What was the incentive behind this?
* are we assuming the legacy clients will change appropriately?

barnabycourt:
@vritant legacy clients are not using this part of the API. The incentive is to remove nested objects that are not needed from the API

PR #1504, opened 2017-03-06
1428146: Do not deadlock on multiple synchronous hypervisor checkins

cnsnyder:


cnsnyder:
This is going to fail the mysql tests until new docker images for the mysql containers have been pushed out. Still working on that piece. Ignore those failures for the moment.

cnsnyder:
This PR is blocked by the merge of https://github.com/candlepin/candlepin/pull/1479.
Once that is merged we can expect the tests to pass once more.

Ceiu:

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -631,20 +631,20 @@ public VirtConsumerMap getHostConsumersMap(Owner owner,
...
-        List<Criterion> ors = new LinkedList<Criterion>();
+        LinkedList<String> hypervisorIdsUpper = new LinkedList<String>();
         for (String hypervisorId : hypervisorIds) {
-            ors.add(Restrictions.eq("hvsr.hypervisorId", hypervisorId).ignoreCase());
+            // Searching for these in a case-insensitive manner
+            hypervisorIdsUpper.add(hypervisorId.toUpperCase());
         }
-        return Restrictions.or(ors.toArray(new Criterion[0]));
+        String query = "SELECT c FROM Consumer c WHERE" +
+                       "(c.owner.key = :ownerKey AND UPPER(c.hypervisorId.hypervisorId) IN (:hypervisorIds))" +
```
Ceiu: Potential IN operator limit issue.

Ceiu:

server/src/main/java/org/candlepin/model/ConsumerCurator.java
cnsnyder: Good catch will take another look.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -626,25 +626,37 @@ public VirtConsumerMap getHostConsumersMap(Owner owner,
...
+            List<String> subList = hypervisorIdsUpper.subList(fromIndex, toIndex);
+            q.setLockMode(LockModeType.PESSIMISTIC_WRITE);
+            q.setParameter("ownerKey", ownerKey);
+            q.setParameter("hypervisorIds", subList);
+            consumers.addAll(q.getResultList());
+            fromIndex = toIndex;
+            toIndex += MAX_IN_QUERY_LENGTH;
         }
-        return Restrictions.or(ors.toArray(new Criterion[0]));
+        return consumers;
```
Ceiu: This will work here, but in the mainline branch going forward, we should avoid re-running the query. There are examples in various curators (AbstractHibernateCurator has an especially generic example) which make use of collection partitioning and query building which builds a query that only runs once.

Ceiu:
ACK

PR #1507, opened 2017-03-07
Remove gutterball

kahowell:
Lots more could probably be done to simplify the code (ex. @cnsnyder
suggests integrating common into server), but this change simply removes
it from the build and repo, and removes references to gutterball.

Note I left a few things alone:

 - Did not remove gutterball strings from translations
 - Did not remove gutterball prefixed properties from cp amqp config


kahowell:
b358ec7..db50183: rebase against master

Ceiu:
With Gutterball gone, we really should merge the bits in common into server, unless there is some need to keep them separated out. As it is now, we end up building the common bits as a separate package, which is seemingly unnecessary now.

I feel like that task should block this, as there's a good chance it won't get done in the foreseeable future otherwise; but I can be convinced otherwise.

vritant:
@kahowell I would recommend we open a doc where the team can list the things that can be done once we do this. most of them could be tackled in latter PRs, but perhaps some are best addressed in this one.

barnabycourt:
@Ceiu I'm validating with it-pnt, they may still be using the common jar file.

kahowell:
@barnabycourt @Ceiu @vritant while I agree we should refactor `common` back into `server` ASAP (and maybe other stuff), I'd rather do it in a follow-up PR; also with that work, we *could* flatten out the project, and have the contents of `server` at the root of the repo (along with everything else), unless someone has a good reason not to do that? Thoughts?

Ceiu:
@barnabycourt I'm sure they do, but I would be surprised if they use it directly. I imagine their usage of it was because it's a dependency for CP proper.

barnabycourt:
@kahowell There still is still a separate project embedded for the checkstyle module to fin resteasy endpoints that have not been swagger annotated. That would be at least one blocker for fully merging everything to the root as long as that is in the project.

PR #1514, opened 2017-03-13
1431639: Allow hyphen for correlation id

wottop:


wottop:
buildbot: retest this please

common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
```diff
@@ -85,14 +85,13 @@ public void doFilter(ServletRequest request, ServletResponse response,
                     correlationId = ((HttpServletRequest) request).getHeader(name);
                 }
             }
-
-            if (StringUtils.isAlphanumeric(correlationId) &&
+            if (correlationId.matches("^([a-zA-Z0-9-])*$") &&
```
wottop: Allowing empty string makes no sense, I will change to '+ '

PR #1515, opened 2017-03-14
Remove null version inclusion on getProductsByVersions and getContentByVersions

barnabycourt:
This change removes the check to always include the version of a
given content or product that was created during the initial migration
to versioned products & content. This removal means that the
initial versions created during the migration will never be matched
for deduplication. The first refresh for a given product or content will
always create a new entity. The existing orphan cleanup job will
clean up the initially migrated data.

The plus side is that we are not performing the null check for every item in the SQL query
and that we do not always return 2 entities that must be field wise
compared for the existing entities once a new version has been imported.


Ceiu:
For clarity, those migrated products with null versions will only be fetched if their product ID matches as well, and they get orphaned and cleaned up all the same. They still behave the same as normal entities.

wottop:
ack

PR #1516, opened 2017-03-15
Awood/swagger corrections

awood:
Corrections and changes to the Swagger annotations to produce a higher quality generated client.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/ConsumerTypeResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerTypeResource.java
@@ -39,6 +39,7 @@ import javax.ws.rs.core.MediaType;
 
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 import io.swagger.annotations.Authorization;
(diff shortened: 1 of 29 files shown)


awood:
Buildbot retest this please

server/src/main/java/org/candlepin/resource/ConsumerTypeResource.java
```diff
@@ -61,7 +61,8 @@ public ConsumerTypeResource(ConsumerTypeCurator consumerTypeCurator, I18n i18n)
         this.i18n = i18n;
     }
 
-    @ApiOperation(notes = "Retrieves a list of Consumer Types", value = "list")
+    @ApiOperation(notes = "Retrieves a list of Consumer Types", value = "list",
+        response = ConsumerType.class, responseContainer = "list")
```
Ceiu: So it looks like the missing piece was the "response = <target class>.class, responseContainer = "list". If that's what we needed, what is the @Wrapped annotation doing here? Looks like the intent was the same.

server/src/main/java/org/candlepin/resource/PoolResource.java
```diff
@@ -194,7 +195,7 @@ public PoolResource(ConsumerCurator consumerCurator, OwnerCurator ownerCurator,
     @Produces(MediaType.APPLICATION_JSON)
     public Pool getPool(@PathParam("pool_id") @Verify(Pool.class) String id,
         @QueryParam("consumer") String consumerUuid,
-        @QueryParam("activeon") String activeOn,
+        @ApiParam("Uses ISO 8601 format") @QueryParam("activeon") String activeOn,
```
Ceiu: Do we want to specify that it's a date/time format here?

Ceiu:
ACK

server/src/main/java/org/candlepin/resource/ConsumerTypeResource.java
awood: Looks like [`Wrapped`]( https://docs.jboss.org/resteasy/docs/1.0.0.GA/javadocs/org/jboss/resteasy/annotations/providers/jaxb/Wrapped.html) is the same idea but RESTEasy specific.  Swagger didn't pay attention to it because I had to put in the response class and response container to get the client to generate correctly.

PR #1518, opened 2017-03-16
1432653: Liquibase scripts need hashes

wottop:


barnabycourt:
This is missing hashes for the other previous versions. 
In addition the changeset for 20170125112728-cleanup-uebercerts-round3-2-0.xml needs the same type of update. 

A quick way to get the checksums is to switch to the branch in the vagrant image and then use a liquibase command like:
`liquibase --driver=org.postgresql.Driver --classpath=/usr/share/java/postgresql-jdbc.jar:/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/ --changeLogFile=db/changelog/changelog-update.xml --url=jdbc:postgresql:candlepin --username=<DB_USER>  --password=<DB_PASSWORD> calculateCheckSum 'db/changelog/20170125112728-cleanup-uebercerts-round3-2-0.xml::20170125112728-1::mstead'`

to calculate the desired checksum.

server/src/main/resources/db/changelog/20161025103454-cleanup-ueber-certs-round2.xml
```diff
@@ -248,7 +248,7 @@
         value="delete c from cp_consumer c where name like 'ueber_cert_consumer';"/>
 
     <changeSet id="20161025103454-1" author="mstead">
-        <validCheckSum>ALL</validCheckSum>
```
barnabycourt: This is missing the full list of previous checksums.

PR #1519, opened 2017-03-16
Optimized how consumers are enriched with compliance details

Ceiu:
- Added a new class, ConsumerEnricher, to replace the now-deleted
  ConsumerInstalledProductEnricher which performs the enrichment
  process using far fewer database and rules calls
- Added a new option for the get_status call in the rules:
  calculateProductComplianceDateRanges. This option will calculate
  the date ranges for each compliant-ish product to remain in its
  current state
- Cleaned up a few bits of old or redundant code


Ceiu:
**Testing:**
- Load a database having consumers with many installed products and entitlements (the prod data dump is a good test for this)
- Deploy current master, without recreating the database
- Pull a consumer, timing the result and writing the output to a file:
```
crog@crog-ws-1:~/devel/candlepin-origin/server (master) $ time curl -k -u admin:admin "https://localhost:8443/candlepin/consumers/d433bad1-7837-4d85-90ad-aa4a734ade21" > master.txt
```
- Checkout this branch (```crog/installed_product_enricher_optimizations```)
- Deploy Candlepin, without recreating the database
- Pull the same consumer, timing the result and writing the output to a file:
```
crog@crog-ws-1:~/devel/candlepin-origin/server (crog/installed_product_enricher_optimizations) $ time curl -k -u admin:admin "https://localhost:8443/candlepin/consumers/d433bad1-7837-4d85-90ad-aa4a734ade21" > update.txt
```
- Verify the time to pull the consumer was no worse than current master [1]
- Verify that the two files have the same content: ```diff master.txt update.txt``` [2]

[1] In my tests, this branch is 15 to 20 times faster than current master for pulling consumers
[2] There is an edge case which is handled differently compared to master which is acceptable for verification: if an installed product is not present in Candlepin, master will return its status as "null," with no date range, whereas this branch will return its status as "red," and also without a date range.

barnabycourt:
@Ceiu This calculates the enriched value for each installed product each time. What would the impact be of saving this back into the ConsumerInstalledProduct table so we avoid the enrichment & run through the rules entirely when the status is already up to date?

Ceiu:
A database change, something to determine whether or not the cached status
is dirty, flipping that bit where necessary and then slight model changes
to write all that info back. The hardest part there would be the
investigation on where we would need to update the dirty flag, and then all
the testing that would need to be added as well.

My hope here is that this patch gets us fast enough so that we don't have
to think about it again until we get around to refactoring/rewriting how
rules calculations work entirely.

barnabycourt:
@Ceiu Regarding database changes, those are very good points. Waiting until we can refactor rules more generally is prudent.

barnabycourt:
LGTM

server/src/main/java/org/candlepin/resource/util/ConsumerInstalledProductEnricher.java
```diff
@@ -1,248 +0,0 @@
...
-        return status;
-    }
-
-    /**
-     * Get the {@link DateRange} that the specified product will be valid (inclusive).
-     *
-     * @param product the product
-     * @return the valid date range of the product.
-     */
-    protected DateRange getValidDateRange(Product product) {
```
awood: Most of the intricate stuff in ConsumerInstalledProductEnricher got pushed out to the JS rules it looks like?

awood:
:+1: from me.  @barnabycourt Merge it when you are ready.

Ceiu:
Yeah. Since the process so heavily relies on the work done in the rules, I
just moved everything out to them to limit how many times we're serializing
stuff.

server/src/main/java/org/candlepin/resource/util/ConsumerInstalledProductEnricher.java
Ceiu: Yeah. Since the process so heavily relies on the work done in the rules, I just moved everything out to them to limit how many times we're serializing stuff.

barnabycourt:
I manually ran the spec tests as the build infrastructure was broken due to an upstream qpid break in the qpid_proton library.

PR #1520, opened 2017-03-16
Add missing post bind bonus pool spec tests

vritant:


Ceiu:
Are we expecting these tests to fail?

vritant:
@Ceiu yeah atleast some tests were written to fail, so they can be a point of a discussion. ive commented them out now, will sort that issue out before I remove the inprogress label.

Ceiu:
@vritant Looks like this is good to go and is even passing. Is the "ignore jenkins failurs" tag just out of date for the latest commit?

PR #1522, opened 2017-03-21
Removed @Formula for  CertificateSerial.revoked

mstead:
No longer use @Formula to determine if a cert
serial is revoked or not. Instead, we store the
state on the object when the referencing certificate
is revoked.

The serial's revoked state is set to true on deletion
of the associated certificate via means of a @PreRemove
callback. Using the callback to update revoked comes
with a drawback. We can not delete the record via JPQL
or the callback will not be fired. It must be deleted
via cascade, or directly via the EntityManager.

Any certificate that wishes to make use of this functionality
must extend the RevokableCertificate base class, and should
follow the deletion process above.

A liquibase changeset is also included with this patch
and when applied, will add a new revoked column to
cp_certificate_serial and will update any serial with
no association to a certificate to be revoked.

## Testing Notes

It is a good idea to start testing this PR with a populated database so that you can verify that any existing cert serials get properly marked as revoked. Running the spec tests on the master branch will end up with some serials in the DB.

```sql
-- Check how many cert serials are not currently being referenced by a certificate
select count(*) from cp_cert_serial cs
                 left outer join cp_ent_certificate ec on ec.serial_id = cs.id
                 left outer join cp_cdn_certificate cdnc on cs.id = cdnc.serial_id
                 left outer join cp_certificate subcert on cs.id = subcert.serial_id
                 left outer join cp_id_cert idcert on cs.id = idcert.serial_id
                 left outer join cp_cont_access_cert cacert on cs.id = cacert.serial_id
                 left outer join cp_ueber_cert ueber on cs.id = ueber.serial_id
                 where ec.id is null and cdnc.id is null and subcert.id is null and idcert.id is null and cacert.id is null and ueber.id is null;

-- check how many cert serials are being referenced from a certificate
select count(*) from cp_cert_serial cs
                 left outer join cp_ent_certificate ec on ec.serial_id = cs.id
                 left outer join cp_cdn_certificate cdnc on cs.id = cdnc.serial_id
                 left outer join cp_certificate subcert on cs.id = subcert.serial_id
                 left outer join cp_id_cert idcert on cs.id = idcert.serial_id
                 left outer join cp_cont_access_cert cacert on cs.id = cacert.serial_id
                 left outer join cp_ueber_cert ueber on cs.id = ueber.serial_id
                 where ec.id is not null OR cdnc.id is not null OR subcert.id is not null OR idcert.id is not null OR cacert.id is not null OR ueber.id is not null;

-- After the changeset is applied, verify that the revoked count is the same as that
-- of the non-referenced query.
select count(*) from cp_cert_serial where revoked = true;

```

### DIFF
--- a/server/src/main/java/org/candlepin/model/CertificateSerial.java
+++ b/server/src/main/java/org/candlepin/model/CertificateSerial.java
@@ -16,12 +16,12 @@ package org.candlepin.model;
 
 import org.candlepin.util.Util;
 
-import org.hibernate.annotations.Formula;
 import org.hibernate.annotations.GenericGenerator;
 
 import java.math.BigInteger;
 import java.util.Date;
 
+import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
@@ -49,20 +49,15 @@ public class CertificateSerial extends AbstractHibernateObject {
     /*
      * A CertificateSerial is considered revoked when no certificates reference it
      * TODO: put different kinds of serials into different tables.  More specifically,
-     * those that we own (can revoke) vs those that we don't
+     * those that we own (can revoke) vs those that we don't.
      */
-    @Formula("(CASE (" +
-        "(SELECT count(entcert.id) FROM cp_ent_certificate entcert where entcert.serial_id = id) + " +
-        "(SELECT count(cdncert.id) FROM cp_cdn_certificate cdncert where cdncert.serial_id = id) + " +
-        "(SELECT count(subcert.id) FROM cp_certificate subcert where subcert.serial_id = id) + " +
-        "(SELECT count(idcert.id) FROM cp_id_cert idcert where idcert.serial_id = id) + " +
-        "(SELECT count(contacccert.id) FROM cp_cont_access_cert contacccert where " +
-        "contacccert.serial_id = id)" +
-        ") WHEN 0 THEN 1 ELSE 0 END)")
+    @NotNull
+    @Column(nullable = false)
     private boolean revoked;
 
     // Set to true if this serial is already a part of the CRL
     @NotNull
+    @Column(nullable = false)
     private boolean collected;
 
     // The expiration.
@@ -101,6 +96,13 @@ public class CertificateSerial extends AbstractHibernateObject {
      */
     public boolean isRevoked() {
         return revoked;
+    }
+
+    /**
+     * @param isRevoked whether or not this serial is revoked.
+     */
+    public void setRevoked(boolean isRevoked) {
+        this.revoked = isRevoked;
     }
 
     /**
--- a/server/src/main/java/org/candlepin/model/CertificateSerialCurator.java
+++ b/server/src/main/java/org/candlepin/model/CertificateSerialCurator.java
@@ -20,13 +20,10 @@ import com.google.common.collect.Iterables;
 import com.google.inject.persist.Transactional;
 
 import org.hibernate.Query;
-import org.hibernate.criterion.Conjunction;
-import org.hibernate.criterion.Criterion;
 import org.hibernate.criterion.DetachedCriteria;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
-import org.hibernate.criterion.Subqueries;
 
 import java.util.Collection;
 import java.util.Date;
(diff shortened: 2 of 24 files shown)


server/src/main/java/org/candlepin/model/CertificateSerial.java
```diff
@@ -49,20 +49,15 @@
     /*
      * A CertificateSerial is considered revoked when no certificates reference it
      * TODO: put different kinds of serials into different tables.  More specifically,
-     * those that we own (can revoke) vs those that we don't
+     * those that we own (can revoke) vs those that we don't.
```
barnabycourt: The above comment about splitting tables can likely be removed given the changes made here.

server/src/main/java/org/candlepin/model/CertificateSerialCurator.java
```diff
@@ -88,7 +80,7 @@ public int deleteExpiredSerials() {
         List<String> ids = this.currentSession()
             .createCriteria(CertificateSerial.class)
             .add(Restrictions.le("expiration", Util.yesterday()))
-            .add(getRevokedCriteria())
+            .add(Restrictions.eq("revoked", true))
```
barnabycourt: This will delete those that have expired AND been marked as revoked. Is that the intention?

server/src/main/java/org/candlepin/model/CertificateSerialCurator.java
```diff
@@ -70,7 +62,7 @@ public CertificateSerialCurator() {
 
         DetachedCriteria criteria = DetachedCriteria.forClass(CertificateSerial.class)
             .add(Restrictions.le("expiration", Util.yesterday()))
-            .add(getRevokedCriteria());
+            .add(Restrictions.eq("revoked", true));
```
barnabycourt: This will delete those that have expired AND been marked as revoked. Is that the intention?

server/src/main/java/org/candlepin/model/RevokableCertificate.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.persistence.Transient;
+
+
+/**
+ * A class that represents a revokable certificate. A revokable certificate should
+ * have an associated {@link CertificateSerial} that can be placed on the Certificate
+ * Revocation List.
+ */
+@MappedSuperclass
+public abstract class RevokableCertificate extends AbstractCertificate {
```
barnabycourt: I believe the proper spelling is Revocable

server/src/main/resources/db/changelog/20170301083925-add-cert-serial-revoked-column.xml
```diff
@@ -0,0 +1,74 @@
...
+            the hibernate formula off of the CertificateSerial entity.
+        </comment>
+        <addColumn tableName="cp_cert_serial">
+            <column name="revoked" type="BOOLEAN" defaultValue="false">
+                <constraints nullable="false" />
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="20170301083925-2" dbms="postgresql,oracle" author="mstead">
```
barnabycourt: This should use a single changeset and changeset ID for both mysql & others but should use a property like 20170206114359-convert-quanity-consumed-exported-to-columns.xml. That will ensure that no problem occurs if someone is migrating from mariadb to postgres due to differences in changelog history.

mstead: I moved away from the 'property' approach because as it turns out, those are global, and if another changelog happens to define the same property, the value defined in the first processed changelog will be used. I've been bitten by this in the past.

Newer versions of liquibase allows specifying dbms on the sql tag. Would be nice if we could upgrade it at some point -- depends on what RHEL has though.

I'll update it to use properies.

barnabycourt:
@mstead One other request for the liquibase changeset. Please add a rollback section so that the changes can be easily backed out.

server/src/main/resources/db/changelog/20170301083925-add-cert-serial-revoked-column.xml
barnabycourt: @mstead That is very good information to have. We can upgrade to version 3 (which has the proper dbms support on the sql tag as it has already been tagged fro the appropriate repositories.

mstead:
retest this please

mstead:
retest this please

PR #1529, opened 2017-03-26
Ensure product uuid does not change upon refresh with no actual changes

vritant:
@awood found that upon refresh in hosted mode, the product UUID changes. To fix that issue, this PR:
1. instead of just validating, resolves all the products of a subscription in the default adapters.
2. when comparing products, does not look if both are locked.

to verify the PR:
1. deploy master in hosted mode.
2. checkout this branch
3.run :  buildr rspec:product_versioning_spec:"should not change uuid of a product after refresh"

the same test should pass when the current branch is deployed.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/util/ResolverUtil.java
+++ b/server/src/main/java/org/candlepin/resource/util/ResolverUtil.java
@@ -31,6 +31,8 @@ import com.google.inject.Inject;
 import org.xnap.commons.i18n.I18n;
 
 import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
 
 /**
  * Convinience util for resolving owners, products, pools & subscriptions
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/resource/util/ResolverUtil.java
```diff
@@ -78,35 +80,57 @@ public Owner resolveOwner(Owner owner) {
...
     }
 
-    public Product resolveProduct(Owner owner, Product product) {
-        String id = null;
-        if (product != null) {
-            id = product.getId();
-        }
-        return resolveProduct(owner, id);
+    public Product resolveProduct(Owner owner, String productId, boolean allowNull) {
+        Product product = new Product();
```
Ceiu: Rather than create new Product instances, I would much rather change the resolveProduct such that we break it up into two methods (resolveProductById, resolveProductByUuid) or takes the ID and UUID as string params, then flip the Product- and ProductData-based methods use those.

server/src/main/java/org/candlepin/resource/util/ResolverUtil.java
```diff
@@ -78,35 +80,57 @@ public Owner resolveOwner(Owner owner) {
...
+        Product argProduct = new Product();
+        ProductData result = null;
+        if (product != null) {
+            argProduct.setId(product.getId());
+            argProduct.setUuid(product.getUuid());
         }
-
-        // TODO: Maybe add UUID resolution as well?
-        return findProduct(owner, productId);
+        if (product != null) {
```
Ceiu: Duplicate null check.

Also, include whitespace between your blocks like this. With our style forcing newlines between blocks in an if-elseif-else, the lack of whitespace here makes them look connected at a glance.

```allowNull``` isn't working properly here. If you set ```allowNull``` to false and then provide a null DTO, the call to resolveProduct that throws the exception is never called.

File: server/src/main/java/org/candlepin/resource/util/ResolverUtil.java [hunk 3]
```diff
@@ -78,35 +80,57 @@ public Owner resolveOwner(Owner owner) {
...
+                    ));
+                }
+            }
+            else {
+                throw new BadRequestException(i18n.tr(
+                   "No product specified, or product lacks identifying information"
+                ));
+            }
+        }
+        else if (!allowNull) {
```
Ceiu: Something about this kinda irks me. Entity resolution wasn't intended to be drop-in input validation. If we're resolving a product, the assumption here is that we should always have something, or it should be completely passthrough.

server/src/main/java/org/candlepin/resource/util/ResolverUtil.java
```diff
@@ -124,25 +148,30 @@ public Pool resolvePool(Pool pool) {
...
+            pool.setDerivedProduct(this.resolveProduct(owner, pool.getDerivedProduct(), true));
         }
 
         HashSet<Product> presolved = new HashSet<Product>();
 
         pool.populateAllTransientProvidedProducts(productCurator);
         for (ProvidedProduct product : pool.getProvidedProductDtos()) {
-            // TODO: Maybe add UUID resolution as well?
-            presolved.add(resolveProduct(owner, product.getProductId()));
+            Product result = resolveProduct(owner, product.getProductId(), true);
```
Ceiu: Why do we want to allow null values in a product list for a pool?

vritant:
Closing this PR. product uuid has to change because it was created via the rest API, not via import or refresh.

PR #1531, opened 2017-03-27
1435807: Optimize cert regen from content update

kahowell:
Don't load pools into memory, instead, use a query to update the
affected entitlements DB-side.

One way to see the effects:
 - set `hibernate.show_sql` to `true` in `persistence.xml`.
 - observe queries made in `buildr rspec:content_resource_spec`

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -1176,11 +1176,16 @@ public class CandlepinPoolManager implements PoolManager {
     @Override
     @Transactional
     public void regenerateCertificatesOf(String productId, boolean lazy) {
-        List<Pool> poolsForProduct = this.listAvailableEntitlementPools(null, null, null,
-            productId, new Date(), false, false, new PoolFilterBuilder(), null, false, false)
-            .getPageData();
-        for (Pool pool : poolsForProduct) {
-            regenerateCertificatesOf(pool.getEntitlements(), lazy);
+        if (lazy) {
+            poolCurator.markCertificatesDirtyForProductId(productId);
+        }
+        else {
+            List<Pool> poolsForProduct = this.listAvailableEntitlementPools(null, null, null,
+                productId, new Date(), false, false, new PoolFilterBuilder(), null, false, false)
+                .getPageData();
+            for (Pool pool : poolsForProduct) {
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1176,11 +1176,16 @@ public void regenerateCertificatesOf(Entitlement e, boolean ueberCertificate,
...
-            regenerateCertificatesOf(pool.getEntitlements(), lazy);
+        if (lazy) {
+            poolCurator.markCertificatesDirtyForProductId(productId);
+        }
+        else {
+            List<Pool> poolsForProduct = this.listAvailableEntitlementPools(null, null, null,
+                productId, new Date(), false, false, new PoolFilterBuilder(), null, false, false)
+                .getPageData();
+            for (Pool pool : poolsForProduct) {
+                regenerateCertificatesOf(pool.getEntitlements(), lazy);
```
Ceiu: We don't need to pass lazy through here, since we already know it's false at this point.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -830,4 +830,12 @@ public void calculateExportedForOwnersPools(Owner owner) {
         q.setParameter("owner", owner);
         q.executeUpdate();
     }
+
+    public void markCertificatesDirtyForProductId(String productId) {
+        String statement = "update Entitlement e set e.dirty=true where e.pool.id in " +
+            "(select p.id from Pool p where p.productId = :productId)";
```
Ceiu: This won't catch every pool the ```listAvailableEntitlementPools``` will fetch; it needs to be also checking the provided product ID as well. See: [PoolCurator.java, Line 290](https://github.com/candlepin/candlepin/blob/c1ca0255d374feff750cc14d3542f7e377021c38/server/src/main/java/org/candlepin/model/PoolCurator.java#L290).

kahowell:
c1ca025..b86ca97: address review feedback

@Ceiu fyi

Lorquas:
retest this please

barnabycourt:
Jenkins failure is due to the job expecting the CP 2.0 layout for qpid configuration instead of the 0.9.54 layout

Ceiu:
ACK

PR #1532, opened 2017-03-27
1436258: Optimize cert regen from content update

kahowell:
Don't load pools into memory, instead, use a query to update the
affected entitlements DB-side.

Analogous to #1531 (this one targets 2.0).

One way to see the effects:
 - set `hibernate.show_sql` to `true` in `persistence.xml`
 - run `buildr rspec:owner_content_resource_spec`

### DIFF
--- a/server/src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
+++ b/server/src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
@@ -377,7 +377,6 @@ public class EntitlementCertificateGenerator {
     @Transactional
     public void regenerateCertificatesOf(Collection<Owner> owners, Collection<Product> products,
         boolean lazy) {
-
         List<Pool> pools = new LinkedList<Pool>();
 
         Set<String> productIds = new HashSet<String>();
@@ -390,9 +389,14 @@ public class EntitlementCertificateGenerator {
         // TODO: This is a very expensive operation. Update pool curator with something to let us
         // do this without hitting the DB several times over.
         for (Owner owner : owners) {
-            pools.addAll(
-                this.poolCurator.listAvailableEntitlementPools(null, owner, productIds, now)
-            );
+            if (lazy) {
+                poolCurator.markCertificatesDirtyForPoolsWithProducts(owner, productIds);
+            }
+            else {
+                pools.addAll(
+                    this.poolCurator.listAvailableEntitlementPools(null, owner, productIds, now)
+                );
+            }
         }
 
         for (Pool pool : pools) {
--- a/server/src/main/java/org/candlepin/model/PoolCurator.java
+++ b/server/src/main/java/org/candlepin/model/PoolCurator.java
@@ -20,6 +20,7 @@ import org.candlepin.model.Pool.PoolType;
 import org.candlepin.model.activationkeys.ActivationKey;
 import org.candlepin.model.activationkeys.ActivationKeyPool;
 
+import com.google.common.collect.Iterables;
 import com.google.inject.Inject;
 import com.google.inject.Injector;
 import com.google.inject.persist.Transactional;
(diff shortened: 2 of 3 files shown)


server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1364,5 +1364,12 @@ public void calculateExportedForOwnersPools(Owner owner) {
         q.executeUpdate();
     }
 
-
+    public void markCertificatesDirtyForOwnerProductId(Owner owner, String productId) {
+        String statement = "update Entitlement e set e.dirty=true where e.pool.id in " +
+            "(select p.id from Pool p where p.product.id in (:productId)) and e.owner = :owner";
```
Ceiu: Like the other branch, we're not testing against derived products. Also, if we only have one product ID, we don't need to use IN here on the inner query.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1364,5 +1364,12 @@ public void calculateExportedForOwnersPools(Owner owner) {
         q.executeUpdate();
     }
 
-
+    public void markCertificatesDirtyForOwnerProductId(Owner owner, String productId) {
```
Ceiu: Pedantic request: can we rename this to ```markCertificatesDirtyFor(Pools|Entitlements)WithProduct``` (Select whichever is preferred for the noun before "with")?

server/src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
```diff
@@ -390,9 +389,16 @@ public void regenerateCertificatesOf(Collection<Owner> owners, Collection<Produc
         // TODO: This is a very expensive operation. Update pool curator with something to let us
         // do this without hitting the DB several times over.
         for (Owner owner : owners) {
-            pools.addAll(
-                this.poolCurator.listAvailableEntitlementPools(null, owner, productIds, now)
-            );
+            if (lazy) {
+                for (Product product : products) {
+                    poolCurator.markCertificatesDirtyForOwnerProductId(owner, product.getId());
```
Ceiu: We could optimize this a bit by modifying this to build up our set of product IDs, then issue one update statement (and disregarding my request to drop the IN on the inner query below).

kahowell:
5352ed1..a0e2ab4: address review feedback

@Ceiu I think that covers it all.

kahowell:
a0e2ab4..0d799d7: fix checkstyle issues

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -69,6 +70,9 @@
     /** The recommended number of expired pools to fetch in a single call to listExpiredPools */
     public static final int EXPIRED_POOL_BLOCK_SIZE = 2048;
 
+    /* cf. oracle's maximum; postgres and mysql don't have a limit */
+    public static final int MAX_IN_CLAUSE_VALUES = 1000;
```
Ceiu: We already have this constant defined in AbstractHibernateCurator (IN_OPERATOR_BLOCK_SIZE).

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1364,5 +1368,47 @@ public void calculateExportedForOwnersPools(Owner owner) {
...
     }
 
+    public void markCertificatesDirtyForPoolsWithProducts(Owner owner, Collection<String> productIds) {
+        if (productIds.size() <= MAX_IN_CLAUSE_VALUES) {
+            markCertificatesDirtyForPoolsWithNormalProducts(owner, productIds);
+            markCertificatesDirtyForPoolsWithProvidedProducts(owner, productIds);
+        }
+        else {
+            Iterator<String> productIdIter = productIds.iterator();
+            List<String> batch = new ArrayList<String>();
```
Ceiu: We should use the Iterables.partition method here, as it creates a view over sections of the original collection instead of having to copy over blocks to a new list. We can reduce this down to (something close to) the following example:

```
for (List<String> block : Iterables.partition(productIds, IN_OPERATOR_BLOCK_SIZE)) {
    this.markCertificatesDirtyForPoolsWithNormalProducts(owner, block);
    this.markCertificatesDirtyForPoolsWithProvidedProducts(owner, block);
}
```

kahowell:
0d799d7..281e459:
 - use `Iterables.partition`
 - use `IN_OPERATOR_BLOCK_SIZE`

@Ceiu thanks for the feedback. I looked for a pre-existing constant but missed it. Also was not familiar with `Iterables.partition`. :smile:

server/src/main/java/org/candlepin/model/PoolCurator.java
Ceiu: Still need to remove this constant

Ceiu:
np
Looks good aside from the lingering constant. Once that's removed this should be ready to be merged.

kahowell:
281e459..615d117: remove lingering constant

kahowell:
615d117..941bd09: replace test reference to `MAX_IN_CLAUSE_VALUES`

Ceiu:

kahowell:
retest this please

Ceiu:

One last attempt to get the tests to run

Ceiu:
ACK

I've tested this manually with fresh Docker images. All tests pass when running locally.

PR #1533, opened 2017-03-29
Removed @Formula for  CertificateSerial.revoked (V2)

mstead:
No longer use @Formula to determine if a cert
serial is revoked or not. Instead, we store the
state on the object when the referencing certificate
is revoked.

The serial's revoked state is set to true on deletion
of the associated certificate via means of a @PreRemove
callback. Using the callback to update revoked comes
with a drawback. We can not delete the record via JPQL
or the callback will not be fired. It must be deleted
via cascade, or directly via the EntityManager.

Any certificate that wishes to make use of this functionality
must extend the RevokableCertificate base class, and should
follow the deletion process above.

A liquibase changeset is also included with this patch
and when applied, will add a new revoked column to
cp_certificate_serial and will update any serial with
no association to a certificate to be revoked.

The generator was removed from CertificateSerial
and the serial number is now generated on object
creation. This allows us to get a usabled serial
number without having to persist the
CertificateSerial first. Since we were fabricating
the serial on our own, there wasn't any need to
be doing this on persist.

## Testing Notes

It is a good idea to start testing this PR with a populated database so that you can verify that any existing cert serials get properly marked as revoked. Running the spec tests on the master branch will end up with some serials in the DB.

```sql
-- Check how many cert serials are not currently being referenced by a certificate
select count(*) from cp_cert_serial cs
                 left outer join cp_ent_certificate ec on ec.serial_id = cs.id
                 left outer join cp_cdn_certificate cdnc on cs.id = cdnc.serial_id
                 left outer join cp_certificate subcert on cs.id = subcert.serial_id
                 left outer join cp_id_cert idcert on cs.id = idcert.serial_id
                 left outer join cp_cont_access_cert cacert on cs.id = cacert.serial_id
                 left outer join cp_ueber_cert ueber on cs.id = ueber.serial_id
                 where ec.id is null and cdnc.id is null and subcert.id is null and idcert.id is null and cacert.id is null and ueber.id is null;

-- check how many cert serials are being referenced from a certificate
select count(*) from cp_cert_serial cs
                 left outer join cp_ent_certificate ec on ec.serial_id = cs.id
                 left outer join cp_cdn_certificate cdnc on cs.id = cdnc.serial_id
                 left outer join cp_certificate subcert on cs.id = subcert.serial_id
                 left outer join cp_id_cert idcert on cs.id = idcert.serial_id
                 left outer join cp_cont_access_cert cacert on cs.id = cacert.serial_id
                 left outer join cp_ueber_cert ueber on cs.id = ueber.serial_id
                 where ec.id is not null OR cdnc.id is not null OR subcert.id is not null OR idcert.id is not null OR cacert.id is not null OR ueber.id is not null;

-- After the changeset is applied, verify that the revoked count is the same as that
-- of the non-referenced query.
select count(*) from cp_cert_serial where revoked = true;

```


mstead:
[Convert org.candlepin.model.CertificateSerial#revoked from an ebedded @Formula to a column](https://trello.com/c/X9F8P3Wr/298-convert-org-candlepin-model-certificateserial-revoked-from-an-ebedded-formula-to-a-column)

server/spec/content_delivery_network_spec.rb
```diff
@@ -70,4 +70,84 @@
     cdn.certificate['cert'].should == "test-cert"
   end
 
+  it 'should allow candlepin to create certificate serial' do
```
barnabycourt: Why is the serial able to be updated via the REST interface?

server/src/main/java/org/candlepin/controller/CdnManager.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Deletes the specified {@link Cdn}.
+     *
+     * @param cdn the cdn to delete.
+     */
+    @Transactional
+    public void deleteCdn(Cdn cdn) {
+        // FIXME When a Cdn is referenced by a Pool, and is deleted, an exception occurs.
```
barnabycourt: Does this still need to be fixed?

server/spec/content_delivery_network_spec.rb
mstead: The API currently allows this and I just added a test for it since it was missing. If the serial is included with the certificate provided in the request body when creating the CDN, candlepin will persist it. If we allow them to set it manually when creating the certificate, why not let them be able to update it?

server/src/main/java/org/candlepin/controller/CdnManager.java
mstead: This was part of an existing issue that I found while making this change. This doesn't need to be addressed as part of this PR as it exists in older candlepin versions. I just added it as a reminder to myself. I'm working on addressing this right now. I'm surprised that this hasn't been hit before. I'm assuming that this is because no one is using it in this way?

server/spec/content_delivery_network_spec.rb
barnabycourt: The test is good to explain the current behavior. The reason I ask about pulling the value from the certificate is that if we used the value from inside the cert there would be no chance of it getting out of sync. I'll file it under ideas for future improvement since this is merely documenting the current API

server/src/main/java/org/candlepin/controller/CdnManager.java
barnabycourt: ack

PR #1535, opened 2017-04-03
Updated spec tests to escape query params and use full date strings

Ceiu:
- Updated several spec tests to use complete dates that include
  time zones to address an issue with pool visibility when created
  on a system with a time zone using a positive offset
- Updated candlepin_api to escape URIs and query parameters
  separately and more completely
- Updated several rspec tests and helper methods to no longer
  include query parameters when building request URIs
- Updated the DateFormat annotation and DateFormatter handler to
  process either ISO8601 date strings or lists of specifically
  supported formats
- Updated ConsumerResource and OwnerResource to use the changes
  made to the DateFormat annotation

### DIFF
--- a/server/src/main/java/org/candlepin/resteasy/DateFormatter.java
+++ b/server/src/main/java/org/candlepin/resteasy/DateFormatter.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2009 - 2012 Red Hat, Inc.
+ * Copyright (c) 2009 - 2017 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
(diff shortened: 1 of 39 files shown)


Ceiu:
Rebased against current master

server/src/main/java/org/candlepin/resteasy/DateFormatter.java
```diff
@@ -15,34 +15,74 @@
...
+                }
+
+                // If we made it here, we're out of formats to use to parse the date
+
+                // TODO: How do we get an i18n instance at this point to translate this?
+                String message = "Unable to parse date parameter: " + value;
+                throw new RuntimeException(message);
+            }
+            else {
+                try {
```
wottop: Does the above loop include ISO 8601? If not, it will not get tried if there are other candidates.

wottop: Just this one comment

Ceiu: The format loop does not include the ISO8601 standard by design. My intent was that if you're specifying a format or formats and you want  ISO8601, you would specify it. Having to tell it not to do that would be clunkier than including it.

Think it needs a revisit?

mstead: I like it as designed. If a format(s) is specified that doesn't work, fail early. I'd suggest that we beef up the error message a little, or add some logging, to describe what it attempted to do before failing.

awood: How about a compromise between what wottop suggested and Ceiu brought up?  I.e. the ISO8601 format is provided as a constant within the `DateFormat` annotation so that it's easy to add.  `@DateFormat(DateFormat.ISO8601)`

Ceiu: @awood Not opposed to that idea, but the difficulty there is we would need to specify all of them in an array and the combine them with the others we want, which wouldn't be usable in the annotation. If it were just one format, or the formatter backing this supported optional bits, that'd be a nice addition.

Also, for clarity, ```@DateFormat``` with no parameters gets us ISO8601 parsing of a date. The clunkiness comes from when we want both ISO8601 and something else (```@DateFormat({"myformathere", ISO8601_LONG, ISO8601_SHORT, etc.})```) vs when it's always a default and we don't want it (```@DateFormat(formats={"myonlyformat"}, omitISO8601=true)```).

Ceiu:
Note: The current rspec failure is a randomish timing collision that should be addressed by updating the test to not fail in this case.

Ceiu:

wottop:
ack

PR #1496, opened 2017-03-02
Create "share" consumers and allow them to bind.

awood:

PR #1536, opened 2017-04-06
Vritant's changes for pool sharing

vritant:
@awood  @mstead 
this is what I have so far, im probably half way through.
I would also like to add some tests myself, coming soon...

discussion points:
- sharing pools with infinite quantity
- cant copy flattened attrs to shared pools.
- some paths skipping on manfiest consumers as well, in this PR, fyi. (  postBindVirtLimit )
- future - optimize product resolves? caching?
- should we allow sharing pools that have restricted to user name? how does that even gets set?
- should we expose order number / contract number / account number to recipient orgs? whats the impact on the client seeing weird account numbers?
- during auto attach, what pools are preferred first? shares or non shares?
- no modifying ent certs need to be generated for share bind imo.
- why set compliance status on shared consumers?
- on decrementing bonus pool quantity for infinite pools, we check both exported and shared quantity now, FYI.
- consumer bind validation - accept pools from post body. 
- can shared consumers "activate subscription" ?
- shared consumers auto heal should be false, yeah?
- do not set owners default service level on share consumers
- do not check for unaccpted subscription terms for share consumers.
- should you be allowed to list pools for shared consumers? if yes, should we exclude unmapped guest pools and other restricted from sharing?


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1583,11 +1583,12 @@ else if (consumer.isShare()) {
         // shares don't need entitlement certificate since they don't talk to the CDN
         if (!consumer.isShare()) {
             handler.handleSelfCertificates(consumer, poolQuantities, entitlements);
+            this.ecGenerator.regenerateCertificatesByEntitlementIds(
```
vritant: @awood please double check

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1488,6 +1517,10 @@ public Response getContentAccessBody(
         @HeaderParam("If-Modified-Since") @DateFormat ("EEE, dd MMM yyyy HH:mm:ss z") Date since) {
         log.debug("Getting content access certificate for consumer: {}", consumerUuid);
         Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
+        if (consumer.isShare()) {
```
vritant: @vritant TODO: update API-doc

vritant:
closing this, code changes will now go in #1496

PR #1537, opened 2017-04-07
Database failfast

dsarman:
This extends the current failfast check that runs every 10 secs and uses it to check if the connection to database is ok.
If it is not, SUSPEND mode is entered and vice versa.
Before the transition from SUSPEND to NORMAL mode due to DB recovery all database connection in the pool will be closed.


kahowell:
retest this please

Ceiu:
Closing this PR, as it's been sitting for a long time without change and it's not yet ready for review. Feel free to reopen this as more work is completed on the task.

PR #1541, opened 2017-04-11
1441700: Fix issues with Status caching invalidation

mstead:
Sometimes the JCache we used for caching the Status
would not invalidate after the configured 5 seconds.

Because of this, we implemented our own caching
mechinism for Status due to cache invalidation
issues we were having with the ehcache-jcache
implementation.

The new StatusCache uses a static Status instance to
track the latest computed candlepin status.
When a status is set in the cache it has a
5 second lifespan.


mstead:
retest this please

mstead:
retest this please

mstead:
[1441700: Fix issues with Status caching invalidation](https://trello.com/c/WfXuYsJs/348-1441700-fix-issues-with-status-caching-invalidation)

server/src/main/java/org/candlepin/cache/StatusCache.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Clear the cached status to allow forced regeneration.
+     */
+    public void clear() {
+        synchronized (LOCK) {
+            reset();
+        }
+    }
+
+    private void reset() {
```
Ceiu: Do we want this to be synchronized as well?

server/src/main/java/org/candlepin/cache/StatusCache.java
```diff
@@ -0,0 +1,83 @@
...
+     * Gets the current Status in the cache.
+     * @return the current {@link Status}, null if it has not yet been set.
+     */
+    public Status getStatus() {
+        synchronized (LOCK) {
+            // Check if we can return our cached status
+            if (cachedStatus != null && cachedStatus.getTimeUTC() != null) {
+                long lastUpdateTimeSkew = System.currentTimeMillis() - cachedStatus.getTimeUTC().getTime();
+                if (lastUpdateTimeSkew <= STATUS_CACHE_TTL) {
+                    log.debug("Returning cached status. Last Update Date: {}, Age: {}",
```
Ceiu: Consistency issue with the debug statements I gave you: the age here doesn't have the "ms" on it, but it does below.

server/src/main/java/org/candlepin/cache/StatusCache.java
mstead: No, should be fine since it is only used within sync blocks. Any public method in this class should be synchronized anyway.

server/src/main/java/org/candlepin/cache/StatusCache.java
mstead: I'll make this change.

mstead:
retest this please

server/src/main/java/org/candlepin/cache/StatusCache.java
Ceiu: ack

Ceiu:
ACK

PR #1542, opened 2017-04-18
Mark modifying ents as dirty if not already

vritant:
HQL didnt help here, there was a bug that always resulted in a weird error "References to collections must be define a SQL alias" 
This happens due to the left outer join with products in the inner class fwiw.
resorting to native query here.

There was also a unit test in PoolManagerTest that did not make sense to me, please take a second look, I think what I did makes sense. we were ensuring a method isnt called before, but that method is only used in tests a.t.m.

### DIFF
--- a/server/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/server/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -575,17 +575,55 @@ public class EntitlementCurator extends AbstractHibernateCurator<Entitlement> {
     }
 
     /**
-     * A version of list Modifying that finds Entitlements that modify
-     * input entitlements.
-     * When dealing with large amount of entitlements for which it is necessary
-     * to determine their modifier products. Today this is used in the revoke case
-     * or when we update ent quantity. In an uncoming PR, we will change this implementation
-     * to directly mark these ents as dirty, instead of returning the ids.
-     * @param entitlement
-     * @return Entitlements that are being modified by the input entitlements
+     * - A version of list Modifying that finds Entitlements that modify
+     * input entitlements, and updates the dirty flag to true.
+     * - useful when dealing with large amount of entitlements for which it is necessary
+     * to determine their modifier products.
+     * - Today this is used in the revoke case or when we update ent quantity.
+     *
+     * pseudo code:
+     * UPDATE entitlements
+     * SET dirty = true
+     * WHERE id IN (
+     *              -- a pass through query to work around
+     *              -- mysql restriction of not updating the same
+     *              -- table that is selected from.
+     *              SELECT * FROM (
+     *                             SELECT entitlement.id
+     *                             FROM relevant tables
+     *                             -- exclude if already dirty
+     *                             WHERE dirty = false
+     *                             -- exclude ents passed in as args
+     *                             AND (ent.id NOT IN (:ein))
+     *                             AND (
+     *                                  EXISTS (
+     *                                          SELECT entIn.id
+     *                                          -- ensure only modifying ents
+     *                                          FROM relevant tables
+     *                                          WHERE (entIn.id IN (:ein))
+     *                                         )
+     *                                  )
+     *                             -- ensure we do not cross in clause limit
+     *                             LIMIT  999
+     *                            ) as e
+     *               )
+     *
+     * @param entitlements
+     * @return number of entitlements that were marked as dirty
      */
-    public Collection<String> batchListModifying(Iterable<Entitlement> entitlements) {
-        List<String> eids = new LinkedList<String>();
+    public int markModifyingEntsDirty(Iterable<Entitlement> entitlements) {
+        return markModifyingEntsDirtyForTesting(getEntityManager(), entitlements);
+    }
+
+    /**
+     * a convenience method for helping with tests, which can pass a mock for entityManager
+     * @param entityManager
+     * @param entitlements
+     * @return number of entitlements that were marked as dirty
+     */
+    public int markModifyingEntsDirtyForTesting(EntityManager entityManager, Iterable<Entitlement>
+        entitlements) {
+        int result = 0;
 
         if (entitlements != null && entitlements.iterator().hasNext()) {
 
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -573,60 +573,68 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
-                "                    LEFT JOIN inPool.providedProducts inProvidedProd" +
-                "                WHERE eIn in (:ein) AND inConsumer = eOut.consumer AND" +
-                "                    inPool.endDate >= outPool.startDate AND" +
-                "                    inPool.startDate <= outPool.endDate AND" +
-                "                    (inProvidedProd.id = outModProdId OR inMktProd.id = outModProdId)" +
-                "        )";
 
-            Query query = this.getEntityManager().createQuery(hql);
+            String sql = "UPDATE cp_entitlement" +
+                "     SET dirty = false" +
```
barnabycourt: Should this be setting dirty=true to force entitlement regeneration?

vritant:
@candlepin/jenkins-bot test this please

server/src/main/java/org/candlepin/model/EntitlementCurator.java
vritant: @barnabycourt all fixed now, this PR is ready.

vritant:
@barnabycourt do you mind if I add the ASAP label given the benefit it could provide to hosted release?

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -573,60 +573,69 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
-                "                WHERE eIn in (:ein) AND inConsumer = eOut.consumer AND" +
-                "                    inPool.endDate >= outPool.startDate AND" +
-                "                    inPool.startDate <= outPool.endDate AND" +
-                "                    (inProvidedProd.id = outModProdId OR inMktProd.id = outModProdId)" +
-                "        )";
 
-            Query query = this.getEntityManager().createQuery(hql);
+            String sql = "UPDATE cp_entitlement" +
+                "     SET dirty = true" +
+                "     WHERE id IN (" +
```
Ceiu: This has potential to hit the IN operator limit.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -573,60 +573,69 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
+                "                            LEFT OUTER JOIN cp2_pool_provided_products providedPrIn" +
+                "                                ON poolIn.id = providedPrIn.pool_id" +
+                "                            LEFT OUTER JOIN cp2_products productIn2" +
+                "                                ON providedPrIn.product_uuid = productIn2.uuid" +
+                "                            WHERE (entIn.id IN (:ein))" +
+                "                               AND consumerIn.id = ent.consumer_id" +
+                "                               AND poolIn.endDate >= pool.startDate" +
+                "                               AND poolIn.startDate <= pool.endDate" +
+                "                               AND (productIn2.product_id = modifiedPr.element " +
+                "                                 OR productIn.product_id = modifiedPr.element)))) as e)";
```
Ceiu: Just to make sure I understand this correctly, I've written what I interpret the intent of this query to be in a sort of pseudo code. Could you verify? If there is anything incorrect in my interpretation, could you provide one?

```
- For each entitlement:
    - Get a set of modified products:
        - Get the entilement's pool
        - Get the pool's product and provided products (Note: NOT its derived products)
        - Get the product's content
        - Get the content's modified products
        - Fetch the modified products for entitlements with pools that haven't yet expired*,
          matching the given owner and with IDs NOT in the set of given entitlements

    - For each modified product:
        - Check for related entitlements
            - Get the entitlements used by the given consumer
            - Get each entitlement's pool
            - Get the pool's products and provided products
            - Fetch the products with IDs matching the modified product and overlapping pool durations
              and the entitlement ID is in the given set of entitlements

        - If there is a related entitlement, set the entitlement's dirty flag


    * This will also match future pools
```

vritant: @Ceiu  seems you got it right, just one difference:
"and the entitlement ID is **NOT** in the given set of entitlements"

if it helps, I used the hibernate generated sql queries from the logs to write this query and then tweaked it to make it work, so perhaps doing the same might make the review easier?

server/src/main/java/org/candlepin/model/EntitlementCurator.java
vritant: @Ceiu  fixed. perhaps I need to write a unit test though.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -573,60 +573,69 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
      * input entitlements.
      * When dealing with large amount of entitlements for which it is necessary
      * to determine their modifier products.
```
vritant: done

vritant:

server/src/main/java/org/candlepin/model/EntitlementCurator.java
vritant: added a unit test too.

Ceiu: Those changes don't address it, from what I can tell. The problem is that the first inner query is feeding this IN operator using the inverse of the entitlement blocks. There's no guarantee we won't pull more entitlement IDs back than what we fed in.

vritant: I am not sure I see the problem, I thought the limit clause should fix it, but perhaps I dont see it yet. could you describe a problem scenario please? or better yet, demonstrate the issue on a db? let us chat online.

vritant:
[During bind, dont fetch modifying ents to mark dirty, instead just mark them via query](https://trello.com/c/1dasfoP8/358-during-bind-dont-fetch-modifying-ents-to-mark-dirty-instead-just-mark-them-via-query)

vritant:
[During bind, dont fetch modifying ents to mark dirty, instead just mark them via query](https://trello.com/c/1dasfoP8/358-during-bind-dont-fetch-modifying-ents-to-mark-dirty-instead-just-mark-them-via-query)

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -594,51 +632,71 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
+                "             LIMIT " + getInBlockSize() + " ) as e)";
+
+            Query query = entityManager.createNativeQuery(sql);
+            query.setParameter("owner", entitlements.iterator().next().getOwner());
+
+            int j = 0;
+            for (List<Entitlement> block : splitEnts) {
+                query.setParameter("not_in_" + j, block);
+                j++;
+            }
```
Ceiu: This loop appears to not work as one would expect. It's always going to pull the last block of entitlements here. Maybe this was intended to go in the loop below?

server/src/main/java/org/candlepin/model/EntitlementCurator.java
Ceiu: To continue on this from before, let's say we have a limit of 10, and our inner query would normally fetch 12 entitlements. The limited query here will safely grab the first 10, but the last two will be dropped, which will give us incorrect results.

This may be a case where doing it in multiple queries is the fastest and safest.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -568,17 +575,146 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
+                "                                ON poolIn.id = providedPrIn.pool_id" +
+                "                            LEFT OUTER JOIN cp2_products productIn2" +
+                "                                ON providedPrIn.product_uuid = productIn2.uuid" +
+                "                            WHERE (entIn.id IN (:ein))" +
+                "                               AND consumerIn.id = ent.consumer_id" +
+                "                               AND poolIn.endDate >= pool.startDate" +
+                "                               AND poolIn.startDate <= pool.endDate" +
+                "                               AND (productIn2.product_id = modifiedPr.element" +
+                "                                 OR productIn.product_id = modifiedPr.element)))" +
+                "             LIMIT " + getInBlockSize() + " ) as e)";
```
Ceiu: I commented on this earlier, but it seems that was on a previous commit. This limit here will protect us from the DB blowing up, but it may cause the query to ignore valid entitlements by truncating the list to make it not explode. In such a case, we're silently doing the wrong thing, which is arguably worse than just crashing out.

Ceiu:
ACK, pending additional comments on the LIMIT clause and successful automated tests.

PR #1543, opened 2017-04-19
Awood/lock less monster interfaces

awood:
Note that most of the code that does anything is commented out because it doesn't quite compile yet.


vritant:
The failing spec tests are not related to this PR and are being fixed in #1544

PR #1546, opened 2017-04-20
Added indexes to various tables (CP 2.0)

Ceiu:
- Added an index to cp2_products.product_id and cp2_content.content_id.
  These should help improve overall performance, as many of our queries
  still do RHID-oriented lookups rather than using the UUID.
- Added an index to cp_pool_products.product_id, which should
  significantly improve migration times


Ceiu:
Reverted ID and preexisting changeset changes.

PR #1552, opened 2017-04-25
1445012: Fix permission issue on getting pools for an AK

mstead:
Fetching the pools for an activation key was
being scoped to a Super Admin. Loosened this
permission to anyone with access to the AK.


Ceiu:
ACK

PR #1553, opened 2017-04-25
Added the skip_existing option to the populate hosted DB task

Ceiu:
- Added support for a skip_existing query parameter/option for
  the populate hosted DB task which will skip fetching and updating
  product information for products which already exist in the local
  database
- When the hosted DB population job runs, it will now verify that the
  products it requested were fetched; for each product it determines
  was omitted, a warning will be printed. Additionally, the final
  job status now includes the number of "unresolved references" hit.


mstead:
retest this please

Ceiu:
The qpid failure looks to be something unrelated to this test (the qpid directory/file not existing).

PR #1557, opened 2017-04-27
Consolidated and refactored the database locking methods

Ceiu:
- Several methods used for applying database/entity locks have been
  generalized and refactored; extraneous methods have been removed
- Fixed a bug that could allow a temporary service level change to
  be permanently applied
- Fixed a bug that prevented consumers from being refreshed when
  the lockAndLoad method was called

### DIFF
--- a/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
+++ b/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
@@ -21,12 +21,13 @@ import org.candlepin.common.paging.Page;
 import org.candlepin.common.paging.PageRequest;
 import org.candlepin.guice.PrincipalProvider;
 
+import org.apache.commons.collections.CollectionUtils;
+
 import com.google.common.collect.Iterables;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.google.inject.persist.Transactional;
 
-import org.apache.commons.collections.CollectionUtils;
 import org.hibernate.Criteria;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
@@ -37,7 +38,13 @@ import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Projection;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
+import org.hibernate.engine.spi.EntityKey;
+import org.hibernate.engine.spi.PersistenceContext;
 import org.hibernate.internal.CriteriaImpl;
+import org.hibernate.internal.SessionImpl;
+import org.hibernate.jpa.AvailableSettings;
+import org.hibernate.metadata.ClassMetadata;
+import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.transform.ResultTransformer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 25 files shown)


server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -767,47 +770,205 @@ protected int safeSQLUpdateWithCollection(String sql, Collection<?> collection,
         return count;
     }
 
-    public List<E> lockAndLoadBatch(Iterable<String> ids, String entityName, String keyName) {
-        List<E> result = new LinkedList<E>();
+    /**
+     * Locks the specified entity using the given lock mode. Note that the entity will not be
+     * refresh as a result of a call to this method. If the entity needs to be locked and
```
awood: s/refresh/refreshed/?

Ceiu:
There's still a bit more cleanup work to do here, pending a post-scrum discussion next week.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
Ceiu: ack, fixed in my local copy.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -767,47 +770,205 @@ protected int safeSQLUpdateWithCollection(String sql, Collection<?> collection,
...
-            //been issued.
-            for (E e : result) {
-                getEntityManager().refresh(e);
-            }
+        if (entities == null) {
+            return Collections.<E>emptyList();
         }
 
-        return result;
+        // We redeclare the collection here so we don't require the final modifier in subclass
```
Ceiu: The bulk of this change was started because there's a need to add sorting to our bulk locking operations to avoid some deadlock situations. This method offloads the bulk of its work to the ```lockAndLoadByColumn``` method, which we know (through blackbox knowledge) is going to put the entities in a sorted set via loop if the iterable it receives isn't a Collection. So, rather than loop through the entities there and pull the IDs, we wrap the whole thing in the anon iterable and do the sorting and ID pulling in one loop, rather than two.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -767,47 +770,205 @@ protected int safeSQLUpdateWithCollection(String sql, Collection<?> collection,
...
+     * @param keyName
+     *  The name of the indexed column/property to examine (i.e. id or key)
+     *
+     * @param value
+     *  The value the specified key must match
+     *
+     * @return
+     *  A locked entity instance, or null if a matching entity could not be found
+     */
+    protected E lockAndLoadByKey(String entityName, String keyName, Object value) {
```
Ceiu: I don't see why not. I'll get that switched over.

Ceiu: Unless I'm overlooking something, it doesn't look like this will be an option here afterall. Switching over to TypedQuery gives us no control over the level 2 cache mode, which puts us one 
step closer to needing a full refresh on everything.

awood: I think it's possible via a "hint"

```
.setHint( "javax.persistence.cache.storeMode", CacheStoreMode.REFRESH )
```

Ceiu:
Rebased against current master

PR #1558, opened 2017-04-27
Added the skip_existing option to the populate hosted DB task (0.9.51)

Ceiu:
- Added support for a skip_existing query parameter/option for
  the populate hosted DB task which will skip fetching and updating
  product information for products which already exist in the local
  database
- When the hosted DB population job runs, it will now verify that the
  products it requested were fetched; for each product it determines
  was omitted, a warning will be printed. Additionally, the final
  job status now includes the number of "unresolved references" hit.


Ceiu:
As with the 0.9.54 build, the qpid failure is not one resulting from these changes -- the build system does not like running that test on pre-2.0 versions.

PR #1559, opened 2017-04-27
Filter unsharable pools when listing pools for share consumers

awood:


vritant:
@awood I dont think this is the right approach, we need to skip java script filterPools / validate_pools_list completely. here is a scenario that demonstrates an example issue, though im sure there can be more such examples:

``` add this to multi_org_spec.rb:
  it 'does list pools for share even if share consumer has one of the same product' do
    prod = create_product(nil, nil, :owner => @owner1['key'])
    pool = create_pool_and_subscription(@owner1['key'], prod.id, 10)
    pools_length = @user_client.list_owner_pools(@owner1['key'], :consumer => share_consumer['uuid']).length
    puts "number of pools that can be shared: #{pools_length}"
    ent = @user_client.consume_pool(pool['id'], :uuid => share_consumer['uuid'], :quantity => 3)[0]
    puts "created an entitlement of quantity: #{ent['quantity']}"
    pools_length =  @user_client.list_owner_pools(@owner1['key'], :consumer => share_consumer['uuid']).length
    puts "number of pools that can be shared now: #{pools_length}"
    ent = @user_client.consume_pool(pool['id'], :uuid => share_consumer['uuid'], :quantity => 4)[0]
    puts "although we can still bind the same pool with quantity: #{ent['quantity']}"
  end
```
 output:
> Multi Org Shares
number of pools that can be shared: 1
created an entitlement of quantity: 3
number of pools that can be shared now: 0
although we can still bind the same pool with quantity: 4

PR #1560, opened 2017-05-02
speeden up testFetchingPoolProvidedProductIdsWithVaryingPoolSetSizes

vritant:


vritant:

Ceiu:
NACKing for a couple reasons:

- The biggest thing we need to test when it comes to blocking on this is verifying that the various collection sizes are safe: zero, less than block size, block size, block size + 1, block size + N, block size * M, block size * M + 1, block size * M + N. As soon as we drop those out and start mocking the results, the test itself loses value. Worse, to properly stub out the various objects for the test, we need to utilize blackbox knowledge in how the method under test works -- effectively tightly coupling the test to that particular implementation. Unless I'm misreading it, this is what I was looking to avoid, as it devalues and overcomplicates the test at the expense of a lot of engineering time.

- @awood brought up an idea that gives purpose to declaring the block size as something configurable and setting it to an optimal size at runtime to match our respective backend's limits. As a side effect of this, we could implement another idea brought up during the discussion in simply changing the block size for the test. I believe he's planning on working on that after his current task, so we'd end up obsoleting much of this in the very near future anyhow.

vritant:
@Ceiu sure, but I do have a question. could you elaborate on "various collection sizes are safe"? what exactly do we want to verify? is it enough to verify that the query compiles ? or perhaps we want to make sure the ids are split into appropriate blocks and none of the block has more than allowed limit?

basically, without focusing on how the test is implemented, I think we should work on a test wish list and discuss about that first.

Ceiu:
In the general case, no, the query compiling/running is not enough. In some of the earlier versions of this type of code (specifically, before we started using Iterables.partition), and even in some of the SQL builders I've written, we've hit issues where there was an off-by-one issue around the boundaries of the blocks. Results would go missing or would get doubly counted in the results. So, if we were to focus on the blocking behavior (which, again, is blackbox knowledge of the implementation), we'd need to know that (a) the query was executed N times, (b) the number of parameters in a given block is M, (c) the arguments aren't duplicated or dropped and (d) the query/ies are compiled and executed successfully. That's a lot of work to test a very specific implementation of a method rather than just testing the method itself with varying sizes of inputs.

I'm more in favor of making the block size not a constant so it can be manipulated, which happens to be a side effect of the RFE al will be working on soon. This way the only change to the test logic becomes the block size as we configure it.

awood:
I'm going to close this because I think the underlying speed issue is addressed with #1564

PR #1561, opened 2017-05-02
1440909: Prioritize host based pools over regular ones for guests

cnsnyder:
This PR implements the following design[1] for candlepin 0.9.54.
In short, this PR causes candlepin to perform an auto attach against just the pools provided by the host the guest is currently known to be on before performing a normal auto attach (for any products non-compliant or partially compliant post the first auto attach).

[1]: https://docs.google.com/document/d/1fM19-FFxUAkt3T2kf_g1ZbjFUZBl5oA_LzlrN5WZtLs/edit#

### DIFF
--- a/server/src/main/java/org/candlepin/controller/Entitler.java
+++ b/server/src/main/java/org/candlepin/controller/Entitler.java
@@ -34,6 +34,8 @@ import org.candlepin.model.Product;
 import org.candlepin.model.ProductPoolAttribute;
 import org.candlepin.model.ProvidedProduct;
 import org.candlepin.policy.EntitlementRefusedException;
+import org.candlepin.policy.js.compliance.ComplianceRules;
+import org.candlepin.policy.js.compliance.ComplianceStatus;
 import org.candlepin.policy.js.entitlement.EntitlementRulesTranslator;
 import org.candlepin.resource.dto.AutobindData;
 import org.candlepin.service.ProductServiceAdapter;
@@ -67,6 +69,7 @@ public class Entitler {
     private EventFactory evtFactory;
     private EventSink sink;
     private ConsumerCurator consumerCurator;
+    private ComplianceRules complianceRules;
     private EntitlementRulesTranslator messageTranslator;
     private EntitlementCurator entitlementCurator;
     private Configuration config;
@@ -79,7 +82,7 @@ public class Entitler {
     public Entitler(PoolManager pm, ConsumerCurator cc, I18n i18n, EventFactory evtFactory,
         EventSink sink, EntitlementRulesTranslator messageTranslator,
         EntitlementCurator entitlementCurator, Configuration config,
-        PoolCurator poolCurator, ProductServiceAdapter productAdapter) {
+        PoolCurator poolCurator, ProductServiceAdapter productAdapter, ComplianceRules complianceRules) {
 
         this.poolManager = pm;
         this.i18n = i18n;
@@ -91,6 +94,7 @@ public class Entitler {
         this.config = config;
         this.poolCurator = poolCurator;
         this.productAdapter = productAdapter;
+        this.complianceRules = complianceRules;
     }
 
     public List<Entitlement> bindByPool(String poolId, String consumeruuid,
@@ -250,11 +254,16 @@ public class Entitler {
             devPool = poolManager.createPool(assembleDevPool(consumer, sku));
             data.setPossiblePools(Arrays.asList(devPool.getId()));
         }
-
         // Attempt to create entitlements:
         try {
             // the pools are only used to bind the guest
-            List<Entitlement> entitlements = poolManager.entitleByProducts(data);
+            // Here is where to get the pools that are provided by the host and attempt to entitleByProducts on those
+            // only, then we should check the compliance of the guest, if no good, entitleByProducts again, this time
+            // without the host pools.
+            List<Entitlement> entitlements = new ArrayList<Entitlement>();
+            if (entitleGuest(data, entitlements)) {
+                entitlements.addAll(poolManager.entitleByProducts(data));
+            }
             log.debug("Created entitlements: " + entitlements);
             return entitlements;
         }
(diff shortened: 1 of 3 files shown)


cnsnyder:
Note: I will squash these commits once the review is in.

cnsnyder:
Also note: Once this has been accepted as a decent way of implementing this RFE I will port to CP 2.0

cnsnyder:
[1440909 - [RFE] Prioritize Host provided pools first during auto attach (and guest migration) for virtual systems](https://trello.com/c/PveYsZFP/349-1440909-rfe-prioritize-host-provided-pools-first-during-auto-attach-and-guest-migration-for-virtual-systems)

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -250,11 +254,44 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+                        if (!status.isCompliant()) {
+                            // We must not be compliant so we'll try to bind by products
+                            data.setPossiblePools(null);
+                            // Only auto attach for the non-compliant and partially compliant products
+                            Set<String> remainingProductsToCheck = status.getNonCompliantProducts();
+                            remainingProductsToCheck.addAll(status.getPartiallyCompliantProducts().keySet());
+                            data.setProductIds(remainingProductsToCheck.toArray(new String[remainingProductsToCheck.size()]));
+                            entitlements.addAll(poolManager.entitleByProducts(data));
+                        }
+                    } else {
```
vritant: any chance we can reduce the size of this if else ladder using a boolean or perhaps turning the if else ladder around? looks like we are doing the same operation under a lot of conditional branching.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -812,6 +813,22 @@ public boolean exists(Pool pool) {
                 .getSingleResult() > 0;
     }
 
+    /**
+     * Uses a database query to list pool ids of pools that require the given host for the given owner
+     * @param host the host consumer that is required
+     * @param owner the owner of the desired pools
+     * @return A list of pool ids that require the given host owned by the given owner
+     */
+    public List<String> getPoolsProvidedByHost(Consumer host, Owner owner) {
```
vritant: do we need the owner as a separate argument? can we get that information from the consumer?

cnsnyder:
@candlepin-pull-request-bot please retest this

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -250,11 +254,44 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
         // Attempt to create entitlements:
         try {
             // the pools are only used to bind the guest
-            List<Entitlement> entitlements = poolManager.entitleByProducts(data);
+            // Here is where to get the pools that are provided by the host and attempt to entitleByProducts on those
+            // only, then we should check the compliance of the guest, if no good, entitleByProducts again, this time
+            // without the host pools.
+            // We should only attempt to use the host pools if there have not already been added any pools.
+            List<Entitlement> entitlements = new ArrayList<Entitlement>();
+            if (consumer.hasFact("virt.uuid") && "true".equalsIgnoreCase(consumer.getFact("virt.is_guest"))
```
cnsnyder: I can't seem to find that method in the 0.9.54 branch. Is that in master only?

cnsnyder:
This should now address all prior comments (either with discussion points or in actual code change).

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -269,6 +279,40 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
         }
     }
 
+    // This is a helper method to attempt to entitle a guest using host provided pools, if available
+    private boolean entitleGuest(AutobindData data, List<Entitlement> entitlements) throws EntitlementRefusedException {
```
vritant: just a suggestion: rename this method to entitleGuestUsingHostRestrictedPools

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -269,6 +279,40 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
         }
     }
 
+    // This is a helper method to attempt to entitle a guest using host provided pools, if available
```
cnsnyder: Check!

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -269,6 +279,40 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+    private boolean entitleGuest(AutobindData data, List<Entitlement> entitlements) throws EntitlementRefusedException {
+        Consumer host;
+        List<String> hostProvidedPools;
+        Consumer consumer = data.getConsumer();
+        Owner owner = consumer.getOwner();
+
+        if (!consumer.hasFact("virt.uuid") || !"true".equalsIgnoreCase(consumer.getFact("virt.is_guest"))
+            || !data.getPossiblePools().isEmpty()) {
```
cnsnyder: It seems most safe to add one.

cnsnyder:
PR updated once more based on review. I've also squashed the commits to one. Please let me know if there is anything else left.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -269,6 +278,57 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+            return true;
+        }
+        hostProvidedPools = poolCurator.getPoolsProvidedByHost(host);
+        if (hostProvidedPools.isEmpty()) {
+            return true;
+        }
+        data.setPossiblePools(hostProvidedPools);
+        entitlements.addAll(poolManager.entitleByProducts(data));
+        ComplianceStatus status;
+        if (data.getOnDate() != null) {
```
vritant: null check is not needed here. getStatus takes care of the case where date is null. please also consider the other arguments eventually passed into getStatus.

vritant:
@cnsnyder I believe tests should pass on candlepin 0.9.54.

barnabycourt:
If this is targetting 0.9.54 there should be a BZ at the beginning of the title

vritant:
@cnsnyder I am tempted to nack this, but I know I have brought this issue up before already, so I will request a second reviewer and let that dev be the judge.
My concern is this : it seems we already have an implementation to increase priority of pools with required host attribute [here](https://github.com/candlepin/candlepin/blob/master/server/src/main/resources/rules/rules.js#L334) . I think we should investigate if that is sufficient, or if its working as designed or not. my concern is especially since we have increased the autobind time in this PR for the average ( owners with not many pools) virtual guest a lot imho. .

cnsnyder:
@vritant We can consider a different implementation but I do know that some change somewhere is necessary as the prioritization of host pools is not working as designed.

mstead:
> @cnsnyder I am tempted to nack this, but I know I have brought this issue up before already, so I will request a second reviewer and let that dev be the judge.
> My concern is this : it seems we already have an implementation to increase priority of pools with required host attribute here . I think we should investigate if that is sufficient, or if its working as designed or not. my concern is especially since we have increased the autobind time in this PR for the average ( owners with not many pools) virtual guest a lot imho. .

@cnsnyder is correct in that the rules are definitely not behaving correctly and something has to change. I do agree with @vritant though that we should take a look at the current pool prioritization to see what is going on there, especially since we've increased the autobind time. The possibility of a double compliance status call and the extra queries is a little concerning.

It would be beneficial to at least find out why the physical pool is getting selected over the host restricted ones in the current rules. As @vritant pointed out, we do have *some* prioritization on it right now.

Perhaps the priority increment for host restricted pools is just too low. What happens if we bump it some. Perhaps installed product coverage is bumping it past the threshold or something. If this isn't the cause, perhaps we could look at leveraging the pool selection code to process the host_restricted pools right away. Might be able to leverage the entitlement group functionality to create a some sort of 'host_restricted' grouping.

Going forward, we will be doing some work to improve the 'prioritization' of pools -- we can make it better then.

If you have trouble making this happen in the rules, let me know and we can chat some more about it.

Don't trash this PR yet, but please do some investigation.

cnsnyder:
Closing in favor of #1583

PR #1562, opened 2017-05-04
Updates for owner delete

wottop:
If share org is deleted, then the share consumer is removed from master org
If REST call comes to delete master org while there is a share, then it will
 fail with a message unless the force query parameter is used.

### DIFF
--- a/server/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/server/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -305,6 +305,25 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
         return this.cpQueryFactory.<Consumer>buildQuery(this.currentSession(), criteria);
     }
 
+    @SuppressWarnings("unchecked")
+    @Transactional
+    public CandlepinQuery<Consumer> listByRecipientOwner(Owner owner) {
+        DetachedCriteria criteria = this.createSecureDetachedCriteria()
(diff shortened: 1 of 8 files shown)


awood:
bot, retest this please

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -338,6 +338,25 @@ public Consumer getConsumer(String uuid) {
...
+
+        return this.cpQueryFactory.<Consumer>buildQuery(this.currentSession(), criteria);
+    }
+
+    public boolean doesShareConsumerExist(Owner owner) {
+        long result = (Long) createSecureCriteria()
+            .add(Restrictions.eq("owner", owner))
+            .add(Restrictions.isNotNull("recipientOwnerKey"))
+            .setProjection(Projections.count("id"))
+            .uniqueResult();
```
awood: `uniqueResult` returns null if there is not a result, not 0.  I would just drop `uniqueResult` entirely, grab the count, and `return count != 0`

awood: This is an invalid comment since `count` will always return something.

wottop:
jenkins: retest this please

awood:
Bot, retest this please

wottop:
[Multi-Org: Organizations](https://trello.com/c/zr7AlfwI/316-multi-org-organizations)

PR #1563, opened 2017-05-05
Vritant/lock less monster

vritant:


server/src/main/java/org/candlepin/bind/BindContext.java
```diff
@@ -0,0 +1,136 @@
...
+public class BindContext {
+
+    private Consumer consumer;
+    private Consumer lockedConsumer;
+    private Map<String, PoolQuantity> poolQuantities;
+    private Map<String, Entitlement> entitlementMap;
+    private Map<String, EntitlementCertificate> certs;
+    private Map<String, Integer> quantities;
+
+    public boolean isQuantityRequested() {
```
barnabycourt: method in the middle of variable declarations?

server/src/main/java/org/candlepin/bind/BindContext.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public Consumer getConsumer() {
+        return consumer;
+    }
+
+    public Map<String, PoolQuantity> getPoolQuantities() {
+        if (poolQuantities == null) {
+            poolQuantities = new HashMap<String, PoolQuantity>();
+            List<Pool> pools = poolCurator.listAllByIds(quantities.keySet()).list();
```
barnabycourt: calling poolCurator.listAllByIds() could run into performance issues if folks don't know that getPoolQuantities() forces db calls.

server/src/main/java/org/candlepin/bind/HandleCertificates.java
```diff
@@ -0,0 +1,91 @@
...
+            Pool pool = poolQuantity.getPool();
+            products.put(pool.getId(), pool.getProduct());
+        }
+
+        certs = ecGenerator.generateEntitlementCertificates(context.getConsumer(),
+            products,
+            poolQuantities,
+            context.getEntitlementMap());
+
+        chain.preProcess(context);
```
barnabycourt: Is each command responsible for calling the next item in the chain or is the BindChain responsible for that? I see both patterns in the commands & BindChain functionality.

server/src/main/java/org/candlepin/bind/HandleCertificates.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public void execute(BindContext context, BindChain chain) {
+
+        if (isApplicable) {
```
barnabycourt: Could checking Applicability be a function of the BindChain?

server/src/main/java/org/candlepin/bind/ToDo.java
```diff
@@ -0,0 +1,53 @@
...
+import org.candlepin.model.PoolQuantity;
+
+import com.google.inject.Inject;
+
+import java.util.Map;
+
+/**
+ * Created by vrjain on 5/5/17.
+ */
+public class ToDo implements BindOperation {
```
barnabycourt: What is the purpose of this class?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1630,102 +1637,35 @@ public Entitlement adjustEntitlementQuantity(Consumer consumer,
...
-        }
-
-        Map<String, PoolQuantity> poolQuantities = new HashMap<String, PoolQuantity>();
-        boolean quantityFound = false;
-        for (Pool pool : pools) {
-            Integer quantity = poolQuantityMap.get(pool.getId());
-            if (quantity > 0) {
-                quantityFound = true;
-            }
+        BindContext context = new BindContext(poolCurator, consumerCurator, i18n);
```
barnabycourt: The changes here are all happening within a transaction context. Unless \@Transactional is removed from the createEnittlements method I don't believe there will be a change in the lock time.

server/src/main/java/org/candlepin/bind/BindChain.java
```diff
@@ -0,0 +1,50 @@
...
+        if (executeIndex < operations.size()) {
+            operations.get(executeIndex).execute(context, this);
+        }
+    }
+
+    public void addOperation(BindOperation operation) {
+        operations.add(operation);
+    }
+
+    // TODO Implement other assorted list methods if desired.
```
awood: You can delete this line 🤣

server/src/main/java/org/candlepin/bind/BindContextFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ * in this software or its documentation.
+ */
+package org.candlepin.bind;
+
+import org.candlepin.model.Consumer;
+
+import java.util.Map;
+
+/**
+ * Created by vrjain on 5/12/17.
```
awood: We should drop these comments and/or replace them with actual JavaDoc.

server/src/main/java/org/candlepin/bind/HandleCertificatesOp.java
```diff
@@ -0,0 +1,94 @@
...
+import org.candlepin.model.PoolQuantity;
+import org.candlepin.model.Product;
+
+import com.google.inject.Inject;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Created by vrjain on 5/4/17.
```
awood: JavaDoc here please

server/src/main/java/org/candlepin/bind/HandleCertificatesOp.java
```diff
@@ -0,0 +1,94 @@
...
+            context.getEntitlementMap(),
+            false);
+
+        chain.preProcess(context);
+    }
+
+    @Override
+    public void execute(BindContext context, BindChain chain) {
+
+        if (isApplicable) {
```
awood: do we really need isApplicable?  It looked to me like the only place that can get set to false is line 56, so why not just set the predicate in this if statement to `context.getConsumer().isShare()`.

server/src/main/java/org/candlepin/model/Entitlement.java
```diff
@@ -73,8 +71,8 @@
     private static final long serialVersionUID = 1L;
 
     @Id
-    @GeneratedValue(generator = "system-uuid")
-    @GenericGenerator(name = "system-uuid", strategy = "uuid")
+    //@GeneratedValue(generator = "system-uuid")
+    //@GenericGenerator(name = "system-uuid", strategy = "uuid")
```
awood: Do we not need these anymore?

server/src/main/java/org/candlepin/bind/BindContext.java
vritant: @barnabycourt apologies, the PR was not ready for review then, its polished now.

vritant: I wanted to publish my progress daily cause this was a multiple day effort.

server/src/main/java/org/candlepin/bind/BindContext.java
vritant: @barnabycourt getPoolQuantities forces db calls just once. after the first invocation, this method is a pure getter which simply responds with he already enriched poolQuantities. this pattern is followed through out the BindContext, so no command is dependent on other commands having to initialize collections.

server/src/main/java/org/candlepin/bind/HandleCertificates.java
vritant: @barnabycourt each command is responsible for calling the next item in the chain. the way it invokes the next item in the chain is by invoking the respective method on the chain itself, which is aware of the next command in line. I believe this design was inspired by the servlet filter interfaces.

server/src/main/java/org/candlepin/bind/HandleCertificates.java
vritant: @barnabycourt perhaps this decision will play itself out once I implement the post bind bonus pool commands, where applicability will be more relevant ( currently implemented commands are almost always applicable )

server/src/main/java/org/candlepin/bind/ToDo.java
vritant: @barnabycourt that was work in progress, to indicate the operations that will be tackled in the next PR.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
vritant: I will run some tests to measure performance.

server/src/main/java/org/candlepin/bind/BindChain.java
vritant: done

server/src/main/java/org/candlepin/bind/BindContextFactory.java
vritant: done

server/src/main/java/org/candlepin/bind/HandleCertificatesOp.java
vritant: done

server/src/main/java/org/candlepin/bind/HandleCertificatesOp.java
vritant: removed

server/src/main/java/org/candlepin/model/Entitlement.java
vritant: done

vritant:
@awood completed the requested changes, and also added some more comments.

server/src/main/java/org/candlepin/bind/HandleCertificatesOp.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This bind operation is responsible to create certificates and regenerate
+ * relevant certificates during a bind.
+ *
+ * TODO: either fetch ents to regenerate in preProcess and then mark them dirty in execute,
+ * otherwise, write a query to directly update ents without fetching them, all in execute method.
```
Ceiu: Has this been addressed?

server/src/main/java/org/candlepin/bind/UpdatePoolsFromStackOperation.java
```diff
@@ -0,0 +1,70 @@
...
+        chain.preProcess(context);
+    }
+
+    @Override
+    public void execute(BindContext context, BindChain chain) {
+        if  (poolUpdates.size() > 0 || poolsToDelete.size() > 0) {
+            // for (PoolUpdate update: poolUpdates) {
+                // update.apply();
+            // }
+            if (poolsToDelete.size() > 0) {
```
Ceiu: This whole block is a bit odd. The update bits are commented out, which makes the initial checks in the branch somewhat extraneous. Even if that update loop is intended to be there, there's no reason to do the count check for both only to do the count checks again per-operation.

server/src/main/java/org/candlepin/model/Entitlement.java
```diff
@@ -150,12 +146,20 @@ public void setId(String id) {
...
      */
     public Entitlement(Pool poolIn, Consumer consumerIn, Integer quantityIn) {
+        this(consumerIn, quantityIn);
         pool = poolIn;
+        updatedOnStart = poolIn.getStartDate().after(new Date());
+    }
+
+    /**
+     * ctor
+     * @param consumerIn consumer associated with the entitlement
```
Ceiu: Docblock needs improvement. Annotation missing for the quantityIn param.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1634,116 +1660,40 @@ public Entitlement adjustEntitlementQuantity(Consumer consumer, Entitlement enti
...
-            }
-
-            poolQuantities.put(pool.getId(), new PoolQuantity(pool, quantity));
-            poolQuantityMap.remove(pool.getId());
+        // fetch and pre-process entities before locks are placed.
+        chain.preProcess(context);
+        if (context.getException() == null) {
+            context.lockPools();
+            // post lock operations
+            chain.execute(context);
```
Ceiu: This looks like it'll only run whatever operation happens to be in slot 1. since each call to .preprocess and .execute only grabs the current index and executes it iff the execute operation count hasn't exceeded the operation count.

At the time of writing, this seems counterintuitive. I'll update this comment if I see something else that clears up why this currently works the way it does.

server/src/main/java/org/candlepin/bind/BindChain.java
```diff
@@ -0,0 +1,51 @@
...
+    private int preProcessIndex = -1;
+    private int executeIndex = -1;
+
+    public BindChain() {
+        operations = new ArrayList<BindOperation>();
+    }
+
+    public void preProcess(BindContext context) {
+        preProcessIndex++;
+        if (preProcessIndex < operations.size()) {
```
Ceiu: Why did we opt for a one-time-use chain that requires manual stepping on the part of the client? Also, this state is not at all synchronized with the operation count, which forces users to fully understand this class to use it properly.

server/src/main/java/org/candlepin/bind/HandleCertificatesOp.java
vritant: @Ceiu , thats an upcoming PR. I will re base #1542 and re-work on it after this is merged.

server/src/main/java/org/candlepin/bind/UpdatePoolsFromStackOperation.java
vritant: @Ceiu that operation is not in use in this PR. this and PostBindBonusPoolOp will be re-worked in an upcoming PR.

server/src/main/java/org/candlepin/model/Entitlement.java
vritant: done

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
vritant: each operation is responsible to invoke the chain for pre-process or execute method.

server/src/main/java/org/candlepin/bind/BindChain.java
vritant: @awood and @Ceiu lets talk online.

server/src/main/java/org/candlepin/bind/HandleCertificatesOp.java
Ceiu: ack

server/src/main/java/org/candlepin/bind/UpdatePoolsFromStackOperation.java
Ceiu: ack

server/src/main/java/org/candlepin/bind/BindChain.java
```diff
@@ -0,0 +1,66 @@
...
+public class BindChain {
+    private List<BindOperation> operations;
+    private static Logger log = LoggerFactory.getLogger(BindChain.class);
+    public BindChain() {
+        operations = new ArrayList<BindOperation>();
+    }
+
+    public void preProcess(BindContext context) {
+        for (BindOperation operation : operations) {
+            log.debug("Starting preprocess of " + operation.getClass().getSimpleName());
```
vritant: ack, sorry im usually good about that skipped my mind.

server/src/main/java/org/candlepin/bind/BindChain.java
```diff
@@ -0,0 +1,66 @@
...
+            if (!operation.preProcess(context)) {
+                log.error("Skipped chain in preprocess of operation " + operation.getClass().getSimpleName());
```
vritant: in the current code its an error.  fixed.

server/src/main/java/org/candlepin/bind/HandleEntitlementsOp.java
```diff
@@ -0,0 +1,92 @@
...
+        this.poolCurator = poolCurator;
+    }
+
+    /**
+     * generates entitlement objects without associating them with pools or consumers.
+     * @param context
+     */
+    @Override
+    public boolean preProcess(BindContext context) {
+        context.getEntitlementMap();
```
vritant: added a comment

server/src/main/java/org/candlepin/bind/HandleEntitlementsOp.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    @Override
+    public boolean execute(BindContext context) {
+        Consumer consumer = context.getLockedConsumer();
+        Map<String, Entitlement> entitlementMap = context.getEntitlementMap();
+        Map<String, PoolQuantity> lockedPools = context.getPoolQuantities();
+        List<Pool> poolsToSave = new LinkedList<Pool>();
+        for (Entry<String, PoolQuantity> entry: context.getPoolQuantities().entrySet()) {
+            Entitlement ent = entitlementMap.get(entry.getKey());
+            Pool pool = lockedPools.get(entry.getKey()).getPool();
```
vritant: but why would I do that?

server/src/main/java/org/candlepin/bind/PreEntitlementRulesCheck.java
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * The java script portion of the rules can be run before the locks have been placed.
+     * if there is a failure, we stop the chain here.
+     * @param context
+     */
+    @Override
+    public boolean preProcess(BindContext context) {
+
+        // fetch pools
+        context.getPoolQuantities();
```
vritant: removed

server/src/main/java/org/candlepin/bind/UpdatePoolsFromStackOperation.java
```diff
@@ -0,0 +1,70 @@
...
+        this.poolRules = poolRules;
+        this.poolManager = poolManager;
+        this.poolUpdates = new ArrayList<PoolUpdate>();
+        this.poolsToDelete = new ArrayList<Pool>();
+    }
+
+    @Override
+    public boolean preProcess(BindContext context) {
+        // Do stuff here
+        return false;
```
vritant: ok

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -571,14 +578,26 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
         List<String> eids = new LinkedList<String>();
 
         if (entitlements != null && entitlements.iterator().hasNext()) {
+
+            String eidClause = "";
+            List<List<Entitlement>> splitEnts = new ArrayList<List<Entitlement>>();
+            Iterable<List<Entitlement>> blocks = Iterables.partition(entitlements, getInBlockSize());
+            int i = 0;
+            for (List<Entitlement> block : blocks) {
+                eidClause += " eOut NOT IN (:not_in_" + i++ + ") AND ";
```
vritant: done

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1593,105 +1620,30 @@ public Entitlement adjustEntitlementQuantity(Consumer consumer, Entitlement enti
...
+            chain.lock(context);
+            // post lock operations
+            chain.execute(context);
         }
 
-        if (!poolQuantityMap.isEmpty()) {
-            throw new IllegalArgumentException(i18n.tr("Subscription pool(s) {0} do not exist.",
-                poolQuantityMap.keySet()));
+        //check again cause either pre process or execute can throw an exception.
+        if (context.getException() != null) {
```
vritant: spoke off line about this, ignoring comment

server/src/main/java/org/candlepin/bind/BindContext.java
Ceiu: If there's any concern that the list returned could be huge, it might be worth considering not calling .list() straight away. None of the operations on the pools hit the DB again, so it's safe to use the cursor iteration here.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1597,105 +1624,30 @@ public Entitlement adjustEntitlementQuantity(Consumer consumer, Entitlement enti
...
-        }
-
-        log.debug("Done locking pools");
+        BindChain chain = new BindChain();
+        BindContext context = bindContextFactory.create(consumer, poolQuantityMap);
+        chain.addOperation(rulesCheckOpFactory.create(caller));
+        chain.addOperation(handleEntitlementsOp);
+        chain.addOperation(postBindBonusPoolsOp);
+        chain.addOperation(handleCertificatesOp);
+        chain.addOperation(complianceOp);
```
mstead: I don't really like the fact that all the operations bleed into the pool manager. The chain is really the only thing that needs to know about the operations. They shouldn't be getting injected into the pool manager IMO.

Perhaps it would be better to inject the chain in the pool manager and do operation injections inside of the chain itself to isolate the operations to the chain.

Then we can simply:
```java
BindContext context = bindContextFactory.create(consumer, poolQuantityMap);
chain.preProcess(context);
...
```

mstead: The BindChain is not really a 'chain'. It is basically a list of operations that get iterated over -- a command so to speak. IMO the chain should be responsible for dictating the pre/execute/post flow, not the pool manager itself. 

Taking this a little further, does the context really need to be exposed to the pool manager?

For consideration, we could execute the command and have it return the result, or throw the exception itself that will bubble up. All this would simply become the following in the pool manager, keeping everything encapsulated in the command/chain class.
```java
// Execute the command/chain with the provided args.
// A context related to this execution is created within the command/chain
// and never gets exposed outside of a single execution.
List<Entitlement> ents = bindcommand.execute(consumer, poolQuantityMap);
```

vritant: no objections, but wanted to make sure @awood is okay with this.

vritant: no objections, but again this is very different from what @awood  designed, so I will create a PR with this specific changes and we can request him to merge this into this PR first.

mstead: Yep, no worries. I'm interested in @awood comments.

I don't think that this change strays from the design. The change is simply hiding details of execution and the execution context from the caller.

A few other comments... If the overall intent is to be able to create a generic chain of varying operations, and then to be able to reuse 'the concept' in other places, then we shouldn't be calling it a "BindChain" and typing the objects it uses as such. As it is right now, it is defining the exact steps required in the bind process.

vritant: @mstead  looking at what I ended up with, I just added a commit to this PR. 
@awood  and @mstead  please review the latest commit .

@mstead  ,  I actually spent the last hour trying to write a generic chain before I saw your comment, but I think that is not a good idea right now, this is what I found:

We can create a generic Operation interface, and OpChain abstract class in an "operations" package, and then implement them as BindChain, while each operation could extend Operation, but the problem lies with BindContext.
both preprocess and execute in the bind chain need the actual BindContext, and not a generic context.
This problem can be solved, but I think we should wait till we have atleast one different type of chain. only then we can pick one of the two following approaches to solve the problem:
1. change BindContext to hold all entities in a map so it can be shared among all kinds of chains.
2. based on what entities the other chain would need,  we could perhaps use the BindContext as is.

I am hesitant to just say that no matter what chains we have, we will use a generic context, cause then we lose encapsulation imo. so I thought it best to simply defer this for future when things will get clearer once we know what our new chain will look like.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1624,30 +1602,13 @@ public Entitlement adjustEntitlementQuantity(Consumer consumer, Entitlement enti
...
-            chain.execute(context);
-        }
-
-        //check again cause either pre process or execute can throw an exception.
-        if (context.getException() != null) {
-            throw context.getException();
-        }
+        Map<String, Entitlement> result = bindChainFactory.
+            create(consumer, poolQuantityMap, caller).
+            run();
```
mstead: If you are just going to call result.values() below, any reason to return Map<String, Entitlement> instead of just a list?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -168,12 +156,7 @@ public CandlepinPoolManager(
         OwnerManager ownerManager,
         PinsetterKernel pinsetterKernel,
         I18n i18n,
-        BindContextFactory bindContextFactory,
-        PreEntitlementRulesCheckOpFactory rulesCheckOpFactory,
-        HandleEntitlementsOp handleEntitlementsOp,
-        PostBindBonusPoolsOp postBindBonusPoolsOp,
-        HandleCertificatesOp handleCertificatesOp,
-        ComplianceOp complianceOp) {
+        BindChainFactory bindChainFactory) {
```
mstead: I like that this factory is now injected instead of a BindChain instance. 👍

server/src/main/java/org/candlepin/bind/BindChain.java
```diff
@@ -14,24 +14,51 @@
...
+        context = bindContextFactory.create(consumer, poolQuantityMap);
+        operations.add(rulesCheckOpFactory.create(caller));
+        operations.add(handleEntitlementsOp);
+        operations.add(postBindBonusPoolsOp);
+        operations.add(handleCertificatesOp);
+        operations.add(complianceOp);
     }
 
-    public void preProcess(BindContext context) {
+    public boolean preProcess(BindContext context) {
```
mstead: method can be protected now (private if not isolated in tests).

server/src/main/java/org/candlepin/bind/BindChain.java
```diff
@@ -51,21 +79,27 @@ public void lock(BindContext context) {
         log.debug("Successfully achieved locks");
     }
 
-    public void execute(BindContext context) {
+    public boolean execute(BindContext context) {
```
mstead: method can be protected now (private if not isolated in tests).

server/src/main/java/org/candlepin/bind/BindChain.java
```diff
@@ -40,9 +67,10 @@ public void preProcess(BindContext context) {
...
                 log.error("Skipped chain in preprocess of operation {}",
                     operation.getClass().getSimpleName());
-                break;
+                return false;
             }
         }
+        return true;
     }
 
     public void lock(BindContext context) {
```
mstead: method can be protected now (private if not isolated in tests).

server/src/main/java/org/candlepin/bind/BindContext.java
vritant: Today, im not worried about this. in future as clients start writing code to actually use the batch bind API, there can be a concern.

mstead:
@vritant Please add a more detailed commit message about what this PR is all about and what changes were made.

vritant:

mstead:
[Implement lock less monster - minimal set](https://trello.com/c/1txlUK4h/359-implement-lock-less-monster-minimal-set)

PR #1564, opened 2017-05-05
Configure in clauses based on the DB we're using.

awood:
Makes unit tests faster too!

The values can be overridden in the Candlepin configuration file, but that's currently undocumented.

### DIFF
--- a/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
+++ b/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
@@ -28,6 +28,7 @@ import org.candlepin.common.config.EncryptedConfiguration;
 import org.candlepin.common.config.MapConfiguration;
 import org.candlepin.common.logging.LoggingConfigurator;
 import org.candlepin.config.ConfigProperties;
+import org.candlepin.config.DatabaseConfigFactory;
 import org.candlepin.controller.SuspendModeTransitioner;
 import org.candlepin.logging.LoggerContextListener;
 import org.candlepin.pinsetter.core.PinsetterContextListener;
@@ -41,7 +42,9 @@ import com.google.inject.Module;
 import com.google.inject.Stage;
 import com.google.inject.util.Modules;
 
+import org.apache.commons.lang.StringUtils;
 import org.hibernate.cfg.beanvalidation.BeanValidationEventListener;
+import org.hibernate.dialect.PostgreSQLDialect;
 import org.hibernate.event.service.spi.EventListenerRegistry;
 import org.hibernate.event.spi.EventType;
 import org.hibernate.internal.SessionFactoryImpl;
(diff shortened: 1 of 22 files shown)


awood:
[speeden up slow unit tests in PoolCuratorTest](https://trello.com/c/KKxkZ12p/371-speeden-up-slow-unit-tests-in-poolcuratortest)

server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -224,9 +226,27 @@ protected Configuration readConfiguration(ServletContext context)
...
+
+        // merge the defaults with the system configuration. PARAMETER ORDER MATTERS.
+        // systemConfig must be FIRST to override subsequent configs.  This set up allows
+        // the systemConfig to override databaseConfig if necessary, but that's not really something
+        // users should be doing unbidden so it is undocumented.
+        return EncryptedConfiguration.merge(systemConfig, databaseConfig, defaults);
+    }
+
+    private DatabaseConfigFactory.SupportedDatabase determinDatabaseConfiguration(String dialect) {
+        if (dialect.matches("(?i)" + DatabaseConfigFactory.SupportedDatabase.MYSQL.getLabel())) {
```
Ceiu: What is the advantage of using a regex here instead of .equalsIgnoreCase? Also, I was under the impression that when you do inline case-insensitivity, it only applies to the group in which the modifier appears. I could be wrong, but this seems like it's just doing a normal comparison.

server/src/main/java/org/candlepin/model/CPRestrictions.java
```diff
@@ -24,21 +27,25 @@
...
         public CPLikeExpression(String property, String value, Character escape, boolean ignoreCase) {
             super(property, value, escape, ignoreCase);
         }
     }
 
-    protected CPRestrictions() {
-        throw new UnsupportedOperationException("CPRestriction should not be instantiated");
+    @Inject
+    public CPRestrictions(Configuration config) {
+        this.config = config;
```
Ceiu: Is it possible to keep this a static factory? Personally, I would prefer to pass the block size in as a parameter and keep the CPRestrictions factory object consistent with Hibernate's Restrictions factory. Also simplifies some constructors and avoids some extra mocking and setup in the classes that use them.

server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
awood: The [dialect value](https://www.mkyong.com/hibernate/hibernate-dialect-collection/) is a class name like "org.hibernate.dialect.PostgreSQLDialect" (which is deprecated) or "org.hibernate.dialect.PostgreSQL82Dialect" which is the recommended value these days.  We happen to use the deprecated value.  Likewise, MySQL has a few different dialects based on the engine used.

I think you are right though.  I will replace it with StringUtils.containsIgnoreCase which is more straightforward.

server/src/main/java/org/candlepin/model/CPRestrictions.java
awood: I couldn't get it to work.  CPRestrictions now requires a reference to the Configuration object.  Guice does allow [static injection](https://github.com/google/guice/wiki/Injections#static-injections) but using it is not recommended and I thought it would making testing more difficult.

I will try again and see if I get it working.

Ceiu: I meant manually. So our calls would look like CPRestrictions.in(param, block_size), and the others would be unaffected.

Ceiu:
ACK

Ceiu:
Checkstyle failures

```
(10:13:07) crog@crog-ws-1:~/devel/candlepin-origin/server (awood/configurable-in-clause) $ buildr clean checkstyle
(in /home/crog/devel/candlepin-origin, development)
Cleaning candlepin:server
Running Checkstyle on candlepin:server
Checking /home/crog/devel/candlepin-origin/server/^src/main/java/
[checkstyle] Running Checkstyle 7.0 on 505 files
[checkstyle] [ERROR] /home/crog/devel/candlepin-origin/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java:245: 'dialect' have incorrect indentation level 16, expected level should be 12. [Indentation]
[checkstyle] [ERROR] /home/crog/devel/candlepin-origin/server/src/main/java/org/candlepin/model/CPRestrictions.java:37:1: Utility classes should not have a public or default constructor. [HideUtilityClassConstructor]
Checking /home/crog/devel/candlepin-origin/server/^src/test/java
[checkstyle] Running Checkstyle 7.0 on 259 files
Checkstyle found errors in candlepin:server
Buildr aborted!
RuntimeError : Checkstyle failed
```

awood:
> Checkstyle failures

enhanced-5324-1433453156-2

Ceiu:
ACK

Tests were verified manually.

PR #1566, opened 2017-05-10
cp_test now checks for Qpid when attempting to configure it

Ceiu:
- When the -q flag is specified to setup Qpid, cp_test will now
  verify the qpid directory exists in the active project before
  attempting to run the setup tasks

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -232,7 +232,8 @@ if [ "$LINTER" == "1" ]; then
 fi
 
 # TODO: keep track of return code?
-cd "$CP_HOME/$PROJECT"
+PROJECT_DIR="$CP_HOME/$PROJECT"
+cd $PROJECT_DIR
 
 if [ "$UNITTEST" -gt 0 ]; then
     echo "Running unit tests $UNITTEST time(s)"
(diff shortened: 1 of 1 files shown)


awood:
LGTM, but jekins-rspec-qpid will always fail in the case that qpid isn't running.  Any thoughts about how to handle that?

Ceiu:
For the test case, it looks like part of the qpid setup is restarting the service. The one hiccup looks to be what happens if (what I assume is) SELinux gets in the way. There's a bit more work and cleanup that could be done in the actual configure-qpid.sh file to detect when we're running in our containers and do the additional/specialized setup as necessary.

Ceiu:
retest this please

docker/candlepin-base/cp-test
```diff
@@ -256,16 +257,23 @@ if [ "$UNITTEST" -gt 0 ]; then
...
-    ./configure-qpid.sh -c
-    # Reconfigure
-    ./configure-qpid.sh
-    popd 
-fi
+    if [ -d "$PROJECT_DIR/bin/qpid" ]; then
+        echo "Setting up Qpid"
+
+        pushd .
+        cd "$PROJECT_DIR/bin/qpid"
```
awood: This can actually be shortened to

```
pushd "$PROJECT_DIR/bin/qpid"
```

The shell will keep the current directory on the stack and you'll return to it when you run `popd`

Ceiu: Fixed.

PR #1570, opened 2017-05-12
1450386: entitlement of quantity 0 gives unlimited entitlements

vritant:


vritant:
[1450386: consuming an entitlement of quantity 0 gives unlimited entitlements](https://trello.com/c/XrZiZGYp/386-1450386-consuming-an-entitlement-of-quantity-0-gives-unlimited-entitlements)

server/spec/entitlement_spec.rb
```diff
@@ -83,6 +83,14 @@
     @system.list_entitlements.first.quantity.should == 10
   end
 
+  it 'should allow consumption of quantity zero' do
+    pool = find_pool_for_product @virt
+    pool['consumed'].should eq(0)
+    ent = @system.consume_pool(pool.id, {:quantity => 0})
```
Ceiu: Should this be an error case? I feel like zero is a special quantity, but maybe not in the "consumed" case. In any event, if this is a situation where we don't allow actually consuming less than 1, we should probably error out here rather than correcting it. Is this something that happens today?

vritant: @Ceiu bindng by zero is allowed today. I think it is meant to represent bind by 1.
I dont know why this is allowed, or if it should be blocked. I thought it is safe to keep the functionality since we explicitly allow it in [this](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/model/Entitlement.java#L152) code that has always been there.

Ceiu: Ah, alright. Thanks.

awood: How strange.  I have no idea why that would be.

Ceiu:
ACKing with the note that this is something that we should review more closely in the future. Correcting what feels like an invalid value seems wrong and it adds unnecessary complexity (and surprises, as we're finding) to our code when we should just be rejecting it outright.

I'll add a task to the backlog for this.

PR #1573, opened 2017-05-15
Establish revocation order for entitlments

wottop:
When the master pool quantity is reduced, the unused share pool counts
 are first reduced. If that is not adequate, then entitlements are revoked
 based on sort order across both master and shared entitlements.

Also removed config entry for revoke order. Now hard coded to LIFO order.


server/src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java
```diff
@@ -234,7 +234,7 @@ private void finishValidation(ValidationResult result, Pool pool, Integer quanti
     public ValidationResult update(Consumer consumer, Entitlement entitlement, Integer change) {
 
         ValidationResult result = new ValidationResult();
-        if (!consumer.getType().isManifest()) {
+        if (!consumer.isManifestDistributor() && !consumer.isShare()) {
```
vritant: oops! my bad, thanks!

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -650,16 +649,18 @@ void removeAndDeletePoolsOnOtherOwners(List<Pool> existingPools, Pool pool) {
             this.poolCurator.flush();
 
             // quantity has changed. delete any excess entitlements from pool
+            // the quantity has not yet been expressed on the pool itself
             if (updatedPool.getQuantityChanged()) {
-                List<Pool> existingPools = Arrays.asList(existingPool);
-                this.deleteExcessEntitlements(existingPools);
+                RevocationPlan revPlan = new RevocationPlan(poolCurator, config);
+                revPlan.setPool(existingPool);
```
vritant: why not set pools in the constructor ?

server/src/main/java/org/candlepin/controller/RevocationPlan.java
```diff
@@ -0,0 +1,167 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Created by wpoteat on 5/11/17.
+ */
+public class RevocationPlan {
```
vritant: RevocationPlan in its current implementation only works for one pool. we should batch the entire operation so that there are fewer queries made.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1628,4 +1628,15 @@ public void calculateSharedForOwnerPools(Owner owner) {
         q.setParameter("owner", owner);
         q.executeUpdate();
     }
+
+    @Transactional
+    public List<Pool> listSharedPoolsFromMasterPool(Pool pool) {
```
vritant: Please avoid using the work "master pool" in this function name, as the pool that was shared could have been a non-master ( non-normal ) bonus pool as well.. 
my recommendation : listSharedPoolsOf(Pool pool) {

server/src/main/java/org/candlepin/controller/RevocationPlan.java
```diff
@@ -0,0 +1,167 @@
...
+        for (Pool sPool : sharedPools.keySet()) {
+            if (over > 0) {
+                long excessCount = sPool.getQuantity() - sPool.getConsumed();
+                long newQuantity = 0;
+                if (excessCount > 0) {
+                    Entitlement shareEnt = sPool.getSourceEntitlement();
+                    if (over >= excessCount) {
+                        newQuantity = sPool.getQuantity() - excessCount;
+                        over = over - excessCount;
+                        newConsumed -= excessCount;
```
vritant: resolved.

server/src/main/java/org/candlepin/controller/RevocationOp.java
```diff
@@ -0,0 +1,171 @@
...
+    private List<Pool> pools;
+    private Set<Entitlement> entitlementsToRevoke;
+    private Map<Entitlement, Long> entitlementsToAdjust;
+    private Map<Pool, List<Pool>> sharedPools;
+    private PoolCurator poolCurator;
+    private long newConsumed;
+
+    public RevocationOp(PoolCurator poolCurator, List<Pool> pools) {
+        entitlementsToRevoke = new HashSet<Entitlement>();
+        entitlementsToAdjust = new HashMap<Entitlement, Long>();
```
vritant: if I understood right, there are only shared ents in this list. would it make sense to change its name to sharedEntitlementsToAdjust?

vritant: I think that would be useful in this case, since this is tricky logic, and this clarifies the intent of this data structure.

server/src/main/java/org/candlepin/controller/RevocationOp.java
```diff
@@ -0,0 +1,171 @@
...
+    private void determineExcessEntitlements(Pool pool) {
+        List<Pool> pools = new ArrayList<Pool>();
+        pools.add(pool);
+        pools.addAll(sharedPools.get(pool));
+        List<Entitlement> freeEntitlements = this.poolCurator.retrieveOrderedEntitlementsOf(pools);
+
+        long existing = pool.getQuantity();
+        for (Entitlement ent : freeEntitlements) {
+            if (newConsumed > existing) {
+                if (!ent.getConsumer().isShare()) {
```
vritant: apologies if I am missing some detail, since we are skipping revokes of shared entitlements here, is that covered else where?

server/src/main/java/org/candlepin/controller/RevocationOp.java
```diff
@@ -0,0 +1,171 @@
...
+                    if (ent.getPool().getType().equals(Pool.PoolType.SHARE_DERIVED)) {
+                        Entitlement source = ent.getPool().getSourceEntitlement();
+                        if (entitlementsToAdjust.get(source) == null) {
```
vritant: a comment here like following would be nice:
" We might have already adjusted the quantity of the source entitlement due to shared pool reduction above"

server/src/main/java/org/candlepin/controller/RevocationOp.java
wottop: I made it 'share' entitlements. The 'shared' entitlements are the ones in OrgB that are consumed by systems.

server/src/main/java/org/candlepin/controller/RevocationOp.java
wottop: These entitlements are the source entitlements for the shared pools. If the adjustment step sets them down to 0 they will get cleaned up. Remember that there is no cert to revoke.

wottop:
[Multi-Org: Revocations](https://trello.com/c/GfYFQp7A/314-multi-org-revocations)

wottop:
jenkins: retest this please

wottop:
jenkins: retest this please

wottop:
Note: I'll squash once we get to an endpoint.

server/src/main/java/org/candlepin/controller/RevocationOp.java
vritant: gotcha

vritant: in that case, before we revoke ents, we should go through the ents to adjust and if any of them are to be set to 0, we should revoke it instead. that we dont leave behind ents with quantity 0.

vritant:
@wottop looks good! please squash.

wottop:
jenkins: retest this please

PR #1575, opened 2017-05-15
1451024: Allow tomcat to read /etc/candlepin (0.9.54)

kahowell:


mstead:
retest this please

mstead:
I've built the RPMs and provided them to the bug reporter. They do not seem to address the issues, even with the changes from the policy guys in the bug mentioned in comment 1.

I've almost got a test system up and running. Just having a little trouble with getting the new selinux-policy rpm deps squared away.

mstead:
Also, I'm curious as to why jenkins is reporting so many changes for this build. Any idea?

kahowell:
a9f72a3..63fcf50:
 - added /var/{cache,lib,log}/candlepin rules
 - use macros for permissions
 - add missing rule

> Also, I'm curious as to why jenkins is reporting so many changes for this build. Any idea?

Because I think it is reporting what the git difference is between the previous run and this one. (Previous run was for a different branch I think).

kahowell:
63fcf50..a70324e: add restorecon for new rules

PR #1577, opened 2017-05-15
1451032: Allow tomcat to read /etc/candlepin

kahowell:


barnabycourt:
The test likely failed because candlepin needs access to /var/lib/candlepin in order to write the CRL file which defaults to /var/lib/candlepin/candlepin-crl.crl

mstead:
retest this please

kahowell:
>The test likely failed because candlepin needs access to /var/lib/candlepin in order to write the CRL file which defaults to /var/lib/candlepin/candlepin-crl.crl

@barnabycourt, doesn't look like we mess with the context for `/var/lib/candlepin`... and on my machine at least it ended up as `system_u:object_r:tomcat_var_lib_t:s0`, so I think we're probably okay on that front, but I did not think to check that previously, so thanks.

kahowell:
db8a12f..e4de42e:
 - added /var/{cache,lib,log}/candlepin rules
 - use macros for permissions
 - add missing rule

kahowell:
e4de42e..e0a8fcb: add restorecon for new rules

kahowell:
e0a8fcb..183ac13: Fix BZ# in commit message

PR #1581, opened 2017-05-18
Added content access list refreshing to the refresh process

Ceiu:
- The refresher will now refresh an owner's content access mode
  and content access mode list during the refresh process. This
  also takes over refreshing content access mode list duties from
  the importer as well.
- Fixed an issue where changing the content access mode or content
  access mode list during import would not cause certificates to be
  properly regenerated.
- Fixed an issue that would cause events to be generated for owner
  creation, even if the owner was not created due to an internal
  error.


server/src/main/resources/db/changelog/changelog-testing.xml
```diff
@@ -2028,7 +2028,7 @@
     </changeSet>
     <changeSet author="awood" id="1413225753032-290">
         <comment>
-             Add the default quartz lock columns.
+             Add the default quartz lock columns.
```
Ceiu: Fixed. We may have an issue with our buildr changeset task, as I don't edit those changelog-* files directly.

wottop:
ack

PR #1582, opened 2017-05-19
[0.9.51] Batch product fetching from adapters when populating hosted db.

mstead:
Product lookups via the adapter take a long time on hosted,
and seems to cause c3p0 to close the DB connection due to
an idle connection timeout. This patch performs batch product
adapter lookups in order to shorten the idle time while
updating the product records.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
+++ b/server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
@@ -14,6 +14,7 @@
  */
 package org.candlepin.pinsetter.tasks;
 
+import com.google.common.collect.Iterables;
 import org.candlepin.common.config.Configuration;
 import org.candlepin.config.ConfigProperties;
 import org.candlepin.model.ContentCurator;
@@ -36,6 +37,8 @@ import org.slf4j.LoggerFactory;
 
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Set;
 
 
@@ -77,6 +80,15 @@ public class PopulateHostedDBTask extends KingpinJob {
             return;
         }
 
+        int confBatchSize = config.getInt(ConfigProperties.POPULATE_HOSTED_DB_JOB_PROD_LOOKUP_BATCH_SIZE);
+        if (confBatchSize < 1) {
+            String error = String.format("Aborting populate DB task. Invalid configuration setting for {0}.",
+                ConfigProperties.POPULATE_HOSTED_DB_JOB_PROD_LOOKUP_BATCH_SIZE);
+            log.warn(error);
+            context.setResult(error);
+            return;
+        }
+
         JobDataMap map = context.getMergedJobDataMap();
         Boolean skipExisting = map.getBoolean(SKIP_EXISTING);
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
```diff
@@ -102,22 +105,36 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
...
+            // Batch fetch and process the product ids to avoid long request
+            // times from the adapter causing c3p0 to drop idle connections.
+            int adapterLookupCount = 0;
+            for (List<String> productIdBatch : Iterables.partition(productIds, confBatchSize)) {
+                // Nested loop to allow DB activity.
+                adapterLookupCount++;
+                log.info("Batch fetching products from adapter: {} of {} batches of {}.", adapterLookupCount,
+                    (int) Math.ceil((double)productIds.size() / (double) confBatchSize), confBatchSize);
+                for (Product product :
+                    this.productService.getProductsByIds(new HashSet<String>(productIdBatch))) {
```
Ceiu: We don't need to drop the block into a HashSet here, since the service takes a generic collection. Even though we know there are inefficiencies on their end, doing an extra copy on our end for a list that is likely already de-duplicated probably isn't worth the overhead (nor the extra line of code from having to wrap a long-line).

server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
```diff
@@ -102,22 +105,36 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
...
-                        this.contentCurator.createOrUpdate(pcontent.getContent());
-                        ++ccount;
+            log.info("Fetching and processing {} products from the adapter.", productIds.size());
+            // Batch fetch and process the product ids to avoid long request
+            // times from the adapter causing c3p0 to drop idle connections.
+            int adapterLookupCount = 0;
+            for (List<String> productIdBatch : Iterables.partition(productIds, confBatchSize)) {
+                // Nested loop to allow DB activity.
+                adapterLookupCount++;
+                log.info("Batch fetching products from adapter: {} of {} batches of {}.", adapterLookupCount,
```
Ceiu: ++adapterLookupCount here. Gotta save that extra instruction! :)

server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
```diff
@@ -102,22 +105,36 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
...
+                    (int) Math.ceil((double)productIds.size() / (double) confBatchSize), confBatchSize);
```
Ceiu: Space after (double)

server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
```diff
@@ -102,22 +105,36 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
             log.info("Importing data for all {} known products...", productIds.size());
         }
 
+        int confBatchSize = config.getInt(ConfigProperties.POPULATE_HOSTED_DB_JOB_PROD_LOOKUP_BATCH_SIZE);
```
Ceiu: Bad things happen if this happens to be 0 or lower.

Ceiu:
Manual testing went well. Verified that it properly handles blocking data as necessary, and successfully completes when the product count is lower than the block size. 

Invalid configuration results in an unclear IllegalArgumentException, though, as I speculated from my review.

server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
mstead: Addressed in new commit.

server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
mstead: Done.

server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
mstead: Done.

server/src/main/java/org/candlepin/pinsetter/tasks/PopulateHostedDBTask.java
mstead: Done.

mstead:
Addressed all your requested changes. Thanks for the review!

PR #1583, opened 2017-05-19
1440909: Choose host_specific/virt_only pools first 

cnsnyder:
This is an implementation of this design [1]. The work to prioritize the host specific and virt-only pools over others is done entirely by the rules.

[1]: https://docs.google.com/document/d/1fM19-FFxUAkt3T2kf_g1ZbjFUZBl5oA_LzlrN5WZtLs/edit#

### DIFF
--- a/server/spec/virt_spec.rb
+++ b/server/spec/virt_spec.rb
@@ -79,7 +79,7 @@ describe 'Standalone Virt-Limit Subscriptions', :type => :virt do
             :sockets=>4
         }
     })
-    # We'd like there two be three subs, two that require a specific host and one that provides both required products
+    # We'd like there to be three subs, two that require a specific host and one that provides both required products
     # in one. These first two are similar to VDC subscriptions, hence the name datacenter.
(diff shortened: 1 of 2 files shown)


cnsnyder:
Will squash once review is complete.

cnsnyder:
After investigating the rspec-qpid failures I've come to the conclusion that that particular job is passing options to cp-test that are not valid for CP 0.9.54. Namely the -q option causes the version of cp-test built into the currently in use docker images to attempt to run some qpid scripts in ../server/bin/qpid which do not exist in 0.9.54. It looks like the qpid_spec was added as part of the fail fast work (which as far as I can tell never hit 0.9.54).

Adding ignore jenkins failure tag.

cnsnyder:
[1440909 - [RFE] Prioritize Host provided pools first during auto attach (and guest migration) for virtual systems](https://trello.com/c/PveYsZFP/349-1440909-rfe-prioritize-host-provided-pools-first-during-auto-attach-and-guest-migration-for-virtual-systems)

cnsnyder:
@vritant I believe I have addressed all comments at this point. Please let me know if there is anything else left at this point.

PR #1584, opened 2017-05-23
1452330: Added refresh behavior from Candlepin 0.9.x (2.0)

Ceiu:
- Fixed a crash that could occur when creating a custom pool without
  a source subscription that would require creation of derived pools
- Derived or bonus pools are no longer created for custom pools
  without source subscription IDs
- Added a spec test for verifying that custom pools are not removed
  during an import or refresh, so long as they were not created with
  a source subscription
- Added a check to refresh to not remove pools that lack an upstream
  pool ID in standalone mode. This temporary fix should avoid an issue
  where pools created through the API were getting deleted during
  refresh.
- Derived/children pools are now properly cleaned up during batch
  deletion.

### DIFF
--- a/server/client/ruby/hostedtest_api.rb
+++ b/server/client/ruby/hostedtest_api.rb
@@ -6,7 +6,7 @@ module HostedTest
   def is_hostedtest_alive?
     if @@hostedtest_alive.nil?
       begin
-        @@hostedtest_alive = @cp.get('/hostedtest/subscriptions/is_alive', {}, 'json', true)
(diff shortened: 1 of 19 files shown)


Ceiu:
[[BZ] Importing/Refreshing manifest deletes custom product pools](https://trello.com/c/NBuOW8EC/396-bz-importing-refreshing-manifest-deletes-custom-product-pools)

Ceiu:
retest this please

PR #1587, opened 2017-05-30
Fix request scoping of the rules version cache provider.

barnabycourt:
The scoping was set on the incorrect object and this
was causing a new JsRunnerRequestCache to be generated
on every call of the JsRunnRequestCacheProvider.get() call
instead of one per request/job scope.


mstead:
@barnabycourt Please provide steps to reproduce.

barnabycourt:
To reproduce:

1. Turn on SQL logging (deploy -l) 
1. Run a command (bind by pool id is a good one) 
1. Note the output for the number of times that the `select max(this_.updated) as y0_ from cp_rules this_` SQL is executed.
1. Redeploy with the PR 
1. Run the same command again. You should see the above SQL only run once for a given REST call.

barnabycourt:
I'm closing this PR as the problem only exhibited in my environment.

PR #1588, opened 2017-05-31
1448897: Include default content access mode when showing org list

wottop:
If the default mode is not included in the mode list, then the list is
appended with the default mode. This is so that the user will see
the default mode as a selection.

### DIFF
--- a/server/src/main/java/org/candlepin/model/Owner.java
+++ b/server/src/main/java/org/candlepin/model/Owner.java
@@ -493,7 +493,13 @@ public class Owner extends AbstractHibernateObject implements Serializable,
      * @return String the value
      */
     public String getContentAccessModeList() {
-        return this.contentAccessModeList;
+        String[] modeList = contentAccessModeList == null ? new String[0] : contentAccessModeList.split(",");
+        String ammend = "";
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/candlepin/model/Owner.java
```diff
@@ -493,7 +493,13 @@ public void setContentAccessMode(String contentAccessMode) {
      * @return String the value
      */
     public String getContentAccessModeList() {
-        return this.contentAccessModeList;
+        String[] modeList = contentAccessModeList == null ? new String[0] : contentAccessModeList.split(",");
+        String ammend = "";
+        if (!ArrayUtils.contains(modeList, ContentAccessCertServiceAdapter.DEFAULT_CONTENT_ACCESS_MODE)) {
```
barnabycourt: Would this mean that the "entitlement" mode is always included? Even if a mode list has been set on the owner that did not include it?

wottop: A distinction without a difference. If setting it to empty =' entitlement', then 'entitlement is on the list.  Otherwise it feels like it is too confusing.

barnabycourt: @wottop sounds like there might need to be a larger discussion about whether empty should really be a valid option.

PR #1594, opened 2017-06-06
Awood/multi org hypervisors

awood:
Allow guests in one org to find hosts in another org.

### DIFF
--- a/server/spec/multi_org_spec.rb
+++ b/server/spec/multi_org_spec.rb
@@ -679,4 +679,74 @@ describe "Multi Org Shares" do
     expect(@cp.get_pool(pool['id']).quantity).to eq(1)
     expect(@user_client.list_pool_entitlements(pool['id']).size).to eq(1)
   end
+
+  it 'uses a host-restricted pool from the sharing org' do
+    skip("candlepin running in hosted mode") if is_hosted?
+    owner_key = random_string('orgA')
+    create_owner( owner_key)
+
+    virt_product = create_product(nil, nil, {
+      :owner => owner_key,
+      :attributes => {:virt_limit => 3}
+    })
+    pool = create_pool_and_subscription(owner_key, virt_product['id'], 10)
+
+    # create host consumer and consume the virt product
+    host = random_string('host')
+    orgAHost = @cp.register(
+      host,
+      :system,
+      nil,
+      {},
+      nil,
+      owner_key,
+      [],
+      [])
+
+    @cp.consume_pool(pool['id'], :quantity => 1, :uuid => orgAHost['uuid'])
+    guest_pool = @cp.list_owner_pools(owner_key).select do |p|
+      p['type'] == "ENTITLEMENT_DERIVED"
+    end.first
+
+    # share guest pool with Org B
+    orgB = random_string('orgB')
+    create_owner(orgB)
+    share_consumer = @cp.register(
+      random_string('orgBShare'),
+      :share,
+      nil,
+      {},
+      nil,
+      owner_key,
+      [],
+      [],
+      nil,
+      [],
+      nil,
+      [],
+      nil,
+      nil,
+      nil,
+      orgB
+    )
+    @cp.consume_pool(guest_pool['id'], :uuid => share_consumer['uuid'])
+
+    virt_uuid = random_string('system.uuid')
+    orgBconsumer = @cp.register(
+      random_string('orgBConsumer'),
+      :system,
+      nil,
+      {'virt.uuid' => virt_uuid, 'virt.is_guest' => 'true'},
+      'admin',
+      orgB,
+      [],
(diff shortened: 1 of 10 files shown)


awood:
bot, retest this please

server/src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java
```diff
@@ -215,9 +230,26 @@ else if (log.isDebugEnabled()) {
...
+     *
+     * @param consumer
+     * @param pools
+     * @return
+     */
+    private Consumer getHost(Consumer consumer, List<Pool> pools) {
+        if (!consumer.hasFact("virt.uuid")) {
+            return null;
+        }
+        List<Owner> potentialOwners = new ArrayList<Owner>(Arrays.asList(consumer.getOwner()));
```
vritant: should this be a set?

server/src/main/java/org/candlepin/policy/criteria/CriteriaRules.java
```diff
@@ -65,7 +65,7 @@ public CriteriaRules(Configuration config, ConsumerCurator consumerCurator) {
         // consumer UUID
         Consumer hostConsumer = null;
         if (consumer.getFact("virt.uuid") != null) {
-            hostConsumer = consumerCurator.getHost(consumer.getFact("virt.uuid"), consumer.getOwner());
+            hostConsumer = consumerCurator.getHost(consumer);
```
vritant: iirc, @mstead  or @awood  had investigated and found that the time difference between when the owner is specified and when not is negligible.
was this investigated against only one of  mysql/postgres or both?

server/spec/multi_org_spec.rb
```diff
@@ -679,4 +679,74 @@
     expect(@cp.get_pool(pool['id']).quantity).to eq(1)
     expect(@user_client.list_pool_entitlements(pool['id']).size).to eq(1)
   end
+
+  it 'uses a host-restricted pool from the sharing org' do
```
vritant: werent we discussing that we would add a spec test where the host in another org is not healed so the heal of the guest fails? is that already covered somehow?

server/src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java
awood: Yeah, probably.  I'll fix that.

awood: Ah.  Can't really make it a set because the owners get passed in to ConsumerCurator's `getHost` method as varargs.  I did that to maintain backwards compatibility with existing calls and to allow for a `getHost(consumer)` call that automatically searches across all orgs.

vritant: makes sense

server/spec/multi_org_spec.rb
awood: Just added one.  The scenario of a guest not causing a heal in a host is actually handled in two places now: CandlepinPoolManager.entitleByProductsForHosts just returns an empty collection if the guest and host are in different orgs **and** Entitler.bindByProducts won't even find the guest's host since we scope the host search to just the guest's organization.  The former check has been in the code 2013 and the latter check I added recently.  I say we leave them both just to keep our bases covered.

server/src/main/java/org/candlepin/policy/criteria/CriteriaRules.java
mstead: My query tests were against the stage database (mysql)

mstead: @vritant @awood It is worth noting that my query tests were done directly with SQL on the DB. We need to make sure that hibernate isn't performing a query any more complex than selecting the GuestId by GuestId.guest_id

My test query (see the design doc for how many records etc...
```sql
select guest_id, uuid, o.account from cp_consumer_guests as g join cp_consumer as h on h.id=g.consumer_id join cp_owner as o on h.owner_id=o.id where g.guest_id='4230b451-7bda-010d-be10-cacf20aff6a7';
```

awood: Hibernate generates the following:

```sql
    select
        this_.consumer_id as y0_ 
    from
        cp_consumer_guests this_ 
    inner join
        cp_consumer gconsumer1_ 
            on this_.consumer_id=gconsumer1_.id 
    where
        (
            this_.guest_id_lower=?
        ) 
    order by
        this_.updated desc limit ?
```

Looks pretty sane to me.

mstead: 👍

PR #1595, opened 2017-06-07
Undo excessive regen added by mistake in 731188f

vritant:
in 731188f, when I split addOrUpdateEntitlements , I mistakingly coded to regenerate *all* the entitlements of a consumer, instead of just the one that we are now updating. this commit undoes that mistake.


awood:
Just to make sure I understand, the problem here was that the entitlements were being regenerated once during entitlement update and then *again* during some later ConsumerResource operation?  Or just that they were being regenerated here and taking up excessive time during the lock instead of being marked dirty to be regenerated outside the log.

vritant:
@awood updated the description to answer the question

PR #1597, opened 2017-06-12
Vritant/list pool cursor

vritant:
use a cursor to fetch pools during a bind


Ceiu:
Couple comments before I address this:
- My initial comment was not a "this needs to change" comment, but was meant to bring attention to the choice of using a cursor instead of list. To reiterate: cursors are great when we're pulling large datasets such that the allocation and copying of all the data would be prohibitively slow. In the case of only a handful of objects, listing may actually be faster than using a cursor.
- Going to wait to review this until after the base PR is merged in, and this is rebased against master. Right now the diff looks like both sets of changes, and from the title, I'm assuming my interest is to be directed at only the cursor-related changes.

vritant:

PR #1600, opened 2017-06-15
1448897: Enforce default content access mode list

wottop:
If no list exists, then it is set to the default list and
 the mode is also set to the default value ['entitlement']
Ensure that the org's content access mode is defined when the
 org is created or updated
Update the mode list and mode columns to not null and default.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/OwnerManager.java
+++ b/server/src/main/java/org/candlepin/controller/OwnerManager.java
@@ -57,24 +57,57 @@ public class OwnerManager {
 
     private static Logger log = LoggerFactory.getLogger(OwnerManager.class);
 
-    @Inject private ConsumerCurator consumerCurator;
-    @Inject private PoolManager poolManager;
-    @Inject private ActivationKeyCurator activationKeyCurator;
-    @Inject private EnvironmentCurator envCurator;
-    @Inject private ExporterMetadataCurator exportCurator;
-    @Inject private ImportRecordCurator importRecordCurator;
-    @Inject private PermissionBlueprintCurator permissionCurator;
-    @Inject private OwnerProductCurator ownerProductCurator;
-    @Inject private ProductManager productManager;
-    @Inject private OwnerContentCurator ownerContentCurator;
-    @Inject private ContentManager contentManager;
-    @Inject private OwnerCurator ownerCurator;
-    @Inject private ContentAccessCertServiceAdapter contentAccessCertService;
-    @Inject private ContentAccessCertificateCurator contentAccessCertCurator;
-    @Inject private OwnerEnvContentAccessCurator ownerEnvContentAccessCurator;
-    @Inject private UeberCertificateCurator uberCertificateCurator;
-    @Inject private OwnerServiceAdapter ownerServiceAdapter;
+    /* PoolManager is injected separately so that we do not create a circular reference with the
+       PoolManager implementation which has an injection of this class. This method causes
+       the injection to occur after the constructor.
+     */
+    @Inject PoolManager poolManager;
+    private ConsumerCurator consumerCurator;
+    private ActivationKeyCurator activationKeyCurator;
+    private EnvironmentCurator envCurator;
+    private ExporterMetadataCurator exportCurator;
+    private ImportRecordCurator importRecordCurator;
+    private PermissionBlueprintCurator permissionCurator;
+    private OwnerProductCurator ownerProductCurator;
+    private ProductManager productManager;
+    private OwnerContentCurator ownerContentCurator;
+    private ContentManager contentManager;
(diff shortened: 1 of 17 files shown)


wottop:
buildbot: retest this please

wottop:
buildbot: retest this please

server/src/main/java/org/candlepin/controller/OwnerManager.java
```diff
@@ -57,26 +57,54 @@
...
+        this.ownerCurator = ownerCurator;
+        this.contentAccessCertService = contentAccessCertService;
+        this.contentAccessCertCurator = contentAccessCertCurator;
+        this.ownerEnvContentAccessCurator = ownerEnvContentAccessCurator;
+        this.uberCertificateCurator = uberCertificateCurator;
+        this.ownerServiceAdapter = ownerServiceAdapter;
+    }
     @Transactional
-    public void cleanupAndDelete(Owner owner, boolean revokeCerts) {
+    public void cleanupAndDelete(Owner owner, PoolManager poolManager, boolean revokeCerts) {
```
Ceiu: Why do we not inject the pool manager anymore?

server/src/main/java/org/candlepin/controller/OwnerManager.java
```diff
@@ -198,19 +226,18 @@ public void refreshContentAccessMode(OwnerServiceAdapter adapter, Owner owner) {
...
         // This shouldn't happen, but in the event our upstream source is having issues, let's
         // not put ourselves in a bad state as well.
-        if (upstreamList != null && !upstreamList.isEmpty()) {
+        if (!StringUtils.isBlank(upstreamList)) {
             // Not empty list. Verify that the upstream mode is present.
             String[] list = upstreamList.split(",");
-
-            if (!StringUtils.isEmpty(upstreamMode) && !ArrayUtils.contains(list, upstreamMode)) {
-                throw new IllegalStateException("Upstream access mode is not present in access mode list");
+            if (upstreamMode == null || !ArrayUtils.contains(list, upstreamMode)) {
```
Ceiu: Are we no longer treating an empty string the same as a null value? What happens if upstream gives us an empty string here?

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1518,9 +1515,9 @@ public Response getContentAccessBody(
             throw new BadRequestException(i18n.tr("Content access body " +
                "can not be requested for a share consumer"));
         }
-        if (consumer.getOwner().contentAccessMode()
-            .equals(ContentAccessCertServiceAdapter.DEFAULT_CONTENT_ACCESS_MODE)) {
-            throw new BadRequestException(i18n.tr("No content access mode assigned"));
+        if (!consumer.getOwner().getContentAccessMode()
```
Ceiu: The output of getContentAccessMode can be null now.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -948,17 +963,17 @@ public Owner updateOwner(@PathParam("owner_key") @Verify(Owner.class) String key
             toUpdate.setAutobindDisabled(owner.getAutobindDisabled());
         }
 
-        if (!ContentAccessCertServiceAdapter.DEFAULT_CONTENT_ACCESS_MODE.equals(owner.contentAccessMode()) &&
-            config.getBoolean(ConfigProperties.STANDALONE)) {
-
+        if (config.getBoolean(ConfigProperties.STANDALONE) &&
+            !toUpdate.getContentAccessMode().equals(owner.getContentAccessMode()) &&
```
Ceiu: Need a null check here.

server/src/main/java/org/candlepin/controller/OwnerManager.java
wottop: Did it for the unit test. Needed to mock.

server/src/main/java/org/candlepin/controller/OwnerManager.java
wottop: We are no longer allowing the owner fields for contentAccessMode and contentAccessModeList to be null or blank

server/src/main/java/org/candlepin/controller/OwnerManager.java
Ceiu: Is there another way to do this that isn't completely painful? I'm not a huge fan of rearranging production code to appease testing frameworks.

Ceiu:
I've updated a comment and replied above, but in general I believe I understand the intent here. However, I would still prefer that we take some steps to ensure that we can't crash out in the event something from our upstream or our backend data sources end up null.

Ceiu:
ACK

PR #1601, opened 2017-06-15
Removes the extra quotes added mistakenly in the docker refactor

cnsnyder:
This is a fix for the deploy script that was broken by the extra quotes in bash_functions.

### DIFF
--- a/bin/bash_functions
+++ b/bin/bash_functions
@@ -128,7 +128,7 @@ cert_from_pkcs12() {
 recreate_postgresql() {
     local db_name="$1"
     local db_user="${2:-$db_name}"
-    local db_host="${DBHOSTNAME:-'localhost'}"
+    local db_host="${DBHOSTNAME:-localhost}"
     # psql must use this env var for password passing
     if [ -n "${DBPASSWORD}" ]; then
      export PGPASSWORD="$DBPASSWORD"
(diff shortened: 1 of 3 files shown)


Lorquas:
looks good to me,  I say merge this if this is blocking other tests

kahowell:
Couple other places:

```
$ git grep -n -- "-'localhost'"
docker/candlepin-base/setup-db.sh:50:  local db_host="${DBHOSTNAME:-'localhost'}"
server/bin/deploy:114:  buildr -s -e $1 app_db_name=$APP_DB_NAME app_db_host=${DBHOSTNAME:-'localhost'} erb $2 $3
```

Ceiu:
One other concern @awood noticed in the error message discussed in chat, is that it appears to be setting up the database as "template1" instead of candlepin. You can verify this by deploying with ```deploy -gt`` and then connecting to the template1 db: ```psql -U candlepin template1```

cnsnyder:
@Ceiu Good point, still investigating.

cnsnyder:
@Ceiu I'm still not certain exactly what the cause of that output was but what is worth mentioning is that the tables template0 and template1 exist in the db of a fresh box made by `vagrany up` on candlepin-2.0-HOTFIX.

Given that I am tempted to say we should be good to merge this.

Ceiu:
ACK

PR #1603, opened 2017-06-16
docker-refactor 0.9.54 edition

cnsnyder:
This is the backport to the 0.9.54 hotfix branch of the docker-refactor bits.

This *should* fix the jenkins jobs when running against 0.9.54-HOTFIX branch.


cnsnyder:
I expect the qpid_spec to fail (the qpid bits are not implemented on 0.9.54 AFAIK)

PR #1606, opened 2017-06-19
Fixed an issue with migrating a database with product content

Ceiu:
- Fixed a Liquibase migration task that would fail when attempting
  to update a database that already contained product content
  when run against MySQL
- Updated the Liquibase buildr task to use a newer version of the
  XML validator
- Added a valid checksum to the ueber cert cleanup task to avoid
  issues with certain migrations


kahowell:
Only jenkins failure was the performance test, so that can be ignored.

Ceiu:
The root problem here was that when adding a new primary key to the cp2_product_content table that already contained data, each database handled the situation of having many rows with the same primary ID (null, in this case) differently. This fix populates the table with unique values for the ID before designating it as a primary key.

To test:
- Deploy Candlepin 0.9.54 or previous with test data on mysql/mariadb
- Check out master
- Deploy without recreating the database. It should fail with a truncate error (mysql) or a null value error (postgresql/oracle)
- Check out this branch
- Deploy without recreating the DB. It should succeed

PR #1609, opened 2017-06-20
Use docker-compose down to clean up old networks

kahowell:


cnsnyder:
Merging despite jenkins failures as they appear to be caused by the issue this PR fixes

PR #1614, opened 2017-06-21
1464133: Put guestId param in consumer json for virtWho instances

wottop:


wottop:
[1458674: Force Candlepin JSON generation for hypervisor checkin to include empty list for guests](https://trello.com/c/bsCziudd/463-1458674-force-candlepin-json-generation-for-hypervisor-checkin-to-include-empty-list-for-guests)

mstead:
retest this please

wottop:
https://bugzilla.redhat.com/show_bug.cgi?id=1464133

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -762,4 +762,14 @@ public void setContentAccessMode(String contentAccessMode) {
         this.contentAccessMode = contentAccessMode;
     }
 
+    /**
+     * This will put in the property so that the virtWho instances won't error
+     *
+     * @return List always empty
+     */
+    @JsonProperty("guestIds")
+    public List getEmptyGuestIdArray() {
```
Ceiu: (duplicate comment, sorry)
Are we ever going to use this property going forward? If so, does it make sense to just call this "getGuestIDs" and have it always return an empty array until we need it to do otherwise? Maybe the doc block is enough documentation here.

Ceiu:
ACK

PR #1619, opened 2017-06-23
Awood/swagger task

awood:
Add buildr plugin to download swagger.json and generate client

This plugin defines several commands:

* swagger:json - This command downloads swagger.json from the location
specified by swagger.json_source in the buildfile.  The swagger.json
file is placed in the top level of the project.  Candlepin **must** be
running for the download to be successful.

* swagger:client - This command attempts to generate a REST client using
swagger-codegen and a swagger.json file located in the top level of the
project.  Several options can be specified such as 'lang' to define the
language to use for the generated client and 'destination' to specify
where to write the resulting client files.

* swagger - This command simply combines the previous two commands for
convenience.


mstead:
@awood The spec tests won't run due to an API issue with rubyzip. According to the docs we can add the dep 'zip-zip' to maintain backwards compatibility. 

See the readme here:
https://github.com/rubyzip/rubyzip

I've tested it and the workaround seems to work fine. We don't seem to be relying on rubyzip ourselves, so I think we can just include zip-zip for now and remove it some time down the road.

mstead:
retest this please

buildfile
```diff
@@ -10,6 +10,7 @@ require 'date'
 require 'json'
 require 'net/http'
 require 'rspec/core/rake_task'
+require 'buildr/jetty'
```
mstead: We aren't using jetty at this point correct? Can we axe this?

PR #1621, opened 2017-06-26
(2.0) Re-enable JobCleaner logging

mstead:
Since JobCleaner runs every 12h by default and
nothing is currently overriding it, there is no
reason not to log it.


wottop:
looks ok, would like the tests to run before acking

wottop:
I have assurances that the tests run.

PR #1622, opened 2017-06-26
Re-enable JobCleaner logging

mstead:
Since JobCleaner runs ever 12h by default and
nothing is currently overriding it, there is no
reason not to log it.


Ceiu:
ack

PR #1625, opened 2017-06-27
1461472: add guestId array to rules

wottop:


wottop:
https://bugzilla.redhat.com/show_bug.cgi?id=1461472

PR #1626, opened 2017-06-28
1461472: Update rules to run against 0.9 versions

wottop:
Detects guestIds as non-existent in the context. This condition
 occurs when newer rules versions are run in 0.9.54 because of
 inclusion in manifests. This will delay the need to make a
 major revision bump.


wottop:
[Bug 1461472 - Unable to list available pools given a large manifest](https://trello.com/c/TVQpnY9n/470-bug-1461472-unable-to-list-available-pools-given-a-large-manifest)

PR #1631, opened 2017-07-07
Added the new org.candlepin.dto package

Ceiu:
- Added the new org.candlepin.dto package with new DTOs, translators
  and DTO factories
- Added the standard DTOFactory object, which acts as a registry for
  DTOTranslators
- Added the DTOTranslator interface, which defines/describes the methods
  for converting a model entity to a DTO
- Added DTOs for Owner, UpstreamConsumer, ConsumerType, Certificate
  and CertificateSerial, and their respective translators
- Added the ModelEntity and TimestampedModelEntity interfaces which
  represent database model entities
- AbstractHibernateObject now implements TimestampedModelEntity
- OwnerProduct and OwnerContent now implement ModelEntity
- Added the APIDTOFactory and the APIv1DTOFactory, both registered in
  the CandlepinModule for injection into the endpoint handlers
- Implemented the new DTOs in their respective resources


Ceiu:
While this is currently labeled "in progress," I do want to stress the only remaining bits I intend to add to this branch are additional tests for the DTOs themselves. As such, I feel the review process can begin now, and the merging be contingent on the addition of the remaining tests.

Also note that putting the actual DTOs into use in the resources will come in another PR after we decide if we'll be doing them individually or after we have our entire DTO model in place.

Ceiu:
I'll be adding even more DTO tests and putting them in place in a couple places to demonstrate intended use.

server/src/main/java/org/candlepin/dto/DTOFactory.java
```diff
@@ -0,0 +1,259 @@
...
+     *  The model class to unregister
+     *
+     * @throw IllegalArgumentException
+     *  if srcClass is null
+     *
+     * @return
+     *  The previously registered translator for the given class, or null a translator had not been
+     *  registered
+     */
+    public DTOTranslator unregisterTranslator(Class srcClass) {
```
Ceiu: I suppose there could be.

server/src/main/java/org/candlepin/dto/api/v1/CertificateSerialDTO.java
```diff
@@ -0,0 +1,179 @@
...
+        this.revoked = revoked;
+        return this;
+    }
+
+    /**
+     * @{inheritDocs}
+     */
+    @Override
+    public String toString() {
+        return String.format("CertificateSerialDTO []");
```
Ceiu: Nah, what's there works fine; it just won't do any substitutions. However, those are placeholders that I forgot to fill in.

server/src/main/java/org/candlepin/dto/DTOFactory.java
```diff
@@ -0,0 +1,259 @@
...
+            if (translator == null) {
+                translator = this.fetchTranslatorByInterfaces(srcClass);
+            }
+        }
+
+        return translator;
+    }
+
+    /**
+     * Recursive implementation for fetching translators by interface for a given class.
```
Ceiu: Kind of. It's only even recursive due to how identifying interfaces works. It still checks the class hierarchy first, then the interfaces in declaration (and then hierarchy) order. But because getParentClass doesn't work with interfaces, it has to be implemented the way it is to walk up the hierarchy of each implemented given interface.

server/src/main/java/org/candlepin/dto/DTOFactory.java
Ceiu: On second thought, I don't believe this would work correctly with those parameterized outputs. In the case where we register some interface I for DTOTranslator<I>, then give it a class C that implements I, we would return DTOTranslator<C> instead of DTOTranslator<I>.

server/src/main/java/org/candlepin/dto/DTOFactory.java
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * Fetches a translator for the given class. If a translator cannot be found, this method
+     * returns null.
+     * <p></p>
+     * This method uses the following algorithm to determine which translator to use:
+     * <pre>
+     * 1. Fetch the source object's class, C
+     * 2. Fetch any translator T registered for class C
+     * 3. If T is not null, go to step X
```
Ceiu: Fixed

server/src/main/java/org/candlepin/model/AbstractHibernateObject.java
```diff
@@ -33,12 +33,17 @@
...
+ *
+ * @param <T>
+ *  Entity type extending this class; should be the name of the subclass
  */
 @MappedSuperclass
 @XmlType(name = "CandlepinObject")
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonFilter("DefaultFilter")
-public abstract class AbstractHibernateObject implements Persisted, Serializable {
+public abstract class AbstractHibernateObject<T extends AbstractHibernateObject>
```
awood: Why are we adding a type argument here?  It doesn't look like there are any generic methods that use `T`.

server/src/main/java/org/candlepin/dto/CandlepinDTO.java
```diff
@@ -0,0 +1,106 @@
...
+
+
+/**
+ * The CandlepinDTO is the base class for all DTOs, which provides common data and functionality.
+ *
+ * @param <T>
+ *  DTO type extending this class; should be the name of the subclass
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public abstract class CandlepinDTO<T extends CandlepinDTO> implements Cloneable, Serializable {
```
awood: I'm confused why `CandlepinDTO` uses a generic type when it doesn't seem to have any generic methods.  I.e. method returning something of type `T`

server/src/main/java/org/candlepin/dto/api/APIDTOFactory.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.candlepin.dto.DTOFactory;
+
+
+
+/**
+ * The APIDTOFactory is an abstract class which denotes factories which are pre-configured with
+ * translators for creating DTOs compatible with the most recent API specification.
+ */
+public abstract class APIDTOFactory extends DTOFactory {
```
Ceiu: It was originally an interface, likely for the same reason you're thinking here. What I ran headlong into was when I declared an APIDTOFactory for injection, it had no knowledge of the methods from DTO factory.

Maybe the answer here is to change DTOFactory itself to an interface and have a SimpleDTOFactory implementation or some such. Thoughts?

server/src/main/java/org/candlepin/dto/api/v1/CertificateDTO.java
```diff
@@ -0,0 +1,167 @@
...
+
+        return false;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int hashCode() {
+        HashCodeBuilder builder = new HashCodeBuilder(37, 7)
```
Ceiu: I believe so. I feel that if .equals is calling up to the parent, .hashCode should as well to keep up with the hashCodes being unequal if the inequality occurs in the parent.

server/src/main/java/org/candlepin/dto/api/v1/CertificateDTO.java
```diff
@@ -0,0 +1,167 @@
...
+            .append(this.getSerial());
+
+        return builder.toHashCode();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Object clone() {
```
Ceiu: Ahh, whoops. This got reverted when I was fighting with covarience earlier in the work -- I'll update that.

Regarding the copy constructor vs clone debate, I have nothing against copy constructors, but I generally prefer a clone method to them for a couple reasons:
- Not needing to know exactly which instance we need to instantiate to get a proper clone
- Not explicitly creating new instances in client code, which has potential to make mocking/testing a pain

What I'd like to do here is keep the clone method, fix the covarience and the obvious inefficiencies I overlooked (somehow forgot that base clone already makes a shallow copy for me), and add copy constructors which make use of the populate method. Would that be acceptable?

server/src/main/java/org/candlepin/model/AbstractHibernateObject.java
Ceiu: The seemingly useless generic here are present to help work around the covarience in the parameters thing I was running into. They basically act as a compile-time marker to ensure that we don't accidentally overload our methods, and ensure we're still passing the right thing in.

server/src/main/java/org/candlepin/dto/CandlepinDTO.java
Ceiu: See above.

server/src/main/java/org/candlepin/dto/api/v1/CertificateSerialDTO.java
```diff
@@ -0,0 +1,184 @@
...
+        copy.revoked = this.isRevoked();
+
+        return copy;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public CertificateSerialDTO populate(CandlepinDTO<CertificateSerialDTO> source) {
```
Ceiu: Github is having trouble showing me the commit for this comment so I can scroll down one line, but I'll just assume it was something I fixed already. Current version looks like this:

```
    @Override
    public CertificateSerialDTO populate(CandlepinDTO<CertificateSerialDTO> source) {
        super.populate(source);
```

server/src/main/java/org/candlepin/dto/api/APIDTOFactory.java
awood: Maybe 

* DTOFactory - interface
* BaseDTOFactory - class implements DTOFactory
* APIDTOFactory - interface
* APIv1DTOFactory - extends BaseDTOFactory and implements APIDTOFactory

server/src/main/java/org/candlepin/dto/api/v1/CertificateDTO.java
awood: Yeah, I can live with that.  Can you add a link to http://jtechies.blogspot.com/2012/07/item-11-override-clone-judiciously.html to the javadoc in the clone method declared in the interface?  Just so people have a reference to how to implement clone properly.

server/src/main/java/org/candlepin/dto/api/APIDTOFactory.java
Ceiu: Done.

server/src/main/java/org/candlepin/dto/api/v1/CertificateDTO.java
Ceiu: Done

server/src/main/java/org/candlepin/dto/DTOTranslator.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param destination
+     *  The DTO to populate with data from the source object
+     *
+     * @throws IllegalArgumentException
+     *  if either source or destination are null or not subclasses of the proper classes
+     *
+     * @return
+     *  The populated DTO
+     */
+    CandlepinDTO populate(ModelEntity<I> source, CandlepinDTO<O> destination);
```
Ceiu: Not sure, but due to the implementation, it was virtually zero effort to add, so I did.

server/src/main/java/org/candlepin/model/AbstractHibernateObject.java
awood: What was the "covarience in the parameters" issue?  Is this an issue with the `populate` method calling super?

Ceiu: Parameters themselves aren't covarient, you'll end up overloading instead of overriding. So, to get type-specific, overrideable methods, those interfaces and types were added.

awood: Okay, I think it would make a little more sense to look like:

CandlepinDTO
```java
public CandlepinDTO populate(T source) {
```

TimestampedCandlepinDTO
```java
public TimestampedCandlepinDTO populate(T source) {
```
and then in OwnerDTO
```java
public OwnerDTO populate(OwnerDTO source) {
```

That compiles for me when I implement it everywhere and has the bonus of removing lines like

```java
        // We're safeguarded here by both the generic definition above, and the runtime-level check
        // in CandlepinDTO's populate method.
        CertificateSerialDTO src = (CertificateSerialDTO) source;
```

Ceiu: It compiles, but it's not working as you expect. The ```public OwnerDTO populate(OwnerDTO source)``` is an overload of populate, so you end up with a ```populate(Object)``` (or maybe ```populate(CandlepinDTO)``` depending on how erasure works with scoped generics) and ```populate(OwnerDTO)```.

That's what I ran into while testing.

awood: I made the following changes and I didn't see any warnings about `@Override` not actually doing anything.  IDEA normally warns about that too.  I don't think there are any erasure issues because the generic type in CandlepinDTO is reified by the time you get to an actual DTO implementation.

```diff
diff --git a/server/src/main/java/org/candlepin/dto/CandlepinDTO.java b/server/src/main/java/org/candlepin/dto/CandlepinDTO.java
index c89320c..fca72bb 100644
--- a/server/src/main/java/org/candlepin/dto/CandlepinDTO.java
+++ b/server/src/main/java/org/candlepin/dto/CandlepinDTO.java
@@ -90,7 +90,7 @@ public abstract class CandlepinDTO<T extends CandlepinDTO> implements Cloneable,
      * @return
      *  this CandlepinDTO instance
      */
-    public CandlepinDTO populate(CandlepinDTO<T> source) {
+    public CandlepinDTO populate(T source) {
         if (source == null) {
             throw new IllegalArgumentException("source is null");
         }
diff --git a/server/src/main/java/org/candlepin/dto/api/v1/CertificateDTO.java b/server/src/main/java/org/candlepin/dto/api/v1/CertificateDTO.java
index 752dfa2..175a04a 100644
--- a/server/src/main/java/org/candlepin/dto/api/v1/CertificateDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/CertificateDTO.java
@@ -150,7 +150,7 @@ public class CertificateDTO extends TimestampedCandlepinDTO<CertificateDTO> {
      * {@inheritDoc}
      */
     @Override
-    public CertificateDTO populate(CandlepinDTO<CertificateDTO> source) {
+    public CertificateDTO populate(CertificateDTO source) {
         super.populate(source);
 
         // We're safeguarded here by both the generic definition above, and the runtime-level check
diff --git a/server/src/main/java/org/candlepin/dto/api/v1/CertificateSerialDTO.java b/server/src/main/java/org/candlepin/dto/api/v1/CertificateSerialDTO.java
index 2cd1537..9e8746a 100644
--- a/server/src/main/java/org/candlepin/dto/api/v1/CertificateSerialDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/CertificateSerialDTO.java
@@ -166,7 +166,7 @@ public class CertificateSerialDTO extends TimestampedCandlepinDTO<CertificateSer
      * {@inheritDoc}
      */
     @Override
-    public CertificateSerialDTO populate(CandlepinDTO<CertificateSerialDTO> source) {
+    public CertificateSerialDTO populate(CertificateSerialDTO source) {
         super.populate(source);
 
         // We're safeguarded here by both the generic definition above, and the runtime-level check
diff --git a/server/src/main/java/org/candlepin/dto/api/v1/ConsumerTypeDTO.java b/server/src/main/java/org/candlepin/dto/api/v1/ConsumerTypeDTO.java
index a2003d4..624e58f 100644
--- a/server/src/main/java/org/candlepin/dto/api/v1/ConsumerTypeDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/ConsumerTypeDTO.java
@@ -134,7 +134,7 @@ public class ConsumerTypeDTO extends TimestampedCandlepinDTO<ConsumerTypeDTO> {
      * {@inheritDoc}
      */
     @Override
-    public ConsumerTypeDTO populate(CandlepinDTO<ConsumerTypeDTO> source) {
+    public ConsumerTypeDTO populate(ConsumerTypeDTO source) {
         super.populate(source);
 
         // We're safeguarded here by both the generic definition above, and the runtime-level check
diff --git a/server/src/main/java/org/candlepin/dto/api/v1/OwnerDTO.java b/server/src/main/java/org/candlepin/dto/api/v1/OwnerDTO.java
index 7339e5e..63adbd3 100644
--- a/server/src/main/java/org/candlepin/dto/api/v1/OwnerDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/OwnerDTO.java
@@ -249,7 +249,7 @@ public class OwnerDTO extends TimestampedCandlepinDTO<OwnerDTO> implements Linka
      * {@inheritDoc}
      */
     @Override
-    public OwnerDTO populate(CandlepinDTO<OwnerDTO> source) {
+    public OwnerDTO populate(OwnerDTO source) {
         super.populate(source);
 
         // We're safeguarded here by both the generic definition above, and the runtime-level check
diff --git a/server/src/main/java/org/candlepin/dto/api/v1/TimestampedCandlepinDTO.java b/server/src/main/java/org/candlepin/dto/api/v1/TimestampedCandlepinDTO.java
index 533ec7c..9135fcb 100644
--- a/server/src/main/java/org/candlepin/dto/api/v1/TimestampedCandlepinDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/TimestampedCandlepinDTO.java
@@ -174,7 +174,7 @@ public abstract class TimestampedCandlepinDTO<T extends TimestampedCandlepinDTO>
      * {@inheritDoc}
      */
     @Override
-    public TimestampedCandlepinDTO populate(CandlepinDTO<T> source) {
+    public TimestampedCandlepinDTO populate(T source) {
         super.populate(source);
 
         // We're safeguarded here by both the generic definition above, and the runtime-level check
diff --git a/server/src/main/java/org/candlepin/dto/api/v1/UpstreamConsumerDTO.java b/server/src/main/java/org/candlepin/dto/api/v1/UpstreamConsumerDTO.java
index 9d87185..b2ef561 100644
--- a/server/src/main/java/org/candlepin/dto/api/v1/UpstreamConsumerDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/UpstreamConsumerDTO.java
@@ -207,7 +207,7 @@ public class UpstreamConsumerDTO extends TimestampedCandlepinDTO<UpstreamConsume
      * {@inheritDoc}
      */
     @Override
-    public UpstreamConsumerDTO populate(CandlepinDTO<UpstreamConsumerDTO> source) {
+    public UpstreamConsumerDTO populate(UpstreamConsumerDTO source) {
         super.populate(source);
 
         // We're safeguarded here by both the generic definition above, and the runtime-level check
```

Ceiu: Done.

PR #1632, opened 2017-07-11
1452694: Candlepin no longer deletes pools for custom subscriptions (2.0)

Ceiu:
- Pools for custom subscriptions will no longer be deleted on manifest
  delete/undo import
- The Per-org products migration has been updated/fixed to also migrate
  subscription information from the subscription table to the pool table
- Bonus pools now inherit upstream pool and customer data from their
  source pool

### DIFF
--- a/server/spec/import_spec.rb
+++ b/server/spec/import_spec.rb
@@ -120,7 +120,8 @@ describe 'Import Test Group:', :serial => true do
     it 'can be undone' do
       # Make a custom pool so we can be sure it does not get wiped
       # out during either the undo or a subsequent re-import:
-      prod = create_product(random_string(), random_string(), {:owner => @import_owner['key']})
+      prod_name = random_string("custom_pool_prod-")
+      prod = create_product(prod_name, prod_name, {:owner => @import_owner['key']})
       custom_pool = @cp.create_pool(@import_owner['key'], prod['id'])
 
       job = @import_owner_client.undo_import(@import_owner['key'])
--- a/server/src/main/java/org/candlepin/pinsetter/tasks/UndoImportsJob.java
+++ b/server/src/main/java/org/candlepin/pinsetter/tasks/UndoImportsJob.java
@@ -131,7 +131,7 @@ public class UndoImportsJob extends UniqueByEntityJob {
 
             List<Pool> pools = this.poolManager.listPoolsByOwner(owner).list();
             for (Pool pool : pools) {
-                if (pool.getSourceSubscription() != null && !pool.getType().isDerivedType()) {
+                if (this.poolManager.isManaged(pool)) {
                     this.poolManager.deletePool(pool);
                 }
             }
--- a/server/src/main/java/org/candlepin/policy/js/JsRunnerProvider.java
+++ b/server/src/main/java/org/candlepin/policy/js/JsRunnerProvider.java
@@ -112,7 +112,7 @@ public class JsRunnerProvider implements Provider<JsRunner> {
                 return;
             }
 
-            log.info("Recompiling rules with timestamp: " + newUpdated);
+            log.info("Recompiling rules with timestamp: {}", newUpdated);
 
             Context context = Context.enter();
             context.setOptimizationLevel(9);
--- a/server/spec/rules_import_spec.rb
+++ b/server/spec/rules_import_spec.rb
@@ -11,7 +11,7 @@ describe 'Rules Import', :serial => true do
     # ones that may have been left in the database:
     @cp.delete_rules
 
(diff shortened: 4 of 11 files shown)


server/spec/import_spec.rb
```diff
@@ -120,7 +120,8 @@ def import_and_wait
     it 'can be undone' do
       # Make a custom pool so we can be sure it does not get wiped
       # out during either the undo or a subsequent re-import:
-      prod = create_product(random_string(), random_string(), {:owner => @import_owner['key']})
+      prod_name = "custom_pool_prod-" + random_string()
```
awood: Just for future reference, `random_string("custom_pool_prod-")` would accomplish the same thing I believe.  Don't worry about changing it but I think the more concise usage would be preferable in the future.

server/src/main/java/org/candlepin/pinsetter/tasks/UndoImportsJob.java
```diff
@@ -129,9 +133,26 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
...
+            // we've received from an import (and, thus, are eligible for deletion). In hosted,
+            // however, we *are* the upstream source, so everything is eligible for removal.
+            // This is pretty hacky, so the way we go about doing this check should eventually be
+            // replaced with something more generic and reliable, and not dependent on the config.
+
+            // Note: this logic is duplicated in CandlepinPoolManager.refreshPoolsWithRegeneration
+
+            // TODO:
+            // Remove the standalone config check and replace it with a check for whether or not the
+            // pool is "managed"  -- however we decide to implement that in the future.
```
awood: Could we go ahead and implement `isManaged()` and just have it return `!standalone || pool.getUpstreamPoolId() != null`?  That way there's one obvious place to go when we decide what being managed actually is.

server/src/main/java/org/candlepin/policy/js/JsRunnerProvider.java
```diff
@@ -112,7 +112,7 @@ public void compileRules(boolean forceRefresh) {
                 return;
             }
 
-            log.info("Recompiling rules with timestamp: " + newUpdated);
+            log.info("Recompiling rules with timestamp: {}", newUpdated);
```
awood: 👏

server/src/main/java/org/candlepin/pinsetter/tasks/UndoImportsJob.java
Ceiu: Do we have a convenient way to have a model object fetch the config without requiring it be passed in explicitly? As it is now, that check is part of the "is managed" bit.

server/spec/import_spec.rb
Ceiu: Gotcha. I'll address this here since there are other changes to make anyway.

server/src/main/java/org/candlepin/pinsetter/tasks/UndoImportsJob.java
awood: I think you could get it out of the resteasy context.  I'm pretty sure we put it in there
but not sure if we really want to pollute the model with looking into the app's config.
Getting it from the context would be like `public boolean isManaged(@Context Config config)` if I recall correctly.

mstead: I'm not really sure that I like the config leaking into the model object. Could we just set 'managed' appropriately on the pool when it is created via the API vs a refresh? Then it would be easy to determine at a database level.

Ceiu: While such a change would be optimal, I feel it would be a rather large change, and one I'm not sure we want to do with this PR. It's something that's better done with the pool hierarchy task.

mstead: @Ceiu  Agreed, I didn't intend it to be done now.

Ceiu: I wonder if we can split the difference and add the isManaged method to CandlepinPoolManager (which UndoImportsJob uses) with a TODO to move it to Pool when the logic is no longer reliant upon the config for proper operation.

awood: @Ceiu something like `poolManager.isManaged(pool)`?  I'd be fine with that.

mstead: 👍 I like that as well.

PR #1634, opened 2017-07-13
Updated CMV to store and verify activation keys (0.9.51)

Ceiu:
- CMV will now pull and verify activation keys
- Fixed a bug in candlepin_api which was sending activation key
  requests to the wrong URL


Ceiu:
retest this please

PR #1638, opened 2017-07-13
Account for misfires in job status

wottop:
If you want to see the misfire happen do the following:

Alter ConfigProperties.java such that org.quartz.threadPool.threadCount = 1

Then add a sleep to the job you want to try out. I did ManifestCleanerJob.java
```
     public void toExecute(JobExecutionContext arg0) throws JobExecutionException {
         File baseDir = new File(config.getString(ConfigProperties.SYNC_WORK_DIR));
        int maxAgeInMinutes = config.getInt(ConfigProperties.MANIFEST_CLEANER_JOB_MAX_AGE_IN_MINUTES);
        try { Thread.sleep(10000); } catch (InterruptedException ie) { };
```

You will need to flood the API with requests to run the job. Here is a the bash I used:

```
#!/bin/bash
for it in `seq 1 150`;
do
    curl -k -u admin:admin -X POST https://localhost:8443/candlepin/jobs/schedule/ManifestCleanerJob
done
```

About 30 seconds after the script finishes running, you will see in the DB jobs that have the "mayFireAgain" message in the result as well as a pending state. Eventually, the jobs will all get picked up again and run [about 10 minutes].

I was not able to get it to hit the "!mayFireAgain" path except by unit test.  I tried setting the default number of retries down to 0. I tried setting the end time on the trigger to only a minute later. I simply got the same retry scenario as when those were not set.

If you have a !mayFireAgain condition, I'd love to see it.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/core/PinsetterTriggerListener.java
+++ b/server/src/main/java/org/candlepin/pinsetter/core/PinsetterTriggerListener.java
@@ -17,11 +17,15 @@ package org.candlepin.pinsetter.core;
 import com.google.inject.Inject;
 import org.candlepin.controller.ModeManager;
 import org.candlepin.model.CandlepinModeChange.Mode;
+import org.candlepin.model.JobCurator;
+import org.candlepin.pinsetter.core.model.JobStatus;
 import org.quartz.JobExecutionContext;
 import org.quartz.Trigger;
 import org.quartz.listeners.TriggerListenerSupport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.text.SimpleDateFormat;
 
 /**
  * This component receives events around job status and performs actions to
@@ -31,10 +35,12 @@ import org.slf4j.LoggerFactory;
 public class PinsetterTriggerListener extends TriggerListenerSupport {
     private static Logger log = LoggerFactory.getLogger(PinsetterTriggerListener.class);
     private ModeManager modeManager;
+    private JobCurator jobCurator;
 
     @Inject
-    public PinsetterTriggerListener(ModeManager modeManager) {
+    public PinsetterTriggerListener(ModeManager modeManager, JobCurator jobCurator) {
         this.modeManager = modeManager;
+        this.jobCurator =  jobCurator;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


wottop:
retest this please

mstead:
@wottop The database transactions are Ok outside of tests? I was expecting that we'd have to utilize a new UnitOfWork since Quartz is invoking the callback, but perhaps I'm wrong?

server/src/main/java/org/candlepin/pinsetter/core/PinsetterTriggerListener.java
```diff
@@ -64,5 +70,23 @@ public void triggerMisfired(Trigger trigger) {
...
+            log.warn("Job {} not allowed to fire again. Setting state to FAILED.", id);
+            js.setState(JobStatus.JobState.FAILED);
+            js.setResult("Failed run. Will not attempt again.");
+        }
+        else {
+            SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
+            js.setResult("Will reattempt job at or after " + sdf.format(trigger.getNextFireTime()));
+            js.setState(JobStatus.JobState.PENDING);
+        }
+        jobCurator.merge(js);
```
mstead: Quartz best practices state that all listener methods contain a try-catch and should handle ALL exceptions [1]. If something happens when committing the JobStatus, we violate this best practice and could potentially introduce issues.

[1] http://www.quartz-scheduler.org/documentation/best-practices.html#listeners-(triggerlistener-joblistener-schedulerlistener

mstead: As stated in another comment, I was expecting that we would need to utilize a new UnitOfWork when committing the changes to the JobStatus as this listener is invoked by Quartz outside of a regular transaction. Do you know if this is the case?

mstead: Is there any code in candlepin that might attempt to restart a job that is in PENDING state? Since this is a misfire, and further scheduling will be handled by Quartz itself, let's make sure that candlepin won't attempt to restart this itself.

server/src/main/java/org/candlepin/pinsetter/core/model/JobStatus.java
```diff
@@ -177,6 +177,7 @@ public void update(JobExecutionContext context) {
             if (runTime > -1) {
                 setState(JobState.FINISHED);
                 this.finishTime = new Date(startTime.getTime() + runTime);
+                setResult("");
```
mstead: Wouldn't this clear any job result that was set by jobs that didn't misfire?

wottop:
[[8] Update PinsetterTriggerListener to set the JobStatus state to FAILED if the misfire occurs and the trigger.mayRefireAgain() returns false.](https://trello.com/c/A824Dmss/471-8-update-pinsettertriggerlistener-to-set-the-jobstatus-state-to-failed-if-the-misfire-occurs-and-the-triggermayrefireagain-retur)

server/src/main/java/org/candlepin/pinsetter/core/PinsetterTriggerListener.java
wottop: the only time we check for the state of PENDING, it has the same outcome as CREATED. UniqueByEntityJob.scheduleJob

wottop: Added try/catch

PR #1640, opened 2017-07-17
Fix custom pom plugin to be compatible with new Buildr version.

awood:
The new version of Buildr added attributes to its POM plugin that we did
not have in our POM plugin.  When the IDEA plugin attempted to access
these attributes, an exception was thrown.  This patch adds in those
attributes and writes them, if defined, into the POM file the same way
that Buildr would itself.


mstead:
retest this please

PR #1644, opened 2017-07-20
Awood/unify entitlement rules

awood:
It's crazy to have code in AbstractEntitlementRules and EntitlementRules when EntilementRules is the only class that extends AbstractEntitlementRules.


wottop:
ack

PR #1647, opened 2017-07-21
1469577: purge orphan pools via liquibase

vritant:


vritant:
[Investigate DB script that can be run by liquibase to clean up rogue pools.](https://trello.com/c/uKtpqwWt/487-investigate-db-script-that-can-be-run-by-liquibase-to-clean-up-rogue-pools)

wottop:
Fails on mysql

server/src/main/resources/db/changelog/20170718095058-purge-orphaned-pools.xml
```diff
@@ -0,0 +1,33 @@
...
+    <changeSet id="20170718095058-1" author="vrjain">
+        <comment> purge orphaned pools </comment>
+        <sql>
+            delete from cp_pool where id in
+            (  select p.id
+               from cp_pool p
+                   left outer join cp_pool_source_stack ss
+               on p.id = ss.derivedpool_id
+               where p.type='STACK_DERIVED'
+                 and ss.derivedpool_id is null
```
Ceiu: This inner query will never match anything. We're joining on ss.derivedpool_id (which will never match using equals if the value is null), but then checking that the value is null.

mstead:
After speaking with mccune, we decided to go with a script + Kbase instead of an automatic cleanup due to complexity and added risk. Closing.

PR #1657, opened 2017-07-28
1474698: Trim String values when converting to num

kahowell:
Quick fix coded on a Friday afternoon (please double-check me :smile:). Happy to cherry-pick to other branches as well if wanted.

### DIFF
--- a/common/src/main/java/org/candlepin/common/config/PropertyConverter.java
+++ b/common/src/main/java/org/candlepin/common/config/PropertyConverter.java
@@ -104,7 +104,7 @@ public class PropertyConverter {
             return (Number) value;
         }
         else {
-            String str = value.toString();
+            String str = value.toString().trim();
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/candlepin/common/config/PropertyConverter.java
```diff
@@ -104,7 +104,7 @@ protected static Number toNumber(Object value, Class<?> clazz) throws Conversion
             return (Number) value;
         }
         else {
-            String str = value.toString();
+            String str = value.toString().trim();
```
Ceiu: Non-NACKing comment: Not a huge fan of the fact we aren't first checking if the value is null, but that unfortunate oversight predates this change.

Ceiu:
ACK

Should be good to merge once Jenkins tests find that this doesn't break anything.

Ceiu:

PR #1658, opened 2017-07-28
Per-org products migration now examines activation keys and environments

Ceiu:
- Products which provide content used by an environment are
  now migrated as part of the product migration step
- Content used by an environment that is not provided by
  any product will be migrated during content migration
- Products specified by an activation key are now migrated
- Added additional data validation for activation key products
  and environment content before migrating data


Ceiu:
The Qpid test itself looks broken, as per usual. Probably safe to ignore that failure (ugh).

wottop:
ack, worked as expected.

PR #1661, opened 2017-07-31
1440910: Choose host_specific/virt_only pools first (forward port to 2.0 of 1440909)

cnsnyder:
This is a change to the rules that allows virtualized systems to
reliably receive entitlements from host_specific and/or virt_only
pools before the consumption of quantity from physical pools.

There is one case that this change could make a bit worse.
In the case that there are not enough host_specific and/or
virt-only pools available to cover the products on the guest
system, the guest will be given entitlements from all available
virtual pools that provide at least one product in addition to
receiving an entitlement from as many physical pools as necessary
to cover the products installed.

If the available physical pools included one that would provide
all the installed products on the guest, all the entitlements
from the virtual pools would be redundant.

Adds a spec test for host pool prioritization

### DIFF
--- a/server/src/main/resources/rules/rules.js
+++ b/server/src/main/resources/rules/rules.js
@@ -1938,26 +1938,34 @@ var Autobind = {
                 return true;
             },
 
-            get_num_host_specific: function() {
+            is_host_specific: function (pool) {
+                // returns true if the given pool is host specific
+                return pool.getAttribute(REQUIRES_HOST_ATTRIBUTE) != null;
+            },
+
+            is_virt_only: function (pool) {
+                // returns true if pool is virt-only
+                return Utils.equalsIgnoreCase('true', pool.getProductAttribute(VIRT_ONLY));
+            },
+
+            count_matching_pools: function (test_func) {
+                // Returns the number of pools for which test_func(pool) evaluates to true
                 var count = 0;
-                for (var i = 0; i < this.pools.length; i++) {
+                for (var i =0; i < this.pools.length; i++) {
                     var pool = this.pools[i];
-                    if (pool.getAttribute(REQUIRES_HOST_ATTRIBUTE) != null) {
+                    if (test_func(pool)){
                         count++;
                     }
                 }
                 return count;
             },
 
-            get_num_virt_only: function() {
-                var count = 0;
-                for (var i = 0; i < this.pools.length; i++) {
-                    var pool = this.pools[i];
-                    if (Utils.equalsIgnoreCase('true',  pool.getProductAttribute(VIRT_ONLY))) {
-                        count++;
-                    }
-                }
-                return count;
+            num_host_specific: function () {
+                return this.count_matching_pools(this.is_host_specific);
+            },
+
+            num_virt_only: function () {
+                return this.count_matching_pools(this.is_virt_only);
             },
 
             /*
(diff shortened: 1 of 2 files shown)


server/src/main/resources/rules/rules.js
```diff
@@ -2447,44 +2455,62 @@ var Autobind = {
...
+                (virt_only_found && group_num_virt_only < best_num_virt_only)) {
+                // Skip this group if we've found virt or host_specific and this group is not.
+                continue;
+            }
+
+            new_best_found = false;
+            if (group_num_host_specific > best_num_host_specific) {
+                host_specific_found = true;
+                new_best_found = true;
+            } else if (group_num_host_specific < best_num_host_specific) {
```
Ceiu: Seems the javascript files sneak around checkstyle. All of these "else if..." bits should be on a new line, as such:

```
if (/*something*/) {
    // Do something
}
else if (/* something else */) {
    // Do the other thing
}
else if (...) {
   ...
```

server/src/main/resources/rules/rules.js
```diff
@@ -2447,44 +2455,62 @@ var Autobind = {
...
+            } else if (group_avg_prio < best_avg_prio) {
+                new_best_found = false;
+            } else if (group_poolquantity < total_poolquantity) {
+                new_best_found = true;
+            } else if (group_poolquantity > total_poolquantity) {
+                new_best_found = false;
+            } else if (stacked && !group.stackable) {
+                new_best_found = true;
+            }
+            if (new_best_found) {
```
Ceiu: Add a line of whitespace before this line so it's clear this conditional isn't part of the previous tree of them.

server/src/main/resources/rules/rules.js
```diff
@@ -2447,44 +2455,62 @@ var Autobind = {
...
+                // Skip this group if we've found virt or host_specific and this group is not.
+                continue;
+            }
+
+            new_best_found = false;
+            if (group_num_host_specific > best_num_host_specific) {
+                host_specific_found = true;
+                new_best_found = true;
+            } else if (group_num_host_specific < best_num_host_specific) {
+                new_best_found = false;
```
Ceiu: Doesn't this risk shutting off the "best found" after turning it on if it happens to come to this group after finding a better group?

This happens a few more times below as well; not going to repeat my comment for them.

cnsnyder: I don't think so. The best group found is stored in 'best'. The vars used for comparison are also updated when the 'best' var is updated. The updated happens at the end of each loop of the for loop. Since each else if block will only be checked if the prior check failed, it shouldn't be possible for further conditions to be checked and to modify the 'new_best_found'.

cnsnyder:
@Ceiu Updated based on feedback! Please let me know if there is anything else remaining that should be changed!

Cheers,
Chris

Ceiu:
ACK

cnsnyder:

cnsnyder:

mstead:
retest this please

cnsnyder:

cnsnyder:
I believe the following PR will fix the registry errors we are seeing in the test failures here: https://github.com/candlepin/candlepin/pull/1673

cnsnyder:

cnsnyder:
Forgot to update the rules version. That should be bumped with this PR.

cnsnyder:
The new commit updates the rules to be the same as they would be in 2.1 if #1685 is merged. The version needed a bump. I've been informed that the rules should be the same in 2.X.

mstead:
Can we squash the commits for this one before it is merged.

barnabycourt:
retest this please

Ceiu:

barnabycourt:
@mstead @Ceiu What is this PR waiting on ?

mstead:
> @mstead @Ceiu What is this PR waiting on ?

I was waiting to have the commits squashed, given the nature of the second commit, I'd be ok with leaving it as 2 commits.

Ceiu:
Since we're no longer waiting for a squash...

PR #1662, opened 2017-07-31
The populate-hosted-db task now fetches activation key products

Ceiu:
- Products referenced by activation keys will now be fetched while
  running the populated-hosted-db task

### DIFF
--- a/server/src/main/java/org/candlepin/model/PoolCurator.java
+++ b/server/src/main/java/org/candlepin/model/PoolCurator.java
@@ -32,6 +32,7 @@ import org.hibernate.Hibernate;
 import org.hibernate.LockOptions;
 import org.hibernate.Query;
 import org.hibernate.ReplicationMode;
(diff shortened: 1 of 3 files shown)


Ceiu:
To test:
- Deploy candlepin-0.9.51-HOTFIX in hosted mode
- Create a new activation key and attach a product ID to it
- Run the pophosteddb task
- Depending on the product ID attached, one of two things will happen:
  - If the product ID is valid, the task will complete successfully and there will be an INFO-level line in candlepin.log along the lines of "Processing product with ID: <pid>", where <pid> is the product ID in question
  - If the product ID is not valid, the task will still complete, however there will be a WARN-level line in candlepin.log indicating that no such product could be found ("Unable to find product for referenced product ID: <pid>").

Ceiu:
Forgot an import statement, it seems. Will fix this in the AM.

wottop:
ack

PR #1663, opened 2017-07-31
Per-org products migration now examines activation keys and environments (master/2.1)

Ceiu:
- Products which provide content used by an environment are
  now migrated as part of the product migration step
- Content used by an environment that is not provided by
  any product will be migrated during content migration
- Products specified by an activation key are now migrated
- Added additional data validation for activation key products
  and environment content before migrating data


Ceiu:
Testing requires a specific database setup, so just ping or e-mail me and I'll forward the DB dump and instructions.

wottop:
ack

PR #1664, opened 2017-08-01
1440924: Adjust message when activation key registration fails [master]

wottop:


wottop:
[[13] 1440924 - [RFE] - better error reporting: When you try to register a client with an activation key, if there are no usable subscriptions](https://trello.com/c/K5GsBmVE/502-13-1440924-rfe-better-error-reporting-when-you-try-to-register-a-client-with-an-activation-key-if-there-are-no-usable-subscripti)

Ceiu:
ACK

PR #1666, opened 2017-08-01
[0.9.54] Added entitlement validation to CMV

Ceiu:
- The CMV tool now validates consumer entitlements for an org
- CMV no longer checks any field named "updated," as these are
  universally problematic timestamp fields that are expected to change
  during a migration anyway
- Added a utility script, generate_consumers.rb, for generating
  consumers and consuming pools in a Candlepin deployment


wottop:
ack

PR #1671, opened 2017-08-04
Break up massive in clauses in listBySourceEntitlements.

awood:


server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -136,13 +137,15 @@ public Pool find(Serializable id) {
             return new ArrayList<Pool>();
         }
 
-        List<Pool> results = createSecureCriteria()
-            .add(Restrictions.in("sourceEntitlement", ents))
-            .setFetchMode("entitlements", FetchMode.JOIN)
-            .list();
+        Iterable<List<Entitlement>> blocks = Iterables.partition(ents, IN_OPERATOR_BLOCK_SIZE);
```
Ceiu: We can probably partition on the QUERY_PARAMETER_LIMIT (may not exist in 0.9.x, but it's set to 32k by default) rather than the IN_OPERATOR_BLOCK_SIZE. If we opt to keep this as-is, we should probably drop use of CPRestrictions, as it won't add anything of value.

Ceiu:
ACK

Will merge once Jenkins does Jenkins things and I throw together a quick manual test for it.

awood:
bot, retest this please

awood:
Looks like Jenkins can't pull down the Docker images?

Ceiu:
spec tests were manually verified

PR #1672, opened 2017-08-08
updating docker registry to new engineering host

Lorquas:
Jenkins slaves have been updated (that work is in the rhsm-ansible repo), this should be all that is needed candlepin side.  This probably needs to be cherrypicked into the HOTFIX and other (?) branches, which @kahowell has written a nice script to do that.


Ceiu:
ACK

PR #1678, opened 2017-08-10
Fix custom pool lookup for bonus pool - 2.0

vritant:
Fix custom pool lookup for bonus pool

the absence of source subscription is not a reliable way to determine if a pool is a custom pool,
since that behaviour varies with the candlepin verion. More reliable is the presence or absence of
an upstream pool id.

CandlepinPoolManager.isManaged is now the single method to determine if a pool is custom, and checks
both source subscription and upstream pool id.

This is ready for a merge, but would like @mstead to confirm if this fix should be pushed in this branch.

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -1075,6 +1075,10 @@ class Candlepin
       pool['derivedProvidedProducts'] = params[:derived_provided_products].collect { |pid| {'productId' => pid} }
     end
 
+    if params[:upstream_pool_id]
+      pool['upstreamPoolId'] = params[:upstream_pool_id]
+    end
+
     return post("/owners/#{owner_key}/pools", pool)
   end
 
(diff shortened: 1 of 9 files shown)


vritant:
[Custom pools in 2.X](https://trello.com/c/zjgmn1a5/524-custom-pools-in-2x)

wottop:
ack the change. pending tests.

wottop:
retest this please

mstead:
@vritant Please add some context to your commit message outlining why this was broken.

barnabycourt:
retest this please

wottop:
@vritant These failures might be real. Standalone system, bonus pool spec tests.

mstead:
I think that the spec tests may be failing for a valid reason. Please look into it to see if the new use of isManaged is the culprit. If it is, we need to determine what we are breaking with this change.

vritant:
Closing cause this fix is fixing a bug that is likely not needing a fix in this branch.

PR #1680, opened 2017-08-11
Warn devs about transaction isolation level on mysql

vritant:
candlepin's mysql deployment requires READ-COMMITTED isolation level, otherwise some spec tests fail and the developers are left with no clue why. This warning should help with that, and is fired only if the current isolation level is not as expected.


mstead:
retest this please

mstead:
@vritant Please add more context to your commit message as to why this is important.

PR #1686, opened 2017-08-16
Remove an unused backreference from IdentityCertificate

barnabycourt:
This consumer backreference caused hibernate to load another copy of the consumer from the database whenever the IdentityCertificate was loaded. For example, a consumer loads the identity which reloads the consumer.


Lorquas:
retest this please

kahowell:
retest this please

PR #1693, opened 2017-08-17
ambiguous subscription resources

vritant:
I'd like to call out the following items this PR does, one should consider before merging:
* discontinues the following APIs and provides replacements if they did not exist already
  * PUT owner/subscriptions
  * GET subscriptions/{id}
  * GET subscriptions/{id}/cert
* SHARED is no longer a pool type, instead we have two new fields 
  * createdByShare ( to identify a pool that was created because of a share)
  * hasSharedAncestor ( to identify shared pools and their descendants )
* The rest API "DELETE pool/{id}" now deletes child pools of the pool id provided.
  * This is necessary if it is to replace "DELETE subscription/{id}"
 * CPM.deletePools no longer  sends out events for revoking entitlements when the pool is deleted.
   * This was already the implementation in CPM.deletePool, and I think we should keep that behavior consistent across candlepin, irrespective of the method.
* PoolCurator method that fetched pools for a subscription id not scoped by owner id were either removed OR are now filtering out shared pools and their descendants.
* Henceforth, not only do we de-priortize shared pools during autobind, we do the same for their descendants as well.

### DIFF
--- a/common/src/main/java/org/candlepin/common/exceptions/ResourceMovedException.java
+++ b/common/src/main/java/org/candlepin/common/exceptions/ResourceMovedException.java
@@ -1,0 +1,39 @@
+/**
+ * Copyright (c) 2009 - 2012 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.common.exceptions;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.ws.rs.core.Response.Status;
+
+/**
+ * Represents a See Other (HTTP 303) error.
+ */
+public class ResourceMovedException extends CandlepinException {
+    String location = null;
+
+    public ResourceMovedException(String argLocation) {
+        super(Status.SEE_OTHER, "The resource has moved to the location: " + argLocation);
+        location = argLocation;
+    }
+
+    @Override
+    public Map<String, String> headers() {
+        HashMap<String, String> negHeaders = new HashMap<String, String>();
+        negHeaders.put("Location", location);
+        return  negHeaders;
+    }
+}
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -1967,7 +1967,10 @@ public class CandlepinPoolManager implements PoolManager {
             }
 
             poolsToDelete.add(pool);
-            entitlementsToRevoke.addAll(pool.getEntitlements());
+            for (Entitlement e: pool.getEntitlements()) {
+                e.setDeletedFromPool(true);
+                entitlementsToRevoke.add(e);
+            }
         }
 
         // Look up the related pools for these subscription IDs.
@@ -2184,11 +2187,6 @@ public class CandlepinPoolManager implements PoolManager {
         return this.poolCurator.getPoolsBySubscriptionId(subscriptionId);
     }
 
-    @Override
-    public Pool getMasterPoolBySubscriptionId(String subscriptionId) {
-        return this.poolCurator.getMasterPoolBySubscriptionId(subscriptionId);
-    }
-
     /**
      * @{inheritDoc}
      */
@@ -2288,7 +2286,10 @@ public class CandlepinPoolManager implements PoolManager {
         // Remove the standalone config check and replace it with a check for whether or not the
         // pool is non-custom  -- however we decide to implement that in the future.
 
-        return pool != null && pool.getSourceSubscription() != null && !pool.getType().isDerivedType() &&
+        return pool != null &&
+            pool.getSourceSubscription() != null &&
+            !pool.getType().isDerivedType() &&
+            !pool.isCreatedByShare() &&
             (pool.getUpstreamPoolId() != null || !this.config.getBoolean(ConfigProperties.STANDALONE, true));
     }
 }
--- a/server/src/main/java/org/candlepin/controller/PoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/PoolManager.java
@@ -285,18 +285,6 @@ public interface PoolManager {
     List<Pool> getPoolsBySubscriptionId(String subscriptionId);
 
     /**
-     * Retrieves the master pool associated with the specified subscription ID. If there is not a
-     * master pool asscoated with the given subscription, this method should return null.
-     *
-     * @param subscriptionId
-     *  The subscription ID to use to lookup a master pool
-     *
-     * @return
-     *  the master pool associated with the specified subscription.
-     */
-    Pool getMasterPoolBySubscriptionId(String subscriptionId);
-
-    /**
      * Retrieves a list consisting of all known master pools.
      *
      * @return
--- a/server/src/main/java/org/candlepin/resource/PoolResource.java
+++ b/server/src/main/java/org/candlepin/resource/PoolResource.java
@@ -33,12 +33,15 @@ import org.candlepin.model.Entitlement;
 import org.candlepin.model.Owner;
 import org.candlepin.model.OwnerCurator;
 import org.candlepin.model.Pool;
+import org.candlepin.model.Pool.PoolType;
 import org.candlepin.model.PoolFilterBuilder;
+import org.candlepin.model.SubscriptionsCertificate;
 import org.candlepin.resource.util.CalculatedAttributesUtil;
 import org.candlepin.resource.util.ResourceDateParser;
 
 import com.google.inject.Inject;
 
+import org.jboss.resteasy.annotations.providers.jaxb.DoNotUseJAXBProvider;
 import org.jboss.resteasy.annotations.providers.jaxb.Wrapped;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 import org.slf4j.Logger;
@@ -46,9 +49,11 @@ import org.slf4j.LoggerFactory;
 import org.xnap.commons.i18n.I18n;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 
+import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
@@ -239,8 +244,14 @@ public class PoolResource {
             throw new NotFoundException(i18n.tr(
                 "Entitlement Pool with ID ''{0}'' could not be found.", id));
         }
-
-        poolManager.deletePool(pool);
+        if (pool.getType() != PoolType.NORMAL) {
+            throw new BadRequestException(i18n.tr("Cannot delete bonus pools, as they are auto generated"));
+        }
+        if (pool.isCreatedByShare()) {
+            throw new BadRequestException(i18n.tr("Cannot delete shared pools, This should be triggered by" +
+                " deleting the share entitlement instead"));
+        }
+        poolManager.deletePools(Collections.singleton(pool));
     }
 
     @ApiOperation(notes = "Retrieve a CDN for a Pool", value = "getPoolCdn")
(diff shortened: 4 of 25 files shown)


server/src/main/java/org/candlepin/resource/PoolResource.java
```diff
@@ -228,7 +233,7 @@ public Pool getPool(@PathParam("pool_id") @Verify(Pool.class) String id,
             "Subscription Pool with ID ''{0}'' could not be found.", id));
     }
 
-    @ApiOperation(notes = "Remove a Pool", value = "deletePool")
+    @ApiOperation(notes = "Remove a Pool", value = "deleteSinglePool")
```
Ceiu: I don't understand what this denotes. Also, like the method name in CPM, the "single pool" bit is already expressed by the non-plural noun "pool" and the parameter list that accepts only one pool.

server/src/main/java/org/candlepin/resource/SubscriptionResource.java
```diff
@@ -150,9 +101,8 @@ protected SubscriptionsCertificate getSubscriptionCertificate(String subscriptio
     @Path("/{subscription_id}")
     @Produces(MediaType.APPLICATION_JSON)
     public Subscription getSubscription(@PathParam("subscription_id") String subscriptionId) {
-        Pool pool = this.getMasterPoolForSubscription(subscriptionId);
 
-        return this.poolManager.fabricateSubscriptionFromPool(pool);
+        throw new ResourceMovedException("pools/{pool_id}");
```
Ceiu: 👍

server/src/main/java/org/candlepin/controller/PoolManager.java
```diff
@@ -89,7 +89,7 @@
      *
      * @param pool
      */
-    void deletePool(Pool pool);
+    void deleteSinglePool(Pool pool);
```
vritant: undoing

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2288,7 +2283,11 @@ public boolean isManaged(Pool pool) {
...
         // pool is non-custom  -- however we decide to implement that in the future.
 
-        return pool != null && pool.getSourceSubscription() != null && !pool.getType().isDerivedType() &&
-            (pool.getUpstreamPoolId() != null || !this.config.getBoolean(ConfigProperties.STANDALONE, true));
+        return pool != null &&
+            pool.getSourceSubscription() != null &&
+            !pool.getType().isDerivedType() &&
+            !pool.isCreatedByShare() &&
+            (pool.getUpstreamPoolId() != null ||
+                !this.config.getBoolean(ConfigProperties.STANDALONE, true));
```
vritant: kk

common/src/main/java/org/candlepin/common/exceptions/ResourceMovedException.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.ws.rs.core.Response.Status;
+
+/**
+ * Represents an See Other (HTTP 303) error.
+ */
+public class ResourceMovedException extends CandlepinException {
+    String location = null;
+
+    public ResourceMovedException(String argLocation) {
+        super(Status.SEE_OTHER, "The resource has moved to the location:" + argLocation);
```
vritant: done

server/src/main/java/org/candlepin/resource/PoolResource.java
vritant: ack

vritant:
[[13] Resolve issue with multiple pools having the same source subscription](https://trello.com/c/icEqeXC1/478-13-resolve-issue-with-multiple-pools-having-the-same-source-subscription)

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1967,7 +1967,10 @@ public void deletePools(Collection<Pool> pools, Set<String> alreadyDeletedPools)
             }
 
             poolsToDelete.add(pool);
-            entitlementsToRevoke.addAll(pool.getEntitlements());
+            for (Entitlement e: pool.getEntitlements()) {
+                e.setDeletedFromPool(true);
+                entitlementsToRevoke.add(e);
+            }
```
Ceiu: This is a good change, and one that I've made myself (in a sense) in the deletePools refactor; but it's one that may have fallout. It might be worth investigating whether or not people will be expecting those events in this case. If they are, we're about to break them -- despite it being incorrect behavior.

server/src/main/resources/rules/rules.js
```diff
@@ -338,7 +337,7 @@ function get_pool_priority(pool, consumer) {
     }
 
     // Decrease the priority of shared pools slightly so that non-shared pools will get consumed first.
-    if (Utils.equalsIgnoreCase('true', pool.getAttribute(SHARE_DERIVED))) {
+    if (Utils.equalsIgnoreCase('true', pool.hasSharedAncestor.toString())) {
```
Ceiu: If this is a boolean value, why are we converting it to a string to check if it's true?

vritant: @Ceiu  LOL. fixed.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
vritant: @crog lets talk post scrum.

Ceiu:
After thinking on it a bit over the weekend, I feel that addressing CPM.deletePool and making it work properly should be done here. While I understand the point of not necessarily wanting to do a dive into every other area that touches it, we've address a potential shortcoming those other methods may not be expecting. And, if they *are* expecting it, they should be calling the right method.

This is one of those thing that will fall off our radars until we have time to do a major pass on the whole feature again, or we're bitten by another bug. I'd rather address it now, while we have a task driving focus on it rather than leaving it until the next time something goes wrong; either on performance or functionality.

vritant:
I disagree. I dont see why replacing deletePool cant be a different PR.  The intent of PR is not to remove ALL usages of deletePool here, but to find a replacement for the subscription resources, and to get rid of the ambiguous PoolCurator methods. Usually, if a PR can be split into chunks, thats better imho.

Also, what to do with deletePool needs some consideration, and is not entirely trivial imo. I dont mind doing it myself, but I dont see the need for that change to be in this PR, the intent here is simply to fix that ambiguity, and it's repercussions.

that being said, I totally agree about the risk of it falling off the radars. @mstead thoughts? is that something I should work on NOW but in a separate PR, or should I stick to this PR?

vritant:

mstead:
> I disagree. I dont see why replacing deletePool cant be a different PR. The intent of PR is not to remove ALL usages of deletePool here, but to find a replacement for the subscription resources, and to get rid of the ambiguous PoolCurator methods. Usually, if a PR can be split into chunks, thats better imho.
> 
> Also, what to do with deletePool needs some consideration, and is not entirely trivial imo. I dont mind doing it myself, but I dont see the need for that change to be in this PR, the intent here is simply to fix that ambiguity, and it's repercussions.
> 
> that being said, I totally agree about the risk of it falling off the radars. @mstead thoughts? is that something I should work on NOW but in a separate PR, or should I stick to this PR?

This PR does not explain the issue around deletePool that I could find other than chat about events. Could you please add some context so it is tracked here. What is is that we need to do here? Depending on the size of that, it might be worth a separate card, and if it is something that should get done ASAP, then we need to get that prioritized for the next sprint. If is is small, then we could roll it into this one.

Ceiu:
@mstead As far as I recall, the issue was that CPM.deletePool was not performing all of the cleanup associated with pool deletion -- it's little more than a pool deletion + attached entitlement revocation.

From what I recall of the discussions, my assumption was that the fix was going to be changing CPM.deletePool to be a wrapper method for CPM.deletePools (note: plural) rather than changing the one problematic situation to call CPM.deletePools instead. I feel like all we're doing is addressing one case when it is likely every caller of deletePool is broken.

mstead:
> @mstead As far as I recall, the issue was that CPM.deletePool was not performing all of the cleanup associated with pool deletion -- it's little more than a pool deletion + attached entitlement revocation.
> 
> From what I recall of the discussions, my assumption was that the fix was going to be changing CPM.deletePool to be a wrapper method for CPM.deletePools (note: plural) rather than changing the one problematic situation to call CPM.deletePools instead. I feel like all we're doing is addressing one case when it is likely every caller of deletePool is broken.

@Ceiu I agree. The wrapper approach seems reasonable to me, and unless I'm missing something, seems to be pretty straight forward and relatively small. In that case, I'm OK with rolling that fix into this PR as long as it is a separate commit.

vritant:
@mstead , let me elaborate on the two methods first. 
* While deletePool deletes a single pool and revokes all its entitlements, deletePools fetches all pools with the same subscription id of the pool passed as argument, and revokes all of those pools ( the entire hierarchy ) and their entitlements.
* current uses of deletePool:
  * CPM.deletePoolsForSubscriptions - basically does what CPM.deletePools does, so should probably be replaced by the new CPM.deletePools.
  * CPM.processPoolUpdates - used in a lot of places, but deletes a single pool that the rules mark for delete
  * CPM.removeAndDeletePoolsOnOtherOwners - at refresh time, delete a single pool if its a normal or bonus pool but on a different owner than the one we are currently refreshing.
  * Entitler.bindByPools - delete a single dev pool if it exists.
  * Entitler.getDryRun - delete a single dev pool if it exists.
  * OwnerManager.cleanupAndDelete - delete all pools of an owner, irrespective of subscription id.
  * UndoImports.toExecute - delete all managed pools of this owner irrespective of the subscription id
  * ConsumerRules.onConsumerDelete - delete user restricted pools.
  * PoolResource.deletePool ( deletes a single pool today, in this PR this behavior is changed to call deletePools ) 
  * SubscriptionResource.deleteSubscription - fetches all pools of a subscription and deletes them all, one at a time. again, something that CPM.deletePools does, so should just use that method instead of re-implementing the wheel.

I think we should put this in a separate PR for the following reasons:
* whether we should make deletePool a wrapper for deletePools, itself is arguable. there are some occasions ( like deleting a dev pool, or deleting all pools of an owner ( in the sharing world it might not be the same as deleting all pools of a subscription ), etc , where there is a use case for deleting pools independent of subscription ids. at the very least, this would need some discussion / analysis.
* If we do make deletePool a wrapper of deletePools, this would require more re factoring in some of the uses above ( ex: SubscriptionResource.deleteSubscription , CPM.deletePoolsForSubscriptions ) , which complicates this PR more than necessary, imo, thats a candidate for a separate PR

mstead:
> @mstead , let me elaborate on the two methods first.
> 
> While deletePool deletes a single pool and revokes all its entitlements, deletePools fetches all pools with the same subscription id of the pool passed as argument, and revokes all of those pools ( the entire hierarchy ) and their entitlements.
> current uses of deletePool:
> CPM.deletePoolsForSubscriptions - basically does what CPM.deletePools does, so should probably be replaced by the new CPM.deletePools.
> CPM.processPoolUpdates - used in a lot of places, but deletes a single pool that the rules mark for delete
> CPM.removeAndDeletePoolsOnOtherOwners - at refresh time, delete a single pool if its a normal or bonus pool but on a different owner than the one we are currently refreshing.
> Entitler.bindByPools - delete a single dev pool if it exists.
> Entitler.getDryRun - delete a single dev pool if it exists.
> OwnerManager.cleanupAndDelete - delete all pools of an owner, irrespective of subscription id.
> UndoImports.toExecute - delete all managed pools of this owner irrespective of the subscription id
> ConsumerRules.onConsumerDelete - delete user restricted pools.
> PoolResource.deletePool ( deletes a single pool today, in this PR this behavior is changed to call deletePools )
> SubscriptionResource.deleteSubscription - fetches all pools of a subscription and deletes them all, one at a time. again, something that CPM.deletePools does, so should just use that method instead of re-implementing the wheel.
> I think we should put this in a separate PR for the following reasons:
> 
> whether we should make deletePool a wrapper for deletePools, itself is arguable. there are some occasions ( like deleting a dev pool, or deleting all pools of an owner ( in the sharing world it might not be the same as deleting all pools of a subscription ), etc , where there is a use case for deleting pools independent of subscription ids. at the very least, this would need some discussion / analysis.
> If we do make deletePool a wrapper of deletePools, this would require more re factoring in some of the uses above ( ex: SubscriptionResource.deleteSubscription , CPM.deletePoolsForSubscriptions ) , which complicates this PR more than necessary, imo, thats a candidate for a separate PR

Given the above info, lets put this in another PR and create a card for it on the backlog. I don't the current PR to become massive. Let's keep it small.

vritant:
@Ceiu any outstanding issues on this PR?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
vritant: W.r.t. this issue, fwiw , I share @Ceiu 's concern in this case, and here are some facts about it.
* there are two implementations which delete pool and their entitlements, and in master today, in deletePool we send out events for the entitlement deletes as well as the pool deletes, but in deletePools we send out events only for the pool deletes, and considers the entitlement deletes redundant.
* Since this was an intentional implementation in deletePools, This PR makes candlepin consistent in this aspect, by skipping those events in deletePool.

Ceiu:
@vritant The only thing I see on here that I don't recall if/what we decided was the bit about entitlement revocation event changes. If that's okay, then I believe I'm good.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
vritant: Based on conversation with @barnabycourt  , This change is okay . @barnabycourt  please confirm.

Ceiu:
I'm ready to merge this if @barnabycourt can confirm @vritant's question above.

Ceiu:
ACK

PR #1696, opened 2017-08-17
Remove an unused backreference from IdentityCertificate.

barnabycourt:
This consumer backreference caused hibernate to load another copy of the consumer from the database whenever the IdentityCertificate was loaded even though lazy was specified. For example, a consumer loads the identity which reloads the consumer.

Back-port of #1686


barnabycourt:
The postgres failure was a false negative, a rebuild passed at: https://rhsm-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/candlepin/job/candlepin-pullrequest-spectests-hosted-postgresql/2308

Ceiu:
ACK

PR #1697, opened 2017-08-17
Make UpstreamConsumer lazy load

barnabycourt:
In order to prevent unnecessary queries for upstream consumer every time an organization is loaded mark the relationship as lazy.


Ceiu:
ACK

PR #1698, opened 2017-08-17
Fix regression in scoping of JsRunnerRequestCache

barnabycourt:
When the @CandlepinScopedRequest annotation was removed from the cache it
caused the JsRunnerRequestCacheProvider to always return a new instance.
This meant that the rules version was queried many times instead of once in a given request.


barnabycourt:
To test, turn on sql logging & watch the sql queries for the max cp rules version during a request.

barnabycourt:
I'm pulling this PR as there appears to be something deeper going on with CandlepinRequestScope

PR #1683, opened 2017-08-11
add unit test for fetching shared pools of a pool

vritant:

PR #1699, opened 2017-08-22
Make UpstreamConsumer lazy load

barnabycourt:
In order to prevent unnecessary queries for upstream consumer ever time an organization is loaded mark the relationship as lazy.

Back port of https://github.com/candlepin/candlepin/pull/1697

### DIFF
--- a/server/src/main/java/org/candlepin/model/Owner.java
+++ b/server/src/main/java/org/candlepin/model/Owner.java
@@ -31,6 +31,7 @@ import java.util.Set;
 import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Entity;
+import javax.persistence.FetchType;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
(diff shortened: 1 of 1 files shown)


barnabycourt:
retest this please

awood:
```
Failures:

  1) Standalone Virt-Limit Subscriptions should revoke guest entitlements when migration happens
     Failure/Error: @guest1_client.list_entitlements.length.should == 0

       expected: 0
            got: 1 (using ==)
     # ./server/spec/virt_spec.rb:220:in `block (2 levels) in <top (required)>'
```

I'm guessing this is an unrelated failure?  The postgres spec tests passed and the MySQL tests passed in #1700.  If you think this is a false positive, let me know and I'll merge it.

awood:
Looks like #1683 is having the same spec test failure.

barnabycourt:
@awood I think the test failure is unrelated to this particular PR but it is definitely a consistent issue in the 2.1 maintenance branch. I'm unable to duplicate it locally but we need to address it.

PR #1702, opened 2017-08-23
1484438: Fix regression in scoping of JsRunnerRequestCache

barnabycourt:
When the @CandlepinScopedRequest annotation was removed from the cache it
caused the JsRunnerRequestCacheProvider to always return a new instance.
This meant that the rules version was queried many times instead of once in a given request.

This is a backport of 1482937 & https://github.com/candlepin/candlepin/pull/1701


barnabycourt:
The mysql tests passed on a second run. See details at: https://rhsm-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/candlepin/job/candlepin-pullrequest-spectests-hosted-mysql/1889/

PR #1704, opened 2017-08-31
1419016: Fix registration issue when owner's autobind is disabled.

mstead:
Allow registration with autobind enabled key when owner
has disabled autobind.

When the disable owner's autobind feature was implemented
we blocked registration in this case, but this was in the end
not desirable.


mstead:
Needed some BZ rejuggling.

PR #1705, opened 2017-08-31
1487374: Fix registration issue when owner's autobind is disabled.

mstead:
Allow registration with autobind enabled key when owner
has disabled autobind.

When the disable owner's autobind feature was implemented
we blocked registration in this case, but this was in the end
not desirable.


mstead:
[[5] 1419016 Registration with auto-attach enabled activationkey fails when Owner has autoattach disabled](https://trello.com/c/GX1HDPlm/295-5-1419016-registration-with-auto-attach-enabled-activationkey-fails-when-owner-has-autoattach-disabled)

vritant:
@mstead only one nit-pick : typo in autoattchDisabledForOwner
looks good otherwise.

PR #1723, opened 2017-09-26
1482949: Limit consumer check-in time updates to specific calls

awood:
This patch adds a ConsumerCheckInFilter that updates a consumer's
check-in time only if the REST resource called has the
UpdateConsumerCheckIn annotation.

The following methods have been annotated:
* ConsumerResource.updateConsumer (PUT /consumers/{uuid})
* ConsumerResource.regenerateEntitlementCertificates (PUT
  /consumers/{uuid}/certificates)
* ConsumerResource.getEntitlementCertificates (GET
  /consumers/{uuid}/certificates)
* ConsumerResource.getEntitlementCertificateSerials (GET
  /consumers/{uuid}/certificates/serials)
* HypervisorResource.hypervisorUpdate (POST /hypervisors)
* HypervisorResource.hypervisorUpdateAsync (POST /hypervisors/{owner})

which should cover the cases of rhsmcertd check-ins, consumer refreshes,
and hypervisor check-ins.

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -919,6 +919,11 @@ class Candlepin
     get("/consumers/#{consumer_id}")
   end
 
+  def get_consumer_release(consumer_id=nil)
+    consumer_id ||= @uuid
+    get("/consumers/#{consumer_id}/release")
+  end
+
   def get_compliance(consumer_id=nil, on_date=nil)
     consumer_id ||= @uuid
     query = "/consumers/#{consumer_id}/compliance"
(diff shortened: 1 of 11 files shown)


awood:
[1482949 - Consumer.lastCheckin is updated more frequently than necessary](https://trello.com/c/TL2G7HJ2/548-1482949-consumerlastcheckin-is-updated-more-frequently-than-necessary)

barnabycourt:
sub-man always checks serials before getting certs, do we need to annotate both getEntitlementCertificates & getEntitlementCertificateSerials?

Ceiu:
This was verified to be functioning properly with the rspec tests provided by PR #1724

Ceiu:
@barnabycourt I would argue that since subman is an external source, we should leave it as-is if both endpoints are expected to result in an updated checkin time.

mstead:
> @barnabycourt I would argue that since subman is an external source, we should leave it as-is if both endpoints are expected to result in an updated checkin time.

I tend to agree.

kahowell:
retest this please

barnabycourt:
My concerns have been resolved & the rest call list has been acked by PM

Ceiu:
ACK

PR #1725, opened 2017-09-28
[M] Updated qpid rspec tests

Ceiu:
- Updated the qpid rspec tests to wait for messages and status
  changes rather than using static sleep timers
- Updated the CandlepinQpid wrapper object to make use of
  blocking calls, timeout configuration and configurable addresses
  and cert/key files

### DIFF
--- a/server/spec/qpid_spec.rb
+++ b/server/spec/qpid_spec.rb
@@ -4,11 +4,11 @@ require 'candlepin_scenarios'
 
 # To run this spec test it is necessary to have Qpid up and configured.
 #
-# In Jenkins environment, this shouldn't be a problem. The cp-test script 
+# In Jenkins environment, this shouldn't be a problem. The cp-test script
 # has -q switch that will run configure-qpid.sh during candlepin server
 # deploy.
 #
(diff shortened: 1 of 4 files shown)


server/spec/qpid_spec.rb
```diff
@@ -4,136 +4,198 @@
...
-     @cq.stop
-     start_tomcat
-     sleep 10
-     assert_mode("SUSPEND")
-     @cq.start
-     sleep 12
-     assert_mode("NORMAL")
+  it 'should start in SUSPEND mode without Qpid' do
+    def start_tomcat()
+       `sudo systemctl start tomcat || sudo supervisorctl start tomcat`
```
barnabycourt: This should check existence of supervisorctl & use that instead of systemctl. Otherwise, systemctl usage will raise dbus erros in the logs when run inside a container.

Lorquas: bashism for this:
```
type systemctl &> /dev/null && sudo systemctl start tomcat || sudo supervisorctl stop tomcat
```
there is probably a nicer lookin' ruby version of this

Ceiu: Done.

PR #1726, opened 2017-09-28
[M] Pool deleted events no longer contain old pool entities

Ceiu:
- Removed the old/previous pool from pool deleted events, as they
  were unused by event consumers and very expensive to serialize


wottop:
ack

PR #1727, opened 2017-09-29
1497248: Execute HypervisorUpdateJobs in the order they were received

cnsnyder:
This patch ensures that for a given owner only one HypervisorUpdateJob
will be run at a time. The next job (for the given owner) to execute will
be the job with oldest created date that is not in a Failed, Cancelled, or
Finished state.


cnsnyder:
Oops meant this for 2.1. Closing and reopening appropriately

PR #1733, opened 2017-10-02
1497288: Mark jobs canceled when shutting down

awood:
Jobs are tracked in both cp_job and in the Quartz tables.  Prior to this
patch, when Pinsetter shutdown, it deleted job detail information from
the Quartz tables but did not alter the job in cp_job.  Upon startup,
the UnpauseJob would attempt to restart the job found in cp_job but no
correlating record would be found in the Quartz tables.

This patch adds a line to mark a job as canceled in the cp_job table as
the job is purged from the Quartz tables.  It also invokes a JobCurator
method that removes orphaned jobs on start up although this measure is
primarily preventative.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
+++ b/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
@@ -57,6 +57,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


Ceiu:
What would the affect of this be in environments that have multiple CP instances running (hosted)? Do we risk killing jobs prematurely or unnecessarily?

awood:
I answered this in person, but I'll record it here for others and for posterity.  These additions should just keep `cp_job` in sync with the Quartz tables.  The line added in the shutdown method just insures that the job is deleted from both the Quartz tables and from `cp_job` when the node shuts down.  The line added in the start up method just reaps orphaned jobs which is something the SweepBarJob does periodically anyway.  It just makes sense to do it on start up because orphaned jobs have no use and just cause exceptions to get thrown (since there is a record in `cp_job` but not a corresponding record in the Quartz tables).

So, no, I don't see any additional risk here.  Orphan jobs are already purged periodically and removing the job from `cp_job` during shutdown just prevents the creation of more orphans (since the job is already being deleted from the Quartz scheduler).

Ceiu:
ACK

PR #1737, opened 2017-10-06
1474443 compliance.created events now use UUID for the 'consumerId' f…

nikosmoum:
…ield instead of an internal db id

### DIFF
--- a/server/src/main/java/org/candlepin/audit/EventFactory.java
+++ b/server/src/main/java/org/candlepin/audit/EventFactory.java
@@ -232,9 +232,13 @@ public class EventFactory {
 
     public Event complianceCreated(Consumer consumer,
         Set<Entitlement> entitlements, ComplianceStatus compliance) {
+        // Instead of an internal db id, compliance.created events now use
+        // UUID for the 'consumerId' and 'entityId' fields, since Katello
+        // is concerned only with the consumer UUID field. This is the first
+        // part of a larger piece of work to simplify Event consumption.
         return new Event(Event.Type.CREATED, Event.Target.COMPLIANCE,
-            consumer.getName(), principalProvider.get(), consumer.getOwner().getId(), consumer.getId(),
-            consumer.getId(), null, buildComplianceDataJson(consumer, entitlements, compliance), null,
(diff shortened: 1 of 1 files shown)


nikosmoum:
[1474443 - compliance.created messages use internal id for 'consumerId'](https://trello.com/c/QnOOT1lg/588-1474443-compliancecreated-messages-use-internal-id-for-consumerid)

server/src/main/java/org/candlepin/audit/EventFactory.java
```diff
@@ -233,7 +233,7 @@ public Event subscriptionExpired(Subscription subscription) {
     public Event complianceCreated(Consumer consumer,
         Set<Entitlement> entitlements, ComplianceStatus compliance) {
         return new Event(Event.Type.CREATED, Event.Target.COMPLIANCE,
-            consumer.getName(), principalProvider.get(), consumer.getOwner().getId(), consumer.getId(),
+            consumer.getName(), principalProvider.get(), consumer.getOwner().getId(), consumer.getUuid(),
             consumer.getId(), null, buildComplianceDataJson(consumer, entitlements, compliance), null,
```
cnsnyder: We'll need the second usage of consumer.getId() replaced with consumer.getUuid() as well.

Ceiu:
The change itself is fine, but a little odd. Definitely worth adding a comment explaining why we're using the UUID in place of the ID; especially since we can't just change the JSON output of the event at the moment.

Also, has any investigation been done about removing the Consumer object itself from the output? Such a change may coincide with a larger event rework, but we should start doing the investigation now as we're making changes here.

nikosmoum:
There has been some investigation by Barnaby I believe, with details in this [document](https://docs.google.com/document/d/1qcucpyMWpkg-b3Zi0hNDzPBCJYbn0RUbqcp4_2dW4eg/edit#), about Katello using only few specific fields, with part of the larger plan being to remove the Consumer object.

nikosmoum:
@cnsnyder I have altered the second call to use getUuid() too, and amended my original commit message.

barnabycourt:
@nikosmoum Please squash the two commits to one.

Ceiu:
ACK

PR #1741, opened 2017-10-09
[2.1] Updated an autobind test to use "skip" instead of "pending"

Ceiu:
- Updated an autobind spec test to use use "skip" rather than "pending"
  when checking if the test is running in hosted mode. This will
  prevent the test from still running and failing with an error when
  running in standalone mode.


wottop:
ack

PR #1742, opened 2017-10-10
Update of Vagrant defaults

jirihnidek:
* Run deploy script with -t option to import test data to database.

### DIFF
--- a/vagrant/README.md
+++ b/vagrant/README.md
@@ -55,7 +55,7 @@ to be stopped or you will have to edit the Vagrantfile and choose different port
 1. `./server/bin/deploy`
 
 ## Deploy with test data
-Candlepin Vagrant deployer deploy candlepin without any test data in database.
+The Candlepin Vagrant deployer will deploy candlepin without any test data in database.
 If you want to use candlepin e.g. for testing subscription-manager, then you
 will need to deploy server with different options:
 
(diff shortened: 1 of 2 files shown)


vagrant/roles/candlepin-user/defaults/main.yml
```diff
@@ -7,4 +7,4 @@ candlepin_user_home:  /home/vagrant
 candlepin_git_repo: https://github.com/candlepin/candlepin.git
 candlepin_checkout: /vagrant
 
-candlepin_deploy_args: "-g -a"
+candlepin_deploy_args: "-g -a -t"
```
barnabycourt: If someone doesn't want to deploy the test data, how would they do so?

jirihnidek: Good point. My change is overkill. I will only change `README.md` with instruction how to deploy with test data.

barnabycourt: That would be great.

mstead:
> This does not appear to update the instructions in the vagrant readme. Is it really the best choice to always deploy the test data by default as opposed to the default being a clean install & letting someone override to set up test data if they want it.
> 
> @mstead What is your take on the default behavior for the CP vagrant environment?

Agree with @barnabycourt  on this.

vagrant/README.md
```diff
@@ -54,6 +54,19 @@ to be stopped or you will have to edit the Vagrantfile and choose different port
 1. `buildr clean test=no package`
 1. `./server/bin/deploy`
 
+## Deploy with test data
+Candlepin Vagrant deployer deploy candlepin without any test data in database.
```
barnabycourt: "The Candlepin Vagrant deployer will deploy.."

PR #1743, opened 2017-10-10
[2.1] Optimized fetching/deleting pools by subscription IDs

Ceiu:
- Updated getPoolsBySubscriptionIds to pull pool IDs directly from
  the cp2_pool_source_sub table, then lookup pools from there
- Deleting consumers no longer triggers consumer status recalcuation
  nor dependent entitlement regeneration
- Added an index on the cp2_pool_source_sub.subscription_id column
  to help speed up subscription ID based lookups

### DIFF
--- a/server/src/main/java/org/candlepin/controller/OwnerManager.java
+++ b/server/src/main/java/org/candlepin/controller/OwnerManager.java
@@ -118,7 +118,10 @@ public class OwnerManager {
 
         for (Consumer consumer : consumers) {
             log.info("Removing all entitlements for consumer: {}", consumer);
-            poolManager.revokeAllEntitlements(consumer, revokeCerts);
+
+            // We're about to delete these consumers; no need to regen/dirty their dependent
+            // entitlements or recalculate status.
+            poolManager.revokeAllEntitlements(consumer, false);
         }
 
         // Actual consumer deletion is done out of the loop above since all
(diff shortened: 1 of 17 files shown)


Ceiu:
[[13] Refresh Performance: Improve select pools by subscription id query.](https://trello.com/c/je2dwvd6/576-13-refresh-performance-improve-select-pools-by-subscription-id-query)

server/src/main/java/org/candlepin/controller/OwnerManager.java
```diff
@@ -118,7 +118,7 @@ public void cleanupAndDelete(Owner owner, boolean revokeCerts) {
 
         for (Consumer consumer : consumers) {
             log.info("Removing all entitlements for consumer: {}", consumer);
-            poolManager.revokeAllEntitlements(consumer, revokeCerts);
+            poolManager.revokeAllEntitlements(consumer, false);
```
mstead: In the cases where we've made a concious decision to set true/false when revoking all ents, do you think it would make sense to add a comment as to why this is? Just a thought, and no worries if you don't agree. Just think it would be easier to understand when reading the code without having to go back to the commit later.

mstead:
Overall, the code looks good to me. However, as I mentioned, I toyed with very similar query changes a little while back and didn't see a big overall improvement though. Let me know what your findings are when you test this against your snapshot.

server/src/main/java/org/candlepin/controller/OwnerManager.java
Ceiu: The only argument I would have against more comments is avoiding pushing another change that will re-trigger the tests.

I can add a quick note to each of these stating that we're deleting the consumer anyway, so there's no value in regenerating entitlements and status.

PR #1744, opened 2017-10-10
[M] Optimized fetching/deleting pools by subscription IDs

Ceiu:
- Updated getPoolsBySubscriptionIds to pull pool IDs directly from
  the cp2_pool_source_sub table, then lookup pools from there
- Deleting consumers no longer triggers consumer status recalcuation
  nor dependent entitlement regeneration
- Added an index on the cp2_pool_source_sub.subscription_id column
  to help speed up subscription ID based lookups


Ceiu:
[[13] Refresh Performance: Improve select pools by subscription id query.](https://trello.com/c/je2dwvd6/576-13-refresh-performance-improve-select-pools-by-subscription-id-query)

PR #1745, opened 2017-10-11
1487758 Added index on cp_consumer.username field

nikosmoum:
In addition to liquibase changes to add the index I have
deleted useless and deprecated @Index and @ForeignKey hibernate
annotations from Consumer since they are DDL specific.


nikosmoum:
[[BZ 1487758]  Missing index on cp_consumer.username causes slow queries by username](https://trello.com/c/Oh2Q0rmp/592-bz-1487758-missing-index-on-cpconsumerusername-causes-slow-queries-by-username)

Ceiu:
ACK

Looks good. Just get the conflicts resolved and this should be good to merge.

PR #1747, opened 2017-10-16
Event Simplification: Removed oldEntity from Events

nikosmoum:
- Removed the oldEntity field from the Event class, also
removed setOldEntity/getOldEntity from EventBuilder and EventFactory.
- Made setEventData on EventBuilder public to be used instead
of setOldEntity when building Events. This method will probably be
re-used similarly when newEntity will be removed in the future and
will be used when creating CREATED/DELETED events. It will be
overloaded for creating MODIFIED events (which currently need both
old/new entity).
- Removed a couple of unused methods (consumerModified, ownerMigrated)

### DIFF
--- a/server/src/main/java/org/candlepin/audit/EventBuilder.java
+++ b/server/src/main/java/org/candlepin/audit/EventBuilder.java
@@ -40,47 +40,42 @@ public class EventBuilder {
         this.factory = factory;
 
         event = new Event(type, target, null, factory.principalProvider.get(),
-                null, null, null, null, null, null, null);
+                null, null, null, null, null, null);
     }
 
-    private void setEventData(Eventful entity) {
-        // Be careful to check for null before setting so we don't overwrite anything useful
-        if (entity instanceof Named && ((Named) entity).getName() != null) {
-            event.setTargetName(((Named) entity).getName());
-        }
-        if (entity instanceof Owned) {
-            Owner entityOwner = ((Owned) entity).getOwner();
-            if (entityOwner != null && entityOwner.getId() != null) {
-                event.setOwnerId(entityOwner.getId());
+    /*  Implementation note (2017/10/16):
+        This method is currently used instead of setOldEntity, since oldEntity was removed,
+        and it will also be used in the future instead of setNewEntity once newEntity is removed.
+        This is part of simplifying the data Events hold.
+     */
+    public EventBuilder setEventData(Eventful entity) {
+        if (entity != null) {
+            // Be careful to check for null before setting so we don't overwrite anything useful
+            if (entity instanceof Named && ((Named) entity).getName() != null) {
+                event.setTargetName(((Named) entity).getName());
             }
-        }
-        if (entity instanceof Entitlement) {
-            event.setReferenceType(Event.ReferenceType.POOL);
-            Pool referencedPool = ((Entitlement) entity).getPool();
-            if (referencedPool != null && referencedPool.getId() != null) {
-                event.setReferenceId(referencedPool.getId());
-            }
-        }
-        if ((String) entity.getId() != null) {
-            event.setEntityId((String) entity.getId());
-            if (entity instanceof ConsumerProperty) {
-                Consumer owningConsumer = ((ConsumerProperty) entity).getConsumer();
-                if (owningConsumer != null && owningConsumer.getId() != null) {
-                    event.setConsumerId(owningConsumer.getId());
+            if (entity instanceof Owned) {
+                Owner entityOwner = ((Owned) entity).getOwner();
+                if (entityOwner != null && entityOwner.getId() != null) {
+                    event.setOwnerId(entityOwner.getId());
                 }
             }
-        }
-    }
-
-    public EventBuilder setOldEntity(Eventful old) {
-        // Allow null, but don't do anything
-        if (old != null) {
-            // If the value is non-null and a value is set, we shouldn't allow it to continue
-            if (event.getType() == Type.CREATED) {
-                throw new IllegalArgumentException("You cannot set the old entity for a creation event");
+            if (entity instanceof Entitlement) {
+                event.setReferenceType(Event.ReferenceType.POOL);
+                Pool referencedPool = ((Entitlement) entity).getPool();
+                if (referencedPool != null && referencedPool.getId() != null) {
+                    event.setReferenceId(referencedPool.getId());
+                }
             }
-            setEventData(old);
-            event.setOldEntity(factory.entityToJson(old));
+            if ((String) entity.getId() != null) {
+                event.setEntityId((String) entity.getId());
+                if (entity instanceof ConsumerProperty) {
+                    Consumer owningConsumer = ((ConsumerProperty) entity).getConsumer();
+                    if (owningConsumer != null && owningConsumer.getId() != null) {
+                        event.setConsumerId(owningConsumer.getId());
+                    }
+                }
+            }
         }
         return this;
     }
--- a/server/src/main/java/org/candlepin/audit/Event.java
+++ b/server/src/main/java/org/candlepin/audit/Event.java
@@ -137,14 +137,6 @@ public class Event implements Persisted {
     @Enumerated(EnumType.STRING)
     private ReferenceType referenceType;
 
-    // Both old/new may be null for creation/deletion events. These are marked
-    // Transient as we decided we do not necessarily want to store the object
-    // state
-    // in our Events table. The Event passing through the message queue will
-    // still
-    // carry them.
-    @Transient
-    private String oldEntity;
     @Transient
     private String newEntity;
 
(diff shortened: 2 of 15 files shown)


nikosmoum:
[Event Simplification: Remove oldEntity from all messages/callers](https://trello.com/c/itPcHuT0/593-event-simplification-remove-oldentity-from-all-messages-callers)

kahowell:
performance ok

barnabycourt:
retest this pease

server/src/main/java/org/candlepin/audit/EventBuilder.java
```diff
@@ -40,47 +40,42 @@ public EventBuilder(EventFactory factory, Target target, Type type) {
...
-    private void setEventData(Eventful entity) {
-        // Be careful to check for null before setting so we don't overwrite anything useful
-        if (entity instanceof Named && ((Named) entity).getName() != null) {
-            event.setTargetName(((Named) entity).getName());
-        }
-        if (entity instanceof Owned) {
-            Owner entityOwner = ((Owned) entity).getOwner();
-            if (entityOwner != null && entityOwner.getId() != null) {
-                event.setOwnerId(entityOwner.getId());
+    /*  Implementation note (2017/10/16):
```
awood: 👏 for using ISO 8601 format

awood:

nikosmoum:
There's genuine spec test failures from these changes. I have a fix for a group of them, still trying to fix the rest.

nikosmoum:

Ceiu:
I've restarted the two failing spec tests to get attempt to get them to pass. Remind me on Monday to show you how to refire individual tests, so you don't need to do the full suite in cases like this.

server/src/main/java/org/candlepin/audit/EventBuilder.java
```diff
@@ -40,47 +40,43 @@ public EventBuilder(EventFactory factory, Target target, Type type) {
...
+                if (entity instanceof ConsumerProperty) {
+                    Consumer owningConsumer = ((ConsumerProperty) entity).getConsumer();
+                    if (owningConsumer != null && owningConsumer.getId() != null) {
+                        event.setConsumerId(owningConsumer.getId());
+                    }
+                }
             }
-            setEventData(old);
-            event.setOldEntity(factory.entityToJson(old));
+            factory.entityToJson(entity);
```
nikosmoum: @mstead Adding this line actually fixed the issue. setOldEntity() used to serialize the entity in order to set it on the oldEntity field, but actually the serialization serves some other additional purpose here (even thought we don't set the produced json string to anything!). Not sure exactly how this magic works, but I got the hint from this [comment](https://github.com/candlepin/candlepin/pull/1747/commits/3bb339522592e958f80f8f8bdac9e881c111a5b3#diff-4cbebfb7e395944b0d84bfc0a4a42cf5L1054).

nikosmoum:

server/src/main/java/org/candlepin/audit/EventBuilder.java
awood:
```java
    protected String entityToJson(Object entity) {
        String newEntityJson = "";
        // TODO: Throw an auditing exception here

        // Drop data on consumer we do not want serialized, Jackson doesn't
        // seem to care about XmlTransient annotations when used here:

        try {
            newEntityJson = mapper.writeValueAsString(entity);
        }
        catch (Exception e) {
            log.warn("Unable to jsonify: " + entity);
            log.error("jsonification failed!", e);
        }
        return newEntityJson;
    }
```

is what should be getting run by `entityToJson`.  I don't see why any side effects would happen in that code.  😕

nikosmoum: I know. It doesn't make sense. But it works. -_-

awood: What was the exact test that was failing before this change and passes now?  I don't like merging code that is magical in how it works.

nikosmoum: 4 test cases in hypervisor_check_in_spec.rb:
- should add consumer to unchanged when comparing empty guest id lists - async
- should add consumer to unchanged when same guest ids are sent - async
- should allow virt-who to update mappings - async
and one more

barnabycourt: We need to know why the forcing of the serialization fixes these issues. Is the entity later being detached from hibernate & then accessed?

nikosmoum: This serialization happens during a HypervisorUpdateJob transaction.
The error happens during a GET request that follows, to check the job's result data (and a consumer object is part of that data). So these are definitely two different hibernate sessions, and the customer is detached. Not sure yet why the serialization helps.

nikosmoum: Apparently, what happens is:
1) We do an asynchronous updateHypervisor request.
2) An UpdateHypervisorJob runs to do the task. Part of the result data of this job is a Consumer object.
3) During this task there is a specific case where sometimes there is nothing to update. If this is the case, the lazy-loaded collections on the Consumer never get loaded.
4) When the job finishes, it saves the result data in binary form in the cp_job.result_data table field (part of which is the Consumer object, without having its collections initialized).
5) When we do a GET /jobs/#{job_id}?result_data=true, the data (and the Consumer object) get fetched, deserialized from binary to Objects and before returning them their serialization to json is attempted by invoking the getters. The Consumer is detached at this point, so an exception is thrown. 

The problem did not exist before because during step 3, the whole object was serialized as 'oldEntity', forcing the collections to get initialized.

As a solution, I used Hibernate.initialize() to initialize the collections during step 3, before serializing into binary and saving to the db.

barnabycourt:
@nikosmoum what does "resultData" : "[cloaked]" mean in the job status?

nikosmoum:
@barnabycourt From what I understand, when we do a GET /jobs/#{job_id} (without further parameters), the result data of the job are not returned (they are cloaked, which means this 'cloaked' string is returned instead of them), and only the status is returned (FINISHED,WAITING, etc.). If we want the result data we need to add ?result_data=true as a param. Not sure if the 'cloaking' serves any other purpose.

barnabycourt:
@nikosmoum If the change is not added to pre-hydrate the methods. Why do the spec tests fail? There are two issues that I see:

1. The json in the job status is missing a bunch of data if it is not loaded via the additional Hibernate.initialize(...). Why is this?
1. At least one of the spec tests is failing even though it is only checking the length of lists not the details that are loaded by the Hibernate.initialize(...). Why?

nikosmoum:
@barnabycourt Not sure if you missed my comment where in the comment thread of the previous changes, which explains question #1:

Apparently, what happens is:

1. We do an asynchronous updateHypervisor request.
2. An UpdateHypervisorJob runs to do the task. Part of the result data of this job is a Consumer object.
3. During this task there is a specific case where sometimes there is nothing to update. If this is the case, the lazy-loaded collections on the Consumer never get loaded.
4. When the job finishes, it saves the result data in binary form in the cp_job.result_data table field (part of which is the Consumer object, without having its collections initialized).
5. When we do a GET /jobs/#{job_id}?result_data=true, the data (and the Consumer object) get fetched, deserialized from binary to Objects and before returning them their serialization to json is attempted by invoking the getters. The Consumer is detached at this point, so an exception is thrown.

The problem did not exist before because during step 3, the whole object was serialized as 'oldEntity', forcing the collections to get initialized. As a solution, I used Hibernate.initialize() to initialize the collections during step 3, before serializing into binary and saving to the db.

As for question #2, If I'm not misunderstanding you, don't the lists have to be loaded in order to access access their size? Which test is it specifically?

nikosmoum:
The 4 spec tests that fail are in hypervisor_check_in_spec.rb:

- should add consumer to unchanged when comparing empty guest id lists - async
- should add consumer to unchanged when same guest ids are sent - async
- should allow virt-who to update mappings - async
- should support multiple orgs reporting the same cluster - async

The general theme of the problem is:
- We do an hypervisorUpdateAsync request.
- A HypervisorUpdateJob runs to do that.
- In a specific case where there's nothing to update, we still have to save the job result data (an unchanged Consumer object). Due to not serializing the Consumer as an 'oldEntity' any more, its 3 collections don't get initialized.
- When HypervisorUpdateJob finishes and its unitOfWork is commited, a listener (PinsetterJobListener.java) is tasked with opening its own unitOfWork, taking the job result data, serializing them to binary and and saving them to the cp_job.resultdata column. This place is too late to force-populate the 3 collections because its a different transaction than the one HypervisorUpdateJob had.
- Later, we do a GET /jobs/#{job_id}?result_data=true, the data (and the Consumer object) get fetched, deserialized from binary to Objects, then the Objects serialized to json (and this is where the LazyInitializationException is thrown, because the getters of the collections are called in a detached state).

nikosmoum:
Squashed commits and resolved conflict.

PR #1748, opened 2017-10-16
Fixed an issue with CPM.deletePools

Ceiu:
- Fixed an exception that could occur in CPM.deletePools when attempting
  to delete pools which had entitlements which had derived pools


vritant:
any reproducer instructions to speeden up a review?

PR #1754, opened 2017-10-17
Delete missing products from activation keys while upgrading.

awood:
migration.


awood:
Github wasn't picking up my force pushes.  Reopened as #1756

PR #1756, opened 2017-10-17
1503176: Delete missing products from activation keys while upgrading.

awood:
migration.


awood:

wottop:
ack

PR #1757, opened 2017-10-17
1503179: Delete missing products from activation keys while upgrading.

awood:
migration.


wottop:
retest this please

wottop:
The failure is legit. When I ran owner_product_resource_spec locally, I see the same error.

awood:

awood:
> The failure is legit. When I ran owner_product_resource_spec locally, I see the same error.

I have no idea how this PR could be related to that failure.  This code is only run during a liquibase migration and, as far as I know, no migrations are performed during the spec test runs.  Also the CI environment is passing it now.

My theory is that the test itself fails intermittently and we would see the same behavior on the base branch.

wottop:
the test passes. ack.

PR #1761, opened 2017-10-17
[2.0] 1502807: bind with an empty string as POST body

vritant:
 ref: BZ: 1502807

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -800,11 +800,10 @@ class Candlepin
     post(path)
   end
 
-  def consume_pools(poolAndQuantities, params={})
+  def consume_pool_empty_body(pool_id, params={})
     uuid = params[:uuid] || @uuid
-    path = "/consumers/#{uuid}/entitlements?async=true"
-
-    return post(path, poolAndQuantities)
+    path = "/consumers/#{uuid}/entitlements?pool=#{pool_id}&quantity=1"
+    return post(path, " ")
   end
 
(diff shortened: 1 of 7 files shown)


vritant:

vritant:
@mstead : the mysql rspec failure is this:

> 15:50:04   1) Standalone Virt-Limit Subscriptions should revoke guest entitlements when migration happens
15:50:04      Failure/Error: @guest1_client.list_entitlements.length.should == 0
15:50:04 
15:50:04        expected: 0
15:50:04             got: 1 (using ==)
15:50:04      # ./server/spec/virt_spec.rb:220:in `block (2 levels) in <top (required)>'

this is a unrelated non-determinstic failure , and I reproduced this is the target branch candlepin-2.0-HOTFIX as well, after some tries.

vritant:
The postgres test failed with:

> 15:55:23 ERROR: for rspecjenkinscandlepincandlepinpullrequestspectestshostedpostgresql202_db_1  UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)
15:55:23 An HTTP request took too long to complete. Retry with --verbose to obtain debug information.
15:55:23 If you encounter this issue regularly because of slow network conditions, consider setting COMPOSE_HTTP_TIMEOUT to a higher value (current value: 60).
15:55:23 return value: 0

jenkins issues imo.

mstead:
NOTE: I've tested this PR with a broken version of subman/python-rhsm and it addresses the issue.

PR #1763, opened 2017-10-17
[M] 1503299: bind with an empty string as POST body

vritant:
 ref: BZ: 1502807

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -802,14 +802,10 @@ class Candlepin
     return post(path, qparams)
   end
 
-  def consume_pools(poolAndQuantities, params={})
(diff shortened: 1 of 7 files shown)


vritant:

vritant:

Ceiu:
I've removed the "ignore Jenkins failures" tag, as we have the means of re-running individual tests without re-running the entire suite (however clumsy it may be).

I've also restarted the test in question to see if it passes. If this is one that's still giving us problems, we should work to patch it sooner than later. I don't want to see us get back into a spot where we need to spend a week fixing Jenkins failures to get things passing reliably again.

mstead:
> I don't want to see us get back into a spot where we need to spend a week fixing Jenkins failures to get things passing reliably again.

Couldn't agree more!

PR #1770, opened 2017-10-18
[M] 1503578: Fixed an error that could occur while deleting subscriptions

Ceiu:

### DIFF
--- a/server/src/main/java/org/candlepin/controller/PoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/PoolManager.java
@@ -318,6 +318,18 @@ public interface PoolManager {
      */
     CandlepinQuery<Pool> getMasterPoolsForOwnerExcludingSubs(Owner owner, Collection<String> excludedSubs);
 
+    /**
+     * Retrieves the master pool associated with the specified subscription ID. If there is not a
+     * master pool asscoated with the given subscription, this method should return null.
+     *
+     * @param subscriptionId
+     *  The subscription ID to use to lookup a master pool
+     *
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/controller/PoolManager.java
```diff
@@ -318,15 +318,26 @@ Entitlement adjustEntitlementQuantity(Consumer consumer, Entitlement entitlement
...
+     * Retrieves the master pool associated with the specified subscription ID. If there is not a
+     * master pool asscoated with the given subscription, this method should return null.
+     *
+     * @param subscriptionId
+     *  The subscription ID to use to lookup a master pool
+     *
+     * @return
+     *  the master pool associated with the specified subscription.
+     */
+    Pool getMasterPoolBySubscriptionId(String subscriptionId);
```
awood: @vritant You removed this method in 9c448315c843c0a20167236af7591359d895613a because of sharing.  I think you and @Ceiu should have a chat.

Ceiu: Definitely. With subscription IDs still being used as global UUIDs, we need a way to link back to the pool in question without necessarily knowing which owner it belongs to. Alternatively, we could just drop this endpoint entirely (which removes the need for this entire PR), and get clients to use the delete pools endpoint. We're already halfway there with some of the deprecations, so that may be the better solution.
Overall though, I wouldn't mind chatting about this, if for no other reason than to understand more about the sharing functionality.

Thoughts?

PR #1771, opened 2017-10-18
[2.1] 1503694: Fixed an error that could occur while deleting subscriptions

Ceiu:

### DIFF
--- a/server/src/main/java/org/candlepin/resource/SubscriptionResource.java
+++ b/server/src/main/java/org/candlepin/resource/SubscriptionResource.java
@@ -32,6 +32,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xnap.commons.i18n.I18n;
 
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


barnabycourt:
Only failing tests are the bugzilla reference & the jenkins-rspec-hosted-mysql. 
The bugzilla reference issues have been fixed.

barnabycourt:
retest this please

Ceiu:

server/src/main/java/org/candlepin/resource/SubscriptionResource.java
```diff
@@ -219,19 +220,15 @@ public Response activateSubscription(
...
-            ++count;
-        }
-
-        if (count == 0) {
+        if (master == null) {
             throw new NotFoundException(
                 i18n.tr("A subscription with the ID \"{0}\" could not be found.", subscriptionId));
         }
+
+        this.poolManager.deletePools(Collections.singletonList(master), null);
```
awood: This is pretty minor, but would you mind switching this to 

```
this.poolManager.deletePools(Collections.singletonList(master))
```

so we don't have that superfluous `null` hanging around.  I think the functionality should be the same since the one parameter method just delegates to the two parameter method but provides the `null` as the second parameter.

Ceiu: But... extra method call overhead.

Changed.

PR #1772, opened 2017-10-18
[2.0] 1503731: Fixed an error that could occur while deleting subscriptions

Ceiu:


cnsnyder:
As this is not used in production, closing this for now.

PR #1773, opened 2017-10-19
[2.1] 1503164: Do not use schema version past what is available in RHEL 7.

awood:
Using a schema version newer than what is packaged in the Liquibase jar
file will cause Liquibase to attempt to access the new schema over http.
In disconnected installations, the lack of network connectivity will
cause the request to fail and Liquibase will fail to run in turn.

Notes on testing:

1. To test this item, you will need to be running Liquibase 3.1 and not 3.5 (which is the current version in Fedora).  You can grab the old Liquibase RPM from the Satellite 6 repository and downgrade.
2. You will need to turn off your machine's network connectivity.
3. Run a `bin/deploy -g`.  Liquibase will run and will fail because some of the changeset files are referencing the 3.5 schema.  The failure occurs when trying to run the `20170220102027-add-id-column-for-productcontent.xml` changeset.

@Ceiu Would you mind checking your editor?  Some of the files were originally using the 2.0 xsd and were then upgraded to 3.5 when you modified them last.  I'm wondering if your editor is trying to be "helpful".

### DIFF
--- a/server/src/main/resources/db/changelog/20170220102027-add-id-column-for-productcontent.xml
+++ b/server/src/main/resources/db/changelog/20170220102027-add-id-column-for-productcontent.xml
@@ -4,7 +4,7 @@
         xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
-        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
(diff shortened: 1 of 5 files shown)


Ceiu:
That wasn't my editor -- that was me. If 3.1 works, no issues, but I needed features from 3.x (don't recall which at this point), and even if you had the right version installed, if you specify an older version in the schema, Liquibase respects and disables that functionality (for better or worse).

wottop:
ack - mysql error is not related

awood:
[Bug 1503164 - Installation fails in disconnected environment](https://trello.com/c/Ne2ApMAr/613-bug-1503164-installation-fails-in-disconnected-environment)

PR #1774, opened 2017-10-20
1504710: Do not use schema version past what is available in RHEL 7.

awood:
Using a schema version newer than what is packaged in the Liquibase jar
file will cause Liquibase to attempt to access the new schema over http.
In disconnected installations, the lack of network connectivity will
cause the request to fail and Liquibase will fail to run in turn.

Notes on testing:

1.  To test this item, you will need to be running Liquibase 3.1 and not 3.5 (which is the current version in Fedora). You can grab the old Liquibase RPM from the Satellite 6 repository and downgrade.
1.  You will need to turn off your machine's network connectivity.
1.  Run a bin/deploy -g. Liquibase will run and will fail because some of the changeset files are referencing the 3.5 schema. The failure occurs when trying to run the 20170220102027-add-id-column-for-productcontent.xml changeset.


wottop:
ack

PR #1775, opened 2017-10-20
Add an index on consumer_id for cp_consumer_content_tags

awood:


awood:
[[5] Investigate if a consumer_id index would help for cp_consumer_content_tags](https://trello.com/c/hlYf9Mqj/601-5-investigate-if-a-consumerid-index-would-help-for-cpconsumercontenttags)

wottop:
ack

PR #1778, opened 2017-10-26
[2.0] Job Cleaner

wottop:


server/src/main/java/org/candlepin/model/JobCurator.java
```diff
@@ -199,6 +202,7 @@ public int cancelOrphanedJobs(List<String> activeIds) {
         return cancelOrphanedJobs(activeIds, 1000L * 60L * 2L); //2 minutes
     }
 
+    @Transactional
     public int cancelOrphanedJobs(List<String> activeIds, Long millis) {
         Date before = new Date(new Date().getTime() - millis);
         String hql = "update JobStatus j " +
```
barnabycourt: This should be splitting the activeIds into batches so as not to send too many in a single query.

PR #1782, opened 2017-10-27
[M] Job Cleaner

wottop:
Transactionalize delete jobs - Missing annotations were causing deletes not to occur
Batch lookups by id - List exceeding the limit

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/tasks/CancelJobJob.java
+++ b/server/src/main/java/org/candlepin/pinsetter/tasks/CancelJobJob.java
@@ -30,6 +30,7 @@ import org.quartz.SchedulerException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -55,7 +56,7 @@ public class CancelJobJob extends KingpinJob {
     @Override
     public void toExecute(JobExecutionContext ctx) throws JobExecutionException {
         Set<String> statusIds = new HashSet<String>();
-        List<JobStatus> jobsToCancel = null;
+        List<JobStatus> jobsToCancel = new ArrayList<JobStatus>();
         try {
             Set<JobKey> keys = pinsetterKernel.getSingleJobKeys();
             for (JobKey key : keys) {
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/candlepin/pinsetter/tasks/CancelJobJob.java
```diff
@@ -55,15 +56,17 @@ public CancelJobJob(JobCurator jobCurator, PinsetterKernel pinsetterKernel) {
...
                 statusIds.add(key.getName());
             }
 
             try {
-                jobsToCancel = jobCurator.findCanceledJobs(statusIds).list();
+                for (List<String> block : this.partition(statusIds)) {
+                    jobsToCancel.addAll(jobCurator.findCanceledJobs(new HashSet<String>(block)).list());
+                }
             }
             catch (HibernateException e) {
```
Ceiu: Few notes about this block:
- Do we still want/need the try/catch if we are properly partitioning the query?
- Instead of partitioning, we can use CPRestrictions.in in the findCanceledJobs method, which is probably faster in this case (fewer hidden queries, allows us to use scrolling should that be desired in the future), and eliminates the need to copy over everything associated with the partition method.
- If we opt to stay with the ID partitioning, we should just use Iterables directly, as I don't think it's a good idea to start exposing the in-block limitations outside of the curators. Moreover, we've moved away from using the constants for that value in favor of something configurable, which this current implementation doesn't support.
- Given that we only use the jobsToCancel within the first-level try-catch, we could just declare it within the block rather than outside of it

server/src/main/java/org/candlepin/pinsetter/tasks/SweepBarJob.java
```diff
@@ -72,9 +72,12 @@ public void toExecute(JobExecutionContext ctx) throws JobExecutionException {
             for (JobKey key : keys) {
                 statusIds.add(key.getName());
             }
-            int canceled = jobCurator.cancelOrphanedJobs(statusIds);
-            if (canceled > 0) {
-                log.info("Canceled " + canceled + " orphaned jobs");
+            int cancelled = 0;
+            for (List<String> block : this.partition(statusIds)) {
```
Ceiu: This partitioning should definitely move to the cancelOrphanedJobs method.

PR #1783, opened 2017-10-27
[M] 1506014: Optimized pool serialization and dependent entitlement lookup

Ceiu:
- Made additional optimizations to how marking dependent entitlements
  is performed, significantly reducing execution time in cases where
  many entitlements owned by many different consumers are updated
- Added the ProductCurator.hydratePoolProvidedProducts method, which
  hydrates a pool's provided and derived provided products and the
  product cache
- Updated CandlepinPoolManager.deletePools to use the new bulk pool
  hydration method to substantially reduce serialization time while
  passing pools and entitlements to the rules engine
- Modified CandlepinPoolManager.processPoolUpdates to bulk its
  merge and flush operations, entitlement fetching and pool
  existence checking
- Added a check to skip stacked pool updating for a consumer
  when there are no stacked pools to update
- Added a method for fetching unentitled stack derived pools to
  avoid recursive calls into deletePools
- Added a method for fetching maps of consumer IDs to stacked pool
  IDs to avoid repeatedly hitting the DB per consumer

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -575,15 +575,30 @@ public class CandlepinPoolManager implements PoolManager {
         return processPoolUpdates(poolEvents, updatedPools);
     }
 
-    protected Set<String> processPoolUpdates(
-        Map<String, EventBuilder> poolEvents, List<PoolUpdate> updatedPools) {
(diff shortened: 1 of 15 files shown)


Ceiu:
[[13] Perf Improvement: Pool.populateTransientProducts very inefficient](https://trello.com/c/WKYo3m1b/604-13-perf-improvement-poolpopulatetransientproducts-very-inefficient)

PR #1784, opened 2017-10-27
[M] Moved job ID partitioning to the curators

Ceiu:
- Moved job ID partitioning to the curators, as the jobs themselves
  should not be concerned with partitioning a collection due to
  query/database limitations


wottop:
ack

PR #1789, opened 2017-10-30
[2.0] Move job ID partitioning to the curators

wottop:


wottop:
spec test fail is not related to this change.

Ceiu:
ACK

PR #1790, opened 2017-10-30
Remove all reference to the Paginate annotation.

awood:
The Paginate annotation is no longer used since resource methods using
CandlepinQuery are now paging-enabled by default, essential enabling
paging universally.


Ceiu:
ACK

Note: the rspec-hosted-postgres status isn't updating due to an unknown error on the Jenkins side; but the only failure is the autobind not favoring non-shared pools, which is being fixed/addressed by #1791.

PR #1791, opened 2017-10-30
[M] Fixed an issue with shared pools not being deprioritized

Ceiu:
- Fixed an issue with the hasSharedAncestor property not being
  serialized, preventing the rules from deprioritizing shared
  pools during pool selection

### DIFF
--- a/server/src/main/resources/rules/rules.js
+++ b/server/src/main/resources/rules/rules.js
@@ -326,11 +326,13 @@ function get_mock_ent_for_pool(pool, consumer) {
 
 function get_pool_priority(pool, consumer) {
     var priority = 100;
+
     // use virt only if possible
     // if the consumer is not virt, the pool will have been filtered out
     if (Utils.equalsIgnoreCase(pool.getProductAttribute(VIRT_ONLY), "true")) {
         priority += 100;
     }
+
     // better still if host_specific
     if (pool.getAttribute(REQUIRES_HOST_ATTRIBUTE) !== null) {
         priority += 150;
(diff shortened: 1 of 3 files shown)


Ceiu:
An alternative to this fix would be changing the rules to use isCreatedByShare rather than exposing hasSharedAncestor to the JSON. Some guidance from @awood or @vritant would be welcome here.

awood:
> An alternative to this fix would be changing the rules to use isCreatedByShare rather than exposing hasSharedAncestor to the JSON. Some guidance from @awood or @vritant would be welcome here.

I would rather have the fix in `isCreatedByShare` to be honest.

server/src/main/resources/rules/rules.js
```diff
@@ -326,36 +326,43 @@ function get_mock_ent_for_pool(pool, consumer) {
...
     }
+
     // better still if host_specific
     if (pool.getAttribute(REQUIRES_HOST_ATTRIBUTE) !== null) {
         priority += 150;
     }
 
     // Decrease the priority of shared pools slightly so that non-shared pools will get consumed first.
-    if (pool.hasSharedAncestor) {
+    if (pool.createdByShare || pool.hasSharedAncestor) {
```
vritant: using hasSharedAncestor should be enough, so I would argue we should use **only** that, cause this use was the sole purpose of this field.

Ceiu: Gotta say, I'm not a fan of those two things. For the sake of this PR, I'll revert it, but going forward we need to pair that down a bit. Having two columns that describe the same thing with only slight differences is going to lead to problems down the road, and has already burned a bit of time trying to understand intent and significant difference.

awood: I think we need to preserve the distinction but decide on better names.  As I understand it:

* createdByShare is for a pool that is at the top of the sharing tree.
* hasSharedAncestor is true if any pool in a pool's lineage or if the pool itself is shared.

Perhaps something like `isShareRoot` and `isConsideredShare`?

PR #1792, opened 2017-11-01
Event Simplification: Added eventData to Events

nikosmoum:
This change is a precursor to removing newEntity.
- Added json blob field 'eventData' to Events which
holds a minimal dictionary of data, specific to events
'pool.created' and 'compliance.created', which are
the only data consumed currently from those on newEntity.
- ActivationListener now uses the new eventData field to
retrieve the subscriptionId for 'pool.created' events.

### DIFF
--- a/server/src/main/java/org/candlepin/audit/EventBuilder.java
+++ b/server/src/main/java/org/candlepin/audit/EventBuilder.java
@@ -18,6 +18,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.candlepin.audit.Event.Target;
 import org.candlepin.audit.Event.Type;
+import org.candlepin.common.exceptions.IseException;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.ConsumerProperty;
 import org.candlepin.model.Entitlement;
@@ -26,6 +27,8 @@ import org.candlepin.model.Named;
 import org.candlepin.model.Owned;
 import org.candlepin.model.Owner;
 import org.candlepin.model.Pool;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -36,6 +39,8 @@ import java.util.Map;
  *
  */
 public class EventBuilder {
+
+    private static final Logger log = LoggerFactory.getLogger(EventBuilder.class);
 
     private final EventFactory factory;
     private final ObjectMapper mapper;
@@ -90,7 +95,8 @@ public class EventBuilder {
                     event.setEventData(mapper.writeValueAsString(eventData));
                 }
                 catch (JsonProcessingException e) {
-                    e.printStackTrace();
+                    log.error("Error while building JSON for pool.created event.");
+                    throw new IseException("Error while building JSON for pool.created event.");
                 }
             }
         }
--- a/server/src/main/java/org/candlepin/audit/EventBuilder.java
+++ b/server/src/main/java/org/candlepin/audit/EventBuilder.java
@@ -40,7 +40,7 @@ public class EventBuilder {
         this.factory = factory;
 
         event = new Event(type, target, null, factory.principalProvider.get(),
-                null, null, null, null, null, null);
+                null, null, null, null, null, null, null);
     }
 
     /*  Implementation note (2017/10/16):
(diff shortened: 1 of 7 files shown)


nikosmoum:
[Event Simplification: Create json blob data field 'eventData' that is event-specific.](https://trello.com/c/2eJ8bPxk/619-event-simplification-create-json-blob-data-field-eventdata-that-is-event-specific)

server/src/main/java/org/candlepin/audit/EventBuilder.java
```diff
@@ -76,6 +76,14 @@ public EventBuilder setEventData(Eventful entity) {
                     }
                 }
             }
+            if (event.getTarget().equals(Target.POOL) && event.getType().equals(Type.CREATED)) {
+                StringBuilder eventDataBuilder = new StringBuilder()
```
Ceiu: I don't want to have us get in the habit of manually building JSON strings. We can use the JsonObjectBuilder from the javax.json package, or we can build maps and pass those to Jackson (ObjectMapper) directly.

server/src/test/java/org/candlepin/audit/EventBuilderTest.java
```diff
@@ -0,0 +1,53 @@
...
+    public void testSetEventDataForPoolSetsEventDataJson() {
+        Pool pool = mock(Pool.class);
+        eventBuilder = new EventBuilder(factory, Event.Target.POOL, Event.Type.CREATED);
+
+        when(pool.getSubscriptionId()).thenReturn("test-subscription-id");
+
+        String expectedEventData = "{\"subscriptionId\": \"test-subscription-id\"}";
+        Event event = eventBuilder.setEventData(pool).buildEvent();
+        assertEquals("The expected event data should have been" +
+            expectedEventData + ", but they weren't", expectedEventData, event.getEventData());
```
Ceiu: The description string is nice here, but you don't need to include the expected value in it. When there's a failure, it'll include the expected and actual values in the output itself.

server/src/main/java/org/candlepin/audit/EventFactory.java
```diff
@@ -210,14 +210,23 @@ public Event subscriptionExpired(Subscription subscription) {
 
     public Event complianceCreated(Consumer consumer,
         Set<Entitlement> entitlements, ComplianceStatus compliance) {
+        StringBuilder eventDataBuilder = new StringBuilder()
+            .append("{\"consumer\": ")
+            .append("{\"uuid\": \"")
```
barnabycourt: Why is this a nested field? It could just use consumer_uuid or some such as the key so users of the eventData don't need to dig further into the tree.

nikosmoum: I had a doubt about it, but I believe I misunderstood what the document said. I thought the status should be a top-level item, while the uuid not. It does make sense if they're both top-level, so I'll change it.

server/src/main/java/org/candlepin/audit/EventFactory.java
```diff
@@ -210,22 +213,23 @@ public Event subscriptionExpired(Subscription subscription) {
...
-            .append("}");
+        Map<String, String> eventData = new HashMap<String, String>();
+        eventData.put("consumer_uuid", consumer.getUuid());
+        eventData.put("status", compliance.getStatus());
+        String eventDataJson = "";
+        try {
+            eventDataJson = mapper.writeValueAsString(eventData);
+        }
+        catch (JsonProcessingException e) {
+            e.printStackTrace();
```
barnabycourt: Why is this eating the exception? Also, if it is to log the issue, it should be using our standard logger.

server/src/main/java/org/candlepin/audit/EventBuilder.java
```diff
@@ -76,6 +88,17 @@ public EventBuilder setEventData(Eventful entity) {
...
                 }
             }
+            if (event.getTarget().equals(Target.POOL) && event.getType().equals(Type.CREATED)) {
+                Map<String, String> eventData = new HashMap<String, String>();
+                eventData.put("subscriptionId", ((Pool) entity).getSubscriptionId());
+                try {
+                    event.setEventData(mapper.writeValueAsString(eventData));
+                }
+                catch (JsonProcessingException e) {
+                    log.error("Error while building JSON for pool.created event.");
```
Ceiu: The logger isn't clear on this, but you can append the exception in the output here by adding it as a parameter without having a matching substitution in the string. I.e.
```
log.error("Error while building JSON for pool.created event", e);
```

server/src/main/java/org/candlepin/audit/EventFactory.java
```diff
@@ -210,14 +214,25 @@ public Event subscriptionExpired(Subscription subscription) {
 
     public Event complianceCreated(Consumer consumer,
         Set<Entitlement> entitlements, ComplianceStatus compliance) {
+        Map<String, String> eventData = new HashMap<String, String>();
+        eventData.put("consumer_uuid", consumer.getUuid());
+        eventData.put("status", compliance.getStatus());
+        String eventDataJson = "";
+        try {
+            eventDataJson = mapper.writeValueAsString(eventData);
```
Ceiu: This is somewhat pedantic, but structurally there is no reason to have eventDataJson exist outside of the try/catch block, since this is an all-or-nothing style of operation.

```
try {
    String eventDataJson = mapper.writeValueAsString(eventData);

    return new Event(...);
}
catch (JsonProcessingException e) {
    ...
}
```

awood:

PR #1793, opened 2017-11-01
[M] 1500843: Fixed several issues with OwnerResource.updatePool

Ceiu:
- OwnerResource.updatePool now properly completes/merges the inbound
  pool object using data from the existing pool before passing it
  to the rules or refresh operations
- PoolRules now removes the VIRT_ONLY attribute from a pool if its
  product no longer carries the restriction during enrichment

### DIFF
--- a/server/src/main/java/org/candlepin/controller/PoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/PoolManager.java
@@ -67,12 +67,11 @@ public interface PoolManager {
     Pool convertToMasterPool(Subscription subscription);
 
     /**
-     * Updates the pools using the information stored in the given pool. Because
-     * the input subscription is used to lookup pools, the ID field must be set
-     * for this method to operate properly.
+     * Applies changes to the given pool to itself and any of its derived pools. This may result in
+     * a deletion of the pool if it has been expired as a result of the changes.
      *
      * @param pool
-     *        The pool to use for updating the associated pools
+     *  The pool to update
      */
     void updateMasterPool(Pool pool);
 
--- a/server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
+++ b/server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
@@ -87,6 +87,7 @@ public class PoolRules {
             log.debug("Increasing pool quantity for instance multiplier: {}", instanceMultiplier);
             result = result * instanceMultiplier;
         }
+
         return result;
     }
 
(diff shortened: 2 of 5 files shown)


Ceiu:
[Fix an issue with updatePools throwing exceptions](https://trello.com/c/pYNI1q8i/622-fix-an-issue-with-updatepools-throwing-exceptions)

Ceiu:
For reviewers: I made a judgement call on which attributes/fields can be modified by updatePool. If properties need to be swapped around to be locked or updateable, add a comment and we can get that done before this is merged.

server/src/main/java/org/candlepin/controller/PoolManager.java
```diff
@@ -66,14 +66,6 @@
...
     Pool convertToMasterPool(Subscription subscription);
 
-    /**
-     * Updates the pools using the information stored in the given pool. Because
-     * the input subscription is used to lookup pools, the ID field must be set
-     * for this method to operate properly.
-     *
-     * @param pool
-     *        The pool to use for updating the associated pools
-     */
```
awood: Why delete the JavaDoc here?  If it's wrong, I'd rather have it corrected than removed entirely.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -114,13 +115,15 @@ private long calculateQuantity(long quantity, Product product, String upstreamPo
...
-            masterPool.getProduct(), masterPool.getUpstreamPoolId()));
 
-        String virtOnly = masterPool.getProductAttributeValue(Product.Attributes.VIRT_ONLY);
+        masterPool.setQuantity(calculateQuantity(masterPool.getQuantity(), masterPool.getProduct(),
+            masterPool.getUpstreamPoolId()));
+
         // The following will make virt_only a pool attribute. That makes the
         // pool explicitly virt_only to subscription manager and any other
         // downstream consumer.
+        // TODO: Should we remove VIRT_ONLY from the pool if the attribute isn't present on the product?
```
awood: I'd like to resolve this question before we merge.  To me the answer seems like "yes" but I'd be interested to hear @vritant opinion.

server/src/main/java/org/candlepin/controller/PoolManager.java
Ceiu: Whoops. Removed it, intended to add it back, got distracted. I'll add that back on.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
vritant: yes, looked at the code, cant find a reason as to why that would be a problem.

Ceiu: Alright. Comment removed and replaced with a line that removes the pool attribute if the product attribute lookup fails.

Ceiu:
@awood @vritant How does the changeable pool properties look here? https://github.com/candlepin/candlepin/pull/1793/files#diff-f9dd0b027cd1a2f440a3d604658b7dbaR1234

As I said above, I'm not sure if this was the right set of properties to allow/disallow, so if anything stands out let me know.

awood:
> @awood @vritant How does the changeable pool properties look here? https://github.com/candlepin/candlepin/pull/1793/files#diff-f9dd0b027cd1a2f440a3d604658b7dbaR1234
> 
> As I said above, I'm not sure if this was the right set of properties to allow/disallow, so if anything stands out let me know.

It looked correct to me.

PR #1794, opened 2017-11-02
1508086: Need to lock manifest consumer when binding

wottop:
https://bugzilla.redhat.com/show_bug.cgi?id=1508086

### DIFF
--- a/server/spec/entitlement_resource_spec.rb
+++ b/server/spec/entitlement_resource_spec.rb
@@ -202,6 +202,36 @@ describe 'Entitlement Resource' do
 
   end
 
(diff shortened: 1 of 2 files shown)


vritant:
The only valid failure from this run was the BZ reference check. in this run that passes, so I am satisfied and merging this PR.

vritant:
I fixed the BZ btw, the target version needed to be set to 2.0

wottop:
[[BZ1508086] Consumer entitlement count does not match the ent count when manually summed from listing entitlements.](https://trello.com/c/vkoPt1Z2/621-consumer-entitlement-count-does-not-match-the-ent-count-when-manually-summed-from-listing-entitlements)

PR #1796, opened 2017-11-02
Event Simplification: Removed newEntity field

nikosmoum:
- Removed the newEntity field from Events
- EventBuilder's setEventData should now be used
for both created and deleted entities where
setNewEntity/setOldentity was used before.
- Overloaded setEventData as a utility to accept
both an old and a new entity for modification events.
- Removed verbose mode from LoggingListener and
the verbose mode property since it is no longer used.
- Added the eventData field to the normal audit log output.

### DIFF
--- a/server/src/main/java/org/candlepin/audit/EventBuilder.java
+++ b/server/src/main/java/org/candlepin/audit/EventBuilder.java
@@ -42,23 +42,41 @@ public class EventBuilder {
 
     private static final Logger log = LoggerFactory.getLogger(EventBuilder.class);
 
-    private final EventFactory factory;
     private final ObjectMapper mapper;
 
     private Event event;
 
     public EventBuilder(EventFactory factory, Target target, Type type) {
-        this.factory = factory;
         this.mapper = new ObjectMapper();
 
         event = new Event(type, target, null, factory.principalProvider.get(),
-                null, null, null, null, null, null, null);
+                null, null, null, null, null, null);
     }
 
-    /*  Implementation note (2017/10/16):
-        This method is currently used instead of setOldEntity, since oldEntity was removed,
-        and it will also be used in the future instead of setNewEntity once newEntity is removed.
-        This is part of simplifying the data Events hold.
+    /**
+     * Utility method used to pass in the old and new version of the modified
+     * entity for {@link Type#MODIFIED} events.
+     * @param oldEntity The target entity before modification
+     * @param newEntity The target entity after modification
+     * @return The builder object
+     */
+    public EventBuilder setEventData(Eventful oldEntity, Eventful newEntity) {
+        if (!event.getType().equals(Type.MODIFIED)) {
+            throw new IseException("This method is only for type MODIFIED Events.");
+        }
+        setEventData(oldEntity);
+        return setEventData(newEntity);
+    }
+
+    /**
+     * This method is used with any type of event and any target entity.
+     * <p>
+     * Note: For {@link Type#MODIFIED} events, it can be called twice consecutively
+     * to first pass in the original, and then the updated entity. Alternatively,
+     * {@link #setEventData(Eventful, Eventful)} can be used in the same use case.
+     * </p>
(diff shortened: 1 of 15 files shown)


nikosmoum:
[Event Simplification: Remove newEntity field from Events](https://trello.com/c/4bIL2S8Q/623-event-simplification-remove-newentity-field-from-events)

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -123,7 +123,6 @@ private ConfigProperties() {
      * will be filtered, meaning they will not enter HORNETQ.
      */
     public static final String AUDIT_FILTER_DEFAULT_POLICY = "candlepin.audit.filter.policy";
-    public static final String AUDIT_LOG_VERBOSE = "candlepin.audit.log_verbose";
```
awood: Can you verify that Candlepin will still start successfully with a configuration file that still has a `candlepin.audit.log_verbose` entry?  I just want to make sure that old configuration files don't cause a start-up error because they don't parse.

server/src/main/java/org/candlepin/audit/EventBuilder.java
```diff
@@ -42,23 +42,40 @@
...
+        if (!event.getType().equals(Type.MODIFIED)) {
+            throw new IseException("This method is only for type MODIFIED Events.");
+        }
+        setEventData(oldEntity);
+        return setEventData(newEntity);
+    }
+
+    /**
+     * This method is used with any type of event and any target entity.
+     * <br><br/>
```
Ceiu: Try to be consistent with your markup here. Also, using ```<p></p>``` will give you a more consistent paragraph-style break than relying on line height from line breaks.

server/src/test/java/org/candlepin/audit/EventBuilderTest.java
```diff
@@ -49,4 +55,14 @@ public void testSetEventDataForPoolSetsEventDataJson() {
...
         assertEquals(expectedEventData, event.getEventData());
     }
+
+    @Test
+    public void testSetEventDataForNonModifiedEventsThrowsIesException() {
+        Pool pool = mock(Pool.class);
+        eventBuilder = new EventBuilder(factory, Event.Target.POOL, Event.Type.CREATED);
+
+        exceptions.expect(IseException.class);
+        exceptions.expectMessage("This method is only for type MODIFIED Events.");
```
Ceiu: You can also accomplish this with the test annotation without setting the rule globally if you don't care about the exception details (which, generally speaking, we don't).

```
@Test(expected = IseException.class)
public void test...
```

nikosmoum: I'm aware of that option, but in this case I think we do care, because the same type of exception (IseException) can be thrown further down the call stack (since setEventData(entity, entity) calls setEventData(entity), which can throw IseException too). We don't want the test to pass for the wrong reason.

server/src/main/java/org/candlepin/config/ConfigProperties.java
nikosmoum: Verified that there's no startup-error when deploying candlepin with ```candlepin.audit.log_verbose=true``` in /etc/candlepin/candlepin.conf

nikosmoum:
retest this please

server/src/main/java/org/candlepin/config/ConfigProperties.java
awood: 💎 💰 👍

awood:

server/src/test/java/org/candlepin/audit/EventBuilderTest.java
Ceiu: That's fine. Just make sure you aren't testing implementation details. We don't want to have to rewrite our tests if the method's underlying implementation is changed.

Ceiu:
ACK

PR #1799, opened 2017-11-03
[M] Improve performance of CancelJobJob

mstead:
Batched up job deletion in the scheduler to improve performance
of CancelJobJob when there are a lot of stale quartz jobs to
cancel.


Ceiu:
ACK

mstead:
[CancelJobJob should delete from the scheduler in bulk.](https://trello.com/c/LX7VSjoT/624-canceljobjob-should-delete-from-the-scheduler-in-bulk)

PR #1801, opened 2017-11-03
[2.0] 1508086: Additional consumer locking on entitlment count

wottop:


vritant:
is this the only issue found where we lack locking?

PR #1807, opened 2017-11-08
[M] Added a per-org products migration validation Liquibase task

Ceiu:
  Liquibase to "debug"
  wrapper
  cpsetup

### DIFF
--- a/server/code/setup/cpdb
+++ b/server/code/setup/cpdb
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 #
-# Copyright (c) 2012 Red Hat, Inc.
+# Copyright (c) 2017 Red Hat, Inc.
 #
 # This software is licensed to you under the GNU General Public License,
 # version 2 (GPLv2). There is NO WARRANTY for this software, express or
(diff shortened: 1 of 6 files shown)


Ceiu:
Testing instructions:

- Deploy a version of Candlepin 0.9.x (0.9.54 is probably best) with test data
- Modify the database to cause a pool, subscription or environment to reference a bad product or content
- Check out this branch
- Do a clean build (compile) within the server branch
- Run cpdb from the server directory with the --validate flag. You should see an error
- Run cpdb again with the --validate and --verbose flags. You should see your error and a pile of other output statements

Ceiu:
[Port per-org-products validation to an independent script that can be run on command + write documentation describing its output.](https://trello.com/c/NvL1FX6b/631-port-per-org-products-validation-to-an-independent-script-that-can-be-run-on-command-write-documentation-describing-its-output)

server/code/setup/cpdb
```diff
@@ -243,9 +274,9 @@ if __name__ == "__main__":
             print("ERROR: Please provide a password for the Candlepin database user.")
             sys.exit(1)
 
-    if (not options.create and not options.update) or \
+    if (not options.create and not options.update and not options.validate) or \
         (options.create and options.update):
```
awood: Unless I'm misreading the logic here would allow for "--validate --create".  Which I think would be okay, but we'd need to move the `if options.validate` portion up to the top to insure the order is correct.

server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationValidationTask.java
```diff
@@ -0,0 +1,195 @@
...
+
+        // Impl Note:
+        // We don't need to check activation key products, since the migration just deletes these if they
+        // exist.
+
+        return passed;
+    }
+
+    /**
+     * Executes the multi-org upgrade task.
```
awood: We're not actually doing the upgrade here, right?  The comment on 170 & 171 is kind of confusing in that regard too.

server/src/main/resources/db/changelog/changelog-validate.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
```
awood: Wrong schema version here, I think.  We want to stick with 3.1 don't we?  Also missing a newline at the end of the file.

server/code/setup/cpdb
Ceiu: Fixed as per online discussion

server/src/main/java/org/candlepin/liquibase/PerOrgProductsMigrationValidationTask.java
Ceiu: Artifacts from copy/pasting the original code. Fixed.

server/src/main/resources/db/changelog/changelog-validate.xml
Ceiu: Fixed

PR #1809, opened 2017-11-09
[2.1] Added a per-org products migration validation Liquibase task

Ceiu:
  Liquibase to "debug"
  wrapper
  cpsetup


Ceiu:
[Port per-org-products validation to an independent script that can be run on command + write documentation describing its output.](https://trello.com/c/NvL1FX6b/631-port-per-org-products-validation-to-an-independent-script-that-can-be-run-on-command-write-documentation-describing-its-output)

PR #1811, opened 2017-11-09
[2.1] 1511615: mode change breaks pinsetter

vritant:
Pause and un pause scheduler, never pause all jobs. cause when we do, quartz pauses the thread group itself. Later it does not resume the async thread group correctly, and as a result no async jobs will run.
To test this PR, please test that:

 *   candlepin is able to schedule async jobs even when it is suspended and those jobs resume and run to completion once candlepin resumes to normal mode. ( as long as tomcat was not shut down inbetween those steps )
 *   cron jobs pause when candlepin moves to suspend mode and resume when candlepin resumes to normal mode.
 *   any scheduled async or cron jobs are cancelled when tomcat is shutdown while candlepin was in suspended mode.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
+++ b/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
@@ -134,6 +134,7 @@ public class PinsetterKernel implements ModeChangeListener {
     /**
      * Starts Pinsetter
      * This method does not return until the this.scheduler is shutdown
+     * Note: candlepin always starts in NORMAL mode.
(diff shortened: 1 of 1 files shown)


vritant:
[1510082: Determine the issue with virt-who checkins and seemingly stuck jobs on satellite ( http://projects.theforeman.org/issues/21110 )](https://trello.com/c/0SK8kujz/603-1510082-determine-the-issue-with-virt-who-checkins-and-seemingly-stuck-jobs-on-satellite-http-projectstheforemanorg-issues-21110)

vritant:
@awood since you reviewed and merged #1810 , could you take a look at this one too please?

PR #1812, opened 2017-11-09
[M] Prioritize cron jobs over async jobs

vritant:
To test PR:
* pause scheduler by making REST request:
POST https://localhost:8443/candlepin/jobs/scheduler ( request body: "false" )
* make any async job request :
curl -X PUT  -k -u admin:admin  "https://localhost:8443/candlepin/owners/admin/subscriptions" 

then use following sql to verify appropriate priorities:

> candlepin=# select distinct priority, trigger_group from qrtz_triggers;
 priority | trigger_group 
----------+---------------
        5 | async group
        7 | cron group

@

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
+++ b/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
@@ -328,10 +328,12 @@ public class PinsetterKernel implements ModeChangeListener {
         }
         try {
             for (JobEntry jobentry : pendingJobs) {
+                //Trigger cron jobs with higher priority than async ( default 5 )
                 Trigger trigger = newTrigger()
                     .withIdentity(jobentry.getJobName(), CRON_GROUP)
                     .withSchedule(cronSchedule(jobentry.getSchedule())
-                        .withMisfireHandlingInstructionDoNothing())
+                    .withMisfireHandlingInstructionDoNothing())
+                    .withPriority(7)
                     .build();
 
                 scheduleJob(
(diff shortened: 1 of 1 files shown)


vritant:
[[8] Prioritise cron jobs over async jobs so they get run first.](https://trello.com/c/XfJXKTu7/640-8-prioritise-cron-jobs-over-async-jobs-so-they-get-run-first)

vritant:

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
```diff
@@ -328,10 +328,12 @@ private void scheduleJobs(List<JobEntry> pendingJobs) {
         }
         try {
             for (JobEntry jobentry : pendingJobs) {
+                //Trigger cron jobs with higher priority than async ( default 5 )
                 Trigger trigger = newTrigger()
                     .withIdentity(jobentry.getJobName(), CRON_GROUP)
                     .withSchedule(cronSchedule(jobentry.getSchedule())
-                        .withMisfireHandlingInstructionDoNothing())
+                    .withMisfireHandlingInstructionDoNothing())
+                    .withPriority(3)
```
awood: I think this is backwards if we want the cron jobs to have priority.  The docs [state](http://www.quartz-scheduler.org/api/2.2.1/org/quartz/TriggerBuilder.html):

```
Set the Trigger's priority.  When more than one Trigger have the same fire time
the scheduler will fire the one with the highest priority first.
```
So the higher the number the higher the priority.

Looking through the [source](https://github.com/quartz-scheduler/quartz/blob/a9c39277fa1ca983d1bc384e81909f37f8fae221/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/StdJDBCConstants.java#L509) seems to confirm this since they have

```
+ "ORDER BY " + COL_NEXT_FIRE_TIME + " ASC, " + COL_PRIORITY + " DESC";
```

where higher a higher priority number would precede a lower one.

vritant: @awood my bad. I misread. fixed.

vritant:
@awood thanks for catching this! would have been really bad otherwise.

awood:

PR #1816, opened 2017-11-13
[M] Force re-trigger cron jobs

vritant:


vritant:
[[13] Implement a way to reset/fix stuck jobs without having to modify quartz DB directly.](https://trello.com/c/OghZlVPX/638-13-implement-a-way-to-reset-fix-stuck-jobs-without-having-to-modify-quartz-db-directly)

vritant:
[[13] Implement a way to reset/fix stuck jobs without having to modify quartz DB directly.](https://trello.com/c/OghZlVPX/638-13-implement-a-way-to-reset-fix-stuck-jobs-without-having-to-modify-quartz-db-directly)

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
```diff
@@ -237,35 +240,44 @@ else if (!isClustered()) {
...
+        catch (ClassNotFoundException e) {
+            throw new RuntimeException(e.getLocalizedMessage(), e);
+        }
+
+        String schedule = this.config.getString("pinsetter." + jobImpl + ".schedule", defvalue);
+
+        if (schedule != null && schedule.length() > 0) {
+            if (log.isDebugEnabled()) {
+                log.debug("Scheduler entry for " + jobImpl + ": " + schedule);
+                return schedule;
```
Ceiu: This only returns the schedule if debug output is enabled. We don't even really need the debug enabled check here.

Also, can we update the log statements throughout to use the substitution syntax?

vritant:
@Ceiu thanks for the catch!

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
```diff
@@ -237,35 +240,42 @@ else if (!isClustered()) {
...
             throw new RuntimeException(e.getLocalizedMessage(), e);
         }
-        scheduleJobs(pendingJobs);
+        catch (ClassNotFoundException e) {
+            throw new RuntimeException(e.getLocalizedMessage(), e);
+        }
+
+        String schedule = this.config.getString("pinsetter." + jobImpl + ".schedule", defvalue);
+
+        if (schedule != null && schedule.length() > 0) {
```
awood: Some comments here would be nice.  I had a bit of trouble reading this.

Something like

```
// Get the schedule string from the configuration if it's there; otherwise use the job's DEFAULT_SCHEDULE
String schedule = this.config.getString("pinsetter." + jobImpl + ".schedule", defvalue);
```

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
```diff
@@ -551,6 +562,37 @@ private void deleteAllJobs() {
...
+        Set<TriggerKey> cronTriggerKeys = null;
+        try {
+            cronTriggerKeys = scheduler.getTriggerKeys(
+                GroupMatcher.triggerGroupEquals(PinsetterKernel.CRON_GROUP));
+            TriggerKey key = null;
+            Iterator<TriggerKey> keysTrigger = cronTriggerKeys.iterator();
+            while (key == null && keysTrigger.hasNext()) {
+                TriggerKey current = keysTrigger.next();
+                if (current.getName().contains(taskName)) {
+                    key = current;
```
awood: If I'm reading this correctly, we're just looping through all the cronTriggerKeys looking for one that matches `taskName`.  Are we expecting more than one Key to match?  If not, we can add a `break`.  If so, is the order of the Keys deterministic?

server/src/main/java/org/candlepin/resource/JobResource.java
```diff
@@ -186,46 +185,68 @@ public SchedulerStatus setSchedulerStatus(boolean running) {
...
+    private Class getCronJobClass(String task) {
         String className = "org.candlepin.pinsetter.tasks." + task;
         try {
-            boolean allowed = false;
             for (String permissibleJob : ConfigProperties.DEFAULT_TASK_LIST) {
                 if (className.equalsIgnoreCase(permissibleJob)) {
-                    allowed = true;
-                    // helps ignore case
-                    className = permissibleJob;
+                    return Class.forName(permissibleJob);
```
awood: If this fails, it will throw a ClassNotFoundException, right?  I don't really see how the code would ever even make it to the `throws` line you have at the end of the method.

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
vritant: @awood there is comment on top of the method. does that suffice?

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
vritant: @awood  we do not expect more than one key to match.
Also, this is a while loop which would stop immediately after _key_ is assigned. I dont like to use _break_ so I went this way for the same purpose. should I add a comment to specify this information?

server/src/main/java/org/candlepin/resource/JobResource.java
vritant: @awood lets talk about this online.

vritant:
@awood  we use scheduler.rescheduleJob, which deletes the current trigger using the trigger key and then creates a new one.

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
vritant: @awood added the missing *

server/src/main/java/org/candlepin/resource/JobResource.java
vritant: talked with @awood , no changes necessary.

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
Ceiu: What is jobImpl here? Is it the (short) name of the job? Fully-qualified class name?

barnabycourt:
retest this please

awood:
retest this please

barnabycourt:
@awood @Ceiu If both of you have approved this change, is it ready to be merged?

PR #1818, opened 2017-11-14
[2.0] 1512879: Force re-trigger cron jobs

vritant:
* Certain cron jobs are getting stuck in BLOCKED state and requires a manual update via SQL.
* This API destroys current trigger and recreates it with a clean state.
* If we need the job to run immediately, we should continue to use the exsisting jobs/{}/schedule API.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
+++ b/server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
@@ -31,6 +31,7 @@ import org.candlepin.model.CandlepinModeChange.Mode;
 import org.candlepin.model.JobCurator;
 import org.candlepin.pinsetter.core.model.JobStatus;
 import org.candlepin.pinsetter.tasks.CancelJobJob;
+import org.candlepin.pinsetter.tasks.KingpinJob;
 import org.candlepin.util.PropertyUtil;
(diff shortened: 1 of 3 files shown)


vritant:
[[13] Implement a way to reset/fix stuck jobs without having to modify quartz DB directly.](https://trello.com/c/OghZlVPX/638-13-implement-a-way-to-reset-fix-stuck-jobs-without-having-to-modify-quartz-db-directly)

server/src/main/java/org/candlepin/pinsetter/core/PinsetterKernel.java
```diff
@@ -235,35 +238,44 @@ else if (!isClustered()) {
...
+        catch (ClassNotFoundException e) {
+            throw new RuntimeException(e.getLocalizedMessage(), e);
+        }
+
+        String schedule = this.config.getString("pinsetter." + jobImpl + ".schedule", defvalue);
+
+        if (schedule != null && schedule.length() > 0) {
+            if (log.isDebugEnabled()) {
+                log.debug("Scheduler entry for " + jobImpl + ": " + schedule);
+                return schedule;
```
Ceiu: Same issue as the other branch here -- the debug enabled check is wrapping the method output.

PR #1819, opened 2017-11-15
1504105: Capture consumer activation key settings before attaching

wottop:


wottop:
[Bug 1504105 - When provisioning a host the release version is not being honored when defined in the activation key.](https://trello.com/c/EHJP3sal/663-bug-1504105-when-provisioning-a-host-the-release-version-is-not-being-honored-when-defined-in-the-activation-key)

wottop:
rspec fail is unrelated. should not be an issue.

wottop:
To test:

Make an activation key that uses release:
curl -k -u admin:admin -H "Content-Type: application/json" -X POST -d @act_key_release.json https://localhost:8443/candlepin/owners/admin/activation_keys

Import a zip file that contains a RHEL (69) subscription:
curl -k -u admin:admin -F upload=@2017-01-09.zip https://localhost:8443/candlepin/owners/admin/imports

Run the good/bad regstration depending on before/after.Good will always work, bad will only work after the fix:
curl -k -u admin:admin 'https://localhost:8443/candlepin/consumers?owner=admin&activation_keys=release-test' -X POST -H "Content-Type: application/json" -d @bad_json.json | grep releaseVer

The files can be found here:
https://drive.google.com/open?id=10WGyTN51t0_hBP1hDvRMV4voNydxf7YY

Lorquas:
retest this please

server/src/main/java/org/candlepin/resource/util/ConsumerBindUtil.java
```diff
@@ -86,6 +90,7 @@ public void handleActivationKeys(Consumer consumer, List<ActivationKey> keys)
             handleActivationKeyOverrides(consumer, key.getContentOverrides());
             handleActivationKeyRelease(consumer, key.getReleaseVer());
             keySuccess &= handleActivationKeyServiceLevel(consumer, key.getServiceLevel(), key.getOwner());
+            consumerCurator.update(consumer);
```
awood: Don't we just want to update the consumer once?  I.e. outside the for loop?

wottop: We need this here because either bind will lockAndLoad the consumer and wipe out the settings that were transferred from the activation key.

awood: Got it.  Didn't realize the `handleActivationKey...` methods had side-effects.

wottop:
retest this please

PR #1821, opened 2017-11-16
[2.1] Zanata translations

vritant:
* changed zanata project version
* followed following process for updating to and from zanata:

1. buildr gettext:extract
2. buildr gettext:merge
3. zanata push
4. zanata pull
5. buildr validate_translation:fix


vritant:
[Determine what net new translations are were added to 0.9.54 and are missing from zanata so we can determine if perusing a sat 6.2 z-stream is required.](https://trello.com/c/CtFMvzkW/660-determine-what-net-new-translations-are-were-added-to-0954-and-are-missing-from-zanata-so-we-can-determine-if-perusing-a-sat-62)

PR #1824, opened 2017-11-22
Add -q option when checking for database existence.

awood:
Newer versions of PostgreSQL print out informational messages when psql
is invoked.  For example, if a user has "\pset pager off" set in their
.psqlrc file, the psql command will print a notice about this setting to
stdout.  This notice causes the conditional we use to check to see
whether the candlepin database exists or not to fail.  Consequently, the
database is not dropped and the subsequent createdb command fails since
the database already exists.  The -q option suppresses these
informational messages.


wottop:
ack, waiting for Jenkins

PR #1825, opened 2017-11-23
[M] 1514484: entitlements are regenerated on content change of provided products

vritant:
1514484: entitlements are regenerated on content change of provided products
* a previous commit had changed subscription resolution to no longer resolve product or content,
  this change undoes that so we can test this.
* added the spec test to test that refresh after content and product change does indeed regenerate ent

### DIFF
--- a/server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
+++ b/server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
@@ -555,6 +555,19 @@ public class PoolRules {
         return incomingProvided;
     }
 
+    private boolean changedProductsInSet(Set<Product> products, Map<String, Product> changedProducts) {
+
+        if (products != null && changedProducts != null) {
+            for (Product product : products) {
+                if (product != null && changedProducts.get(product.getId()) != null) {
+                    return true;
+                }
+            }
+
+        }
+        return false;
+    }
+
     private boolean checkForChangedProducts(Product incomingProduct, Set<Product> incomingProvided,
         Pool existingPool, Map<String, Product> changedProducts) {
 
@@ -569,8 +582,12 @@ public class PoolRules {
             !currentProvided.equals(incomingProvided);
 
         // Check if the existing product is in the set of changed products
-        if (!productsChanged && changedProducts != null && pid != null) {
-            productsChanged = (changedProducts.get(pid) != null);
+        if (!productsChanged && changedProducts != null) {
+            if (pid != null) {
+                productsChanged = (changedProducts.get(pid) != null);
+            }
+            productsChanged = productsChanged ||
+                changedProductsInSet(incomingProvided, changedProducts);
         }
 
         if (productsChanged) {
@@ -589,7 +606,7 @@ public class PoolRules {
             productsChanged = !pool.getDerivedProduct().getId()
                 .equals(existingPool.getDerivedProduct().getId());
 
-            productsChanged = productsChanged ||
+            productsChanged |=
                 (changedProducts != null && changedProducts.containsKey(pool.getDerivedProduct().getId()));
         }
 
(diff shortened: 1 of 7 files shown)


Lorquas:
retest this please

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -555,6 +555,21 @@ private void syncBranding(Pool pool, Pool existingPool) {
         return incomingProvided;
     }
 
+    private boolean changedProductsInSet(Set<Product> products,
+        Map<String, Product> changedProducts) {
+
+        if (products != null && changedProducts != null) {
+            for (Product product: products) {
```
Ceiu: Checkstyle failure here: need a space between product and the colon. Should also pull the method declaration into one line as well.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -555,6 +555,21 @@ private void syncBranding(Pool pool, Pool existingPool) {
...
+                if (product != null && product.getId() != null &&
```
Ceiu: You can skip the null check on the ID here. If it's null, it won't be found in the products list, nor will it cause an error.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -569,8 +584,12 @@ private boolean checkForChangedProducts(Product incomingProduct, Set<Product> in
             !currentProvided.equals(incomingProvided);
 
         // Check if the existing product is in the set of changed products
-        if (!productsChanged && changedProducts != null && pid != null) {
-            productsChanged = (changedProducts.get(pid) != null);
+        if (!productsChanged && changedProducts != null) {
+            if (pid != null) {
+                productsChanged = (changedProducts.get(pid) != null);
+            }
+            productsChanged = productsChanged ||
```
Ceiu: We can condense this whole block down to a (wrapped?) one-liner:

```
productsChanged |= (pid != null && changedProducts.get(pid) != null) || changedProductsInSet(...);
```

If you opt not to change it, then you need to fix the bug on line 589 which could revert the previous state of productsChanged if the declaration is true on 581 and the pid check is false on 589.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -607,7 +626,8 @@ private boolean checkForChangedDerivedProducts(Pool pool, Pool existingPool,
             }
         }
 
-        productsChanged = productsChanged || !currentProvided.equals(incomingProvided);
+        productsChanged = productsChanged || !currentProvided.equals(incomingProvided) ||
```
Ceiu: Not required, but you can use |= here.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
vritant: are you sure there is a bug?

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
vritant: looks like our checkstyle is fine with these changes. I changed it anyway

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
vritant: im only 50% sure on this, but during debugging I *thought* I saw some null objects. I am reluctant to try and prove it, may I ask what is the harm in an additional check or two even if that is not the case?

vritant:
@Ceiu changes done

Ceiu:
ACK

PR #1827, opened 2017-11-24
[2.1] 1514483: entitlements are regenerated on content change of provided products

vritant:
1514483: entitlements are regenerated on content change of provided products

* a previous commit had changed subscription resolution to no longer resolve product or content,
  this change undoes that so we can test this.
* added the spec test to test that refresh after content and product change does indeed regenerate ent

### DIFF
--- a/server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
+++ b/server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
@@ -547,6 +547,19 @@ public class PoolRules {
         return incomingProvided;
     }
 
+    private boolean changedProductsInSet(Set<Product> products, Map<String, Product> changedProducts) {
+
+        if (products != null && changedProducts != null) {
+            for (Product product : products) {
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -561,8 +576,12 @@ private boolean checkForChangedProducts(Product incomingProduct, Set<Product> in
             !currentProvided.equals(incomingProvided);
 
         // Check if the existing product is in the set of changed products
-        if (!productsChanged && changedProducts != null && pid != null) {
-            productsChanged = (changedProducts.get(pid) != null);
+        if (!productsChanged && changedProducts != null) {
+            if (pid != null) {
+                productsChanged = (changedProducts.get(pid) != null);
```
Ceiu: Potential issue here. See the PR for master for more details.

Ceiu:
ACK

PR #1829, opened 2017-11-29
[M] 1500837: Fixed an exception that could occur when updating pools

Ceiu:
- Fixed an NPE that could occur when calculating the effective
  quantity while running pool rules or subscription reconciliation


Ceiu:
This bug has actually already been addressed from an endpoint context. This PR just adds some more sanity checks on the quantity and multiplier usages, as it's possible for them to be null, which would trigger a NPE while unboxing.

Ceiu:
[Bug 1500837 - Runtime Error null at org.candlepin.policy.js.pool.PoolRules.createAndEnrichPools:116](https://trello.com/c/k9YuyV0s/671-bug-1500837-runtime-error-null-at-orgcandlepinpolicyjspoolpoolrulescreateandenrichpools116)

PR #1830, opened 2017-12-01
1515093: Set the KeyPair object to lazy load.

awood:
Joining on the cp_key_pair table can take hundreds of milliseconds in
tables with many rows.  The KeyPair object is not even used in most
cases, so switching it to lazy loading will shave some time off our
queries at the expense of an additional query in the cases where we
actually need the KeyPair.


awood:
[[8] BZ 1517848 Investigate whether there is a quick win by making Consumer.keyPair lazy load.](https://trello.com/c/SqCD4sl6/683-8-bz-1517848-investigate-whether-there-is-a-quick-win-by-making-consumerkeypair-lazy-load)

wottop:
ack

PR #1831, opened 2017-12-01
[0.9.54] 1519900: Added a per-org products migration validation Liquibase task

Ceiu:
  Liquibase to "debug"
  wrapper
  cpsetup

### DIFF
--- a/server/code/setup/cpdb
+++ b/server/code/setup/cpdb
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 #
-# Copyright (c) 2012 Red Hat, Inc.
+# Copyright (c) 2017 Red Hat, Inc.
 #
 # This software is licensed to you under the GNU General Public License,
 # version 2 (GPLv2). There is NO WARRANTY for this software, express or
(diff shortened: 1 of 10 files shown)


Ceiu:
This is a backport of the updated cpdb validation and logging functionality from 2.0, including the validation task for the per-org products migration.

Ceiu:
[Back port cpdb --validate changes to 0.9.54](https://trello.com/c/kdp6wF29/690-back-port-cpdb-validate-changes-to-0954)

mstead:
@Ceiu Please make sure that this PR addresses the concerns in:
https://bugzilla.redhat.com/show_bug.cgi?id=1519904

Leave them a comment as to what they can expect and some details around what the script will output if the migration has already been attempted and what they should do to find the validation errors in this case.

mstead:
@Ceiu adding this comment to track what we talked about.

'cpdb --validate' will not run the changeset once it has run successfully. We should make the changeset run each time UNLESS we have already migrated.

Ceiu:
Added "runAlways" to the changeset, as the preconditions are what we care about here.

PR #1834, opened 2017-12-04
Update zanata URL (2.0)

kahowell:
We're now using a different zanata instance for translations.
Also set the version to candlepin-2.0-HOTFIX


Ceiu:
ACK

PR #1836, opened 2017-12-04
Update zanata URL (0.9.51)

kahowell:
We're now using a different zanata instance for translations.
Also set the version to candlepin-0.9.51-HOTFIX


Ceiu:
ACK

PR #1839, opened 2017-12-06
[M] Forward port of Liquibase migration validation improvements

Ceiu:
- The per-org products migration validation Liquibase task now always
  runs, even if it has previously passed, so long as the preconditions
  succeed
- Improved the output message when the precondition fails and causes
  the validation task to not be run


wottop:
ack

PR #1849, opened 2017-12-12
Misc. validations on consumer registration. 

vritant:
found during DTO work, insignificant validations, committed separately to have an independent conversation

* in the create share consumer case if you don't provide and empty collection for installed products, we get an NPE. fixed now.
* today if you provide an id-cert in the POST register request, candlepin accepts it. this should not be allowed, each consumer should have a unique id cert generated by candlepin.
* today non-share consumers are allowed to specify recipient owners which does not make sense.
* we also accept entitlement count and persist it for some reason, intuitively that does not make sense.


wottop:
retest this please

wottop:
Fails checkstyle

wottop:
ack

PR #1850, opened 2017-12-12
[M] 1502193: Lock update consumer to avoid constraint violation with hypervisor checkin

wottop:
To test, run the new spec test against master. You will be able to see the consumerUpdate failure.
The added locking will keep the violation from occurring.

### DIFF
--- a/server/spec/hypervisor_check_in_spec.rb
+++ b/server/spec/hypervisor_check_in_spec.rb
@@ -902,4 +902,26 @@ describe 'Hypervisor Resource', :type => :virt do
     user = user_client(owner, random_string("user"))
     job_detail = async_update_hypervisor(owner, user, 'hypervisor.bind.com', hypervisor_id, ["blah"])
   end
+
+ it 'should be able to update a consumer with a hypervisor_id, and check it in as a hypervisor concurrently' do
+   host_hyp_id = random_string('host')
+   mapping = get_host_guest_mapping(host_hyp_id, [])
+   consumer = @user.register(@expected_host_name, :hypervisor, nil, {"test_fact" => "fact_value"},
+                  nil, nil, [], [], nil, [])
+   threads = []
+   i = 0
+   loop do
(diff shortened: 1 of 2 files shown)


server/spec/hypervisor_check_in_spec.rb
```diff
@@ -902,4 +902,35 @@ def create_virtwho_client(user)
     user = user_client(owner, random_string("user"))
     job_detail = async_update_hypervisor(owner, user, 'hypervisor.bind.com', hypervisor_id, ["blah"])
   end
+
+ it 'should be able to update a consumer and check it in as a hypervisor' do
+   host_hyp_id = random_string('host')
+   mapping = get_host_guest_mapping(host_hyp_id, [])
+   consumer = @user.register(@expected_host_name, :hypervisor, nil, {"test_fact" => "fact_value"},
+                  nil, nil, [], [], nil, [])
+   t1a = Thread.new{@cp.update_consumer({:uuid => consumer.uuid, :hypervisorId => host_hyp_id})}
```
cnsnyder: Any change we could loop over this?

server/spec/hypervisor_check_in_spec.rb
```diff
@@ -902,4 +902,35 @@ def create_virtwho_client(user)
     user = user_client(owner, random_string("user"))
     job_detail = async_update_hypervisor(owner, user, 'hypervisor.bind.com', hypervisor_id, ["blah"])
   end
+
+ it 'should be able to update a consumer and check it in as a hypervisor' do
```
cnsnyder: I would suggest changing this to "should be able to update a consumer with a hypervisor_id, and check it in as a hypervisor concurrently".

wottop:
[[8] BZ: 1502193 - periodic mysql exceptions when updating consumers in production](https://trello.com/c/kr5KOoAR/600-8-bz-1502193-periodic-mysql-exceptions-when-updating-consumers-in-production)

Ceiu:
We'll revisit this once we have time to evaluate the locking mechanism properly, and/or can use the DTOs to avoid pulling the consumer type with every consumer.

PR #1851, opened 2017-12-12
[2.1] Optimized the delete orphaned pools Liquibase task

Ceiu:
- Optimized the delete orphaned pools Liquibase task, using a separate
  set of queries to work around inefficiencies and limitations with
  how MySQL/MariaDB handles temporary tables and deletes


Ceiu:
If there are any Oracle issues from this PR, they will be addressed in a separate, larger PR that addresses all of our current Oracle problems.

kahowell:
Checkstyle should be fixed via #1855.

PR #1852, opened 2017-12-12
[M] Optimized the delete orphaned pools Liquibase task

Ceiu:
- Optimized the delete orphaned pools Liquibase task, using a separate
  set of queries to work around inefficiencies and limitations with
  how MySQL/MariaDB handles temporary tables and deletes

### DIFF
--- a/server/src/main/resources/db/changelog/20170718095058-purge-orphaned-pools.xml
+++ b/server/src/main/resources/db/changelog/20170718095058-purge-orphaned-pools.xml
@@ -6,556 +6,149 @@
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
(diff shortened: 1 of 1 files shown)


awood:
Looks nice, but I have a few questions.  We also run under `hsql`.  I'm presuming these migrations are not relevant for `hsql` but I want to verify.

Also, is there a good way to test  this?

Lastly, I know changing the author alters the checksum but I don't know if it also changes the "identity" of the changeset itself.  I'm guessing it does not since you added the valid checksum element in there, but again just wanted to verify.

Ceiu:
Regarding hsqldb, since we always build a fresh instance for each test, there's nothing to migrate/delete, so we skip it there.

A good way to test this is by throwing it at a prod-sized database and comparing runtimes. @mstead  and @vritant both have another DB that may be usable for testing as well.

I wasn't aware the author was part of the checksum (and I'm not sure I see the value in that, but whatever), but all of the changesets that were modified in any way got their old checksums added as valid ones, so we should be safe there. But, I may also be misunderstanding your question here.

PR #1853, opened 2017-12-12
getting vagrant to work

Lorquas:
The vagrant wasn't working.  Now it does.  Quick test is:
```
vagrant destroy
# faster than configuring nfs
CANDLEPIN_VAGRANT_NO_NFS='false' vagrant up
```

Fixes include:

- qpid_proton 1.18.x doesn't build due to a C compilation error, dropping to 1.17.0 - not sure what this effects, need input here.

- turned off guest management of vagrant's hostnamanger.  We don't actually need the guest to be managed and it was breaking dns resolution for many sites.

- Installing bundler and json_pure as both the candlepin user and root. Bundler will not install it if root has it (which we need for the executable), but then buildr fails because it cannot find the bundler gem.  Installing it in both places fixes this issue.  Open to better ideas.

- Using the bundler ansible module since it seems ansible 2 is the main thing now and it works a little faster.

Note:  If these changes to qpid are acceptable, I can merge this to the QE repos (which use the same ansible roles).  Currently spinning up of new candlepin servers is broken and this is blocking tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem 'stringex'
 gem 'digest-murmurhash'
 gem 'httpclient'
 gem 'activesupport', '~> 4.2'
-gem 'qpid_proton'
+gem 'qpid_proton', '~> 0.17.0'
 
 # Remove this once we are fully using the new Ruby bindings
 gem 'rest-client', '~> 1.6.0'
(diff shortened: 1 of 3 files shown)


Lorquas:
barnaby is also fixing this... will just update his....

PR #1857, opened 2017-12-13
[2.1] 1525204: Lock update consumer avoid to constraint violation with hypervi…

wottop:
…sor checkin

To test, run the new spec test against 2.1 HOTFIX. You will be able to see the consumerUpdate failure.
The added locking will keep the violation from occurring.

### DIFF
--- a/server/spec/hypervisor_check_in_spec.rb
+++ b/server/spec/hypervisor_check_in_spec.rb
@@ -902,4 +902,26 @@ describe 'Hypervisor Resource', :type => :virt do
     user = user_client(owner, random_string("user"))
     job_detail = async_update_hypervisor(owner, user, 'hypervisor.bind.com', hypervisor_id, ["blah"])
   end
+
+  it 'should be able to update a consumer with a hypervisor_id, and check it in as a hypervisor concurrently' do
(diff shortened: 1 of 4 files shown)


kahowell:
Looks like checkstyle should pass once #1855 lands.

wottop:
[[8] BZ: 1502193 - periodic mysql exceptions when updating consumers in production](https://trello.com/c/kr5KOoAR/600-8-bz-1502193-periodic-mysql-exceptions-when-updating-consumers-in-production)

wottop:
retest this please

barnabycourt:
Adding in progress flag until we have appropriate approvals for the release

barnabycourt:
This needs to  be updated to include a fix similar to the one in https://github.com/candlepin/candlepin/compare/candlepin-2.2-HOTFIX...crog/2.2/1531628 to change the fetchmode of the type attribute or other shared sub-objects of the consumer that are not being changed so a pessimistic lock is not held across multiple orgs or consumers inadvertently

Ceiu:
We'll revisit this once we have time to evaluate the locking mechanism properly, and/or can use the DTOs to avoid pulling the consumer type with every consumer.

PR #1858, opened 2017-12-15
[M] 1514517: Correctly update distributor's modified product content

mstead:
When a regular consumer binds to an entitlement providing product
modifiers, candlepin matches on product/provided products to
determine if the modified content should be added to the
entitlement certificate.

However, when a distributor binds to the same entitlement, candlepin
will also try to match on the derived product and derived
provided products.

This patch addresses a bug where candlepin was not considering
derived products when checking to see which modifier entitlements
to regenerate when pools/entitlements are deleted.

### DIFF
--- a/server/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/server/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -577,92 +577,63 @@ public class EntitlementCurator extends AbstractHibernateCurator<Entitlement> {
      * Marks all dependent entitlements of the given entitlements dirty, excluding those specified
      * as input. The result is the number of entitlements marked dirty by this operation.
      *
-     * @param entitlements
-     *  The collection of entitlements for which to mark dependent entitlements dirty
+     * @param entitlementIds
+     *  The collection of entitlement Ids for which to mark dependent entitlements dirty
      *
      * @return
      *  the number of entitlements marked dirty as a result of this operation
      */
-    public int markDependentEntitlementsDirty(Iterable<Entitlement> entitlements) {
-        return this.markDependentEntitlementsDirty(entitlements, true);
-    }
+    public int markDependentEntitlementsDirty(Iterable<String> entitlementIds) {
+        if (entitlementIds != null && entitlementIds.iterator().hasNext()) {
+            Set<String> eids = new HashSet<String>();
+            if (entitlementIds != null && entitlementIds.iterator().hasNext()) {
+                // Find all entitlements that are modified by any of the specified entitlements.
+                eids.addAll(findDependentEntitlementsByProvidedProduct(entitlementIds));
 
-    /**
-     * Marks all dependent entitlements of the given entitlements dirty, returning the number of
-     * entitlements marked dirty as a result of this operation.
-     *
-     * @param entitlements
-     *  The collection of entitlements for which to mark dependent entitlements dirty
-     *
-     * @param excludeInput
-     *  Whether or not the input entitlements should be excluded from the set of dependent
-     *  entitlements
-     *
-     * @return
-     *  the number of entitlements marked dirty as a result of this operation
-     */
-    public int markDependentEntitlementsDirty(Iterable<Entitlement> entitlements, boolean excludeInput) {
-        if (entitlements != null && entitlements.iterator().hasNext()) {
-            Set<String> entitlementIds = new HashSet<String>();
-
-            // Deduplicate our entitlement IDs, discarding null/incomplete entitlements
-            for (Entitlement entitlement : entitlements) {
-                // TODO: Should we throw an exception if we find a malformed/incomplete entitlement
-                // or just continue silently ignoring them?
-                if (entitlement != null && entitlement.getId() != null) {
-                    entitlementIds.add(entitlement.getId());
+                // Distributors require modified product content matched on derived products to flow
+                // downstream in the manifest. Determine if there are any modified entitlements based
+                // on derived products of any distributors and include those as well.
+                //
+                // We do this in a separate query as it gives us a chance to skip the second query
+                // if it does not apply to any of the specified entitlements.
+                //
+                // We do the filter here to avoid any unnecessary work in the
+                // following query and to reduce the number of entitlements to check.
+                Set<String> distributorEntitlements = filterDistributorEntitlementIds(entitlementIds);
+                if (!distributorEntitlements.isEmpty()) {
+                    eids.addAll(findDependentEntitlementsByDerivedProvidedProduct(distributorEntitlements));
                 }
             }
 
-            // Impl note:
-            // We do this in direct SQL, as it lets us take a sane path from base to dependent
-            // entitlements, rather than the lunacy that would be required with HQL, JPQL or
-            // CriteriaBuilder.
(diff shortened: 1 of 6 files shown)


mstead:
[[BZ 1514514] Generated manifests can have content not entitled to](https://trello.com/c/zx73SdIX/669-bz-1514514-generated-manifests-can-have-content-not-entitled-to)

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -987,4 +924,132 @@ public Entitlement findUpstreamEntitlementForStack(Consumer consumer, String sta
...
+            "  AND e2.dirty = false " +
+            "  AND e1.id IN (:entitlement_ids)" +
+            "  AND e2.id NOT IN (:entitlement_ids)";
+
+        private EntityManager em;
+        private int blockSize;
+
+        public DependentEntitlementQuery(EntityManager em, int blockSize) {
+            this.em = em;
+            this.blockSize = blockSize;
```
Ceiu: Block size should not be a parameter here -- this object should determine the proper block size based on the query it's going to be executing.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -987,4 +924,132 @@ public Entitlement findUpstreamEntitlementForStack(Consumer consumer, String sta
 
         return (Entitlement) activeNowQuery.uniqueResult();
     }
+
+    /**
+     * Defines the queries used to find the collection of modifier entitlements of
+     * a given set of entitlements.
+     */
+    private class DependentEntitlementQuery {
```
Ceiu: Why are we encapsulating this into a class? Nothing here seems particularly reusable, nor does it hold/manage any state for us.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -614,55 +596,59 @@ public int markDependentEntitlementsDirty(Iterable<Entitlement> entitlements, bo
...
+    public Set<String> filterDistributorEntitlementIds(Iterable<String> entsIdsToFilter) {
+        Set<String> filteredIds = new HashSet<String>();
+        if (entsIdsToFilter != null && entsIdsToFilter.iterator().hasNext()) {
+            String querySql = "SELECT DISTINCT e.id " +
+                "FROM Entitlement e JOIN e.consumer c JOIN c.type t " +
+                "WHERE t.manifest = true AND e.id IN (:entitlement_ids)";
 
-            Set<String> dirtyIds = new HashSet<String>();
+            Query query = this.getEntityManager().createQuery(querySql);
+            int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit() / 2);
```
Ceiu: You don't need to divide the block size by 2 here for the latter case. That was done in the previous query because it had two in-blocks, so we needed to make sure that each block was no more than half the total number of parameters for the entire query.

mstead:
> Also, I'm not hugely in favor of removing the string-ID based versions of the query as the operative methods. Given that we are slowly moving away from nesting objects, it is more likely as we go forward that we will be working with collections of IDs rather than collections of objects, so it makes more sense to me to design our curator methods such that they take collections of strings, then add helper methods which unbox object collections as necessary.

Not sure what you are getting at here. The queries deal with IDs as they did before. The new queries I added do the same -- deal with string IDs. I must be missing something.

Ceiu:
Sorry. I mean the method definitions on the curators themselves.

mstead:
> Sorry. I mean the method definitions on the curators themselves.

Are you referencing the method below? i.e Don't pass the Consumer, consumer id instead? And the other existing that are passing the model objects in?

https://github.com/candlepin/candlepin/pull/1858/files#diff-8864d89395315f6a5f8456d0e504fd91R1005

If so, I can change that and I can update the curator methods I've touched to do the same.

mstead:
NOTE: Will rebase once review has been completed.

mstead:
retest this please

server/src/main/java/org/candlepin/model/EntitlementCurator.java
Ceiu: This comment still applies -- we don't need to divide the query limit by two here.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -987,4 +875,175 @@ public Entitlement findUpstreamEntitlementForStack(Consumer consumer, String sta
...
+            "  AND e2.id NOT IN (:entitlement_ids)";
+
+        Query query = getEntityManager().createNativeQuery(queryStr);
+
+        Set<String> result = new HashSet<String>();
+        if (entitlementIds == null || !entitlementIds.iterator().hasNext()) {
+            return result;
+        }
+
+        int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit());
```
Ceiu: This query, however, *does* need the division by two, as we have two blocks of params in the query.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -987,4 +875,175 @@ public Entitlement findUpstreamEntitlementForStack(Consumer consumer, String sta
...
+                "JOIN cp2_product_content pc ON pc.content_uuid = cmp.content_uuid " +
+                // Dependent product => dependent pool
+                "JOIN cp2_pool_provided_products ppp2 ON ppp2.product_uuid = pc.product_uuid " +
+                // Dependent pool => dependent entitlement
+                "JOIN cp_entitlement e ON e.pool_id = ppp2.pool_id " +
+                "WHERE e.consumer_id = :consumer_id " +
+                "  AND ppp1.pool_id IN (:pool_ids) ";
+
+
+            int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit());
```
Ceiu: Need to subtract 1 from the query limit here.

mstead:
REMINDER: I'd like to squash commits before this is actually merged.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -577,92 +577,63 @@ private boolean poolOverlapsRange(Pool p, Date startDate, Date endDate) {
...
+    public Set<String> filterDistributorEntitlementIds(Iterable<String> entsIdsToFilter) {
+        Set<String> filteredIds = new HashSet<String>();
+        if (entsIdsToFilter != null && entsIdsToFilter.iterator().hasNext()) {
+            String querySql = "SELECT DISTINCT e.id " +
+                "FROM Entitlement e JOIN e.consumer c JOIN c.type t " +
+                "WHERE t.manifest = true AND e.id IN (:entitlement_ids)";
 
-            Set<String> dirtyIds = new HashSet<String>();
+            Query query = this.getEntityManager().createQuery(querySql);
+            int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit() - 1);
```
Ceiu: Don't need the minus one here. Just the query param limit.

Ceiu:
ACK

PR #1860, opened 2017-12-15
[DTO] Implemented PoolDTO, EntitlementDTO and BrandingDTO

nikosmoum:
- Created PoolDTO/PoolTranslator and internal DTOs: ProvidedProductDTO, SourceSubscriptionDTO and SourceStackDTO.
- Created EntitlementDTO/EntitlementTranslator.
- Created BrandingDTO/BrandingTranslator, which is an internal DTO used by EntitlementDTO, and can be used by SubscriptionDTO once that gets introduced.
- Changed endpoints in PoolResource, ActivationKeyResource, OwnerResource, ConsumerResource and EntitlementResource to use the new DTOs when receiving/sending data.
- Removed JAXB/Jackson and annotations from Branding, ProvidedProduct, SourceStack, SourceSubscription model objects.
- API Changes: All the new DTO classes (internal and external), no longer use the @HateoasInclude annotation and therefore are serialized fully as nested objects of their parent serialized entities, instead of only few select fields of them.
- Removed some unused @Index/@ForeignKey hibernate annotations from Entitlement and SourceStack model classes.

### DIFF
--- a/server/src/main/java/org/candlepin/dto/api/v1/EntitlementDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/EntitlementDTO.java
@@ -1,0 +1,495 @@
+/**
+ * Copyright (c) 2009 - 2017 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.dto.api.v1;
+
+import com.fasterxml.jackson.annotation.JsonFilter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.candlepin.common.jackson.HateoasInclude;
+import org.candlepin.model.Consumer;
+import org.candlepin.util.SetView;
+import org.candlepin.util.Util;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlRootElement;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * A DTO representation of the Entitlement entity
+ */
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.PROPERTY)
+@ApiModel(parent = TimestampedCandlepinDTO.class, description = "DTO representing an entitlement")
+@JsonFilter("EntitlementFilter")
+public class EntitlementDTO extends TimestampedCandlepinDTO<EntitlementDTO> implements LinkableDTO {
+
+    private static final long serialVersionUID = 1L;
+
+    private String id;
+    private OwnerDTO owner;
+    private Consumer consumer; //TODO: Turn into ConsumerDTO once that is introduced
+    private PoolDTO pool;
+    private Integer quantity;
+    private Boolean deletedFromPool;
+    private Set<CertificateDTO> certificates;
+    private Date startDate;
+    private Date endDate;
+
+    /**
+     * Initializes a new EntitlementDTO instance with null values.
+     */
+    public EntitlementDTO() {
+        // Intentionally left empty
+    }
+
+    /**
+     * Initializes a new EntitlementDTO instance which is a shallow copy of the provided
+     * source entity.
+     *
+     * @param source
+     *  The source entity to copy
+     */
+    public EntitlementDTO(EntitlementDTO source) {
+        super(source);
+    }
+
+    /**
+     * Retrieves the id field of this EntitlementDTO object.
+     *
+     * @return the id field of this EntitlementDTO object.
+     */
+    @HateoasInclude
+    public String getId() {
+        return this.id;
+    }
+
+    /**
+     * Sets the id to set on this EntitlementDTO object.
+     *
+     * @param id the id to set on this EntitlementDTO object.
+     *
+     * @return a reference to this EntitlementDTO object.
+     */
+    public EntitlementDTO setId(String id) {
+        this.id = id;
+        return this;
+    }
+
+    /**
+     * Returns the owner of this entitlement.
+     *
+     * @return the owner of this entitlement.
+     */
+    @JsonIgnore
+    public OwnerDTO getOwner() {
+        return this.owner;
+    }
+
+    /**
+     * Sets the owner of this entitlement.
+     *
+     * @param owner the owner to set.
+     *
+     * @return a reference to this EntitlementDTO object.
+     */
+    @JsonProperty
+    public EntitlementDTO setOwner(OwnerDTO owner) {
+        this.owner = owner;
+        return this;
+    }
+
+    /**
+     * Returns the consumer of this entitlement.
+     *
+     * @return return the associated Consumer.
+     */
+    //TODO: Replace with ConsumerDTO once that gets introduced.
+    public Consumer getConsumer() {
+        return consumer;
+    }
+
+    /**
+     * Associates the given consumer with this entitlement.
+     *
+     * @param consumer consumer to associate.
+     *
+     * @return a reference to this EntitlementDTO object.
+     */
+    //TODO: Replace with ConsumerDTO once that gets introduced.
+    public EntitlementDTO setConsumer(Consumer consumer) {
+        this.consumer = consumer;
+        return this;
+    }
+
+    /**
+     * Returns the pool of this entitlement.
+     *
+     * @return the pool of this entitlement.
+     */
+    public PoolDTO getPool() {
+        return this.pool;
+    }
+
+    /**
+     * Sets the pool of this entitlement.
+     *
+     * @param pool the pool to set.
+     *
+     * @return a reference to this EntitlementDTO object.
+     */
+    public EntitlementDTO setPool(PoolDTO pool) {
+        this.pool = pool;
+        return this;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @HateoasInclude
+    @Override
+    public String getHref() {
+        return this.getId() != null ? String.format("/entitlements/%s", this.getId()) : null;
+    }
+
+    /**
+     * Retrieves the quantity of this entitlement.
+     *
+     * @return the quantity of this entitlement.
+     */
+    public Integer getQuantity() {
+        return quantity;
+    }
+
+    /**
+     * Sets the quantity of this entitlement.
+     *
+     * @param quantity the quantity to set.
+     *
+     * @return a reference to this EntitlementDTO object.
+     */
+    public EntitlementDTO setQuantity(Integer quantity) {
+        this.quantity = quantity;
+        return this;
+    }
+
+    /**
+     * Returns true if this entitlement is deleted from the pool, or false otherwise.
+     *
+     * @return if this entitlement is deleted from the pool or not.
+     */
+    @JsonIgnore
+    public Boolean isDeletedFromPool() {
+        return deletedFromPool;
+    }
+
+    /**
+     * Sets if this entitlement is is deleted from the pool or not.
+     *
+     * @param deletedFromPool if this entitlement is deleted from the pool or not.
+     *
+     * @return a reference to this EntitlementDTO object.
+     */
+    @JsonProperty
+    public EntitlementDTO setDeletedFromPool(Boolean deletedFromPool) {
+        this.deletedFromPool = deletedFromPool;
+        return this;
+    }
+
+    /**
+     * Retrieves a view of the certificates associated with the entitlement represented by this DTO.
+     * If the certificates have not yet been defined, this method returns null.
+     * <p></p>
+     * Note that the collection returned by this method is a view of the collection backing this
+     * set of certificates. Elements cannot be added to the collection, but elements may be removed.
+     * Changes made to the collection will be reflected by this certificates data instance.
+     *
+     * @return
+     *  the certificates associated with this key, or null if the certificates have not yet been defined
+     */
+    public Set<CertificateDTO> getCertificates() {
+        return this.certificates != null ? new SetView<CertificateDTO>(this.certificates) : null;
+    }
+
+    /**
+     * Sets the certificates of the entitlement represented by this DTO.
+     *
+     * @param certificates
+     *  A collection of certificate DTOs to attach to this entitlement DTO, or null to clear the content
+     *
+     * @throws IllegalArgumentException
+     *  if the collection contains null or incomplete certificate DTOs
+     *
+     * @return
+     *  a reference to this EntitlementDTO object.
+     */
+    public EntitlementDTO setCertificates(Set<CertificateDTO> certificates) {
+        if (certificates != null) {
+            if (this.certificates == null) {
+                this.certificates = new HashSet<CertificateDTO>();
+            }
+            else {
+                this.certificates.clear();
+            }
+
+            for (CertificateDTO dto : certificates) {
+                if (isNullOrIncomplete(dto)) {
+                    throw new IllegalArgumentException(
+                        "collection contains null or incomplete certificates");
+                }
+            }
+            this.certificates.addAll(certificates);
+        }
+        else {
+            this.certificates = null;
+        }
+        return this;
+    }
+
+    /**
+     * Adds the given certificate to this entitlement DTO.
+     *
+     * @param certificate the certificate DTO to add to this entitlement DTO.
+     *
+     * @throws IllegalArgumentException
+     *  if the certificate is null or incomplete
+     *
+     * @return true if this certificate was not already contained in this entitlement DTO.
+     */
+    @JsonIgnore
+    public boolean addCertificate(CertificateDTO certificate) {
+        if (isNullOrIncomplete(certificate)) {
+            throw new IllegalArgumentException("certificate is null or incomplete");
+        }
+
+        if (this.certificates == null) {
+            this.certificates = new HashSet<CertificateDTO>();
+        }
+
+        return this.certificates.add(certificate);
+    }
+
+    /*
+     * Utility method to validate certificate input.
+     */
+    private boolean isNullOrIncomplete(CertificateDTO certificate) {
+        return certificate == null || certificate.getSerial() == null ||
+            certificate.getId() == null || certificate.getId().isEmpty() ||
+            certificate.getKey() == null || certificate.getKey().isEmpty() ||
+            certificate.getCert() == null || certificate.getCert().isEmpty();
+    }
+
+    /**
+     * Returns the start date of this entitlement.
+     *
+     * @return Returns the startDate from the pool of this entitlement.
+     */
+    @JsonProperty
+    public Date getStartDate() {
+        return startDate;
+    }
+
+    /**
+     * Sets the start date of this entitlement.
+     *
+     * @param startDate the startDate of this entitlement.
+     *
+     * @return a reference to this EntitlementDTO object.
+     */
+    @JsonIgnore
+    public EntitlementDTO setStartDate(Date startDate) {
+        this.startDate = startDate;
+        return this;
+    }
+
+    /**
+     * Returns the end date of this entitlement.
+     *
+     * @return Returns the endDate of this entitlement.
+     */
+    @JsonProperty
+    public Date getEndDate() {
+        return endDate;
+    }
+
+    /**
+     * Sets the end date of this entitlement.
+     *
+     * @param endDate the endDate of this entitlement.
+     *
+     * @return a reference to this EntitlementDTO object.
+     */
+    @JsonIgnore
+    public EntitlementDTO setEndDate(Date endDate) {
+        this.endDate = endDate;
+        return this;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String toString() {
+        return String.format("EntitlementDTO [id: %s, product id: %s, pool id: %s, consumer uuid: %s]",
+            this.id,
+            this.pool != null ? pool.getProductId() : null,
+            this.pool != null ? pool.getId() : null,
+            this.consumer != null ? consumer.getUuid() : null);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+
+        if (obj instanceof EntitlementDTO && super.equals(obj)) {
+            EntitlementDTO that = (EntitlementDTO) obj;
+
+            // Pull the nested object IDs, as we're not interested in verifying that the objects
+            // themselves are equal; just so long as they point to the same object.
+            String thisOwnerId = this.getOwner() != null ? this.getOwner().getId() : null;
+            String thatOwnerId = that.getOwner() != null ? that.getOwner().getId() : null;
+
+            String thisPoolId = this.getPool() != null ? this.getPool().getId() : null;
+            String thatPoolId = that.getPool() != null ? that.getPool().getId() : null;
+
+            String thisConsumerId = this.getConsumer() != null ? this.getConsumer().getUuid() : null;
+            String thatConsumerId = that.getConsumer() != null ? that.getConsumer().getUuid() : null;
+
+            EqualsBuilder builder = new EqualsBuilder()
+                .append(this.getId(), that.getId())
+                .append(thisOwnerId, thatOwnerId)
+                .append(thisPoolId, thatPoolId)
+                .append(thisConsumerId, thatConsumerId)
+                .append(this.getQuantity(), that.getQuantity())
+                .append(this.isDeletedFromPool(), that.isDeletedFromPool())
+                .append(this.getEndDate(), that.getEndDate())
+                .append(this.getStartDate(), that.getStartDate());
+
+            // Note that we're using the boolean operator here as a shorthand way to skip checks
+            // when the equality check has already failed.
+            boolean equals = builder.isEquals();
+
+            equals = equals && Util.collectionsAreEqual(this.getCertificates(), that.getCertificates(),
+                new Comparator<CertificateDTO>() {
+                    @Override
+                    public int compare(CertificateDTO c1, CertificateDTO c2) {
+                        if (c1 == c2) {
+                            return 0;
+                        }
+
+                        if (c1 != null && c1.getId() != null) {
+                            return c1.getId().compareTo(c2.getId());
+                        }
+
+                        return 1;
+                    }
+                });
+
+            return equals;
+        }
+
+        return false;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int hashCode() {
+        int certsHashCode = 0;
+        Collection<CertificateDTO> certificateDTOs = this.getCertificates();
+
+        if (certificateDTOs != null) {
+            for (CertificateDTO dto : certificateDTOs) {
+                certsHashCode = 31 * certsHashCode +
+                    (dto != null && dto.getId() != null ? dto.getId().hashCode() : 0);
+            }
+        }
+
+        HashCodeBuilder builder = new HashCodeBuilder(37, 7)
+            .append(super.hashCode())
+            .append(this.getId())
+            .append(this.getOwner() != null ? this.getOwner().getId() : null)
+            .append(this.getPool() != null ? this.getPool().getId() : null)
+            .append(this.getConsumer() != null ? this.getConsumer().getUuid() : null)
+            .append(this.getQuantity())
+            .append(this.isDeletedFromPool())
+            .append(this.getEndDate())
+            .append(this.getStartDate())
+            .append(certsHashCode);
+
+        return builder.toHashCode();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public EntitlementDTO clone() {
+        EntitlementDTO copy = (EntitlementDTO) super.clone();
+
+        OwnerDTO owner = this.getOwner();
+        copy.owner = owner != null ? owner.clone() : null;
+        PoolDTO pool = this.getPool();
+        copy.pool = pool != null ? pool.clone() : null;
+        Consumer consumer = this.getConsumer();
+        //TODO change Consumer to ConsumerDTO.
+        copy.consumer = consumer != null ?
+            new Consumer(consumer.getName(), consumer.getUsername(),
+                consumer.getOwner(), consumer.getType()) : null;
+        if (consumer != null) {
+            copy.consumer.setUuid(consumer.getUuid());
+        }
+
+        copy.certificates = this.getCertificates();
+
+        copy.endDate = this.endDate != null ? (Date) this.endDate.clone() : null;
+        copy.startDate = this.startDate != null ? (Date) this.startDate.clone() : null;
+
+        return copy;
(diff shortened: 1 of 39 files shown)


nikosmoum:
[[8] Add API DTO for Pool](https://trello.com/c/rXP0wwSH/634-8-add-api-dto-for-pool)

server/src/main/java/org/candlepin/resource/EntitlementResource.java
```diff
@@ -200,7 +210,8 @@ public Entitlement getEntitlement(
     @Path("{entitlement_id}")
     public void updateEntitlement(
         @PathParam("entitlement_id") @Verify(Entitlement.class) String id,
-        @ApiParam(name = "update", required = true) Entitlement update) {
+        @ApiParam(name = "update", required = true) EntitlementDTO update) {
+        //TODO Why do we accept a whole Entitlement object here, if we only use the quantity field???
```
Ceiu: Probably just a generic method originating from CRUD method specs, but this is a very valid question. If we only ever update the quantity, we'd be better served by dropping this endpoint and adding an updateQuantity to replace it.

server/src/test/java/org/candlepin/dto/api/v1/CertificateTranslatorTest.java
```diff
@@ -55,6 +55,7 @@ protected CertificateTranslator initObjectTranslator() {
 
     @Override
     protected Certificate initSourceObject() {
+        //FIXME: This should not be returning an IdentityCertificate, since this class is generic.
```
Ceiu: Unless the identity cert explicitly does something it shouldn't, it's fine here. So long as it's not misbehaving, any object adhering to the Certificate interface is acceptable.

vritant:
> API Changes: All the new DTO classes (internal and external), no longer use the @hateoasinclude annotation and therefore are serialized fully as nested objects of their parent serialized entities, instead of only few select fields of them.

Does this not conflict with our guideline that the JSON output in this branch should not change?

nikosmoum:
@vritant It does and I did ask @Ceiu about this and he said that I shouldn't worry about it and if it causes any problems we may have to bring them over again. But the idea was that eventually we'll not include nested object at all, so we don't need it. I'm still not sure about it though. If we expose extra data with these changes, I assume it would be harder to remove the nested object further down the line.

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
```diff
@@ -0,0 +1,1846 @@
...
+        }
+    }
+
+    /**
+     * Internal DTO object for SourceStack
+     */
+    public static class SourceStackDTO {
+        private final String id;
+        private final String sourceStackId;
+        private final Consumer sourceConsumer; //TODO Change to ConsumerDTO once its introduced.
```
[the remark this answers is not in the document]
nikosmoum: The Pool's sourceStack object doesn't get serialized. It's only available for deserialization.

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Utility method that translates the Pool's nested objects.
+     *
+     * @param modelTranslator the translator used to translate the nested objects.
+     *
+     * @param source the pool model object we are translating.
+     *
+     * @param dest the pool DTO object that the source model object is being translated to.
+     */
+    private void processNestedObjects(ModelTranslator modelTranslator, Pool source, PoolDTO dest) {
```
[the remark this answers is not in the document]
nikosmoum: checkstyle complained that the method was longer than 100 lines.

server/src/main/java/org/candlepin/dto/api/v1/EntitlementDTO.java
```diff
@@ -0,0 +1,541 @@
...
+            this.certificates = new HashSet<CertificateDTO>();
+        }
+
+        return this.certificates.add(certificate);
+    }
+
+    /*
+     * Utility method to validate certificate input.
+     */
+    private boolean isNullOrIncomplete(CertificateDTO certificate) {
```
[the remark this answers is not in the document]
nikosmoum: There's also two overloaded methods with this name in PoolDTO for two different types, so I'd prefer to leave it like that for consistency.

server/src/main/java/org/candlepin/dto/api/v1/EntitlementDTO.java
```diff
@@ -0,0 +1,541 @@
...
+        EntitlementDTO copy = (EntitlementDTO) super.clone();
+
+        OwnerDTO owner = this.getOwner();
+        copy.owner = owner != null ? owner.clone() : null;
+        PoolDTO pool = this.getPool();
+        copy.pool = pool != null ? pool.clone() : null;
+        Consumer consumer = this.getConsumer();
+        //TODO change Consumer to ConsumerDTO.
+        copy.consumer = consumer != null ?
+            new Consumer(consumer.getName(), consumer.getUsername(),
```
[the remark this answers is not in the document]
nikosmoum: I definitely would not use new here if it wasn't for the fact that Consumer does not implement clone(), and did not want to introduce it just for the sake of using it here, since this code will get replaced by ConsumerDTO which I expect to have a clone() method.

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
```diff
@@ -0,0 +1,200 @@
...
+
+        Entitlement sourceEntitlement = source.getSourceEntitlement();
+        dest.setSourceEntitlement(sourceEntitlement != null ?
+            modelTranslator.translate(sourceEntitlement, EntitlementDTO.class) : null);
+
+        SourceSubscription sourceSubscription = source.getSourceSubscription();
+        if (sourceSubscription != null) {
+            dest.setSourceSubscription(
+                new PoolDTO.SourceSubscriptionDTO(
+                sourceSubscription.getId(),
```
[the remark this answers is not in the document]
nikosmoum: checkstyle says otherwise! if I indent once more it complains, while leaving it like this is fine.

server/src/test/java/org/candlepin/dto/api/v1/CertificateTranslatorTest.java
nikosmoum: It's not about the behaviour of it, but the type. Unfortunately, when using these initSourceObject methods, we use the objects they create to populate other, parent objects, which expect a certain type. For example, when using this initSourceObject to populate an EntitlementCertificate in EntitlementTranslatorTest, I have to cast its result to EntitlementCertificate (because that is what is expected), and get "ClassCastException: IdentityCertificate cannot be cast to EntitlementCertificate". This forced me not to use this method, but to create an EntitlementCertificate manually.

vritant:
@Ceiu @mstead  comments on the above please?
FYI there are spec tests that would fail if we skip the nested objects in some objects today.
I dont have a preference either way, as long as we are consistent in what we do across all DTOs.
imo it seems simpler to ensure output matches existing output today, and am already working with that intent on my consumer PR.

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
```diff
@@ -0,0 +1,1846 @@
...
+    }
+
+    /**
+     * Return the "top level" product this pool is for.
+     * Note that pools can also provide access to other products.
+     * See getProvidedProducts().
+     *
+     * @return Top level product ID.
+     */
+    public String getProductId() {
```
Ceiu: We definitely don't need both product ID+name and the product object. We should pick one or the other.

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
```diff
@@ -0,0 +1,1846 @@
...
+
+    /**
+     * Sets the derived product name of this pool.
+     *
+     * @param derivedProductName set the derived product name of this pool.
+     *
+     * @return a reference to this PoolDTO object.
+     */
+    @JsonIgnore
+    public PoolDTO setDerivedProductName(String derivedProductName) {
```
Ceiu: Same with derived product. We need only the derived product object or the ID+name, not both.

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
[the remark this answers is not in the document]
Ceiu: Things like that are code smells to suggest maybe shortening the method. In cases where it's valid, however, it'd be preferable to simply suppress the warning rather than modifying the code to work around it in a way that doesn't actually reduce the overall footprint.

```@SuppressWarnings("checkstyle:methodlength")```

File: server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java (hunk 7 above)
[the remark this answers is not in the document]
Comment (Ceiu) [reply server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java]: ACK

Checkstyle is dumb sometimes. If it gets to be too troublesome, you can suppress it, but this will work for now.

File: server/src/main/java/org/candlepin/resource/OwnerResource.java [hunk 10]
```diff
@@ -459,24 +598,354 @@ else if (ownerDto.getKey() != null) {
...
+                        ProvidedProduct newProd = new ProvidedProduct();
+                        newProd.setProductId(providedProductDTO.getProductId());
+                        newProd.setProductName(providedProductDTO.getProductName());
+                        products.add(newProd);
+                    }
+                }
+                // We are using the entity's setProvidedProductDtos() instead of
+                // setProvidedProducts() to populate the providedProducts from the
+                // PoolDTO because the former was marked as @JsonProperty("providedProducts").
+                entity.setProvidedProductDtos(products);
```
Ceiu: I don't believe we use the provided product DTOs on deserialization at all. We may be able to ignore these entirely.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1405,35 +1891,39 @@ public Pool createPool(@PathParam("owner_key") @Verify(Owner.class) String owner
...
+                i18n.tr("Unable to find a pool with the ID \"{0}\"", newPoolDTO.getId()));
         }
 
-        // Ignore what the client has specified as the owner, since we already know the owner here
-        newPool.setOwner(currentPool.getOwner());
-
         // Verify the pool type is one that allows modifications
-        if (currentPool.getType() != PoolType.NORMAL || newPool.getType() != PoolType.NORMAL) {
+        if (currentPool.getType() != PoolType.NORMAL ||
+            !newPoolDTO.getType().equals(PoolType.NORMAL.toString())) {
```
Ceiu: This is a bit odd. We don't care about the input pool's type, since it can't be changed by the client. That in mind, it may not be worth having it pre-converted to a string, as that'll happen automatically anyway and we can ignore the field on deserialization.

Also, if we do want to use the type in this manner, we need to worry about the type being null.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1405,35 +1891,39 @@ public Pool createPool(@PathParam("owner_key") @Verify(Owner.class) String owner
...
             throw new BadRequestException(i18n.tr("Cannot update bonus pools, as they are auto generated"));
         }
 
-        if (currentPool.isCreatedByShare() || newPool.isCreatedByShare()) {
+        if (currentPool.isCreatedByShare() || newPoolDTO.isCreatedByShare()) {
```
Ceiu: We're not interested in this value on the DTO -- only use the actual entity as the source of truth for state validation.

server/src/test/java/org/candlepin/dto/api/v1/CertificateTranslatorTest.java
Ceiu: Subclasses can redeclare a method to be more specific than its parent, which is what you should do in such a case. As far as this particular method goes, returning an IdentityCertificate is still valid here, as the contract only states that it returns any Certificate.

server/src/test/java/org/candlepin/dto/api/v1/EntitlementTranslatorTest.java
```diff
@@ -0,0 +1,141 @@
...
+            assertEquals(source.getId(), dest.getId());
+            assertEquals(source.getQuantity(), dest.getQuantity());
+            assertEquals(source.getEndDateOverride(), dest.getEndDateOverride());
+            assertEquals(source.isDirty(), dest.isDirty());
+            assertEquals(source.isUpdatedOnStart(), dest.isUpdatedOnStart());
+            assertEquals(source.deletedFromPool(), dest.isDeletedFromPool());
+
+            if (childrenGenerated) {
+                this.ownerTranslatorTest
+                        .verifyOutput(source.getOwner(), dest.getOwner(), true);
```
Ceiu: double indentation

server/src/test/java/org/candlepin/dto/api/v1/PoolDTOTest.java
```diff
@@ -0,0 +1,425 @@
...
+        dto.hasProductAttribute(null);
+    }
+
+    @Test
+    public void testAddProvidedProductWithAbsentProvidedProduct() {
+        PoolDTO dto = new PoolDTO();
+
+        PoolDTO.ProvidedProductDTO product =
+            new PoolDTO.ProvidedProductDTO(
+            "test-id-provided-product-1",
```
Ceiu: Both of these input values can probably fit on line 273 with the instantiation.

server/src/test/java/org/candlepin/dto/api/v1/PoolDTOTest.java
```diff
@@ -0,0 +1,425 @@
...
+        assertTrue(dto.addProvidedProduct(product));
+    }
+
+    @Test
+    public void testAddProvidedProductWithPresentProvidedProduct() {
+        PoolDTO dto = new PoolDTO();
+
+        PoolDTO.ProvidedProductDTO product =
+            new PoolDTO.ProvidedProductDTO(
+            "test-id-provided-product-2",
```
Ceiu: Indentation. Keep in mind that checkstyle loses to actual readability.

server/src/test/java/org/candlepin/dto/api/v1/PoolDTOTest.java
```diff
@@ -0,0 +1,425 @@
...
+            "test-name-provided-product-2");
+        assertTrue(dto.addProvidedProduct(product));
+
+        PoolDTO.ProvidedProductDTO product2 =
+            new PoolDTO.ProvidedProductDTO(
+            "test-id-provided-product-2",
```
Ceiu: Indentation. Skipping further references to this issue.

server/src/test/java/org/candlepin/dto/api/v1/ProductTranslatorTest.java
```diff
@@ -36,7 +36,7 @@
 
 
 /**
- * Test suite for the UpstreamConsumerTranslator class
+ * Test suite for the ProductTranslator class
```
Ceiu: :+1:

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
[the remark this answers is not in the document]
nikosmoum: Are you referring to an existing annotation, or something we'll introduce?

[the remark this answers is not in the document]
Ceiu: Existing ones. I don't recall details exactly, but I think it's something like a global ```@JsonIgnore``` on the class itself, or it's a slightly different one. I know it exists, though. :)

nikosmoum:
@Ceiu Following our discussion, am I understanding correctly that the 'microDTOs' should probably be nested classes which contain only the required fields, 2 of them which already exist (SourceStackDTO, SourceSubscriptionDTO), and the other one (nested ProductDTO, for product/derivedProduct) will have to be introduced, right?

server/src/main/java/org/candlepin/resource/OwnerResource.java
nikosmoum: @Ceiu We do use the providedProducts and derivedProvidedProducts fields on deserialization (at least there are many spec tests that use them). (But not the providedProductDtos/derivedProvidedProductDtos fields).

nikosmoum:
retest this please

nikosmoum:
retest this please

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -264,6 +291,75 @@ private Owner findOwner(String key) {
...
+     *  if the owner with the given id and key was not found in the system.
+     */
+    private Owner findOwnerByIdOrKey(String id, String key) {
+        Owner owner = null;
+
+        if (id != null) {
+            owner = this.ownerCurator.find(id);
+        }
+
+        if (owner == null && key != null) {
```
Ceiu: We need to be careful with these kinds of lookups. Since this will be initiated by client-input, we're looking to pick one or the other according to which has higher accuracy. In this case, use ID if it's present, otherwise use the key. This leaves us with a lookup like:

```
if (id != null) {
   // by-id lookup
   if (owner == null) {
      // error: could not find owner by id
   }
}
else if (key != null) {
   ... // key lookup
   if (owner == null) {
      // error: could not find owner by key
}
else {
   // error: no identifiers provided
}
```

Since that's basically just wrapped versions of the other lookups, we can avoid some extra translation bits and whatnot by just calling out to each:

```
if (id != null) {
   return this.findOwnerById(id);
}
else if (key != null) {
   return this.findOwnerByKey(key);
}
else {
   // error: no identifying info
}
```

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -274,6 +370,48 @@ private Consumer findConsumer(String consumerUuid) {
...
+     *
+     * @param poolId the id of the pool we are searching for.
+     *
+     * @return the pool that was found in the system based on the given id.
+     *
+     * @throws NotFoundException
+     *  if the pool with the given id was not found in the system.
+     */
+    private Pool findPool(String poolId) {
+        Pool pool = poolManager.find(poolId);
```
Ceiu: I'd need to check, but this might fail in when poolId is null or empty. If so we need to wrap this (and other lookups in similar methods) in a null or empty check.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -459,24 +597,314 @@ else if (ownerDto.getKey() != null) {
...
+     *  The entity instance to populate
+     *
+     * @param dto
+     *  The DTO containing the data with which to populate the entity
+     *
+     * @throws IllegalArgumentException
+     *  if either entity or dto are null
+     */
+    @SuppressWarnings("checkstyle:methodlength")
+    protected void populateEntity(Pool entity, PoolDTO dto) {
```
Ceiu: We need to step through this a bit more carefully and ignore fields we don't allow the client to set.

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
```diff
@@ -0,0 +1,1904 @@
...
+
+    /**
+     * Internal DTO object for Product
+     */
+    public static class ProductDTO {
+        private final String id;
+        private final String name;
+
+        @JsonCreator
+        public ProductDTO(
```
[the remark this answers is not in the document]
nikosmoum: I agree. I originally thought of naming this InternalProductDTO. Don't know why I changed it.

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
```diff
@@ -0,0 +1,1904 @@
...
+
+    private Set<BrandingDTO> branding;
+
+    private Map<String, String> calculatedAttributes;
+    private String upstreamPoolId;
+    private String upstreamEntitlementId;
+    private String upstreamConsumerId;
+
+    private String productName;
+    private String productId;
```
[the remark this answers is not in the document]
nikosmoum: We need the mini product object, since deserialization used it.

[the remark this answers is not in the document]
Ceiu: It shouldn't; and if there are such places, we need to standardize them on which fields they'll use from the DTOs as well.

The whole point of these DTOs is to get rid of these extraneous bits of data, duplicate fields and whatnot. Even if we're actually using some of them, this is our time to fix that.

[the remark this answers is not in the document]
nikosmoum: Are you saying that we want to get rid of the mini product in the future from the API, so we should get rid of it now from the DTO, and write some kind of custom deserializer again for productId/productName fields?

server/src/main/java/org/candlepin/resource/OwnerResource.java
nikosmoum: I wrote it this way because even though the id might not be null and we do a lookup with it, the lookup might fail, in which case we fall back to the key. If we use an 'else if', then the fallback to the key will never run. Or is it that we don't care about it?

Ceiu: The else-if will use the key only if the ID is not present, which is what we want. We want to use the most specific identifier provided, not a mixture of them. If we're provided an owner object that has both, but they don't align, then that's malformed data -- we can't be certain which the client intended to use.

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
```diff
@@ -0,0 +1,1904 @@
...
+
+            return builder.toHashCode();
+        }
+    }
+
+    /**
+     * Internal DTO object for ProvidedProduct
+     */
+    @JsonFilter("ProvidedProductFilter")
+    public static class ProvidedProductDTO {
```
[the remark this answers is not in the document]
nikosmoum: They're not identical to the client though. From what I see on the spec test candlepin_api.rb, the id field on each of these fields has a different name. For the product/derivedProduct, it is called 'id', while for the providedProducts/derivedProvidedProducts collections, it is called 'productId'. Also, I cannot amend this behaviour by having a single internal ProductDTO class and using two constructors because adding two @JsonCreator annotations is not supported.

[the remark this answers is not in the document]
nikosmoum: @Ceiu The question I wanted to ask is about this^. Is keeping the two identical internal DTOs fine, since I don't see an alternative solution?

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
```diff
@@ -0,0 +1,1904 @@
...
+            if (subscriptionSubKey == null || subscriptionSubKey.isEmpty()) {
+                throw new IllegalArgumentException("The subscriptionSubKey is null or empty.");
+            }
+
+            this.id = id;
+            this.subscriptionId = subscriptionId;
+            this.subscriptionSubKey = subscriptionSubKey;
+        }
+
+        public String getId() {
```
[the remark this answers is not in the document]
Ceiu: It's not necessary to have both methods for storing the data. As of current master, this is what a pool looks like once serialized:
```
{
  "id" : "ff808081611e541901611e549c630196",
  "type" : "NORMAL",
  "owner" : {
    "id" : "ff808081611e541901611e5438400003",
    "key" : "admin",
    "displayName" : "Admin Owner",
    "href" : "/owners/admin"
  },
  "activeSubscription" : true,
  "createdByShare" : false,
  "hasSharedAncestor" : false,
  "sourceEntitlement" : null,
  "quantity" : 5,
  "startDate" : "2018-01-22T00:00:00+0000",
  "endDate" : "2019-01-22T00:00:00+0000",
  "attributes" : [ ],
  "restrictedToUsername" : null,
  "contractNumber" : "0",
  "accountNumber" : "12331131231",
  "orderNumber" : "order-8675309",
  "consumed" : 0,
  "exported" : 0,
  "shared" : 0,
  "branding" : [ {
    "productId" : "32060",
    "name" : "Branded Admin OS Instance Based one socket",
    "type" : "OS",
    "created" : "2018-01-22T14:45:17+0000",
    "updated" : "2018-01-22T14:45:17+0000"
  } ],
  "calculatedAttributes" : {
    "compliance_type" : "Instance Based"
  },
  "upstreamPoolId" : "upstream_-Iwldslh6",
  "upstreamEntitlementId" : null,
  "upstreamConsumerId" : null,
  "productId" : "adminos-onesocketib",
  "productName" : "Admin OS Instance Based one socket",
  "productAttributes" : [ {
    "name" : "host_limited",
    "value" : "true"
  }, {
    "name" : "instance_multiplier",
    "value" : "2"
  }, {
    "name" : "multi-entitlement",
    "value" : "yes"
  }, {
    "name" : "virt_limit",
    "value" : "1"
  }, {
    "name" : "variant",
    "value" : "ALL"
  }, {
    "name" : "stacking_id",
    "value" : "15"
  }, {
    "name" : "support_level",
    "value" : "Standard"
  }, {
    "name" : "sockets",
    "value" : "1"
  }, {
    "name" : "support_type",
    "value" : "L1-L3"
  }, {
    "name" : "arch",
    "value" : "ALL"
  }, {
    "name" : "type",
    "value" : "MKT"
  }, {
    "name" : "version",
    "value" : "1.0"
  } ],
  "stacked" : true,
  "stackId" : "15",
  "developmentPool" : false,
  "href" : "/pools/ff808081611e541901611e549c630196",
  "derivedProductId" : null,
  "derivedProductAttributes" : [ ],
  "derivedProductName" : null,
  "created" : "2018-01-22T14:45:17+0000",
  "updated" : "2018-01-22T14:45:17+0000",
  "providedProducts" : [ {
    "productId" : "32060",
    "productName" : "Awesome OS Instance Server Bits"
  }, {
    "productId" : "5050",
    "productName" : "Admin OS Premium Architecture Bits"
  }, {
    "productId" : "5051",
    "productName" : "Admin OS Developer Bits"
  } ],
  "derivedProvidedProducts" : [ ],
  "subscriptionSubKey" : "master",
  "sourceStackId" : null,
  "subscriptionId" : "source_sub_-WALN73Zi"
}
```

Note that there's no "product" object here (nor source subscription or source stack). The data that's not present need not be on the DTO at all. If we're currently expecting it on input, that's strange but is something we're free to fix now. One of the biggest drivers of the DTOs is to no longer need all of these weird workarounds and cludges to express data differently between the client view and the database.

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
```diff
@@ -0,0 +1,1548 @@
...
+        return this;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @HateoasInclude
+    @Override
+    public String getHref() {
+        return this.id != null ? String.format("/pools/%s", this.id) : null;
```
Ceiu: For methods like this which pull from other bits of data, since we're making it extensible, try to use the methods instead:
```    return this.getId() != null ? String.format(...) : null;```

That way subclasses need not override both methods should the source of the ID change.

File: server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java [hunk 26]
```diff
@@ -0,0 +1,1548 @@
...
+
+        copy.setAttributes(this.getAttributes());
+        copy.setCalculatedAttributes(this.getCalculatedAttributes());
+        copy.setProductAttributes(this.getProductAttributes());
+        copy.setDerivedProductAttributes(this.getDerivedProductAttributes());
+        copy.setBranding(this.getBranding());
+        copy.setProvidedProducts(this.getProvidedProducts());
+        copy.setDerivedProvidedProducts(this.getDerivedProvidedProducts());
+
+        copy.endDate = this.endDate != null ? new Date(this.endDate.getTime()) : null;
```
Ceiu: Use this.endDate.clone() instead so we're guaranteed to get the same class of date back.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -459,24 +667,250 @@ else if (ownerDto.getKey() != null) {
...
+        if (dto.getEndDate() != null) {
+            entity.setEndDate(dto.getEndDate());
         }
 
-        return product;
+        if (dto.getQuantity() != null) {
+            entity.setQuantity(dto.getQuantity());
+        }
+
+        if (dto.getConsumed() != null) {
```
Ceiu: Clients aren't allowed to set this field.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -459,24 +667,250 @@ else if (ownerDto.getKey() != null) {
...
+            entity.setConsumed(dto.getConsumed());
+        }
+
+        if (dto.getExported() != null) {
```
Ceiu: Same here

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -459,24 +667,250 @@ else if (ownerDto.getKey() != null) {
...
+            entity.setExported(dto.getExported());
+        }
+
+        if (dto.getShared() != null) {
+            entity.setShared(dto.getShared());
```
Ceiu: Shouldn't be setting this.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -459,24 +667,250 @@ else if (ownerDto.getKey() != null) {
...
+        if (dto.getAccountNumber() != null) {
+            entity.setAccountNumber(dto.getAccountNumber());
+        }
+
+        if (dto.getOrderNumber() != null) {
+            entity.setOrderNumber(dto.getOrderNumber());
+        }
+
+        if (dto.isActiveSubscription() != null) {
+            entity.setActiveSubscription(dto.isActiveSubscription());
```
Ceiu: Shouldn't be setting this.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -459,24 +667,250 @@ else if (ownerDto.getKey() != null) {
...
+        }
+
+        if (dto.isCreatedByShare() != null) {
+            entity.setCreatedByShare(dto.isCreatedByShare());
```
Ceiu: Shouldn't be setting this.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -459,24 +667,250 @@ else if (ownerDto.getKey() != null) {
...
+        }
+
+        if (dto.hasSharedAncestor() != null) {
+            entity.setHasSharedAncestor(dto.hasSharedAncestor());
```
Ceiu: Shouldn't be setting this.

server/src/main/java/org/candlepin/dto/api/v1/EntitlementDTO.java
```diff
@@ -0,0 +1,495 @@
...
+        return this;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @HateoasInclude
+    @Override
+    public String getHref() {
+        return "/entitlements/" + getId();
```
Ceiu: Need to update this to use safe URI generation.

server/src/main/java/org/candlepin/dto/api/v1/PoolDTO.java
```diff
@@ -0,0 +1,1548 @@
...
+import java.util.Set;
+
+/**
+ * A DTO representation of the Pool entity
+ */
+@XmlRootElement(name = "pool")
+@XmlAccessorType(XmlAccessType.PROPERTY)
+@ApiModel(parent = TimestampedCandlepinDTO.class, description = "DTO representing a Pool")
+@JsonFilter("PoolFilter")
+public class PoolDTO extends TimestampedCandlepinDTO<PoolDTO> implements  LinkableDTO {
```
Ceiu: Double space before LinkableDTO

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1382,17 +1825,25 @@ public JobDetail refreshPools(
...
+
+        // Correct owner
+        Owner owner = findOwnerByKey(ownerKey);
         pool.setOwner(owner);
 
+        // Populate the rest of the pool
+        this.populateEntity(pool, inputPoolDTO);
+
+        // Resolve products
         pool = resolverUtil.resolvePool(pool);
```
Ceiu: This is probably unnecessary now -- all of the resolution is done by the population of the entity from the DTO.

PR #1861, opened 2017-12-15
[M] Fixed several Oracle-related SQL errors in Liquibase tasks

Ceiu:
- Several Liquibase change sets have been updated to address errors
  which occurred when running against Oracle databases
- Optimized a few queries to work faster and/or require less
  separation between the database plaforms
- Removed an unneeded rollback task in the update-revoked-cert-serial
  task


wottop:
I cannot get through the Oracle build process in Master, much less this branch.
liquibase-3.5.3-1.fc24.noarch
oracle-xe-11.2.0-1.0.x86_64

Ceiu:
@wottop Are you still having issues with this? I'm not able to reproduce.

Ceiu:
@wottop Apparently newer versions of Liquibase don't like one of the types we were using for our timestamps. I've pushed another change here that should address that on Oracle. Let me know if this solves the issue you're seeing.

wottop:
ack, waiting for jenkins

PR #1862, opened 2017-12-15
1479353: Add 'after' flag to pool listing call to better find future …

wottop:
…pools

### DIFF
--- a/server/spec/owner_resource_spec.rb
+++ b/server/spec/owner_resource_spec.rb
@@ -864,10 +864,31 @@ describe 'Owner Resource Future Pool Tests' do
     pools[0].id.should eq(@future_pool2.id)
   end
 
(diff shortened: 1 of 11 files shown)


wottop:
[1479353: Add support for looking up pools available 'after' a specified date.](https://trello.com/c/4ezql5BU/695-1479353-add-support-for-looking-up-pools-available-after-a-specified-date)

cnsnyder:
I've added a PR against subman that implements the requested feature using the existing API.
We'll need some changes to subman either way to implement the RFE and so it is done.
Cheers

PR: https://github.com/candlepin/subscription-manager/pull/1751

cnsnyder:
I see that we will need this as done for the subman RFE (after taking a second look).

LGTM! Thank you!

PR #1863, opened 2017-12-18
[DTO] ConsumerDTO and dependent DTOs

vritant:
Added ConsumerDTO and dependent DTOs

* added new translate method for collections
* added following DTOs and Translators:
  * Capability
  * ConsumerInstalledProduct
  * Consumer
  * Environment
  * GuestId
  * HypervisorId
* added unit tests for add, remove methods for collections
* added hateos annotations and json filter annotations to DTOs

### DIFF
--- a/server/src/main/java/org/candlepin/dto/ModelTranslator.java
+++ b/server/src/main/java/org/candlepin/dto/ModelTranslator.java
@@ -16,8 +16,6 @@ package org.candlepin.dto;
 
 import org.candlepin.model.CandlepinQuery;
 
-
-
 /**
  * The ModelTranslator interface defines common functionality required by every ModelTranslator
  * implementation.
--- a/server/src/main/java/org/candlepin/dto/SimpleModelTranslator.java
+++ b/server/src/main/java/org/candlepin/dto/SimpleModelTranslator.java
@@ -20,9 +20,9 @@ import org.candlepin.util.ElementTransformer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Map;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.Set;
 
 
--- a/server/src/main/java/org/candlepin/dto/api/v1/CapabilityDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/CapabilityDTO.java
@@ -1,0 +1,149 @@
+/**
+ * Copyright (c) 2009 - 2017 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.dto.api.v1;
+
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+
+import io.swagger.annotations.ApiModel;
+
+/**
+ * A DTO representation of the Consumer Capability entity
+ */
+@ApiModel(parent = TimestampedCandlepinDTO.class, description = "DTO representing a consumer capability")
+public class CapabilityDTO extends TimestampedCandlepinDTO<CapabilityDTO> {
+    public static final long serialVersionUID = 1L;
+
+    protected String id;
+    protected String name;
+
+    /**
+     * Initializes a new CapabilityDTO instance with null values.
+     */
+    public CapabilityDTO() {
+        // Intentionally left empty
+    }
+
+    /**
+     * Initializes a new CapabilityDTO instance which is a shallow copy of the provided
+     * source entity.
+     *
+     * @param source
+     *  The source entity to copy
+     */
+    public CapabilityDTO(CapabilityDTO source) {
+        super(source);
+    }
+
+    /**
+     * Retrieves the id field of this CapabilityDTO object.
+     *
+     * @return the id of the capability, or null if the id has not yet been defined
+     */
+    public String getId() {
+        return this.id;
+    }
+
+    /**
+     * Sets the id to set on this CapabilityDTO object.
+     *
+     * @param id the id to set on this CapabilityDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public CapabilityDTO setId(String id) {
+        this.id = id;
+        return this;
+    }
+
+    /**
+     * Retrieves the name field of this CapabilityDTO object.
+     *
+     * @return the name of the capability, or null if the id has not yet been defined
+     */
+    public String getName() {
+        return this.name;
+    }
+
+    /**
+     * Sets the name to set on this CapabilityDTO object.
+     *
+     * @param name the name to set on this CapabilityDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public CapabilityDTO setName(String name) {
+        this.name = name;
+        return this;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String toString() {
+        return String.format(
+            "CapabilityDTO [id: %s, name: %s]",
+            this.getId(), this.getName());
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+
+        if (obj instanceof CapabilityDTO && super.equals(obj)) {
+            CapabilityDTO that = (CapabilityDTO) obj;
+
+            EqualsBuilder builder = new EqualsBuilder()
+                .append(this.getId(), that.getId())
+                .append(this.getName(), that.getName());
+
+            return builder.isEquals();
+        }
+
+        return false;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int hashCode() {
+        HashCodeBuilder builder = new HashCodeBuilder(37, 7)
+            .append(super.hashCode())
+            .append(this.getId())
+            .append(this.getName());
+
+        return builder.toHashCode();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public CapabilityDTO populate(CapabilityDTO source) {
+        super.populate(source);
+
+        this.setId(source.getId());
+        this.setName(source.getName());
+
+        return this;
+    }
+}
--- a/server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
@@ -1,0 +1,1009 @@
+/**
+ * Copyright (c) 2009 - 2017 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.dto.api.v1;
+
+import org.candlepin.common.jackson.HateoasArrayExclude;
+import org.candlepin.common.jackson.HateoasInclude;
+import org.candlepin.jackson.SingleValueWrapDeserializer;
+import org.candlepin.jackson.SingleValueWrapSerializer;
+import org.candlepin.util.ListView;
+import org.candlepin.util.MapView;
+import org.candlepin.util.SetView;
+
+import com.fasterxml.jackson.annotation.JsonFilter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.apache.commons.lang3.StringUtils;
+
+import io.swagger.annotations.ApiModel;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlRootElement;
+
+/**
+ * A DTO representation of the Consumer entity
+ */
+@ApiModel(parent = TimestampedCandlepinDTO.class, description = "DTO representing an upstream consumer")
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.PROPERTY)
+@JsonFilter("ConsumerFilter")
+public class ConsumerDTO extends TimestampedCandlepinDTO<ConsumerDTO> implements LinkableDTO {
+    public static final long serialVersionUID = 1L;
+
+    protected String id;
+    protected String uuid;
+    protected String name;
+    protected String username;
+    protected String entitlementStatus;
+    protected String serviceLevel;
+    protected String releaseVer;
+    protected OwnerDTO owner;
+    protected EnvironmentDTO environment;
+    protected Long entitlementCount;
+    protected Map<String, String> facts;
+    protected Date lastCheckin;
+    protected Set<ConsumerInstalledProductDTO> installedProducts;
+    protected Boolean canActivate;
+    protected Set<CapabilityDTO> capabilities;
+    protected HypervisorIdDTO hypervisorId;
+    protected Set<String> contentTags;
+    protected Boolean autoheal;
+    protected String recipientOwnerKey;
+    protected String annotations;
+    protected String contentAccessMode;
+    protected ConsumerTypeDTO type;
+    protected CertificateDTO idCert;
+    protected List<GuestIdDTO> guestIds;
+
+    /**
+     * Serialization utility class for wrapping the 'releaseVer' field in a JSON object.
+     */
+    private static class ReleaseVersionWrapSerializer extends SingleValueWrapSerializer {
+        public ReleaseVersionWrapSerializer() {
+            super("releaseVer");
+        }
+    }
+
+    /**
+     * Deserialization utility class for for unwrapping the 'releaseVer' field from a JSON object.
+     */
+    private static class ReleaseVersionWrapDeserializer extends SingleValueWrapDeserializer {
+        public ReleaseVersionWrapDeserializer() {
+            super("releaseVer");
+        }
+    }
+
+    /**
+     * Initializes a new ConsumerDTO instance with null values.
+     */
+    public ConsumerDTO() {
+        // Intentionally left blank
+    }
+
+    /**
+     * Initializes a new ConsumerDTO instance which is a shallow copy of the provided
+     * source entity.
+     *
+     * @param source
+     *  The source entity to copy
+     */
+    public ConsumerDTO(ConsumerDTO source) {
+        super(source);
+    }
+
+    /**
+     * Retrieves the id field of this ConsumerDTO object.
+     *
+     * @return the id of the consumer, or null if the id has not yet been defined
+     */
+    @HateoasInclude
+    public String getId() {
+        return this.id;
+    }
+
+    /**
+     * Sets the id to set on this ConsumerDTO object.
+     *
+     * @param id the id to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setId(String id) {
+        this.id = id;
+        return this;
+    }
+
+    /**
+     * Retrieves the uuid field of this ConsumerDTO object.
+     *
+     * @return the uuid of the consumer, or null if the uuid has not yet been defined
+     */
+    @HateoasInclude
+    public String getUuid() {
+        return this.uuid;
+    }
+
+    /**
+     * Sets the uuid to set on this ConsumerDTO object.
+     *
+     * @param uuid the id to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setUuid(String uuid) {
+        this.uuid = uuid;
+        return this;
+    }
+
+    /**
+     * Retrieves the name field of this ConsumerDTO object.
+     *
+     * @return the name of the consumer, or null if the name has not yet been defined
+     */
+    @HateoasInclude
+    public String getName() {
+        return this.name;
+    }
+
+    /**
+     * Sets the name to set on this ConsumerDTO object.
+     *
+     * @param name the name to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setName(String name) {
+        this.name = name;
+        return this;
+    }
+
+    /**
+     * Retrieves the username field of this ConsumerDTO object.
+     *
+     * @return the username of the consumer, or null if the username has not yet been defined
+     */
+    public String getUsername() {
+        return this.username;
+    }
+
+    /**
+     * Sets the username to set on this ConsumerDTO object.
+     *
+     * @param username the username to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setUsername(String username) {
+        this.username = username;
+        return this;
+    }
+
+    /**
+     * Retrieves the entitlement status field of this ConsumerDTO object.
+     *
+     * @return the entitlementStatus of the consumer, or null if it has not yet been defined
+     */
+    public String getEntitlementStatus() {
+        return entitlementStatus;
+    }
+
+    /**
+     * Sets the entitlement status to set on this ConsumerDTO object.
+     *
+     * @param entitlementStatus the entitlement status to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setEntitlementStatus(String entitlementStatus) {
+        this.entitlementStatus = entitlementStatus;
+        return this;
+    }
+
+    /**
+     * Retrieves the service level field of this ConsumerDTO object.
+     *
+     * @return the serviceLevel of the consumer, or null if it has not yet been defined
+     */
+    public String getServiceLevel() {
+        return this.serviceLevel;
+    }
+
+    /**
+     * Sets the service level to set on this ConsumerDTO object.
+     *
+     * @param serviceLevel the service level to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setServiceLevel(String serviceLevel) {
+        this.serviceLevel = serviceLevel;
+        return this;
+    }
+
+    /**
+     * Retrieves the releaseVer field of this ConsumerDTO object.
+     *
+     * @return the releaseVer of the consumer, or null if it has not yet been defined
+     */
+    @JsonSerialize(using = ConsumerDTO.ReleaseVersionWrapSerializer.class,
+        nullsUsing = ConsumerDTO.ReleaseVersionWrapSerializer.class)
+    @JsonProperty("releaseVer")
+    public String getReleaseVersion() {
+        return releaseVer;
+    }
+
+    /**
+     * Sets the releaseVer to set on this ConsumerDTO object.
+     *
+     * @param releaseVer the release version to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    @JsonDeserialize(using = ConsumerDTO.ReleaseVersionWrapDeserializer.class)
+    @JsonProperty("releaseVer")
+    public ConsumerDTO setReleaseVersion(String releaseVer) {
+        this.releaseVer = releaseVer;
+        return this;
+    }
+
+    /**
+     * Retrieves the owner field of this ConsumerDTO object.
+     *
+     * @return the owner of the consumer, or null if it has not yet been defined
+     */
+    public OwnerDTO getOwner() {
+        return this.owner;
+    }
+
+    /**
+     * Sets the owner to set on this ConsumerDTO object.
+     *
+     * @param owner the owner to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setOwner(OwnerDTO owner) {
+        this.owner = owner;
+        return this;
+    }
+
+    /**
+     * Retrieves the environment field of this ConsumerDTO object.
+     *
+     * @return the environment of the consumer, or null if it has not yet been defined
+     */
+    public EnvironmentDTO getEnvironment() {
+        return this.environment;
+    }
+
+    /**
+     * Sets the environment to set on this ConsumerDTO object.
+     *
+     * @param environment the environment to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setEnvironment(EnvironmentDTO environment) {
+        this.environment = environment;
+        return this;
+    }
+
+    /**
+     * Retrieves the entitlement count field of this ConsumerDTO object.
+     *
+     * @return the entitlementCount of the consumer, or null if it has not yet been defined
+     */
+    public Long getEntitlementCount() {
+        return this.entitlementCount;
+    }
+
+    /**
+     * Sets the entitlement count to set on this ConsumerDTO object.
+     *
+     * @param entitlementCount the entitlement count to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setEntitlementCount(Long entitlementCount) {
+        this.entitlementCount = entitlementCount;
+        return this;
+    }
+
+    /**
+     * Retrieves the facts field of this ConsumerDTO object.
+     *
+     * @return the facts of the consumer, or null if it has not yet been defined
+     */
+    @HateoasArrayExclude
+    public Map<String, String> getFacts() {
+        return this.facts != null ? new MapView<String, String>(facts) : null;
+    }
+
+    /**
+     * Sets the facts to set on this ConsumerDTO object.
+     *
+     * @param facts the facts to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setFacts(Map<String, String> facts) {
+        if (facts != null) {
+            if (this.facts == null) {
+                this.facts = new HashMap<String, String>();
+            }
+            else {
+                this.facts.clear();
+            }
+
+            this.facts.putAll(facts);
+        }
+        else {
+            this.facts = null;
+        }
+        return this;
+    }
+
+    /**
+     * Sets a fact to set on this ConsumerDTO object.
+     *
+     * @param key the key of the fact to set on this ConsumerDTO object.
+     * @param value the value of the fact to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setFact(String key, String value) {
+        if (key == null) {
+            throw new IllegalArgumentException("key is null");
+        }
+
+        if (facts == null) {
+            facts = new HashMap<String, String>();
+        }
+
+        this.facts.put(key, value);
+        return this;
+    }
+
+    /**
+     * Retrieves the last checkin field of this ConsumerDTO object.
+     *
+     * @return the lastCheckin of the consumer, or null if it has not yet been defined
+     */
+    public Date getLastCheckin() {
+        return this.lastCheckin;
+    }
+
+    /**
+     * Sets the last checkin date to set on this ConsumerDTO object.
+     *
+     * @param lastCheckin the last checkin date to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setLastCheckin(Date lastCheckin) {
+        this.lastCheckin = lastCheckin;
+        return this;
+    }
+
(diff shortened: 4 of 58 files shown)


server/src/main/java/org/candlepin/dto/ModelTranslator.java
```diff
@@ -160,6 +160,30 @@
...
+     *  The non null collection to populate the translated entities with.
+     *
+     * #throws TranslationException
+     *  if a translator cannot be found for the input object or any of its nested objects
+     *
+     * @return
+     *  the same collection passed as argument outputCollection, populated with translated DTO objects, if
+     *  any.
+     */
+    <I, O> Collection<O> translate(Collection<I> input, Class<O> outputClass, Collection<O> outputCollection);
```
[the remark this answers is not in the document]
vritant: @Ceiu ,
I agree with you that:
 * this method is used only in two instances of all entities that have been ported to DTOs yet.
 * the intent of this method with its current implementation was convenience only, not efficiency.
 * the code savings is only a handful of lines
 * we do not have a critical need of this at present
 * also I agree, to improve the efficiency of this method is non-trivial.

I'd like to add that:
 * we do have a lot of APIs which return collections in candlepin over all, so there is scope for more use in currently implemented DTOs and the remaining entities.
 * while the lines saved currently is very less, this does not obstruct work in future to implement this with more efficiency, this implementation does not hurt us, as the new implementation could simply replace current implementation if/when it is implemented.

May I suggest we add a TODO comment on the implementation in this PR to capture the points you mentioned?

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,985 @@
...
+    }
+
+    /**
+     * Initializes a new ConsumerDTO. convenience constructor for tests.
+     * @param name the consumer name to set on the consumer
+     * @param userName the username to set on the consumer
+     * @param owner the owner to set on the consumer
+     * @param type the consumer type to set on the consumer
+     */
+    public ConsumerDTO(String name, String userName, OwnerDTO owner, ConsumerTypeDTO type) {
```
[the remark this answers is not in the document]
vritant: Currently, the only use is in tests.

server/src/main/java/org/candlepin/dto/api/v1/CapabilityDTO.java
```diff
@@ -0,0 +1,158 @@
...
+        this.name = name;
+        return this;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String toString() {
+
```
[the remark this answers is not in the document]
vritant: removed

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,985 @@
...
+
+    /**
+     * Retrieves the releaseVer field of this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    @JsonSerialize(using = ConsumerDTO.ReleaseVersionWrapSerializer.class,
+        nullsUsing = ConsumerDTO.ReleaseVersionWrapSerializer.class)
+    @JsonProperty("releaseVer")
+    public String getReleaseVer() {
```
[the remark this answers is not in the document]
vritant: done

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,985 @@
...
+    /**
+     * Sets the releaseVer to set on this ConsumerDTO object.
+     *
+     * @param releaseVer the release version to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    @JsonDeserialize(using = ConsumerDTO.ReleaseVersionWrapDeserializer.class)
+    @JsonProperty("releaseVer")
+    public ConsumerDTO setReleaseVer(String releaseVer) {
```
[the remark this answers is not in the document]
vritant: done

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,985 @@
...
+    }
+
+    /**
+     * Retrieves the facts field of this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    @HateoasArrayExclude
+    public Map<String, String> getFacts() {
+        return facts;
```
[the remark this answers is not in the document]
vritant: will fix.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,985 @@
...
+
+    /**
+     * Sets the facts to set on this ConsumerDTO object.
+     *
+     * @param facts the facts to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setFacts(Map<String, String> facts) {
+        this.facts = facts;
```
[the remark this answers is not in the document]
vritant: will fix.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,985 @@
...
+        return this;
+    }
+
+    /**
+     * Retrieves the consumer capabilities field of this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public Set<CapabilityDTO> getCapabilities() {
+        return this.capabilities;
```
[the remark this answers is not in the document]
vritant: will fix.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,985 @@
...
+
+    /**
+     * Sets the capabilities on this ConsumerDTO object.
+     *
+     * @param capabilities the capabilities to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setCapabilities(Set<CapabilityDTO> capabilities) {
+        this.capabilities = capabilities;
```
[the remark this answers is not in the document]
vritant: will fix.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,985 @@
...
+        return this;
+    }
+
+    /**
+     * This will put in the property so that the virtWho instances won't error
+     *
+     * @return List always empty
+     */
+    @JsonProperty("guestIds")
+    public List<String> getEmptyGuestIdArray() {
```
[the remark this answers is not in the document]
vritant: per the comment on the method in the existing model code, the need is virt-who clients.
This works for me, but please share alternative suggestions.

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /**
+     * Retrieves the attributes field of this GuestIdDTO object.
+     *
+     * @return the attributes field of this GuestIdDTO object.
+     */
+    @HateoasArrayExclude
+    public Map<String, String> getAttributes() {
+        return attributes;
```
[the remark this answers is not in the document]
vritant: will fix.

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * Sets the attributes on this GuestIdDTO object.
+     *
+     * @param attributes the attributes to set on this GuestIdDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public GuestIdDTO setAttributes(Map<String, String> attributes) {
+        this.attributes = attributes;
```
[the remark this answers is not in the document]
vritant: will fix.

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public GuestIdDTO clone() {
+        GuestIdDTO copy = super.clone();
+        Map<String, String> attributes = this.getAttributes();
+        if (attributes != null) {
+            copy.setAttributes(new HashMap<String, String>(attributes));
```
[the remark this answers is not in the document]
vritant: will fix.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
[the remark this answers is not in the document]
Ceiu: What I mean is that if we need to always return a guest ID collection, maybe we need to update the DTO accordingly. This may be a good candidate for discussion when @cnsnyder returns so we can determine exactly what is needed, why it's needed and the cleanest way forward.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -430,7 +441,143 @@ public Consumer getConsumer(
...
+
+        if (dto.getCapabilities() != null) {
+            Set<ConsumerCapability> capabilities = new HashSet<ConsumerCapability>();
+            for (CapabilityDTO capabilityDTO : dto.getCapabilities()) {
+                if (capabilityDTO != null) {
+                    capabilities.add(new ConsumerCapability(entity, capabilityDTO.getName()));
+                }
+            }
+            entity.setCapabilities(capabilities);
+        }
```
Ceiu: Lot of refactoring here to be done with the encapsulation changes

server/src/main/java/org/candlepin/resource/GuestIdResource.java
```diff
@@ -157,38 +214,41 @@ public void updateGuest(
         @PathParam("consumer_uuid") @Verify(Consumer.class) String consumerUuid,
         @ApiParam("guest virtual uuid")
         @PathParam("guest_id") String guestId,
-        @ApiParam(name = "updated", required = true, value = "updated guest data to use") GuestId updated) {
+        @ApiParam(name = "updated", required = true, value = "updated guest data to use")
+        GuestIdDTO updatedDTO) {
```
Ceiu: Don't need to wrap here

server/src/test/java/org/candlepin/dto/AbstractDTOTest.java
```diff
@@ -97,6 +102,16 @@ public T getDTOInstance() {
         }
     }
 
+    public void addCollectionMethod(Class<T> dtoClass, String field, String prefix,
```
Ceiu: I'm not opposed to generalizing all of the collection-based tests, but I'd like to see this cleaned up a bit and not as hacky as it is. Right now the field name is somewhat misused to fetch a value using one of two values, and the method lookup is dependent on a very specific pluralization, which may not always hold true. I.e. content is already plural, some pluralization might require an -es, -ies or just an -s suffix.

More work needs to be done here to make this work cleanly, and is a large part of why it wasn't done initially.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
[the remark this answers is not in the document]
Ceiu: After starting on the second pass of the PoolDTO PR, I've noted that I positively responded to additional constructors there.

Seems like we (I) need to figure out which way we want to go on that. Do we want extra constructors, or should we limit it to only constructors for populating a DTO/entity with its required fields. If we go with the latter, we get the bonus of being able to slap a ```@JsonCreator``` annotation on it for slightly faster/cleaner creation.

server/src/main/java/org/candlepin/dto/ModelTranslator.java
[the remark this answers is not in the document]
vritant: I.m.o. this particular instance is a simple implementation that could save some lines of code, 
I will remove it never the less.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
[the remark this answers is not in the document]
vritant: * Based on the comment in the code, my understanding is that the absence of this collection would break (at least) older version of virt-who clients. Also we should ensure the dto work does not change the output of the APIs as they are today.
* we do need a getter that returns the actual guest ids, at the very least for test purposes, also to populate the entity itself.
* I like the idea to skip translation. translation seems like a waste anyway, so I changed the implementation to not translate guestIds.

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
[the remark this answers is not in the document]
vritant: fixed

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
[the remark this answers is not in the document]
vritant: fixed

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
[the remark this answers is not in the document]
vritant: fixed

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,984 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @HateoasInclude
+    public String getHref() {
+        return this.uuid != null ? String.format("/consumers/%s", this.uuid) : null;
+    }
+
+    public void setHref(String href) {
```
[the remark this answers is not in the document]
vritant: removed it entirely.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,984 @@
...
+            }
+        }
+
+        Set<CapabilityDTO> capabilities = this.getCapabilities();
+        if (capabilities != null) {
+            copy.capabilities = new HashSet<CapabilityDTO>();
+            for (CapabilityDTO capabilityDTO: capabilities) {
+                copy.capabilities.add(capabilityDTO.clone());
+            }
+        }
```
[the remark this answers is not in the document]
vritant: removed

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentDTO.java
```diff
@@ -0,0 +1,260 @@
...
+
+    /**
+     * Sets the environmentCOntent to set on this EnvironmentDTO object.
+     *
+     * @param environmentContent the environementContent to set on this EnvironmentDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public EnvironmentDTO setEnvironmentContent(Set<EnvironmentContentDTO> environmentContent) {
+        this.environmentContent = environmentContent;
```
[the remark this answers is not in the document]
vritant: done

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentTranslator.java
```diff
@@ -0,0 +1,95 @@
...
+                    if (envContent != null) {
+                        EnvironmentContentDTO dto = envContentTranslator.translate(translator, envContent);
+                        if (dto != null) {
+                            dest.getEnvironmentContent().add(dto);
+                        }
+                    }
+                }
+            }
+        }
+        else {
```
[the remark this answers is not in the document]
vritant: perhaps I am missing something obvious. will chat with you online.

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
```diff
@@ -0,0 +1,208 @@
...
+        return this;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String toString() {
+        return String.format(
+            "GuestIdDTO [id: %s, guestId: %s]",
```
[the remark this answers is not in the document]
vritant: done

server/src/main/java/org/candlepin/resource/ConsumerResource.java
vritant: im missing something obvious here as well. will chat with you online

server/src/main/java/org/candlepin/resource/GuestIdResource.java
vritant: we do.
resource/GuestIdResource.java:217: Line is longer than 110 characters (found 114). [LineLength]

server/src/test/java/org/candlepin/dto/AbstractDTOTest.java
vritant: I hear you. noting this as a topic of discussion.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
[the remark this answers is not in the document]
vritant: noted as a point for discussion

server/src/main/java/org/candlepin/dto/SimpleModelTranslator.java
```diff
@@ -310,6 +312,29 @@ private Class findNearestMappedInterface(Class source, Set<Class> mappedClasses)
         return output;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public <I, O> Collection<O> translate(Collection<I> inputCollection, Class<O> outputClass,
```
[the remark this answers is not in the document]
vritant: removed

server/src/test/java/org/candlepin/dto/AbstractDTOTest.java
Ceiu: One way of approaching this problem is changing the detection to use the prefix (add/remove/get) with a similar noun (i.e. Products or Content), and then add them to a separate list of things to test (as you're doing)

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
```diff
@@ -0,0 +1,216 @@
...
+    protected String id;
+    protected String guestId;
+    protected Map<String, String> attributes;
+
+    /**
+     * Initializes a new GuestIdDTO instance with null values.
+     */
+    public GuestIdDTO() {
+    }
+
```
nikosmoum: Probably a good idea to add a comment saying what this constructor is used for (e.g. tests).

server/src/main/java/org/candlepin/dto/api/v1/GuestIdTranslator.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.candlepin.dto.ModelTranslator;
+import org.candlepin.model.GuestId;
+
+/**
+ * The GuestIdTranslator provides translation from GuestIdCapability model objects to
+ * GuestIdDTOs
+ */
+public class GuestIdTranslator extends
+    TimestampedEntityTranslator<GuestId, GuestIdDTO> {
```
nikosmoum: No need to break, since this fits on a single line.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * The ConsumerTranslator provides translation from Consumer model objects to
+ * ConsumerDTOs
+ */
+public class ConsumerTranslator extends
+    TimestampedEntityTranslator<Consumer, ConsumerDTO> {
```
nikosmoum: No need to break, since this fits on a single line.

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentTranslator.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * The CapabilityTranslator provides translation from ConsumerCapability model objects to
+ * CapabilityDTOs
+ */
+public class EnvironmentTranslator extends
+    TimestampedEntityTranslator<Environment, EnvironmentDTO> {
```
nikosmoum: No need to break, since this fits on a single line.

server/src/main/java/org/candlepin/dto/api/v1/HypervisorIdTranslator.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.candlepin.dto.ModelTranslator;
+import org.candlepin.model.HypervisorId;
+
+/**
+ * The HypervisorIdTranslator provides translation from HypervisorId model objects to
+ * HypervisorIdDTOs
+ */
+public class HypervisorIdTranslator extends
+    TimestampedEntityTranslator<HypervisorId, HypervisorIdDTO> {
```
nikosmoum: No need to break, since this fits on a single line.

server/src/main/java/org/candlepin/dto/api/v1/HypervisorIdDTO.java
```diff
@@ -0,0 +1,182 @@
...
+
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+
+import io.swagger.annotations.ApiModel;
+
+/**
+ * A DTO representation of the HypervisorId entity
+ */
+@ApiModel(parent = TimestampedCandlepinDTO.class, description = "DTO representing a consumer capability")
```
nikosmoum: Wrong API description.

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
```diff
@@ -0,0 +1,216 @@
...
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlRootElement;
+
+/**
+ * A DTO representation of the GuestId entity
+ */
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.PROPERTY)
+@ApiModel(parent = TimestampedCandlepinDTO.class, description = "DTO representing a consumer capability")
```
nikosmoum: Wrong API description.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,1030 @@
...
+        this.entitlementCount = 0L;
+        this.facts = new HashMap<String, String>();
+        this.installedProducts = new HashSet<ConsumerInstalledProductDTO>();
+        this.guestIds = new ArrayList<GuestIdDTO>();
+    }
+
+    /**
+     * Retrieves the id field of this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
```
nikosmoum: Wrong return type comment.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,1030 @@
...
+     */
+    public ConsumerDTO setId(String id) {
+        this.id = id;
+        return this;
+    }
+
+    /**
+     * Retrieves the uuid field of this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
```
nikosmoum: Wrong return type comment.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -0,0 +1,1030 @@
...
+     */
+    public ConsumerDTO setUuid(String uuid) {
+        this.uuid = uuid;
+        return this;
+    }
+
+    /**
+     * Retrieves the name field of this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
```
nikosmoum: Wrong return type comment. I won't add more comments for this, but there are quite a few methods (mostly getters) with this comment in this class that don't actually return a ConsumerDTO.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerInstalledProductDTO.java
```diff
@@ -0,0 +1,316 @@
...
+     */
+    public ConsumerInstalledProductDTO setStartDate(Date startDate) {
+        this.startDate = startDate;
+        return this;
+    }
+
+    /**
+     * Retrieves the endDate field of this ConsumerInstalledProduct object.
+     *
+     * @return a reference to this DTO object.
```
nikosmoum: Wrong return type comment.(same in other methods of this class)

server/src/main/java/org/candlepin/resource/ConsumerTypeResource.java
```diff
@@ -151,7 +151,7 @@ public ConsumerTypeDTO create(
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     public ConsumerTypeDTO update(
-        @ApiParam(name = "consumerType", required = true) ConsumerTypeDTO dto) throws BadRequestException {
+        @ApiParam(name = "type", required = true) ConsumerTypeDTO dto) throws BadRequestException {
```
nikosmoum: Why change the name of the API's property on the swagger output? Accident due to mass refactor/rename with IDE? Same with the other two changes in this file.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -372,7 +372,7 @@ public void setOwner(Owner owner) {
     public String toString() {
         String consumerType = (this.getType() != null) ? this.getType().getLabel() : "null";
 
-        return String.format("Consumer [id: %s, uuid: %s, consumerType: %s, name: %s]",
+        return String.format("Consumer [id: %s, uuid: %s, type: %s, name: %s]",
```
nikosmoum: Why change the name the property? Accident due to mass refactor/rename with IDE?

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -577,7 +577,7 @@ public Consumer getHost(String guestId, Owner... owners) {
      * This is an insecure query, because we need to know whether or not the
      * consumer exists
      *
-     * We do not require that the hypervisor be consumerType hypervisor
+     * We do not require that the hypervisor be type hypervisor
```
nikosmoum: Why change this? Accident due to mass refactor/rename with IDE?

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentDTO.java
```diff
@@ -0,0 +1,276 @@
...
+            String thatOid = that.getOwner() != null ? that.getOwner().getId() : null;
+
+            EqualsBuilder builder = new EqualsBuilder()
+                .append(this.getId(), that.getId())
+                .append(this.getName(), that.getName())
+                .append(this.getDescription(), that.getDescription())
+                .append(thisOid, thatOid);
+
+            return builder.isEquals() &&
+                Util.collectionsAreEqual(this.getEnvironmentContent(), that.getEnvironmentContent());
```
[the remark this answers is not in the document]
vritant: @nikosmoum I kept the implementation similar to ProductDTO equals, which does similar computation on content. thoughts?

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
vritant: I assume it was the other constructor. ill add the comment.

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentDTO.java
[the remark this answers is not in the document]
nikosmoum: crog mentioned this on the PoolDTO so I thought that's the way we do it. Not sure if if ProductDTO is an exception for some reason, or if it was missed. @Ceiu ^?

server/src/main/java/org/candlepin/model/Consumer.java
vritant: probably. undo-ing

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentDTO.java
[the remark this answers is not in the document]
Ceiu: ProductDTO uses the collectionsAreEqual on the sets because they're sets of ProductContentDTO objects, which themselves do the ID comparison. 

As an aside, we're looking at the UUIDs in ProductContentDTO, which might be wrong. It probably needs to look at UUIDs if they exist, and then fall back to IDs. More thought is needed there, I think.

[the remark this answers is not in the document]
nikosmoum: Actually, I just noticed that I implemented that suggestion wrongly on my PR :/

vritant:
[[8] Add API DTO for Consumer](https://trello.com/c/jthiyML4/635-8-add-api-dto-for-consumer)

Ceiu:
In the interest of time, I think we should pull the abstract collection stuff out, and just test the collections normally in their respective test classes. I still think abstract collection tests have value, and we should do it eventually, but I want more time to plan it out and implement something that is easy to use and fits well with the rest of the abstract DTO test framework.

vritant:

server/src/main/java/org/candlepin/dto/api/v1/ConsumerInstalledProductDTO.java
```diff
@@ -0,0 +1,316 @@
...
+            .append(this.getEndDate());
+
+        return builder.toHashCode();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public ConsumerInstalledProductDTO clone() {
```
Ceiu: Need to clone the date objects here.

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentContentDTO.java
```diff
@@ -0,0 +1,188 @@
...
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+
+import io.swagger.annotations.ApiModel;
+
+/**
+ * A DTO representation of the Environment Content entity
+ */
+@ApiModel(parent = TimestampedCandlepinDTO.class, description = "DTO representing an environment content")
+public class EnvironmentContentDTO extends TimestampedCandlepinDTO<EnvironmentContentDTO> {
```
Ceiu: This class could probably be a private class within EnvironmentDTO. We could also drop its ID and timestamps from the output, as none of them are important to end users. See ProductDTO.ProductContentDTO for an similar example.

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentDTO.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public EnvironmentDTO clone() {
+        EnvironmentDTO copy = super.clone();
+        copy.owner = owner != null ? owner.clone() : null;
+        copy.setEnvironmentContent(this.getEnvironmentContent());
```
Ceiu: This is only acceptable if, and only if, the EnvironmentContentDTO is immutable (which it does not appear to be). This is okay if the implementation there is changed to closer match ProductContentDTO.

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /**
+     * Sets the attributes on this GuestIdDTO object.
+     *
+     * @param attributes the attributes to set on this GuestIdDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public GuestIdDTO setAttributes(Map<String, String> attributes) {
```
Ceiu: Add TODOs denoting the absence of the add/remove attribute methods and the intention to add them if we have a need in the future.

server/src/main/java/org/candlepin/model/GuestId.java
```diff
@@ -163,10 +166,12 @@ public boolean equals(Object other) {
         if (!(other instanceof GuestId)) {
             return false;
         }
+
         GuestId that = (GuestId) other;
-        if (this.getGuestId().equalsIgnoreCase(that.getGuestId()) &&
-            this.getAttributes().equals(that.getAttributes())) {
-            return true;
+        if (this.getGuestId().equalsIgnoreCase(that.getGuestId())) {
+            EqualsBuilder builder = new EqualsBuilder()
```
Ceiu: This is just a map of strings, and the EqualsBuilder is just calling into .equals as the replaced code was doing. We're not gaining anything with this change.

server/src/main/java/org/candlepin/model/GuestIdCurator.java
```diff
@@ -35,14 +30,11 @@ public GuestIdCurator() {
         super(GuestId.class);
     }
 
-    public Page<List<GuestId>> listByConsumer(Consumer consumer, PageRequest pageRequest) {
-        Criteria criteria = this.currentSession().createCriteria(GuestId.class)
+    public CandlepinQuery<GuestId> listByConsumer(Consumer consumer) {
+        DetachedCriteria criteria = this.createSecureDetachedCriteria()
```
Ceiu: :+1:

server/src/main/java/org/candlepin/resource/GuestIdResource.java
Ceiu: ack

server/src/main/java/org/candlepin/resource/GuestIdResource.java
```diff
@@ -167,38 +224,41 @@ public void updateGuest(
...
-                    updated.getGuestId(), guestId));
+                    updatedDTO.getGuestId(), guestId));
         }
 
         Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
-        updated.setConsumer(consumer);
+        GuestId guestIdEntity = new GuestId();
+        populateEntity(guestIdEntity, updatedDTO);
+        guestIdEntity.setConsumer(consumer);
         GuestId toUpdate = guestIdCurator.findByGuestIdAndOrg(guestId, consumer.getOwner());
```
Ceiu: This could be done in such a way that we aren't creating a new entity. Lookup the actual DB entity, apply the changes directly to it, then merge. That way we don't have to juggle two instances of the entity and don't waste time allocating a new instance.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -664,8 +651,48 @@ else if (ownerDto.getKey() != null) {
...
+        Owner owner = null;
+        if (ownerDto.getId() != null) {
+            // look up by ID
+            owner = this.ownerCurator.find(ownerDto.getId());
+        }
+        else if (ownerDto.getKey() != null) {
+            // look up by key
+            owner = this.ownerCurator.lookupByKey(ownerDto.getKey());
+        }
+        if (owner == null) {
```
Ceiu: Whitespace between block structures. The lack of space makes them flow together at a glance.

Ceiu:
**ConsumerDTO.java:852**
Use this.getUuid() rather than this.uuid so subclasses don't need to override .getHref if .getUuid is changed.

**ConsumerResource.java:482**
We don't allow clients to set the UUID. Possibly other fields here, too.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerInstalledProductDTO.java
vritant: done

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentContentDTO.java
vritant: done

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentDTO.java
vritant: changed to mimic ProductContentDTO

server/src/main/java/org/candlepin/dto/api/v1/GuestIdDTO.java
vritant: done

server/src/main/java/org/candlepin/model/GuestId.java
vritant: done

server/src/main/java/org/candlepin/resource/GuestIdResource.java
vritant: done

vritant:
> ConsumerDTO.java:852
Use this.getUuid() rather than this.uuid so subclasses don't need to override .getHref if .getUuid is changed.

done
> ConsumerResource.java:482
We don't allow clients to set the UUID. Possibly other fields here, too.

surprisingly enough, we do allow clients to set the UUID. any other specific fields you have in mind?

server/src/main/java/org/candlepin/resource/GuestIdResource.java
vritant: walking back on this one per our conversation online.

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentDTO.java
```diff
@@ -0,0 +1,408 @@
...
+     * @param dto
+     *  The product content DTO to add to this product
+     *
+     * @throws IllegalArgumentException
+     *  if content is null
+     *
+     * @return
+     *  true if adding the content resulted in a change to this product; false otherwise
+     */
+    public boolean addContent(ContentDTO dto, boolean enabled) {
```
Ceiu: This needs to accept a null enabled value, as this field is used as an environment override only if it's not null.

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentTranslator.java
```diff
@@ -0,0 +1,113 @@
...
+
+                    for (EnvironmentContent ec : envContent) {
+                        if (ec != null) {
+                            ContentDTO dto = contentTranslator.translate(translator, ec.getContent());
+
+                            if (dto != null) {
+                                log.debug("VRITANT {}", dest);
+                                log.debug("VRITANT {}", dto);
+                                log.debug("VRITANT {}", ec.getEnabled());
+                                dest.addContent(dto, ec.getEnabled().booleanValue());
```
Ceiu: Looked into this a bit more and it's as you thought: ec.getEnabled is returning null, which is causing the problem. For environment content, we need to accept null as a value, so once EnvironmentContentDTO is updated accordingly, we can drop the .booleanValue from this and everything should be good.

server/src/main/java/org/candlepin/resource/GuestIdResource.java
Ceiu: ack

vritant: walking back on this one per our conversation online.

Ceiu:
ack

PR #1865, opened 2017-12-22
[DTO] Fixed content detection in products

vritant:
what was changed and why:
* addContentToUpstreamSubscriptions was removed, which breaks two tests in entitlement_certificate_v3_spec.rb because the upstream subscriptions were not being updated with the new content, as is currently done in master. this PR adds it back.
* we added a ProductContentDTO comparator in ProductManager but never used it, fixed in this PR.
* ProductManager simply uses ContentManager.isChangedBy. uuid comparision for content is now skipped ( as it is in master ) which fixes the spec tests in refresh_pools_spec.rb
* to verify it was broken, compare results in target branch and PR branch : 
  * deploy in hosted mode
  * buildr rspec:refresh_pools:"regenerates entitlements when"

### DIFF
--- a/server/src/main/java/org/candlepin/controller/ContentManager.java
+++ b/server/src/main/java/org/candlepin/controller/ContentManager.java
@@ -726,7 +726,7 @@ public class ContentManager {
      * @return
      *  true if this content would be changed by the given DTO; false otherwise
      */
-    public boolean isChangedBy(Content entity, ContentDTO dto) {
+    public static boolean isChangedBy(Content entity, ContentDTO dto) {
         if (dto.getId() != null && !dto.getId().equals(entity.getId())) {
             return true;
         }
@@ -800,7 +800,7 @@ public class ContentManager {
      * @return
      *  true if this content would be changed by the given DTO; false otherwise
      */
-    public boolean isChangedBy(Content entity, ContentData dto) {
+    public static boolean isChangedBy(Content entity, ContentData dto) {
         if (dto.getId() != null && !dto.getId().equals(entity.getId())) {
             return true;
         }
--- a/server/src/main/java/org/candlepin/controller/ProductManager.java
+++ b/server/src/main/java/org/candlepin/controller/ProductManager.java
@@ -740,7 +740,7 @@ public class ProductManager {
      * @return
      *  true if this product would be changed by the given DTO; false otherwise
      */
-    private boolean isChangedBy(Product entity, ProductDTO dto) {
+    public static boolean isChangedBy(Product entity, ProductDTO dto) {
         // Check simple properties first
         if (dto.getId() != null && !dto.getId().equals(entity.getId())) {
             return true;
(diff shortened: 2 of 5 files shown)


vritant:
@nikosmoum this is the fix for the problem you pinged me about.

Ceiu:
NACK

This is a problem in how the current hosted test juggles data, not with how the code itself works. This operates properly when built on the refactored hosted test bits (comparator addition aside).

Ceiu:
@vritant Alright, so after some thought, here's how I would like to handle this issue in this branch:

- Update *all* of the isChangedBy methods to be public static (\*grumble\*)
- Update the ProductManager.isChangedBy methods (both of them) such that the content check uses the block given below [1]
- Rearrange the code in the ProductManager such that both isChangedBy methods are at the bottom of the class, rather than between some of the applyProductChanges methods
- Update the existing calls to isChangedBy that reference it locally via ```this.isChangedBy``` to remove the ```this.``` component

[1]
```
                        if (content != null && cdto != null) {
                            if (cdto.getUuid() != null ?
                                cdto.getUuid().equals(content.getUuid()) :
                                (cdto.getId() != null && cdto.getId().equals(content.getId()))) {
                                // At this point, we've either matched the UUIDs (which means we're
                                // referencing identical products) or the UUID isn't present on the DTO, but
                                // the IDs match (which means we're pointing toward the same product).

                                return (update.isEnabled() != null &&
                                    !update.isEnabled().equals(existing.isEnabled())) ||
                                    ContentManager.isChangedBy(content, cdto) ? 1 : 0;
                            }
                        }
```

vritant:
@Ceiu  noting it here so I don't forget, but we can chat on bjeans.
the discrepancy there is that productManager.isChangedBy does not look for uuid changes, but if we go this route, we will be looking for uuid changes in content.

this causes test failures in this branch.

if we start looking for uuid changes in content in master, imo we would get the same failures there as well.

one of the test that fails:
buildr rspec:refresh:"regenerates entitlements when modifiedProductIds of content change"

nikosmoum:
This should probably get merged before the PoolDTO/ConsumerDTO stuff, to make the tests pass?

Ceiu:
@vritant The hosted test adapter needs to clear any product or content UUIDs it receives. Once that's done, the changes requested above work as expected.

vritant:
@Ceiu  I made the changes in this PR and  #1881 for master , please review that as well.

vritant:
[[8] Add API DTO for Consumer](https://trello.com/c/jthiyML4/635-8-add-api-dto-for-consumer)

Ceiu:
ACK

vritant:
@Ceiu the tests failed because the rebase on consumer_dto added lastRefreshed field to the model but the dto work for that was missing.

I fixed this, also added #1884  so after every re-base/merge we can see if anything is broken in the feature branch.

vritant:

Ceiu:
@vritant #1882 already addresses the Owner DTO stuff

Ceiu:
ACK... for real this time.

PR #1866, opened 2017-12-22
mark changesMade on name change in performConsumerUpdates

vritant:
* To verify the fix, run updatedNameRegeneratesIdCert unit test before and after the fix.
* The current test is not ideal because it relies on the object being manipulated in the method, improved it in dtos branch.


Ceiu:
ack

PR #1867, opened 2018-01-02
Update Owner with date of last refresh.

awood:


Ceiu:
ACK pending successful test runs

PR #1868, opened 2018-01-04
1525626: Environment content changes in Golden Ticket

wottop:


wottop:
[1525626 - when content in environment changes, content_access cert is not updated](https://trello.com/c/JwMFEh1q/703-1525626-when-content-in-environment-changes-contentaccess-cert-is-not-updated)

PR #1873, opened 2018-01-05
[m] 1525627: Environment content changes in Golden Ticket

wottop:


wottop:
[1525626 - when content in environment changes, content_access cert is not updated](https://trello.com/c/JwMFEh1q/703-1525626-when-content-in-environment-changes-contentaccess-cert-is-not-updated)

wottop:
retest this please

server/src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -362,6 +368,12 @@ public JobDetail demoteContent(
         return contentIds;
     }
 
+    @Transactional
+    private void clearContentAccessCerts(Environment env) {
+        ownerEnvContentAccessCurator.removeAllForEnvironment(env.getId());
+        ownerEnvContentAccessCurator.refresh();
```
awood: How does this work?  I don't see a specific `refresh()` method in OwnerEnvContentAccessCurator, so I'm supposing it goes to `refresh(E... entities)` in the superclass.  But since you're not passing in any entities, that looks like it would just be a no-op.

wottop: I believe it clears the cache for the curator. If it is not called, the next query through the curator returns the old versions and the content changes don't get expressed.

PR #1874, opened 2018-01-10
[M] 1531106: Round 4 of ueber data cleanup.

vritant:
Since pre-org migration is now stricter,
we need to clean up ueber certs while the data is in
pre migrated format. the sql here is identical to
round 3 in 0.9.54

the scenario that breaks: upgrade from 0.9.54.X where X < 18 to 2.1, master, 2.2.
for testing steps, please use comments in #1459 

reviewer: please verify upgrade from 0.9.54.17 to 0.9.54-HOTFIX as well, and  confirm my finding that we don't have an issue there.

### DIFF
--- a/server/src/main/resources/db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml
+++ b/server/src/main/resources/db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml
@@ -1,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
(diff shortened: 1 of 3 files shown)


awood:
Also I had trouble running this as noted in #1878

mstead:
> Liquibase still uses the expanded variables for checksum calculation, so putting each query into a separate variable doesn't actually save us work or clean things up. I would prefer that we just have two changesets for each block that differs.

Another option would be to use the single changeset, but use the 'dbms' option on the SQL tag. I've no preference on either approach.

`<sql dbms="postgres"> ... </sql>`
... other postgres queries ...

`<sql dbms="mysql"> ... </sql>`
... other mysql queries ...

awood:
Works for me with both postgres and mysql.

Ceiu:
ACK

PR #1877, opened 2018-01-10
[2.1] 1533211: Set the default enabled value for Environment Content to Null

barnabycourt:
We depend on the database value of null, true, or false, to determine
whether the environment has overridden the default value. When the
DTO was added in Candlepin 2.x it was set to true by default and thus
the environment was always overriding the value from the engineering
product.


awood:
👍

PR #1878, opened 2018-01-11
[2.1] 1531107: Round 4 of ueber data cleanup.

vritant:
Since pre-org migration is now stricter,
we need to clean up ueber certs while the data is in
pre migrated format. the SQL here is identical to
round 3 in 0.9.54

### DIFF
--- a/server/src/main/resources/db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml
+++ b/server/src/main/resources/db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml
@@ -1,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <!-- POSTGRES, ORACLE, HSQLDB QUERIES -->
+    <property
+            dbms="postgresql,oracle,hsqldb"
+            name="round4_premigration_cleanup_ueber_subs"
+            value="delete from cp_subscription where id in (select s.id from cp_subscription s inner join cp_product p on s.product_id = p.id where p.name LIKE '%_ueber_product');" />
+
+    <property
+            dbms="postgresql,oracle,hsqldb"
+            name="round4_premigration_cleanup_ueber_ent_certs"
+            value="delete from cp_ent_certificate where entitlement_id in (select e.id from cp_entitlement e inner join cp_pool p on e.pool_id = p.id where p.productname LIKE '%_ueber_product');" />
+
+    <property
+            dbms="postgresql,oracle,hsqldb"
+            name="round4_premigration_cleanup_ueber_ents"
+            value="delete from cp_entitlement where pool_id in (select id from cp_pool p where p.productname LIKE '%_ueber_product');" />
+
+    <property
+            dbms="postgresql,oracle,hsqldb"
+            name="round4_premigration_cleanup_ueber_pools"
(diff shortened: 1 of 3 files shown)


server/src/main/resources/db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml
```diff
@@ -0,0 +1,100 @@
...
+
+    <property
+            dbms="mysql"
+            name="round4_premigration_cleanup_ueber_products"
+            value="delete p from cp_product p where p.name LIKE '%_ueber_product';" />
+
+    <property
+            dbms="mysql"
+            name="round4_premigration_cleanup_ueber_consumers"
+            value="delete c from cp_consumer c where c.name = 'ueber_cert_consumer';" />
```
awood: Why can't we let Liquibase do this for use?

```xml
<delete tablename="cp_consumer">
  <where>name = 'ueber_cert_consumer'</where>
</delete>
```

I tested this on Postgres and MySQL and it worked.

Postgres: 
```
DEBUG 1/11/18 2:03 PM: liquibase: db/changelog/changelog-update.xml: db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml::20180110092945-1::vrjain: Executing EXECUTE database command: DELETE FROM public.cp_consumer WHERE name = 'ueber_cert_consumer'
```

MySQL: 
```
DEBUG 1/11/18 2:01 PM: liquibase: db/changelog/changelog-update.xml: db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml::20180110092945-1::vrjain: Executing EXECUTE database command: DELETE FROM candlepin.cp_consumer WHERE name = 'ueber_cert_consumer'
```

server/src/main/resources/db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml
```diff
@@ -0,0 +1,100 @@
...
+
+    <changeSet id="20180110092945-1" author="vrjain">
```
awood: I'm seeing errors when I'm running this.  Here's my steps:

1. Deploy a clean Candlepin with `/bin/deploy -g`
2. Deploy Candlepin (so the update changelog runs) with `/bin/deploy`

I'm seeing errors like that `cp_subscription` doesn't exist and that `cp_pool` doesn't have a `productname` column.  I had to comment out a lot of statements just to test the delete element thing I proposed earlier.

mstead: @vritant This is happening because during  liquibase DB creation the changeset that cleans up the old pre 2.0 tables / fields gets run (20170130144529-perorgproducts-phase-2.xml), but not your new one. On update, it sees that the new one hasn't yet been run yet and explodes.

You could probably use a precondition to get around this or add the changeset to the create changelog (though it seems redundant).

server/src/main/resources/db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml
vritant: @awood I simply copied the existing changelog from 0.9.54-HOTFIX. we could re-write them though. please let me know.

server/src/main/resources/db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml
vritant: @awood nice catch!
@mstead to me it seems cleaner to add preconditions, to do this during create seems like cheating. updated the code and added pre-condition, and verified this does not fail anymore.

also verified this PR still cleans up bad data even with the preconditions.

also added the pre-condition to the other branch as well.

server/src/main/resources/db/changelog/20180110092945-cleanup-uebercerts-round4-premigration.xml
```diff
@@ -0,0 +1,106 @@
...
+        <preConditions onFail="MARK_RAN">
+            <or>
```
awood: Sorry, to be picky, but shouldn't this be an `and`?

PR #1879, opened 2018-01-11
[M] Add version requirement to Gemfile.lock

awood:
`buildr` keeps making this change every time I run it.  I'm not sure why.  I'm using buildr 1.5.3 on Ruby 2.3.4.

The change itself seem innocuous since we already have the `~> 0.17.0` requirement in `Gemfile`.  If no one else is seeing this issue, maybe it's something in my environment.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -114,7 +114,7 @@ DEPENDENCIES
   pry
   pry-byebug
   pry-stack_explorer
-  qpid_proton
+  qpid_proton (~> 0.17.0)
   rest-client (~> 1.6.0)
   rspec (~> 3.0)
   rubocop (= 0.36.0)


awood:

Gemfile.lock
```diff
@@ -114,7 +114,7 @@ DEPENDENCIES
   pry
   pry-byebug
   pry-stack_explorer
-  qpid_proton
+  qpid_proton (~> 0.17.0)
```
barnabycourt: This causes a failure using the vagrant image due to what appears to be a conflict with the rpm installed qpid_proton 0.18 on Centos/7. If I updated both the gemfile & temfile.lock to (~> 0.18) it worked for me with vagrant up.

awood: Okay, I'll just scrap this change then since I'm not precisely sure the underlying reason for it.

awood: I thought we had to stick with 0.17 for some reason?  Fedora is up to 0.18 right now, so I'd be happy to move to it as long as it doesn't break anything.

To be honest, the qpid spec tests fail pretty frequently when I run them locally anyway.  Maybe we could move them to a specific test suite for Jenkins only or make them optional?  This qpid gem has been a chronic pain in the neck for awhile now.

awood:
Closed in favor of #1889

PR #1880, opened 2018-01-12
[M] Ensure unique share consumer between two owners

vritant:


vritant:
[[13] [multi-org] ensure one sharing consumer for two owners](https://trello.com/c/BA5KGWeH/641-13-multi-org-ensure-one-sharing-consumer-for-two-owners)

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -483,8 +483,14 @@ public Consumer create(
             // Share consumers do not need identity certificates so refuse to create them.
             identityCertCreation = false;
             validateShareConsumer(consumer, principal, keys);
+            // if there exists a share consumer between the two orgs, return it.
+            Consumer existingConsumer = consumerCurator.searchOwnerConsumers(owner,
+                null, null, null, null, null, null, null, null,
+                consumer.getRecipientOwnerKey()).uniqueResult();
```
mstead: Let's create a ConsumerCurator method specifically for finding a share consumer. All the null parameters screams break out into a method. I'm sure we are doing the same sort of thing in other places, but let's fix this one.

wottop:
I'm ack after you fix mstead's suggestion.

wottop:
ack

PR #1882, opened 2018-01-16
[DTO] Added the lastRefreshed field to the OwnerDTO

Ceiu:


Ceiu:
This branch depends on #1865 to fix some Jenkins failures, but #1865 depends on this PR to fix its failures. One needs to be merged before the other.

PR #1883, opened 2018-01-16
1531628: Reduced scope of consumer lock to exclude type in some queries

cnsnyder:
- Set the fetch mode of the consumer type to be SELECT (lazy) in the
  getHypervisorsBulk query, as the query also requests a pessimistic
  lock on the consumer. The lazy fetch mode on type excludes its
  table/rows from being included in the lock, preventing the locking
  logic from extending outside of the consumer/owner boundry.

- Performs all consumer updates (and creates etc) in one flush at the
  end of the HypervisorUpdateJob.

- Removes guestid created and removed events from checkForGuestsUpdate

- HypervisorUpdateJob handles all relevant consumer creation aspects


cnsnyder:
[1531627 - Two or more hypervisor update jobs can cause lockwait timeouts](https://trello.com/c/cQPkN7ss/711-1531627-two-or-more-hypervisor-update-jobs-can-cause-lockwait-timeouts)

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1521,9 +1521,11 @@ public UeberCertificate getUeberCertificate(@Context Principal principal,
         @PathParam("owner_key") @Verify(Owner.class) String ownerKey,
         @QueryParam("hypervisor_id") List<String> hypervisorIds) {
 
-        return (hypervisorIds == null || hypervisorIds.isEmpty()) ?
+        CandlepinQuery<Consumer> out = (hypervisorIds == null || hypervisorIds.isEmpty()) ?
             this.consumerCurator.getHypervisorsForOwner(ownerKey) :
             this.consumerCurator.getHypervisorsBulk(hypervisorIds, ownerKey);
```
barnabycourt: this is still using ownerKey even though the getHypervisorsBulk was changed to use the owner ID

server/client/ruby/lock_wait_repro.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
Ceiu: Probably don't need to include this repro script in master

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -183,6 +183,20 @@ public E create(E entity) {
...
+     * @return newly created entity.
+     */
+    @Transactional
+    public E create(E entity, boolean flush) {
+        getEntityManager().persist(entity);
+        if (flush) {
+            flush();
+        }
+        return entity;
+    }
```
Ceiu: Let's take this opportunity to clean up and sync our various methods here. Move the save method up next to these (or move these down to it), update the ```create(E)``` and ```save(E)``` (since create and save are synonyms) methods to call into this new create method with flush set to true. This way all of them will be doing the same thing, instead of having multiple implementations that all vary slightly.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -77,22 +80,36 @@
...
+
         this.ownerCurator = ownerCurator;
         this.consumerCurator = consumerCurator;
+        this.consumerTypeCurator = consumerTypeCurator;
         this.consumerResource = consumerResource;
         this.i18n = i18n;
+        this.subAdapter = subAdapter;
+        this.complianceRules = complianceRules;
+
+        this.hypervisorType = consumerTypeCurator.lookupByLabel(ConsumerTypeEnum.HYPERVISOR.getLabel());
```
Ceiu: We need to be a bit careful here in actual production code: we're broken if this returns null.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -251,31 +268,58 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
...
                             hypervisorId, ownerKey, knownHost.getHypervisorId().getReporterId(),
                             jobReporterId);
                     }
-                    if (consumerResource.performConsumerUpdates(incoming, knownHost, guestConsumersMap,
-                        false)) {
-                        consumerCurator.update(knownHost);
-                        result.updated(updateCheckinTime(knownHost));
+
+                    boolean changesMade = consumerResource.checkForFactsUpdate(knownHost, incoming);
+                    changesMade = consumerResource.checkForGuestsUpdate(knownHost, incoming) || changesMade;
```
Ceiu: We should add a TODO here to denote that we need to refactor this a bit so we're not calling into a resource for common work.

Also, slight aside, we're starting to diverge pretty heavily from what master has/does in this regard. That'll likely be a problem going forward.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -251,31 +268,58 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
...
+                    result.created(consumer);
+                }
+                else if (updatedConsumers.contains(consumer)) {
+                    consumerCurator.update(consumer, false);
+                    result.updated(consumer);
+                } else {
+                    result.unchanged(consumer);
+                }
+
+            }
```
Ceiu: Most of this block was just PoC code that wasn't intended to be used verbatim. There's a lot of cleanup to do here. Looking a bit closer at the pieces we have, what I would advise is the following:

- Get rid of the two new sets on line 235 and 236
- Move the result.(changed|updated|unchanged) lines back to their respective position in the body of the job.
- Update this block to use the collections maintained by the result to identify which DB action we're taking. To make things a bit simpler, I'd also add utility methods to HypervisorUpdateResult like .is/wasUpdated/Created/Unchanged so you don't need to constantly fetch the collection before using it, but you could also just fetch once, store locally and use.
- Since we're setting the last checkin time, we need to update even unchanged consumers here, so the loop really boils down to: 
```
consumer.setLastCheckin(now);
result.wasCreated(consumer) ? consumerCurator.create(consumer, false) : consumerCurator.update(consumer, false);
```

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -560,6 +559,7 @@ public Consumer create(
             complianceRules.getStatus(consumer, null, false, false);
             consumerCurator.update(consumer);
 
+            sink.emitConsumerCreated(consumer);
```
Ceiu: :+1:

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1045,6 +1045,11 @@ public boolean performConsumerUpdates(Consumer updated, Consumer toUpdate,
         return performConsumerUpdates(updated, toUpdate, guestConsumerMap, true);
     }
 
+    public boolean performConsumerUpdates(Consumer updated, Consumer toUpdate, boolean isIdCert) {
+
```
Ceiu: Extraneous whitespace

server/src/main/java/org/candlepin/resource/dto/HypervisorUpdateResult.java
```diff
@@ -46,6 +46,12 @@ public HypervisorUpdateResult() {
         this.failed = new HashSet<String>();
     }
 
+    public boolean wasCreated(Consumer c) { return created.contains(c); }
+
+    public boolean wasUpdated(Consumer c) { return updated.contains(c); }
+
+    public boolean wasUnchanged(Consumer c) { return unchanged.contains(c); }
```
Ceiu: Does this make it through the linter? We don't have any one-liners like this anywhere else; should use normal formatting here.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -393,14 +371,15 @@ public static String decompress(byte[] bytes) {
...
+    private Consumer createConsumerForHypervisorId(String incHypervisorId, Owner owner, Principal principal,
+                                                   Consumer incoming) {
         Consumer consumer = new Consumer();
         consumer.setName(sanitizeHypervisorId(incHypervisorId));
         consumer.setType(this.hypervisorType);
         consumer.setFact("uname.machine", "x86_64");
         consumer.setGuestIds(new ArrayList<GuestId>());
         consumer.setOwner(owner);
-        consumer.setAutoheal(true);
+        consumer.setAutoheal(!owner.getAutobindDisabled());
```
barnabycourt: I don't think this is what this setting is for. Isn't the owner.AutobindDisabled used for maintenance mode?

cnsnyder: My intention was to set the autobind to the default for the org. I can stick with the default of "true" for now if so desired.

cnsnyder: Changed back to default of "true".

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
Ceiu: Still need a contingency plan here.

cnsnyder:
Not sure what's going on but the failed mysql test "Hypervisor Resource should revoke guest entitlements when migration happens" passes on my locally deployed version of this branch.

server/src/main/java/org/candlepin/model/VirtConsumerMap.java
```diff
@@ -74,6 +74,10 @@ public Consumer get(String virtUuid) {
         return null;
     }
 
+    public java.util.Collection<Consumer> getConsumers() {
```
Ceiu: A bit more PoC code that's left over. We should do a proper import and not fully qualify this on the fly.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -77,22 +80,39 @@
...
         this.ownerCurator = ownerCurator;
         this.consumerCurator = consumerCurator;
         this.consumerResource = consumerResource;
         this.i18n = i18n;
+        this.subAdapter = subAdapter;
+        this.complianceRules = complianceRules;
+
+        this.hypervisorType = consumerTypeCurator.lookupByLabel(ConsumerTypeEnum.HYPERVISOR.getLabel());
+        if (this.hypervisorType == null) {
+            consumerTypeCurator.create(new ConsumerType(ConsumerTypeEnum.HYPERVISOR));
```
Ceiu: This is good, but hypervisorType is still null at this point. I believe create kicks out the instance, so we can just set it inline here.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -251,31 +267,50 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
...
                     reportedOnConsumer.getHypervisorId().setReporterId(jobReporterId);
                 }
                 else if (jobReporterId == null) {
                     log.debug("hypervisor checkin reported asynchronously without reporter id " +
                         "for hypervisor:{} of owner:{}", hypervisorId, ownerKey);
                 }
             }
+
+            Date now = new Date();
+            java.util.Collection<Consumer> c = hypervisorConsumersMap.getConsumers();
```
Ceiu: Another place where we shouldn't use the fully-qualified class here -- should do a proper import.

Ceiu:
ACK

PR #1885, opened 2018-01-18
[WIP] Product share resolves

vritant:


server/src/main/java/org/candlepin/controller/CandlepinOwnerProductShareManager.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Manages Shared product operations.
+ */
+public class CandlepinOwnerProductShareManager implements OwnerProductShareManager {
```
awood: Is there (or do we anticipate there will be) more than one implementation of the OwnerProductShareManager interface?  I'm guessing that there's just going to be the one implementation so unless we need an interface for testing purposes or something, I don't see the advantage in having an interface.

server/src/main/java/org/candlepin/controller/CandlepinOwnerProductShareManager.java
```diff
@@ -0,0 +1,200 @@
...
+        List<Product> existingProducts;
+        Set<String> missingProducts = new HashSet<String>();
+
+        if (productIds != null && !productIds.isEmpty()) {
+            // add all requested now so we can remove the found ones later.
+            missingProducts.addAll(productIds);
+            existingProducts = ownerProductCurator.getProductsByIds(recipientOwner, productIds).list();
+        }
+        else {
+            existingProducts = ownerProductCurator.getProductsByOwner(recipientOwner).list();
```
awood: Is this meant to handle the case where a caller wants to resolve *all* of an owner products?  The semantics here seem odd.  Sending in an empty or null list of productIds seems like it should be an error (or at the very least a no-op) instead of an expected/accepted way of doing things.

server/src/main/java/org/candlepin/controller/CandlepinOwnerProductShareManager.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        for (Product product : existingProducts) {
+            result.put(product.getId(), product);
+            missingProducts.remove(product.getId());
```
awood: If you wanted to be fancy you could probably do this all in one go with something like

```
missingProducts = missingProducts.removeAll(result.getKeys());
```

That's pseudo-code, but I imagine there's some method to remove a bunch of items at once from a set and I think `getKeys` there would return a set.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -969,13 +971,34 @@ public Pool convertToMasterPool(Subscription sub) {
...
 
+    /**
+     * Refreshes all pools effected in an owner given the products that changed.
+     * all products that effect non-master pools should effect master pools as well,
+     * so refreshing master pools should fix child pools.
+     * @param owner
+     * @param productMap
+     */
+    @Override
+    public void refreshEffectedPools(Owner owner, Map<String, Product> productMap) {
```
awood: Should be `refreshAffectedPools`

vritant:

server/src/main/java/org/candlepin/controller/OwnerProductShareManager.java
```diff
@@ -0,0 +1,247 @@
...
+     * @return recipient's ownerProduct if it exists, or resolve the most recently shared product from shared
+     * products if shares are requested to be included.
+     */
+    public List<Product> resolveProductsByIds(Owner recipientOwner, Collection<String> productIds,
+        boolean includeShares) {
+        Map<String, Product> result = new HashMap<String, Product>();
+        List<Product> existingProducts;
+        Set<String> missingProducts = new HashSet<String>();
+
+        if (productIds != null && !productIds.isEmpty()) {
```
awood: I'm still confused why `productIds` would ever be null or empty.  If that's an expected case, I think we should have a different public method named something like `resolveAllProducts(Owner owner, boolean includeShares)` and then have it and `resolveProductsByIds` delegate to the same private method.

server/src/main/java/org/candlepin/controller/OwnerProductShareManager.java
```diff
@@ -0,0 +1,247 @@
...
+     *   as 'refreshDue' so we can update all recipient owner's pools with this product and regenerate
+     *   respective ents. else we are done.
+     *
+     * @param sharingOwner the owner that is sharing the sharedProducts
+     * @param recipient the recipient owner of the share
+     * @param sharedProducts the products that are being shared from sharingOwner to recipient
+     * @return set of ResolvedProducts that also indicate if a refresh is required in the recipient for
+     * that owner
+     */
+    public Map<String, ResolvedProduct> resolveProductsAndUpdateProductShares(Owner sharingOwner,
```
awood: I would really like to see a more object-oriented approach here.  I have a general idea of what this code is meant to be doing and I'm having a hard time keeping this all in my head at once.

For example, resolving products and updating product shares seem, to me, to orthogonal to each other.  It is convenient to perform the actions at once, but logically speaking they're unrelated.  Could we break them apart?

server/src/main/java/org/candlepin/model/OwnerProductShareCurator.java
```diff
@@ -0,0 +1,97 @@
...
+        Collection<String> productIds) {
+        return findProductSharesByOwner(owner, true, activeOnly, productIds);
+    }
+
+    public List<OwnerProductShare> findProductSharesBySharer(Owner owner, boolean activeOnly,
+        Collection<String> productIds) {
+        return findProductSharesByOwner(owner, false, activeOnly, productIds);
+    }
+
+    private List<OwnerProductShare> findProductSharesByOwner(Owner owner, boolean isRecipient,
```
awood: I don't think it's a great idea to have the meaning of the method call hinge on a boolean like this.  Why not have a `findProductSharesGivenToOwner` and a `findProductSharesMadeByOwner` to encapsulate the difference between

```java
if (isRecipient) {
    jpql += "ps.recipientOwner.id = :owner_id ";
}
else {
    jpql += "ps.sharingOwner.id = :owner_id ";
}
```

Otherwise, it becomes a constant guessing game when a developer is reading the code as to what exactly they're getting back.

server/src/main/java/org/candlepin/model/OwnerProductShareCurator.java
```diff
@@ -0,0 +1,97 @@
...
+    public boolean removeShares(List<OwnerProductShare> shares) {
+        int rows = 0;
+        if (CollectionUtils.isNotEmpty(shares)) {
+            List<String> ids = new LinkedList<String>();
+            for (OwnerProductShare share : shares) {
+                ids.add(share.getId());
+            }
+
+            String jpql = "DELETE FROM OwnerProductShare s " +
+                "WHERE s.id in (:ids)";
```
awood: Do we need to worry about the in clause limit here?

server/src/test/java/org/candlepin/model/OwnerProductShareCuratorTest.java
```diff
@@ -0,0 +1,171 @@
...
+            assertEquals(share.getRecipientOwner(), recipientOwner);
+            if (share.getProduct() == product) {
+                productFound = true;
+            }
+            if (share.getProduct() == product2) {
+                product2Found = true;
+            }
+        }
+        assertTrue(productFound);
+        if (bothExpected) {
```
awood: Maybe

```java
assertEquals(bothExpected, product2Found)
```

server/src/main/java/org/candlepin/controller/OwnerProductShareManager.java
vritant: note from discussion: move productid logic to resource

server/src/main/java/org/candlepin/model/OwnerProductShareCurator.java
vritant: note from discussion, this is private method, so its okay.

server/src/main/java/org/candlepin/controller/OwnerProductShareManager.java
vritant: done

server/src/main/java/org/candlepin/model/OwnerProductShareCurator.java
vritant: done

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -395,12 +394,36 @@ private Owner resolveOwner(Owner owner) {
...
-        Map<String, Product> changedProducts) {
+        Map<String, Product> changedProducts, boolean refreshShares) {
+
+        List<Pool> allSubscriptionPools = poolCurator.getPoolsBySubscriptionId(pool.getSubscriptionId())
+            .list();
+        List<Pool> subscriptionPools = new LinkedList<>();
+
+        // choose the subset of pools depending on what we are refreshing
+        for (Pool poolToAdd : allSubscriptionPools) {
+            if ((refreshShares && poolToAdd.hasSharedAncestor()) ||
```
awood: This predicate might be clearer as

```java
if (refreshShares == poolToAdd.hasSharedAncestor) {
```

barnabycourt:
Adding back the in-progress given discussions about direction today.

PR #1886, opened 2018-01-18
Remove pin to qpid_proton 0.17.x gem

awood:
This change wasn't reflected in the Gemfile.lock and was causing the file to be updated continually.  Remove it for now until we figure out something better.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem 'stringex'
 gem 'digest-murmurhash'
 gem 'httpclient'
 gem 'activesupport', '~> 4.2'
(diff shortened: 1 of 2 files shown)


awood:

barnabycourt:
@awood This worked for me locally with vagrant but seems to have caused the jenkins environment a lot of problems.

awood:
> Odd to remove the pin from the gem file but add a pin to the gemfile.lock.

The Gemfile.lock is always going to have a version assigned to it.  I've got another idea on how to get rid of this dumb thing so I'm going to try that.

PR #1889, opened 2018-01-24
Do not use the rubygems provide proton bindings on Vagrant.

awood:
Also includes a script that uses the python proton bindings so that developers can easily spy on the queue and watch events flow through.  Theoretically, this script might be enough to spec test with, but I have difficulty with it starting too slow to catch events in the spec test.  We'd have to start it, have it not block, and then watch for the event coming through.

I thought it would just be easier to have the vagrant image rely on the qpid_proton provided via Yum while developers can do a `bundle install`

This PR also returns the Qpid version to 0.17.0 which is what Jenkins wants.  Developers should just use `bundle install --without=proton` to avoid the issue altogether.  Theoretically, it would be possible to segregate the two tests that actually require the qpid_proton library and only skip them instead of the entire qpid suite which tests for other things that don't require the library.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ buildconf/scripts/import_products.json
 .rvmrc
 .ruby-version
 .ruby-gemset
+.bundle
 
 # Vagrant
 .vagrant/
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,10 @@ gem 'stringex'
 gem 'digest-murmurhash'
 gem 'httpclient'
 gem 'activesupport', '~> 4.2'
-gem 'qpid_proton', '~> 0.17.0'
+
+group 'proton' do
+  gem 'qpid_proton'
+end
 
 # Remove this once we are fully using the new Ruby bindings
 gem 'rest-client', '~> 1.6.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -122,4 +122,4 @@ DEPENDENCIES
   webrick
 
 BUNDLED WITH
-   1.14.6
+   1.16.1
--- a/server/bin/qpid_watch.py
+++ b/server/bin/qpid_watch.py
@@ -1,0 +1,139 @@
+#! /usr/bin/python
+from __future__ import print_function, unicode_literals
+
+import sys
+import argparse
+import logging
+import proton
+import proton.handlers
+import proton.reactor
+import proton.utils
+
+logging.basicConfig()
+log = logging.getLogger("qpid_watch")
+
+
+class TimeoutError(Exception):
+    pass
+
+
+class QpidWatcher(proton.handlers.MessagingHandler):
+    def __init__(self, args):
+        super(QpidWatcher, self).__init__()
+
+        if args.ssl_cert and args.ssl_key:
+            self.ssl_domain = proton.SSLDomain(proton.SSLDomain.MODE_CLIENT)
+            self.ssl_domain.set_credentials(args.ssl_cert, args.ssl_key, None)
+            # Bad practice, but this is meant for dev testing only.
+            self.ssl_domain.set_peer_authentication(proton.SSLDomain.ANONYMOUS_PEER)
+        else:
+            self.ssl_domain = None
+
+        self.received = 0
+        self.expected = args.messages
+        self.url = args.address
+        self.source = args.source
+        self.timeout = args.timeout
+        self.subject = args.subject
+        self.show_message = args.show_message
+
+    def on_start(self, event):
+        conn = event.container.connect(url=self.url, ssl_domain=self.ssl_domain)
+
+        if self.timeout > 0:
+            self.last_event = event.container.mark()
+            event.container.schedule(self.timeout, self)
+
+        # Note that Copy() does not actually consume the message off the queue!
+        self.receiver = event.container.create_receiver(conn, self.source, options=proton.reactor.Copy())
+
+    def on_timer_task(self, event):
+        elapsed = event.container.mark() - self.last_event
+        # elapsed is in milliseconds
+        if elapsed > self.timeout * 1000:
+            raise TimeoutError("No activity for %d seconds" % self.timeout)
+        else:
+            # Reschedule for the next timeout period
+            event.container.schedule(self.timeout, self)
+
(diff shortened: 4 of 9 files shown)


Lorquas:
So this works for me with a few caveats.

1) in the vagrant case, the vagrant image's networking is all messed up.  I'm not sure if this is because of a bug in vagrant-hostmanager or how we're using it or my box.  It cannot hosts for the yum repos on the first try or rubygems.org and the provision has to be run multiple times.

I've worked around this by commenting out the hostmanger block in the `Vagrantfile` and adding this hacky work around:

```diff
diff --git a/vagrant/candlepin.yml b/vagrant/candlepin.yml
index 4864e52f4..0baeb4c6c 100644
--- a/vagrant/candlepin.yml
+++ b/vagrant/candlepin.yml
@@ -5,6 +5,15 @@
   environment:
     JAVA_HOME: /usr/lib/jvm/java-1.8.0/
 
+  pre_tasks:
+    - name: fix resolv.conf
+      lineinfile:
+        path: /etc/resolv.conf
+        state: absent
+        line: "search example.com"
+      become: yes
+      become_user: root
+
   roles:
     - role: candlepin-root
       become: yes
```

I think this requires more investigation but it has nothing to do with the topic of this PR.  Once these above changes are made the vagrant seems to work on the first pass and candlepin is up.  The curl command against `/candlepin/status` returns the expected result.

2) This blocking QE deployments of candlepin.  Switching to this branch in our automation fixes our issues.  This a more generic question about bundler but in the vagrant's ansible this fix runs: ` command: /usr/bin/bundle install --without=proton` in the `vagrant/roles/candlepin-user/tasks/main.yml` file.  Our automation uses a newer version of ansible, and we have this section like: 

```yaml
- name: Run Bundle install
  bundler:
    executable: /usr/bin/bundle
    state: latest
    user_install: true
    chdir: "{{candlepin_checkout}}"
    # extra_args: "--without=proton"
  tags:
    - candlepin-user
```
adding those extra args seem to break it since this module is running a bundle update first and not a raw bundle install; everything seems to work though.  Should this be ok?

awood:
> adding those extra args seem to break it since this module is running a bundle update first and not a raw bundle install; everything seems to work though. Should this be ok?

QE deployments are using a version of Ansible with a special "bundler" task?  What if you use `state: present` instead of `state: latest`?  The Ansible documents seem to imply that "present" would translate to "bundle install" instead of "bundle update".

Lorquas:
Ah yup, I just ran a new deploy and changing it to present works with the extra_args.  Good catch.

Ceiu:
ACK pending final conformation from @Lorquas 

```
(14:15:54) crog@cl-vm-1:~/devel/candlepin-origin/server(awood/qpid_proton_pin) $ gem uninstall qpid_proton      
Successfully uninstalled qpid_proton-0.17.0
(14:16:11) crog@cl-vm-1:~/devel/candlepin-origin/server(awood/qpid_proton_pin) $ bundle install --without proton
Using rake 0.9.2.2
Using i18n 0.8.4
Using minitest 5.10.2
Using thread_safe 0.3.6
Using tzinfo 1.2.3
Using activesupport 4.2.8
Using ast 2.3.0
Using rjb 1.5.4
Using atoulme-Antwrap 0.7.5
Using debug_inspector 0.0.3
Using binding_of_caller 0.7.2
Using builder 3.2.2
Using bundler 1.16.1
Using diff-lcs 1.2.5
Using json_pure 1.8.3
Using minitar 0.5.4
Using net-ssh 3.1.1
Using net-sftp 2.1.2
Using rubyzip 1.2.0
Using xml-simple 1.1.5
Using buildr 1.5.3
Using buildr-findBugs 0.1.1
Using byebug 9.0.6
Using coderay 1.1.1
Using digest-murmurhash 1.1.1
Using httpclient 2.8.3
Using method_source 0.8.2
Using mime-types 1.25.1
Using oauth 0.5.3
Using parallel 1.11.2
Using parallel_tests 2.14.1
Using parser 2.4.0.0
Using pmd 1.0.1
Using powerpack 0.1.1
Using slop 3.6.0
Using pry 0.10.4
Using pry-byebug 3.4.2
Using pry-stack_explorer 0.4.9.2
Using rainbow 2.2.2
Using rest-client 1.6.9
Using rspec-support 3.6.0
Using rspec-core 3.6.0
Using rspec-expectations 3.6.0
Using rspec-mocks 3.6.0
Using rspec 3.6.0
Using ruby-progressbar 1.8.1
Using rubocop 0.36.0
Using stringex 2.7.1
Using webrick 1.3.1
Bundle complete! 19 Gemfile dependencies, 49 gems now installed.
Gems in the group proton were not installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.

...

(14:19:15) crog@cl-vm-1:~/devel/candlepin-origin/server(awood/qpid_proton_pin) $ buildr clean rspec:qpid
(in /home/crog/devel/candlepin-origin, development)
Cleaning candlepin:server
/home/crog/.rvm/rubies/ruby-2.3.3/bin/ruby -I/home/crog/.rvm/gems/ruby-2.3.3/gems/rspec-core-3.6.0/lib:/home/crog/.rvm/gems/ruby-2.3.3/gems/rspec-support-3.6.0/lib /home/crog/.rvm/gems/ruby-2.3.3/gems/rspec-core-3.6.0/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb --color --format documentation --require /home/crog/devel/candlepin-origin/tasks/failure_formatter --format ModifiedRSpec::FailuresFormatter -I /home/crog/devel/candlepin-origin/server/spec -I /home/crog/devel/candlepin-origin/server/client/ruby /home/crog/devel/candlepin-origin/server/spec/qpid_spec.rb

Qpid Broker
  qpid_proton is not available.  Skipping these tests (PENDING: No reason given)

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) Qpid Broker qpid_proton is not available.  Skipping these tests
     # No reason given
     # ./server/spec/qpid_spec.rb:244


Finished in 0.00105 seconds (files took 0.32956 seconds to load)
1 example, 0 failures, 1 pending

Completed in 1.257s
```

Ceiu:
ack

PR #1890, opened 2018-01-25
[DTO] Implemented CdnDTO/CdnTranslator

nikosmoum:
- Implemented CdnDTO/CdnTranslator.
- Translated endpoints that use Cdn to the DTO framework.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/CdnResource.java
+++ b/server/src/main/java/org/candlepin/resource/CdnResource.java
@@ -29,7 +29,6 @@ import org.candlepin.model.CdnCurator;
 
 import com.google.inject.Inject;
 
-import org.apache.commons.lang.StringUtils;
 import org.candlepin.model.CertificateSerial;
 import org.xnap.commons.i18n.I18n;
 
@@ -108,6 +107,10 @@ public class CdnResource {
         Cdn cndToCreate = new Cdn();
         this.populateEntity(cndToCreate, cdnDTOInput);
 
+        if (cdnDTOInput.getLabel() != null) {
+            cndToCreate.setLabel(cdnDTOInput.getLabel());
+        }
+
         return this.translator.translate(cdnManager.createCdn(cndToCreate), CdnDTO.class);
     }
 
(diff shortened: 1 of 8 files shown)


nikosmoum:
[Add API DTO for Cdn.](https://trello.com/c/YTdIJdbN/739-add-api-dto-for-cdn)

server/src/main/java/org/candlepin/resource/CdnResource.java
```diff
@@ -88,36 +96,110 @@ public void delete(@PathParam("label") String label,
...
+
+        if (dto == null) {
+            throw new IllegalArgumentException("the Cdn dto is null");
+        }
+
+        if (dto.getName() != null) {
+            entity.setName(dto.getName());
+        }
+
+        if (dto.getLabel() != null) {
```
Ceiu: We don't allow setting the label. This appears to be our external ID for this entity.

nikosmoum: Then who creates the label? CdnResource.create() explicitly says that it expects a unique label, otherwise will not create the Cdn:
```
    public CdnDTO create(
        @ApiParam(name = "cdn", required = true) CdnDTO cdnDTOInput,
        @Context Principal principal) {
        Cdn existing = curator.lookupByLabel(cdnDTOInput.getLabel());
        if (existing != null) {
            throw new BadRequestException(i18n.tr(
                "A CDN with the label {0} already exists", cdnDTOInput.getLabel()));
        }
```

Ceiu: Ahh, I misread the diff here and had it in my head we were using it in both update and create. Let me address update specifically, then.

server/src/main/java/org/candlepin/resource/CdnResource.java
```diff
@@ -88,36 +96,110 @@ public void delete(@PathParam("label") String label,
...
 
     @ApiOperation(notes = "Updates a CDN @return a Cdn object", value = "update")
     @PUT
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/{label}")
-    public Cdn update(@PathParam("label") String label,
-        @ApiParam(name = "cdn", required = true) Cdn cdn,
+    public CdnDTO update(@PathParam("label") String label,
+        @ApiParam(name = "cdn", required = true) Cdn cdnDTOInput,
```
Ceiu: Missed this one the first time around: we're still accepting an entity on input -- this should accept a CdnDTO, which will likely mean the rest of this method changes as well.

Also, given that we have a CdnManager already, this is probably a good time to update this resource to only do validation and offload any actual work to the manager.

server/src/main/java/org/candlepin/resource/CdnResource.java
```diff
@@ -88,36 +96,110 @@ public void delete(@PathParam("label") String label,
...
+        Cdn existing = curator.lookupByLabel(cdnDTOInput.getLabel());
         if (existing != null) {
-            throw new BadRequestException(i18n.tr("A CDN with the label {0} already exists", cdn.getLabel()));
+            throw new BadRequestException(i18n.tr(
+                "A CDN with the label {0} already exists", cdnDTOInput.getLabel()));
         }
-        return cdnManager.createCdn(cdn);
+
+        Cdn cndToCreate = new Cdn();
+        this.populateEntity(cndToCreate, cdnDTOInput);
```
Ceiu: We're no longer persisting the entity created here. Need to call into CdnCurator or Manager to create the instance. The latter is preferred here.

nikosmoum: cdnManager.createCdn() is called in the next line, inline with the translate() call.

server/src/main/java/org/candlepin/resource/CdnResource.java
nikosmoum: I'll remove setting the label in populateEntity, so that I can use it on both update and create methods.

server/src/main/java/org/candlepin/resource/CdnResource.java
Ceiu: ack

Ceiu:
ack

PR #1894, opened 2018-02-05
[DTO] Updated EntitlementDTO to use ConsumerDTOs instead of Consumer entities

Ceiu:


Ceiu:
[[2] [DTO]Change EntitlementDTO's consumer field type from 'Consumer' to 'ConsumerDTO', once ConsumerDTO is merged in.](https://trello.com/c/xrSszIRT/750-2-dtochange-entitlementdtos-consumer-field-type-from-consumer-to-consumerdto-once-consumerdto-is-merged-in)

server/src/test/java/org/candlepin/dto/api/v1/EntitlementDTOTest.java
```diff
@@ -57,11 +57,11 @@ public EntitlementDTOTest() {
         pool.setProductName("pool_product_name");
 
         //TODO: change to ConsumerDTO once it gets introduced.
```
nikosmoum: This TODO can be removed now.

server/src/test/java/org/candlepin/dto/api/v1/EntitlementDTOTest.java
```diff
@@ -57,11 +57,11 @@ public EntitlementDTOTest() {
...
-        Consumer consumer = new Consumer();
+        ConsumerDTO consumer = new ConsumerDTO();
         consumer.setId("consumer_id");
         consumer.setName("consumer_name");
         consumer.setUsername("consumer_username");
```
nikosmoum: We need to set the uuid too here (e.g. `consumer.setUuid("consumer_uuid");`), since EntitlementDTO's toString/populate/equals/hashcode methods use the consumerDTO's uuid, and the tests for those methods for the consumer field are getting ignored.

Ceiu:
@nikosmoum done

PR #1895, opened 2018-02-06
[M] Several corrections to fix issues found during static analysis

Ceiu:


server/src/main/java/org/candlepin/pki/impl/BouncyCastlePKIReader.java
```diff
@@ -190,9 +182,6 @@ else if (caKeyObj instanceof PrivateKey) {
                         caKeyObj.getClass().getName());
                 }
             }
-            finally {
-                reader.close();
```
wottop: Don't we need to close the reader somewhere?

Ceiu: This switched to the "new" try-with-resources block, which automatically handles any closeables within the try declaration on line 165.

server/src/main/java/org/candlepin/util/CrlFileUtil.java
```diff
@@ -133,18 +133,20 @@ public void initializeCRLFile(File file, Collection<BigInteger> revoke) throws I
...
-        BufferedWriter w = new BufferedWriter(new FileWriter(tempFile));
 
+        BufferedReader r = null;
+        BufferedWriter w = null;
         String line = null;
 
         try {
+            r = new BufferedReader(new FileReader(file));
+            w = new BufferedWriter(new FileWriter(tempFile));
+
```
nikosmoum: Could use a multi-try-with-resources
```
try (r = new BufferedReader(new FileReader(file));
     w = new BufferedWriter(new FileWriter(tempFile))) {
```
to cut down the boilerplate.

server/src/main/webapp/docs/lib/swagger-oauth.js
```diff
@@ -107,44 +110,49 @@ function handleLogin() {
     var redirectUrl = window.oAuthRedirectUrl || defaultRedirectUrl;
     var url = null;
     var scopes = []
-    var o = popup.find('input:checked'); 
+    var o = popup.find('input:checked');
     var OAuthSchemeKeys = [];
     var state;
-    for(k =0; k < o.length; k++) {
+
+    for (k =0; k < o.length; k++) {
```
nikosmoum: Missing whitespace between = and 0

server/src/main/java/org/candlepin/jackson/DateSerializer.java
```diff
@@ -31,20 +31,18 @@
...
-    // This SimpleDateFormat is iso 8601 without milliseconds
-    public static final SimpleDateFormat ISO_8601_WITHOUT_MILLISECONDS;
-
-    static {
-        ISO_8601_WITHOUT_MILLISECONDS = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
-        ISO_8601_WITHOUT_MILLISECONDS.setTimeZone(TimeZone.getTimeZone("UTC"));
+    public DateSerializer() {
+        // This SimpleDateFormat is ISO 8601 without milliseconds
+        this.dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
+        this.dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
```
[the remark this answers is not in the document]
Ceiu: Fortunately, we don't create this more than once per object mapper, so the per-instance declaration isn't substantially more expensive than the static one. The main reason we made this change, however, is DateFormat (and much of the old Java date bits) is not thread safe. Granted, this doesn't really matter for the same reason changing it to an instance variable doesn't change our performance here, but since the impact is so low, and I was already spending the dev time doing stuff like this, best practices wins out.

server/src/main/java/org/candlepin/util/CrlFileUtil.java
Ceiu: There was a reason I didn't at the time, but I can't think of that reason now. Guess it couldn't have been that important. Fixed.

nikosmoum:
@Ceiu Do you want anyone else to review or should I merge?

kahowell:
[Several corrections to fix issues found during static analysis](https://trello.com/c/jEZrPWuO/781-several-corrections-to-fix-issues-found-during-static-analysis)

Ceiu:
If this is ready to be merged, feel free to merge. If someone else would like to review, they are welcome to do so.

PR #1897, opened 2018-02-07
Awood/schema check test

awood:
If you want to test this test, try
* Deleting the `schemaLocation` attribute from the `databaseChangeLog` element.
* Changing the `schemaLocation` value to something that doesn't match the regex (e.g. `xsi:schemaLocation="foo"`.
* Changing the schema URL to something that is not "3.1".

Any of those changes will result in a test failure with a unique message.

This PR also converts all existing changesets to use version 3.1.  This should not be an issue since we've made version switches in the past with no ill effects.


awood:
[[5] Write a test to verify we don't use a liquibase XSD version over 3.1](https://trello.com/c/hI8mX4Yl/755-5-write-a-test-to-verify-we-dont-use-a-liquibase-xsd-version-over-31)

awood:

Ceiu:
ack

PR #1898, opened 2018-02-08
[M] Updated Jackson version to 2.9.4

Ceiu:
- Updated the version of Jackson to 2.9.4
- Updated a couple uses of XmlElement to use JsonIgnore instead
- Added additional testing output for a couple of JSON tests
- SourceSubscription is no longer processed from clients

### DIFF
--- a/server/src/main/java/org/candlepin/model/Pool.java
+++ b/server/src/main/java/org/candlepin/model/Pool.java
@@ -69,7 +69,6 @@ import javax.validation.constraints.Size;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlRootElement;
-import javax.xml.bind.annotation.XmlTransient;
 
 
 
@@ -243,7 +242,7 @@ public class Pool extends AbstractHibernateObject implements Persisted, Owned, N
     @OneToOne(mappedBy = "derivedPool", targetEntity = SourceStack.class)
     @Cascade({org.hibernate.annotations.CascadeType.ALL,
         org.hibernate.annotations.CascadeType.DELETE_ORPHAN})
-    @XmlTransient
+    @JsonIgnore
     private SourceStack sourceStack;
 
     /**
@@ -253,7 +252,6 @@ public class Pool extends AbstractHibernateObject implements Persisted, Owned, N
     @OneToOne(mappedBy = "pool", targetEntity = SourceSubscription.class)
     @Cascade({org.hibernate.annotations.CascadeType.ALL,
         org.hibernate.annotations.CascadeType.DELETE_ORPHAN})
-    @XmlTransient
     private SourceSubscription sourceSubscription;
 
     @Column(nullable = false)
@@ -644,7 +642,7 @@ public class Pool extends AbstractHibernateObject implements Persisted, Owned, N
      * @return
      *  the value set for the given attribute, or null if the attribute is not set
      */
-    @XmlTransient
+    @JsonIgnore
     public String getAttributeValue(String key) {
         if (key == null) {
             throw new IllegalArgumentException("key is null");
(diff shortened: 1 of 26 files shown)


Ceiu:
[[13] Update Jackson Libraries](https://trello.com/c/LGDpfBeP/751-13-update-jackson-libraries)

server/src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java
```diff
@@ -959,6 +970,9 @@ private void adjustHostedBonusPoolQuantity(PoolManager poolManager, Consumer c,
     public void postEntitlement(PoolManager poolManager, Consumer consumer,
         Map<String, Entitlement> entitlements, List<Pool> subPoolsForStackIds, boolean isUpdate,
         Map<String, PoolQuantity> poolQuantityMap) {
+
+        // TODO: Why are we calling a separate method here instead of just doing that method's work here?
```
kahowell: Nitpick: only formatting changes and this comment in this file - should happen in a separate PR...

server/src/test/java/org/candlepin/policy/js/RulesObjectMapperTest.java
```diff
@@ -121,12 +120,13 @@ public void filterTimestampsOffAttributes() {
...
+
         // Shouldn't see timestamps:
-        assertFalse(output.contains("\"created\""));
-        assertFalse(output.contains("\"updated\""));
+        assertFalse(output, output.contains("\"created\""));
+        assertFalse(output, output.contains("\"updated\""));
 
         // Shouldn't see a productId:
-        assertFalse(output.contains("PRODID"));
+        assertFalse(output, output.contains("PRODID"));
```
kahowell: Nitpick: these changes are good, but are unrelated to the changes necessary for Jackson... Separate PR please?

server/src/test/java/org/candlepin/sync/ConsumerExporterTest.java
```diff
@@ -70,7 +70,8 @@ public void testConsumerExport() throws IOException {
         json.append("\"urlWeb\":\"/subscriptions\",");
         json.append("\"urlApi\":\"/candlepin\",");
         json.append("\"contentAccessMode\":\"access_mode\"}");
-        assertTrue(TestUtil.isJsonEqual(json.toString(), writer.toString()));
+        assertTrue(json.toString() + "\n" + writer.toString(),
```
kahowell: Nitpick: these changes are good, but are unrelated to the changes necessary for Jackson... Separate PR please?

server/src/main/java/org/candlepin/model/Pool.java
```diff
@@ -1397,22 +1395,25 @@ public void setBranding(Set<Branding> branding) {
         this.branding = branding;
     }
 
-    @XmlTransient
+    @JsonIgnore
     public SourceSubscription getSourceSubscription() {
         return sourceSubscription;
     }
 
+    @JsonIgnore
```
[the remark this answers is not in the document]
Ceiu: I'm not sure this mandates a comment in code, but the reason for annotating in this way are due to a number of factors:

- Going forward, this is how we should be annotating things. Jackson configuration (or behavior) can change how implied functionality works, so if we omit it, the behavior could change out from under us in the future, as has happened in a number of places in this PR.
- Throwing the annotation on the field works, but becomes inconsistent with cases where we want a getter or setter to be used, but not both. In those cases, the annotations go on the getter/setter rather than the field.
- In this particular case, the first iteration actually allowed the setter to be used (but not the getter), but in the name of sane JSON use, that was dropped and this was changed to ignore it if present.

Ceiu:
Failures have been addressed (missing underscore was the culprit), but I'm going to push back on the separate PR for formatting changes as we discussed online.

server/src/main/java/org/candlepin/model/Pool.java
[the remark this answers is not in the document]
kahowell: ack

PR #1899, opened 2018-02-09
[DTO] Implemented ConsumerDTO & OwnerDTO APIs for the Rules framework.

nikosmoum:
- Added RulesConsumerDTO which is a DTO of the Consumer entity as used by the Rules API.
- Now translating Consumer entity instances passed into the Rules API to RulesConsumerDTO.

### DIFF
--- a/server/src/main/java/org/candlepin/dto/rules/v1/ConsumerTranslator.java
+++ b/server/src/main/java/org/candlepin/dto/rules/v1/ConsumerTranslator.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2009 - 2017 Red Hat, Inc.
+ * Copyright (c) 2009 - 2018 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -12,23 +12,20 @@
  * granted to use or replicate Red Hat trademarks that are incorporated
  * in this software or its documentation.
  */
-package org.candlepin.dto.api.v1;
+package org.candlepin.dto.rules.v1;
 
 import org.candlepin.dto.ModelTranslator;
-import org.candlepin.dto.ObjectTranslator;
 import org.candlepin.dto.TimestampedEntityTranslator;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.ConsumerCapability;
 import org.candlepin.model.ConsumerInstalledProduct;
-import org.candlepin.model.Release;
 
-import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Set;
 
 /**
  * The ConsumerTranslator provides translation from Consumer model objects to
- * ConsumerDTOs
+ * ConsumerDTOs, as used by the Rules framework.
  */
 public class ConsumerTranslator extends TimestampedEntityTranslator<Consumer, ConsumerDTO> {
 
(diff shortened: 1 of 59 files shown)


nikosmoum:
[[13] Create ConsumerDTO API for Rules framework](https://trello.com/c/N8qe31LE/763-13-create-consumerdto-api-for-rules-framework)

nikosmoum:
Note for reviewers: The RulesConsumerDTO is really a subset/cut-down version of the original ConsumerDTO, which only contains the fields that rules.js actually needs.

server/src/main/java/org/candlepin/dto/api/v1/RulesConsumerDTO.java
```diff
@@ -0,0 +1,474 @@
...
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.dto.api.v1;
```
Ceiu: This should live in the rules namespace: ```org.candlepin.dto.rules.v1```

server/src/main/java/org/candlepin/dto/api/v1/RulesConsumerDTO.java
```diff
@@ -0,0 +1,474 @@
...
+
+/**
+ * A DTO representation of the Consumer entity, as used by the Rules framework.
+ */
+@ApiModel(parent = TimestampedCandlepinDTO.class,
+    description = "DTO representing an upstream consumer, as used by the Rules framework")
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.PROPERTY)
+@JsonFilter("ConsumerFilter")
+public class RulesConsumerDTO extends TimestampedCandlepinDTO<RulesConsumerDTO> implements LinkableDTO {
```
Ceiu: With the package change, we're free to call this ConsumerDTO

server/src/main/java/org/candlepin/dto/api/v1/RulesConsumerTranslator.java
```diff
@@ -0,0 +1,118 @@
...
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.dto.api.v1;
```
Ceiu: Same deal here: update package, simplify name

server/src/main/java/org/candlepin/dto/rules/v1/ConsumerTranslator.java
```diff
@@ -12,10 +12,15 @@
...
-package org.candlepin.dto.api.v1;
+package org.candlepin.dto.rules.v1;
 
 import org.candlepin.dto.ModelTranslator;
 import org.candlepin.dto.ObjectTranslator;
+import org.candlepin.dto.api.v1.CapabilityDTO;
+import org.candlepin.dto.api.v1.ConsumerInstalledProductDTO;
+import org.candlepin.dto.api.v1.ConsumerTypeDTO;
+import org.candlepin.dto.api.v1.OwnerDTO;
+import org.candlepin.dto.api.v1.TimestampedEntityTranslator;
```
Ceiu: Each DTO package should be isolated and not using the DTOs from the other packages directly. In the case of these objects, most of them will end up being minimal objects in the rules, so we don't need to have serialization of most fields like the API needs.

server/src/test/java/org/candlepin/dto/api/v1/CapabilityTranslatorTest.java
```diff
@@ -34,18 +34,18 @@
     protected CapabilityTranslator capabilityTranslator = new CapabilityTranslator();
 
     @Override
-    protected void initModelTranslator(ModelTranslator modelTranslator) {
+    public void initModelTranslator(ModelTranslator modelTranslator) {
```
Ceiu: What's driving the change to increase visibility here? These are intended to be used only by subclasses.

nikosmoum: It's because we're using the DTOs from the dto.api.v1 package, so it would be justified to be able to use the test utility methods from the same package. Since we're going to have separate DTOs isolated in the dto.rules.v1 package though as you mentioned in the other comment, I'll revert these changes, and either introduce new test classes in that package (rules), or do manual setup/validation if they're minimal enough.

server/src/main/java/org/candlepin/dto/rules/v1/ConsumerTranslator.java
nikosmoum: I removed dependencies to DTOs from api.v1, but not from common classes like api.v1.LinkableDTO and api.v1.TimestampedEntityTranslator, since those will probably get moved at some point anyway.

Ceiu: Probably not, honestly. Linkable is a REST-specific thing that only matters on the API layer, and I doubt the rules care about the timestamps.

The hard/tedious bit about the rules DTOs is that we should sift through the rules and figure out what fields we actually use. These will be tiny DTOs as a result, but that's actually a good thing -- it'll make translating them very quick.

server/src/test/java/org/candlepin/dto/api/v1/CapabilityTranslatorTest.java
Ceiu: ack

server/src/main/java/org/candlepin/dto/rules/v1/ConsumerTranslator.java
nikosmoum: Ah you're right about the LinkableDTO, I should just remove that. The TimestampedEntityTranslator/TimestampedCandlepinDTO dependencies I kept because I can see that rules.js actually uses consumer.created (but not updated), so I guess I can remove that too and instead put the created field directly on the class.

Ceiu: That's kind of surprising. If we have a legit use case, then there's an argument to be made that TimestampedCandlepinDTO should move up to the general dto layer. Since it's abstract already, that'll be a pretty trivial change.

nikosmoum: done

server/src/main/java/org/candlepin/dto/rules/v1/OwnerDTO.java
```diff
@@ -0,0 +1,116 @@
...
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlRootElement;
+
+/**
+ * A DTO representation of the Owner entity as used by the Rules framework.
+ */
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.PROPERTY)
+public class OwnerDTO extends CandlepinDTO<OwnerDTO> {
```
Ceiu: You sure we won't need id, name, and/or account here?

Ceiu:
ACKing the changes, but I'm a little worried about a DTO that lacks a clear identifier. I feel like we should, at minimum, include the model's DB ID field to ensure we can differentiate between two DTOs that have the same value for the field we're using.

server/src/main/java/org/candlepin/dto/rules/v1/OwnerDTO.java
nikosmoum: 'defaultServiceLevel' is the only owner field I see being accessed in rules.js, so I'm pretty sure, unless I'm missing something else.

nikosmoum:
@Ceiu I added the id field on OwnerDTO just to be sure.

Ceiu:
ack

PR #1901, opened 2018-02-12
Various DTO fixes in equals/hashcode/populate methods and ignored tests.

nikosmoum:
- Fixed equals/hashcode/populate methods in ConsumerDTO, EnvironmentDTO, UpstreamConsumerDTO.
- Fixed a lot of DTO unit tests that were silently ignored due to missing fields during setup. Also fixed the inheritDoc annotation.


nikosmoum:
@Ceiu The tests without proper field names or missing fields in the setup are being ignored instead of failing due to the use of junit's assumeFalse (instead of assertFalse) in AbstractDTOTest. Is this something we should consider to change? Why was it used in the first place?

nikosmoum:
[[5] Fix various ignored DTO Tests.](https://trello.com/c/2aZAv8zK/768-5-fix-various-ignored-dto-tests)

Ceiu:
ack

PR #1902, opened 2018-02-12
PoolQuantity DTO

wottop:


server/src/main/java/org/candlepin/dto/api/v1/PoolQuantityDTO.java
```diff
@@ -0,0 +1,187 @@
...
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.dto.api.v1;
```
Ceiu: This is (probably) a DTO for the rules stuff, so it should go in the rules package/namespace.

Ceiu:
The code itself looks good outside of the package bit. As we discussed online, it may not be something needed in the DTO framework itself. It might be custom enough that it's a one-off used as more internal processing by the rules before a result is returned.

nikosmoum:
@Ceiu @wottop PoolQuantity is used in one API endpoint in ConsumerResource.java (/{consumer_uuid}/entitlements/dry-run), and in the rules framework (EntitlementRules.java). That is why there is another card in the backlog for the Rules framework, while this was to be for the API endpoint.

Ceiu:
I still don't know that this needs to change at all. PoolQuantity isn't a model object, so there's really no reason to translate the existing DTO to another DTO. This is probably just a case where we leave it as-is.

kahowell:
[Add API DTO for PoolQuantity.](https://trello.com/c/u9FibawV/767-add-api-dto-for-poolquantity)

Ceiu:
I'm gonna take a look at the path from the data in the rules to the end point from which this is returned and figure out how we should address this DTO.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -534,6 +539,10 @@ protected void verifyDevProducts(Consumer consumer, String expectedSku, Develope
                 }
             }
         }
+        List<PoolQuantityDTO> result = new ArrayList<PoolQuantityDTO>();
+        for (PoolQuantity pq : poolQuantities) {
+            result.add(this.modelTranslator.translate(pq, PoolQuantityDTO.class));
+        }
```
Ceiu: The translation should be done at the resource layer.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1966,13 +1966,13 @@ public Response bind(
...
 
-        List<PoolQuantity> dryRunPools = new ArrayList<PoolQuantity>();
+        List<PoolQuantityDTO> dryRunPools = new ArrayList<PoolQuantityDTO>();
 
         try {
             consumerBindUtil.validateServiceLevel(consumer.getOwner(), serviceLevel);
             dryRunPools = entitler.getDryRun(consumer, serviceLevel);
             if (dryRunPools == null) {
-                return new ArrayList<PoolQuantity>();
+                return new ArrayList<PoolQuantityDTO>();
```
Ceiu: We can refactor this block a bit such that we're not creating the collection only to throw it out and, in some cases, recreate it.

Ceiu: What I was referring to here was the creation of the arraylist. We could do something like:

```
List<PoolQuantity> dryRunPools = null;

try {
    ...
    dryRunPools = entitler.getDryRun(...);
}
// more catch stuff here

if (dryRunPools != null) {
  // do translation here
}
else {
  return Collections.<PoolQuantityDTO>emptyList();
}
```

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1971,21 +1972,26 @@ public Response bind(
...
         catch (ForbiddenException fe) {
-            return dryRunPools;
+            return Collections.<PoolQuantityDTO>emptyList();
         }
         catch (BadRequestException bre) {
             throw bre;
         }
         catch (RuntimeException re) {
-            return dryRunPools;
+            return Collections.<PoolQuantityDTO>emptyList();
```
Ceiu: We can skip returning from the exception handlers here and just fall through to the new translation block.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1971,21 +1972,26 @@ public Response bind(
...
+        }
+        if (dryRunPools != null) {
+            List<PoolQuantityDTO> dryRunPoolDtos = new ArrayList<PoolQuantityDTO>();
+            for (PoolQuantity pq : dryRunPools) {
+                dryRunPoolDtos.add(this.translator.translate(pq, PoolQuantityDTO.class));
+            }
+            return dryRunPoolDtos;
```
Ceiu: This is what I'd do here as well, but eventually we should replace this block with either the streaming stuff, a translation iterator or whatever we come up with for bulk translation. For now maybe a TODO to improve this in the future.

PR #1903, opened 2018-02-12
Awood/1478549 update hornetq

awood:
Move from HornetQ to Apache ActiveMQ Artemis (which is the next incarnation of HornetQ) in order to address an open CVE.

While looking at the CVE list, I also updated Logback since that was there too.

There is one consideration to keep in mind.  I changed the directory that stores the journal, etc. from /var/lib/candlepin/hornetq to /var/lib/candlepin/activemq-artemis.  Using the old directory with ActiveMQ Artemis will result in a deployment failures since Artemis won't read the old journal.  Changing the directory name should, in my opinion, be sufficient but it will leave the old `hornetq` directory.  We have the option of adding a `%pre` section to the Candlepin spec file that will remove any existing `/var/lib/candlepin/hornetq` directory.

It is also possible, but very unlikely, that someone will not be using the default value (which I have changed) for `candlepin.audit.hornetq.base_dir`.  If they have this value set, they'll hit deployment errors since Candlepin will start up and try to read old journals.  I don't know of a great way around this issue since the incompatibility is a one time issue (i.e. once the issue is resolved it won't reoccur so we don't need Candlepin to continually check configuration settings on subsequent deployments).

### DIFF
--- a/buildfile
+++ b/buildfile
@@ -138,7 +138,7 @@ HSQLDB_OLD = 'org.hsqldb:hsqldb:jar:1.8.0.10'
 ORACLE = ['com.oracle:ojdbc6:jar:11.2.0', 'org.quartz-scheduler:quartz-oracle:jar:2.1.5']
 
 COMMONS = ['commons-codec:commons-codec:jar:1.4',
-           'commons-collections:commons-collections:jar:3.2',
+           'commons-collections:commons-collections:jar:3.2.2',
            'commons-io:commons-io:jar:1.4',
            'commons-lang:commons-lang:jar:2.5']
 
@@ -164,7 +164,7 @@ GUICE =  [group('guice-assistedinject', 'guice-multibindings',
            'aopalliance:aopalliance:jar:1.0',
            'javax.inject:javax.inject:jar:1']
 
-COLLECTIONS = 'com.google.guava:guava:jar:13.0'
+COLLECTIONS = 'com.google.guava:guava:jar:19.0'
 
 OAUTH= [group('oauth',
               'oauth-provider',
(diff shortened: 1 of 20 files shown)


awood:
[Investigate (&fix?) 1528625 - subscription-manager facts command is showing incomplete facts. [13]](https://trello.com/c/WMiC8Vvb/713-investigate-fix-1528625-subscription-manager-facts-command-is-showing-incomplete-facts-13)

awood:
[[8] 1478549 - Update HornetQ library to address CVE-2015-3208 hornetq: XXE/SSRF in XPath selector [rhn_satellite_6.4]](https://trello.com/c/oxXAcaKr/764-8-1478549-update-hornetq-library-to-address-cve-2015-3208-hornetq-xxe-ssrf-in-xpath-selector-rhnsatellite64)

awood:
> server/bin/deploy needs to be updated to clear journal from new location

Do you mean it should delete `/var/lib/candlepin/hornetq` since it is now unused?

> Question: if Artemis doesn't read the old HornetQ journal, what type of data will be lost? Is this something we need to consider for upgrades?

Old messages that haven't been picked up.  a) I don't know of a way to even handle this in an upgrade since we can't exactly block an upgrade until those messages have been picked up and b) in the past we have not been squeamish about clearing those messages if necessary (e.g. when HornetQ was filling up the disk) and c) I've not put in anything that will remove/clear that directory so the messages won't be destroyed although they will remain inaccessible under normal circumstances.

awood:

mstead:
Verified that this is behaving as it does before the upgrade (including paging).

@awood This thread [1] appears to hint at the idea that an old hornetQ queue can be migrated to Artemis' new style using the import/export tooling. Looks like it would require a not yet released version of Artemis though that addresses some bug fixes. Might be worth seeing if we can give that a try to see if it works. If we can, we may be able to automatically leverage it.

[1] http://activemq.2283324.n4.nabble.com/Migratiion-of-Queue-data-from-HornetQ-to-Artemis-2-x-td4735200.html

kahowell:
> > server/bin/deploy needs to be updated to clear journal from new location

> Do you mean it should delete /var/lib/candlepin/hornetq since it is now unused?

No, I was referring to these lines in the `gendb` function: https://github.com/candlepin/candlepin/blob/61972248a02dcdc16c54b176c601283c487efebc/server/bin/deploy#L27-L30

> This thread [1] appears to hint at the idea that an old hornetQ queue can be migrated to Artemis' new style using the import/export tooling. Looks like it would require a not yet released version of Artemis though that addresses some bug fixes.

@mstead my opinion is let's keep that idea in mind for later; i.e. if we later discover a need to migrate messages, we can come back to it. Given what @awood mentioned about not caring too much about clearing the journal historically, I'm happy with not worrying about it for now.

mstead:
> @mstead my opinion is let's keep that idea in mind for later; i.e. if we later discover a need to migrate messages, we can come back to it. Given what @awood mentioned about not caring too much about clearing the journal historically, I'm happy with not worrying about it for now.

@kahowell Yep. That works for me. Verified with sat folk that they are ok with requiring a 1 time re-sync due to lost messages. They have the tooling to makes this an easy process.

awood:

kahowell:
@awood @mstead FYI one of the previous Jenkins mysql rspec test runs was successful except for some cleanup steps:

```
20:36:04 Removing network rspecjenkinscandlepincandlepinpullrequestspectestsmysql787_default
20:36:04 Network rspecjenkinscandlepincandlepinpullrequestspectestsmysql787_default not found.
```

See https://rhsm-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/candlepin/job/candlepin-pullrequest-spectests-mysql/787/console

Other test failures look like typical test instability that we've seen, so I think we can ignore Jenkins failures on this one. Merging.

PR #1905, opened 2018-02-14
Awood/1545363 bouncycastle

awood:


awood:
[Update BouncyCastle to address CVE](https://trello.com/c/phVxIDxy/775-update-bouncycastle-to-address-cve)

barnabycourt:
Al, do we have an estimate of the performance impact of the switch from SHA1 to SHA256?

awood:
> Al, do we have an estimate of the performance impact of the switch from SHA1 to SHA256?

It should be minimal.  Somewhere around a few hundred extra milliseconds.  See https://automationrhapsody.com/md5-sha-1-sha-256-sha-512-speed-performance/ or

```
% openssl speed sha1 sha256
Doing sha1 for 3s on 16 size blocks: 23817253 sha1's in 2.99s
Doing sha1 for 3s on 64 size blocks: 14319312 sha1's in 3.00s
Doing sha1 for 3s on 256 size blocks: 6842520 sha1's in 3.00s
Doing sha1 for 3s on 1024 size blocks: 2222278 sha1's in 3.00s
Doing sha1 for 3s on 8192 size blocks: 289066 sha1's in 2.99s
Doing sha1 for 3s on 16384 size blocks: 149680 sha1's in 3.00s
Doing sha256 for 3s on 16 size blocks: 13490938 sha256's in 2.99s
Doing sha256 for 3s on 64 size blocks: 7623985 sha256's in 3.00s
Doing sha256 for 3s on 256 size blocks: 3277832 sha256's in 3.00s
Doing sha256 for 3s on 1024 size blocks: 1021067 sha256's in 2.99s
Doing sha256 for 3s on 8192 size blocks: 136766 sha256's in 3.00s
Doing sha256 for 3s on 16384 size blocks: 68266 sha256's in 3.00s
OpenSSL 1.1.0g-fips  2 Nov 2017
built on: reproducible build, date unspecified
options:bn(64,64) md2(char) rc4(16x,int) des(int) aes(partial) idea(int) blowfish(ptr) 
compiler: gcc -DZLIB -DDSO_DLFCN -DHAVE_DLFCN_H -DNDEBUG -DOPENSSL_THREADS -DOPENSSL_NO_STATIC_ENGINE -DOPENSSL_PIC -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DRC4_ASM -DMD5_ASM -DAES_ASM -DVPAES_ASM -DBSAES_ASM -DGHASH_ASM -DECP_NISTZ256_ASM -DPADLOCK_ASM -DPOLY1305_ASM -DPURIFY -DSYSTEM_CIPHERS_FILE="/etc/crypto-policies/back-ends/openssl.config" -DOPENSSLDIR="\"/etc/pki/tls\"" -DENGINESDIR="\"/usr/lib64/engines-1.1\""  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -mtune=generic -Wa,--noexecstack
The 'numbers' are in 1000s of bytes per second processed.
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes  16384 bytes
sha1            127450.18k   305478.66k   583895.04k   758537.56k   791982.83k   817452.37k
sha256           72192.31k   162645.01k   279708.33k   349689.84k   373462.36k   372823.38k
```

But SHA1 is end-of-life so we need to get ahead of the curve.  Akamai is slowly deprecating it for example: https://www.securityweek.com/akamai-kill-support-sha-1-certificates

server/src/main/java/org/candlepin/pki/impl/BouncyCastleProviderLoader.java
```diff
@@ -0,0 +1,38 @@
...
+ * in this software or its documentation.
+ */
+package org.candlepin.pki.impl;
+
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+
+import java.security.Security;
+
+/** When this class is loaded, the Bouncy Castle provider will be installed into the JVM.  The
+ * provider can also be added via a call to a static method if needed in a test for example.
```
kahowell: Please consider mentioning in this comment that `addProvider` can be called more than once with the same provider with no issue (from [the javadocs](https://docs.oracle.com/javase/7/docs/api/java/security/Security.html#addProvider(java.security.Provider)), it will just return -1).

When I first looked I wondered if multiple calls might cause issues...

I think ideally we'd ensure our code only calls `addProvider` once, at initialization, where initialization can be servlet context creation, unit test loading, etc. but after reading the `java.security.Provider.addProvider` javadoc, I think it's no big deal to call it a few places (as is currently written).

awood: Done.

I would rather not squash in that "fix random stuff" commit because a lot of it is not logically related to a switch to BouncyCastle.  Like there are changes in that commit to the Huffman encoding process that it independent of the crypto provider.  I just made the changes to get rid of a lot of warnings before I started the migration in earnest.

PR #1906, opened 2018-02-15
[DTO] Implemented ConsumerDTO & ConsumerTypeDTO APIs for import/export.

nikosmoum:
- Removed the old ConsumerDto that was used in manifest import/export.
- Implemented ConsumerDTO to replace the previous 'ConsumerDto' object throughout import/export endpoints and framework.
- Implemented ConsumerTypeDTO to use throughout import/export endpoints and framework.


nikosmoum:
[[13] Move manifest import/export of Consumer entity to the DTO framework.](https://trello.com/c/KWDmTd6D/771-13-move-manifest-import-export-of-consumer-entity-to-the-dto-framework)

nikosmoum:
Note: I have used '@JsonIgnoreProperties(ignoreUnknown = false)' on ConsumerDTO in order to override the same property which is set to true on the CandlepinDTO (superclass). That is because when we configure the objectmapper with FAIL_ON_UNKNOWN_IMPORT_PROPERTIES to true, it no longer fails, because the value ignoreUnknown=true on CandlepinDTO overrides the objectmapper config.

This might be something to discuss: why does CandlepinDTO have that setting? If it is useful for the API, is it also useful for the Manifest? Should it be somewhere else then?

Ceiu:
That was largely carryover from the AbstractHibernateObject object. Given that we can configure that setting on the ObjectMapper directly, it makes more sense to drop it from the objects and leave it entirely up to the mapper configuration.

nikosmoum:
Removing that setting from CandlepinDTO seems to have catastrophic consequences for the API. Most spec tests fail. The reason this was different for Consumer during import/export, is because we weren't actually using the Consumer model entity, but a 'ConsumerDto' class which isn't a subclass of AbstractHibernateObject.

Ceiu:
It looks like we're keeping the ignoreproperties? With the consumer object being replaced with ConsumerDTO, is this still a problem? What other areas need to be addressed to get the JSON annotations cleaned up?

Ceiu:
Other than that, this looks good, so once we decide on a direction there, we're safe to merge this.

nikosmoum:
@Ceiu I think we should keep the ignoreproperties annotation on CandlepinDTO, since I assume the original one was set on AbstractHibernateObject to override the mapper config, and we want to keep consistent with how the API operates. The ConsumerDto.java class I removed was the only one that didn't have that annotation probably by design (because there's a unit test that changes the mapper config explicitly checks an exception will be thrown). For that reason we should also keep the ignoreproperties on ConsumerDTO.java, which overrides the one on its superclass. (I am trying to guess why the code is the way it is, please let me know if I'm wrong)

server/src/test/java/org/candlepin/dto/api/v1/ConsumerTranslatorTest.java
```diff
@@ -38,7 +38,7 @@
 import java.util.Set;
 
 /**
- * Test suite for the UpstreamConsumerTranslator class
+ * Test suite for the ConsumerTranslator class
```
vritant: my bad

vritant:
looks straightforward, but I would rather the file names for each ConsumerDTO be different. I understand that they are packaged separately, which is great, but naming them something different makes it easier to identify from an IDE, removing the need to check which package each file is from, which is well worth the redundancy in my opinion, especially since we have so many consumer dtos.

@Ceiu  @nikosmoum  any objections?

nikosmoum:
I don't have a specific preference to this. Both options are fine as long as we are consistent.

Ceiu:
I'd prefer to keep the names minimal, as there won't be many situations where we're using a given DTO from multiple packages in the same context. Outside of the global translator registration, I doubt we'll even run into this.

vritant:
okay @Ceiu . @nikosmoum , lets keep it.

PR #1908, opened 2018-02-15
[M] Updated the hibernate-validator library to the version 6.0.7

Ceiu:


Ceiu:
Looks like the following API calls are failing performance tests with this change:
```
"API call: GET consumers/{uuid}/content_overrides............ [FAILED] current avg: 290ms, base line avg: 139ms, dev: 108%, allowed dev: 101.6%",
"API call: GET owners/{ownerkey}/info........................ [FAILED] current avg: 2625ms, base line avg: 1743ms, dev: 50%, allowed dev: 48.6%"
```

The owner info call being the worse of the two, adding ~1s to the average request time. I'm not sure how or why the validator change is adding time to these endpoints, so maybe there is something we can do there to alleviate that. If not, however, we need to either accept that there is going to a performance hit on these two calls, or not perform this update at all.

awood:
Would you consider a3f590e9b33862107c21ebdecc790bcc5e1a2467 as an alternative?

PR #1910, opened 2018-02-16
Product Certificate DTO

wottop:


wottop:
retest this please

Ceiu:
ACKing this on the condition we add another task to re-examine the structure of the certificate DTOs. Looks like I goofed on the first couple and included too much in the CertificateDTO. This really should be a subclass of one of the CertificateDTOs to minimize duplicated code. But we can fix that in another PR when the entire set gets reorganized.

server/src/main/java/org/candlepin/dto/api/v1/ProductCertificateDTO.java
```diff
@@ -0,0 +1,196 @@
...
+        super(source);
+    }
+
+    /**
+     * Retrieves the id field of this ProductCertificateDTO object.
+     *
+     * @return the id field of this ProductCertificateDTO object.
+     */
+    @HateoasInclude
+    public String getId() {
```
nikosmoum: This method shouldn't have `@HateoasInclude`, and should be marked with `@JsonIgnore` since it was also ignored in the model entity.

server/src/main/java/org/candlepin/dto/api/v1/ProductCertificateDTO.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Sets the id to set on this ProductCertificateDTO object.
+     *
+     * @param id the id to set on this ProductCertificateDTO object.
+     *
+     * @return a reference to this ProductCertificateDTO object.
+     */
+    public ProductCertificateDTO setId(String id) {
```
nikosmoum: This should also be marked with `@JsonIgnore` to make sure it isn't set during deserialization, since the model entity didn't have a setter either.

server/src/main/java/org/candlepin/dto/api/v1/ProductCertificateDTO.java
```diff
@@ -0,0 +1,196 @@
...
+    public ProductCertificateDTO setKey(String key) {
+        this.key = key;
+        return this;
+    }
+
+    public String getCert() {
+        return this.cert;
+    }
+
+    public ProductCertificateDTO setCert(String cert) {
```
nikosmoum: Minor suggestion (feel free to ignore): these setters/getters should also have comments just for consistency.

nikosmoum:
@Ceiu I'm not sure if inheritance is a good approach here since there are some differences. EntitlementCertificate and IdentityCertificate, on which CertificateDTO is based on, do serialize/deserialize the id field, while ProductCertificate is ignoring it.

nikosmoum:
[Add API DTO for ProductCertificate](https://trello.com/c/DUH0JBrL/778-add-api-dto-for-productcertificate)

Ceiu:
@nikosmoum Ahh, maybe that's why they're setup the way they are. Some of those details are starting to fade from memory. :)

PR #1913, opened 2018-02-16
Replace explicit types with the diamond operator

awood:
Note that if you are using the IntelliJ checkstyle plugin, you will need to `mvn install` the `checks` project and then point the plugin to the JAR in the "Third-Party Checks" section (illustrated below).

settings_002

### DIFF
--- a/server/src/main/java/org/candlepin/policy/js/compliance/ComplianceStatus.java
+++ b/server/src/main/java/org/candlepin/policy/js/compliance/ComplianceStatus.java
@@ -48,12 +48,12 @@ public class ComplianceStatus {
     private Set<ComplianceReason> reasons;
 
     public ComplianceStatus() {
-        this.nonCompliantProducts = new HashSet<String>();
-        this.compliantProducts = new HashMap<String, Set<Entitlement>>();
-        this.partiallyCompliantProducts = new HashMap<String, Set<Entitlement>>();
-        this.partialStacks = new HashMap<String, Set<Entitlement>>();
-        this.productComplianceDateRanges = new HashMap<String, DateRange>();
-        this.reasons = new HashSet<ComplianceReason>();
+        this.nonCompliantProducts = new HashSet<>();
+        this.compliantProducts = new HashMap<>();
+        this.partiallyCompliantProducts = new HashMap<>();
+        this.partialStacks = new HashMap<>();
+        this.productComplianceDateRanges = new HashMap<>();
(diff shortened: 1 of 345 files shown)


server/src/main/java/org/candlepin/policy/js/compliance/ComplianceStatus.java
```diff
@@ -101,15 +101,15 @@ public void addNonCompliantProduct(String productId) {
 
     public void addPartiallyCompliantProduct(String productId, Entitlement entitlement) {
         if (!partiallyCompliantProducts.containsKey(productId)) {
-            partiallyCompliantProducts.put(productId, new HashSet<Entitlement>());
+            partiallyCompliantProducts.put(productId, new HashSet<>());
```
Ceiu: Depending on version requirements, we may need to be careful using the diamond operator in contexts like this. I believe in Java7, this doesn't resolve properly/at all, so it may not work. In Java 8, using it like this on a generic method also fails, since it ends up resolving to Object. We're safe here since the method is not generic, but we may see some confusion when we see the same code structure fail in one place and not in another.

server/src/main/java/org/candlepin/util/Util.java
```diff
@@ -107,19 +107,19 @@ public static String generateDbUUID() {
     }
 
     public static <E> List<E> newList() {
-        return new ArrayList<E>();
+        return new ArrayList<>();
```
Ceiu: This actually works...??

awood: I'm assuming that the compiler can infer the type from the method signature?  But why do we even have that method.  Seems like a waste of time and it's only referenced from unit test classes.

kahowell:
[Replace explicit types with the diamond operator](https://trello.com/c/1EVj2tvA/784-replace-explicit-types-with-the-diamond-operator)

vritant:
@awood marking this in progress until we merge dtos branch in.

server/src/main/java/org/candlepin/util/Util.java
Ceiu: Definitely -- especially the list-based ones where Collections and Arrays natively provide better functionality.

awood:

kahowell:
@awood FYI looks like this needs rebase now.

awood:

awood:

Ceiu:
ACK

PR #1914, opened 2018-02-16
1531627: Reduce possibility of lock waits on Async Hypervisor Check in

cnsnyder:
This is essentially a forward port of bz 1531628.
The biggest difference here is this version of the fix reuses existing guest id entries where applicable to ensure a guest is only mapped to one host (for a given organization).

Note: I will remove the lock wait repro script and squash the commits post-review.

### DIFF
--- a/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
+++ b/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
@@ -179,7 +179,20 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
      */
     @Transactional
     public E create(E entity) {
-        save(entity);
+        return create(entity, true);
+    }
+
+    /**
+     * @param entity to be created.
+     * @param flush whether or not to flush after the persist.
+     * @return newly created entity.
+     */
+    @Transactional
+    public E create(E entity, boolean flush) {
+        getEntityManager().persist(entity);
+        if (flush) {
+            flush();
+        }
         return entity;
     }
 
@@ -545,8 +558,7 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
 
     @Transactional
     protected void save(E anObject) {
-        getEntityManager().persist(anObject);
-        flush();
+        create(anObject, true);
     }
 
     public void flush() {
--- a/server/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/server/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -27,6 +27,7 @@ import com.google.inject.persist.Transactional;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.hibernate.Criteria;
+import org.hibernate.FetchMode;
 import org.hibernate.Hibernate;
 import org.hibernate.Query;
 import org.hibernate.ReplicationMode;
@@ -42,6 +43,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -52,6 +54,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 
 import javax.persistence.LockModeType;
+import javax.persistence.TypedQuery;
 
 
 
@@ -75,11 +78,10 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
 
     @Transactional
     @Override
-    public Consumer create(Consumer entity) {
+    public Consumer create(Consumer entity, boolean flush) {
         entity.ensureUUID();
         this.validateFacts(entity);
-
-        return super.create(entity);
+        return super.create(entity, flush);
     }
 
     @Override
@@ -395,7 +397,7 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
         Consumer existingConsumer = find(updatedConsumer.getId());
 
         if (existingConsumer == null) {
-            return this.create(updatedConsumer);
+            return this.create(updatedConsumer, flush);
         }
 
         // TODO: Are any of these read-only?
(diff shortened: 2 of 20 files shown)


cnsnyder:
[[20] 1531627 - Two or more hypervisor update jobs can cause lockwait timeouts](https://trello.com/c/iGjKyakT/743-20-1531627-two-or-more-hypervisor-update-jobs-can-cause-lockwait-timeouts)

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -76,10 +79,19 @@ public ConsumerCurator() {
...
+        return super.create(commonCreate(entity), true);
+    }
+
+    @Transactional
+    @Override
+    public Consumer create(Consumer entity, boolean flush) {
+        return super.create(commonCreate(entity), flush);
+    }
+
+    public Consumer commonCreate(Consumer entity) {
```
Ceiu: Since the methods in AbstractHibernateCurator are generic, the only reason you'd override them at all in this subclass is to add stuff like commonCreate does. A better way to implement all of this without adding additional overhead would be to remove commonCreate and create, and just add the ensureUUID and validateFacts call to the two-param create method.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -542,6 +564,73 @@ public Consumer getHost(String guestId, Owner... owners) {
...
+
+        String hql = "SELECT gid FROM GuestId gid " +
+            "WHERE gid.consumer.owner = :owner AND gid.guestIdLower IN (:guest_ids)" +
+            "ORDER BY gid.updated DESC";
+
+        TypedQuery<GuestId> query = this.getEntityManager().createQuery(hql, GuestId.class);
+        Map<String, GuestId> output = new HashMap<String, GuestId>();
+
+        for (List<String> block : this.partition(guestIds)) {
+            List<String> sanitized = new ArrayList<String>(block);
```
Ceiu: Whoops... made a mistake here. That needs to be block.size(), not block. As-is, we'll double our block size and potentially exceed our param limit here.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -542,6 +564,73 @@ public Consumer getHost(String guestId, Owner... owners) {
...
+    private boolean safeDateAfter(Date d1, Date d2) {
+        return d1 != null && (d2 == null || d1.after(d2));
+    }
+
+
+//    @Transactional
+//    public VirtConsumerMap getGuestsHostBulk(List<String> guestIds, Owner... owners) {
+//        VirtConsumerMap vcMap = new VirtConsumerMap();
+//
+//    }
```
Ceiu: Dead block of code that can be removed.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -267,46 +280,42 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
...
+                        !jobReporterId.contentEquals(reportedOnConsumer.getHypervisorId().getReporterId()))) {
                     reportedOnConsumer.getHypervisorId().setReporterId(jobReporterId);
                 }
                 else if (jobReporterId == null) {
                     log.debug("hypervisor checkin reported asynchronously without reporter id " +
                         "for hypervisor:{} of owner:{}", hypervisorId, ownerKey);
                 }
             }
+
+            for (Consumer consumer : hypervisorConsumersMap.getConsumers()) {
```
Ceiu: I think we need to set the lastUpdatedTime or lastCheckinTIme or something here. What'd we end up doing in the 2.2 version of this change?

server/src/main/java/org/candlepin/resource/util/GuestMigration.java
```diff
@@ -96,27 +95,16 @@ public GuestMigration buildMigrationManifest(Consumer incoming, Consumer existin
             for (GuestId guestId : removedGuests) {
                 existingGuests.remove(guestId);
                 log.debug("Guest ID removed: {}", guestId);
-                sink.queueEvent(eventFactory.guestIdDeleted(guestId));
+//                sink.queueEvent(eventFactory.guestIdDeleted(guestId));
```
Ceiu: Dead code?

server/src/main/java/org/candlepin/resource/util/GuestMigration.java
```diff
@@ -96,27 +95,16 @@ public GuestMigration buildMigrationManifest(Consumer incoming, Consumer existin
...
-                    // TODO is it possible for real guests to not resolve to a host? It occurs in unit tests.
-                    Consumer oldHost = consumerCurator.getHost(guest);
-                    if (oldHost != null) {
-                        manifest.addOldMapping(oldHost, guest);
-                    }
-                }
-
                 log.debug("New guest ID added: {}", guestId);
-                sink.queueEvent(eventFactory.guestIdCreated(guestId));
+//                sink.queueEvent(eventFactory.guestIdCreated(guestId));
```
Ceiu: Dead?

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
cnsnyder: We did update the lastCheckinTime here in the 2.2 version. I've set the lastCheckinTime during createConsumerForHypervisorId in this branch, to avoid multiple update calls generated by hibernate (we do one update in GuestMigration and a create for new consumers at the end of the job).

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -79,19 +79,17 @@ public ConsumerCurator() {
     @Transactional
     @Override
     public Consumer create(Consumer entity) {
```
Ceiu: Almost. This method need not exist -- only the two-method version with flush. In AbstractHibernateCurator we know it calls flush(entity, bool), so we can just override that one method and cover both.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -179,7 +179,20 @@ public E secureFind(Serializable id) {
      */
     @Transactional
     public E create(E entity) {
-        save(entity);
+        return create(entity, true);
```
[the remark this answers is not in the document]
Ceiu: That's the default behavior today. The point of these additions/changes is to allow a caller to specify that they will flush at a later time without breaking the other existing calls.

awood:
I'm having a little trouble getting the big picture on this PR.  Would you mind fleshing out the commit message a little bit more describing what was causing the problem and how your patch's changes fix it

PR #1917, opened 2018-02-23
[M] Added several improvements to the request scanner

Ceiu:
- Added the showall option, to display info for every processed
  request
- Added the idsonly option for displaying only the request IDs for
  processed requests
- Added checks and reporting for failed requests, where failed is
  a request returning a response code outside the 200 range
- Added the --uri and --method options for filtering requests to
  those matching the method and/or URI filters specified

### DIFF
--- a/server/bin/request_scanner.rb
+++ b/server/bin/request_scanner.rb
@@ -32,6 +32,7 @@ def process_log(filename)
   requests = {}
   proc_requests = {}
   slow_requests = {}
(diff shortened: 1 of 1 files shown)


Ceiu:
Performance failure is due to it being aborted early, and as this isn't a change to Candlepin itself, there is literally zero possibility it will actually impact performance.

wottop:
ack

PR #1922, opened 2018-02-27
Pull corrected translations from Zanata.

awood:
We had several translations that incorrectly escaped single quotes.  I
corrected the translations, pushed them to Zanata, and then pulled them
back.  Zanata apparently performs some clean-up (deleting trailing
whitespace, etc) so I made the pull to avoid conflicts in the future.

### DIFF
--- a/common/po/fr.po
+++ b/common/po/fr.po
@@ -1,5 +1,6 @@
 # croe <croe@redhat.com>, 2017. #zanata
 # ljanda <ljanda@redhat.com>, 2017. #zanata
+# awood <awood@redhat.com>, 2018. #zanata
 # khowell <khowell@redhat.com>, 2018. #zanata
 msgid ""
 msgstr ""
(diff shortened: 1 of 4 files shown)


common/po/fr.po
```diff
@@ -9,8 +10,8 @@ msgstr ""
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"PO-Revision-Date: 2018-01-25 08:40+0000\n"
-"Last-Translator: khowell <khowell@redhat.com>\n"
+"PO-Revision-Date: 2018-02-27 04:43+0000\n"
+"Last-Translator: Copied by Zanata <copied-by-zanata@zanata.org>\n"
```
Ceiu: Is this something we want to avoid somehow?

Ceiu:
ACK

The only question I have is if we want to (and are able to) avoid having the last translator/reviewer clobbered by Zanata's copying/juggling operation.

common/po/fr.po
kahowell: No biggie IMO.

PR #1923, opened 2018-02-27
Correct incorrect punctuation in user messages.

awood:
Single quotes need to be escaped with single quotes in po files. This is
an unusual escape sequence and it is not obvious to translators that the
translated key also needs to use two single quotes.  Double quotes are
escaped with a backslash.

Besides, quoting these strings in single quotes was grammatically
incorrect in the first place.

### DIFF
--- a/server/spec/consumer_resource_dev_spec.rb
+++ b/server/spec/consumer_resource_dev_spec.rb
@@ -88,7 +88,7 @@ it 'should not allow entitlement for consumer past expiration' do
     consumer = @user.register(random_string('system'), type=:system, nil, facts = {:dev_sku => "dev_product"}, @username,
               @owner['key'], [], [], nil, [], nil, [], created_date)
     consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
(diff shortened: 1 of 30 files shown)


awood:

awood:

server/spec/consumer_resource_dev_spec.rb
```diff
@@ -88,7 +88,8 @@ def auto_attach_and_verify_dev_product(expected_product_id, old_ent_id=nil)
     consumer = @user.register(random_string('system'), type=:system, nil, facts = {:dev_sku => "dev_product"}, @username,
               @owner['key'], [], [], nil, [], nil, [], created_date)
     consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
-    expected_error = "Unable to attach subscription for the product 'dev_product': Subscriptions for dev_product expired on:"
+    expected_error = "Unable to attach subscription for the product \"dev_product\": Subscriptions for
+    dev_product expired on:"
```
Ceiu: This line break is causing the failures with the spec tests

awood: Fixed

PR #1925, opened 2018-03-05
Custom UserType for job result data

awood:
This PR is the first half of the effort to stop storing serialized objects in cp_job's resultData column.  The idea is that once this is merged we can modify the ResultDataUserType so that it can load either serialized objects (for legacy records) **or** UTF8 JSON stored in the column as bytes.  Storing the JSON in a `bytea` column seems a little hacky but it's backwards compatible and won't require a DB migration.  The DB migration would be tricky since migrating it will require writing a custom task in Java that deserializes the data from the column, re-serializes the data as JSON, and then writes it to the column of type `text`.

I'm filing this partly to avoid bit-rot while I'm working on something else this week and also to get opinions on whether this solution is acceptable to the team.

### DIFF
--- a/common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
+++ b/common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
@@ -60,6 +60,7 @@ public class LoggingFilter implements Filter {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public void doFilter(ServletRequest request, ServletResponse response,
         FilterChain chain) throws IOException, ServletException {
 
--- a/common/src/test/java/org/candlepin/common/jackson/DynamicPropertyFilterTest.java
+++ b/common/src/test/java/org/candlepin/common/jackson/DynamicPropertyFilterTest.java
@@ -66,6 +66,7 @@ public class DynamicPropertyFilterTest {
     }
 
     @Test
+    @SuppressWarnings("unchecked")
     public void nonEmptyIsSerializable() {
         //Implicitly return true
         when(dynamicFilterData.isAttributeExcluded(anyList()))
--- a/server/src/main/java/org/candlepin/hibernate/ResultDataUserType.java
+++ b/server/src/main/java/org/candlepin/hibernate/ResultDataUserType.java
@@ -1,0 +1,143 @@
+/**
+ * Copyright (c) 2009 - 2018 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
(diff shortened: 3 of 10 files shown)


awood:

wottop:
I'm good with it. Will wait for more opinions to merge.

kahowell:
@awood, so the idea is when we are writing JSON, we will set result data to a `String`, so the representation will end up as a serialized Java `String`?

I was curious, so I did some quick little tests to see how the overhead of ObjectOutputStream is...

```groovy
// test1.groovy
def serialize(i) {
  def test = "foobar" + i
  def baos = new ByteArrayOutputStream()
  def oos = new ObjectOutputStream(baos)
  oos.writeObject(test)
  def data = baos.toByteArray()
}

for (def i = 0; i < 100000; i++) {
  serialize(i)
}
```
vs.
```groovy
// test2.groovy 
def serialize(i) {
  def test = "foobar" + 1
  def data = test.getBytes();
}

for (def i = 0; i < 100000; i++) {
  serialize(i)
}
```

```
$ time groovy test1.groovy 

real	0m0.808s
user	0m1.841s
sys	0m0.104s
```

```
$ time groovy test2.groovy 

real	0m0.579s
user	0m1.126s
sys	0m0.072s
```

I think we *could* in the future optimize this if we wanted, but meh why bother for now. (I'm mostly just thinking aloud).

So, +1 from me.

@mstead, thoughts on the PR?

awood:
> @awood, so the idea is when we are writing JSON, we will set result data to a String, so the representation will end up as a serialized Java String?

My thought was, in order to remains backwards compatible, we keep the `result_data` column as a bytea.  Older candlepins will serialize Java objects to it like they always did and newer candlepins will just utf8 encode JSON.

The UserType would then be modified to decode the row data as UTF-8 and attempt to read it as JSON.  If that fails (because the row data is actually a Java object) then we try to deserialize it.  On the storage side, we modify the UserType to just send the object through a Jackson ObjectMapper to turn it into JSON.

nikosmoum:
Looks good. So when we read the `result_data` from the DB as a UTF-8 String, and then read it as JSON, we can mark the JobStatusDTO's resultData field with @JsonRawValue in order for it to be included literally instead of serialized as an object, which will save us a serialization.

vritant:

mstead:
> I think we could in the future optimize this if we wanted, but meh why bother for now. (I'm mostly just thinking aloud).
> 
> So, +1 from me.
> 
> @mstead, thoughts on the PR?

I'm Ok with this. Let's optimize this if/when we need to. :+1:

PR #1926, opened 2018-03-05
Add SubjectKeyIdentifierWriter extension point back into Candlepin.

awood:
Hosted-mode Candlepin instances can have special requirements around the
SubjectKeyIdentifier extension written into certificates.  This
extension point was removed in the process of upgrading to a newer
release of BouncyCastle.

### DIFF
--- a/server/src/main/java/org/candlepin/guice/DefaultConfig.java
+++ b/server/src/main/java/org/candlepin/guice/DefaultConfig.java
@@ -14,6 +14,8 @@
  */
 package org.candlepin.guice;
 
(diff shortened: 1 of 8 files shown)


vritant:
Where is this coming from? any context?

vritant:
does it make sense to write any unit tests?

awood:
> does it make sense to write any unit tests?

Honestly, I don't really think it does.  The default implementation doesn't really have any logic; it's just calling out to BouncyCastle.

vritant:
I figured.

PR #1927, opened 2018-03-09
1546236: Properly handle connections when they are recreated

mstead:
https://bugzilla.redhat.com/show_bug.cgi?id=1546236

Whenever the QpidConnection needed to be rebuilt, the existing
connection was being disregarded and a new one created. The old
connections were not getting cleaned up causing candlepin to
exhaust all connections available to Qpid.

This was primarily surfacing when HornetQ would try to resend
failed messages but the Qpid queue was in FLOW_STOPPED. In
this case, QpidConnection would be marked as stale because of the
exception, and would create a new connection to Qpid, leaving
the other around. In the case of FLOW_STOPPED, a new connection
isn't required - we just need to wait for the queue to catch up.

Since the exceptions thrown by Qpid for the various failure
cases, we could never tell exactly when to recreate the connection
or when the connection was Ok.

This patch addresses this PR by monitoring the Qpid status and
only creating new connections when required as well as cleaning
up after itself.

Notes:
- Pulled Qpid connection checking via QMF into an isolated service
  separate from SuspendModeTransitioner.
- Added a listener framework for objects to be notified when the latest
  Qpid status is determined.
- SuspendModeTransitioner now listens for Qpid status updates instead
  of running the checks itself.
- QpidConnection now listens for status updates so that it can better
  manage its connection state (connect, reconnect)
  * When Qpid is FLOW_STOPPPED, block message sending to Qpid until
    it is notified that Qpid is able to accept the message.
  * When notified that Qpid went down, close the old connection so that
    it is no longer hanging around.
  * When notified that Qpid is back up, re-create the connection again.

When deploying the patch, apply the following to the candlepin config to see the logging info:

```
log4j.logger.org.candlepin.controller.QpidStatusMonitor=DEBUG
log4j.logger.org.candlepin.controller.SuspendModeTransitioner=DEBUG
log4j.logger.org.candlepin.audit.QpidConnection=DEBUG
```

### DIFF
--- a/server/src/main/java/org/candlepin/controller/QpidStatusMonitor.java
+++ b/server/src/main/java/org/candlepin/controller/QpidStatusMonitor.java
@@ -1,0 +1,153 @@
+/**
+ * Copyright (c) 2009 - 2018 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.controller;
+
+import com.google.inject.Inject;
+import org.candlepin.audit.QpidConnection;
+import org.candlepin.audit.QpidQmf;
+import org.candlepin.audit.QpidStatus;
+import org.candlepin.common.config.Configuration;
+import org.candlepin.config.ConfigProperties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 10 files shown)


mstead:
retest this please

mstead:
retest this please

server/src/main/java/org/candlepin/controller/QpidStatusMonitor.java
```diff
@@ -0,0 +1,152 @@
...
+             * hope that this error won't infinitely recur with each scheduled execution
+             */
+            return;
+        }
+
+        log.debug("Qpid connection status - Old: {} New: {}", this.lastStatus, status);
+        try {
+            // If a listener throws an exception, log it, and move to the next listener.
+            // We don't want to skip other listeners just because another failed.
+            fireStatusChange(this.lastStatus, status);
```
awood: I'm probably reading this wrong, but the way I read it, this code does the opposite of what the comment says.  In other words, a call is made to `fireStatusChange` which enters a loop.  If any listener in the loop throws an exception, the looping ends, and we proceed to the catch statement.  I would think to match the comment, we would want something like

```java
for (QpidStatusListener listener: this.listeners) {
    try {
        listener.onStatusUpdate(oldStatus, newStatus);
    } catch (Exception e) {
        log.warn("Failed to notify status listener", e);
    }
}
```

mstead: :+1:  I'll fix that up. Nice catch. Appreciated.

mstead:
@awood Fixed the issue you noted.

mstead:
Still working on making jenkins happy. Tests are fine when run locally, but not on jenkins.

PR #1928, opened 2018-03-13
Decouple from BouncyCastle

awood:
The goal of this PR is to decouple how we read PEM encoded private keys from the BouncyCastle libraries.

A few notes:
* Reading encrypted PKCS8 files does not currently work due to a [bug](https://bugs.openjdk.java.net/browse/JDK-8076999) in OpenJDK.  We only support encrypted PKCS8 files today in the sense that the BouncyCastle PEMParser class will read them, but we were doing some opinionated casting in the old `BouncyCastlePKIReader.loadPrivateKey()` that looks like it would fail if we did try to provide a PKCS8 format private key.  Satellite installations would be using an OpenSSL style PKCS1 normally anyway.
* A lot of the code is extracted/adapted from PEMParser and PemReader.  The code around peeling off the `-----BEGIN...` headers, for example.
* This PR still uses BouncyCastle for the PKCS1 parsers.  When it comes time to switch, we can just replace the implementations of `PKCS1PrivateKeyPemParser` and `PKCS1EncryptedPrivateKeyPemParser` to use the replacement crypto library.  In fact, `PKCS1PrivateKeyPemParser` doesn't actually perform any crypto operations that the JDK itself can't do; it just needs BouncyCastle to parse the ASN1 so it can get the RSA primitives it needs to build the key.


awood:

server/src/main/java/org/candlepin/pki/PrivateKeyReader.java
```diff
@@ -0,0 +1,325 @@
...
+            Enumeration asn1 = seq.getObjects();
+
+            BigInteger v = ((ASN1Integer) asn1.nextElement()).getValue();
+            if (v.intValue() != 0 && v.intValue() != 1)
+            {
+                throw new IllegalArgumentException("wrong version for RSA private key");
+            }
+
+            // See RFC 8017 A.1.2
+            BigInteger version = v;
```
kahowell: Nitpick: unused var `version`.

server/src/main/java/org/candlepin/pki/PrivateKeyReader.java
```diff
@@ -0,0 +1,325 @@
...
+            BigInteger publicExponent = ((ASN1Integer) asn1.nextElement()).getValue();
+            BigInteger privateExponent = ((ASN1Integer) asn1.nextElement()).getValue();
+            BigInteger primeP = ((ASN1Integer) asn1.nextElement()).getValue();
+            BigInteger primeQ = ((ASN1Integer) asn1.nextElement()).getValue();
+            BigInteger primeExponentP = ((ASN1Integer) asn1.nextElement()).getValue();
+            BigInteger primeExponentQ = ((ASN1Integer) asn1.nextElement()).getValue();
+            BigInteger coefficient = ((ASN1Integer) asn1.nextElement()).getValue();
+
+            if (asn1.hasMoreElements()) {
+                ASN1Sequence otherPrimeInfos = (ASN1Sequence) asn1.nextElement();
```
kahowell: Nitpick: unused var `otherPrimeInfos`

server/src/test/resources/keys/README
```diff
@@ -0,0 +1 @@
+The string 'password' is the password for all the encrypted keys.
```
kahowell: This is not true for `DESTestKey.pem`. Could just make a note of that here.

server/src/main/java/org/candlepin/pki/impl/BouncyCastlePKIUtility.java
```diff
@@ -78,8 +78,7 @@
  * objects in PEM format (for saving to the db or sending to the client), and
  * decode raw ASN.1 DER values (as read from a Certificate/CRL).
  *
- * All code that imports bouncycastle should live either in this module,
- * or in {@link BouncyCastlePKIReader}
+ * All code that imports bouncycastle should live in this module.
```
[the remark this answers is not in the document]
awood: My goal in this PR is not to eliminate all usages of BouncyCastle, but to eliminate usages of `PEMParser` and `PemReader` since these are very high level classes that use a lot of BouncyCastle internals.  Even still, for PKCS1 style keys, we're continuing to use the BouncyCastle ASN1 parsing tools.

server/src/test/resources/keys/README
awood: I'm going to change the key instead :)

awood:

awood:

awood:

PR #1929, opened 2018-03-13
1365248: Reconcile registered consumers and hypevisors

wottop:
Will affect libvirt scenarios
Uses the system_uuid, when available, to match an existing
 consumer to a new hypervisor or an existing hypervisor to a new consumer.
Will work when the system_uuid on the hypervisor is used
 as the submitted hypervisor id.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -567,10 +567,23 @@ public class ConsumerResource {
             entity.setGuestIds(guestIds);
         }
 
-        if (dto.getHypervisorId() != null) {
-            Owner owner = ownerCurator.findOwnerById(entity.getOwnerId());
-            HypervisorId hypervisorId = new HypervisorId(entity, owner,
-                dto.getHypervisorId().getHypervisorId(), dto.getHypervisorId().getReporterId());
+        if (dto.getHypervisorId() != null && entity.getOwnerId() != null) {
+            HypervisorId hypervisorId = new HypervisorId(
+                entity,
+                ownerCurator.findOwnerById(entity.getOwnerId()),
+                dto.getHypervisorId().getHypervisorId(),
+                dto.getHypervisorId().getReporterId());
+            entity.setHypervisorId(hypervisorId);
+        }
+
+        if (dto.getHypervisorId() == null &&
+            dto.getFact("system_uuid") != null &&
+            !"true".equals(dto.getFact("virt.is_guest")) &&
+            entity.getOwnerId() != null) {
+            HypervisorId hypervisorId = new HypervisorId(
+                entity,
+                ownerCurator.findOwnerById(entity.getOwnerId()),
+                dto.getFact("system_uuid"));
             entity.setHypervisorId(hypervisorId);
         }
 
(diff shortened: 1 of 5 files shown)


barnabycourt:
Performance test running at https://rhsm-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/candlepin/job/CandlepinPerformance/180/

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -108,7 +106,10 @@ public HypervisorResource(ConsumerResource consumerResource, ConsumerCurator con
         this.translator = translator;
         this.guestIdResource = guestIdResource;
 
-        this.hypervisorType = consumerTypeCurator.lookupByLabel(ConsumerTypeEnum.HYPERVISOR.getLabel(), true);
+        this.hypervisorType = consumerTypeCurator.lookupByLabel(ConsumerTypeEnum.HYPERVISOR.getLabel());
```
Ceiu: This looks like a rebasing artifact here. We should keep the original line.

server/src/test/java/org/candlepin/resource/HypervisorResourceTest.java
```diff
@@ -127,12 +127,16 @@ public void setupTest() {
...
 
         this.hypervisorType = new ConsumerType(ConsumerTypeEnum.HYPERVISOR);
+
         this.hypervisorType.setId("test-hypervisor-ctype");
 
         this.mockConsumerType(this.hypervisorType);
 
         this.translator = new StandardTranslator(this.consumerTypeCurator);
 
+        when(consumerTypeCurator.lookupByLabel(eq(ConsumerTypeEnum.HYPERVISOR.getLabel())))
```
Ceiu: Another rebasing issue here. Pretty sure this is handled or the mocking needs to be changed. The existing code in master should work for this.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -606,7 +613,25 @@ public ConsumerDTO create(
...
+        if (ownerKey != null && dto.getFact("system_uuid") != null &&
+            !"true".equalsIgnoreCase(dto.getFact("virt.is_guest"))) {
+            Owner owner = ownerCurator.lookupByKey(ownerKey);
+            if (owner != null) {
+                consumer = consumerCurator.getHypervisor(dto.getFact("system_uuid"), owner);
+                if (consumer != null) {
+                    consumer.setIdCert(generateIdCert(consumer, false));
+                    this.updateConsumer(consumer.getUuid(), dto, principal);
+                    return translator.translate(
+                        consumerCurator.findByUuid(consumer.getUuid()), ConsumerDTO.class);
```
Ceiu: Do we need to do another lookup here? We have the consumer already, is there a reason we're not just translating it straight away?

Ceiu:
This is soft-blocked on #1930 to avoid another rebase of a much larger commit. This PR will likely require yet another rebase after that is merged in, unfortunately.

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -110,6 +108,9 @@ public HypervisorResource(ConsumerResource consumerResource, ConsumerCurator con
         this.guestIdResource = guestIdResource;
 
         this.hypervisorType = consumerTypeCurator.lookupByLabel(ConsumerTypeEnum.HYPERVISOR.getLabel(), true);
+        if (this.hypervisorType == null) {
```
Ceiu: Rebasing artifact; This block is unnecessary with the above boolean -- it'll do this creation if the type isn't found.

PR #1931, opened 2018-03-15
Add reasons back to compliance.created events

kahowell:
Katello only uses name and message. See https://github.com/Katello/katello/blob/8bc1ef327073f03149b4dd29e0284b65c4b5626a/app/services/katello/candlepin/consumer.rb#L125


Ceiu:
As a side note, the code itself is good, but do we care that we're re-introducing inconsistency in the events by having reasons on create but not on the others?

kahowell:
@Ceiu squashed

barnabycourt:
Merging as all the tests have passed. QPID has not updated jenkins but it did pass.

PR #1936, opened 2018-03-19
[M] 1558060: Added precondition check to add-share-consumer task

Ceiu:
- Added a pre-condition check to the Liquibase task to add a share
  consumer. This will prevent a failure that could occur in the
  event a consumer type named "share" already exists.


mstead:
ACK once jenkins is happy again. :(

PR #1937, opened 2018-03-19
[2.1] 1558065: Added precondition check to add-share-consumer task

Ceiu:
- Added a pre-condition check to the Liquibase task to add a share
  consumer. This will prevent a failure that could occur in the
  event a consumer type named "share" already exists.


mstead:
ACK once Jenkins is happy again.

Ceiu:
This PR is temporarily blocked due to a quoting problem in Jenkins. It should be addressed with an update to the Zanata translations and the updates in PR #1938

PR #1938, opened 2018-03-19
[2.1] Correct incorrect punctuation in user messages.

Ceiu:
Single quotes need to be escaped with single quotes in po files. This is
an unusual escape sequence and it is not obvious to translators that the
translated key also needs to use two single quotes.  Double quotes are
escaped with a backslash.

Besides, quoting these strings in single quotes was grammatically
incorrect in the first place.


Ceiu:
This is a backport of #1923

awood:
Doesn't look like the backport did what we needed it to do?  The checkstyle job is still registering a lot of complaints.

kahowell:
> Doesn't look like the backport did what we needed it to do?  The checkstyle job is still registering a lot of complaints.

Let's retest once #1939 lands.

I think it's still worth landing this backport so we can have consistency between master and the 2.1 branch in translations and punctuation.

kahowell:
retest this please

PR #1940, opened 2018-03-20
Optimized DTO usage in API endpoints

nikosmoum:
- The following endpoints were optimized to do data assignment from DTO to entity (as opposed to from entity to entity) where possible, and also do data validation before data assignment:
  ConsumerResource.create
  ConsumerResource.updateConsumer
  OwnerResource.createOwner
  OwnerResource.updateOwner
  OwnerResource.createActivationKey
  OwnerResource.createEnv
  OwnerResource.createPool
  OwnerResource.updatePool

- Updated a few forgotten endpoints in ProductResource & ContentResource to use OwnerDTO and ContentDTO instead of the equivalent model entities.
- Refactored a few tests in ContentResourceTest that are testing deprecated endpoints.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -50,6 +50,7 @@ import org.candlepin.dto.api.v1.CertificateDTO;
 import org.candlepin.dto.api.v1.ConsumerDTO;
 import org.candlepin.dto.api.v1.ConsumerInstalledProductDTO;
 import org.candlepin.dto.api.v1.GuestIdDTO;
+import org.candlepin.dto.api.v1.HypervisorIdDTO;
 import org.candlepin.dto.api.v1.OwnerDTO;
 import org.candlepin.dto.api.v1.PoolQuantityDTO;
 import org.candlepin.model.CandlepinQuery;
@@ -290,6 +291,25 @@ public class ConsumerResource {
 
     /**
      * Sanitizes inbound consumer facts, truncating long facts and dropping invalid or untracked
+     * facts. The consumer DTO will be updated in-place.
+     *
+     * @param consumerDTO
+     *  The consumer DTO to populate with sanitized facts
+     */
+    public void sanitizeConsumerFacts(ConsumerDTO consumerDTO) {
+        if (consumerDTO != null) {
+            Map<String, String> facts = consumerDTO.getFacts();
+
+            if (facts != null && facts.size() > 0) {
+                log.debug("Sanitizing facts for consumer {}", consumerDTO.getName());
+                Map<String, String> sanitized = sanitizeConsumerFacts(facts);
+                consumerDTO.setFacts(sanitized);
+            }
+        }
+    }
+
+    /**
+     * Sanitizes inbound consumer facts, truncating long facts and dropping invalid or untracked
      * facts. The consumer will be updated in-place.
      *
      * @param consumer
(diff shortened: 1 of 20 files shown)


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -675,20 +727,17 @@ public Consumer createConsumerFromEntity(Consumer consumer, Principal principal,
             }
         }
 
-        if (consumer.getEntitlementCount() != 0) {
+        if (consumer.getEntitlementCount() != null || consumer.getEntitlementCount() != 0) {
```
Ceiu: These type of checks should be unnecessary now -- they won't be leaving the dto.

nikosmoum:
Note: On the updatePool endpoint, I removed a few lines that assign values from the managed entity to the non-managed entity from the section commented as "Forcefully set fields we don't allow the client to change.", but I did not remove all of them, since they seem to be used further down in the CandlepinPoolManager.updateMasterPool call and its nested calls. 

I removed these which didn't seem to be accessed:
```
            entity.setConsumed(currentEntity.getConsumed());
            entity.setExported(currentEntity.getExported());
            entity.setShared(currentEntity.getShared());
            entity.setEntitlements(currentEntity.getEntitlements());
            entity.setRestrictedToUsername(currentEntity.getRestrictedToUsername());
            entity.setSourceEntitlement(currentEntity.getSourceEntitlement());
            entity.setSourceStack(currentEntity.getSourceStack());
            entity.setActiveSubscription(currentEntity.getActiveSubscription());
```

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1769,10 +1895,7 @@ public PoolDTO createPool(@PathParam("owner_key") @Verify(Owner.class) String ow
         pool.setOwner(owner);
 
         // Populate the rest of the pool
-        this.populateEntity(pool, inputPoolDTO);
-
-        // Resolve products
-        //pool = resolverUtil.resolvePool(pool);
+        this.populateEntityOnCreate(pool, inputPoolDTO);
```
Ceiu: We shouldn't need separate methods for create/update like this. Do what we can in one, pull the unique bits back into the callers for the rest.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1826,67 +1947,7 @@ public void updatePool(@PathParam("owner_key") @Verify(Owner.class) String owner
...
-            newPool.setAttributes(currentPool.getAttributes());
-        }
-
-        if (newPool.getBranding() == null) {
-            newPool.setBranding(currentPool.getBranding());
-        }
-
-        // Resolve the bits we may not have set
-        newPool = resolverUtil.resolvePool(newPool);
+        this.populateEntityOnUpdate(newPool, newPoolDTO, currentPool);
```
Ceiu: Same here; also not a fan of the three-way merge looking thing here. We're at a point where we don't need to juggle multiple instances of the entities to accomplish this task.

server/src/main/java/org/candlepin/resource/util/GuestMigration.java
```diff
@@ -66,6 +69,34 @@ public void migrate(boolean flush) {
...
+     * @return guestMigration object that contains all information related to the migration
+     */
+    public GuestMigration buildMigrationManifest(ConsumerDTO incoming, Consumer existing) {
+        if (incoming.getGuestIds() == null) {
+            log.debug("Guests not included in this consumer update, skipping update.");
+            migrationPending = false;
+            return this;
+        }
+
+        Consumer incomingConsumerEntity = new Consumer();
```
Ceiu: There is almost as much work done here just to convert the consumer as there would be if the DTO were used directly. This is the type of thing we want to avoid.

server/src/main/java/org/candlepin/service/SubscriptionServiceAdapter.java
```diff
@@ -83,6 +84,14 @@
      */
     boolean canActivateSubscription(Consumer consumer);
 
+    /**
+     * Can this consumer activate a subscription?
+     *
+     * @param consumer
+     * @return <code>true</code> if and only if this consumer can activate a subscription
+     */
+    boolean canActivateSubscription(ConsumerDTO consumer);
```
Ceiu: No API DTOs in the adapters

server/src/main/java/org/candlepin/resource/util/GuestMigration.java
nikosmoum: @Ceiu The original version of this method is called from quite a few other places using a Consumer (as opposed to ConsumerDTO) for a parameter, so both versions of a parameter need to be available. And this whole class, and its nested MigrationManifest class, are very coupled to Consumer and GuestId. The only alternative to this wrapper that translates to a Consumer would be to write a whole new class called something like 'GuestMigrationDTO' that is using ConsumerDTO and GuestIdDTO alongside it. Would that be an acceptable solution?

server/src/main/java/org/candlepin/resource/util/GuestMigration.java
```diff
@@ -86,15 +86,36 @@ public GuestMigration buildMigrationManifest(ConsumerDTO incoming, Consumer exis
             return this;
         }
 
-        Consumer incomingConsumerEntity = new Consumer();
-        List<GuestId> guestIds = new ArrayList<>();
-        for (GuestIdDTO dto : incoming.getGuestIds()) {
-            GuestId entity = new GuestId();
-            GuestIdResource.populateEntity(entity, dto);
-            guestIds.add(entity);
+        manifest = new MigrationManifest(existing);
```
Ceiu: This code looks correct enough, but something about this isn't sitting well with me, and I can't quite put my finger on what the issue is. I'll mention this as a post-scrum discussion on Monday and try to figure out where we take this class and the surrounding/provided functionality.

Ceiu:
Per our discussion today, we'll hold this PR until @mstead has a chance to look through at least the GuestMigration bits and we get his feedback.

nikosmoum:
@Ceiu @mstead I had to rebase this PR because of the Consumer environment ID changes. Should be good to review now.

mstead:
> Per our discussion today, we'll hold this PR until @mstead has a chance to look through at least the GuestMigration bits and we get his feedback.

@Ceiu @nikosmoum So my thoughts... First off, GuestId's equals should probably not be considering the attributes as **_from what I can tell_** we really only care about the actual String id match when determining if GuestIds were added and removed. If we make this change, we should do it in a separate commit in case we need to pull it for some reason. 

Second, there's really nothing in the GuestMigration class that offends me too much TBH. The one thing that sticks out though is that we could potentially be needlessly creating a bunch of GuestId objects in cases where GuestIds are reported that already exist on the correct host.  Perhaps there is a quick fix there by comparing string IDs and only creating the model objects when they are net new. I'm not really sure it is worth the effort of addressing this unless we determine that it is causing an issue.

Thoughts?

Ceiu:
I'm fine with it being done in a separate PR; as that's a refactor to GuestId and GuestMigration rather than being focused on DTO usage itself.

mstead:
> @Ceiu @nikosmoum So my thoughts... First off, GuestId's equals should probably not be considering the attributes as from what I can tell we really only care about the actual String id match when determining if GuestIds were added and removed. If we make this change, we should do it in a separate commit in case we need to pull it for some reason.

Actually, I could be wrong on this part. Need to verify something....

mstead:
> I'm fine with it being done in a separate PR; as that's a refactor to GuestId and GuestMigration rather than being focused on DTO usage itself.

Agreed. Lets merge this as it is.

PR #1942, opened 2018-03-20
Unify custom tasks under the LiquibaseCustomTask interfaces.

awood:
We can change the FixDuplicatePools class name without an issue since
the actual changelog XML only references the
FixDuplicatePoolsLiquibaseWrapper class name.

### DIFF
--- a/server/src/main/java/org/candlepin/liquibase/FixDuplicatePoolsTask.java
+++ b/server/src/main/java/org/candlepin/liquibase/FixDuplicatePoolsTask.java
@@ -14,10 +14,12 @@
  */
 package org.candlepin.liquibase;
 
-import liquibase.database.jvm.JdbcConnection;
+import liquibase.database.Database;
+import liquibase.exception.DatabaseException;
 
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


awood:

server/src/main/java/org/candlepin/liquibase/FixDuplicatePoolsTask.java
```diff
@@ -156,10 +149,11 @@ private int updateEntsForIds(List<String> dupeIds, String goodId) throws Excepti
             }
             else if (!subOwners.get(current).equals(rs.getString(4))) {
                 // Make sure owners are the same
-                log.error("Owners '{}' and '{}' both have pools from subscription: {}",
+                logger.error("Owners '{}' and '{}' both have pools from subscription: {}",
```
Ceiu: Does this work? I thought the liquibase logger didn't support this style of interpolation?

Ceiu:
Code looks good; just want to do a quick test to see if that log line actually functions or not. If it doesn't, this seems like a good opportunity to fix it.

server/src/main/java/org/candlepin/liquibase/FixDuplicatePoolsTask.java
Ceiu: Just tested it, it does not.

```
this.logger.info("THIS IS A {}", "test");
```

Outputs:

```
20150210094558-35::crog: THIS IS A {}
```

awood: I replaced it with `String.format`

Ceiu:
ack

PR #1943, opened 2018-03-22
1558114: Add total count of records as a header when paging is used.

barnabycourt:


Ceiu:
ACK

My only concern with this is that our current paging isn't stateful, nor are we setup to be in this context in the near future. This could end up being incorrect in certain circumstances. If this isn't a concern, this is good to merge.

barnabycourt:
Statefullness is not a concer. It is understood that the counts could change as paging is occuring.

Ceiu:
Alright, if it's been considered and acknowledged as a risk, I'm fine with it. Merging.

PR #1946, opened 2018-03-26
[M] Added pool ID filtering functionality to OwnerPoolResource.listPools

Ceiu:
- The listPools end point now supports filtering on pool IDs
- listPools now supports multiple matches filters


server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1478,6 +1478,9 @@ public int countConsumers(
         @ApiParam("Will only return pools with a start date after the supplied date. " +
                 "Overrides the activeOn date.")
         @QueryParam("after") @DateFormat Date after,
+        @ApiParam("One or more pool IDs to use to filter the output; only pools with IDs matching " +
+                "those provided will be returned; may be specified multiple times")
+        @QueryParam("poolids") List<String> poolIds,
```
mstead: Should this param be named in singular form as we do with others?

mstead: Well, looking at the other resources, we seem to do both. In that case, up to yourself.

PR #1948, opened 2018-03-27
Enable a bypass of resolving custom products when adding subscriptions to the test adapters

barnabycourt:
This enables the testing of the true hosted behavior where the entire
subscription definition comes in via the adapter. Prior to this change, product data is loaded from 
Candlepin via the resolver before being put into the HostedTestSubscriptionServiceAdapter in 
memory cache. This change enables setting things exactly as passed in the request JSON.


barnabycourt:
retest this please

barnabycourt:
@awood This is going to be used by a custom jmeter script that will be in a follow on PR to caracalla for simulating various hosted org refresh use cases. 

It could also be used by the spec tests if a decision is made to move away from the current model of loading products/content using custom products/content for hosted tests.

PR #1949, opened 2018-03-28
Manifest import/export now uses ProductDTO

nikosmoum:


server/src/main/java/org/candlepin/sync/EntitlementImporter.java
```diff
@@ -123,13 +126,13 @@ public Subscription importObject(ObjectMapper mapper, Reader reader, Owner owner
             }
         }
 
-        Product product = this.findProduct(productsById, entitlement.getPool().getProductId());
-        subscription.setProduct(product.toDTO());
+        ProductDTO productDTO = this.findProduct(productsById, entitlement.getPool().getProductId());
+        subscription.setProduct(productDTOtoProductData(productDTO));
```
Ceiu: Maybe we should use another shim translator here, so when we finally clean that stuff up, we don't have remnants of conversion code laying about. May not be worth the effort though. Thoughts?

nikosmoum: Yes, I think I'll move these methods in translators under the shim package. I didn't because we agreed not to use "reverse translators" that translate from DTO to entity, but since ProductData is also a DTO and not an entity it should be fine.

nikosmoum:
@Ceiu Introduced new shim translators for ProductDTO->ProductData.

Ceiu:
ack

PR #1951, opened 2018-03-29
1550190: Fixed Artemis resend on error to prevent lost messages

mstead:
Before this patch, resend message on error wasn't working
all the time, which could sometimes lead to lost messages
that wouldn't make it to Qpid.

This patch moves from having one ClientSession receiving
messages for all queues, to using a ClientSession per
queue/listener.

Prior to this patch, client sessions were set to
auto-acknowledge and auto-commit messages. This is no
longer the case. Each listener's ListenerWrapper now
manages its session by committing on success and rolling
back on failure.

Another important change is that as soon as a listener
recieves a message, that message should be immediately
acknowledged so that the session can keep track of the
number of times a message was sent (despite success or
failure). This is key to allowing retry on error to
function correctly.

### Testing Notes
All steps to test this PR can be found in the BZ:
    https://bugzilla.redhat.com/show_bug.cgi?id=1550190

With this patch applied, you should notice that retries will begin to back off by 10 seconds up to a max of the amount configured in candlepin conf. By default the max is 1 minute.

Once Qpid comes back online, messages will be received based on the next retry time.

### DIFF
--- a/server/src/main/java/org/candlepin/audit/EventReceiver.java
+++ b/server/src/main/java/org/candlepin/audit/EventReceiver.java
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) 2009 - 2016 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.audit;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.activemq.artemis.api.core.ActiveMQException;
+import org.apache.activemq.artemis.api.core.ActiveMQExceptionType;
+import org.apache.activemq.artemis.api.core.client.ClientConsumer;
+import org.apache.activemq.artemis.api.core.client.ClientSession;
+import org.apache.activemq.artemis.api.core.client.ClientSessionFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Receives messages from an Artemis Queue. Each receiver creates and handles its own session.
+ */
+public class EventReceiver {
+    private static Logger log = LoggerFactory.getLogger(EventReceiver.class);
+    private ClientSession session;
+
+    public EventReceiver(EventListener listener, ClientSessionFactory sessionFactory,
+        ObjectMapper mapper) throws ActiveMQException {
+        String queueName = EventSource.QUEUE_ADDRESS + "." + listener.getClass().getCanonicalName();
+        log.debug("registering listener for {}", queueName);
+
+        // The client session is created without auto-acking enabled. This means
+        // that the client handlers will have to manage the session themselves.
+        // The session management will be done by each individual ListenerWrapper.
+        //
+        // A message ack batch size of 0 is specified to prevent duplicate messages
+        // if the server goes down before the batch ack size is reached.
+        session = sessionFactory.createSession(false, false, 0);
+
+        try {
+            // Create a durable queue that will be persisted to disk:
+            session.createQueue(EventSource.QUEUE_ADDRESS, queueName, true);
+            log.debug("created new event queue: {}", queueName);
+        }
+        catch (ActiveMQException e) {
+            // if the queue exists already we already created it in a previous run,
+            // so that's fine.
+            if (e.getType() != ActiveMQExceptionType.QUEUE_EXISTS) {
+                throw e;
+            }
+        }
+
+        ClientConsumer consumer = session.createConsumer(queueName);
+        consumer.setMessageHandler(new ListenerWrapper(listener, mapper, session));
+        session.start();
+    }
+
+    /**
+     * Close the current session.
+     */
+    public void close() {
+        // Use a separate try/catch to ensure that both methods
+        // are at least tried.
+        try {
+            this.session.stop();
+        }
+        catch (ActiveMQException e) {
+            log.warn("Error stopping client session", e);
+        }
+
+        try {
+            this.session.close();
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/candlepin/audit/EventReceiver.java
```diff
@@ -0,0 +1,84 @@
...
+        // The session management will be done by each individual ListenerWrapper.
+        //
+        // A message ack batch size of 0 is specified to prevent duplicate messages
+        // if the server goes down before the batch ack size is reached.
+        session = sessionFactory.createSession(false, false, 0);
+
+        try {
+            // Create a durable queue that will be persisted to disk:
+            session.createQueue(EventSource.QUEUE_ADDRESS, queueName, true);
+            log.debug("created new event queue " + queueName);
```
awood:
```
log.debug("created new event queue {}", queueName);
```
please

server/src/main/java/org/candlepin/audit/EventReceiver.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.activemq.artemis.api.core.client.ClientConsumer;
+import org.apache.activemq.artemis.api.core.client.ClientSession;
+import org.apache.activemq.artemis.api.core.client.ClientSessionFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Receives messages from an Artemis Queue. Each receiver creates and handles its own session.
+ */
+public class EventReceiver {
```
awood: Should this class maybe implement `Closeable`?  It already implements a `void close()` so if we formally implement `Closeable` that would allow us to use it in a [try-with-resources statement](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html).

server/src/main/java/org/candlepin/audit/EventSource.java
```diff
@@ -67,39 +60,15 @@ protected ClientSessionFactory createSessionFactory() throws Exception {
...
+    }
 
-            ClientConsumer consumer = session.createConsumer(queueName);
-            consumer.setMessageHandler(new ListenerWrapper(listener, mapper));
-        }
-        catch (ActiveMQException e) {
-            log.error("Unable to register listener :" + listener, e);
-        }
+    private void closeEventReceivers() {
+        this.eventReceivers.forEach((receiver) -> { receiver.close(); });
```
awood: We can strip this down quite a bit to just

```java
this.eventReceivers.forEach(EventReceiver::close);
```

server/src/main/java/org/candlepin/audit/ListenerWrapper.java
```diff
@@ -14,60 +14,75 @@
...
-            msg.acknowledge();
-            log.debug("ActiveMQ message acknowledged for listener: " + listener);
+            // Finally commit the session so that the message is taken out of the queue.
+            session.commit();
         }
-        catch (ActiveMQException e) {
-            log.error("Unable to acknowledge ActiveMQ msg", e);
+        catch (Exception e) {
+            // Log a warning instead of a full stack trace to reduce log size.
+            String messageId = msg == null ? "" : Long.toString(msg.getMessageID());
```
awood: Can we put these ternary predicates in parenthesis?

```
String messageId = (msg == null) ? "" : Long.toString(msg.getMessageID());
```

server/src/main/java/org/candlepin/audit/ListenerWrapper.java
```diff
@@ -14,60 +14,75 @@
...
+            String reason = e.getCause() == null ? e.getMessage() : e.getCause().getMessage();
+            log.error("Unable to process message {}: {}", messageId, reason);
+
+            // If debugging is enabled log a more in depth message.
+            log.debug("Unable to process message. Rolling back client session.\n" + body, e);
```
awood:
```
log.debug("Unable to process message. Rolling back client session.\n{}", body, e);
```
will save the JVM from having to do the string concatenation unnecessarily.

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -106,6 +106,22 @@ private ConfigProperties() {
      */
     public static final String ACTIVEMQ_MAX_PAGE_SIZE = "candlepin.audit.hornetq.max_page_size";
 
+    /**
+     * For more on Artemis redelivery delays, see:
+     * https://activemq.apache.org/artemis/docs/2.4.0/undelivered-messages.html
+     */
+
+    // The number of milliseconds to wait before redelivering failed messages to the listeners.
```
awood: Let's mark these comments as javadoc `/**` so they'll show up in generated documentation.

server/src/main/java/org/candlepin/audit/EventReceiver.java
mstead: ???

mstead: :+1:

server/src/main/java/org/candlepin/audit/EventSource.java
mstead: :+1:

server/src/main/java/org/candlepin/audit/ListenerWrapper.java
mstead: :+1:

server/src/main/java/org/candlepin/audit/ListenerWrapper.java
mstead: hrm.... I don't think that the debug method supports that does it (i.e Throwable as the last filter arg)? This was moved from another class, and I thought that's why it was this way.

server/src/main/java/org/candlepin/config/ConfigProperties.java
mstead: :+1:

server/src/main/java/org/candlepin/audit/EventReceiver.java
mstead: After our chat, I won't be doing this since EventReceiver will not be used in a try block. Just making note.

server/src/main/java/org/candlepin/audit/ListenerWrapper.java
awood: Internally, Logback looks at the last argument of a call to `log.debug(String, Object...)` and if it's a Throwable, it does the [right thing](https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java#L141).

PR #1952, opened 2018-03-29
[M] 1562097: Renamed Event.consumerId to consumerUuid

Ceiu:
- Renamed Event.consumerId to consumerUuid, as the consumer ID isn't
  usable outside of Candlepin, and event consumers are expecting the UUID
- Renamed the cp_event.consumerid column to consumer_uuid in accordance
  with the field name change


wottop:
ack

PR #1954, opened 2018-04-03
[M] 1563003: Removed extraneous validation on default service level

Ceiu:
- Default service level is no longer validated during owner creation
  or update

### DIFF
--- a/server/src/main/java/org/candlepin/resource/OwnerResource.java
+++ b/server/src/main/java/org/candlepin/resource/OwnerResource.java
@@ -536,13 +536,8 @@ public class OwnerResource {
         }
 
         if (dto.getDefaultServiceLevel() != null) {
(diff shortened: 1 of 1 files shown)


Ceiu:
Seems reverting this change (re)introduces a bit of an inconsistency. Our tests indicate that we've been expecting this to fail on update, but either don't expect it or weren't testing it during creation. I want to have time to look into this more before we merge this and potentially make this situation worse.

vritant:
@Ceiu @cnsnyder adding myself as additional reviewer.
This fixes bug: 1565066 as well imo.

Ceiu:
Closing this PR as the "fix" here is not to change behavior, but to clean up the error message and add a new error for attempting to provide a default service level on owner creation.

A new PR will be posted with these fixes shortly.

Ceiu:
New PR: #1965

PR #1955, opened 2018-04-03
Factored the ansible candlepin roles into a dependency

Lorquas:
I pulled the candlepin roles out into a separate repo.  This simplifies the whole vagrant setup and it allows QE (and anybody else who wants to develop candlepin) to keep up with the latest ansible roles.

The repo is currently living here: https://github.com/Lorquas/ansible-role-candlepin
I would like to move this repo into the candlepin github project eventually.  Right now it is developer focused (installing candlepin with git) but it would be nice to make that role more flexible for different use cases in the future.

This PR currently works as is.  If we decide to move the canlepin role repo the requirements.yml will need to be changed.


Lorquas:
Something that is untested, is the yorkit bits of the role.  I left them untouched but don't know enough about it to test.  Any help would be appreciated.

Lorquas:
@barnabycourt  I think it definitely should be on the candlepin team page, but I didn't want to add a project all willy nilly without some sort of discussion.  I'm also unsure of the preferred way of setting up admins and groups for this team.

Lorquas:
Updated the repo location.

kahowell:
> @kahowell what is your take?

Looks great. @Lorquas thanks for this PR!

Merging

PR #1956, opened 2018-04-03
HornetQ: Disable monitoring for in-vm connections

kahowell:
Essentially, this makes our client `connection-ttl` and `client-failure-check-period` the same as if we were running on a newer version of HornetQ (see https://github.com/hornetq/hornetq/pull/1825).

This should help us to avoid getting into a bad state due to the behavior described in https://issues.jboss.org/browse/HORNETQ-1314.

See https://docs.jboss.org/hornetq/2.2.5.Final/user-manual/en/html/connection-ttl.html for details about what the parameters do.

### DIFF
--- a/server/src/main/java/org/candlepin/audit/HornetqEventDispatcher.java
+++ b/server/src/main/java/org/candlepin/audit/HornetqEventDispatcher.java
@@ -72,6 +72,8 @@ public class HornetqEventDispatcher  {
         ServerLocator locator = HornetQClient.createServerLocatorWithoutHA(
             new TransportConfiguration(InVMConnectorFactory.class.getName()));
         locator.setMinLargeMessageSize(largeMsgSize);
+        locator.setConnectionTTL(-1);
+        locator.setClientFailureCheckPeriod(-1);
         return locator.createSessionFactory();
     }
 


Ceiu:
I'm not sure I understand the connection between the referenced HornetQ change, the title and the changes made within CP. It looks like we're just shutting off the timeouts entirely, but I don't see how that changes between a virtualized and native environment. Could you provide more details as to how everything connects here?

kahowell:
I'm marking as in-progress until we determine whether this works as expected (some recent testing suggests no :disappointed: )

kahowell:
> I'm not sure I understand the connection between the referenced HornetQ change, the title and the changes made within CP. It looks like we're just shutting off the timeouts entirely, but I don't see how that  changes between a virtualized and native environment. Could you provide more details as to how everything connects here?

@Ceiu yeah, we've seen instances in some cases of `HQ212037: Connection failure has been detected: HQ119014: Did not receive data from invm:0` which then leads to an NPE when the associated `org.candlepin.audit.HornetqEventDispatcher` tries to send an event. I suspected we are *not* seeing this in newer candlepin installations because of changes in https://issues.jboss.org/browse/HORNETQ-1314 (newer candlepin uses newer HornetQ, which has HORNETQ-1314 merged) ...

HornetQ looks for dead connections, and frees them in attempt to keep a HornetQ server from leaking resources. In the case of an in-VM (i.e. in-JVM) connection (vs. TCP), the connection is a reliable connection, and thus does not benefit from the connection checking. My assumption is that if we disable the checks by setting the two parameters here (per the documentation), then Candlepin will not get into a state where the client times out and its connection is cleaned up (which again, is fine since it is an in-JVM connection). The method I used to set those two parameters needs some additional investigation, as I've found that setting these parameters on the server locator may not work the way I expected.

kahowell:
Closing this PR in favor of planned changes suggested in https://bugzilla.redhat.com/show_bug.cgi?id=1564087

PR #1957, opened 2018-04-04
[M] 1563436: Restore ConsumerTypeDTO constructor for legacy consumer JSON

Ceiu:
- Return the removed ConsumerTypeDTO constructor which accepts a
  label to address an issue where an exception is thrown when a
  consumer type is sent as a label
- Removed an additional DTO constructor that slipped through the
  previous removal pass


vritant:
Could you add a spec test to demonstrate the necessity of the constructor, please?

Ceiu:
@vritant Done.

PR #1960, opened 2018-04-05
[M] Added the new UserDTO for the API v1 DTO package

Ceiu:
- Added the new UserDTO
- Added the UserTranslator for translating User entities to UserDTOs

### DIFF
--- a/server/src/main/java/org/candlepin/dto/api/v1/UserDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/UserDTO.java
@@ -1,0 +1,252 @@
+/**
+ * Copyright (c) 2009 - 2016 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.dto.api.v1;
+
+import org.candlepin.dto.TimestampedCandlepinDTO;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+
+
+
+/**
+ * DTO representing the user data exposed to the API layer.
+ *
+ * <pre>
+ * {
+ *   "id": "string",
+ *   "username": "string",
+ *   "superAdmin": false,
+ *   "created": "2018-04-02T19:47:31.758Z",
+ *   "updated": "2018-04-02T19:47:31.758Z"
+ * }
+ * </pre>
+ */
+@ApiModel(parent = TimestampedCandlepinDTO.class, description = "User information for a given user")
+public class UserDTO extends TimestampedCandlepinDTO<UserDTO> {
+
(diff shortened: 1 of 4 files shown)


nikosmoum:
Is this in progress, or is there another task for applying this DTO to the Resource layer?

Ceiu:
Another task. Given the size of some of the actual usage/refactor tasks, and the way we've started breaking up the manifest DTO tasks, this is probably the safer, more consistent way to do it.

server/src/main/java/org/candlepin/dto/api/v1/UserDTO.java
```diff
@@ -0,0 +1,203 @@
...
+@ApiModel(parent = TimestampedCandlepinDTO.class, description = "User information for a given user")
+public class UserDTO extends TimestampedCandlepinDTO<UserDTO> {
+
+    @ApiModelProperty(required = true, example = "ff808081554a3e4101554a3e9033005d")
+    protected String id;
+    @ApiModelProperty(required = true, example = "admin")
+    protected String username;
+    @ApiModelProperty(required = false, example = "true")
+    protected Boolean superAdmin;
+
```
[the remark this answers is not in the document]
Ceiu: Kicking out the hashed password is a security risk we don't need to be doing. In the unlikely event someone actually uses that data, we can discuss returning it again; but under no (normal) circumstance should this type of thing be built into the core APIs and be kicked out to users.

However, the lack of an input password is definitely an oversight and something that needs to be present.

server/src/main/java/org/candlepin/dto/api/v1/UserDTO.java
```diff
@@ -0,0 +1,249 @@
...
+     */
+    public UserDTO setPassword(String password) {
+        // Impl note:
+        // While this would be a horrid practice in any other language, in Java, there's not a whole
+        // lot we can do about this. As soon as the plain-text password is in the JVM, we have no
+        // way of guaranteeing a (shadow) copy of that variable won't be stored in the garbage
+        // collector or whatever goofy optimization mapping Java is doing under the hood.
+        // As such, what is typically a very unsafe practice in most languages is par for the course
+        // here in Java land. This is really not a great deal worse than instantly hashing it as we
+        // do in the User entity given how it was used.
```
nikosmoum: Is it not possible to use a `char[]` as the password type, since an array can be overwritten when the password is no longer needed, and it won't stay in the string pool? (obviously not something for this PR, just a general thought)

server/src/main/java/org/candlepin/dto/api/v1/UserDTO.java
```diff
@@ -0,0 +1,249 @@
...
+     * Sets or clears the plain-text password to use for this user.
+     *
+     * @param password
+     *  The password, in plain text, to set for this user, or null to clear any previously set
+     *  password
+     *
+     * @return
+     *  a reference to this DTO
+     */
+    public UserDTO setPassword(String password) {
```
nikosmoum: I believe you need to add a `@JsonProperty` on this setter explicitly, because the `@JsonIgnore` annotation on the getter marks this whole field (field, setter, getter) as ignored.

server/src/main/java/org/candlepin/dto/api/v1/UserDTO.java
```diff
@@ -0,0 +1,249 @@
...
+        this.username = username;
+        return this;
+    }
+
+    /**
+     * Fetches the password set for this user. If the password hasn't been set, this method returns
+     * null.
+     * <p></p>
+     * <strong>
+     *  Note that for obvious reasons, this be serialized for output under any sane circumstances.
```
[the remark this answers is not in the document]
Ceiu: Whoops. Fixed.

server/src/main/java/org/candlepin/dto/api/v1/UserDTO.java
Ceiu: No, because the risk here isn't that we can't clear the string when we're done with it (though it helps minimize this potential for a memory-based attack), it's that if the GC runs while the char array is being maintained with a valid password, depending on the GC implementation, you could end up with another copy of the array in memory as the GC shuffles things around.

Either way, we're kinda stuck with a string here due to the interface we're using.

server/src/main/java/org/candlepin/dto/api/v1/UserDTO.java
Ceiu: Fixed

server/src/main/java/org/candlepin/dto/api/v1/UserDTO.java
nikosmoum: I see

PR #1961, opened 2018-04-05
[M] Made several corrections and standardizations to the DTOs

Ceiu:
- Updated the clone method for several DTOs to use getters and setters
  for assigning values to the cloned entities for future compatibility
  with DTO subclasses
- Standardized the naming convention of the identity certificate field
- Removed extraneous clone methods which don't perform any operations
- Updated some toString methods to use the same capitalization and
  spacing with field labels


Ceiu:
Note for reviewers: Jenkins hasn't failed here. The tests actually completed successfully, but just failed to update github on the status.

PR #1963, opened 2018-04-09
[M] Implemented DistributorVersionDTO for manifest import/export.

nikosmoum:
- DistributorVersionDTO is used instead of DistributorVersion when importing/exporting a manifest.

### DIFF
--- a/server/src/main/java/org/candlepin/dto/manifest/v1/DistributorVersionDTO.java
+++ b/server/src/main/java/org/candlepin/dto/manifest/v1/DistributorVersionDTO.java
@@ -1,0 +1,347 @@
+/**
+ * Copyright (c) 2009 - 2018 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/candlepin/dto/manifest/v1/DistributorVersionDTO.java
```diff
@@ -0,0 +1,340 @@
...
+@XmlAccessorType(XmlAccessType.PROPERTY)
+public class DistributorVersionDTO extends TimestampedCandlepinDTO<DistributorVersionDTO> {
+    public static final long serialVersionUID = 1L;
+
+    /**
+     * Internal DTO object for DistributorVersionCapability
+     */
+    @XmlRootElement(name = "distributorversioncapability")
+    @XmlAccessorType(XmlAccessType.PROPERTY)
+    public static class DistributorVersionCapabilityDTO {
```
Ceiu: These should still be CP DTOs for generic processing handiness

server/src/main/java/org/candlepin/dto/manifest/v1/DistributorVersionDTO.java
```diff
@@ -0,0 +1,340 @@
...
+
+        public String getId() {
+            return this.id;
+        }
+
+        public void setId(String id) {
+            this.id = id;
+        }
+
+        public String getName() {
```
Ceiu: This isn't a deal breaker, but there's no mutator for name. The only thing that must be immutable on these type of inner-DTOs are object references. If there's a reason to not have it, that's fine; but otherwise we should try to adhere to the DTO design requirements as much as possible.

nikosmoum:
@Ceiu the inner DTO now extends CandlepinDTO, and has a mutator for the name field.

PR #1964, opened 2018-04-09
[m] Lock less binds phase 2

vritant:
## Description
Today, all bind time pool operations are performed via a bind chain that passes around a shared bind context. The preProcess method performs most of the heavy weight operations, but does not perform any database operations untill the execute method is executed after the locks are placed.
In this PR, I have:
* created a DelayedPoolOp to capture the operations to perform once the locks are placed.
* split PostBindBonusPoolsOp into PostBindBonusPoolsOp and CheckBonusPoolsQuantitiesOp.
* Currently only PostBindBonusPoolsOp is lock-less in it's preProcessing. A following PR would tackle CheckBonusPoolsQuantitiesOp.
* PostBindBonusPoolsOp temporarily assigns the pools and entitlements in its preProcess step so pools and entitlements can be associated to create stack derived pools and keep the stack updated.

### Test setup
* create an owner
* create a single pool with attributes virt-limit, stack_id and multi-entitlement.
* make sure quantity of pool > 1000
* create 100 hypervisor consumers
* using caracalla/bind.jmx , in 100 threads bind each consumer against the same pool. repeat this in a loop 10 times.
* The first bind should create stacked derived pools for each consumer, and following 9 binds should update the stack.

### Result
| Branch | Average  | Min | Max | Std Dev. | Error % |
| ------------- | ------------- |------------- |------------- |------------- |------------- |
| lock-less-2 | 5773  | 313 | 28667 | 4875.64 | 0.007 |
| lock-less-2 | 5924  | 384 | 33304 | 5032.88 | 0.008 |
| Master | 7196  | 282 | 32198 | 5255.77 | 0 |
| Master | 7941  | 318 | 34714 | 5656.93 | 0 |

### Errors observed in performance tests
w.r.t the 0.008% error observed in this branch, upon closer inspection of the logs, it was observed that this happens because the performance test was written in a way that we could have multiple requests from the same consumer trying to bind against the same pool. when that happens, we can get a ConstraintViolationException because the two requests have completed their pre-process step together, before a lock was placed on the consumer. This has been a known and acceptable risk of lock-less implementation, and the previously agreed upon solution was to fail all but one concurrent requests so the clients can try again later. Added an exception handling block in BindChain to do the same.

### DIFF
--- a/server/src/main/java/org/candlepin/bind/PostBindBonusPoolsOp.java
+++ b/server/src/main/java/org/candlepin/bind/PostBindBonusPoolsOp.java
@@ -16,12 +16,25 @@ package org.candlepin.bind;
 
 import org.candlepin.controller.PoolManager;
 import org.candlepin.model.Consumer;
+import org.candlepin.model.ConsumerType;
+import org.candlepin.model.ConsumerType.ConsumerTypeEnum;
+import org.candlepin.model.ConsumerTypeCurator;
 import org.candlepin.model.Entitlement;
+import org.candlepin.model.Pool;
+import org.candlepin.model.PoolCurator;
 import org.candlepin.model.PoolQuantity;
+import org.candlepin.policy.js.entitlement.Enforcer;
 
 import com.google.inject.Inject;
 
+import org.apache.commons.collections.CollectionUtils;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
 
 /**
  * This BindOperation will be replaced shortly in an upcoming PR, where we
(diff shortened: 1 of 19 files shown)


vritant:

vritant:
Currently looking into why that failure is introduced in the performace tests, but please begin the review.

vritant:
The failure issue is now resolved, noted the resolution in the PR description above.

server/src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java
```diff
@@ -638,7 +599,8 @@ private void postUnbindVirtLimit(PoolManager poolManager, Entitlement entitlemen
                     for (int idex = 0; idex < pools.size(); idex++) {
                         Pool derivedPool = pools.get(idex);
                         if (derivedPool.getAttributeValue(Pool.Attributes.DERIVED_POOL) != null) {
-                            poolManager.updatePoolQuantity(derivedPool, virtQuantity);
+                            poolManager.setPoolQuantity(derivedPool, derivedPool.adjustQuantity
```
Ceiu: formatting gripe: can we move the whole "derivedPool.adjustQuantity" down so the argument list stays with at least the method name?

vritant:
@Ceiu  will accomodate these notes after @wottop  has turned his feedback in  too.

barnabycourt:
A bunch of questions about the performance resutlts:

1. For the performance changes, what was the mean time.
1. what was the percentage improvement? 
1. For customers that use VDC, what is the mean / max number of hypervisors in a VDC pool?
1. Were there any changes in the baseline performance? For example, f you do not have > 1000 hypervisors. What is the performance if 10 hypervisors?

barnabycourt:
@vritant Can the source data / caracalla tests be massaged or added to in such a way that we do not run into the error case?

server/src/main/java/org/candlepin/bind/DelayedPoolOp.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+/**
+ * Holds and represents operations to be performed at a later time.
+ * Currently, only holds relevant operations for PostBindBonusPoolsOp
+ */
+public class DelayedPoolOp {
```
awood: I'd prefer a different name here.  This class doesn't implement `BindOperation` so ending it with `Op` is kind of confusion.  Could I suggest `BindCallback` or `PoolCallback` or something like that?

server/src/main/java/org/candlepin/bind/PostBindBonusPoolsOp.java
```diff
@@ -30,27 +43,81 @@
...
+                    subPoolsForStackIds,
+                    entitlements.values());
+            }
+        }
+        else {
+            subPoolsForStackIds = new ArrayList<>();
+        }
+
+        delayedPoolOp.appendDelayedOp(
+            enforcer.postEntitlement(poolManager,
```
awood: The indentation is kind of off here.

```java
delayedPoolOp.appendDelayedOp(enforcer.postEntitlement(
    poolManager,
    consumer,
    context.getOwner(),
    entitlements,
    subPoolsForStackIds,
    false,
    poolQuantities
));
```
would be more clear

server/src/main/java/org/candlepin/policy/js/pool/PoolHelper.java
```diff
@@ -143,18 +149,18 @@ private PoolHelper() {
...
+                        sourceEntitlements.get(pool.getId()));
                 }
             }
-            poolsToCreate.add(consumerSpecificPool);
+            delayedPoolOp.addPoolToCreate(consumerSpecificPool);
         }
 
         if (CollectionUtils.isNotEmpty(poolsToUpdateFromStack)) {
-            poolManager.updatePoolsFromStack(consumer, poolsToUpdateFromStack);
+            poolManager.updatePoolsFromStackWithoutDeletingStack(consumer, poolsToUpdateFromStack, null);
```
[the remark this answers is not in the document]
vritant: @wottop  because this method invocation intentionally does not delete stack derived pools, and hence there is no need for a delayed execution. I renamed the method to WithoutDeletingStack  to convey that information, but perhaps there is a better name needed If I got this question :)
Any suggestion?

server/src/main/java/org/candlepin/bind/DelayedPoolOp.java
vritant: @awood we do have a precedence in RevocationOp that does end with Op and does not implement BindOperation. Going forward I can see us having similar Op classes for refresh code that will not implement BindOperation, but might implement something like RefreshOp ....

that being said, if you hate the name I am not married to it.. any objection to PoolOperationCallback ?

awood: PoolOperationCallback works for me!

vritant:

barnabycourt:
given the current discussions about performance the increasing of the minimum time is challenging. Is the change in minimum time across the board or only in specific cases? Can we articulate the cases in which the minimum time has gone up, how much it has gone up, and the expected change overall throughput that will be observed under high load?

vritant:
@barnabycourt  wrote some ruby scripts to measure the performance for the simple cases.

To measure performance, created a hundred hypervisors that bind against single pool sequentially. each of the bind should create a stack derived pool for the consumer.
Then, the same consumers bind against the same pool again, thus updating the stack by adding another entitlement to it.
Five such samples were drawn from each branch, redeploying candlepin with a fresh database and a tomcat restart separated each sample.

| Branch |  Test case | Average  | Median | Range | Min | Max |
| ------------- | ------------- |------------- |------------- |------------- |------------- |------------- |
| master | stack creation | 15.4347234 | 15.715739 | 1.349932 | 14.628198 | 15.978130 |
| lock-less-2 | stack creation | 14.7419556 | 14.360263 | 3.086409 | 13.298753 | 16.385162 |
| master | stack update | 16.0100528 | 15.648755 | 2.401198 | 14.937440 | 17.338638 |
| lock-less-2 | stack update | 15.7537312 | 15.509354 | 2.139505 | 15.088833 | 17.228338 |

barnabycourt:
@vritant This is good information. @kahowell what is your take? Should we merge this as is or should we hold off & get a before/after test run from IT-PNT using scratch builds after the 2.3/2.4 issues have been resolved?

kahowell:
> @vritant This is good information. @kahowell what is your take? Should we merge this as is or should we hold off & get a before/after test run from IT-PNT using scratch builds after the 2.3/2.4 issues have been resolved?

I'd like to run it through IT-PnT's tests just to see what the effects on throughput are.

kahowell:
> I'd like to run it through IT-PnT's tests just to see what the effects on throughput are.

We changed our minds, getting a build that will run in today's test setup is non-trivial, and this change is low-to-no risk to throughput. Therefore, merging :smile:

PR #1965, opened 2018-04-09
[M] Improved error output surrounding owner creation and service levels

Ceiu:
- Fixed a bug that caused the errors that occurred during owner
  creation to be lacking the owner key in the error message
- OwnerResource.createOwner now checks that the owner key is set
  before attempting to create the owner
- OwnerResource.createOwner now checks that the default service
  level is not set before attempting to create the owner


Ceiu:
This PR is the changes that came from discussion and investigation of the issue mentioned in BZ 1563003. Since we are not changing the behavior back to the old, incorrect behavior, this commit does not make a reference to the BZ in question.

PR #1967, opened 2018-04-11
Insert owner level logging filter only when needed.

awood:
The LoggerAndMDCFilter is responsible for logging messages at a given
level for specific owners.  We use this tool in production to get DEBUG
level messages for specific owners without having to log everyone else's
requests.  The filter is fast: the decide() method takes 0.1ms or less,
but it is still being run for *every* logging call.  Those tenths of a
millisecond add up over millions of invocations and all for a feature
that is used only rarely.

This patch inserts the LoggerAndMDCFilter when owner level logging is
requested and removes it when owner level logging is disabled.  It also
checks at servlet start-up time to see if owner level logging is
enabled anywhere and adds the filter if so.

----
Notes for testing:

Enable owner-level logging like so:
```
curl -k -v -u admin:admin  -X PUT "https://localhost:8443/candlepin/owners/snowwhite/log?level=DEBUG" 
```

Disable like so:
```
curl -k -v -u admin:admin  -X DELETE https://localhost:8443/candlepin/owners/snowwhite/log   
```

The following should be true:
* If any owner has logging enabled at servlet start up, the filter should be inserted.
* If owner logging is requested for an owner, the filter should only be inserted if it is not already present in the filter list
* If owner logging is disabled for an owner, the filter should be removed if and only if there are no other owners using owner logging.

I tested
* Enable logging for `admin`.  A message stating the filter has been added should appear.
* Enable logging for `snowwhite`.  No message should appear since the filter was already inserted.
* Disable logging for `snowwhite`. No message should appear since `admin` is still using owner logging.
* Disable logging for `admin`. A message should appear since no one else is using owner logging.
* Disable logging for `admin` again. No message should appear since the filter is not present in the filter list at all.
* Enable logging for `admin` and restart the server.  A message should appear since owner logging is still enabled for `admin`.

### DIFF
--- a/server/src/main/java/org/candlepin/model/OwnerCurator.java
+++ b/server/src/main/java/org/candlepin/model/OwnerCurator.java
@@ -78,6 +78,13 @@ public class OwnerCurator extends AbstractHibernateCurator<Owner> {
         return owner;
     }
 
+    public boolean ownerLoggingEnabled() {
+        Long ownersWithLogging = (Long) getEntityManager()
+            .createQuery("SELECT COUNT(o) FROM Owner o WHERE logLevel IS NOT NULL")
+            .getSingleResult();
+        return ownersWithLogging > 0;
+    }
+
     @Transactional
     public Owner replicate(Owner owner) {
         this.currentSession().replicate(owner, ReplicationMode.EXCEPTION);
--- a/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
+++ b/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
@@ -189,6 +189,7 @@ public class CandlepinContextListener extends GuiceResteasyBootstrapServletConte
         pinsetterListener.contextInitialized();
 
         loggerListener = injector.getInstance(LoggerContextListener.class);
+        loggerListener.contextInitialized();
 
         /**
          * Custom ModelConverter to handle our specific serialization requirements
--- a/server/src/main/java/org/candlepin/logging/LoggerAndMDCFilter.java
+++ b/server/src/main/java/org/candlepin/logging/LoggerAndMDCFilter.java
@@ -14,14 +14,21 @@
  */
 package org.candlepin.logging;
 
+import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 import org.slf4j.Marker;
 
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.classic.spi.TurboFilterList;
 import ch.qos.logback.classic.turbo.MatchingFilter;
+import ch.qos.logback.classic.turbo.TurboFilter;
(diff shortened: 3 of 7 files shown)


Ceiu:
Manual test looks good -- behavior works as expected. Just need to address the manual owner update endpoints.

server/src/main/java/org/candlepin/model/OwnerCurator.java
```diff
@@ -78,6 +78,13 @@ public Owner findOwnerById(String ownerId) {
         return owner;
     }
 
+    public boolean ownerLoggingEnabled() {
+        Long ownersWithLogging = (Long) getEntityManager()
+            .createQuery("SELECT COUNT(o) FROM Owner o WHERE logLevel IS NOT NULL")
```
barnabycourt: This is likely to fall over in production given the number of owners that are going to be scanned.

awood: I think it should still be pretty darn fast:

```
MariaDB [candlepin]> select count(*) from cp_owner where loglevel is not null;
+----------+
| count(*) |
+----------+
|        5 |
+----------+
1 row in set (0.29 sec)

MariaDB [candlepin]> select count(*) from cp_owner;
+----------+
| count(*) |
+----------+
|  1188919 |
+----------+
1 row in set (0.23 sec)
```

That's from a production dump from August.  So scanning 1.2 million owners takes about a quarter of a second which they should make back up within probably under an hour of not having the filter running.

barnabycourt: fair enough

barnabycourt:
Al, Does this have a scenario where the filter could be enabled on one node but not others? Specifically I am thinking of hosted where there are many nodes. The scenario I am thinking of is this:

1. A cluster of 5 servers is set up. 
1. No Org level filters are enabled.
1. All nodes are brought online (no log filters are inserted)
1. One organization has logging enabled.

I believe this will result in the log filter being inserted on one of the nodes but not the others. Am I correct? If so, wouldn't some of the logging be missed if requests are load balanced to other nodes?

awood:
> Al, Does this have a scenario where the filter could be enabled on one node but not others? Specifically I am thinking of hosted where there are many nodes. The scenario I am thinking of is this:
> 
>     A cluster of 5 servers is set up.
>     No Org level filters are enabled.
>     All nodes are brought online (no log filters are inserted)
>     One organization has logging enabled.
> 
> I believe this will result in the log filter being inserted on one of the nodes but not the others. Am I correct? If so, wouldn't some of the logging be missed if requests are load balanced to other nodes?

Yes, you are correct.  I neglected to consider clustering.  I'm not seeing any good way around that problem since we don't have anything like JGroups or some other intra-cluster communication mechanism.

Well, I can either just close this PR entirely or we could use some config value that always inserts the org-level filter during servlet start-up.  That would allow single node instances to benefit from this feature.  But for single node instances the benefit probably doesn't outweigh the additional complexity.  Let me know what you think.

PR #1971, opened 2018-04-16
Improved standardization of curator method verbiage

Ceiu:
- Renamed the generic ID-based lookup "find" and "secureFind" to
  "get" and "secureGet" respectively
- Renamed methods using the "lookup" prefix to use a "get" prefix
- Refactored or removed a few methods which were made redundant
  either by this change or by other changes made over the years
- Added the new "getNaturalIdLoader" method to AbstractHibernateCurator
  to assist with fetching entities using their natural IDs

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -382,7 +382,7 @@ public class CandlepinPoolManager implements PoolManager {
 
         if (owner.getKey() != null) {
             String ownerKey = owner.getKey();
-            owner = ownerCurator.lookupByKey(owner.getKey());
+            owner = ownerCurator.getByKey(owner.getKey());
 
             if (owner == null) {
                 throw new IllegalStateException(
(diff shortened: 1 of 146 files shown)


Ceiu:
So far the initial work here is just to start to standardize the verbiage used for our methods. I opted for "get" over "find" as a prefix due to two reasons: (1) we have more methods prefixed this way than not and (2) ```get``` is used by Hibernate (which we tend to use) where ```find``` is the JPA equivalent. Should the latter distinction become important later, it'll be nice to be able to use a prefix to quickly denote that.

I'm still sifting through the curators to see if we can make use of Hibernate's ```load``` method to create uninitialized proxies to avoid loading entire objects, but the renaming and refactoring work can be reviewed while I continue that.

Ceiu:
After looking through literally every use of a method named ".get," not a single one of them looked to be good places for a Hibernate proxy. Either we immediately use multiple fields on the object, or existence isn't guaranteed -- both of which make for the exact scenario in which you *don't* use a proxy. That being the case, I'll be omitting it from this PR and will be keeping it in mind for future work.

PR #1974, opened 2018-04-20
re-factor ConsumerResourceTest

vritant:


Ceiu:
Soooo much better. Thank you.

PR #1975, opened 2018-04-24
[M] 1517978: Jobs now respect owner-specific logging level where appropriate

Ceiu:
- Jobs run within the context of a specific owner now respect and use
  owner/org log levels

### DIFF
--- a/server/src/main/java/org/candlepin/controller/ManifestManager.java
+++ b/server/src/main/java/org/candlepin/controller/ManifestManager.java
@@ -114,11 +114,12 @@ public class ManifestManager {
      *                      a new export of the target consumer.
      * @return the details of the async export job.
      */
-    public JobDetail generateManifestAsync(String consumerUuid, String ownerKey, String cdnLabel,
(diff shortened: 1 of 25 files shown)


Ceiu:
**Testing instructions**

Update candlepin.conf (or the autoconf bits, whatever works) such that the following will be true:
- The standard log level is "INFO"
- Candlepin is running in hosted mode
- Candlepin is using the test hosted mode adapters

```
log4j.logger.org.candlepin=INFO
candlepin.standalone=false
module.config.hosted.configuration.module=org.candlepin.hostedtest.AdapterOverrideModule
```

Deploy Candlepin with test data and the hosted test adapters:
```$ ./bin/deploy -gtH```

Run a refresh against the "admin" owner and inspect candlepin.log. You should see only warnings and info lines.
```$ ./client/ruby/cpc refresh_pools admin```

Set the owner's log level to "debug"
```$ ./client/ruby/cpc set_owner_log_level admin debug```

Re-run the refresh against the admin owner and inspect candlepin.log. You should now see warnings, info and debug lines.
```$ ./client/ruby/cpc refresh_pools admin```

PR #1978, opened 2018-04-27
[M] Added the RoleDTO and PermissionBlueprintDTO

Ceiu:
This is just for implementation the new DTOs; application is a separate task


wottop:
ack

PR #1979, opened 2018-04-30
Candlepin requires Tomcat 7

wottop:


vritant:
would appreciate some verification steps if you have any handy?

vritant:
removing my name from reviewers list since I got unblocked on other items

PR #1984, opened 2018-05-07
Use JSS for basic crypto functions

awood:
This is the first of several PRs that I'll open against the `jss` feature branch.  This PR separates out our crypto functionality by provider and then adds a JSS provider.  The application itself still uses BouncyCastle, but I've gone ahead and implemented the code to create a basic X.509 CRL (the more complex CRL streaming code will be in a later PR), write PEM encoded ASN.1, and create a basic X.509 certificate (there's a TODO for adding all our extensions).


awood:

awood:

server/src/main/java/org/candlepin/pki/PKIUtility.java
```diff
@@ -41,48 +40,54 @@
...
 import java.util.List;
 import java.util.Set;
 
+import javax.inject.Inject;
+
 /**
  * PKIUtility
  */
-public abstract class PKIUtility {
+public class PKIUtility {
```
Ceiu: What value are we getting out of this class if it's just passthrough to the implemented interface? Seems like we could drop this entirely, inject the PKIProviderUtility as needed and skip this class entirely.

server/src/main/java/org/candlepin/pki/impl/JSSPKIUtility.java
```diff
@@ -0,0 +1,337 @@
...
+            certInfo.set(X509CertInfo.SUBJECT, new CertificateSubjectName(new X500Name(dn)));
+            certInfo.set(X509CertInfo.KEY,
+                new CertificateX509Key(X509Key.parse(new DerValue(publicKeyEncoded))));
+            certInfo.set(X509CertInfo.ALGORITHM_ID,
+                new CertificateAlgorithmId(AlgorithmId.get(SIGNING_ALG_ID)));
+            certInfo.set(X509CertInfo.VERSION, new CertificateVersion(CertificateVersion.V3));
+
+            X509CertImpl certImpl = new X509CertImpl(certInfo);
+
+            // TODO extensions here
```
Ceiu: Is this needed now, or is this for to-be-determined extensions as need arrives in the future?

server/src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            //add the new path
+            final String[] newPaths = Arrays.copyOf(paths, paths.length + 1);
+            newPaths[newPaths.length - 1] = pathToAdd;
+            usrPathsField.set(null, newPaths);
+            for (String x : newPaths) {
+                log.info("New Path " + x);
+            }
+            log.info("Library path " + System.getProperty("java.library.path"));
```
Ceiu: I'm guessing a lot of this is legacy carry over? Should be using {} instead of concatenation.

This applies elsewhere throughout the PR, but I'm not going to point them all out individually.

server/src/main/java/org/candlepin/pki/impl/JSSPKIUtility.java
```diff
@@ -0,0 +1,354 @@
...
+            CertificateFactory cf = CertificateFactory.getInstance("X.509");
+            return (X509CRL) cf.generateCRL(new ByteArrayInputStream(crlImpl.getEncoded()));
+        }
+        catch (GeneralSecurityException | IOException | InvalidBERException e) {
+            throw new RuntimeException("Error creating CRL", e);
+        }
+    }
+
+    private byte[] getPemEncoded(byte[] der, String type) throws IOException {
+        try (ByteArrayOutputStream out = new ByteArrayOutputStream()) {
```
Ceiu: OT comment:
Does our linter handle this very well? I can see this being confused for an inline assignment.

server/src/main/java/org/candlepin/resource/CrlResource.java
```diff
@@ -84,7 +84,7 @@ public Response getCurrentCrl(@Context Principal principal) throws CRLException
             // Create an empty CRL if we didn't have anything to write
             if (!crlFile.exists() || crlFile.length() < 1) {
                 pkiUtility.writePemEncoded(
-                    pkiUtility.createX509CRL(new LinkedList<>(), BigInteger.ZERO),
+                    pkiUtility.createX509CRL(new LinkedList<>(), BigInteger.ONE),
```
Ceiu: Did we have an off-by-one issue here?

awood: No, there was another spot in the code where we were calling `createX509` but with BigInteger.ONE.  This just makes it consistent across the codebase.

server/src/main/java/org/candlepin/pki/PKIUtility.java
awood: PKIUtility implements crypto methods using the JCA interfaces (i.e. stuff that comes with the JDK like verifying digital signatures), but there's other stuff we need to do that requires us to dip into the classes of the crypto provider.  PKIUtility is meant to be the one-stop-shop and uses composition to access the provider specific implementations (PKIProviderUtility).  The goal is to give us the flexibility in the future to switch providers again if we need to.

I will agree that the architecture/naming is sub-optimal.  For this first pass through I was mainly focusing on getting JSS to work without destroying all the BouncyCastle stuff as well.  Let me know if you have some ideas about organization/architecture/etc.

server/src/main/java/org/candlepin/pki/impl/JSSPKIUtility.java
awood: It's for the next sprint basically.

server/src/main/java/org/candlepin/pki/impl/JSSPKIUtility.java
awood: It doesn't complain at least.  We have the option of doing

```java
try (
  OutputStream x = new ByteArrayOutputStream();
) {
```

I can change it to that if you want.

Ceiu: Not at all -- I don't mind the new try-with-resources syntax, I just expected our linter to be a little flaky with a "new" syntax that looks similar to old disallowed ones.

server/src/main/java/org/candlepin/pki/SubjectKeyIdentifierWriter.java
```diff
@@ -16,26 +16,54 @@
...
+     * string.
+     * </p><p>
+     * Returning the encoded KeyIdentifier instead of just the KeyIdentifier is arguably dumb, but that's
+     * how the API contract was written (and it made sense with the way BouncyCastle created extensions) so
+     * we need to abide by it.
+     * </p>
+     * @param clientKeyPair the KeyPair of the certificate's subject
+     * @param extensions Other Extensions being placed on this certificate
+     * @return DER encoded octet string of the subject key identifier as a byte array
+     * @throws IOException thrown if error reading extensions or encoding the KeyIdentifier
```
Ceiu: :+1: Hooray for docs!

Ceiu:
ack

awood:

awood:

Ceiu:
Should this PR be squashed before merging?

PR #1985, opened 2018-05-08
Make various components singletons

kahowell:
In order to reduce memory overhead by not needlessly creating duplicate objects.

### DIFF
--- a/server/src/main/java/org/candlepin/guice/I18nProvider.java
+++ b/server/src/main/java/org/candlepin/guice/I18nProvider.java
@@ -17,7 +17,6 @@ package org.candlepin.guice;
 import org.candlepin.common.guice.CommonI18nProvider;
 
 import com.google.inject.Inject;
-import com.google.inject.Injector;
 import com.google.inject.Provider;
 import com.google.inject.ProvisionException;
 
@@ -30,24 +29,29 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import javax.inject.Singleton;
 import javax.servlet.http.HttpServletRequest;
 
 /**
  * I18nProvider
  */
+@Singleton
 public class I18nProvider extends CommonI18nProvider implements Provider<I18n> {
     private static Logger log = LoggerFactory.getLogger(I18nProvider.class);
     private static Map<Locale, I18n> cache = new ConcurrentHashMap<>();
-
-    private I18n i18n;
+    private final Provider<HttpServletRequest> request;
 
     @Inject
-    public I18nProvider(Injector injector) {
-        HttpServletRequest request = null;
+    public I18nProvider(Provider<HttpServletRequest> request) {
+        this.request = request;
+    }
+
+    @Override
+    public I18n get() {
         Locale locale = null;
 
         try {
-            request = injector.getInstance(HttpServletRequest.class);
+            locale = request.get().getLocale();
         }
         catch (ProvisionException e) {
             // This can happen in pinsetter, or anything else not in an http
@@ -55,34 +59,28 @@ public class I18nProvider extends CommonI18nProvider implements Provider<I18n> {
             // just ignore it.
         }
 
-        if (request != null) {
-            locale = request.getLocale();
-        }
-
         locale = (locale == null) ? Locale.US : locale;
 
         // If the locale does not exist, xnap is pretty inefficient.
         // This cache will hold the records more efficiently.
-        //
-        // Make sure to keep the access wrapped in synchronized so we can
-        // share across threads!
-        synchronized (cache) {
-            i18n = cache.get(locale);
-            if (i18n == null) {
-                log.debug("Getting i18n engine for locale {}", locale);
 
-                i18n = I18nFactory.getI18n(getClass(), getBaseName(), locale, I18nFactory.FALLBACK);
-                cache.put(locale, i18n);
+        // see https://en.wikipedia.org/wiki/Double-checked_locking
+        I18n i18n = cache.get(locale);
+        if (i18n == null) {
+            synchronized (cache) {
+                i18n = cache.get(locale);
+                if (i18n == null) {
+                    log.debug("Getting i18n engine for locale {}", locale);
+
+                    i18n = I18nFactory.getI18n(getClass(), getBaseName(), locale, I18nFactory.FALLBACK);
+                    cache.put(locale, i18n);
+                }
             }
         }
-    }
-
-    @Override
-    public I18n get() {
         return i18n;
     }
 
     public String getTestString() {
-        return i18n.tr("Bad Request");
+        return get().tr("Bad Request");
     }
 }
--- a/server/src/main/java/org/candlepin/model/HostCache.java
+++ b/server/src/main/java/org/candlepin/model/HostCache.java
@@ -1,0 +1,31 @@
+/**
+ * Copyright (c) 2009 - 2018 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.model;
+
+import org.candlepin.guice.CandlepinRequestScoped;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * Cache for host consumers (see https://github.com/candlepin/candlepin/pull/1334 for original rationale)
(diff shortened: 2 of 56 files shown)


server/src/main/java/org/candlepin/guice/I18nProvider.java
```diff
@@ -35,54 +31,52 @@
...
-        // Make sure to keep the access wrapped in synchronized so we can
-        // share across threads!
-        synchronized (cache) {
-            i18n = cache.get(locale);
-            if (i18n == null) {
-                log.debug("Getting i18n engine for locale {}", locale);
 
-                i18n = I18nFactory.getI18n(getClass(), getBaseName(), locale, I18nFactory.FALLBACK);
-                cache.put(locale, i18n);
+        // see https://en.wikipedia.org/wiki/Double-checked_locking
```
awood: I read that Wikipedia page and it seems like there are a lot of subtle things that can go awry with double-check locking in [Java](https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java).  Specifically the 

```
1.  Thread A notices that the value is not initialized, so it obtains the lock
    and begins to initialize the value.
2.  Due to the semantics of some programming languages, the code generated by the compiler is 
    allowed to update the shared variable to point to a partially constructed object before A has
    finished performing the initialization. For example, in Java if a call to a constructor has been
    inlined then the shared variable may immediately be updated once the storage has been
    allocated but before the inlined constructor initializes the object.[6]
3.  Thread B notices that the shared variable has been initialized (or so it appears), and returns
    its value. Because thread B believes the value is already initialized, it does not acquire the
    lock. If B uses the object before all of the initialization done by A is seen by B (either because
    A has not finished initializing it or because some of the initialized values in the object have
    not yet percolated to the memory B uses (cache coherence)), the program will likely crash.
```

This scenario may not apply to use since we are storing `i18n` in the `cache` map rather than in a field (thus sidestepping step 2), but I just wanted to make sure it's something we've considered.

See also http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html and the "Best Practices for Lazy Initialization" section of http://www.oracle.com/technetwork/articles/javase/bloch-effective-08-qa-140880.html

Personally, I like this part:

**Q**: What's most important to understand about best practices for lazy initialization?

**A**: The single most important piece of advice is "Don't do it unless you need to." 

Why do we need to lazy initialize this cache at all?  We know what locales we support.  Why not populate the map in a static block and be done with it?

Or we could just not do it ourselves and use https://github.com/google/guava/wiki/CachesExplained with a CacheLoader.

If we don't do any of that, I think we should be using an immutable map from Guava or an unmodifiable map from Collections and either map should be marked volatile instead of the ConcurrentHashMap.  See https://stackoverflow.com/a/23807722/6124862

Ceiu: Due to the operations we do inside the synch block, I think we're good on points 2 and 3 -- we don't have an object to insert into the map until the constructor returns, so both cases of partial initialization are avoided here.

However, that aside, the point you made about static init rather than JIT is a valid one that comes down to a question of which locals we support today and what the memory cost of loading all of them in straight away is in the general case (probably a wash in IT, potentially significant in Sat).

barnabycourt: @Ceiu Our translation bundles are currently ~2MB on disk. Static init for 2 MB does not seem like too much in the grand scheme of things when our overall memory footprint is 1GB+ in production systems.

kahowell:
db3945107..5b60cbc9a: Fix checkstyle issues and unit tests.

server/src/main/java/org/candlepin/guice/I18nProvider.java
kahowell: @awood @Ceiu @barnabycourt I agree that we should initialize all locales at once during application initialization, but I'd like to tackle that in a separate PR later on, just to keep this PR scoped to just scope changes (pun intended).

kahowell:
> Could we use javax.inject.Singleton instead?

@awood, sure. I'll update this PR shortly with that change.

kahowell:
I saw some troubling failures in spec tests. Will investigate more tomorrow.

server/src/main/java/org/candlepin/model/HostCache.java
```diff
@@ -0,0 +1,26 @@
...
+package org.candlepin.model;
+
+import org.candlepin.guice.CandlepinRequestScoped;
+
+import java.util.HashMap;
+
+/**
+ * Cache for host consumers (see https://github.com/candlepin/candlepin/pull/1334 for original rationale)
+ */
+@CandlepinRequestScoped
```
Ceiu: What happens when this cache is used from outside a request scope (i.e. from within the jobs)? Do we get an exception for being out of scope, or do we get a new instance on every call? We could probably work around this by keeping it session scoped rather than request scope in either case.

While I'm thinking on it, this is also an opportunity to make a proper type-safe minicache object that acts like a cache instead of a store. That way we can reduce our memory load in the off chance we flood this cache with hosts in a memory-heavy operation. If not this PR, I feel we should at least throw a task on the backlog to do so.

kahowell: > What happens when this cache is used from outside a request scope (i.e. from within the jobs)?

Javadoc for CandlepinRequestScope says "works even in our Quartz jobs".

> While I'm thinking on it, this is also an opportunity to make a proper type-safe minicache object that acts like a cache instead of a store. That way we can reduce our memory load in the off chance we flood this cache with hosts in a memory-heavy operation. If not this PR, I feel we should at least throw a task on the backlog to do so.

We can certainly look at using the caching mechanisms we introduced in #1336 , but I'd be against building anything too complicated. Created a card to discuss further.

kahowell:
> we'll only ever leverage the cache for very specific loops using the same set of owners as input.

Unfortunately, this is necessary (discussed later).

> While the pair is safer and isolates us from the migration case, I feel we should only be using just one owner ID in the pair. This would mean we'd need to check every possible pairing of guest ID and potential owner ID for the cache, and then generate a new pair from the consumer's actual owner.

I spent some time looking through having the cache be against a single owner at a time, and I realized unfortunately, this would violate the contract that the cache should fulfill, due to subtleties with the use of the `updated` field: see
https://github.com/candlepin/candlepin/blob/021ab0f4f6de957eb802742df1abf856c34e76e3/server/src/main/java/org/candlepin/model/ConsumerCurator.java#L538-L544

Considering what would be necessary to implement a smarter caching mechanism while keeping it correct blows the scope of this change... It's also worth noting that the only time it matters is when sharing is involved, so currently only in some of the spec tests...

I also looked into reverting #1594 but, unsurprisingly, that's an unclean revert.

Anyways, I'll plan for us to address in subsequent work.

> As this gets more and more complex, it might be worth checking what the use case of the cache was initially, and whether or not we can more easily work around it with other logic, like batching the host lookups entirely.

Unfortunately, the cache does still have an impact on the number of times that the DB is hit during a single request (see #1334 for original rationale)... I verified by profiling to confirm a difference in the number of calls to the query generated by `getHost`. I saw 3 with the caching enabled, and 5 without; without digging further, I'm not sure what drove the numbers into the 30s in the test case discussed in the original PR...

Ceiu:
@kahowell I don't understand the relationship between the multi-owner cache key and the updated field. Could you explain that a bit?

Ultimately, if this is still necessary, than it is what it is. I'm still skeptical, but if we have a task to clean it up properly, I'm not chuffed enough about it to hold this PR up.

kahowell:
> @kahowell I don't understand the relationship between the multi-owner cache key and the updated field. Could you explain that a bit?

Imagine two calls: `getHost(consumer, 'orgA', 'orgB')` and `getHost(consumer, 'orgB', 'orgC')`. Say that there are entries in all three orgs: OrgA's entry was updated at 3pm, OrgB's entry was updated at 2pm, and orgC's entry was updated at 1pm.

The query we are using will show from the first call that the host is the entry from OrgA, and the second call will show that the second call is the entry from OrgB.

Because the query only gives information about the latest org, we have not put OrgB's record in the cache, and therefore, will miss the cache for the second call. The query could be modified to return all, and then we could cache all the timestamps, but this is a significant change...

Ceiu:
And all that is fine. In the per-owner lookup, the process would check the cache for: guest-orga (miss), guest-orgb (miss), then hit the DB. DB returns org A's consumer, which gets cached as guest-orga. Second call checks guest-orgb (miss), then guest-orgc (miss), hitting the DB again and storing the consumer in guest-orgb.

The problem here is not these misses here, as the current code has the same "missing" problem as well: First call checks guest-orga-orgb (miss), fetches from DB and stores in the same key. Second call checks guest-orgb-orgc and misses as well. Our DB hit rate hasn't improved at all, and will only match the individual lookup in cases where we happen to repeat the exact call multiple times, which we shouldn't be doing (as this implies heavy misuse of the method).

Now, where all three implementations run into issues in the sharing case is when the first call is performed twice, but the last update time between org A and org B changes such that the other should be returned. Our cache lookup will fetch the last one when it shouldn't. Now the probability of this happening *should* be zero, as (1) we're in a transaction so such a change shouldn't even be visible to this thread and (2) if such an update is critical to the proper operation of this method, it should be gated with a lock on something to prevent it from happening (which I believe it is).

This implementation is an improvement over the original in that it won't return blatantly incorrect values, as our example above would (second call would return org A's consumer, which is wrong), so, following what we discussed earlier in that this PR doesn't make things worse and we have a task to fix this up later once we know what's happening with sharing, I'm content to continue with this as-is.

PR #1987, opened 2018-05-10
1576868: Checksum issues in liquibase script

wottop:


wottop:
retest this please

wottop:
retest this please

Ceiu:
I'm not seeing the fix here. When I run this locally against a fresh DB, deploying first on 2.2.3-1 and then 2.3.6-1 (or 2.3-HOTFIX), I don't have any issue. When I run this against my prod dump using the same process, I hit a different set of changeset mishaps.

I think that this may be a case where only IT's database is in a bad state, and they would be better off manually fixing the md5sum field rather than us adding another changeset allowance to these.

kahowell:
Closing in favor of #2011.

PR #1990, opened 2018-05-14
[m] Upgrade resteasy to 3.5.1

vritant:
related to #1996  .


mstead:
Overall, this looks good to me. One small concern I had was that the param name is no longer making it to the exception, which could be a little bit o fa PITA if the request has a lot of parameters. After chatting with Al, I'm ok with it -- especially considering the boilerplate that was required to get the name.

@awood If all is well, please merge once you are done looking at this one.

server/src/main/java/org/candlepin/resteasy/parameter/KeyValueParameter.java
```diff
@@ -31,30 +31,24 @@
...
-
-    @Override
-    void parse() throws CandlepinParameterParseException {
-        String[] parts = this.paramValue.split(":", 2); // Maximum of two parts
+    public KeyValueParameter(String inValue) {
+        String[] parts = inValue.split(":", 2); // Maximum of two parts
         if (parts.length == 1) {
-            throw new CandlepinParameterParseException(this.paramName,
-                this.paramName + "=name:value");
+            throw new CandlepinParameterParseException("name:value");
```
awood: Looks like the price of switching KeyValueParameter to just being a regular, bindable class is that we don't have the parameter name anymore to show in an error message.  E.g. "Bad format; Use fact=name:value" becomes "Bad format; Use name:value".  I'm okay with that, but just wanted to make sure that was a conscious choice.

vritant: @awood yes, this was done consciously.

PR #1991, opened 2018-05-14
[2.3] Make various components singletons

kahowell:
Same changes as #1985 , cherry-picked to 2.3.

There were only 3 conflicts, which were easy enough to resolve:
 * `server/src/main/java/org/candlepin/model/ExporterMetadataCurator.java`
 * `server/src/main/java/org/candlepin/model/IdentityCertificateCurator.java`
 * `server/src/main/java/org/candlepin/model/activationkeys/ActivationKeyCurator.java`


Ceiu:
ack -- looks about right.

kahowell:
4e6762628..fd54772e5: squashed

PR #1996, opened 2018-05-15
[m] removed ContentTypeHackFilter

vritant:
related to #1990


Ceiu:
I admit this is rather petty, but could we fix the typo in the commit message (ContenType => ContentType)? If we ever look through the logs for this change, we'll never find it with the string "ContentType"

awood:
@Ceiu done

vritant:
@awood , the bug was:
>  RESTEasy 2.2.1GA requires GET requests with no body to still have a content type set,  which at least some of our client stuff (the ruby lib) doesn't do. If you don't have the content type set, you'll end up getting an HTTP 415 response. This servlet filter simply sets a content type if none is detected.

based on this, GET requests should have errored out without this filter and the header. I removed the filter but the GET requests still did not fail ( even with the older resteasy 3.0.10 ).
based on this, I think we have not needed this filter for a long time now, hence the separate PR against master.

PR #2003, opened 2018-05-17
[m] Upgrade hibernate to 5.2.17

vritant:


vritant:

server/src/test/java/org/candlepin/test/SessionWrapper.java
```diff
@@ -16,65 +16,83 @@
...
+import javax.persistence.criteria.Selection;
+import javax.persistence.metamodel.Metamodel;
 
 /**
  * The SessionWrapper is a utility class intended to be used in places where we need to spy on an
  * existing session instance, but are unable to due to the "final" nature of Hibernate's
  * SessionImpl class.
  */
-public class SessionWrapper implements Session, SessionImplementor {
+public class SessionWrapper implements SessionImplementor {
```
wottop: If you extend AbstractSessionImpl, wouldn't it reduce the number of no-op implementations you added?

vritant: perhaps. reduces the class SessionWrapper from a 1600 line class to 900 line class...but we still have a *lot* of boilerplate code.
but that class implicitly introduces more interfaces than needed, so Imo we should just stick with this.
thoughts?

wottop: You convinced me. ack.

PR #2007, opened 2018-05-18
1578967: Fixed edge case that was causing Qpid connections to remain open

mstead:
When QpidQMF reports that qpid is in FLOW_STOPPED, the connection should
remain open. When the state is reported as CONNECTED, there is no need to
open a connection since it is already open.

Also added some tests around QpidConnection since none existed prior to
to this PR.

**To Test**
Follow the steps outlined in the BZ: 
https://bugzilla.redhat.com/show_bug.cgi?id=1578967


vritant:
@mstead this looks good to me, but I don't think the reproducer steps in the bug are relevant for this PR
please confirm if what I think is correct, if not we should discuss.

vritant:
was able to verify after talking to @mstead .

PR #2009, opened 2018-05-22
ENT-432 Converted Owner.key to a hibernate @NaturalId

nikosmoum:
- OwnerCurator.getByKey now uses session.bySimpleNaturalId to fetch an owner, and the old version of that method is renamed to getByKeySecure.
- Added explicit access check in ConsumerResource.create API due to it being marked as @SecurityHole.
- Deleted two @SecurityHole annotations that were in conflict with @Verify annotations (these are mutually exclusive).
- Deleted a couple of unused methods from ResolverUtil.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/OwnerProductResource.java
+++ b/server/src/main/java/org/candlepin/resource/OwnerProductResource.java
@@ -16,7 +16,6 @@ package org.candlepin.resource;
 
 import org.apache.commons.lang3.StringUtils;
 import org.candlepin.auth.Verify;
-import org.candlepin.common.auth.SecurityHole;
 import org.candlepin.common.config.Configuration;
 import org.candlepin.common.exceptions.BadRequestException;
 import org.candlepin.common.exceptions.ForbiddenException;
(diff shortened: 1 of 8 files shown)


nikosmoum:
It will be useful for reviewers to read [this document](https://docs.google.com/document/d/1_ucehD0e1jSFmVLrYhE0C8AkTqmy6FqYii9Zm4IyiqU/edit#) I compiled while investigating for this issue. No need to read all of it, just until the edit note.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -208,7 +207,6 @@ protected Content fetchContent(Owner owner, String contentId) {
     @GET
     @Path("/{product_id}")
     @Produces(MediaType.APPLICATION_JSON)
-    @SecurityHole
```
vritant: could you clarify why this is okay please?

vritant: nvm, found your answer in the PR description

PR #1986, opened 2018-05-10
1572669: Checksum issues in liquibase script

wottop:

PR #2010, opened 2018-05-22
[M] 1572669: Update liquibase checksums

kahowell:
See #2008. I ran those scripts against candlepin-2.2.3-1:

    server/bin/check_liquibase_checksums --base candlepin-2.2.3-1 --fix

Use of the script catches and fixes any `validCheckSum` which does not
match the current checksum, and also catches and fixes any changeset
which has a different checksum between HEAD and candlepin-2.2.3-1 but
does not have `validCheckSum` specified (it should).

Some of those caught are for deployment against Oracle databases,
which we test far less often.

Then I ran similarly against candlepin-0.9.54-HOTFIX, but no changes
resulted.

I did a little bit of cleanup afterwards (mainly removing stale
comments).

Obsoletes #1986.

### DIFF
--- a/server/src/main/resources/db/changelog/20130425102131-distributor-capabilities.xml
+++ b/server/src/main/resources/db/changelog/20130425102131-distributor-capabilities.xml
@@ -32,7 +32,11 @@
     </changeSet>
 
     <changeSet id="20130425102131-2" author="wpoteat" dbms="oracle,postgresql">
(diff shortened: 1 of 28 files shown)


kahowell:
Note that I included new checksums for liquibase-3.6.1, since that is the version that you get if you install via RPM on Fedora right now. It's of little functional consequence; Satellite repos use 3.1.0, and IT-PnT is using 3.5.0. I feel it doesn't hurt to future-proof though :-)

wottop:
ack

PR #2015, opened 2018-05-31
[M] 1584259: Detach consumers from persistence context to avoid OOM - ENT-583

awood:
consumerStackedEnts map.  During status calculation, the consumer
facts are accessed and during those accesses, the Hibernate proxy object
is replaced by the actual map of facts.  The consumerStackedEnts map
would accordingly grow as each status was calculated.

consumerStackedEnts map.  The consumers are then fetched again, the
status is updated, and then immediately detached to keep the persistence
context at roughly a constant size.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -2296,8 +2296,7 @@ public class CandlepinPoolManager implements PoolManager {
 
                 // Recalculate status for affected consumers
                 for (List<String> subList : Lists.partition(consumers, 1000)) {
-                    for (String consumerId : subList) {
-                        Consumer consumer = consumerCurator.get(consumerId);
+                    for (Consumer consumer : this.consumerCurator.getConsumers(subList).list()) {
                         this.complianceRules.getStatus(consumer);
                         this.consumerCurator.detach(consumer);
                     }
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -74,6 +74,7 @@ import org.candlepin.util.Util;
 import org.candlepin.util.Traceable;
 import org.candlepin.util.TraceableParam;
 
+import com.google.common.collect.Lists;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 
@@ -97,8 +98,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-
-
+import java.util.stream.Collectors;
 
 /**
  * PoolManager
@@ -2168,6 +2168,7 @@ public class CandlepinPoolManager implements PoolManager {
                 log.info("Revoking {} entitlements...", entitlements.size());
                 this.entitlementCurator.batchDelete(entitlements);
                 this.entitlementCurator.flush();
+                this.entitlementCurator.batchDetach(entitlements);
                 log.info("Entitlements successfully revoked");
             }
             else {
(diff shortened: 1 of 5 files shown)


awood:
Before:
* 252s to run
* perm generation maxed out at 5.2GB
expired-pools-master-before-252s
---
After:
* 337s to run
* perm generation maxed out at 1.9GB
expired-pools-master-after-337s

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2282,17 +2285,26 @@ public void deletePools(Collection<Pool> pools, Collection<String> alreadyDelete
...
-                for (Consumer consumer : consumerStackedEnts.keySet()) {
-                    this.complianceRules.getStatus(consumer);
+                /* Save the consumer IDs and then detach all the keys in the consumerStackEnts map.
+                 * Otherwise during the status calculations, the facts proxy objects objects will be resolved
+                 * and the memory use will grow linearly with the number of consumers instead of remaining
+                 * constant as we calculate the status of each consumer.  See BZ 1584259 */
+                List<String> consumers = consumerStackedEnts.keySet().stream()
+                    .map(Consumer::getId)
+                    .collect(Collectors.toList());
+                this.consumerCurator.batchDetach(consumerStackedEnts.keySet());
```
Ceiu: I'm still a little fuzzy on the particulars of the streaming API, but if I understand it correctly, we can avoid a second iteration through the collection by adding a forEach before the map to do the detaching:
```
List<String> consumers = ...
    .forEach(c -> this.consumerCurator.detach(c))
    .map(Consumer::getId)
    ...
```

My intent here is to do the detach and ID fetching in-line, rather than as two separate iterations through the collection.

As a side note, I just learned that we can apparently reference functions via the double colon syntax instead of manually calling them? Does this mean in my above example, we could just as easily do ```ConsumerCurator::detach``` rather than the longer lambda syntax? Feels like the scoping is wrong for such a ref.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2282,17 +2285,26 @@ public void deletePools(Collection<Pool> pools, Collection<String> alreadyDelete
...
 
-                    if (++i % 1000 == 0) {
-                        this.consumerCurator.flush();
+                log.info("Recomputing status for {} consumers", consumers.size());
+
+                // Recalculate status for affected consumers
+                for (List<String> subList : Lists.partition(consumers, 1000)) {
+                    for (String consumerId : subList) {
+                        Consumer consumer = consumerCurator.get(consumerId);
```
Ceiu: We briefly talked about this on Friday, but since we've evicted the consumers from the cache, this is going to hit the DB for every consumer again. If we changed this loop around a bit such that we can bulk up the consumer fetching for blocks of consumers, we'll save a bunch of time on DB hits/latency (in exchange for some of that memory we've been abusing):

```
for (/* partition loop */) {
    for (Consumer consumer : this.consumerCurator.getConsumers(subList)) {
        this.complianceRules.getStatus(consumer);
        this.consumerCurator.detach(consumer);
    }
    this.consumerCurator.flush();
}
```

If this is still heavy on memory, we could tune the block size accordingly; but anything higher than one should make a measurable difference.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2227,6 +2229,7 @@ public void deletePools(Collection<Pool> pools, Collection<String> alreadyDelete
                 }
 
                 this.poolCurator.updateAll(poolsToSave, false, false);
+                this.poolCurator.batchDetach(poolsToSave);
```
Ceiu: This one concerns me a bit, as there's one code path where we call into deletePools but it's not the last step of the path. If we detach these pools which are being saved here and they happen to be ones the caller is operating on, and the caller makes changes that are expected to be automatically persisted at the end of the transaction, we are broken.

I'll have to step through that path manually to see if that's a concern, but it's definitely something to note.

awood: I will remove this detach.  It's not essential and I didn't see it was passed in as a parameter.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
awood: That doesn't work.  `forEach` is a [void function](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#forEach-java.util.function.Consumer-).  It could be done by overloading the map function such that it detached the item and then grabbed the id, but that's adding a side-effect into what is meant to be functional code for a time savings that's probably measured in milliseconds.  Not worth it in my opinion.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
awood: `getConsumers` doesn't actually exist.  Is there maybe a method in AbstractHibernateCurator you were thinking of?

Ceiu: Nope, it'd have to be written, but it's fairly trivial to write. Should be no more than five lines of code since we're using the consumer ID.

awood:
Bulk fetching looks good to me.

awood:
After batched commit from @Ceiu 
* 312s to run
* perm generation maxed out at 1.4GB
expired-pools-master-batched-312s

PR #2016, opened 2018-05-31
[2.1] 1584256: Detach consumers from persistence context to avoid OOM - ENT-583

awood:
consumerStackedEnts map.  During status calculation, the consumer
facts are accessed and during those accesses, the Hibernate proxy object
is replaced by the actual map of facts.  The consumerStackedEnts map
would accordingly grow as each status was calculated.

consumerStackedEnts map.  The consumers are then fetched again, the
status is updated, and then immediately detached to keep the persistence
context at roughly a constant size.


awood:
Before:
expired-pools-2 1-before-713s
* 713s to run
* perm generation maxed out at 5.2 GB

---
After (612s):
expired-pools-2 1-after-612s
* 612s to run
* perm generation maxed out at 2GB

awood:

PR #2017, opened 2018-06-01
ENT-93 Renamed and moved SchedulerStatus to the dto package

nikosmoum:


Ceiu:
Simple enough change, but let's also update any DTOs that land in the DTO packages so they conform to the standards and designs used in the others. We won't need a translator for this one, but we should still add the DTO test and have it extend CandlepinDTO.

server/src/main/java/org/candlepin/resource/JobResource.java
```diff
@@ -145,20 +144,15 @@ else if (!StringUtils.isEmpty(principalName)) {
         }
 
         return this.translator.translateQuery(
-            statuses.transform(new ElementTransformer<JobStatus, JobStatus>() {
-                @Override
-                public JobStatus transform(JobStatus jobStatus) {
-                    return jobStatus.cloakResultData(true);
-                }
-            }), JobStatusDTO.class);
+            statuses.transform(jobStatus -> jobStatus.cloakResultData(true)), JobStatusDTO.class);
```
Ceiu: :+1:

I didn't realize the compiler was smart enough to do all of this with a single operator. Good to know.

nikosmoum: Yep, its because ElementTransformer happens to be a functional interface (has exactly 1 abstract method, so it knows whatever comes after -> is basically the implementation of this one method).

Ceiu:
ack

PR #2018, opened 2018-06-01
[M] Moved Status from the model package to the DTO package

Ceiu:
- Converted Status to StatusDTO and moved it to the DTO package
- Updated the StatusDTO object to conform with other DTO design guidelines,
  such as returning self references from mutators, adding clone and
  populate methods, and not containing references to model objects
- Updated the CandlepinContextListener to no longer reference the Status
  object directly
- Added the new CandlepinCapabilities object which stores the capabilities
  set in the Resteasy context rather than in a static array on Status


Ceiu:
Reviewers: Give some extra attention to the context/capability changes, as I would like verification that this doesn't have its own set of shortcomings to address.

server/src/main/java/org/candlepin/dto/api/v1/StatusDTO.java
```diff
@@ -0,0 +1,368 @@
...
+     * @param capabilities
+     *  The capabilities to set, or null to clear any existing capabilities
+     *
+     * @return
+     *  A reference to this DTO
+     */
+    public StatusDTO setManagerCapabilities(Collection<String> capabilities) {
+        if (capabilities != null) {
+            if (this.capabilities == null) {
+                this.capabilities = new HashSet<String>();
```
nikosmoum: Something else that I missed. I know it's not that important, but we don't have to set the type when creating new objects any more. We can just use the diamond operator `new HashSet<>`.

server/src/test/java/org/candlepin/dto/api/v1/StatusDTOTest.java
```diff
@@ -0,0 +1,146 @@
...
+public class StatusDTOTest extends AbstractDTOTest<StatusDTO> {
+
+
+    protected Map<String, Object> values;
+    protected OwnerDTOTest ownerDTOTest = new OwnerDTOTest();
+    protected EnvironmentDTOTest environmentDTOTest = new EnvironmentDTOTest();
+    protected ConsumerInstalledProductDTOTest cipDTOTest = new ConsumerInstalledProductDTOTest();
+    protected CapabilityDTOTest capabilityDTOTest = new CapabilityDTOTest();
+    protected HypervisorIdDTOTest hypervisorIdDTOTest = new HypervisorIdDTOTest();
+    protected GuestIdDTOTest guestIdDTOTest = new GuestIdDTOTest();
```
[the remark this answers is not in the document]
Ceiu: Yep. Fixed.

server/src/main/java/org/candlepin/dto/api/v1/StatusDTO.java
Ceiu: Force of habit. I imagine that'll be a continuing theme for some time. Fixed.

PR #2019, opened 2018-06-04
[2.4] 1584258: Detach consumers from persistence context to avoid OOM - ENT-583

awood:
consumerStackedEnts map.  During status calculation, the consumer facts
are accessed and during those accesses, the Hibernate proxy object is
replaced by the actual map of facts.  The consumerStackedEnts map would
accordingly grow as each status was calculated.

consumerStackedEnts map.  The consumers are then fetched again, receive
an updated status, and are then immediately detached to keep the
persistence context at roughly a constant size.


awood:
Before:
expired-pools-2 4-before-216s
* 216s to run
* perm generation maximum 2GB
---
After:
* 287s to run
* perm generation maximum 4.7GB
expired-pools-2 4-after-287s

PR #2026, opened 2018-06-05
1510920: Allow consumer to cancel jobs started by consumer

wottop:
The permissions for JobResource were previously modified to allow a consumer to check
 the status of a job it started.  This scenario was introduced so that the virt-who
 process on a registered system could query for the job status. To assist the virt-who
 choreography, the consumer will be allowed to cancel its own jobs.

### DIFF
--- a/server/src/main/java/org/candlepin/auth/permissions/JobStatusPermission.java
+++ b/server/src/main/java/org/candlepin/auth/permissions/JobStatusPermission.java
@@ -28,17 +28,19 @@ import org.hibernate.criterion.Restrictions;
 import java.util.List;
 
 /**
- * Allows users/consumers access to any JobStatus from jobs that they
- * created/initiated.
+ * Determines users and consumers specific access to JobStatus.
+ *
+ * Users can view all jobs in their org, but only cancel jobs they started.
+ * Consumers can view and cancel jobs they started, but only those.
  *
  */
-public class CheckJobStatusPermission extends TypedPermission<JobStatus> {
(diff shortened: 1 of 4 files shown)


wottop:
retest this please

mstead:
@wottop Please add some context to the commit message as to what this is fixing and why it is being done. The BZ conversations seem unrelated though it is.

wottop:
retest this please

server/src/main/java/org/candlepin/auth/permissions/JobStatusPermission.java
```diff
@@ -32,13 +32,13 @@
  * created/initiated.
  *
  */
```
mstead: Please update the javadoc to reflect that this is a general JobStatus permission and give a little info about what it allows access to.

server/src/main/java/org/candlepin/auth/permissions/JobStatusPermission.java
```diff
@@ -99,11 +97,21 @@ public boolean canAccessTarget(JobStatus target, SubResource subResource, Access
...
+            }
+            else {
+                return Access.READ_ONLY.provides(required);
+            }
+        }
+        else if ("consumer".equalsIgnoreCase(principalType)) {
+            // if this is a consumer it must match the principle name access
+            return Access.ALL.provides(required) && principalNameMatch;
+        }
+        else {
```
mstead: Being a little picky, but you can save a couple of lines since the last else is not required -- you can just return false.

wottop:
retest this please

PR #2030, opened 2018-06-07
[M] 1588202: Fixed status endpoint not setting the timeUTC field

Ceiu:
- StatusResource now sets the timeUTC field of the StatusDTO object
  before returning it.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/StatusResource.java
+++ b/server/src/main/java/org/candlepin/resource/StatusResource.java
@@ -36,6 +36,7 @@ import org.slf4j.LoggerFactory;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 
+import java.util.Date;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


Ceiu:
The spec failure is an issue currently present in master and will be fixed in another PR.

server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -155,7 +157,8 @@ public StatusDTO status() {
             .setMode(mode != null ? mode.toString() : null)
             .setModeReason(modeChangeReason != null ? modeChangeReason.toString() : null)
             .setModeChangeTime(modeChange.getChangeTime())
-            .setManagerCapabilities(caps);
+            .setManagerCapabilities(caps)
+            .setTimeUTC(new Date());
```
awood: I'm worried this won't handle leap seconds appropriately. :nerd_face:

PR #2031, opened 2018-06-08
[M] Upgraded postgresql-jdbc driver to the latest ENT-521

mstead:
BZ 1566200 occurs because of an issue with the postgres JDBC driver which causes some of the operations that liquibase executes against Postgres 9.6 to fail.

cpdb sets the classpath when running liquibase to the RPM installed version of the driver which causes the breakage.

The postgres driver will be updated for RHEL 7.5.z and 7.6 ( https://bugzilla.redhat.com/show_bug.cgi?id=1566200#c2 ) which will address the installer issue.

Candlepin runs on its own version of the driver which is included in the WAR. Although it is very old, it does not cause any issues during normal operation against postgres 9.6.

This PR updates the version of the driver to the latest ONLY in the master branch since it is really outdated, however, this in no way addresses the BZ. We will have to wait for the driver updates to hit RHEL 7 before we can close out the bug.

### DIFF
--- a/buildfile
+++ b/buildfile
@@ -120,7 +120,7 @@ HIBERNATE = [group('hibernate-core', 'hibernate-entitymanager', 'hibernate-c3p0'
              'org.jboss.logging:jboss-logging:jar:3.3.0.Final'
              ] + JAVAX
 
-POSTGRESQL = 'postgresql:postgresql:jar:9.0-801.jdbc4'
+POSTGRESQL = 'org.postgresql:postgresql:jar:42.2.2'
 
 SWAGGER = [group('swagger-jaxrs', 'swagger-core','swagger-models','swagger-annotations',
                      :under => 'io.swagger',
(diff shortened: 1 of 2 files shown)


buildfile
```diff
@@ -120,7 +120,7 @@ HIBERNATE = [group('hibernate-core', 'hibernate-entitymanager', 'hibernate-c3p0'
              'org.jboss.logging:jboss-logging:jar:3.3.0.Final'
              ] + JAVAX
 
-POSTGRESQL = 'postgresql:postgresql:jar:9.0-801.jdbc4'
+	     POSTGRESQL = 'org.postgresql:postgresql:jar:42.2.2'
```
awood: Why the change in indentation?  This looks kind of funny now.

mstead: @awood Oops. That wasn't intentional! Fixing!

mstead: @awood Fixed the indent issue.

mstead:
@awood All fixed up. Thanks for looking.

PR #2032, opened 2018-06-11
Remove revoke op  ENT-413

vritant:
@reviewer: While this was supposed to be a PR for the last step of lockless 2, CheckBonusPoolQuantitiesOp does not lend itself to the lock-less framework easily. To do so would require a complete overhaul of the CandlepinPoolManager.revokeEntitlements implementation, but once I got into the weeds of it, it became more and more clear that the restructuring that will be introduced is not warranted by the gains we could gain from it.

nevertheless, I have removed RevokeOP and the sharing logic implemented in it since we are now undoing sharing related implementation anyway.


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -697,6 +695,49 @@ protected int cleanupExpiredPoolsImpl() {
         return entitlementsToRegen;
     }
 
+    Set<Pool> revokeEntitlementsFromOverflowingPools(List<Pool> pools) {
+        Collection<Pool> overflowing = new ArrayList<>();
+        for (Pool pool : pools) {
```
awood: This would be a really good place to use the stream api.

```
Collection<Pool> overflowing = pools.stream().filter(Pool::isOverflowing).collect(Collectors.toList())
```

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -697,6 +695,49 @@ protected int cleanupExpiredPoolsImpl() {
...
+        }
+        List<Entitlement> entitlementsToRevoke = new ArrayList<>();
+        overflowing = poolCurator.lockAndLoad(overflowing);
+
+        List<Entitlement> overFlowingEnts = this.poolCurator.retrieveOrderedEntitlementsOf(overflowing);
+        Map<String, List<Entitlement>> entMap = new HashMap<>();
+        for (Entitlement entitlement : overFlowingEnts) {
+            List<Entitlement> poolEnts = entMap.get(entitlement.getPool().getId());
+            if (poolEnts == null) {
+                poolEnts = new ArrayList<Entitlement>();
```
awood: Maybe use [`computeIfAbsent`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-) here.

```
entMap.computeIfAbsent(entitlement.getPool().getId(), k -> new ArrayList<Entitlement>()).add(entitlement);
```

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -697,6 +695,49 @@ protected int cleanupExpiredPoolsImpl() {
...
+            poolEnts.add(entitlement);
+        }
+
+        for (Pool pool : overflowing) {
+            // we then start revoking the existing entitlements
+            List<Entitlement> entitlements = entMap.get(pool.getId());
+            long newConsumed = pool.getConsumed();
+
+            for (Entitlement ent : entitlements) {
+                if (newConsumed > pool.getQuantity()) {
```
awood: As written, this is a basically non-deterministic for revocations, right?  Would it make sense to apply some sort of heuristic to keep things consistent?  For example, sort the entitlements by quantity ascending and revoke the smallest entitlements first.

vritant:
@awood done

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -697,6 +695,41 @@ protected int cleanupExpiredPoolsImpl() {
         return entitlementsToRegen;
     }
 
+    Set<Pool> revokeEntitlementsFromOverflowingPools(List<Pool> pools) {
```
awood: Is this default visibility for a reason?  Not a big deal, but I don't normally see methods that we keep package-protected.

PR #2034, opened 2018-06-12
[M] ENT-600 Add a sleep to fix failing spec test against mysql

mstead:
Because of the lack of DATETIME precision in MySQL, it appears
this this test will fail randomly when an export and re-import
occur at just the right moment causing a 409 MANIFEST_SAME conflict
due to matching time.

**NOTE**
Once this PR runs through the tests and we don't see the failure in other PRs against master, I'll create the PRs for 2.4 and 2.5


Ceiu:
I wasn't able to trigger the issue on master, but the sleep also doesn't cause problems in this branch; and given what we know about mysql and sub-second intervals, we know this could be a problem.

ACK.

PR #2037, opened 2018-06-18
[2.3] Add a sleep to fix failing spec test against mysql

mstead:
Because of the lack of DATETIME precision in MySQL, it appears
this this test will fail randomly when an export and re-import
occur at just the right moment causing a 409 MANIFEST_SAME conflict
due to matching time.

### DIFF
--- a/server/spec/import_excess_spec.rb
+++ b/server/spec/import_excess_spec.rb
@@ -36,6 +36,8 @@ describe 'Import Update', :serial => true do
     all_pools = @user.list_pools :owner => @import_owner.id
     all_pools.size.should == 3
 
+    # Sleep to create a gap between when the manifest was initially imported
+    sleep 1
     # Now lets import
     updated_export = @exporter.create_candlepin_export_update()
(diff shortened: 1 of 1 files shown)


vritant:
found another non deterministic failure :(

15:34:07   1) Product Versioning should cleanup orphans without interfering with normal actions
15:34:07      Failure/Error: product = @cp.create_product(owner2["key"], id, id)
15:34:07 
15:34:07      RestClient::InternalServerError:
15:34:07        500 Internal Server Error

mstead:
retest this please

mstead:
> found another non deterministic failure :(
> 
> 15:34:07 1) Product Versioning should cleanup orphans without interfering with normal actions
> 15:34:07 Failure/Error: product = @cp.create_product(owner2["key"], id, id)
> 15:34:07
> 15:34:07 RestClient::InternalServerError:
> 15:34:07 500 Internal Server Error

I haven't seen this one before. Note taken. If we start seeing this more, perhaps we can tackle that one later.

Ceiu:
ack

PR #2041, opened 2018-06-19
[M] Added the ContentOverrideDTO and ContentOverrideTranslator (ENT-86, ENT-87)

Ceiu:
- Added the new ContentOverrideDTO and its translator
- Moved ActivationKeyContentOverrideDTO to its own class as a subclass
  of the new ContentOverrideDTO
- Added the ConsumerContentOverrideDTO as a subclass of the new
  ContentOverrideDTO
- Updated the OwnerResource to use the new DTO where appropriate
- Updated several resources to use the new DTOs where the old ContentOverride
  entities were being output
- Moved the content override validation from the rules to native
  Java and deprecated the validation code within rules.js
- Removed the OverrideRules class, as it is no longer used
- Fixed several issues found in a number of DTOs and translators
  surrounding content overrides
- Updated several tests to use the new DTOs and fewer mocks where
  appropriate


server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyContentOverrideDTO.java
```diff
@@ -0,0 +1,152 @@
...
+        return copy;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public ActivationKeyContentOverrideDTO populate(ActivationKeyContentOverrideDTO source) {
+        super.populate(source);
+
```
nikosmoum: Need to add `this.setActivationKey(source.getActivationKey());` here.

server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyContentOverrideDTO.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Initializes a new ActivationKeyContentOverrideDTO instance which is a shallow copy of the
+     * provided source entity.
+     *
+     * @param source
+     *  The source entity to copy
+     */
+    public ActivationKeyContentOverrideDTO(ActivationKeyContentOverrideDTO source) {
+        this.populate(source);
```
nikosmoum: Should probably call `super(source);` instead of populate

server/src/main/java/org/candlepin/dto/api/v1/ConsumerContentOverrideDTO.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Initializes a new ConsumerContentOverrideDTO instance which is a shallow copy of the provided
+     * source entity.
+     *
+     * @param source
+     *  The source entity to copy
+     */
+    public ConsumerContentOverrideDTO(ConsumerContentOverrideDTO source) {
+        this.populate(source);
```
nikosmoum: Should probably call `super(source);` instead of populate

server/src/main/java/org/candlepin/dto/api/v1/ContentOverrideDTO.java
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * Initializes a new ContentOverrideDTO instance which is a shallow copy of the provided
+     * source entity.
+     *
+     * @param source
+     *  The source entity to copy
+     */
+    public ContentOverrideDTO(T source) {
+        this.populate(source);
```
nikosmoum: Same as the subclasses, this should be `super(source)`

server/src/main/java/org/candlepin/dto/api/v1/ConsumerContentOverrideDTO.java
```diff
@@ -0,0 +1,155 @@
...
+        return copy;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public ConsumerContentOverrideDTO populate(ConsumerContentOverrideDTO source) {
+        super.populate(source);
+
```
nikosmoum: Should also add `this.setConsumer(source.getConsumer());`

server/src/test/java/org/candlepin/dto/api/v1/ActivationKeyTranslatorTest.java
```diff
@@ -89,21 +105,75 @@ protected void verifyOutput(ActivationKey source, ActivationKeyDTO dest, boolean
...
-                    for (String prodDto : dest.getProductIds()) {
+                for (ActivationKeyContentOverride override : overrides) {
+                    assertNotNull(override);
 
-                        assertNotNull(prodDto);
+                    for (ActivationKeyContentOverrideDTO odto : overrideDTOs) {
+                        assertNotNull(odto);
 
-                        if (prodDto.equals(prod.getId())) {
-                            // Nothing else to assert on, since prodDto only holds the product id.
```
nikosmoum: what was I thinking haha

server/src/main/resources/rules/rules.js
```diff
@@ -3361,10 +3361,13 @@ var PoolType = {
...
 /**
- * Namespace for determining ability to override a specific content set
- *  value
+ * Namespace for determining ability to override a specific content set value
  *
  * The list of disallowed value names are listed in this method
+ *
+ * DEPRECATED:
+ *  This functionality has been moved to ContentOverrideValidator as of Candlepin 2.5.5. This method
+ *  will no longer be called by Candlepin and is considered deprecated.
```
nikosmoum: This is more like a question; do we have to bump the minor version of the file even if it is a comment that changed, or not? (my intuition says no, but I just wanted to make sure)

server/src/main/java/org/candlepin/dto/api/v1/ConsumerContentOverrideDTO.java
```diff
@@ -0,0 +1,155 @@
...
+     */
+    @Override
+    public String toString() {
+        ConsumerDTO consumer = this.getConsumer();
+
+        return String.format("ConsumerContentOverrideDTO [consumer: %s, content: %s, name: %s, value: %s]",
+            this.getContentLabel(),
+            this.getName(),
+            this.getValue(),
+            (consumer != null ? consumer.getId() : null));
```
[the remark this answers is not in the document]
Ceiu: Fixed.

server/src/test/java/org/candlepin/dto/api/v1/ActivationKeyContentOverrideDTOTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+
+
+/**
+ * Test suite for the ActivationKeyContentOverrideDTO class
+ */
+public abstract class ActivationKeyContentOverrideDTOTest extends
+    AbstractDTOTest<ActivationKeyContentOverrideDTO> {
```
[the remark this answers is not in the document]
Ceiu: Fixed.

server/src/test/java/org/candlepin/dto/api/v1/ConsumerContentOverrideDTOTest.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+
+
+/**
+ * Test suite for the ConsumerContentOverrideDTO class
+ */
+public abstract class ConsumerContentOverrideDTOTest extends AbstractDTOTest<ConsumerContentOverrideDTO> {
```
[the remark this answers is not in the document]
Ceiu: Fixed.

server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyContentOverrideDTO.java
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String toString() {
+        ActivationKeyDTO key = this.getActivationKey();
+
+        return String.format("ActivationKeyContentOverrideDTO [key: %s, content: %s, name: %s, value: %s]",
+            this.getContentLabel(), this.getName(), this.getValue(), (key != null ? key.getId() : null));
```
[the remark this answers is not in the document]
Ceiu: Fixed.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerContentOverrideDTO.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Fetches the activation key that's overriding the content. If the activation key has not been
+     * set, this method returns null.
+     *
+     * @return
+     *  The activation key overriding the content, or null if the activation key has not been set
+     */
+    @JsonIgnore
+    public ConsumerDTO getConsumer() {
```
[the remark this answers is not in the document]
Ceiu: Fixed.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerContentOverrideDTO.java
```diff
@@ -0,0 +1,155 @@
...
+     * Sets or clears the activation key that's overriding the content. If the provided key is null,
+     * any key currently set will be cleared.
+     *
+     * @param consumer
+     *  The activation key that's overriding the content, or null to clear the key
+     *
+     * @return
+     *  A reference to this DTO
+     */
+    @JsonProperty("key")
```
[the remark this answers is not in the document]
Ceiu: Fixed.

server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyTranslator.java
```diff
@@ -70,58 +71,76 @@ public ActivationKeyDTO populate(ModelTranslator modelTranslator,
...
+        if (overrides != null) {
+            Set<ActivationKeyContentOverrideDTO> overrideDTOs = new HashSet<>();
+
+            for (ActivationKeyContentOverride override : overrides) {
+                if (override != null) {
+                    overrideDTOs.add(new ActivationKeyContentOverrideDTO()
+                        .setActivationKey(dest)
+                        .setContentLabel(override.getContentLabel())
+                        .setName(override.getName())
+                        .setValue(override.getValue()));
```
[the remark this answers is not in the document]
Ceiu: This was a tough decision to make, and went through a couple of revisions. At one point in this branch there was an explicit translator for AK content overrides, but it ran into the problem of a circular creation loop since the state between the translators isn't shared. We probably *could* drop this down to a generic content override and use those, but then we're intentionally building incomplete objects, and we'd have to change the collection on the AK DTOs.

After weighing all the pros and cons, I opted for this solution.

server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyContentOverrideDTO.java
Ceiu: Fixed.

server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyContentOverrideDTO.java
Ceiu: Fixed.

server/src/main/java/org/candlepin/dto/api/v1/ContentOverrideDTO.java
Ceiu: Fixed.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerContentOverrideDTO.java
Ceiu: Fixed.

server/src/main/resources/rules/rules.js
Ceiu: Yup, definitely should. The version check is used to determine if the rules are worth replacing at all. If we don't bump the version, the comment won't make it into the rules file.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerContentOverrideDTO.java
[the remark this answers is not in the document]
nikosmoum: I did a quick lookup and it looks to me that the tests aren't actually setting this field themselves (at least the couple I checked. I didn't do a thorough check). But then I looked at ContentOverrideResource and it looks like the resources themselves aren't even looking for this field, and instead, they're retrieving the uuid/id from the URL (in the case of consumerCO they retrieve the consumer_uuid, and in the case of the akCO they're using the activation_key_id), and then load the actual entity from the db:
 https://github.com/candlepin/candlepin/blob/7099ceee48270e8d153c5858192f4792f05be68e/server/src/main/java/org/candlepin/resource/ContentOverrideResource.java#L93-L95

Currently running the spec tests locally with this field completely ignored in both DTOs since it seems they aren't needed on client input.

Ceiu:
Merged

PR #2046, opened 2018-06-25
[M] Updated UserResource to use UserDTOs instead of User entities (ENT-414)

Ceiu:
- UserResource now uses UserDTOs for input and output, and outputs
  OwnerDTOs and RoleDTOs where applicable
- UserResource now does a bit more input validation when performing
  user operations via username, and will return the appropriate
  response if the username is invalid or doesn't exist
- The exception message when attempting to create a user that
  already exists is now properly translated to non-English
  langages where appropriate
- Updated the unit and spec tests in accordance with the above
  changes


server/src/main/java/org/candlepin/resource/UserResource.java
```diff
@@ -141,31 +207,38 @@ public User createUser(@ApiParam(hidden = true) User user) {
...
+
+        if (dto.isSuperAdmin() != null) {
+            user.setSuperAdmin(dto.isSuperAdmin());
+        }
+
+        return this.modelTranslator.translate(userService.updateUser(user), UserDTO.class);
+    }
 
     @ApiOperation(notes = "Removes a User", value = "deleteUser")
     @ApiResponses({ @ApiResponse(code = 410, message = "") })
```
nikosmoum: If we're throwing `NotFoundException` instead of `GoneException` now, then we should change this to 404. Maybe even add in the 400 since it could throw `BadRequestException`.

common/pom.xml
```diff
@@ -67,6 +67,7 @@
     <com.fasterxml.jackson.module-jackson-module-jsonSchema.version>2.9.4</com.fasterxml.jackson.module-jackson-module-jsonSchema.version>
     <com.fasterxml.jackson.module-jackson-module-jaxb-annotations.version>2.9.4</com.fasterxml.jackson.module-jackson-module-jaxb-annotations.version>
     <com.fasterxml.jackson.datatype-jackson-datatype-hibernate5.version>2.9.4</com.fasterxml.jackson.datatype-jackson-datatype-hibernate5.version>
+    <com.fasterxml.jackson.datatype-jackson-datatype-jdk8.version>2.9.4</com.fasterxml.jackson.datatype-jackson-datatype-jdk8.version>
```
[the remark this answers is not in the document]
Ceiu: The pom.xml files are automatically updated based on changes made to the buildfile. While you're probably right, changing it will just cause it to get overwritten on the next build.

server/src/main/java/org/candlepin/resource/UserResource.java
Ceiu: Done.

PR #2050, opened 2018-07-03
ENT-616: Update SubjectKeyIdentifierWriter for isolation [m]

aparadkar:
- add toOctetString default method
- keeps IT from having to use BouncyCastle directly
- moved getSubjectKeyidentifier to the interface

### DIFF
--- a/server/src/main/java/org/candlepin/pki/SubjectKeyIdentifierWriter.java
+++ b/server/src/main/java/org/candlepin/pki/SubjectKeyIdentifierWriter.java
@@ -16,7 +16,6 @@ package org.candlepin.pki;
 
 import org.bouncycastle.asn1.ASN1Encoding;
 import org.bouncycastle.asn1.DEROctetString;
-import org.bouncycastle.cert.jcajce.JcaX509ExtensionUtils;
 
 import java.io.IOException;
 import java.security.KeyPair;
@@ -31,7 +30,7 @@ import java.util.Set;
  * extension so it may be necessary to hijack the extension as an alternative communications channel.
  */
 public interface SubjectKeyIdentifierWriter {
-     /**
+    /**
      * This method returns the SubjectKeyIdentifier extension that will be written into a certificate.
      *
      * @param clientKeyPair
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/pki/SubjectKeyIdentifierWriter.java
```diff
@@ -36,6 +40,14 @@
      * @throws IOException thrown if error reading cert
      * @throws NoSuchAlgorithmException thrown if JcaX509ExtensionUtils can't be created
      */
-    byte[] getSubjectKeyIdentifier(KeyPair clientKeyPair, Set<X509ExtensionWrapper> extensions)
-        throws IOException, NoSuchAlgorithmException;
+    default byte[] getSubjectKeyIdentifier(KeyPair clientKeyPair, Set<X509ExtensionWrapper> extensions)
```
awood: I would prefer that we keep this method definition inside DefaultSubjectKeyIdentifierWriter.  The reason being that as it stands, DefaultSubjectKeyIndentifierWriter is just an empty class.

server/src/main/java/org/candlepin/pki/SubjectKeyIdentifierWriter.java
```diff
@@ -36,6 +40,14 @@
...
+            throws IOException, NoSuchAlgorithmException {
+        return new JcaX509ExtensionUtils().createSubjectKeyIdentifier(clientKeyPair.getPublic()).getEncoded();
+    }
+
+    default byte[] toOctetString(byte[] bytes_to_encode) throws IOException {
```
awood: Please use camel case for your parameters instead of snake care.  E.g. "byteToEncode" instead of "bytes_to_encode".  Snake case is the idiomatic pattern in Python, but Java favors camel case.

server/src/main/java/org/candlepin/pki/SubjectKeyIdentifierWriter.java
```diff
@@ -36,6 +40,14 @@
...
+        DEROctetString octetStr = new DEROctetString(bytes_to_encode);
+        return octetStr.getEncoded(ASN1Encoding.DER);
```
awood: I don't think you need to specify ASN1Encoding.DER since that's implicit in the object type, DEROctetString.  Try without and see if the unit tests still pass.

server/src/test/java/org/candlepin/pki/SubjectKeyIdentifierWriterTest.java
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+public class SubjectKeyIdentifierWriterTest {
+
+    @Test
+    public void testToOctetString() throws Exception{
+        DefaultSubjectKeyIdentifierWriter writer = new DefaultSubjectKeyIdentifierWriter();
+
+        byte[] foobar_in_asn1 = {0x4, 0x6, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72};
```
awood: I'd pad these to consistently use 2 hex characters and I'd leave a comment here like

"0x04 is the DER tag for an octet string and 0x06 is the subsequent length of the octet string"

PR #2051, opened 2018-07-05
Switch hashing algorithms from SHA-256 to Murmur3 for compliance hash

awood:
Compliance hashes do not need to be cryptographically secure, so using
SHA-256 is overkill.  Murmur3 is slightly faster and results in modest
performance improvement (roughly 10%).

---
Running ExpiredPoolsJob on a DB with about 10,500 consumers that need status recalculations goes from 195s to 181s on my machine (give or take).

### DIFF
--- a/server/src/main/java/org/candlepin/policy/js/compliance/hash/Hasher.java
+++ b/server/src/main/java/org/candlepin/policy/js/compliance/hash/Hasher.java
@@ -15,10 +15,14 @@
 
 package org.candlepin.policy.js.compliance.hash;
 
-import org.apache.commons.codec.digest.DigestUtils;
+import com.google.common.hash.HashCode;
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hashing;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.nio.charset.Charset;
 import java.util.Collection;
 
 /**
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/policy/js/compliance/hash/Hasher.java
```diff
@@ -28,19 +32,28 @@
...
+    );
+
     public Hasher() {
         sink = new StringBuilder();
     }
 
     /**
-     * Products an SHA256 hash of anything that was put into this hasher.
+     * Produces two concatenated Murmur3 hashes of anything that was put into this hasher to produce a 256
+     * bit digest.  We concatenate two 128 bit digests for legacy reasons: previously we used SHA256 which
```
kahowell: Would it be faster to concatenate the same hash twice (i.e. 

```java
HASH_FUNCTION = Hashing.murmur3_128();
//...
return c.toString() + c.toString()
```
)? Or something similar?

awood: Slightly faster, yes.  Probably on the order of another 1 or 2%.  I was basing my approach on [this](https://github.com/google/guava/blob/master/guava/src/com/google/common/hash/Hashing.java#L75).  I think I would like to keep it so that the hash is distributed across 256 bits instead of just 128 doubled up.

awood:
@candlepin-pull-request-bot Retest this please

awood:
@candlepin-pull-request-bot Retest this please

kahowell:
ACK.

PR #2056, opened 2018-07-16
Create/Update Consumer now updates syspurpose data

mstead:
This patch makes the necessary changes to store the system
purpose data on a Consumer via the create/update consumer APIs.

* Added table for syspurpose add ons
* Added sp_status and sp_status_hash to cp_consumer
* Renamed column names to be prefixed with sp_ for added clarity.
* Renamed offering on Consumer class and DTO to role.
* Implemented Consumer and ConsumerDTO changes for sys purpose
* Made changes to ConsumerResource's create/update APIs to update
  system purpose fields.


server/client/ruby/candlepin_api.rb
```diff
@@ -79,14 +79,16 @@ def register(name, type=:system, uuid=nil, facts={}, username=nil,
               environment=nil, capabilities=[], hypervisor_id=nil,
               content_tags=[], created_date=nil, last_checkin_date=nil,
               annotations=nil, recipient_owner_key=nil, user_agent=nil,
-              entitlement_count=0, id_cert=nil, offering=nil, usage=nil)
+              entitlement_count=0, id_cert=nil, role=nil, usage=nil,
```
awood: `offering` is a dead letter now I guess?

server/spec/consumer_resource_spec.rb
```diff
@@ -91,15 +108,33 @@
...
+      add_ons = consumer['addOns']
+      add_ons.size.should == 2
+      add_ons.should include('add-on 1')
+      add_ons.should include('add-on 2')
+
+      update_args = {
+          :uuid => consumer['uuid'],
+          :role => 'updatedrole',
+          :usage => 'updatedusage',
+          :addOns => ['add-on 2', 'add-on 4']
```
awood: Should we add SLA to this test too since that's part of the system purpose matrix?

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -293,6 +295,58 @@ public ConsumerDTO setUsage(String usage) {
...
+     * Sets the system purpose add ons on this ConsumerDTO object.
+     *
+     * @param addOns the add ons to set on this ConsumerDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ConsumerDTO setAddOns(Set<String> addOns) {
+        if (addOns != null) {
+            if (this.addOns == null) {
+                this.addOns = new HashSet<>();
```
awood: Do you think it would make sense to make `addOns` an unmodifiable set via `Collections` since it looks like the set is meant to be cleared on an update instead of modified.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -124,15 +124,30 @@
...
+    @CollectionTable(name = "cp_sp_add_on", joinColumns = @JoinColumn(name = "consumer_id"))
+    @Column(name = "add_on")
+    private Set<String> addOns = new HashSet<>();
+
+    @Column(name = "sp_status", length = 32)
+    @Size(max = 32)
+    private String systemPurposeStatus;
+
+    /**
+     * Represents an SHA256 hex string of the last calculated system purpose status.
```
awood: It's murmur3 now that #2051 is merged.  I would just say "a 256 bit hash digest" to keep it vague.

server/src/main/resources/db/changelog/20180430101420-add-intent-fields-to-consumer.xml
```diff
@@ -7,14 +7,61 @@
         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
     <changeSet id="20180430101420-1" author="vrjain">
-        <comment> add-intent-fields-to-consumer</comment>
+        <comment> Add system purpose role column to cp_consumer</comment>
         <addColumn tableName="cp_consumer">
-            <column name="offering" type="VARCHAR(255)"/>
+            <column name="sp_role" type="VARCHAR(255)"/>
```
awood: Just making sure, but this changeSet hasn't been release anywhere, right?

server/src/main/resources/db/changelog/changelog-testing.xml
```diff
@@ -2028,7 +2028,7 @@
     </changeSet>
     <changeSet author="awood" id="1413225753032-290">
         <comment>
-            Add the default quartz lock columns.
+             Add the default quartz lock columns.
```
awood: I think this is a spurious change.  I've seen it pop up when I run the liquibase task though so I don't have a problem leaving it in.

server/src/main/resources/db/changelog/20180430101420-add-intent-fields-to-consumer.xml
```diff
@@ -7,14 +7,61 @@
...
+    <changeSet id="20180430101420-2" author="vrjain">
+        <comment> Add system purpose usage column to cp_consumer</comment>
+        <addColumn tableName="cp_consumer">
+            <column name="sp_usage" type="VARCHAR(255)"/>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="20180430101420-3" author="mstead">
+        <comment>Add system-purpose status column to cp_consumer.</comment>
+        <addColumn tableName="cp_consumer">
```
barnabycourt: Due to migration time, if each of the column adds in a single table is bundled in a single changeset it would likely update faster.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -1033,6 +1092,12 @@ public ConsumerDTO clone() {
         copy.setCapabilities(this.getCapabilities());
         copy.setContentTags(this.getContentTags());
 
+        copy.setServiceLevel(this.getServiceLevel());
+        copy.setRole(this.getRole());
+        copy.setUsage(this.getUsage());
+        copy.setAddOns(this.getAddOns());
+        copy.setSystemPurposeStatus(this.getSystemPurposeStatus());
```
nikosmoum: No need to explicitly assign Strings like role, usage, servicelevel, systempurposestatus etc. in the clone method, since they're already on the copy when `ConsumerDTO copy = super.clone();` is called, and they're immutable, so they can be safely shared between the original and the cloned objects. Only setAddons() is required here because its a collection.

server/src/test/java/org/candlepin/dto/api/v1/ConsumerDTOTest.java
```diff
@@ -94,7 +99,9 @@ public ConsumerDTOTest() {
         this.values.put("EntitlementStatus", "test-entitlement-status");
         this.values.put("ServiceLevel", "test-service-level");
         this.values.put("Usage", "test-usage");
-        this.values.put("Offering", "test-offering");
+        this.values.put("Role", "test-role");
+        this.values.put("SystemPurposeStatus", "matched");
+        this.values.put("AddOns", addOns);
```
nikosmoum: Need to similarly update ConsumerTranslatorTest.java to set and verify the 4 new fields there too (unfortunately the test framework doesn't fail automatically if you don't set the new fields manually on the test, so its something we have to remember all the time).

server/client/ruby/candlepin_api.rb
mstead: Ya, it appears that offering has now become 'role'.

mstead:
> Found a couple of minor DTO changes needed. Also have a question: I'm fairly certain we don't need to be updating the Rules' ConsumerDTO, but do we need to add these fields to the manifest ConsumerDTO? (is there a reason for them to be exported?)

As I understand, all selection logic will be outside of the rules.js, so the rules DTO update shouldn't be required.

Since the consumer that is added to the manifest is a distributor, the system purpose attributes will not be set on them. IMO there's no reason to update the manifest DTO.

@awood @vritant @kahowell  can you confirm my above statements?

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
mstead: Done. Will be included in next update of this PR. Thanks!

server/src/test/java/org/candlepin/dto/api/v1/ConsumerDTOTest.java
mstead: Done. Will be included in next update of this PR. Thanks!

server/spec/consumer_resource_spec.rb
mstead: Done. Will be included in next update of this PR. Thanks!

nikosmoum:
@mstead From my discussion with vritant today, I understand that the selection (filtering & prioritization of pools) is going to continue to be in the rules.js, while the syspurpose status calculation will happen in Java. (@vritant correct me if I'm wrong!)

server/src/main/java/org/candlepin/model/Consumer.java
mstead: @awood Is the compliance status now murmur3 as well? If so, I'll update the comment for complianceStatusHash at the same time.

mstead: Looks like it is. I'll change the comment there as well.

mstead: Done. Will be included in next update of this PR. Thanks!

server/src/main/resources/db/changelog/changelog-testing.xml
mstead: Let's leave it as is so that it doesn't keep showing up as a change.

mstead:
> @mstead From my discussion with vritant today, I understand that the selection (filtering & prioritization of pools) is going to continue to be in the rules.js, while the syspurpose status calculation will happen in Java. (@vritant correct me if I'm wrong!)

If that's the case then I think we need the rules ConsumerDTO will need to be updated.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
mstead: I don't think that we can do that since the collection is passed to the SetView and requires the ability to remove elements. As per the contract defined by the DTO framework [here](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/dto/CandlepinDTO.java#L32), we have to return views.

There's a card to do some work on the views so perhaps we can address your concern then as this is done in other places as well. I'd also like to chat about this with @Ceiu when he's back. Thoughts?

server/src/main/resources/db/changelog/20180430101420-add-intent-fields-to-consumer.xml
mstead: I looked into this some last night. If the DB has autocommit enabled, then it doesn't matter if the operations are bundled into a single changeset. For this reason, liquibase recommends separating them since it can leave the DB in a bad state if something fails -- [See these docs](https://www.liquibase.org/documentation/changeset.html). I believe that MySQL has autocommit enabled by default.

Given this info, do we still want to bundle them?

server/src/main/resources/db/changelog/20180430101420-add-intent-fields-to-consumer.xml
mstead: Correct. That's why I reused the existing instead of creating yet another.

mstead:
> @mstead From my discussion with vritant today, I understand that the selection (filtering & prioritization of pools) is going to continue to be in the rules.js, while the syspurpose status calculation will happen in Java. (@vritant correct me if I'm wrong!)
> 
> If that's the case then I think we need the rules ConsumerDTO will need to be updated.

Done.

mstead:
retest this please

PR #2058, opened 2018-07-23
1598850: Hypervisor checkins now return HypervisorUpdateResultDTO ENT-709

nikosmoum:
- Created HypervisorUpdateResultDTO, HypervisorConsumerDTO & HypervisorConsumerTranslator.
- Both sync & async checkins now return a HypervisorUpdateResultDTO, which contains the data:
'name'
'uuid'
'owner' -> 'key'


nikosmoum:
Fixed unit tests & added some extra assertions in virt checkin spec tests.

PR #2063, opened 2018-07-27
[2.4] 1608426: API call to get list of consumer UUIDs entitled from a pool [ENT-740]

awood:
Also delete pointless override of find method in PoolCurator.


wottop:
ack

PR #2066, opened 2018-07-30
Fix issue with intermittent rspec test failure.

awood:
This test was originally just scheduling a job and then attempting to
cancel it.  Unfortunately, if the job ran fast enough, it would be
complete before the cancellation attempt landed and Candlepin would
return a 400 error.  This patch fixes the issue by pausing the
scheduler, scheduling a job, and then attempts to cancel it.  The patch
is based on a similar test elsewhere in the spec file.


wottop:
ack

PR #2067, opened 2018-08-06
[M] Adapter Refactor Work

Ceiu:
Several updates and changes to the service adapters

- Several methods in the service adapters have been changed to
  avoid passing full objects through where an object ID or key
  will suffice
- Added the new model subpackage for the service adapters. The
  new service model interfaces provide an interface any object
  passing through the adapter must implement; allowing a complete
  separation from the API or DTO models for data hitting the
  adapters
- Updated the RoleResource to use DTOs for its input and output
- Codified the expected behavior of role and user update objects
  when presented with subobjects
- Codified the expected behavior of the User.getPermissions
  method
- Renamed all remaining instances of get/setMetadataExpire to
  get/setMetadataExpiration
- Updated the KatelloUserServiceAdapter and DefaultUserServiceAdapter
  in accordance with the UserServiceAdapter changes
- Made various corrections to code which were passing objects into
  the adapters where it now expects string-based keys or IDs
- Refactored the PermissionFactory class to work with the service
  adapter model
- Updated the UserAuth class work with the the service adapter model
- Updated the hosted test infrastructure to use the new service
  model
- Refactored the hosted test subscription adapter and resource to
  keep its own copies of products and content and no longer create
  downstream/local copies instead
- Removed the requirement for a local instance of a product/content
  before updating or linking them from the hosted test resource
- Added the ImportedEntityCompiler which gathers entities from various
  sources and compiles them into single, de-duplicated, validated
  collections for easier importing
- Updated the Entitler and CandlepinPoolManager to use the new
  ImportedEntityCompiler
- Updated several classes to reference the service adapter model
  when interacting with data originating from the adapters
- Refactored the PermissionFactory a bit; it should now be injected
  rather than directly instantiated, and will perform a temporary
  form of entity resolution internally
- User no longer provides permissions directly, nor does it provide
  the getAccessibleOwners utility method, as this required a lot of
  unnecessary churn to function, wasn't widely used and belongs
  elsewhere
- Removed the getId methods from the service model interfaces, as
  external data sources should not be expected to have our internal
  DB IDs for the model objects
- Changed Consumer.getOwner to lazily pull the owner using the
  owner ID field rather than using a transient reference
- Updated all tests in accordance with the changes above
- Fixed several failing tests in accordance with the changes above
- Updated standalone tests to work with the new service model
  and surrounding changes
- Added new translators to convert unknown service model objects
  to API DTOs
- Updated spec tests in accordance with the adapter changes
- Extended the API operations allowed by the hosted test
  endpoints
- Changed how RoleDTO manages its users and permissions
  internally
- Changed how Consumer references owner and owner ID

### DIFF
--- a/server/client/ruby/hostedtest_api.rb
+++ b/server/client/ruby/hostedtest_api.rb
@@ -3,10 +3,14 @@ module HostedTest
   @@hosted_mode = nil
   @@hostedtest_alive = nil
 
+  # FIXME: This is broken. There's no distiction between up and downstream data sources, which creates
+  # a lot of problems when it comes to product and content mapping. Hosted test resources should be
+  # entirely upstream, and not rely on, nor require, anything downstream for proper functionality.
+
   def is_hostedtest_alive?
     if @@hostedtest_alive.nil?
       begin
-        @@hostedtest_alive = @cp.get('/hostedtest/subscriptions/is_alive', {}, 'json', true)
+        @@hostedtest_alive = @cp.get('/hostedtest/alive', {}, 'text/plain', true)
       rescue RestClient::ResourceNotFound
         @@hostedttest_alive = false
       end
@@ -15,7 +19,6 @@ module HostedTest
   end
 
   def create_hostedtest_subscription(owner_key, product_id, quantity=1, params={})
-
     provided_products = params[:provided_products] || []
     start_date = params[:start_date] || DateTime.now
     end_date = params[:end_date] || start_date + 365
@@ -63,7 +66,8 @@ module HostedTest
   end
 
   def update_hostedtest_subscription(subscription)
-    return @cp.put("/hostedtest/subscriptions", {}, subscription)
+    id = subscription.id
+    return @cp.put("/hostedtest/subscriptions/#{id}", {}, subscription)
   end
 
   def get_all_hostedtest_subscriptions()
@@ -78,8 +82,8 @@ module HostedTest
     return @cp.delete("/hostedtest/subscriptions/#{id}", {}, nil, true)
   end
 
-  def delete_all_hostedtest_subscriptions()
-    @cp.delete('/hostedtest/subscriptions/', {}, nil, true)
+  def clear_upstream_data()
+    @cp.delete('/hostedtest', {}, nil, true)
   end
 
   def is_hosted?
@@ -109,7 +113,7 @@ module HostedTest
       content_ids.each do |id|
         data[id] = enabled
       end
-      @cp.post("/hostedtest/subscriptions/owners/#{owner_key}/products/#{product_id}/batch_content", {}, data)
+      @cp.post("/hostedtest/products/#{product_id}/batch_content", {}, data)
     else
       @cp.add_batch_content_to_product(owner_key, product_id, content_ids, true)
     end
@@ -117,7 +121,7 @@ module HostedTest
 
   def add_content_to_product(owner_key, product_id, content_id, enabled=true)
     if is_hosted?
-      @cp.post("/hostedtest/subscriptions/owners/#{owner_key}/products/#{product_id}/content/#{content_id}", {:enabled => enabled})
+      @cp.post("/hostedtest/products/#{product_id}/content/#{content_id}", {:enabled => enabled})
     else
       @cp.add_content_to_product(owner_key, product_id, content_id, true)
     end
@@ -125,16 +129,19 @@ module HostedTest
 
   def update_product(owner_key, product_id, params={})
     if is_hosted?
+      # FIXME: This is broken
+
       product = {
         :id => product_id
       }
+
       product[:name] = params[:name] if params[:name]
       product[:multiplier] = params[:multiplier] if params[:multiplier]
       product[:attributes] = params[:attributes] if params[:attributes]
       product[:dependentProductIds] = params[:dependentProductIds] if params[:dependentProductIds]
       product[:relies_on] = params[:relies_on] if params[:relies_on]
 
-      @cp.put("/hostedtest/subscriptions/owners/#{owner_key}/products/#{product_id}", {}, product)
+      @cp.put("/hostedtest/products/#{product_id}", {}, product)
     else
       @cp.update_product(owner_key, product_id, params)
     end
@@ -155,6 +162,7 @@ module HostedTest
     params[:order_number] = order_number
     params[:quantity] = quantity
     params[:provided_products] = provided_products
+
     pool = nil
     if is_hosted?
       ensure_hostedtest_resource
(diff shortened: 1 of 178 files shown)


kahowell:
ok to test

server/src/main/java/org/candlepin/hostedtest/HostedTestSubscriptionServiceAdapter.java
```diff
@@ -38,146 +60,961 @@
...
+
+        return this.productMap.get(productId);
+    }
+
+    public ContentInfo createContent(ContentInfo cinfo) {
+        if (cinfo == null) {
+            throw new IllegalArgumentException("cinfo is null");
+        }
+
+        if (cinfo.getId() == null || cinfo.getId().matches("\\A\\s*\\z")) {
```
awood: `StringUtils.isBlank`?

server/src/main/java/org/candlepin/hostedtest/HostedTestSubscriptionServiceAdapter.java
```diff
@@ -38,146 +60,961 @@
...
-        idMap.put(s.getId(), s);
-        if (!ownerMap.containsKey(s.getOwner().getKey())) {
-            ownerMap.put(s.getOwner().getKey(), new ArrayList<>());
+
+    protected void updateSubscriptionProductMappings(Subscription sdata) {
+        if (sdata == null) {
+            throw new IllegalArgumentException("sdata is null");
+        }
+
+        if (sdata.getId() == null || sdata.getId().matches("\\A\\s*\\z")) {
```
awood: `StringUtils.isBlank`?

server/src/main/java/org/candlepin/hostedtest/HostedTestSubscriptionServiceAdapter.java
```diff
@@ -38,146 +60,961 @@
...
+            }
+        }
+    }
+
+    protected void updateProductContentMappings(ProductData pdata) {
+        if (pdata == null) {
+            throw new IllegalArgumentException("pdata is null");
+        }
+
+        if (pdata.getId() == null || pdata.getId().matches("\\A\\s*\\z")) {
```
awood: `StringUtils.isBlank`?

server/src/main/java/org/candlepin/model/CdnCurator.java
```diff
@@ -35,6 +35,8 @@ public CdnCurator() {
      * @return CDN whose label matches the given value.
      */
     public Cdn getByLabel(String label) {
+        // TODO: This is dangerous. We're expecting a unique result, but there is no guarantee the
```
awood: I think `uniqueResult` will throw an exception if it's not unique (both in [Hibernate](https://docs.jboss.org/hibernate/orm/5.1/javadocs/org/hibernate/Query.html#uniqueResult--) and in [JPA's version](https://docs.oracle.com/javaee/7/api/javax/persistence/Query.html#getSingleResult--)).  So we can catch the exception and do something or just let a runtime error get thrown.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -387,20 +388,48 @@ public String getOwnerId() {
...
-        }
+    @Override
+    @JsonIgnore
+    @Formula("SELECT o FROM Owner WHERE o.id = owner_id")
+    @Basic(fetch = FetchType.LAZY)
+    public Owner getOwner() {
+        return null;
     }
 
+    // /**
```
awood: Delete this?  Seems obsolete now.

server/src/main/java/org/candlepin/model/User.java
```diff
@@ -219,19 +206,21 @@ public String getPassword() {
...
-            sub = SubResource.NONE;
-        }
-        for (Permission p : this.getPermissions()) {
-            if (p.canAccess(p.getOwner(), sub, accessLevel)) {
-                owners.add(p.getOwner());
-            }
-        }
-        return owners;
-    }
+    // @XmlTransient
```
awood: Obsolete?

server/src/main/java/org/candlepin/model/User.java
```diff
@@ -254,32 +243,24 @@ public void removeRole(Role r) {
...
+                role.removeUser(this);
+            }
+        }
     }
 
     /**
      * @return if the user has the SUPER_ADMIN role
      */
-    public boolean isSuperAdmin() {
+    public Boolean isSuperAdmin() {
```
awood: Unless "isSuperAdmin" returning null is meaningful, I vote that we do not change this.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -784,7 +784,14 @@ public Consumer createConsumerFromDTO(ConsumerDTO consumer, ConsumerType type, P
         }
 
         Consumer consumerToCreate = new Consumer();
-        consumerToCreate.setOwner(owner);
+
+
+        log.debug("SETTING OWNER ID FROM OWNER: {} => {}", owner, owner.getId());
```
awood: Can we either delete these or un-all-caps them?

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1146,7 +1153,7 @@ private void verifyPersonConsumer(ConsumerDTO consumer, ConsumerType type, Owner
         }
 
         if (user == null) {
-            throw new NotFoundException(i18n.tr("User with ID \"{0}\" could not be found."));
+            throw new NotFoundException(i18n.tr(this.i18n.tr("User not found: {0}", username)));
```
awood: Why the nested calls to `i18n.tr` now?

server/src/main/java/org/candlepin/resource/RoleResource.java
```diff
@@ -99,150 +164,124 @@ public Role createRole(@ApiParam(name = "role", required = true) Role role) {
...
+        @ApiParam(name = "permissionBlueprint", required = true) PermissionBlueprintDTO permission) {
 
-        Role existingRole = lookupRole(roleId);
+        // Validate role name
+        this.fetchRoleByName(roleName);
 
         // Don't allow NONE permissions to be created, this is currently just for
         // internal use:
-        if (permission.getAccess().equals(Access.NONE)) {
+        if (Access.NONE.name().equals(permission.getAccess())) {
```
awood: :+1:

server/src/main/java/org/candlepin/auth/permissions/PermissionFactory.java
```diff
@@ -37,45 +51,226 @@
...
+     */
+    @Deprecated
+    protected static class Resolver {
+
+        protected OwnerCurator ownerCurator;
+        protected Map<String, Owner> ownerCache;
+
+        public Resolver(OwnerCurator ownerCurator) {
+            this.ownerCurator = ownerCurator;
+            this.ownerCache = new HashMap<>();
```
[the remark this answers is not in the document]
Ceiu: The class definition is static, but the instances aren't, and they're short-lived. Each resolver should only live for a single request and shouldn't have multiple threads accessing it, so we should be safe on both accounts here.

server/src/main/java/org/candlepin/auth/permissions/PermissionFactory.java
```diff
@@ -37,45 +51,226 @@
...
+                Owner owner = this.ownerCache.get(oinfo.getKey());
+
+                if (owner == null) {
+                    owner = this.ownerCurator.getByKey(oinfo.getKey());
+
+                    if (owner == null) {
+                        throw new IllegalStateException("No such owner: " + oinfo.getKey());
+                    }
+
+                    ownerCache.put(owner.getKey(), owner);
```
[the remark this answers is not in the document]
Ceiu: We should be safe here. The lifetime of this object will isolate us in all but cases where the permission objects detach the entity during construction, which shouldn't ever happen.

server/src/main/java/org/candlepin/model/User.java
Ceiu: This is a bit of an odd one. If we want to make life easy on things going to and from the adapter layer, we need to implement UserInfo. While we could make UserInfo's definition kick out a primitive, our UserDTO would then need extra work to get around it. 

Also, we need some sentinel value to represent "no change," which is currently null in most other cases.

server/client/ruby/hostedtest_api.rb
```diff
@@ -275,4 +283,204 @@ def cleanup_subscriptions
...
+    subscription = {
+      :startDate => start_date,
+      :endDate   => end_date,
+      :product =>  { :id => product_id },
+      :owner =>  { :key => owner_key },
+      :quantity => 1
+    }
+
+    # Merge, but convert some snake-case keys to camel case
+    keys = [:account_number, :contract_number, :order_number, :upstream_pool_id,
```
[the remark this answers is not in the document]
Ceiu: Fixed

server/src/main/java/org/candlepin/auth/UserAuth.java
```diff
@@ -14,46 +14,53 @@
...
         }
-        else {
-            Principal principal = new UserPrincipal(username, user.getPermissions(), false);
 
-            return principal;
-        }
+        // TODO: This creates a lot of object churn. We should probably update this later in a way
+        // that can do permission checking without creating piles of objects that we just throw away
+        // without ever using them in the general case.
+        return user.isSuperAdmin() != null && user.isSuperAdmin() ?
```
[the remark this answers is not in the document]
Ceiu: Seems my comment got lost here...

This one's a bit tough. It's a Boolean to play nice with the service model interface and any type of object we might send in our out of the adapters. We could change it back to a primitive, but then the service models and DTOs have no way of specifying the "no change" state (which we typically use null for, at present).

I'm open to suggestions here, but the driving functionality is the need to be able to indicate a value isn't changing.

server/src/main/java/org/candlepin/auth/permissions/PermissionFactory.java
```diff
@@ -37,45 +51,226 @@
...
+            return null;
+        }
+    }
+
+
+    /**
+     * Micro-interface used to create simple permission builders from a user and a permission
+     * blueprint.
+     */
+    public interface PermissionBuilder {
```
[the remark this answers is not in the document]
Ceiu: Done

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -586,6 +527,8 @@ public void sendEvents(List<Entitlement> entitlements) {
         public DeveloperProducts(String expectedSku, Map<String, Product> products) {
             this.sku = products.remove(expectedSku);
             this.provided = products;
+
+            log.debug("Test?? {}", this.sku);
```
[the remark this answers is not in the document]
Ceiu: Removed

server/src/main/java/org/candlepin/dto/api/v1/PermissionBlueprintInfoTranslator.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public PermissionBlueprintDTO translate(ModelTranslator translator, PermissionBlueprintInfo source) {
+        return source != null ? this.populate(translator, source, new PermissionBlueprintDTO()) : null;
+    }
+
+    /**
+     * {@inheritDoc}
```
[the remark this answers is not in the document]
Ceiu: Oh it does now? Good to know. I'll probably still be in the habit of adding them in these cases to signify that the documentation wasn't forgotten, but I won't feel as bad if I forget them now.

server/src/main/java/org/candlepin/dto/api/v1/RoleDTO.java
```diff
@@ -308,7 +326,19 @@ public boolean removePermission(String permissionId) {
...
 
-        return this.permissions != null && this.permissions.remove(permissionId) != null;
+        if (this.permissions != null) {
+            Iterator<PermissionBlueprintDTO> iterator = this.permissions.iterator();
+
+            while (iterator.hasNext()) {
+                PermissionBlueprintDTO existing = iterator.next();
+
+                if (permissionId.equals(existing.getId())) {
+                    iterator.remove();
```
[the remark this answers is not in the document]
Ceiu: Fixed

server/src/main/java/org/candlepin/hostedtest/HostedTestSubscriptionResource.java
```diff
@@ -109,28 +113,126 @@ public Boolean isAlive() {
...
+            }
+            else {
+                this.adapter.createProduct(product);
+            }
+        }
+    }
+
+    private void addProductsToMap(ProductData product, Map<String, ProductData> pmap) {
+        if (product != null) {
+            if (product.getId() == null || product.getId().matches("\\A\\s*\\z")) {
```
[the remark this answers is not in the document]
Ceiu: Fixed

server/src/main/java/org/candlepin/model/CdnCurator.java
Ceiu: We can, but the problem isn't the lack of exception handling, it's that we expect a non-unique field to be unique. If someone's deployment ends up with two Cdns with the same label, they're going to have issues regardless of whether or not we catch that exception. This is something that we should probably fix by adding a DB restriction, or allowing more than a single object to be output from this method.

server/src/main/java/org/candlepin/model/User.java
Ceiu: Yup. Removed.

server/src/main/java/org/candlepin/model/Consumer.java
Ceiu: Oh yeah. I need to verify that the formula call works as I expect it to and add tests around it.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
Ceiu: Removed

server/src/main/java/org/candlepin/resource/ConsumerResource.java
Ceiu: Refactoring artifact, I'd guess. Fixed.

server/src/main/java/org/candlepin/auth/permissions/PermissionFactory.java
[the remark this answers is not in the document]
awood: For some reason I thought that the PermissionFactory singleton was holding a reference to a Resolver but that's not the case.  Please disregard.

awood:
I think we should get another couple of reviewers on this.  @mstead, @wottop, @kahowell, @cnsnyder ?

server/src/main/java/org/candlepin/dto/StandardTranslator.java
```diff
@@ -188,12 +201,16 @@ public StandardTranslator(ConsumerTypeCurator consumerTypeCurator,
             new ProductCertificateTranslator(), ProductCertificate.class, ProductCertificateDTO.class);
         this.registerTranslator(
             new RoleTranslator(), Role.class, RoleDTO.class);
+        this.registerTranslator(
+            new RoleInfoTranslator(), RoleInfo.class, RoleDTO.class);
```
nikosmoum: I'm a bit unsure how the *Info classes are going to be translated. Surely when the DTO framework is calling `getClass()` on the input, it will always resolve to `Role` (because that will be the concrete class of the instance at runtime), and not `RoleInfo` (which is just an interface), resulting in always using the `RoleTranslator` and never the 'RoleInfoTranslator'. So the `*InfoTranslator` classes are never going to be actually used. Unless I'm missing something (I might, because I'm not familiar with the adapters at all).

server/src/main/java/org/candlepin/resource/RoleResource.java
```diff
@@ -99,150 +164,124 @@ public Role createRole(@ApiParam(name = "role", required = true) Role role) {
         " permissions.", value = "updateRole")
     @ApiResponses({ @ApiResponse(code = 404, message = "") })
     @PUT
-    @Path("{role_id}")
+    @Path("{role_name}")
```
nikosmoum: Are we free to change the API here? (from role_id to role_name)

Ceiu: Yup, no one uses it (yet), so we're free to change it to be more in-line with the user APIs and other such objects which pass through the adapters (and, thus, can't be expected to know about our internal DB IDs).

server/src/main/java/org/candlepin/dto/StandardTranslator.java
Ceiu: You're correct in the one-off translation case where we provide the instance and the destination class, but where this comes into play is when calling the getStreamMappers method, which requires the source class as a parameter.

I suppose this is a use case for adding an overload to .translate which accepts a source class for doing explicit translation. But that can come in another PR that could also address a bug in the lookup for object translators (both input and output scan "up" the hierarchy, but output should scan "down" or we break).

PR #2073, opened 2018-08-16
[2.4] 1618391: Manifest import no longer fails when product changes ENT-787

kahowell:
Cherry-picked and then resolved conflicts.

* When a product changes in a manfiest resulting in entitlement revocation and subsequent deletion of entitlement derived pools,
  some pools that are deleted subsequently were attempted to being updated. This branch skips those deleted pools
* Some product changes resulted in creating brandings on a product, and when that pool was being locked, created an error since
  Hibernate had not flushed the brandings yet


kahowell:
Originally from #1976

PR #2074, opened 2018-08-16
Remove use of parallelStream when not necessary.

awood:
Using parallelStream incurs a lot of overhead around thread
synchronization.  In this case, these collections are likely to be on
the order of a hundred or so elements and the actual operation itself is
going to be very fast.  The additional overhead of parallelStream will
end up degrading performance.  If we expect these lists of compliant
products, partially compliant products, and partial stacks to get very
large, parallelizing this would be worth reconsidering.

See https://stackoverflow.com/a/20375622/6124862 and
https://www.ibm.com/developerworks/library/j-java-streams-5-brian-goetz/index.html
(specifically the "NQ model" section)


Ceiu:
ack

PR #2079, opened 2018-08-17
1616408: selinux - allow tomcat to connect to postgres ENT-791

kahowell:
Recent/upcoming selinux-policy disallows this by default.


wottop:
retest this please

PR #2085, opened 2018-08-21
ENT-109 Implemented ActivationKeyDTO for Rules

nikosmoum:


server/src/main/java/org/candlepin/dto/rules/v1/ActivationKeyDTO.java
```diff
@@ -0,0 +1,415 @@
...
+                .append(this.getQuantity());
+
+            return builder.toHashCode();
+        }
+    }
+
+    /**
+     * Internal DTO that represents a Pool object on an ActivationKeyPool object.
+     */
+    public static class InternalPoolDTO extends CandlepinDTO<InternalPoolDTO> {
```
Ceiu: Not sure how much of a fan I am of this class. I assume the intent here is to reduce the amount of data we have to serialize, but it starts to get away from the standardization the DTOs are trying to fix. I'll need to think on this a bit.

nikosmoum: Yes, the intent is to minimize the data, since there already is a PoolDTO for rules, but it has way more fields than we actually need (the existing PoolDTO has 12 fields, while in this case we only use 3 of them). I also don't like the way this is written, and struggled but didn't find a better way.

Ceiu: Since the rules are effectively on their deathbed anyway, we can go ahead and merge this as-is, I suppose. Since we're in need of a rebase anyway, could you add a note or some such on it to indicate that it's a non-standard solution to this problem and one we shouldn't get in the habit of doing?

nikosmoum: Done

Ceiu:
ack

PR #2095, opened 2018-08-23
[2.3] Updated gen-certs for compatibility with Tomcat 8.5+

Ceiu:
- gen-certs will now add an explicit entry for the candlepin_ca
  cert for compatibility with Tomcat 8.5+ when using the same
  file as both a keystore and trust store


wottop:
ack

PR #2096, opened 2018-08-23
Newer versions of mariadb don't recognize dbms="mysql"

wottop:


Ceiu:
Rather than go through and list each one individually like this, do we want to define a type that covers both ("mysql_compat"? "mysql_like"?)? I feel like in the vast majority of cases, we're going to be compatible with both for the foreseeable future.

PR #2099, opened 2018-08-28
Correct over-zealous search and replace.

awood:


wottop:
ack, my bad. I corrected this in the create by missed the update

PR #2100, opened 2018-08-30
ENT-113 Replace @JsonIgnoreProperties with ObjectMapper config:

nikosmoum:
- Removed all @JsonIgnoreProperties(ignoreUnknown) annotations from DTOs/models.
- Configured the API json provider's ObjectMapper to not fail during deserialization when unknown properties are encountered.


Ceiu:
There are a handful of other object mappers which were not configured to ignore unknown properties. Some of them are only used for serialization, so those are fine, but there are a couple (rules and sync) which probably should be closer evaluated.

Ceiu:
ack

PR #2101, opened 2018-08-31
[M] Removed the v1 DTO shims from manifest import and export (ENT-334)

Ceiu:
- Added the new SubscriptionDTO object to act as a full
  representation of a pool during manifest import
- EntitlementImporter no longer uses the Subscription class,
  nor attempts to validate its products, entitlements and
  other DTOs against existing model entities
- Optimized out some unnecessary sorting and collection
  conversion in SubscriptionReconciler
- Updated several tests to use the new manifest DTOs rather
  than the old DTO shims


Ceiu:
Done. Also cleaned up some of the field duplication with updated and lastmodified.

PR #2105, opened 2018-09-04
Remove Oracle related code.

awood:
Oracle support is no longer mandated.


awood:
@nikosmoum I honestly can't remember why I marked this as in progress.  I think because I thought I needed to get some stuff merged for the CI environment first.  But it turns out that stuff had already been merged, so this should be good to go (or good for review at least!)

server/code/schema/README.txt
```diff
@@ -1,3 +0,0 @@
-This directory contains third party schema.
```
nikosmoum: Looks like the whole 'schema' directory got removed (as opposed to just the oracle script). I'm unable to see how/if they're currently used. Do we not need these scripts any more?

awood: Yes, they are unused.  Everything is in liquibase now.

PR #2110, opened 2018-09-05
ENT-798 Immutable Products & Hibernate L2 Cache support

barnabycourt:
Better enable Hibernate L2 Cache support by properly annotating Products & Content as immutable. 

Also change some of the product hydration & caching to depend more on the Hibernate L2 cache instead of Candlepin specific product caches.


server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -168,24 +158,38 @@ public Product getByName(Owner owner, String name) {
...
+            else {
+                productsNotInCache.add(uuid);
+            }
+        }
+        if (productsNotInCache.size() > 0) {
+            log.debug("Loading objects that were not already in the cache: " + productsNotInCache.size());
+            Session session = this.currentSession();
+            java.util.List entities = session.byMultipleIds(this.entityType())
+                .enableSessionCheck(true)
+                .multiLoad(new ArrayList(productsNotInCache));
```
Ceiu: It bugs me that Hibernate's object here doesn't accept a generic collection, but at least they provide a variadic alternative. It's slightly faster here to use ```productsNotInCache.toArray()``` rather than building another collection.

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -168,24 +158,38 @@ public Product getByName(Owner owner, String name) {
...
+            products.addAll(entities);
+        }
 
-        productCache.putAll(freshProducts);
-        products.addAll(freshProducts.values());
+        // Hydrate all the objects fully this is because a lot of serialization happens outside of
+        // the transactional boundry when we do not have a valid session.
```
Ceiu: :+1:

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -228,20 +232,20 @@ public Product getByName(Owner owner, String name) {
...
+            Map<String, Set<String>> poolProvidedProducts = new HashMap<>();
+            Map<String, Set<String>> poolDerivedProvidedProducts = new HashMap<>();
+            Map<String, Product> allProducts = new HashMap<>();
 
             Set<String> poolIds = new HashSet<>();
-            Set<String> productUuids = new HashSet<>();
 
             for (Pool pool : pools) {
                 poolIds.add(pool.getId());
+                poolProvidedProducts.put(pool.getId(), new HashSet<>());
```
Ceiu: If we're pre-populating the pool->product maps, then the poolIds set is unnecessary, since we can just use the keyset of either map in its place.

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -261,20 +265,14 @@ public int hydratePoolProvidedProducts(Iterable<Pool> pools) {
                     String poolId = (String) row[0];
                     String productUuid = (String) row[1];
 
-                    Set<Object> ppSet = poolProvidedProducts.get(poolId);
+                    Set<String> ppSet = poolProvidedProducts.get(poolId);
                     if (ppSet == null) {
```
Ceiu: This block is redundant if we're pre-populating the pool->product map.

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -283,80 +281,38 @@ public int hydratePoolProvidedProducts(Iterable<Pool> pools) {
                     String poolId = (String) row[0];
                     String productUuid = (String) row[1];
 
-                    Set<Object> dpSet = poolDerivedProvidedProducts.get(poolId);
+                    Set<String> dpSet = poolDerivedProvidedProducts.get(poolId);
                     if (dpSet == null) {
```
Ceiu: This block is redundant if we're pre-populating the pool->product map.

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -398,7 +354,15 @@ public Product create(Product entity) {
 
         this.validateProductReferences(entity);
 
-        return super.create(entity);
+        Product newProduct = super.create(entity, false);
+
+        for (ProductContent productContent: entity.getProductContent()) {
```
Ceiu: Missing space before the colon

server/src/main/java/org/candlepin/model/ProductCurator.java
Ceiu: Only this remains. Other changes look good.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -308,7 +307,6 @@ public ProductDTO addBatchContent(
             throw new ForbiddenException(i18n.tr("product \"{0}\" is locked", product.getId()));
         }
 
-        this.productCurator.lock(product);
```
Ceiu: I'm a little concerned about removing these locks here. Admittedly, having the resource lock the product is a bit of a hack anyway, so something should be done to improve this; but the scenario where two update requests for the same product at the same time worries me. Feels like we'll end up with a typical race condition where both succeed but one update is clobbered.

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -168,24 +158,38 @@ public Product getByName(Owner owner, String name) {
...
-        products.addAll(productsFromCache.values());
-
-        notInCache.addAll(productUuids);
-        notInCache.removeAll(productsFromCache.keySet());
-
-        //Now find, hydrate and cache all the products that has been
-        //missing in the cache
-        Map<String, Product> freshProducts = this.getHydratedProductsByUuid(notInCache);
+        Set<String> productsNotInCache = new HashSet<>();
+        for (String uuid :productUuids) {
```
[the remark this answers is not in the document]
Ceiu: This seems to have slipped back in. Can fix during squash.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -373,7 +386,6 @@ public ProductDTO removeBatchContent(
...
         ProductDTO pdto = this.translator.translate(product, ProductDTO.class);
```
Ceiu: Missed a lock here

Ceiu:
ack

PR #2113, opened 2018-09-13
[M] 1628594: Update epoch on Java requirement [ ENT-881 ]

kahowell:
See note about epoch at https://fedora-java.github.io/howto/latest/#_jvm


wottop:
ack

PR #2115, opened 2018-09-14
Ignore unknown JSON properties in checkin data

mstead:
Because we are no longer ignoring unknown properties
in the ConsumerDTO via annotations, async checkins will
fail due to additional properties sent by virt-who.

This PR creates a custom ObjectMapper in the update
job so that it can be customized to ignore unknown
properties when processing the JSON.

###Testing
Check out and deploy master. Then run:
```bash
curl -X POST -k -u admin:admin -d '{"hypervisors": [{"hypervisorId": {"hypervisorId": "fe16edca-cd38-4f98-bb8f-e322d9718dad"}, "guestIds": [{"guestId": "b9cedf43-6554-4ed9-9faf-97fec58bb016", "state": 5}]}]}' -H "Content-Type: text/plain" https://localhost:8443/candlepin/hypervisors/admin
```
The job should fail.

Check out this branch and deploy. Then run the same command and it should pass.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
+++ b/server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
@@ -16,6 +16,8 @@ package org.candlepin.pinsetter.tasks;
 
 import static org.quartz.JobBuilder.*;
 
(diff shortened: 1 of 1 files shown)


wottop:
ack

PR #2119, opened 2018-09-21
[M] Updated the update-server-xml script for changes required in F28

Ceiu:
- Updated the update-server-xml.py script such that it generates
  non-deprecated server configuration which does not function
  when running Tomcat on Fedora 28
- update-server.xml.py now accepts a "--tomcat-version" option, to
  specify the version of Tomcat to target, changing the generated
  configuration accordingly
- Updated the is_different method to check attributes and children
- Attributes are now added to the node by the node editors themselves
  rather than the methods provided by the AbstractBaseEditor
- Deploy script updated in accordance with the above changes
- JBoss setup and configuration removed from the deploy script

### DIFF
--- a/server/bin/deploy
+++ b/server/bin/deploy
@@ -211,54 +211,47 @@ if [ $(id -u) -ne 0 ]; then
     fi
 fi
 
-if [ "$JBOSS_HOME" != "" ]; then
-    # we're using JBOSS AS
-    isrpm $JBOSS_HOME
+# use tomcat 6, unless it doesnt exist and tomcat does
+TC=tomcat6
+TC_VERSION=6
 
-    if [ "$?" -ne 0 ]; then
-        START="$JBOSS_HOME/bin/run.sh 2>&1 > console.log &"
-        STOP="$JBOSS_HOME/bin/shutdown.sh --shutdown"
-        DEPLOY="$JBOSS_HOME/server/default/deploy/"
-    else
-        START="$SUDO /sbin/service jbossas start"
-        STOP="$SUDO /sbin/service jbossas stop"
-        DEPLOY="$JBOSS_HOME/server/default/deploy/"
-    fi
+if ( [ -f /usr/sbin/tomcat ] && (! [ -f /usr/sbin/tomcat6 ] ) );
+then
+    TC=tomcat
 
-    CONTAINER_CONF_DIR="$JBOSS_HOME/conf"
-    CLEAN="$JBOSS_HOME/server/default/deploy/candlepin*"
-else
-    # use tomcat 6, unless it doesnt exist and tomcat does
-    TC=tomcat6
-    if ( [ -f /usr/sbin/tomcat ] && (! [ -f /usr/sbin/tomcat6 ] ) );
-    then
-        TC=tomcat
-    fi
-    if [ -z $TC_HOME ]; then
-        TC_HOME=/var/lib/$TC
-    fi
-
-    isrpm $TC_HOME
-
-    if [ "$?" -ne 0 ]; then
-        START="$TC_HOME/bin/catalina.sh jpda start"
-        STOP="$TC_HOME/bin/catalina.sh stop"
-        CONTAINER_CONF_DIR="$TC_HOME/conf"
-    else
-        START="$SUDO /sbin/service $TC start"
-        STOP="$SUDO /sbin/service $TC stop"
-        CONTAINER_CONF_DIR="/etc/$TC"
-    fi
-
-    # Check if we're in a container using supervisord instead:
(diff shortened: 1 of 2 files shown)


server/bin/update-server-xml.py
```diff
@@ -57,27 +57,74 @@ def _add_attributes(self, node, attributes):
...
+        if cnode.name != nnode.name:
+            logger.debug("Node names differ: %s != %s" % (cnode.name, nnode.name))
+            return True
+
+        # Compare attributes immediately present on both nodes...
+        cattribs = {}
+        nattribs = {}
+
+        if cnode.properties:
+            for prop in cnode.properties:
```
awood: Use a dictionary comprehension.  It's more pythonic and more succinct.

```
cattribs = {n.name: n.content for n in cnode.properties if n.type == "attribute"}
```

server/bin/update-server-xml.py
```diff
@@ -137,42 +189,73 @@ def new_node(self):
...
+            "TLS_ECDH_RSA_WITH_AES_256_CBC_SHA",
+            "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA",
         ])
 
-        # This is a list of tuples instead of a dict so we can preserve the attribute
-        # ordering.  OrderedDict didn't get added until 2.7.
-        return [
+        # Setup our node configuration
+        connector = libxml2.newNode("Connector")
+        self._add_attributes(connector, [
```
awood: We can probably actually use an OrderedDict now if we wanted.  Pretty much everything is 2.7+ these days.

server/bin/update-server-xml.py
Ceiu: How does that handle the case of cnode.properties being non-iterable?

awood: Yeah, sorry, I left out that the `if cnode.properties` needs to remain.

Ceiu:

server/bin/deploy
```diff
@@ -335,7 +328,7 @@ bash $SELF_DIR/gen-certs $GEN_CERTS_ARGS
 update_keystore
 
 # update server.xml
-$SUDO python $PROJECT_DIR/bin/update-server-xml.py $CONTAINER_CONF_DIR
+$SUDO python $PROJECT_DIR/bin/update-server-xml.py --version $TC_VERSION $CONTAINER_CONF_DIR
```
awood: Sorry if this is nit-picky, but can we use `--tomcat-version`?  `--version` is almost always "show me the version of this program" so using it as an actual functional option creates confusing semantics.

server/bin/update-server-xml.py
```diff
@@ -57,27 +57,70 @@ def _add_attributes(self, node, attributes):
...
     def _is_different(self, node):
-        current_attributes = {}
-        for property in node.properties:
-            if property.type == "attribute":
-                current_attributes[property.name] = property.content
-        return node.name != self.new_node.name or current_attributes != dict(self.attributes)
+        return self._is_different_impl(node, self.new_node)
 
     def _update(self, existing_nodes):
         different_nodes = filter(self._is_different, existing_nodes)
```
awood: Maybe instead of having a wrapper `_is_different` and `_is_different_impl`, we can just adjust this filter statement to

```
different_nodes = filter(lambda x: self._is_different(x, self.new_node), existing_nodes)
```

And then have the body of `_is_different` be replaced with the body of the current `_is_different_imple`.

server/bin/deploy
Ceiu: Done.

server/bin/update-server-xml.py
Ceiu: Doesn't that just offload the wrapping to the caller instead of the implementation?

awood: Yes, but I thought it was a little weird to have both `is_different` and `is_different_impl` methods.  Both times when reviewing, I got to `is_different` and sat scratching my head for a minute wondering "why is this even here?"

Ceiu: Hrmm... Would you be okay if I moved the impl to an inner function within is_different (assuming calling conventions allow it)?

Ceiu: Or, now that I think of it, a "static" or global function that handles the checking of two arbitrary nodes, and the method which calls into it to properly separate/name the two variations.

awood: > Or, now that I think of it, a "static" or global function that handles the checking of two arbitrary nodes, and the method which calls into it to properly separate/name the two variations.

Sure.

Ceiu: Done.

PR #2120, opened 2018-09-21
[M] 1631590: Added limits to the size of link headers (ENT-893)

Ceiu:


kahowell:
@Ceiu, please use 1632761 as the BZ reference. Thanks!

Ceiu:
Reopening with the proper branch name to keep the git logs clear.

PR #2123, opened 2018-09-25
[M] 1632761: Added limits to the size of link headers (ENT-893)

Ceiu:


kahowell:
Works as advertised :+1: awesome work!

PR #2127, opened 2018-10-03
[2.3] 1635807: Added limits to the size of link headers

kahowell:

A backport of #2120 to 2.3.


Ceiu:
ack

PR #2128, opened 2018-10-08
[2.1] 1634040: Fixed an issue with object desync during pool cleanup

Ceiu:
- Fixed an issue where consumer instances could become desync'd
  during CPM.deletePools while recalculating compliance status
  for consumers in some cases, leading to an EntityExistsException

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -70,7 +70,7 @@ import org.candlepin.util.Util;
 import org.candlepin.util.Traceable;
 import org.candlepin.util.TraceableParam;
 
-import com.google.common.collect.Lists;
+import com.google.common.collect.Iterables;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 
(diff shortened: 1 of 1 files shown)


Ceiu:
Note for reviewers: this patch undoes one part of the patch that reduces memory usage. In theory, this patch should still trend toward the same memory usage, but it may not get there as fast. However, it may be worth validating that statement if someone has a dataset in mind that could run into memory limits.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2224,21 +2225,11 @@ public void deletePools(Collection<Pool> pools, Collection<String> alreadyDelete
...
-                log.info("Recomputing status for {} consumers", consumers.size());
+                log.info("Recomputing status for {} consumers", consumerStackedEnts.keySet().size());
 
                 // Recalculate status for affected consumers
-                for (List<String> subList : Lists.partition(consumers, 1000)) {
-                    for (Consumer consumer : this.consumerCurator.getConsumers(subList).list()) {
+                for (List<Consumer> subList : Iterables.partition(consumerStackedEnts.keySet(), 1000)) {
+                    for (Consumer consumer : subList) {
                         this.complianceRules.getStatus(consumer);
                         this.consumerCurator.detach(consumer);
```
awood: I would like to keep the comment starting on line 2227 about the purpose of the detach above this line.

Ceiu: Done. I've reworded it a bit to make sense in the new context and moved it to the detach line as requested here.

awood:
I ran ExpiredPools job on a database dump and memory usage is holding stable.  I'd like to run this on a customer with some crazy amount of consumers off a single pool, but I don't have a dataset and I'm satisfied with what I've seen on the database I do have.

PR #2129, opened 2018-10-10
[M] Fixed an issue with object desync during pool cleanup

Ceiu:
- Fixed an issue where consumer instances could become desync'd
  during CPM.deletePools while recalculating compliance status
  for consumers in some cases, leading to an EntityExistsException


Ceiu:

PR #2130, opened 2018-10-11
[M] ActiveMQ sessions are now closed after a commit or rollback (ENT-880)

Ceiu:
- EventSink now closes its sessions after performing either a
  commit or rollback to help avoid leaving dangling sessions
  around indefinitely
- Minor code rearrangements and adjustments to align/standardize
  operation order, style and flow

### DIFF
--- a/server/src/main/java/org/candlepin/audit/EventSinkImpl.java
+++ b/server/src/main/java/org/candlepin/audit/EventSinkImpl.java
@@ -54,6 +54,7 @@ import javax.inject.Inject;
 @Singleton
 public class EventSinkImpl implements EventSink {
     private static Logger log = LoggerFactory.getLogger(EventSinkImpl.class);
+
     private EventFactory eventFactory;
     private ClientSessionFactory factory;
     private Configuration config;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/candlepin/audit/EventSinkImpl.java
```diff
@@ -196,9 +208,12 @@ public void queueEvent(Event event) {
...
     @Override
     public void sendEvents() {
+        log.debug("Committing ActiveMQ transaction");
+
         try {
-            log.debug("Committing ActiveMQ transaction.");
-            getClientSession().commit();
+            ClientSession session = this.getClientSession();
+            session.commit();
+            session.close();
```
mstead: Put the close() into a finally block so that an attempt is always made to close the session, even on commit failure. If closing the session fails, there's not really anything we can do -- log and move on.

server/src/main/java/org/candlepin/audit/EventSinkImpl.java
```diff
@@ -209,10 +224,12 @@ public void sendEvents() {
...
     @Override
     public void rollback() {
-        log.warn("Rolling back ActiveMQ transaction.");
+        log.warn("Rolling back ActiveMQ transaction");
+
         try {
-            ClientSession session = getClientSession();
+            ClientSession session = this.getClientSession();
             session.rollback();
+            session.close();
```
mstead: Put the close() into a finally block so that an attempt is always made to close the session, even on rollback fails. If closing the session fails, there's not really anything we can do -- log and move on.

mstead:
Code looks good, I want to do a little manual testing before I merge.

mstead:
retest this please

mstead:
@Ceiu I got this PR running against an external Artemis server and checked the session counts in the Artemis web console.

Without this patch, when I deploy with test data there are 29 Sessions and 29 producers listed (only 4 sessions are expected to be around while candlepin is running).

With this patch applied, I see the 4 expected sessions, and a new one appears briefly when a request is made.

:+1: 

Will merge once tests pass.

PR #2131, opened 2018-10-11
1618398: Remove service level checking for consumers [ENT-809]

wottop:


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2044,7 +2046,8 @@ public Response bind(
         List<PoolQuantity> dryRunPools = new ArrayList<>();
 
         try {
-            consumerBindUtil.validateServiceLevel(consumer.getOwnerId(), serviceLevel);
+            // BZ 1618398 Remove validation check on consumer service level
+            // consumerBindUtil.validateServiceLevel(consumer.getOwnerId(), serviceLevel);
```
nikosmoum: The method `ConsumerBindUtil.validateServiceLevel` now seems deprecated since it doesn't look like its used anywhere else. Maybe we should just remove it (unless we have a reason to use it again in the future?).

wottop: It gets used when the default service level is set on an org or when an activation key is created.

PR #2133, opened 2018-10-16
1591383: When content access mode is org_environment, disable auto-at… [ENT-814]

wottop:
…tach for org

Add detail to the exception message to indicate content access
Consolidate content access enabled check in owner

### DIFF
--- a/server/src/main/java/org/candlepin/controller/Entitler.java
+++ b/server/src/main/java/org/candlepin/controller/Entitler.java
@@ -204,11 +204,14 @@ public class Entitler {
         Consumer consumer = data.getConsumer();
         Owner owner = data.getOwner();
 
-        if (!consumer.isDev() && owner.isAutobindDisabled()) {
-            log.info("Skipping auto-attach for consumer '{}'. Auto-attach is disabled for owner {}.",
-                consumer, owner.getKey());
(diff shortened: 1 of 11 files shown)


server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -204,11 +204,14 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
         Consumer consumer = data.getConsumer();
         Owner owner = data.getOwner();
 
-        if (!consumer.isDev() && owner.isAutobindDisabled()) {
-            log.info("Skipping auto-attach for consumer '{}'. Auto-attach is disabled for owner {}.",
-                consumer, owner.getKey());
-            throw new AutobindDisabledForOwnerException(i18n.tr("Auto-attach is disabled for owner \"{0}\".",
-                owner.getKey()));
+        if (!consumer.isDev() && owner.isAutobindDisabled() || owner.isContentAccessEnabled()) {
```
Ceiu: Just for clarification, is the intent here to be that the contentAccessEnabled be a valid condition even when the isDev and isAutobindDisabled checks fail?

That is, what we have here is equivalent to:
```if ((!consumer.isDev() && owner.isAutobindDisabled()) || owner.isContentAccessEnabled()) {```

Is that what's intended?

Ceiu:
ack

PR #2135, opened 2018-10-23
ENT-884 Reuse Objectmappers when possible:

nikosmoum:
* Changed all ObjectMapper instance references to static ones
* Changed SyncUtils to a singleton to make sure the ObjectMapper its
wrapping is reused instead of recreated

### DIFF
--- a/server/src/main/java/org/candlepin/audit/EventBuilder.java
+++ b/server/src/main/java/org/candlepin/audit/EventBuilder.java
@@ -14,6 +14,7 @@
  */
 package org.candlepin.audit;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import org.candlepin.audit.Event.Target;
 import org.candlepin.audit.Event.Type;
 import org.candlepin.common.exceptions.IseException;
@@ -25,9 +26,7 @@ import org.candlepin.model.Named;
 import org.candlepin.model.Owned;
 import org.candlepin.model.Pool;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-
+import org.candlepin.util.Util;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -43,13 +42,9 @@ public class EventBuilder {
 
     private static final Logger log = LoggerFactory.getLogger(EventBuilder.class);
 
-    private final ObjectMapper mapper;
-
     private Event event;
 
     public EventBuilder(EventFactory factory, Target target, Type type) {
-        this.mapper = new ObjectMapper();
-
         event = new Event(type, target, null, factory.principalProvider.get(),
                 null, null, null, null, null, null);
     }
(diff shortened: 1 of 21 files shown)


nikosmoum:
@awood I have changed the references from static to Guice-managed for all except two:
- ResultDataUserType, because it is not instantiated by Guice, but by hibernate, and it requires a default non-args constructor. Leaving it static in this case makes sense (unless there is another solution I haven't thought of).
- SyncUtils, because the creation of the ObjectMapper it wraps is dependent on another member-variable of the class (a Configuration instance) that is used elsewhere in the SyncUtils class, and thus is not only the ObjectMapper instance's dependency.

server/src/main/java/org/candlepin/audit/EventBuilder.java
```diff
@@ -121,13 +115,11 @@ public EventBuilder setEventData(Eventful entity) {
                 Map<String, String> eventData = new HashMap<>();
                 eventData.put("subscriptionId", ((Pool) entity).getSubscriptionId());
 
-                try {
-                    event.setEventData(mapper.writeValueAsString(eventData));
-                }
-                catch (JsonProcessingException e) {
-                    log.error("Error while building JSON for pool.created event.", e);
+                String output = Util.toJson(eventData);
+                if (output == null) {
```
awood: This seems odd to me.  Shouldn't `Util.toJson` throw an exception if it is unable to serialize to JSON?

awood:
>     * SyncUtils, because the creation of the ObjectMapper it wraps is dependent on another member-variable of the class (a Configuration instance) that is used elsewhere in the SyncUtils class, and thus is not only the ObjectMapper instance's dependency.

Looks good.  Thanks for doing this.  I know it's not fun to go through and correct a bunch of tests.

For this SyncUtils class, the configuration is actually bound to a single instance in the [context listener](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java#L291) so the Configuration object in SyncUtils will be exactly the same as the one available to you in the CandlepinModule class.  I'm fine either way; just wanted to let you know it's an option to build it in CandlepinModule.

server/src/main/java/org/candlepin/audit/EventBuilder.java
nikosmoum: `Util.toJson` is actually swallowing the exception, logging an error and returning an empty string (don't know why, it might be useful in the other places using this utility method?), so I tried to leave it as it is, and put the burden of throwing the exception here in `EventBuilder`. I did do a mistake though, `if (output == null)` should be `if (output.equals(""))`.

awood: I count 3 uses in actual code
1. org.candlepin.audit.Event's constructor
1. org.candlepin.audit.Event's setPrincipal
1. org.candlepin.audit.auth.Principal.toString

All of them are pretty simple.  Would you mind having Util.toJson just being

```java
    public static String toJson(Object anObject) throws JsonProcessingException {
            return mapper.writeValueAsString(anObject);
    }
```

And then those three callers can catch the exception, log it, and set their variables to "".  Meanwhile, EventBuilder can wrap the JsonProcessingException in an IseException and throw it.

nikosmoum: Done

PR #2138, opened 2018-10-26
[M] 1650635: Fix addons & roles compliance mismatch in rules (ENT-930)

nikosmoum:
Fixes a few things on the back of previous work
to make roles & addons behave similarly to products
in terms of compliance:

- Fix deserialization error by providing a non-compliance reason
of addons and roles in the form of CSV for the whole pool stack,
instead of in the form of an array.
- Alter rules' use of product attribute name from 'role' to 'roles'
as agreed with IT.
- Check for role/addon compliance against each separate value in the
CSV of the role/addon product attribute strings.
- Added proper Compliance Reason output strings for role/addon
mismatch to show to the user.
- Added spec tests that check mismatched role/addons does not
result to failure and the proper reasons are returned.

### DIFF
--- a/server/src/main/resources/rules/rules.js
+++ b/server/src/main/resources/rules/rules.js
@@ -1,4 +1,4 @@
-// Version: 5.29
+// Version: 5.30
 
 /*
  * Default Candlepin rule set.
@@ -55,6 +55,8 @@ var ARCH_FACT = "uname.machine";
 var IS_VIRT_GUEST_FACT = "virt.is_guest";
 var STORAGE_BAND_USAGE = "band.storage.usage";
 var PROD_ARCHITECTURE_SEPARATOR = ",";
+var PROD_ROLE_SEPARATOR = ",";
+var PROD_ADDON_SEPARATOR = ",";
 
 // Product attribute names
 var SOCKETS_ATTRIBUTE = "sockets";
@@ -74,7 +76,7 @@ var VCPU_ATTRIBUTE = "vcpu";
 var MULTI_ENTITLEMENT_ATTRIBUTE = "multi-entitlement";
 var STACKING_ID_ATTRIBUTE = "stacking_id";
 var STORAGE_BAND_ATTRIBUTE = "storage_band";
-var ROLE_ATTRIBUTE = "role";
+var ROLES_ATTRIBUTE = "roles";
 var ADDONS_ATTRIBUTE = "addons";
 
 // caller types
@@ -115,7 +117,7 @@ var PHYSICAL_ATTRIBUTES = [
     ARCH_ATTRIBUTE,
     GUEST_LIMIT_ATTRIBUTE,
     STORAGE_BAND_ATTRIBUTE,
-    ROLE_ATTRIBUTE,
+    ROLES_ATTRIBUTE,
     ADDONS_ATTRIBUTE
 ];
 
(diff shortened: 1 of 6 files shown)


nikosmoum:
@jturel  This should fix the rules issue. One thing to note: the Product attribute name for role is 'roles', not 'role' (in the server, since a Product can support multiple roles in a CSV format, similarly to addons, even though a consumer can have only a single role specified, which is why the syspurpose attribute on the client is called 'role').

To import the rules, either do a POST in the /rules API with the contents of the new rules file, or alternatively, you can paste the whole rules.js contents into the cp_rules.rules_blob column in the database.

jturel:
@nikosmoum thank you! once the build goes green I'll give the rules a test.

jturel:
@nikosmoum looks like some other nice fixes have come in but now this needs a rebase. I was hoping to demo syspurpose in our community demo this Thursday.. Any chance this will be merged before then so that I can request a build?

nikosmoum:
@jturel looking at it now, I will try to get it in by Thursday but I wouldn't bet on it

jturel:
Thanks, it's not a real problem - I can demo the feature next time which will give plenty of time for this to be finished, but now I can plan accordingly.

nikosmoum:
Notes for reviewers: 
* I did a rebase because another PR (#2143) that fixed the main deserialization error this PR was trying to fix was merged first. The rest of the changes here are still valuable.
* I want to bring attention to a specific point in this PR. I'd like to verify that the logic being tested by the new unit tests I added in AutobindRulesTest.java is correct. Specifically, the logic tested by the `testSelectBestPoolsShouldNotSelectPoolWithMismatchedRole` and the `testSelectBestPoolsShouldNotSelectPoolWithMismatchedAddon` (look to their comments for details).

server/src/main/resources/rules/rules.js
```diff
@@ -74,7 +76,7 @@ var VCPU_ATTRIBUTE = "vcpu";
 var MULTI_ENTITLEMENT_ATTRIBUTE = "multi-entitlement";
 var STACKING_ID_ATTRIBUTE = "stacking_id";
 var STORAGE_BAND_ATTRIBUTE = "storage_band";
-var ROLE_ATTRIBUTE = "role";
+var ROLES_ATTRIBUTE = "roles";
```
jturel: Is this an internal change or will this change something about the output of the compliance event that might require katello to change?

nikosmoum: The output of the syspurpose compliance event won't change, because it is about the compliance of a specific consumer, who can only have a single role specified. This change was required because the Product attribute, as agreed with IT, is called `roles`, and not `role` as seen [here](https://github.com/candlepin/candlepin/blob/7c6859c68821767e45e689eb1278bed090ed85c4/server/src/main/java/org/candlepin/model/Product.java#L135). That is because a Product/SKU can have multiple roles (comma-separated list, similarly to how the `arch` and `addons` attributes work).

@kahowell can you verify that my understanding is correct if possible (this is what I understand from reading the code, and what we've agreed with IT)?

nikosmoum:
Hi @jturel Yes, another PR will address this. The original issue and fix of this PR was fixing a behaviour that shouldn't be there in the first place. Instead, the proper fixes of that behavior will be in other PRs. The issues involved (for master) are: https://bugzilla.redhat.com/show_bug.cgi?id=1652737 and https://bugzilla.redhat.com/show_bug.cgi?id=1652967 (not sure when PRs for these will be up, because we're currently working on the 2.3 clones of these for a build for IT, but soon enough I hope).

PR #2141, opened 2018-11-02
[2.5] 1645673: Fixed an issue with syspurpose changes not updating status

Ceiu:
- Fixed an issue with changes to system purpose not causing the
  syspurpose compliance status hash to be updated, events
  generated, nor system purpose status to be retained.


wottop:
ack

PR #2143, opened 2018-11-02
1640215: Fix incorrect partial compliance status on consumers

awood:
If a consumer does not specify a syspurpose attribute (role, usage,
etc.), do not mark them partially compliant if they attach to a pool
that provides that attribute.

This commit also fixes a Jackson error that I discovered while writing
the test cases.  Jackson was attempting to deserialize compliance reason
JSON to a Map<String, String> when the JSON had arrays as the values
instead.

### DIFF
--- a/server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
+++ b/server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
@@ -98,7 +98,6 @@ public class SystemPurposeComplianceRules {
             .collect(Collectors.toSet());
 
         for (Entitlement entitlement : entitlements) {
-
             String unsatisfedRole = consumer.getRole();
             Set<String> unsatisfiedAddons = new HashSet<>(consumer.getAddOns());
             String preferredSla = consumer.getServiceLevel();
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
```diff
@@ -140,19 +139,21 @@ public SystemPurposeComplianceStatus getStatus(Consumer consumer,
                 }
                 unsatisfiedAddons.removeAll(addonsFound);
 
-                if (product.hasAttribute(Product.Attributes.SUPPORT_LEVEL)) {
+                if (StringUtils.isNotEmpty(preferredSla) &&
```
Ceiu: If there's something else to fix in this PR, we could also reorganize this code a bit to only do this check once. Both of the places it was added to could be nested within a single check, since both are contingent on it being non-empty.

awood: I'm not sure I understand.  One spot is `StringUtils.isNotEmpty(preferredSla)` and the other is `StringUtils.isNotEmpty(preferredUsage)`.

Ceiu: Oof... disregard. This is what happens when I review code while dealing with migraine-induced double vision.

PR #2145, opened 2018-11-09
Add support for running rspec tests via Gradle

barnabycourt:
Enable running the rspec tests.

This assumes that the ruby environment is set up separately and 'rspec' can be run from the command line. 

Supports specifying the spec file name using the -- spec option. 
Supports specifying the test name using the --test option.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -1,3 +1,4 @@
+import org.gradle.api.tasks.options.Option
 
 apply plugin: 'war'
 apply plugin: Gettext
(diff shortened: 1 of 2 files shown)


server/build.gradle
```diff
@@ -48,6 +49,60 @@ configurations {
...
+
+    @Option(option = 'test', description = 'Set the name of the spec test to run')
+    void setTest(final String test_name) {
+        this.test = test_name
+    }
+
+    @TaskAction
+    runRspec() {
+        def rspec_args = [
+                "--format", "documentation",
```
awood: Can we add "--force-color"?  That gets the colors back in the output making it easy to tell when a test fails (since it prints red)

PR #2146, opened 2018-11-13
Add ability to run msgmerge via gradle & update the readme.

barnabycourt:
Also, mask generated source files form checkstyle


wottop:
ack

PR #2149, opened 2018-11-15
[2.5] 1642621: Remove service level checking for consumers [ENT-929]

nikosmoum:
This is a patch of #2131 for the 2.5-HOTFIX branch.


wottop:
ack

PR #2150, opened 2018-11-15
Catch exception in the scenario where guest migration has no consumers

wottop:


mstead:
@wottop Please provide steps to reproduce and expand on the commit message as to why this is needed and what the commit is about.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -322,7 +322,13 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
                     if (factsUpdated || guestMigration.isMigrationPending() || typeUpdated ||
                         hypervisorIdUpdated) {
                         knownHost.setLastCheckin(new Date());
-                        guestMigration.migrate(false);
+                        try {
+                            guestMigration.migrate(false);
+                        }
+                        catch (IllegalStateException ise) {
```
Ceiu: This is only thrown when we don't have a migrationPending. Maybe instead of silently throwing out the exception, we could update the check on line 322 to require the isMigrationPending to return true, as well as any of the other flags. I.e.:
```if (guestMigration.isMigrationPending() && (factsUpdated || typeUpdated || hypervisorIdUpdated)) { ...```

kahowell: >This is only thrown when we don't have a migrationPending. Maybe instead of throwing out the exception, we could update the check on line 322 to require the isMigrationPending to return true, as well as any of the other flags. I.e.:
if (guestMigration.isMigrationPending() && (factsUpdated || typeUpdated || hypervisorIdUpdated)) { ...

@wottop and I chatted a bit about this yesterday, and we assumed that perhaps there's still some sort of race condition where you could still get the ISE, so I recommended the try, catch. Do you think we can rule out a race condition here? If so, the check would be fine.

I like the idea of just changing the `migrate()` to just debug log.

Ceiu: If a race condition is still a concern, then I'd argue the silent (or semi-silent) noop in migrate is the superior option. If we hit that condition that we no longer need to migrate, then log it, and do nothing.

Either way, the state doesn't look to be a bad one, so if we're at a point where we're silently discarding it anyway, maybe it doesn't need to exist.

PR #2152, opened 2018-11-19
[M] 1631553: Use Product attribute names for owner syspurpose [ENT-900]

nikosmoum:


jturel:
This looks good to me. Small change from our side as we only use the output of this API in one location!

wottop:
ack

PR #2156, opened 2018-11-20
[2.3] 1651713: Unneeded guest migration call leads to exception

wottop:


barnabycourt:
Adding the inprogress label until we get the green on the build currently under test

Ceiu:
ack

PR #2157, opened 2018-11-20
[2.4] 1651710: Unneeded guest migration call leads to exception

wottop:

### DIFF
--- a/server/spec/hypervisor_check_in_spec.rb
+++ b/server/spec/hypervisor_check_in_spec.rb
@@ -669,18 +669,19 @@ describe 'Hypervisor Resource', :type => :virt do
     return { host_hyp_id => guest_id_list }
   end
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -291,7 +291,9 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
 
                     if (factsUpdated || guestMigration.isMigrationPending() || typeUpdated) {
                         knownHost.setLastCheckin(new Date());
-                        guestMigration.migrate(false);
+                        if (guestMigration.isMigrationPending()) {
```
Ceiu: This check is extraneous. We've already checked it in the parent conditional, and the method checks it before it executes.

Ceiu:
ack

PR #2158, opened 2018-11-20
1556930: Index to make hypervisor uuid lookup faster

wottop:
Postgres allows the condition to better focus the index.
MySql does not, so the index will need to be broader.

### DIFF
--- a/server/src/main/resources/db/changelog/20181116110941-virt-fact-index.xml
+++ b/server/src/main/resources/db/changelog/20181116110941-virt-fact-index.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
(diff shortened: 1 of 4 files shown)


barnabycourt:
@wottop I believe the index recommended in the KCS was limited to aparticular key. Is there a reason to index all values? The KCS recommended the index:
su - postgres -c "psql candlepin -c \"create index cp_consumer_fact_mapkey_idx on cp_consumer_facts (mapkey, element) where mapkey='virt.uuid';\""

barnabycourt:
@wottop This is waiting on investigation from shayne regarding performance impacts in mariadb and whether or not they would prefer that the index be omitted from mariadb entirely.

tdarby:
Shayne was able to create this index in our test env in just under an hour.  Given that this will require what is considered a significant outage, and will add >20GB to the db size, I recommend not doing this for mysql at this time.  We can revisit when we have a larger outage scheduled.

Ceiu:
ack

PR #2169, opened 2018-12-04
1588101: dynamicly filter activation key's pools subfields [ENT-972]

nikosmoum:
- Now using ?includes=pools.poolId or ?includes=pools.quantity on the
GET /owners/{owner_key}/activation_keys API will only show the requested subfield,
instead of the whole pools field.
- Also similarly fixed: using ?include=principal.name or
?include=principal.type on the GET /owners/{owner_key}/events API.
- By extending `CandlepinDTO`, which has a `@JsonFilter("DTOFilter")` on it, like all other
DTO classes, forces jackson to serialize properties of these classes using filters (even though
the `DTOFilter` does not exist, the default filter (`DynamicPropertyFilter`) will be used).

### DIFF
--- a/server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyDTO.java
@@ -14,6 +14,7 @@
  */
 package org.candlepin.dto.api.v1;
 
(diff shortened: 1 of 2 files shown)


nikosmoum:
@awood Explanation added.

PR #2171, opened 2018-12-05
Allow tomcat to read the candlepin CA cert

jturel:
In our Katello deployment we have recently moved the candlepin CA cert and key to live under /etc/candlepin/certs and I noticed that when using selinux the app just doesn't work - /candlepin/status gives a 404

I found some denials in audit.log:

```
[root@nightly-test certs]# audit2allow -a

#============= tomcat_t ==============
allow tomcat_t candlepin_etc_certs_ca_cert_r_t:file { getattr open read };
allow tomcat_t cgroup_t:dir search;
allow tomcat_t cgroup_t:file { getattr open read };
```

I applied these and everything seems to work fine, and I think I've translated these in the proper way to the policy in this repo.

### DIFF
--- a/server/selinux/candlepin.te
+++ b/server/selinux/candlepin.te
@@ -75,10 +75,12 @@ miscfiles_read_localization(candlepin_t)
 
 require {
     type candlepin_etc_certs_ca_cert_r_t;
+    type cgroup_t;
     type httpd_t;
     type tomcat_t;
     type postgresql_port_t;
     class file { read getattr open };
+    class dir search;
 }
 
 # for httpd/katello, access to the candlepin ca certs
@@ -87,6 +89,9 @@ allow httpd_t candlepin_etc_certs_ca_cert_r_t: file { read getattr open };
 allow httpd_t candlepin_etc_certs_rw_t:dir search;
 
 # allow tomcat to access candlepin ca certs, etc.
+allow tomcat_t cgroup_t:dir search;
+allow tomcat_t cgroup_t:file { getattr open read };
+allow tomcat_t candlepin_etc_certs_ca_cert_r_t: file { read getattr open };
 allow tomcat_t candlepin_etc_certs_rw_t:dir { list_dir_perms };
 allow tomcat_t candlepin_etc_certs_rw_t:file { read_file_perms };
 allow tomcat_t candlepin_etc_rw_t:dir { list_dir_perms };


ehelms:
I also seem to now encounter this. Could this get looked at?

awood:
@candlepin-pull-request-bot okay to test

awood:
@jturel Where were the certs before?  On my deployment, they've always been under /etc/candlepin/certs

I'm a little confused by this bug.  Tomcat is using the cert stored in the `keystore` file symlinked in to `/etc/tomcat` from `/etc/candlepin` and not having any problem with that (otherwise you would be getting something besides a 404) so I'm guessing the issue is with the Candlepin code reading the .crt and .key file for when it comes time to do a signing.

Would you mind posting some more context: the exact AVC denial and some more context from the Candlepin log?  I imagine there's a stack trace in there from where deployment failed.

jturel:
The AVC denial taken from last week:

```
type=AVC msg=audit(1543981267.690:4266): avc:  denied  { read } for  pid=19475 comm="java" name="candlepin-ca.crt" dev="vda1" ino=2491120 scontext=system_u:system_r:tomcat_t:s0 tcontext=system_u:object_r:candlepin_etc_certs_ca_cert_r_t:s0 tclass=file
type=SYSCALL msg=audit(1543981267.690:4266): arch=c000003e syscall=2 success=no exit=-13 a0=7f40257d8cb0 a1=0 a2=1b6 a3=7f4069052780 items=0 ppid=1 pid=19475 auid=4294967295 uid=53 gid=53 euid=53 suid=53 fsuid=53 egid=53 sgid=53 fsgid=53 tty=(none) ses=4294967295 comm="java" exe="/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.191.b12-1.el7_6.x86_64/jre/bin/java" subj=system_u:system_r:tomcat_t:s0 key=(null)
type=PROCTITLE msg=audit(1543981267.690:4266): proctitle=2F7573722F6C69622F6A766D2F6A72652F62696E2F6A617661002D586D73313032346D002D586D78343039366D002D636C61737370617468002F7573722F73686172652F746F6D6361742F62696E2F626F6F7473747261702E6A61723A2F7573722F73686172652F746F6D6361742F62696E2F746F6D6361742D6A756C692E6A
```

The only thing that looks different in the logs is this which appears shortly after starting tomcat:

```
Dec 10, 2018 3:51:11 PM org.apache.catalina.loader.WebappClassLoaderBase loadClass
INFO: Illegal access: this web application instance has been stopped already.  Could not load com.mchange.v2.resourcepool.BasicResourcePool$AsyncTestIdleResourceTask.  The eventual following stack trace is caused by an error thrown for debugging purposes as well as to attempt to terminate the thread which caused the illegal access, and has no functional impact.
java.lang.IllegalStateException
        at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1777)                                                                      
        at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1735)                                                                      
        at com.mchange.v2.resourcepool.BasicResourcePool.checkIdleResources(BasicResourcePool.java:1635)                                                                    
        at com.mchange.v2.resourcepool.BasicResourcePool.access$2000(BasicResourcePool.java:44)                                                                             
        at com.mchange.v2.resourcepool.BasicResourcePool$CheckIdleResourcesTask.run(BasicResourcePool.java:2176)                                                            
        at java.util.TimerThread.mainLoop(Timer.java:555)
        at java.util.TimerThread.run(Timer.java:505)
```

@awood the certs were previously (I think) in /etc/pki/katello somewhere. @ehelms can you confirm?

ekohl:
We did move them around in our Puppet managed config. Puppet does look at the selinux types it should set for the file, essentially doing a restorecon.

ehelms:
Another data point, if we put the certificates in place before `candlepin-selinux` is installed, then the certificates end up with the right context. If the order is reversed we end up with the denials.

@awood Background is, Katello used to generate a CA cert and key and put them in a central location for all services needing them to use. We recently moved to putting certs to their default service location even if it means duplication of those certs on the system. So we just switched to putting the CA cert and key back to `/etc/candlepin`.

awood:

awood:
Thanks for the PR!

PR #2174, opened 2018-12-06
1656612: Concurrent content addition error [2.5] [ENT-1004]

wottop:
This spec test run against the base branch will show an exception where the row has been removed. This change will correct that.


wottop:

PR #2175, opened 2018-12-06
1588097: Expose owner of ComplianceStatusDTO's pools [ENT-783]

nikosmoum:
- Now the output of GET /consumers/{consumer_uuid}/compliance will have the
  owner field of compliant products' entitlements' pools populated.


wottop:
ack

PR #2177, opened 2018-12-06
1656611: Concurrent content addition error [2.1] [ENT-1004]

wottop:
This spec test run against the base branch will show a single content associated to the product out of 5. This change will correct that.

### DIFF
--- a/server/spec/owner_content_resource_spec.rb
+++ b/server/spec/owner_content_resource_spec.rb
@@ -124,4 +124,19 @@ describe 'Owner Content Resource' do
     # Verify the serial has changed (the entitlement has been regenerated)
     original_serial.should_not == new_serial
   end
+
+  it 'concurrently updates content in the org' do
+    content1 = create_content("name" => "content1")
+    content2 = create_content("name" => "content2")
+    content3 = create_content("name" => "content3")
+    t6 = Thread.new{@cp.update_content(@owner['key'], content1['id'], {"name" => "content4"})}
+    t7 = Thread.new{@cp.update_content(@owner['key'], content2['id'], {"name" => "content5"})}
+    t8 = Thread.new{@cp.update_content(@owner['key'], content3['id'], {"name" => "content6"})}
+    t6.join
+    t7.join
+    t8.join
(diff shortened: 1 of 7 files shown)


Ceiu:
The code itself is fine and seems to address part of the problem, however I'd prefer to avoid locking on the whole owner if we can. I'm working on a PoC to lock on the owner-product relation instead to see if that will work in place of a global owner lock.

Also, we should do the same locking on content, as the problem exists there as well.

Ceiu:
As stated above, this code works for the very specific path called out in the BZ. However, the hammer we're using to solve the problem is a bit too big and doesn't cover the other paths which are broken by the same issue.

I wrote a quick PoC to verify we can lock only on the relation rather than the whole owner and get the same, corrected, result:

OwnerProductCurator:
```
    /**
     * Locks the owner-product relation for the given owner ID and product ID. This should be done
     * before the product is fetched to ensure the proper state is fetched.
     *
     * @param ownerId
     *  The ID of the owner for the owner-product relation to lock
     *
     * @param productId
     *  The ID of the product for the owner-product relation to lock
     *
     * @return
     *  true if the owner-product relation was found and locked successfully; false otherwise
     */
    public boolean lockOwnerProductRelation(String ownerId, String productId) {
        String jpql = "SELECT op FROM OwnerProduct op " +
            "WHERE op.owner.id = :owner_id AND op.product.id = :product_id";

        try {
            OwnerProduct op = this.getEntityManager()
                .createQuery(jpql, OwnerProduct.class)
                .setParameter("owner_id", ownerId)
                .setParameter("product_id", productId)
                .setLockMode(javax.persistence.LockModeType.PESSIMISTIC_WRITE)
                .getSingleResult();

            return true;
        }
        catch (javax.persistence.NoResultException e) {
            return false;
        }
    }
```

OwnerProductResource:
```
        // Lock the relation so we don't do multiple updates in parallel and clobber one of them
        boolean locked = this.ownerProductCurator.lockOwnerProductRelation(owner.getId(), productId);
```

The locking should also occur in all of the update and delete methods of both products and content, before the product or content to be updated is fetched.

Lorquas:

wottop:

PR #2178, opened 2018-12-06
1599456: fix physical system activation key registration [ENT-742]

nikosmoum:
- Now a physical system using activation keys to register will not fail
  with an authorization error.


wottop:
ack

PR #2187, opened 2018-12-12
fixing container builds

Lorquas:
Trying to get the containers to build more reliably.  Work in progress, adding fixes as I find them...

- rvm needs an extra gpg key to install properly

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -12,7 +12,9 @@
 # Exit on any error:
 set -e
 
-source /etc/profile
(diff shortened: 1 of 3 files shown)


Lorquas:
One thought,  we could add the -c option to the candlepin-jobs that builds this.  This would enable caching and make this thing build much faster if the jenkins node already has these images.  Can anybody think of any downsides to that?

Ceiu:
I see no obvious problems with doing that. Ideally any deltas will have different hashes so the cache will be invalidated in such a case. Unless there's a case that's not immediately jumping to mind, I think it'll be okay.

Lorquas:
retest this please

PR #2194, opened 2018-12-19
1660878: changed the possible syspurpose status states [ENT-1026]

nikosmoum:

### DIFF
--- a/server/src/main/java/org/candlepin/policy/SystemPurposeComplianceStatus.java
+++ b/server/src/main/java/org/candlepin/policy/SystemPurposeComplianceStatus.java
@@ -29,12 +29,12 @@ import java.util.Set;
  * ComplianceStatus
  *
  * Represents the system purpose compliance status for a given consumer. Carries information
- * about which roles, SLA, usage and add-ons are fully entitled, not entitled, or partially entitled.
+ * about which roles, SLA, usage and add-ons are entitled or not entitled, or if none are specified.
  */
 public class SystemPurposeComplianceStatus {
-    public static final String GREEN = "valid";
-    public static final String YELLOW = "partial";
-    public static final String RED = "invalid";
+    public static final String MATCHED = "matched";
+    public static final String MISMATCHED = "mismatched";
+    public static final String NOT_SPECIFIED = "not specified";
 
     // Date this compliance was set on
     private Date date;
@@ -46,8 +46,6 @@ public class SystemPurposeComplianceStatus {
     private Map<String, Set<Entitlement>> compliantAddOns;
     private Map<String, Set<Entitlement>> compliantSLA;
     private Map<String, Set<Entitlement>> compliantUsage;
-    private Map<String, Set<Entitlement>> nonPreferredSLA;
-    private Map<String, Set<Entitlement>> nonPreferredUsage;
     private Set<String> reasons;
     private I18n i18n;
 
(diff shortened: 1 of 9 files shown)


nikosmoum:
Note: do not merge until the in progress label is removed.

nikosmoum:
@Ceiu I have changed the wording on the `getStatus()` comment to make the status states more clear.

Some more info on the change, if you or anyone else plan to do manual testing, and how the old status labels map to the new:

- The 'valid' status will map directly to 'matched', and the 'invalid' status will map directly to 'mismatched'.
- The 'not specified' is a new status. Previously, if the consumer did not specify any syspurpose attributes, the status would be 'valid'.
- The 'partial' status is where the trouble was, because we cannot map all 'partial' statuses into either 'matched' or 'mismatched'. There some cases which the current 'partial' status should really be 'matched', and other cases where it should really be 'mismatched'. Examples:

Example1:
- System specifies only one syspurpose attribute: SLA 'mysla', and has some installed products.
- System auto-attaches (or attaches multiple pools manually), and gets subscribed to 3 pools, with pool1 covering 'mysla', and pool2 and pool3 covering its products.
- Based on the design, I would expect the syspurpose status to be 'valid/matched', since 'mysla' is covered by pool1, but it turns out it is 'partial' if it happens that any other pool(s) the system is attached to (pool2 or pool3 in this case) happens to have an SLA which isn't 'mysla'. This is a very likely scenario to happen and seems outright wrong.
- In this case, the status that used to be 'partial' should be 'matched'.
- This scenario is applicable for the usage attribute as well.
- This scenario has one more problem: for every attached pool that has a mismatched SLA or usage, the status contains an additional reason (of non-compliance). In this example the status would contain something along the lines of this:
```
  {
    compliantSLA: mysla,
    reasons: {
                'mysla does not match with the sla of pool2',
                'mysla does not match with the sla of pool3'
             }
  }
```
So, not only is it wrong to have non-compliance reasons in this case (because the status should be valid/matched), but there are actually multiple reasons added, which is redundant.


Example2:
- System specifies only 2 addons as syspurpose attributes and nothing else: 'addon1' and 'addon2'.
- System attaches to a pool that covers only 'addon1'.
- Based on the design, I would expect that the syspurpose status is 'invalid/mismatched', since there is still a syspurpose attribute (addon2) that isn't matched (I assume we treat each addon specified as a separate attribute). But currently the status is 'partial'. 
- In this case, the status that used to be 'partial' should be 'mismatched'.
In general for addons, I would expect to have a 'matched' status if all specified addons are satisfied, and a 'mismatched' status if at least one specified addon is not satisfied.

This change will rectify the problems in both those scenarios.

server/src/main/java/org/candlepin/policy/SystemPurposeComplianceStatus.java
```diff
@@ -269,44 +211,37 @@ public boolean isCompliant() {
...
-     * * Role is specified, and not even one entitlement provides the preferred role
-     * * Add ons are specified, and none of the add ons are provided are provided by the pools
-     * * usage is specified, and entitlements provide a single usage that does not match preference
-     * * usage is specified, no entitlements provide the usage preference
+     * The status is 'Mismatched' if at least one syspurpose attribute (SLA, role, addons, usage)
+     * is specified, and from those specified, at least one is not satisfied by the consumer's entitlements.
+     * (This includes the scenario of multiple addons being specified, where only some but not all of them
+     * are satisfied)
      *
+     * The status is 'Not Specified' when NONE of the attributes were specified by the consumer at all.
```
Ceiu: :+1:
Much clearer on intent according to docs here

Ceiu:
ack

manual testing passed as expected; overriding Jenkins shenanigans

PR #2200, opened 2019-01-03
[F] Added the JobBuilder, JobExecutionContext and CandlepinAsyncJob classes (ENT-1001)

Ceiu:
- Updated JobManager to use the new JobBuilder in place of the Quartz
  JobDetail class

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobBuilder.java
+++ b/server/src/main/java/org/candlepin/async/JobBuilder.java
@@ -14,6 +14,10 @@
  */
 package org.candlepin.async;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
 
 
 /**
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/candlepin/async/JobBuilder.java
```diff
@@ -14,13 +14,287 @@
...
+        return this;
+    }
+
+    /**
+     * Fetches the name of this job. If the job name has not yet been set, this method returns
+     * null.
+     *
+     * @return
+     *  The name of this job, or null if the name has not been set
+     */
```
mstead: Javadoc references the name attribute, should be the group.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -199,41 +208,40 @@ private void sendJobMessage(JobMessage jobMessage) throws ActiveMQException, Jso
...
         }
 
-        // TODO: Rewrite this method to use the upcoming JobBuilder
-
         AsyncJobStatus job = new AsyncJobStatus();
 
-        job.setJobClass(detail.getJobClass());
-        job.setName(detail.getKey().getName());
-        job.setGroup(detail.getKey().getGroup());
+        job.setJobClass(builder.getJobClass());
```
mstead: We should store a job key instead of the actual job class so that we have flexibility to change the class name if we choose -- storing the actual class name has bitten us with Quartz in the past. When receiving a job message, candlepin should look up the Job class via some sort of registry / config.

server/src/main/java/org/candlepin/async/JobBuilder.java
Ceiu: Fixed

server/src/main/java/org/candlepin/async/JobManager.java
Ceiu: Done. The way I've sorted this for the moment is adding a static map to JobManager which can be updated via calls to the static methods registerJob and unregisterJob, and keys can be resolved by calling getJobClass. The idea for the fix is that actual job classes will have a static block that looks like this:

```
public class RefreshPoolsJob {
    static {
        JobManager.registerJob("refresh_pools", RefreshPoolsJob.class);
    }

    ...
}
```

mstead: :+1:

PR #2164, opened 2018-11-28
Add gradle support for deploy script.

barnabycourt:
* Add candlepin.conf generation
* Add Logdriver support
* Switch server/bin/deploy to gradle
* Update the README

PR #2202, opened 2019-01-10
Modify cp-test to be gradle-aware

kahowell:
For now, the cp-test script should support both... This will allow us to
roll out gradle across the various hotfix branches without doing them
all at once.


kahowell:
Will continue fixes on #2164.

PR #2203, opened 2019-01-11
1482942: Remove unnecessary deleted consumer checks

Januson:
- Remove unnecessary deleted consumer check from ConsumerAuth.createPrincipal. This check is done by ConsumerStore.lookup
 - Refactor ConsumerStore.lookup so that the deleted consumer check is run only when the consumer was not found

### DIFF
--- a/server/src/main/java/org/candlepin/auth/ConsumerAuth.java
+++ b/server/src/main/java/org/candlepin/auth/ConsumerAuth.java
@@ -14,15 +14,13 @@
  */
 package org.candlepin.auth;
 
+import com.google.inject.Inject;
 import org.candlepin.common.exceptions.GoneException;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.ConsumerCurator;
 import org.candlepin.model.DeletedConsumerCurator;
 import org.candlepin.model.Owner;
 import org.candlepin.model.OwnerCurator;
-
-import com.google.inject.Inject;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xnap.commons.i18n.I18n;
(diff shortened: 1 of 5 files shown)


server/spec/consumer_resource_spec.rb
```diff
@@ -197,11 +197,11 @@
     end.should raise_exception(RestClient::Gone)
   end
 
-  it 'should return a 410 for a consumer with an invalid identity cert' do
+  it 'should return a 401 for a consumer with an invalid identity cert' do
```
Ceiu: :+1:

server/src/main/java/org/candlepin/auth/ConsumerAuth.java
```diff
@@ -39,41 +35,37 @@
...
+     * consumerUuid.
+     *
+     * @param consumerUuid requested consumer
+     * @return created principal
+     */
     public ConsumerPrincipal createPrincipal(String consumerUuid) {
-        ConsumerPrincipal principal = null;
-
-        if (consumerUuid != null) {
-            // If this UUID has been deleted, return a 410.
```
Ceiu: While I'm very much in favor of this being less transparent for data security reasons, I'd be interested to know if there's an old, explicit reason we returned a 410 in this case. Given that it's more work to check for the consumer's state here, I have to assume someone requested it at one point.

server/src/main/java/org/candlepin/auth/ConsumerAuth.java
```diff
@@ -39,41 +35,37 @@
...
-                }
-            }
+        final Consumer consumer = this.consumerCurator.getConsumer(consumerUuid);
+        if (consumer == null) {
+            return null;
         }
 
+        final Owner owner = this.ownerCurator.findOwnerById(consumer.getOwnerId());
+        final ConsumerPrincipal principal = new ConsumerPrincipal(consumer, owner);
+        if (log.isDebugEnabled()) {
```
Ceiu: Since we're addressing this code anyway, we can drop the isDebugEnabled check here. It's already done implicitly by the .debug check, and our output code is trivial here.

server/src/main/java/org/candlepin/auth/SSLAuth.java
```diff
@@ -44,11 +42,11 @@
...
     @Inject
     SSLAuth(ConsumerCurator consumerCurator, OwnerCurator ownerCurator,
-        DeletedConsumerCurator deletedConsumerCurator,
-        Provider<I18n> i18nProvider) {
-        super(consumerCurator, ownerCurator, deletedConsumerCurator, i18nProvider);
+        DeletedConsumerCurator deletedConsumerCurator) {
+        super(consumerCurator, ownerCurator, deletedConsumerCurator);
     }
 
+    @Override
```
Ceiu: :+1:

server/src/main/java/org/candlepin/auth/ConsumerAuth.java
awood: I wrote this bit 7 years ago and unfortunately, I can't remember anything about it.  However, looking at [subscription-manager](https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/scripts/rhsmcertd_worker.py#L128), I do think we want to continue returning 410 if the consumer has been deleted.

```python
if ge.deleted_id == uuid:
    log.critical("Consumer profile \"%s\" has been deleted from the server. Its local certificates will now be archived", uuid)
    managerlib.clean_all_data()
    log.critical("Certificates archived to '/etc/pki/consumer.old'. Contact your system administrator if you need more           information.")
```

Januson: Done. I reverted the changes so that the return code stays 410.

server/src/main/java/org/candlepin/auth/ConsumerAuth.java
```diff
@@ -51,30 +49,35 @@
...
+            return null;
         }
 
+        final Owner owner = this.ownerCurator.findOwnerById(consumer.getOwnerId());
+        final ConsumerPrincipal principal = new ConsumerPrincipal(consumer, owner);
+        log.debug("principal created for consumer {}", principal.getConsumer().getUuid());
         return principal;
     }
 
+    private boolean wasDeleted(final String consumerUuid) {
```
Ceiu: This is stylistic thing, but I'm not seeing the value of pulling small one-off operations like this out of the main code. Depending on how the compiler handles this, we may inherit the overhead of a method call to get some self-documenting code, when a comment can achieve the same thing without the risk of any additional code or runtime complexity.

Similarly, declaring variables as final doesn't have a ton of value in this context outside of programmer hints (especially in Java8 where the inner class ref doesn't require it anymore).

Ceiu:
ack

PR #2206, opened 2019-01-14
All Jobs Should Not Retry On Any Persistence Failure[ENT-733]

Januson:
- Update RefreshPoolsJobTest and UndoImportsJob to retry on LockTimeoutException instead of generic PersistenceException


Januson:
@mstead @nikosmoum Please re-review. I updated the PR so that it affects retry of all jobs.

server/src/main/java/org/candlepin/pinsetter/tasks/KingpinJob.java
```diff
@@ -131,23 +130,40 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
...
+            Throwable cause = e.getCause();
+            while (cause != null) {
+                if (SQLException.class.isAssignableFrom(cause.getClass())) {
+                    log.warn("Caught a runtime exception wrapping an SQLException.");
+                    refireCheck(context, e);
+                }
+                cause = cause.getCause();
             }
+            // Otherwise throw as we would normally for any generic Exception:
+            log.error("RefreshPoolsJob encountered a problem.", e);
```
nikosmoum: Now that this code lives in KingpinJob, the logging should change to output the subclass' name here, using `this.getClass().getName()`.

server/src/main/java/org/candlepin/pinsetter/tasks/KingpinJob.java
```diff
@@ -131,23 +130,40 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
...
+            throw new JobExecutionException(e.toString(), e, false);
+        }
+        catch (JobExecutionException e) {
+            throw e;
+        }
+        // Catch any other exception that is fired and re-throw as a
+        // JobExecutionException so that the job will be properly
+        // cleaned up on failure.
+        catch (Exception e) {
+            log.error("RefreshPoolsJob encountered a problem.", e);
```
nikosmoum: Same here for logging.

PR #2207, opened 2019-01-14
[M] 1623491: Get syspurpose attributes an owner's consumers [ENT-821]

nikosmoum:
- Created new API /owners/{owner_key}/consumers_system_purpose used
to retrieve an aggregate of all the syspurpose attributes that the
owner's consumers have specified.

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -295,6 +295,10 @@ class Candlepin
     get("/owners/#{owner}/system_purpose")
   end
 
+  def get_owner_consumers_syspurpose(owner)
+    get("/owners/#{owner}/consumers_system_purpose")
+  end
+
   def get_owner_hypervisors(owner, hypervisor_ids = [])
     url = "/owners/#{owner}/hypervisors"
     params = {
--- a/server/spec/owner_resource_spec.rb
+++ b/server/spec/owner_resource_spec.rb
@@ -5,6 +5,7 @@ require 'candlepin_scenarios'
 require 'rubygems'
 require 'rest_client'
 require 'time'
+require 'json'
 
 describe 'Owner Resource' do
   include CandlepinMethods
@@ -589,7 +590,7 @@ describe 'Owner Resource' do
     # output for owners, so we don't need to verify their presence.
   end
 
-  it 'accepts system purpose attributes' do
+  it 'lists system purpose attributes of its products' do
     owner_key = random_string("owner")
     @cp.create_owner(owner_key)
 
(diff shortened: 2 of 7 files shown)


nikosmoum:
@jturel I would appreciate a review when you can since you were driving this API addition!

jturel:
I'll take a look 👍

jturel:
Thanks for the API! It looks sane to me but I have just one clarifying question: if I have two consumers which both have, for example, 'addon1' will the output for the addons key have one occurrence of addon1? Same question for the other fields. I assume yes, but just want to be sure.

You may want to update your spec test to assert that behavior just in case it should be inadvertently changed in the future.

nikosmoum:
@jturel The output will have only one 'addon1' occurrence if more than one consumer has specified it. I'll make sure to put a test case for this. Thanks for the review!

nikosmoum:
## Performance Testing

This is a performance comparison between the original commit and the second commit.

The original commit fetches all the Consumers of the owner (in this case, all 51k+ of them) with a single commit and scrapes the syspurpose attributes off them. The second commit, instead, is doing 4 queries (one for each attribute) using DISTINCT to get the unique aggregate values of each.

### Set Up

A JMeter test was used to register around 51k consumers for the same owner. The consumers all specified syspurpose data based on the valid_fields.json (plus a few custom addon values which are missing from it). Each consumer specified a random role/usage/sla and 2 addons each based on the these values:
```
{
    "role": [
      "Red Hat Enterprise Linux Server",
      "Red Hat Enterprise Linux Workstation",
      "Red Hat Enterprise Linux Compute Node",
      "Red Hat Enterprise Linux Client/Desktop"
    ],
    "service_level_agreement": [
      "Premium",
      "Standard",
      "Self-Support"
    ],
    "usage": [
      "Production",
      "Development/Test",
      "Disaster Recovery"
    ],
    "addons": [
      "addon1",
      "addon2",
      "addon3",
      "addon4",
      "addon5",
      "addon6",
    ]
}
```

### Commit 1 Results (Fetching all Consumers & scrape the attributes of them):

A single query cannot be completed, due to memory problems (fetching 51k Consumers in memory...):

"Runtime Error Ran out of memory retrieving query results. at org.postgresql.core.PGStream.receiveTupleV3:408"


### Commit 2 Results (Using 4 queries, one for each attribute):

A single query of the endpoint has an average response of 360ms, and under load an average response of 4735ms:

Average of 100 requests (sequential):

| Label                      | # Samples | Average | Median | 90% Line | 95% Line | 99% Line | Min | Max | Error % | Throughput | Received KB/sec | Sent KB/sec | 
|----------------------------|-----------|---------|--------|----------|----------|----------|-----|-----|---------|------------|-----------------|-------------| 
| request | 100       | 360     | 354    | 392      | 411      | 525      | 317 | 541 | 0.000%  | 2.76878    | 2.23            | 0.62        | 

Average of 100 requests (10 loops of 10 concurrent requests each):

| Label                      | # Samples | Average | Median | 90% Line | 95% Line | 99% Line | Min  | Max  | Error % | Throughput | Received KB/sec | Sent KB/sec | 
|----------------------------|-----------|---------|--------|----------|----------|----------|------|------|---------|------------|-----------------|-------------| 
| request | 100       | 4735    | 4802   | 5604     | 5732     | 6079     | 2308 | 6116 | 0.000%  | 2.06458    | 1.66            | 0.46        |

wottop:
It looks good to me. Is the squash tag still needed?

nikosmoum:
@wottop yes, I will squash shortly!

wottop:
ack

PR #2208, opened 2019-01-15
Allow passing credentials to cpsetup via environment variables[ENT-940]

Januson:
Credentials passed as an argument take precedence. When no credentials are passed then environment variables are checked. When environment variables are missing, script fallbacks to hardcoded default values.

- Allow passing a db name as an environment variable CANDLEPING_DB_NAME
- Allow passing a db user as an environment variable CANDLEPIN_DB_USER
- Allow passing a db password as an environment variable CANDLEPING_DB_PASSWORD

### DIFF
--- a/server/code/setup/cpsetup
+++ b/server/code/setup/cpsetup
@@ -210,11 +210,11 @@ def main(argv):
                   action="store_true", dest="skipdbcfg", default=False,
                   help="don't configure the /etc/candlepin/candlepin.conf file")
     parser.add_option("-u", "--user",
-                  dest="dbuser", default="candlepin",
-                  help="the database user to use")
+                  dest="dbuser", default=os.getenv('CANDLEPIN_DB_USER', 'candlepin'),
+                  help="Database user. When missing environment variable 'CANDLEPIN_DB_USER' is read. Defaults to 'candlepin'.")
     parser.add_option("-d", "--database",
-                  dest="db", default="candlepin",
-                  help="the database name to use")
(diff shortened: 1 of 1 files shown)


Ceiu:
What's driving this request? I'm not a huge fan of using environment variables as input into a script like this, nor having *some* things work via environment variables while others cannot.

Januson:
@Ceiu I asked @kahowell. It is a request from ehelms for containerization efforts.

Ceiu:
After speaking with ehelms to figure out what they need out of cpsetup, I heard a couple things:

- There are places where they are needing to doubly-specify config options or passwords, as they need to modify both the config file(s) *and* pass in parameters to cpsetup and cpdb. The environment variable thing sounds like a workaround to a limitation on how this is currently designed.
- There are some functions they need that are not currently performed by cpsetup or cpdb, requiring them do call out to liquibase directly to properly setup a container

That said, I think what we have here only loosely covers their issues, and only in a quick-and-dirty way. I'm okay with merging this as-is to get them something more immediately, but only on the condition that we add a card to revert this patch and then better address the problems they're running into (as well as maybe some of the TODO's we have all over this area). That would cover:

1) Skipping writing a config file for cases where the config file will be generated/provided elsewhere
2) Adding functionality to allow cpsetup or cpdb to read a config file for its operating config
3) Adding the missing functionality they're currently getting from liquibase itself

@kahowell Thoughts?

kahowell:
@Ceiu sounds good to me... feel free to create cards on the backlog for follow-on work.

Ceiu:
ack

PR #2209, opened 2019-01-15
[F] Moved the amqp management  and message sending out of JobManager [ENT-1090]

Ceiu:
- Added a new interface, JobMessageDispatcher, to allow for multiple
  implementations of the message sending backend without needing to
  modify JobManager itself
- Moved the AMQP broker connection management out of the JobManager
  and into the new ArtemisJobMessageDispatcher class

### DIFF
--- a/server/src/main/java/org/candlepin/async/impl/ArtemisJobMessageDispatcher.java
+++ b/server/src/main/java/org/candlepin/async/impl/ArtemisJobMessageDispatcher.java
@@ -1,0 +1,117 @@
+/**
+ * Copyright (c) 2009 - 2019 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.async.impl;
+
+import org.candlepin.async.JobMessage;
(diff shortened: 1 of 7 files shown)


Ceiu:
Will need more work pending on discussions surrounding the overlap in functionality between the ActiveMQConnection object and a lot of the broker management being performed in the Artemis connection classes.

server/src/main/java/org/candlepin/async/impl/ArtemisJobMessageDispatcher.java
```diff
@@ -0,0 +1,154 @@
...
+     * Fetches the current client session, creating and intializing a new instance as necessary.
+     *
+     * @return
+     *  the current ClientSession instance for this job manager
+     */
+    ClientSession getClientSession() throws Exception {
+        if (this.session == null || this.session.isClosed()) {
+            log.debug("Creating new ActiveMQ session for async job messages...");
+
+            ClientSessionFactory csf = this.getClientSessionFactory();
```
mstead: This is where i think the connection would come into play:
`this.connection.getSenderSession()`

mstead:
This is looking pretty good. I'll be interested in seeing the connection changes.

server/src/main/java/org/candlepin/async/impl/ArtemisJobMessageDispatcher.java
mstead: You could also bundle the session/producer into an object and retrieve those straight from the connection as well. This could be very similar to what we do in EventMessageSender the internal class that we use in EventSinkImpl.

Anyway, something to consider.

mstead:
> Added another commit with the new connection object. Take a look at that and we can decide if this needs more work or not.

This looks good to me. Once this is merged it should be pretty simple for me to integrate this with the receiver bits I'm working on. Let me know when this is squashed and ready to merge.

As an aside, the new job framework seems to be lacking test coverage. We should probably get some test coverage around the code soon so that we don't accumulate too much debt. Thoughts?

Ceiu:
> As an aside, the new job framework seems to be lacking test coverage. We should probably get some test coverage around the code soon so that we don't accumulate too much debt. Thoughts?

I was thinking this as well; Was going to mention it to you next we spoke.

PR #2210, opened 2019-01-17
[gradle] Move the rspec custom task to buildSrc

barnabycourt:
In order to optimize compile time move the Rspec custom task class to buildSrc so that it will be compiled once & then cached between build runs.


Ceiu:
Looks good to me, but I do not have the means to manually test this at the moment (I think). Given my lack of familiarity with the platform, I would feel more comfortable with another ack on this than just my own.

barnabycourt:
@Ceiu To manually test you can use the deploy script or run "../gradlew rspec" from the candlepin/server folder on any system with enough ruby installed to run the rspec tests.

Ceiu:
ah ha
runs great locally; ack

PR #2212, opened 2019-01-18
[M] Added additional spec tests for pool refresh scenarios

Ceiu:
Added some additional spec tests in an attempt to make a reproducer for a reported issue. Didn't end up catching said issue, but the tests themselves still have some value.


wottop:
ack

PR #2213, opened 2019-01-22
1520502: Move to syslog/journald[ENT-397]

Januson:
Candlepin root logger now  logs to the syslog. All log files (access.log, audit.log, candlepin.log and error.log) are rotated daily. Old log files are rotated in a format "${filename}.log-${YYYY-MM-dd}.gz"

- Configure Candlepin to log into the syslog
- Replace logrotate with Logback rolling for audit.log, candlepin.log and error.log
- Remove unused property AUDIT_LOG_FILE from ConfigProperties
- Fix rotation of Tomcat's access.log


Ceiu:
ack

PR #2214, opened 2019-01-22
[2.4] 1668454: Fix problems with CPM.deletePools with large datasets

Ceiu:
- Fixed several queries which would fail when presented with large
  collections of inputs
- Reduced the chunk size of the cleanupExpiredPools operation to
  avoid hitting additional query size limits elsewhere


wottop:
ack

PR #2218, opened 2019-01-23
Added ability to receive async job messages from Artemis

mstead:
* ActiveMQSessionFactory now provides client sessions for all message
  sending/receiving in candlepin (Jobs/Events)
* A single MessageSource is now used for both Event and Job Messages
  + Message receivers are now added to the event source via an injected
    factory allowing for customization down the road.
* Refactored the receiver class hierarchy so there was a common base class
* Implemented a job message receiver to process messages when they hit the
  queue.
  + If the receiver encounters a PreJobExecutionException the message is
    considered to be no longer valid and is discarded.
  + An assumption is made that if any exception happens while executing
    the job, it will be added to the AsyncJobStatus result and
    JobManager.executeJob will return this status with the failure info.
    The message is then discarded.
* Created a temporary resource and test class to allow us to manually test
  sending/receiving/executing jobs. We will remove them when the feature is
  complete.
* ActiveMQStatusMonitor now manages its own connection for connection testing
  only.
* ActiveMQSessionFactory now has the ability to create transacted
  client sessions as this was needed for event message sending.

## Testing
1. Enable logging of the job message receiver.
```
log4j.logger.org.candlepin.async.JobMessageReceiver=DEBUG
```
2. Deploy candlepin using this PR.

3. Use the test resource to send a job message. Tail the logs to see that the message was received.
```bash
curl -X PUT -k -u admin:admin https://localhost:8443/candlepin/async/test
```
4. Changes were made to the Artemis connections for sending and receiving. Some additional things to check are:
  * Ensure that both artemis and qpidd can be restarted (will need to use the external artemis broker).
    * Deploy with 'bin/deploy -gtaqb'
  * Ensure that both event and job messaging are working.
    * for events, the easiest way is to just check that records are being added to cp_events.


Ceiu:
ack

As you noted, there are a few real minor things to update yet (refactors, repackaging, etc.), but given the state of this feature, we have more than enough time to do those in other future PRs.

PR #2219, opened 2019-01-23
1556932: Index to make hypervisor uuid lookup faster in Postgresql [E…

wottop:
…NT-390] [2.5]


barnabycourt:
The rspec failure is unrelated to this index.

PR #2220, opened 2019-01-25
[M] 1650992: Increased certificate key size to 4096 bits (ENT-1069)

Ceiu:
Testing notes:
Several rspec tests cover the CRL functionality, specifically the cert_revocation_spec, but also the client_v1_size_spec. At one point I saw failures in the CRL processing (unexpected EOL), but after regenerating a new CRL, no amount of switching between 2048- and 4096-bit builds could re-trigger them. May be unrelated, but something to be aware of while testing.

Additionally, it's important to verify that subscription manager can perform various actions against this branch, including status, register and attach; both with a newly registered system and a system that was already registered against CP with a 2048-bit key size.

When this branch is confirmed safe, the other two PRs will go up for the 2.5 and 2.3 branches.

### DIFF
--- a/server/src/main/java/org/candlepin/pki/impl/ProviderBasedPKIUtility.java
+++ b/server/src/main/java/org/candlepin/pki/impl/ProviderBasedPKIUtility.java
@@ -65,7 +65,7 @@ public abstract class ProviderBasedPKIUtility implements PKIUtility {
     private static Logger log = LoggerFactory.getLogger(ProviderBasedPKIUtility.class);
 
     // TODO : configurable?
-    public static final int RSA_KEY_SIZE = 2048;
+    public static final int RSA_KEY_SIZE = 4096;
 
     protected CertificateReader reader;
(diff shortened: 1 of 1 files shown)


wottop:
So far: [SubMan and Candlepin are on the same machine]
Registers the machine against master CP and set to DEFAULT
Passes against master when set to DEFAULT
Fails against master when then set to FUTURE
I have to change back to DEFAULT or Candlepin will not restart after deploying your branch.
Then it fails again with the original id cert 
Cannot register anew when set to FUTURE: Error: CA certificate for subscription service has not been installed. [It is there and was working before].

Should I try this again with separate machines so candlepin is not running under FUTURE?

Ceiu:
Nah, I think that's a nack, unfortunately. I was seeing odd issues that magically resolved themselves which was already giving me concerns, and you hitting more testing failures tells me there's more to this than just increasing the bit size.

I'll close this branch and update the card tomorrow with my comments and your results.

PR #2221, opened 2019-01-29
1651651: Unneeded guest migration call leads to exception

barnabycourt:


Ceiu:
visual ack

awood:
Looks good to me also.  +1 once the tests pass.

barnabycourt:
The tests that failed are known bad on this branch. Merging per comment from Al

PR #2222, opened 2019-02-05
[M] Added the org_migrator utility (ENT-1110)

Ceiu:
- Added the new organization migration utility, which can export
  a single org's data to an archive to be re-imported later or
  migrated to another database
- Added the cp_connectors utility, which provides simple setup
  of various backends for communicating with Candlepin

More work is eventually needed here, but this gets us most of the way there.


wottop:
ack

PR #2223, opened 2019-02-06
[M] 1667238,1655703: Increase syspurpose impact on auto-attach; ENT-1137,ENT-1073

nikosmoum:
- Pool priority (and consequently, syspurpose attribute matches) will
now have a higher impact than a virt_only match when breaking ties
during auto-attach.
- Syspurpose attributes will now always have a higher impact than
all other attributes (requires_host,virt_only,ram,cores,sockets,
cpu,vcpu) during auto-attach prioritization scoring.
- A match of the usage attribute will now have a higher impact than
not having a role, when the consumer has not specified any role
(1669185).
- Added helpful debug logging in the auto-attach rules code.

### DIFF
--- a/server/src/main/resources/rules/rules.js
+++ b/server/src/main/resources/rules/rules.js
@@ -1,4 +1,4 @@
-// Version: 5.31
+// Version: 5.32
 
 /*
  * Default Candlepin rule set.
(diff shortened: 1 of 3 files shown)


server/src/main/resources/rules/rules.js
```diff
@@ -477,19 +477,23 @@ function get_mock_ent_for_pool(pool, consumer) {
...
-        'products': 20,
-        'roles': 10,
-        'addons': 5,
-        'support_level': 3,
-        'usage': 1
+        'products': 5600,
+        'roles': 2800,
+        'addons': 1400,
+        'support_level': 700,
+        'usage': 350
```
Ceiu: Out of curiosity, where did these values come from?

server/src/main/resources/rules/rules.js
```diff
@@ -502,53 +506,41 @@ function get_pool_priority(pool, consumer) {
         var null_rule_score = 0;
         var mismatch_rule_score = 0;
 
+        log.debug("Evaluating attribute " + attr + " with weight " + attrs[attr]);
```
Ceiu: Even though this is in the Javascript, the logger still supports some of the string building/formatting placeholders we use in Java. I think there may be a parameter limit or something to that effect, but for the most part, that syntax can (and probably should) be used.

server/src/main/resources/rules/rules.js
nikosmoum: They were the initial values considered during the design, but for unknown reasons they were not used: https://docs.google.com/document/d/1q-gMSby30mDsTu0Yjlb3yV3Tt_x9KW_yZSY39lNsK0Q/edit#

But they are not entirely random, the relationship between the previous and the next weight is important, as the comment above them explains. If you add up all the max scores of all the attributes that are less important than usage(requires_host,virt_only,ram,cores,sockets,vcpu), they make up 330 points (150+100+20+20+20+20), so usage cannot be less than that sum. Same for support_level (330 + 350 < 700), and every one after that.

The idea is that each attribute is at least as important as all the attributes below it, combined.

server/src/main/resources/rules/rules.js
nikosmoum: Changed the log statements to use string formatting

Ceiu:
Did this PR also need any other changes for the rules syncing?

nikosmoum:
@Ceiu No, the 2.3 branch was missing a couple of changes, which were added to the PR: https://github.com/candlepin/candlepin/pull/2224. The master branch is up to date.

Ceiu:
ack

PR #2226, opened 2019-02-11
[2.3] 1674415: changed the possible syspurpose status states

nikosmoum:


Ceiu:
ack

PR #2227, opened 2019-02-11
[2.5] 1674419: changed the possible syspurpose status states

nikosmoum:


kahowell:
@jturel FYI, this PR backports some changes to candlepin-2.5. Given that Katello is consuming candlepin-2.5, I think this requires some (small) changes in Katello. So once we merge this PR, Katello will require updates upon pulling in an updated candlepin-2.5 branch.

kahowell:
I'm marking this as in progress; we'll coordinate making the change with Katello, so we can find a good timing for the change.

PR #2228, opened 2019-02-11
[M] 1601517: Added input partitioning to select ConsumerCurator methods

Ceiu:
- Added input partioning to ConsumerCurator.getConsumers, .findByUuids,
  and .findByUuidsAndOwner. This will avoid hitting the query
  parameter limit with large consumer sets.

### DIFF
--- a/server/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/server/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -280,20 +280,38 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
     }
 
     @Transactional
-    public CandlepinQuery<Consumer> findByUuids(Collection<String> uuids) {
-        DetachedCriteria criteria = this.createSecureDetachedCriteria()
-            .add(Restrictions.in("uuid", uuids));
+    public Collection<Consumer> findByUuids(Collection<String> uuids) {
+        Set<Consumer> consumers = new HashSet<>();
 
-        return this.cpQueryFactory.<Consumer>buildQuery(this.currentSession(), criteria);
+        for (List<String> block : this.partition(uuids)) {
+            // Unfortunately, this needs to be a secure criteria due to the contexts in which this
+            // is called.
+            Criteria criteria = this.createSecureCriteria()
+                .add(Restrictions.in("uuid", block));
+
+            consumers.addAll(criteria.list());
+        }
+
+        return consumers;
     }
 
     @Transactional
-    public CandlepinQuery<Consumer> findByUuidsAndOwner(Collection<String> uuids, String ownerId) {
-        DetachedCriteria criteria = DetachedCriteria.forClass(Consumer.class)
-            .add(Restrictions.eq("ownerId", ownerId))
-            .add(Restrictions.in("uuid", uuids));
+    public Collection<Consumer> findByUuidsAndOwner(Collection<String> uuids, String ownerId) {
+        Set<Consumer> consumers = new HashSet<>();
 
-        return this.cpQueryFactory.<Consumer>buildQuery(this.currentSession(), criteria);
+        javax.persistence.Query query = this.getEntityManager()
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -280,20 +280,38 @@ public Consumer findByUuid(String uuid) {
...
 
     @Transactional
-    public CandlepinQuery<Consumer> findByUuids(Collection<String> uuids) {
-        DetachedCriteria criteria = this.createSecureDetachedCriteria()
-            .add(Restrictions.in("uuid", uuids));
+    public Collection<Consumer> findByUuids(Collection<String> uuids) {
+        Set<Consumer> consumers = new HashSet<>();
 
-        return this.cpQueryFactory.<Consumer>buildQuery(this.currentSession(), criteria);
+        for (List<String> block : this.partition(uuids)) {
```
Januson: Why is block unused? Isn't it then doing the same thing on every iteration?

server/src/test/java/org/candlepin/model/ConsumerCuratorTest.java
```diff
@@ -314,10 +310,50 @@ public void testGetConsumersNullInput() {
...
     }
 
+    @Test
+    public void testGetConsumerInputPartioning() {
+        int partitionBlockSize = 5;
+
+        ConsumerCurator curator = Mockito.spy(consumerCurator);
+        when(curator.getInBlockSize()).thenReturn(partitionBlockSize);
+
+        Map<String, Consumer> consumers = new HashMap<>();
```
Januson: This test setup is quite noisy and since it is repeated in multiple tests we could help readability by extracting it to methods.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: Fixed

server/src/test/java/org/candlepin/model/ConsumerCuratorTest.java
Ceiu: I had considered it, but ended up not for the following reasons:

- Only two methods use it
- The point at which you pull common code out would only catch the consumer creation loop, having an input for the curator and output of the map. This then requires that the tests know the details of how the UUIDs are generated for them to work properly when generating the ID/UUID lists for method input, strongly coupling the two together.
- Getting *all* of the setup would require additional inputs of the "input" and "expected" sets so they can be populated for the call and verification steps. This basically ties them exactly to working in this one very specific way, and only serves to divide the code for the sake of dividing the code.

I get the DRY arguments, but I would argue that context switches to understand why a given setup is being performed, and convoluted method I/O doesn't make code easier to read in this case.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -317,15 +335,20 @@ public Consumer getConsumer(String uuid) {
      * @return
      *  A query to fetch the consumers with the specified consumer IDs
      */
-    public CandlepinQuery<Consumer> getConsumers(Collection<String> consumerIds) {
+    public Collection<Consumer> getConsumers(Collection<String> consumerIds) {
+        Set<Consumer> consumers = new HashSet<>();
+        javax.persistence.Query query = this.getEntityManager()
+            .createQuery("SELECT c FROM Consumer c WHERE c.id IN (:cids)");
```
barnabycourt: Should this be loaded using session.byMultipleIds(this.entityType()).enableSessionCheck(true).multiLoad(<id list>) similarly to what we are doing in the ProductCurator so the partitioning & checking from the cache can be offloaded to hibernate?

Ceiu: The difference in cache usage shouldn't be the same, but I made the change so we can offload the partitioning bit to Hibernate and not worry about yet another problem with the IN operator.

PR #2224, opened 2019-02-06
[2.3] 1669199,1656876: Increase syspurpose impact on auto-attach; ENT-1137,ENT-1073

nikosmoum:
- Pool priority (and consequently, syspurpose attribute matches) will
now have a higher impact than a virt_only match when breaking ties
during auto-attach.
- Syspurpose attributes will now always have a higher impact than
all other attributes (requires_host,virt_only,ram,cores,sockets,
cpu,vcpu) during auto-attach prioritization scoring.
- A match of the usage attribute will now have a higher impact than
not having a role, when the consumer has not specified any role
(1673329).
- Added helpful debug logging in the auto-attach rules code.

PR #2231, opened 2019-02-12
1677340: An existing entitlement with exempt sla will block other auto attaches. [ENT-1158]

wottop:
Also corrects logging compatibility issue with rules in Candlepin 2.1


server/src/main/resources/rules/rules.js
```diff
@@ -1,4 +1,4 @@
-// Version: 5.31
+// Version: 5.32
```
nikosmoum: Not a review, but as a general note: there is currently another PR with rules changes (https://github.com/candlepin/candlepin/pull/2223), which also increases the version to 5.32, so whichever of these gets merged first, the other will have to make sure to update it to 5.33. Similarly if this PR also gets patched into the candlepin-2.3-HOTFIX branch, as there is a PR for that branch too (https://github.com/candlepin/candlepin/pull/2224).

wottop:
The spec failure is unrelated to this PR. Please ignore.

PR #2233, opened 2019-02-19
1659345: Hypervisor reconciliation uses different attribute

wottop:
I initially attempted to back port all the changes Vritant had made in this area. There were too many changes necessary for that to work.

I instead updated it to use the properly named attribute.


wottop:
Fixing this in 2.3 has not yet been deemed important. Will hold it here until that is decided.

wottop:
May reopen if it is deemed necessary in this version of Candlepin

PR #2234, opened 2019-02-19
[M] Addressed issues surrounding locking with unmanaged entities

Ceiu:
- Fixed a couple bugs with AbstractHibernateCurator.lockAndLoad
  not properly handling entities which have been deleted or
  created, but not yet flushed
- lockAndLoad now throws an exception if provided an entity which
  is immutable
- Removed the JPA criteria query building in favor of Hibernate's
  native MultiIdentifierLoadAccess entity fetching
- Updated several unit tests in accordance with the above changes,
  and added new tests for the created and deleted cases

### DIFF
--- a/server/src/main/java/org/candlepin/controller/OwnerManager.java
+++ b/server/src/main/java/org/candlepin/controller/OwnerManager.java
@@ -223,12 +223,15 @@ public class OwnerManager {
         }
 
         // Lock the owner
-        owner = ownerCurator.lockAndLoad(owner);
+        Owner locked = ownerCurator.lockAndLoad(owner);
+        if (locked == null) {
+            throw new IllegalStateException("Unable to obtain exclusive lock on owner: " + owner);
+        }
 
         // Fetch the upstream list and mode
-        String upstreamList = adapter.getContentAccessModeList(owner.getKey());
-        String upstreamMode = adapter.getContentAccessMode(owner.getKey());
-        String currentMode = owner.getContentAccessMode();
+        String upstreamList = adapter.getContentAccessModeList(locked.getKey());
+        String upstreamMode = adapter.getContentAccessMode(locked.getKey());
+        String currentMode = locked.getContentAccessMode();
 
         // This shouldn't happen, but in the event our upstream source is having issues, let's
         // not put ourselves in a bad state as well.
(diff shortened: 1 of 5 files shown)


Ceiu:
If this gets cleared for merge, we should also backport this to 2.3, 2.4 and 2.5.

server/src/main/java/org/candlepin/controller/OwnerManager.java
```diff
@@ -288,15 +291,18 @@ public void refreshContentAccessMode(OwnerServiceAdapter adapter, Owner owner) {
     @Transactional
     public void refreshOwnerForContentAccess(Owner owner) {
         // we need to update the owner's consumers if the content access mode has changed
-        owner = ownerCurator.lockAndLoad(owner);
+        Owner locked = ownerCurator.lockAndLoad(owner);
+        if (locked == null) {
+            throw new IllegalStateException("Unable to obtain exclusive lock for owner: " + owner);
+        }
```
mstead: Checking if the locked entity is null and throwing an ISE seems to be a common pattern. Does it make sense to have Curator.lockAndLoad(Entity e)  throw the exception instead of having to do the null check wherever it is called?

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -1175,43 +1153,15 @@ protected E lockAndLoadById(Class<E> entityClass, Serializable id) {
...
-                    executable.setParameter(param, block);
-                    result.addAll(executable.getResultList());
-                }
+                result.addAll(((Session) session)
+                    .byMultipleIds(this.entityType())
+                    .with(new LockOptions(LockMode.PESSIMISTIC_WRITE))
+                    .with(CacheMode.REFRESH)
+                    .enableSessionCheck(false)
+                    .enableOrderedReturn(false)
+                    .multiLoad(new ArrayList(idSet)));
```
mstead: With this new query, we no longer have to partition the ids and make multiple lookups?

Ceiu: Nope, it handles that for us internally.

server/src/main/java/org/candlepin/controller/OwnerManager.java
Ceiu: No, because it's not a universal thing. In a couple of those cases, a null output could probably fall back to "do nothing" behavior safely, but given that it was written with the intent to always do something, I felt an exception would be a better drop-in replacement than silently skipping it. At least for now.

There are a couple cases (as you noticed in other comments) where we don't care if the object comes back or not.

mstead: ACK

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
mstead: Beauty! :+1:

PR #2238, opened 2019-02-25
Execute Jobs When Message Is Received From Artemis[ENT-1089]

Januson:
- Implement construction of job from job key
- Update job state during execution(QUEUED->RUNNING->COMPLETED/FAILED)
- Handle job execution transactions
- Setup MDC for job logging
- Update TestJob1 so that it can sleep/fail according to the given parameters
- Add job_exec_source column to the cp2_async_jobs table

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -14,25 +14,27 @@
  */
 package org.candlepin.async;
 
+import com.google.inject.Injector;
+import com.google.inject.persist.UnitOfWork;
+import org.candlepin.audit.EventSink;
 import org.candlepin.auth.Principal;
 import org.candlepin.common.filter.LoggingFilter;
+import org.candlepin.guice.CandlepinRequestScope;
 import org.candlepin.guice.PrincipalProvider;
 import org.candlepin.model.AsyncJobStatus;
 import org.candlepin.model.AsyncJobStatus.JobState;
 import org.candlepin.model.AsyncJobStatusCurator;
+import org.candlepin.pinsetter.core.model.JobStatus;
 import org.candlepin.util.Util;
-
-import org.apache.log4j.MDC;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.HashMap;
-import java.util.Map;
+import org.slf4j.MDC;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
-
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
 
 
 /**
@@ -40,16 +42,18 @@ import javax.inject.Singleton;
  */
 @Singleton
 public class JobManager {
+
     private static Logger log = LoggerFactory.getLogger(JobManager.class);
 
     /** Stores our mapping of job keys to job classes */
     private static Map<String, Class<? extends AsyncJob>> jobs = new HashMap<>();
 
+    private final AsyncJobStatusCurator jobCurator;
+    private final JobMessageDispatcher dispatcher;
+    private final CandlepinRequestScope candlepinRequestScope;
+    private final PrincipalProvider principalProvider;
+    private final Injector injector;
 
-    private AsyncJobStatusCurator jobCurator;
-    private JobMessageDispatcher dispatcher;
-
-    private PrincipalProvider principalProvider;
 
 
     /**
@@ -125,18 +129,21 @@ public class JobManager {
         return jobs.get(jobKey);
     }
 
-
     /**
      * Creates a new JobManager instance
      */
     @Inject
-    public JobManager(AsyncJobStatusCurator jobCurator, JobMessageDispatcher dispatcher,
-        PrincipalProvider principalProvider) {
-
-        this.jobCurator = jobCurator;
-        this.dispatcher = dispatcher;
-
-        this.principalProvider = principalProvider;
+    public JobManager(
+        final AsyncJobStatusCurator jobCurator,
+        final JobMessageDispatcher dispatcher,
+        final PrincipalProvider principalProvider,
+        final CandlepinRequestScope scope,
+        final Injector injector) {
+        this.jobCurator = Objects.requireNonNull(jobCurator);
+        this.dispatcher = Objects.requireNonNull(dispatcher);
+        this.candlepinRequestScope = Objects.requireNonNull(scope);
+        this.principalProvider = Objects.requireNonNull(principalProvider);
+        this.injector = Objects.requireNonNull(injector);
     }
 
     /**
@@ -165,7 +172,7 @@ public class JobManager {
 
         // Add environment-specific metadata
         job.setOrigin(Util.getHostname());
-        job.setCorrelationId((String) MDC.get(LoggingFilter.CSID));
+        job.setCorrelationId(MDC.get(LoggingFilter.CSID));
 
         Principal principal = this.principalProvider.get();
         job.setPrincipal(principal != null ? principal.getName() : null);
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/candlepin/async/StatusUpdatingJob.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public Object execute(final JobExecutionContext jdata) throws JobExecutionException {
+        setRunning();
+        try {
+            final Object result = job.execute(jdata);
+            updateStatus(AsyncJobStatus.JobState.COMPLETED, result);
+            return result;
+        }
+        catch (JobExecutionException e) {
```
barnabycourt: Is there any exception where the job should not be recorded as Failed?

server/src/main/java/org/candlepin/async/LoggingJob.java
```diff
@@ -0,0 +1,66 @@
...
+        MDC.put("requestType", "job");
+        if (jdata != null) {
+            final JobDataMap map = jdata.getJobData();
+
+            String requestUuid = null;
+            String orgKey = null;
+            String orgLogLevel = null;
+
+            if (map != null) {
+                requestUuid = map.getAsString("requestUuid");
```
barnabycourt: Do we have a constant for "requestUuid" somewhere?

server/src/main/java/org/candlepin/async/LoggingJob.java
```diff
@@ -0,0 +1,66 @@
...
+public class LoggingJob implements AsyncJob {
+
+    private final AsyncJob job;
+
+    public LoggingJob(final AsyncJob job) {
+        this.job = requireNonNull(job);
+    }
+
+    @Override
+    public Object execute(final JobExecutionContext jdata) throws JobExecutionException {
```
barnabycourt: I think this should be setting the CorrelationId from the AsyncJobStatus on the logging context. Perhaps this should be done somewhere else though.

server/src/main/java/org/candlepin/async/UnitOfWorkJob.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.async;
+
+import com.google.inject.persist.UnitOfWork;
+
+import static java.util.Objects.requireNonNull;
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -188,12 +110,9 @@ private AsyncJobStatus buildJobStatus(JobBuilder builder) {
      * in the queue or currently executing, a new job will not be queued and the existing job's
      * job status will be returned instead.
      *
-     * @param builder
-     *  A JobBuilder instance representing the configuration of the job to queue
-     *
-     * @return
-     *  an AsyncJobStatus instance representing the queued job's status, or the status of the
-     *  existing job if it already exists
+     * @param builder A JobBuilder instance representing the configuration of the job to queue
```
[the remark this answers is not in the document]
Januson: Done. This was auto-formatter's doing. I am ok with both formats.

mstead:
> There are some interesting implementation choices here, but some that don't really fit with the intention of the job infrastructure/design. Namely, using jobs themselves to perform required job environment initialization/configuration is a corruption of the intent of the jobs. Moreover, chaining them together in a way that forces context to be shared between several classes is bad for readability and starts coding us into corners when one part of the chain changes in a way that breaks something else along it (see deletePools or revokeEntitlements for examples of how badly this can get over time). While doing all of the setup in the JobManager may feel excessively monolithic, that's just where the code belongs; Jobs are not decorators for other jobs.

I agree with @Ceiu on this. The job manager should be responsible defining the 'process' of running a job -- setting up the UOW and sending event messages. If this is encapsulated in a job, it becomes increasingly difficult to manage.

The running of jobs should be streamlined, and IMO this behavior should be defined by the manager when the job is executed -- not in the job itself. A job should only be concerned about the work that is to be done, and nothing more.

server/src/main/java/org/candlepin/async/LoggingJob.java
Januson: Depends on whether we want CorellationId for the whole request or just the job execution. I think it should be set when the message is received.

server/src/main/java/org/candlepin/async/LoggingJob.java
Januson: I was looking around since the other two values had a constant but haven't found any. I can create one. I am just not sure where to put it. Is there any common place for constants?

server/src/main/java/org/candlepin/async/JobKey.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Represents the binding from jobKey to the jobClass. Job key has to be
+ * registered before it can be used. Attempts to use unregistered key will
+ * result in {@link IllegalArgumentException}.
+ */
+public class JobKey {
```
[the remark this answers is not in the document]
Januson: Idea was to break the two way static dependency. It might not be needed anymore since everything is now in JobManager.

server/src/main/java/org/candlepin/async/StatusUpdatingJob.java
Januson: From looking at the old job implementation and KingpinJob it seems there is no such case.

Ceiu: Let's have this catch any exception to mark the job as failed.

server/src/main/java/org/candlepin/async/JobKey.java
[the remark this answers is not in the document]
Ceiu: Yeah, let's move this back to the JobManager

server/src/main/java/org/candlepin/async/LoggingJob.java
Ceiu: I may be misunderstanding here, but the objective of the correlation ID/correlated service ID is to track all Candlepin operations (log output) associated with a given operation from the client. As part of that goal, the CSID should be set during the execution of a job that was created from a request with a non-null CSID.

Januson: That's what I mean. Message is part of the user request so CSID should be set when the message is received right?

server/src/main/java/org/candlepin/async/JobKey.java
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/StatusUpdatingJob.java
Januson: Done.

server/src/main/java/org/candlepin/async/LoggingJob.java
Ceiu: Kind of. When we queue a job, if there's a CSID set in the MDC, then we pull it from there and put it in the JobStatus. From there, when the job is executed, we need to set it again as the job is running in a different context.

Ceiu:
I've approved the changes, though there is still some work to be done on this particular task. As such, I've added two additional cards on the backlog to address how org-specific logging configuration should be done (currently using a copy/paste of the old code, which is quirky), and addressing the transaction concerns we have with a unit-of-work wrapping our session management for us.

PR #2241, opened 2019-02-26
[2.5] 1683305: Synchronize java.util.Date serializations; ENT-1183

nikosmoum:

### DIFF
--- a/server/src/test/java/org/candlepin/resteasy/JsonProviderTest.java
+++ b/server/src/test/java/org/candlepin/resteasy/JsonProviderTest.java
@@ -14,8 +14,10 @@
  */
 package org.candlepin.resteasy;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import org.candlepin.common.config.Configuration;
 import org.candlepin.jackson.ProductCachedSerializationModule;
@@ -26,6 +28,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationConfig;
 import com.fasterxml.jackson.databind.SerializationFeature;
 
+import org.junit.ComparisonFailure;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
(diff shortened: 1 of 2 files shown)


server/src/test/java/org/candlepin/resteasy/JsonProviderTest.java
```diff
@@ -43,6 +49,13 @@
     @Mock private Configuration config;
     @Mock private ProductCurator productCurator;
 
+    private AtomicBoolean failed;
+
+    @Before
+    public void setUp() {
+        failed = new AtomicBoolean(false);
```
Ceiu: You should be able to define this within the scope of the test using it; potentially declaring it as final for Java8 and earlier to reference it within the inner class.

server/src/test/java/org/candlepin/resteasy/JsonProviderTest.java
```diff
@@ -68,6 +81,45 @@ public void serializedDateDoesNotIncludeMilliseconds() throws JsonProcessingExce
...
+    public void canSerializeDatesWithLargeTimestampsConcurrently() {
+        JsonProvider provider = new JsonProvider(config,
+            new ProductCachedSerializationModule(productCurator));
+        ObjectMapper mapper = provider.locateMapper(Object.class, MediaType.APPLICATION_JSON_TYPE);
+
+        ExecutorService ex = Executors.newFixedThreadPool(1000);
+        for (int i = 0; i < 500000; i++) {
+            ex.execute(() -> {
+                try {
+                    mapper.writeValueAsString(new Date(getRandomTimestampWithMilliSeconds()));
```
Ceiu: Should we also verify that the value written is as expected?

server/src/test/java/org/candlepin/resteasy/JsonProviderTest.java
nikosmoum: Done.

server/src/test/java/org/candlepin/resteasy/JsonProviderTest.java
nikosmoum: Done. This is actually good because running this test with the old code, there's quite a few dates that are not serialized accurately (I suspect because of similar threading issues), while the new code fixes those (all dates are now serialized accurately).

PR #2242, opened 2019-02-26
Awood/dependency fixup

awood:
Fix a dependency issue and fix some tests that were failing due to a cert with a too short key.


awood:

nikosmoum:
I know this has already been merged, but since I'm reading up on our build process, I thought we need to update the pom.xml files every time we change dependencies in the buildfile, or our builds will get messed up?

Ceiu:
Yes and no, though probably more yes than no. While the pom.xml isn't updated, it gets automatically updated as part of our other systems (such as deploy and build). So it's not strictly necessary, but it should be done since it'll cause pom.xml to constantly look like it's been modified.

awood:
Crap.  I forgot to check in the updated pom.  I will do that.

awood:
See #2245

PR #2244, opened 2019-02-26
Make cpdb and cpsetup python3 compatible

zjhuntin:
These are the easiest pieces that I could find to change. Some help/guidance with setting up how test these would be much appreciated.

### DIFF
--- a/server/code/setup/cpdb
+++ b/server/code/setup/cpdb
@@ -16,13 +16,16 @@
 
 """ A CLI utility for managing the Candlepin database. """
 
+from __future__ import print_function
+from optparse import OptionParser
 
-import commands
 import sys
 import os
 
-from optparse import OptionParser
-from subprocess import Popen, PIPE
+try:
+    from commands import getstatusoutput
+except ImportError:
+    from subprocess import getstatusoutput
 
 if os.path.exists('/usr/sbin/tomcat') and not os.path.exists('/usr/sbin/tomcat6'):
     TOMCAT = 'tomcat'
@@ -33,7 +36,7 @@ JBOSS_CLASSPATH = "/var/lib/jbossas/server/production/deploy/candlepin.war/WEB-I
 TOMCAT_CLASSPATH = "/var/lib/" + TOMCAT + "/webapps/candlepin/WEB-INF/classes/"
 
 def run_command(command):
-    (status, output) = commands.getstatusoutput(command)
+    (status, output) = getstatusoutput(command)
     if status > 0:
         error_out(command, status, output)
     return output
@@ -101,10 +104,10 @@ class DbSetup(object):
         # Add in output level. By default, Liquibase defaults to "off"
         liquibase_options += " --logLevel=%s" % ('debug' if self.verbose else 'severe')
 
-        print liquibase_options
+        print(liquibase_options)
 
         output = run_command("liquibase %s migrate -Dcommunity=%s" % (liquibase_options, self.community))
-        print output
+        print(output)
 
 
 class PostgresqlSetup(DbSetup):
@@ -138,7 +141,7 @@ class PostgresqlSetup(DbSetup):
             if self.port:
                 command = "%s -p %s" % (command, self.port)
         command = "%s %s" % (command, self.db)
-        (status, output) = commands.getstatusoutput(command)
+        (status, output) = getstatusoutput(command)
 
         if status > 0 and output.find("already exists") > 0:
             print("Candlepin database already exists, skipping...")
@@ -149,7 +152,7 @@ class PostgresqlSetup(DbSetup):
     def drop(self):
         print("Dropping candlepin database")
         command = "dropdb -U %s %s" % (self.username, self.db)
-        (status, output) = commands.getstatusoutput(command)
+        (status, output) = getstatusoutput(command)
         if status > 0 and output.find('does not exist') > 0:
             return
         elif status > 0:
--- a/server/code/setup/cpsetup
+++ b/server/code/setup/cpsetup
@@ -21,14 +21,24 @@ This script should be idempotent, as puppet will re-run this to keep
 the server functional.
 """
 
+from __future__ import print_function
+from optparse import OptionParser
+
 import sys
-import commands
 import os.path, os
 import socket
 import re
 import time
-import httplib
-from optparse import OptionParser
+
+try:
+    from commands import getstatusoutput
+except ImportError:
+    from subprocess import getstatusoutput
+
+try:
+    from httplib import HTTPConnection
+except ImportError:
+    from http.client import HTTPConnection
 
 CANDLEPIN_CONF = '/etc/candlepin/candlepin.conf'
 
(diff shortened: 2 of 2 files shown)


server/code/setup/cpdb
```diff
@@ -101,10 +101,10 @@ class DbSetup(object):
         # Add in output level. By default, Liquibase defaults to "off"
         liquibase_options += " --logLevel=%s" % ('debug' if self.verbose else 'severe')
 
-        print liquibase_options
+        print(liquibase_options)
```
evgeni: please add `from __future__ import print_function` at the very beginning of this file for best Py2 compatibility.

server/code/setup/cpsetup
```diff
@@ -268,7 +268,7 @@ def main(argv):
     if not options.skipdbcfg:
         write_candlepin_conf(options)
     else:
-        print "** Skipping configuration file setup"
+        print("** Skipping configuration file setup")
```
evgeni: please add `from __future__ import print_function` at the very beginning of this file for best Py2 compatibility.

server/code/setup/cpsetup
```diff
@@ -22,12 +22,12 @@ the server functional.
...
 
 import sys
-import commands
+import subprocess
 import os.path, os
 import socket
 import re
 import time
-import httplib
+import http.client
```
evgeni: this library is Python3 only, but I guess we want this script to work across all Pythons?

```python
try:
    from httplib import HTTPConnection
except ImportError
    from http.client import HTTPConnection
```

server/code/setup/cpsetup
```diff
@@ -119,7 +119,7 @@ class TomcatSetup(object):
 
     def wait_for_startup(self):
         print("Waiting for tomcat to restart...")
-        conn = httplib.HTTPConnection('localhost:8080')
+        conn = http.client.HTTPConnection('localhost:8080')
```
evgeni: following the above import change, this becomes
```suggestion
        conn = HTTPConnection('localhost:8080')
```

server/code/setup/cpdb
```diff
@@ -17,7 +17,7 @@
 """ A CLI utility for managing the Candlepin database. """
 
 
-import commands
+import subprocess
```
evgeni: `subprocess` doesn't have `getstatusoutput` on Python2, so you'd need something like:

```python
try:
    from commands import getstatusoutput
except ImportError:
    from subprocess import getstatusoutput
```

server/code/setup/cpdb
```diff
@@ -33,7 +33,7 @@ JBOSS_CLASSPATH = "/var/lib/jbossas/server/production/deploy/candlepin.war/WEB-I
 TOMCAT_CLASSPATH = "/var/lib/" + TOMCAT + "/webapps/candlepin/WEB-INF/classes/"
 
 def run_command(command):
-    (status, output) = commands.getstatusoutput(command)
+    (status, output) = subprocess.getstatusoutput(command)
```
evgeni: This is tricky, as the return values of `commands.getstatusoutput` and `subprocess.getstatusoutput` actually differ:

# `commands.getstatusoutput(cmd)`

Execute the string `cmd` in a shell with `os.popen()` and return a 2-tuple `(status, output)`. `cmd` is actually run as `{ cmd ; } 2>&1`, so that the returned output will contain output or error messages. A trailing newline is stripped from the output. The exit status for the command can be interpreted according to the rules for the C function `wait()`.

# `subprocess.getstatusoutput(cmd)`

Return `(exitcode, output)` of executing `cmd` in a shell.

Execute the string `cmd` in a shell with `Popen.check_output()` and return a 2-tuple `(exitcode, output)`. The locale encoding is used; see the notes on Frequently Used Arguments for more details.

A trailing newline is stripped from the output. The exit code for the command can be interpreted as the return code of subprocess.

The function now returns `(exitcode, output)` instead of `(status, output)` as it did in Python 3.3.3 and earlier. exitcode has the same value as `returncode`.

# difference

The difference is that in Python2 we had to pass `status` through [`os.WEXITSTATUS`](https://docs.python.org/2/library/os.html#os.WEXITSTATUS) to get the real exit code, whereas we don't have to do this since Python 3.4. That said, the old code is *technically* wrong (as it won't catch all possible error conditions), but the new one is correct.

See https://github.com/Katello/katello-client-bootstrap/commit/5e7e0130d4581a07beb5c0ecc58c2f3653781300 and https://github.com/Katello/katello-client-bootstrap/commit/d1c308a3b8648cc4c2b12043fdcd7f22371bcd56 for an example how we used that in `katello-client-bootstrap`.

server/code/setup/cpdb
```diff
@@ -138,7 +138,7 @@ class PostgresqlSetup(DbSetup):
             if self.port:
                 command = "%s -p %s" % (command, self.port)
         command = "%s %s" % (command, self.db)
-        (status, output) = commands.getstatusoutput(command)
+        (status, output) = subprocess.getstatusoutput(command)
```
evgeni: same `commands` vs `subprocess` comment as above

server/code/setup/cpdb
```diff
@@ -149,7 +149,7 @@ class PostgresqlSetup(DbSetup):
     def drop(self):
         print("Dropping candlepin database")
         command = "dropdb -U %s %s" % (self.username, self.db)
-        (status, output) = commands.getstatusoutput(command)
+        (status, output) = subprocess.getstatusoutput(command)
```
evgeni: same `commands` vs `subprocess` comment as above

server/code/setup/cpsetup
```diff
@@ -38,7 +38,7 @@ else:
     TOMCAT = 'tomcat6'
 
 def run_command(command):
-    (status, output) = commands.getstatusoutput(command)
+    (status, output) = subprocess.getstatusoutput(command)
```
evgeni: same `commands` vs `subprocess` comment as above

awood:

awood:
@zjhuntin The comments from @evgeni are spot on.  Would you mind addressing them?

zjhuntin:
@evgeni Thanks so much for the in-depth review. 

@awood I made the changes as suggested, I wasn't 100% sure how to address the difference between `exitstatuscode`

server/code/setup/cpsetup
```diff
@@ -22,13 +22,19 @@ the server functional.
...
-import commands
+import subprocess
 import os.path, os
 import socket
 import re
 import time
-import httplib
+
 from optparse import OptionParser
+from __future__ import print_function
```
evgeni: this needs to be the very first import you do in a file.

server/code/setup/cpdb
```diff
@@ -16,13 +16,18 @@
 
 """ A CLI utility for managing the Candlepin database. """
 
+from __future__ import print_function
+from optparse import OptionParser
+from subprocess import Popen, PIPE
```
evgeni: these seem unused?

server/code/setup/cpdb
```diff
@@ -16,13 +16,18 @@
...
 
-import commands
+import subprocess
```
evgeni: this seems unused?

server/code/setup/cpsetup
```diff
@@ -21,14 +21,20 @@ This script should be idempotent, as puppet will re-run this to keep
 the server functional.
 """
 
+from __future__ import print_function
+from optparse import OptionParser
+
 import sys
-import commands
+import subprocess
```
evgeni:
```suggestion
try:
    from commands import getstatusoutput
except ImportError:
    from subprocess import getstatusoutput
```

PR #2246, opened 2019-02-27
[M] Made some improvements to the org_migrator utility

Ceiu:
- It is no longer require to specify the account on import
- Removed an extraneous entry in the help info
- Added some explicit conversion for juggling mysql and
  postgresql exports
- Reworked the binary decoder mechanism to be more generic
- Added support for fetching column types as strings
- Exports will now include column types for each table
- The export operation will no longer attempt to run if the
  org ID resolution fails
- Added the --list operation to list orgs in a given DB

### DIFF
--- a/server/bin/org_migrator.py
+++ b/server/bin/org_migrator.py
@@ -8,7 +8,6 @@ import logging
 from optparse import OptionParser
 import os
 import re
-import sys
 import zipfile
 
 import cp_connectors as cp
@@ -18,31 +17,23 @@ log = logging.getLogger('org_migrator')
 
 
 
-# def get_table_columns(table_name):
-#     columns = None
-#     cursor = None
-
-#     try:
-#         cursor = db.execQuery('SELECT COLUMN_NAME FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME=%s;', (table_name,))
-
-#         for row in cursor:
-#             if columns is None:
-#                 columns = []
-
-#             columns.append(row[0])
-
-#         return columns
-#     finally:
-#         if cursor is not None:
-#             cursor.close()
-
-def get_cursor_columns(cursor):
+def get_cursor_columns(db, cursor):
     if hasattr(cursor, 'column_names'):
         return cursor.column_names
     elif hasattr(cursor, 'description'):
         return [col.name for col in cursor.description]
     else:
         raise Exception('Cannot determine column names for cursor')
+
+
+def get_cursor_column_types(db, cursor):
+    if hasattr(cursor, 'column_names'):
+        return [db.get_type_as_string(col[1]) for col in cursor.description]
+    elif hasattr(cursor, 'description'):
+        return [db.get_type_as_string(col.type_code) for col in cursor.description]
+    else:
+        raise Exception('Cannot determine column types for cursor')
+
 
 def resolve_org(db, org):
     org_id = None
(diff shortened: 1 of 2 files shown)


Ceiu:
I'm not sure if this is ready or not. I feel like there's more that could be done, but I'm debating putting time in now or when we run into the situation that requires it. Namely, importing into differing backends or versions causes some headaches which should be mitigatable by checking the table structure before we import.

server/bin/org_migrator.py
```diff
@@ -778,45 +798,41 @@ def parse_options():
...
+        help="The hostname/address of the database server; defaults to 'localhost'")
     parser.add_option("--port", action="store", default=None,
-            help="The port to use when connecting to the database server")
+        help="The port to use when connecting to the database server")
     parser.add_option("--db", action="store", default='candlepin',
-            help="The database to use; defaults to 'candlepin'")
+        help="The database to use; defaults to 'candlepin'")
     parser.add_option("--file", action="store", default='export.zip',
-            help="The name of the file to export to or import from; defaults to 'export.zip''")
+        help="The name of the file to export to or import from; defaults to 'export.zip''")
```
nikosmoum: There's an extra ' (single bracket) at the end of the sentence

server/bin/org_migrator.py
```diff
@@ -827,34 +843,56 @@ def main():
...
-        log.info('Using backend: %s', db.backend())
-
-        org_id = resolve_org(db, args[0])
-        log.info('Resolved org "%s" to org ID: %s', args[0], org_id)
+        log.info('Using database: %s @ %s', db.backend(), options.host)
 
         if options.act_import:
-            importer = OrgImporter(db, options.file, org_id)
+            # Use the provided argument if the --file is still default
+            if options.file == 'export.zip' and args[0]:
```
nikosmoum: If I run it like this (assuming I have an export.zip in my current dir) `org_migrator.py --import`, this throws `IndexError: list index out of range`. Need to check the length of args first.

server/bin/org_migrator.py
```diff
@@ -18,32 +18,22 @@
...
-
-#             columns.append(row[0])
-
-#         return columns
-#     finally:
-#         if cursor is not None:
-#             cursor.close()
-
-def get_cursor_columns(cursor):
+def get_cursor_columns(db, cursor):
```
nikosmoum: Let's remove the imports of `os` and `sys` at the top of the file since they're not used.

Ceiu: done

server/bin/org_migrator.py
Ceiu: Fixed

server/bin/org_migrator.py
Ceiu: Fixed

server/bin/org_migrator.py
nikosmoum: I get this now:
```
[vagrant@candlepin vagrant]$ ./server/bin/org_migrator.py --import
2019-03-20 15:12:23,688 INFO    org_migrator     Using database: PostgreSQL @ localhost
Traceback (most recent call last):
  File "./server/bin/org_migrator.py", line 903, in <module>
    main()
  File "./server/bin/org_migrator.py", line 860, in main
    if not os.access(options.file, os.R_OK) or not os.isfile(options.file):
AttributeError: 'module' object has no attribute 'isfile'
```
(this is with python2.7)
I think you might need to do `os.path.isfile` instead

PR #2249, opened 2019-03-01
[2.3] 1684508: Remove redundant role & addon pools from stacks; ENT-1199

nikosmoum:
- Now, during auto-attach, stackable pools that provide either
identical or unsatisfactory roles & addons will be removed from
the stack (assuming they also dont provide necessery products).

### DIFF
--- a/server/src/main/resources/rules/rules.js
+++ b/server/src/main/resources/rules/rules.js
@@ -1,4 +1,4 @@
-// Version: 5.33
+// Version: 5.34
 
 /*
  * Default Candlepin rule set.
@@ -2134,6 +2134,7 @@ var Autobind = {
              * are removed.
              */
             validate: function(context) {
+                log.debug("Running validate for stack_id: {}", this.stack_id);
                 var all_ents = this.get_all_ents(this.pools).concat(this.attached_ents);
 
                 if (all_ents.length == 0) {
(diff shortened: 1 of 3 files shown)


server/src/main/resources/rules/rules.js
```diff
@@ -3082,14 +3073,14 @@ var Autobind = {
...
-                    this.get_common_addons(addons, ent_group).length > 0 ||
+                    ent_group.get_common_role(role) != null ||
+                    ent_group.get_common_addons(addons).length > 0 ||
                     (ent_group.attached_ents !== null && ent_group.attached_ents.length > 0)) {
                     valid_groups.push(ent_group);
                     ent_group.remove_extra_attrs(role, addons);
                     ent_group.prune_pools(role, addons);
                 } else {
-                    log.debug("Group " + ent_group.stack_id + " provides no needed products");
+                    log.debug("Group " + ent_group.stack_id + " provides no needed products, roles or addons");
```
Ceiu: Try to get into the habit of using the format syntax rather than concatenation for these.

server/src/main/resources/rules/rules.js
```diff
@@ -2955,22 +2949,19 @@ var Autobind = {
...
                 }
             }
-            group.addons = addons_in_common;
 
-            var role_in_common = this.get_common_role(role, group);
-            if (role_in_common.indexOf(role) !== -1) {
+            if (group.get_common_role(role) != null) {
                 log.debug("get_best_entitlement_groups: Remove role we just satisfied from list of specified roles...");
-                role = "";
+                role = null;
```
Ceiu: :+1:

nikosmoum:
@Ceiu I think the most relevant reproducer comment is this: https://bugzilla.redhat.com/show_bug.cgi?id=1683169#c2

In short:
If you auto-attach with role and/or addons specified, you'll get all of the pools in a stack that provide those role/addons, even if you only need 1 pool.

To explain in more detail:
- You have a stage account with 3 different pools, who all provide the same SKU, and all of them have the same stack id and they stack on an attribute like sockets. They all have available quantity of 1. They all happen to provide the product, role and addons that your system has specified. They're basically identical. (but they don't have to be completely identical, all of them providing the same role and/or addons that the consumer has specified is the key point).
- You have a system that has a product, role and addon specified, and it would be satisfied with a quantity of 1, from any ONE of those 3 pools.
- You auto-attach, but instead get all 3 of them attached, even though you need only 1 (if you remove 2 of them, leaving only 1, and run subman status, you still get a 'valid' status).

In the code, the main problem was in the `prune_pools` method:
- Before, every time we removed temporarily a pool, and then checked if we can still cover the consumer, we did NOT check if the remaining pools in the stack can still cover the consumer's role and addons, but instead, wrongly, we were checking if the pool we just temporarily removed covers the consumer's role/addons: https://github.com/candlepin/candlepin/blob/b4a6c489a4b63c319ec1166f7dc4a3a99b6b10d0/server/src/main/resources/rules/rules.js#L2393
- Now, with the fix, after we temporarily remove a pool, we check if the remaining pools in the stack cover the consumer's role/addons: https://github.com/candlepin/candlepin/blob/cad35b7f68e6bd9d5d5c161b0a33b6697dc2af15/server/src/main/resources/rules/rules.js#L2381-L2382

If I am not explaining clearly, I'd be happy to answer more clarifying questions. Thanks for reviewing!

PR #2251, opened 2019-03-05
Expose useful data on System Purpose Compliance event

jturel:
This is very useful for Katello because we report system purpsose status (matched/mismatched/not specified) on each field. With this additional data we can figure that out. And since it should be identical fields to the /consumers/:uuid/purpose_compliance API we can parse both with the same code in case we need to process compliance status outside of an event which could happen.

I haven't tested this!


kahowell:
ok to test

PR #2258, opened 2019-03-12
[M] 1687535: Better syspurpose matches during auto-attach; ENT-1213

nikosmoum:
- Now during auto-attach, syspurpose attribute mismatches will
have a lower impact, as to only be significant when breaking
ties between otherwise similar pools. Auto-attach will choose
the pool which covers the most syspurpose attributes, even if
that pool comes with other non-specified syspurpose attributes.

### DIFF
--- a/server/src/main/resources/rules/rules.js
+++ b/server/src/main/resources/rules/rules.js
@@ -1,4 +1,4 @@
-// Version: 5.34
+// Version: 5.35
 
 /*
  * Default Candlepin rule set.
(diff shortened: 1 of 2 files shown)


nikosmoum:
Notes for reviewers:

Here is an example auto-attach scenario of what this change is fixing. This scenario is the most extreme case I can think of that will showcase the problem:

- A system has installed product1 and has specified role1, addon1, sla1, and usage1.
- SKU1 provides product1, role2, addon2, sla2, and usage1.
- SKU2 provides product1 only.

Without this fix, auto-attach will attach SKU2, which only covers product1 (even though SKU1 would cover both product1 and usage1), because the other attribute mismatches of SKU1 (role2, addon2, and sla2) have a high enough impact in our algorithm.

With this fix, SKU1 will be attached because it better covers the system (even though we are "overconsuming" role2, addon2 and sla2).
With this fix, the 'mismatch rule' in our algorithm should only have an impact when breaking ties like these:

- A system has installed product1 and has specified role1, addon1, sla1, and usage1.
- SKU1 provides product1, role1, addon1, sla1, and no usage.
- SKU2 provides product1, role1, addon1, sla1 and usage2.

In this case both SKUs can cover the system as well, but the 'mismatch rule' giving some negative score to SKU2 helps attach SKU1, to avoid "overconsumption" of usage2.

Ceiu:
ack

PR #2259, opened 2019-03-12
[2.3] 1687779: Better syspurpose matches during auto-attach; ENT-1213

nikosmoum:
- Now during auto-attach, syspurpose attribute mismatches will
have a lower impact, as to only be significant when breaking
ties between otherwise similar pools. Auto-attach will choose
the pool which covers the most syspurpose attributes, even if
that pool comes with other non-specified syspurpose attributes.


Ceiu:
ack

PR #2260, opened 2019-03-12
Added ability to create testing yum/dnf repositories.

jirihnidek:
* Repositories are created according definition of 'content' in
  test_data.json
* Added definitions of dummy packages to test_data.json
  - Dummy packages are empty packages without any content (no files)
  - No dependencies on anything
  - Easy to test on any platform (RHEL, CentOS 6/7, Fedora, ...)
* Product certificate is created from temporary product using REST API
  - Temporary product contains only one contet and ID of temporary
    product is the same as ID of content
  - Certitifate for temporary product is downloaded
  - Temporary product is removed
* Python script for creating testing repos also contains functions
  for creating GPG keys and signing keys, but it is not used ATM
* It requires more packages on the system running candlepin (https://github.com/candlepin/ansible-role-candlepin will be updated according these requirements):
    - createrepo
    - rpm-build
    - rpm-sign
    - python-requests
    - expect
* To test new option run: `./server/bin/deploy -gtar` on the candlepin server
* Generated RPMs are signed with GPG key
* GPG key has to be protected by passphrase, because rpmsing
  tries to read passphrase despite key is not protected by
  any passphrase
* Tool called expect is used to "type" secret passphrase
* Generated RPMs are cached in rpmbuild directory
* Modified some paths in products
* Modified gpg_url to:
   "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-candlepin"
* It is not possible to use: "gpg_url": "/RPM-GPG-KEY-candlepin",
  because subscription-manager generate following
  gpgkey: http://ip_addr:8080/RPM-GPG-KEY-candlepin
  in redhat.repo and yum/dnf/pkcon/microdnf are not willing
  to download GPG key using http. It is allowed to use only
  https protocol.
* GPG key has to be downloaded and installed on registered
  system to be able to install signed RPMs. Note: yum and dnf
  allow to install RPMs without gpg sign, when --nogpgcheck
  is used, but pkcon nor microdnf do not have such option
* Added hint to final message of the deploy script how to
  install GPG key on registered system

### DIFF
--- a/server/bin/create_test_repos.py
+++ b/server/bin/create_test_repos.py
@@ -1,0 +1,469 @@
+#!/usr/bin/env python
+
+"""
+This script is used for generating testing RPM repositories
+"""
+
+import sys
+import os
+import json
+import string
+import subprocess
+import shutil
+import tempfile
+import requests
+
+
+RPM_SPEC_FILE_TEMPLATE = """
+Summary:        ${Name} package
+Name:           ${name}
+Version:        ${version}
+Release:        ${release}
+License:        GPL License
+Packager:       Jiri Hnidek <jhnidek redhat com>
+Vendor:         Red Hat
+Group:          Networking/Daemons
+URL:            http://www.tricky-testing-animals.org/
+BuildArch:      noarch
+ 
+%%description
+This is a ${name} package.
+ 
+%%files
+ 
+%%changelog
+* Fri Mar 1 2019 Jiri Hnidek <jhnidek redhat com> 0
+- ${name}
+"""
+
+GPG_NAME_REAL = "candlepin"
+
+GPG_NAME_EMAIL = "noreply@candlepinproject.org"
+
+GPG_BATCH_GEN_SCRIPT_CONTENT = """
+Key-Type: 1
+Key-Length: 2048
+Subkey-Type: 1
+Subkey-Length: 2048
+Name-Real: %s
+Name-Email: %s
+Expire-Date: 0
+%%no-protection
+%%transient-key
+%%commit
+""" % (GPG_NAME_REAL, GPG_NAME_EMAIL)
+
+GPG_EXPORTED_CANDLEPIN_KEY = os.path.expanduser("~") + "/RPM-GPG-KEY-candlepin"
+
+RPMMACROS_CONTENT = """
+%%_signature gpg
+%%_gpg_name %s
+""" % GPG_NAME_REAL
+
+RPMBUILD_ROOT_DIR = os.path.expanduser("~") + "/rpmbuild"
+
+REPO_ROOT_DIR = "/var/lib/tomcat/webapps/ROOT"
+
+CANLDEPIN_SERVER_BASE_URL = "https://localhost:8443/candlepin/"
+
+CANDLEPIN_USER = 'admin'
+CANDLEPIN_PASS = 'admin'
+
+
+def run_command(command, verbose=False):
+    """
+    Try to run command in own process
+    """
+
+    # Run command in subprocess
+    process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+
+    # Print output of command
+    if verbose is True:
+        for line in process.stdout.readlines():
+            print line
+
+    # Wait for result of process
+    ret = process.wait()
+
+    return ret
+
+
+def read_test_data(filename):
+    """
+    Try to read test data from json file.
+    """
+    try:
+        with open(filename) as fp:
+            try:
+                data = json.load(fp)
+            except ValueError as err:
+                print("Unable to read file %s with test data: %s" % (filename, str(err)))
+                return None
+    except IOError as err:
+        print("Unable to read file: %s with test data: %s" % (filename, str(err)))
+        return None
+    return data
+
+
+def get_repo_definitions(test_data):
+    """
+    Get list of repositories from test_data
+    """
+
+    if test_data is None:
+        return []
+
+    try:
+        repo_definitions = test_data['content']
+    except KeyError as err:
+        print('Info: Test data does not include any global definition of repositories')
+        repo_definitions = []
+
+    # There can be also some content specific for owners
+    try:
+        owners = test_data['owners']
+    except KeyError as err:
+        return repo_definitions
+    
+    for owner in owners:
+        try:
+            owner_content = owner['content']
+        except KeyError as err:
+            continue
+        else:
+            repo_definitions.extend(owner_content)
+    
+    return repo_definitions
+
+
+def add_packages_to_repo(packages_list, repo_path, repository):
+    """
+    This function tries to add list of packages to the repository
+    """
+    for pkg_name in packages_list:
+        print("\tAdding package %s" % pkg_name)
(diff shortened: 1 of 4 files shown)


server/bin/test_data.json
```diff
@@ -27,57 +93,59 @@
             "type": "yum",
             "vendor": "test-vendor",
             "content_url": "/foo/path",
-            "gpg_url": "/foo/path/gpg/",
```
Ceiu: What's driving the removal of these? I'm somewhat concerned what impact this will have on some of QE's test suite. Is it possible to leave these in place without breaking the RPM building bits? Perhaps using --nogpgcheck?

server/bin/create_test_repos.py
```diff
@@ -0,0 +1,469 @@
...
+
+RPM_SPEC_FILE_TEMPLATE = """
+Summary:        ${Name} package
+Name:           ${name}
+Version:        ${version}
+Release:        ${release}
+License:        GPL License
+Packager:       Jiri Hnidek <jhnidek redhat com>
+Vendor:         Red Hat
+Group:          Networking/Daemons
```
awood: You can drop the "Group" line.  It's [obsolete and deprecated](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_tags_and_sections).  Also, I don't know if you want to attach your name as packager.  I would use `whoami` or something.  I say this solely because I can envision a scenario several years from now where people start contacting you with questions about packages you "maintain."

server/bin/create_test_repos.py
```diff
@@ -0,0 +1,469 @@
...
+- ${name}
+"""
+
+GPG_NAME_REAL = "candlepin"
+
+GPG_NAME_EMAIL = "noreply@candlepinproject.org"
+
+GPG_BATCH_GEN_SCRIPT_CONTENT = """
+Key-Type: 1
+Key-Length: 2048
```
awood: Please use 4096 bit keys.  We're already seeing failures in places now that JVMs/operating systems are becoming more adamant about using longer keys.

server/bin/create_test_repos.py
```diff
@@ -0,0 +1,469 @@
+#!/usr/bin/env python
```
awood: `from __future__ import print_function, division, absolute_import`?

server/bin/create_test_repos.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+def add_packages_to_repo(packages_list, repo_path, repository):
+    """
+    This function tries to add list of packages to the repository
+    """
+    for pkg_name in packages_list:
+        print("\tAdding package %s" % pkg_name)
+        pkg_file_name = pkg_name + '-1-0.noarch.rpm'
+        src_pkg_file_path = RPMBUILD_ROOT_DIR + '/RPMS/noarch/' + pkg_file_name
```
awood: Let's use `os.path.join` just for completeness sake.  I doubt this will ever be run on Windows, but seeing a raw concatenation of directory names makes my hair stand up on end for some reason.

server/bin/create_test_repos.py
```diff
@@ -0,0 +1,469 @@
...
+        if not os.path.exists(REPO_ROOT_DIR):
+            os.makedirs(REPO_ROOT_DIR)
+    except OSError as err:
+        print("Unable to create directory: %s, %s" % (REPO_ROOT_DIR, str(err)))
+        return None
+
+    for repo in repo_definitions:
+        repo_path = REPO_ROOT_DIR + repo['content_url']
+        print("creating repository %s in %s" % (repo['name'], repo_path ))
+        if not os.path.exists(repo_path + '/RPMS'):
```
awood: `os.path.join` please

server/bin/create_test_repos.py
```diff
@@ -0,0 +1,469 @@
...
+            os.makedirs(repo_path + '/RPMS')
+
+        add_packages_to_repo(repo['packages'], repo_path, repo)
+
+        # Create repository with RPM packages
+        run_command('createrepo %s' % repo_path)
```
awood: This is just a feeling I have, but I get the sense that `createrepo_c` is the way the wind is blowing.  It's under active development and it doesn't require yum as `createrepo` does.

server/bin/test_data.json
jirihnidek: Automatic signing of generated RPMs is very complicated, because GPG requires using passhrase (see my comment in commit message). Thus I decided to remove `gpg_url` from `test_data.json`

I can create RPMs without GPG signs (it is default behavior of rpmbuild). The problem is on different place. This 'gpg_url' influences definition of repository in `/etc/yum.repos.d/redhat.repo` created automatically by e.g. subscription-manager plugin on the registered system. The `/etc/yum.repos.d/redhat.repo` is automatically generated from entitlement certificates installed in `/etc/pki/entitlement` (you can see content of cert usint `rct cc cert.pem`). The entitlement certificate is generated by candlepin server, when system attaches som subscription using e.g.: `subscription-manager attach --pool <pool_id>`

When definition repository in `redhat.repo` does include not empty string of GPG and rpm is not signed, then yum/dnf/PackageKit/microdnf will not allow to install such RPM to the system.

I will discuss this with QE team.

server/bin/create_test_repos.py
jirihnidek: OK. I will change it as you are suggesting. :-)

server/bin/create_test_repos.py
jirihnidek: OK, np.

server/bin/create_test_repos.py
jirihnidek: Good point!

server/bin/create_test_repos.py
jirihnidek: Good point.

server/bin/create_test_repos.py
jirihnidek: Interesting. The `createrepo` is just symlink on `createrepo_c` on Fedora 29. The `createrepo` is bash script from different rpm package on CentOS 7 (we use it in our Vagrant environment). Syntax should be almost the same, but `createrepo` command is probably considered deprecated. Good point. Thanks.

server/bin/create_test_repos.py
jirihnidek: Sure think.

server/bin/test_data.json
Ceiu: Would it be possible to have values here that are valid enough to not break the GPG signing, but still have values present for testing?

jirihnidek: I think that will not be possible. I will try to do more experiments with signing RPMs using GPG key.

Ceiu:

awood:

jirihnidek:
@awood @Ceiu Please, do not merge it to master. When golden ticket is used, then paths to content is different. E.g. when content access mode is granted to snowwhite, then all paths to content begins with:
"snowwhite". Thus I propose to solve this issue by creating symbolic links for all owners. Thoughts?

jirihnidek:
@awood @Ceiu It should be ready for review/merge again.

awood:

PR #2262, opened 2019-03-12
Add spec setup for python shebang in cpdb and cpsetup

zjhuntin:


server/code/setup/cpdb
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/python
```
ehelms: Does pathfix add a shebang or simply fix existing shebangs?

zjhuntin: You're right, it fixes existing shebangs, sorry about that.

kahowell:
ok to test

server/candlepin.spec.tmpl
```diff
@@ -87,6 +88,11 @@ SELinux policy module supporting candlepin
 
 %prep
 %setup -q %{SOURCE0}
+%if 0%{?rhel} == 8
+pathfix.py -pni "%{__python3} %{py3_shbang_opts}" . code/setup/cpdb code/setup/cpsetup
```
kahowell: Please remove '.' from the arguments. Otherwise, this ends up evaluating all the source files and looking for shebangs (harmless, but not necessary).

PR #2263, opened 2019-03-12
1591704: Disable status checks when org_environment is content access…

wottop:
… mode


barnabycourt:
One drive by comment, should a disabled status be gray instead of blue?

wottop:
Spec test failure is not related and has failed intermittently in the past.

PR #2265, opened 2019-03-13
Added support for unique job constraints during job queuing (ENT-844)

Ceiu:
- JobManager.queueJob will now check for matching jobs using the
  incoming job's unique constraints, and will return any existing
  active job(s) rather than queuing the new job if one is found
- Renamed PreJobExecutionException to JobInitializationException
- Removed the async job constraints table in favor of a single
  varchar field on the async job status for a hash of the
  constraints
- Added additional indexes on the async job status table to
  improve the querying performance for job lookups

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -14,8 +14,6 @@
  */
 package org.candlepin.async;
 
-import com.google.inject.Injector;
-import com.google.inject.persist.UnitOfWork;
 import org.candlepin.audit.EventSink;
 import org.candlepin.auth.Principal;
 import org.candlepin.common.filter.LoggingFilter;
@@ -26,15 +24,25 @@ import org.candlepin.model.AsyncJobStatus.JobState;
 import org.candlepin.model.AsyncJobStatusCurator;
 import org.candlepin.pinsetter.core.model.JobStatus;
 import org.candlepin.util.Util;
+
+import com.google.inject.Injector;
+import com.google.inject.persist.UnitOfWork;
+
+import org.apache.commons.codec.binary.Hex;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
-import javax.inject.Inject;
-import javax.inject.Singleton;
+import java.security.MessageDigest;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
 
 
 /**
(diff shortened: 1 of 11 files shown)


server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -139,13 +147,54 @@ public JobManager(
...
+     * Calculates a hash for the unique job constraints of a given job. If the provided map of
+     * constraints is null or empty, this method returns null.
+     *
+     * @param constraints
+     *  A map of unique job constraints for which to calculate a hash
+     *
+     * @return
+     *  a hash of the provided job constraints, or null if no constraints were provided
+     */
+    private String calculateUniqueJobConstraintHash(Map<String, String> constraints) {
```
nikosmoum: I'm not very familiar with the jobs framework (the new, or the old), so the following scenario might never happen, but here's what I want to ask:
- If you have job1 with a constraint on owner_id=my_onwer currently running, you don't want another job of this type with the same constraint running concurrently.
- And job2 is queued with two constraints: owner_id=my_owner, consumer_id=my_consumer.
- These 2 jobs are going to have different hashes from what I understand, so we'll end up running them concurrently.

Does that not violate the logic behind job1's constraint (because we end up running 2 jobs against the same owner_id, even though one of them is more specific in its restriction)? So the main question is: there a possibility of the same Job type being run with different sets/subsets/supersets of constraints?

Ceiu: That's a good question and part of what @mstead and I discussed Tuesday, and the focus of the query I brought up post-scrum yesterday. The answer, however, is not as clear since it's net-new functionality.

The intent behind converting it to a hash is to allow for arbitrary constraints per-job, while making the constraint restriction feel intuitive, regardless of the job order. This brought up a number of questions regarding the restriction, like the one you've posted above. Namely:

- Should *any* constraint match conflict?
- Should matching all of the constraints on the inbound job, but not on the existing job conflict (non-commutative matching)?
- Should order of the constraints matter?

For this implementation, we've opted with the answers, no, no and no. But as to whether or not that's correct or best... I'm not so sure. Just in responding to this post I've come up with a scenario where we may want to revert to what I had on Tuesday and following what you've said above. I.e.:

Imagine we have a job "refresh_owner_product" which updates an owner's product definition from upstream. Generally it would probably have constraints owner_id and product_id, as we probably wouldn't mind two different products being refreshed in parallel. However, two potential issues come up here:
- What if we wanted to allow the job to refresh *all* products for the org by not specifying a product. In such a case, we'd want the job to be constrained by the only constraint we passed in; which would be only the owner_id
- What if we decided that refreshing two products in parallel causes some kind of locking problem (maybe updating pools or something), but still want to be sure no one else refreshes the product at the same time (deduplication issues, perhaps). In this case, we'd want to make sure any constraint collisions blocks.

I'm less sold on the second scenario, but the first is definitely something I can see being a problem going forward. There's also a minor issue when it comes to debugging with the hashing method in that we 'lose' the inputs, so it'll be impossible for a human to know why two hashes collide without hands-on debugging.

Maybe this is something we should have a larger discussion on before continuing.

Ceiu:
Unrelated to the discussion above, I'm working on rebasing both the feature branch and this branch against master so we can get those checkstyle fixes in to unbreak Jenkins.

mstead:
Code looks good to me, however there's some good points raised in the comments above. After spending some time trying to write some notes, I think this warrants a larger discussion.

Ceiu:
Closing this PR as we're going to reapproach this with formalized use cases before pushing this through as-is.

PR #2267, opened 2019-03-14
Update deploy and cpdb to use CP-provided Liquibase[ENT-1209]

Januson:
- Add custom script for running CP-provided Liquibase
- Update deploy and cpdb to use the new script instead of locally installed Liquibase
- Remove postgresql-jdbc from candlepin.spec.tmpl

### DIFF
--- a/server/bin/liquibase.sh
+++ b/server/bin/liquibase.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Source functions library
+_prefer_jre="true"
+. /usr/share/java-utils/java-functions
+
+# Source system prefs
+if [ -f /etc/java/liquibase.conf ] ; then
+  . /etc/java/liquibase.conf
+fi
+
+# Source user prefs
+if [ -f $HOME/.liquibaserc ] ; then
+  . $HOME/.liquibaserc
+fi
+
+# Configuration
+MAIN_CLASS=liquibase.integration.commandline.Main
+BASE_FLAGS=""
+BASE_OPTIONS=""
+
+CP_CLASSPATH=${CP_LIQUIBASE_CLASSPATH:-/var/lib/tomcat/webapps/candlepin/WEB-INF/lib}
+CLASSPATH=$(JARS=("$CP_CLASSPATH"/*.jar); IFS=:; echo "${JARS[*]}")
+
+# Set parameters
+set_jvm
+set_flags ${BASE_FLAGS}
+set_options ${BASE_OPTIONS}
+
+# Let's start
+run "$@"
+
+exit 0
--- a/server/candlepin.spec.tmpl
+++ b/server/candlepin.spec.tmpl
@@ -45,10 +45,8 @@ Requires: tomcat >= 0:7.0.0
 %else
 Requires: pki-servlet-container >= 0:7.0.0
 %endif
-Requires: liquibase >= 0:3.0.0
 Requires: jss
-# Need JDBC driver for cpdb
-Requires: postgresql-jdbc
+Requires: javapackages-tools
 
 %description
 Candlepin is an open source entitlement management system.
--- a/server/code/setup/cpdb
+++ b/server/code/setup/cpdb
@@ -19,8 +19,9 @@
 from __future__ import print_function
 from optparse import OptionParser
 
-import sys
+import glob
 import os
+import sys
 
 try:
     from commands import getstatusoutput
@@ -34,6 +35,7 @@ else:
 
 JBOSS_CLASSPATH = "/var/lib/jbossas/server/production/deploy/candlepin.war/WEB-INF/classes/"
 TOMCAT_CLASSPATH = "/var/lib/" + TOMCAT + "/webapps/candlepin/WEB-INF/classes/"
+JAR_PATH = "/var/lib/" + TOMCAT + "/webapps/candlepin/WEB-INF/lib/"
 
 def run_command(command):
     (status, output) = getstatusoutput(command)
@@ -79,7 +81,7 @@ class DbSetup(object):
 
     def _run_liquibase(self, changelog_path):
         # Figure out what to append to the classpath for liquibase:
-        classpath = self.jdbc_jar
+        classpath = ":".join(glob.glob(JAR_PATH + "*postgresql*.jar"))
 
         if os.path.exists('target/classes'):
             classpath = "%s:%s" % (classpath, 'target/classes/')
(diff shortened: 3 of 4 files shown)


server/bin/liquibase.sh
```diff
@@ -0,0 +1,39 @@
...
+if [ -f /etc/java/liquibase.conf ] ; then
+  . /etc/java/liquibase.conf
+fi
+
+# Source user prefs
+if [ -f $HOME/.liquibaserc ] ; then
+  . $HOME/.liquibaserc
+fi
+
+mkdir -p /var/tmp/candlepin
```
kahowell: There are two contexts where we'll to invoke liquibase with slightly different classpaths:
 - `cpdb` when run from a Satellite installation
 - in a development setup via `server/bin/deploy`

My suggestion for handling this (feel free to take another approach), is to default to the Satellite use-case (i.e. exploded war's deps are in `/var/lib/tomcat/webapps/candlepin/WEB-INF/lib`), and then introduce an environment variable, such as `CP_LIQUIBASE_CLASSPATH`  which can override the default path, and then move some of this logic to `server/bin/deploy`, having `server/bin/deploy` do the extraction and setting `CP_LIQUIBASE_CLASSPATH` appropriately.

server/code/setup/cpdb
```diff
@@ -106,7 +108,7 @@ class DbSetup(object):
 
         print(liquibase_options)
 
-        output = run_command("liquibase %s migrate -Dcommunity=%s" % (liquibase_options, self.community))
+        output = run_command("/vagrant/server/bin/liquibase.sh %s migrate -Dcommunity=%s" % (liquibase_options, self.community))
```
kahowell: Since `cpdb` is run from a Satellite installation, the proper thing to do here is to add `liquibase.sh` to the candlepin RPM in a known location. You can reference how `cpdb` and `cpsetup` are added. (Note: there's an extra step added recently to fix the shebangs that won't be necessary for liquibase.sh).

I would suggest a location of `/usr/share/candlepin/liquibase.sh`

server/code/setup/cpdb
```diff
@@ -34,6 +35,7 @@ else:
 
 JBOSS_CLASSPATH = "/var/lib/jbossas/server/production/deploy/candlepin.war/WEB-INF/classes/"
 TOMCAT_CLASSPATH = "/var/lib/" + TOMCAT + "/webapps/candlepin/WEB-INF/classes/"
+JAR_PATH = "/var/tmp/candlepin/WEB-INF/lib/"
```
kahowell: The path on an installed candlepin (as seen via `rpm -ql candlepin` is `/var/lib/tomcat/webapps/candlepin/WEB-INF/lib`), please build the path similar to how the `TOMCAT_CLASSPATH` is built.

server/code/setup/cpdb
```diff
@@ -19,8 +19,9 @@
 from __future__ import print_function
 from optparse import OptionParser
 
-import sys
+import glob
```
kahowell: It looks like you're not using glob, actually, since the path doesn't have any globs.

server/candlepin.spec.tmpl
```diff
@@ -43,8 +43,6 @@ Requires: wget
 Requires: tomcat >= 0:7.0.0
 Requires: liquibase >= 0:3.0.0
```
kahowell: Since we can remove the liquibase dependency with this PR (yay! :smile:), please replace the `Requires: liquibase...` line with a `Requires: javapackages-tools` line (it provides `/usr/share/java-utils/java-functions` and should get pulled in transitively by tomcat, but it's better to be explicit about our dependency).

server/code/setup/cpdb
Januson: Done. I forgot to add the glob to the path.

server/candlepin.spec.tmpl
Januson: Done.

server/code/setup/cpdb
Januson: Done.

server/code/setup/cpdb
Januson: Done. I used the location you suggested.

server/bin/liquibase.sh
Januson: Done. Classpath can be overridden with `CP_LIQUIBASE_CLASSPATH`.

server/bin/liquibase.sh
```diff
@@ -0,0 +1,34 @@
...
+MAIN_CLASS=liquibase.integration.commandline.Main
+BASE_FLAGS=""
+BASE_OPTIONS=""
+
+CP_CLASSPATH=${CP_LIQUIBASE_CLASSPATH:-/var/lib/tomcat/webapps/candlepin/WEB-INF/lib}
+CLASSPATH=$(JARS=("$CP_CLASSPATH"/*.jar); IFS=:; echo "${JARS[*]}")
+
+# Set parameters
+set_jvm
+set_classpath "${CLASSPATH}"
```
kahowell: Please remove the `set_classpath` line. Apparently just setting `CLASSPATH` environment variable, as you did on line 23 is sufficient. Removing this line also fixes the

```
/usr/bin/build-classpath: Unable to find xmvn-resolve.
/usr/bin/build-classpath: Make sure that xmvn-resolve package is installed.
/usr/bin/build-classpath: error: Some specified jars were not found
```

messages.

server/candlepin.spec.tmpl
```diff
@@ -43,8 +43,7 @@ Requires: wget
...
```
kahowell: Please remove the `Requires: liquibase >= 0:3.0.0` line.

server/candlepin.spec.tmpl
```diff
@@ -127,6 +126,7 @@ sed -i 's/#LOGROTATE-3.8#//' %{buildroot}%{_sysconfdir}/logrotate.d/%{name}
 
 %{__install} -m 755 code/setup/cpsetup %{buildroot}/%{_datadir}/%{name}/cpsetup
 %{__install} -m 755 code/setup/cpdb %{buildroot}/%{_datadir}/%{name}/cpdb
+%{__install} -m 755 bin/liquibase.sh %{buildroot}/%{_datadir}/%{name}/liquibase.sh
```
kahowell: This needs a corresponding entry in the `%files` section.

Please add `%{_datadir}/%{name}/liquibase.sh` to the `%files` section.

server/bin/liquibase.sh
Januson: Done.

server/candlepin.spec.tmpl
Januson: Done.

server/candlepin.spec.tmpl
Januson: Done.

kahowell:
@Januson FYI, i had to make one tiny change: see a300106 .

wottop:
Jenkins completed

PR #2270, opened 2019-03-15
Require pki-servlet-container instead of tomcat if building for RHEL8

zjhuntin:


server/candlepin.spec.tmpl
```diff
@@ -40,7 +40,11 @@ BuildRequires: /usr/bin/pathfix.py
 
 Requires: java >= 1:1.8.0
 Requires: wget
+%if 0%{?rhel} == 8
+Requires: pki-servlet-container
```
ehelms: I wonder if this should be a bit more generic and require `bundled(tomcat)`

kahowell: I had the same question but according to https://fedoraproject.org/wiki/Bundled_Libraries?rd=Packaging:Bundled_Libraries , at least in Fedora, the virtual bundled provides are not intended for use in requires statements, and are mainly for making package maintenance easier.

kahowell:
Fixes #2266 ; we should choose one or the other.

zjhuntin:
Closing if favor of https://github.com/candlepin/candlepin/pull/2266 as I think it's more specific and will serve Candlepin better.

PR #2271, opened 2019-03-19
Added additional tests to verify import functionality

Ceiu:
- Added an additional pair of spec tests to verify
  content access mode can be exported and imported
  as the only change in a manifest
- Fixed an oversight in the import spec tests that
  was using a non-async export operation during
  async tests


Ceiu:
If this is good to merge, this needs to be forward ported to 2.4, 2.4.8, 2.5 and master.

Ceiu:

wottop:
ack

PR #2272, opened 2019-03-22
Syspurose Status filter on date fixes

nikosmoum:
- Now the syspurpose status is calculated & persisted during
consumer creation.
- Now the syspurpose status is calculated & persisted correctly
during a bind operation (consumed entitlements are filtered by
their valid date).
- Now the syspurpose status will not be recalculated & persisted
during a status call with specified on_date.
- Removed redundant db call when filtering entitlements
with a specified date for the syspurpose status calculation.
- Do not calculate syspurpose status for distributors.


Ceiu:
ack

PR #2273, opened 2019-03-25
[2.3] 1644707: RevocationOp no longer refreshes pools while locking (ENT-1221)

Ceiu:
- RevocationOp no longer uses lockAndLoad to both lock and refresh
  pools, instead only using lock and leaving pools in their current
  state
- Added additional trace output during refresh
- Removed an extraneous output line from the class
  DefaultEntitlementCertServiceAdapter

### DIFF
--- a/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
+++ b/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
@@ -802,6 +802,37 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
     }
 
     /**
+     * Locks a collection of entities with a pessimisitic write lock. Note that none of the entities
+     * will be refreshed as a result of a call to this method. If an entity needs to be locked and
+     * refreshed, used the lockAndLoad method family instead.
+     *
+     * @param entities
+     *  A collection of entities to lock
+     *
+     * @return
+     *  the provided collection of now-locked entities
+     */
+    public Collection<E> lock(Iterable<E> entities) {
+        Map<Serializable, E> entityMap = new TreeMap<>();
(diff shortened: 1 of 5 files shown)


Ceiu:
Note for reviewers: the trace output in CPM doesn't work as nicely as it would appear due to lacking functionality in the job system. The orgLogLevel bits aren't backported to this branch, so a refresh job (currently) does not set the log level properly, causing the output to be lost if Candlepin's entire log level is not set to Trace. This will be fixed automatically if the orgLogLevel change is backported.

Secondly, and probably more importantly, there should be another look backward from the change in RevocationOp through both code paths that use it to verify that the code doesn't need to reload the entities at that point. Nothing jumped out to me when I looked through it, and testing shows it's safe; however another check for sanity is probably a good thing here.

Finally, this will need to get forward ported to 2.4, 2.5 and master if this PR is cleared to be merged.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -801,6 +801,34 @@ protected int safeSQLUpdateWithCollection(String sql, Collection<?> collection,
...
+        Map<Serializable, E> entityMap = new TreeMap<>();
+
+        SessionImpl session = (SessionImpl) this.currentSession();
+        ClassMetadata metadata = session.getSessionFactory().getClassMetadata(this.entityType);
+
+        for (E entity : entities) {
+            entityMap.put(metadata.getIdentifier(entity, session), entity);
+        }
+
+        for (E entity : entityMap.values()) {
```
awood: Please add a comment about the intent here per our conversation.

PR #2277, opened 2019-03-26
[2.5] Added additional tests to verify import functionality

Ceiu:
- Added an additional pair of spec tests to verify
  content access mode can be exported and imported
  as the only change in a manifest
- Fixed an oversight in the import spec tests that
  was using a non-async export operation during
  async tests


wottop:
ack

PR #2284, opened 2019-03-28
1410601: allow hypervisor id update when it changes in the hypervisor report

wottop:


server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -296,7 +300,8 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
                     if (knownHost.getHypervisorId() != null && !hypervisorId.equalsIgnoreCase(knownHost
                         .getHypervisorId().getHypervisorId())) {
                         hypervisorIdUpdated = true;
-                        knownHost.setHypervisorId(incoming.getHypervisorId());
+                        log.debug("Changing hypervisor id to [" + hypervisorId + "]");
+                        knownHost.getHypervisorId().setHypervisorId(hypervisorId);
```
Ceiu: This is curious to me. The existing code updates the HypervisorId object and assigns it to the consumer, whereas the new code changes the existing HypervisorId object. Both, in theory, should work the same. Do you know what's going wrong in the original version?

wottop: Yes, the incoming one does not have an id or owner id. It gives a non-null error when it is stored.

Ceiu: Ahh, okay. Thanks!

PR #2286, opened 2019-04-01
1591315: Less confusing warnings, when golden ticket is used; ENT-671

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1591315
* This BZ bug report was originaly opened against sub-man and most
  of issues were really related to sub-man. These issues were
  fixed in: https://github.com/candlepin/subscription-manager/pull/2056
* It was necessary to fix only one case in candlepin server.
* When system is registererd with activation key,  auto-attch is false
  and subscriptions attached that cover installed products but all
  pools associated with the key are exhausted/consumed, then no
  exception should not be raised, when golden ticket is used.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/util/ConsumerBindUtil.java
+++ b/server/src/main/java/org/candlepin/resource/util/ConsumerBindUtil.java
@@ -82,6 +82,7 @@ public class ConsumerBindUtil {
         // Process activation keys.
 
         boolean listSuccess = false;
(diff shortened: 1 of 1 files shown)


jirihnidek:
@Ceiu Thanks for PR review. Should I create PR against other branches too?

Ceiu:
Yeah, for all the branches that this needs to be backported to.

PR #2293, opened 2019-04-05
[2.4] 1696632: Synchronize java.util.Date serializations

nikosmoum:


nikosmoum:
Closing since there is no Satellite bz to backport to the 2.4 branch for this issue yet. If one gets created & acked, will reopen this backport PR.

PR #2297, opened 2019-04-08
[2.3] 1679286: Removed AHC.lockAndLoad to address quirks with refresh (ENT-1245)

Ceiu:
- Entirely removed AbstractHibernateCurator.lockAndLoadById(s), as
  they were becoming increasingly resource intensive and difficult to
  maintain attempting to cover every possible edgecase that can occur
  with regards to entity state. No replacement will be added, due to
  the high risk involved with hidden/generic refresh behavior.
- AHC.lockAndLoad no longer attempts to bypass the cache or perform
  individual refreshes on the entities it returns
- Updated several places where lockAndLoad's behavior was being
  expected such that the callers now perform explicit refreshes where
  it is necessary to do so
- Added additional overloads for refresh and lock to address new
  usages

### DIFF
--- a/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
+++ b/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
@@ -28,7 +28,6 @@ import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.google.inject.persist.Transactional;
 
-import org.hibernate.CacheMode;
 import org.hibernate.Criteria;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
@@ -41,27 +40,20 @@ import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Projection;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
-import org.hibernate.engine.spi.EntityEntry;
-import org.hibernate.engine.spi.EntityKey;
-import org.hibernate.engine.spi.PersistenceContext;
-import org.hibernate.engine.spi.Status;
 import org.hibernate.internal.CriteriaImpl;
 import org.hibernate.internal.SessionImpl;
 import org.hibernate.metadata.ClassMetadata;
-import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.transform.ResultTransformer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xnap.commons.i18n.I18n;
 
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.SortedMap;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -715,13 +707,19 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
         return entities;
     }
 
-    public void refresh(E... entities) {
+    public void refresh(Iterable<E> entities) {
         if (entities != null) {
             EntityManager manager = this.getEntityManager();
 
             for (E entity : entities) {
                 manager.refresh(entity);
             }
+        }
+    }
+
+    public void refresh(E... entities) {
+        if (entities != null) {
+            this.refresh(Arrays.asList(entities));
         }
     }
 
@@ -832,6 +830,10 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
         return entityMap.values();
     }
 
+    public Collection<E> lock(E... entities) {
+        return entities != null ? this.lock(Arrays.asList(entities)) : null;
+    }
+
     /**
      * Locks the specified entity with a pessimistic write lock. Note that the entity will not be
      * refreshed as a result of a call to this method. If the entity needs to be locked and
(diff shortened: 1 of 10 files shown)


Ceiu:
The bug that just won't die...

Notes for reviewers:
This change scares me. A lot. In the interest in avoiding the insanity that is attempting to deal with automated cascading refreshes, I've opted to axe the lockAndLoad family of functions entirely. These have never worked as implied, and given the complexity involved in getting it right without modifying Hibernate directly [1], I feel it's both better and safer to let the caller decide if they *really* need to refresh a given entity or not.

However, while the simplicity we get from no longer dealing with this method, the worrying part is the code that was seemingly relying on it. Most of it is simple enough, but the problem comes from the idea that we're now (re)opening the doors to race conditions with simultaneous operations. The OwnerManager changes are probably safe, as is entitlement revocation in CPM, but some of the other locks on pools... yikes. The correct answer is to be locking the moment we load a given entity, or perhaps clearing the session entirely and requerying everything; but with the amount of state passing done in CPM with regards to pools... I don't believe that is possible without a substantial restructuring. This is likely a "cross our fingers and hope for the best" short-term solution, with the long-term solution being something along the lines of lock-less where we refactor a huge part of the codebase where locks are critical to the functions being performed.



[1] If we were to continue patching lockAndLoad in a way that *doesn't* explode when attempting to deal with certain sets of data, we would need to recursively checking the state of an entity *and every child entity it contains using reflection* to ensure that we don't attempt to refresh something that is net-new or has been deleted, as Hibernate doesn't just ignore those. Even then, we aren't safe from concurrent operations deleting a not-yet-locked entity from the DB, putting another thread in a desync'd state and causing it to explode when Hibernate's lookup of the deleted entity returns nothing.

server/src/test/java/org/candlepin/controller/PoolManagerTest.java
```diff
@@ -212,19 +213,21 @@ public void init() throws Exception {
...
+
+        // Make sure lock methods are passthrough
+        doAnswer(returnsFirstArg()).when(this.mockPoolCurator).lock(any(Pool.class));
+        doAnswer(returnsFirstArg()).when(this.mockProductCurator).lock(any(Product.class));
+        doAnswer(returnsFirstArg()).when(this.entitlementCurator).lock(any(Entitlement.class));
+        // doAnswer(returnsFirstArg()).when(this.certCuratorMock).lock(any(Persisted.class));
+        doAnswer(returnsFirstArg()).when(this.consumerCuratorMock).lock(any(Consumer.class));
+        doAnswer(returnsFirstArg()).when(this.mockOwnerCurator).lock(any(Owner.class));
+        // doAnswer(returnsFirstArg()).when(this.mockOwnerContentCurator).lock(any(OwnerContent.class));
+        // doAnswer(returnsFirstArg()).when(this.mockOwnerProductCurator).lock(any(OwnerProduct.class));
```
nikosmoum: Commented out code can be removed (unless there is a reason for it, which could be explained in a comment).

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -1088,86 +964,16 @@ protected E lockAndLoadById(Class<E> entityClass, Serializable id) {
...
+        for (Serializable id : ids) {
+            idSet.add(id);
         }
 
-        // Should we be returning a view of the list, rather than the fully mutable list here?
-        return result;
+        return this.currentSession()
+            .byMultipleIds(this.entityType())
+            .with(new LockOptions(LockMode.PESSIMISTIC_WRITE))
+            .multiLoad(new ArrayList(idSet));
```
nikosmoum: Also, is there a reason we don't want the `.with(CacheMode.REFRESH)` as in the `lockAndLoadById` method?

Ceiu:
This is a complete rework on the same ideas. Instructions for testing the targeted bug fix as well as review will be posted later today.

Ceiu:
To test the actual failing refresh issue:

1. Grab server/bin/org_migrator.py and server/bin/cp_connectors.py from the current master branch and save copies locally.  I put these in server/bin for consistency, but this is not strictly required.
1. Message me to get a copy of the upstream data dump and org export for the org hitting the problem
1. Deploy candlepin-2.3-HOTFIX in hosted mode on mysql: ```deploy -gmHa```
1. Dump the initial state of the database using mysqldump: ```mysqldump -u candlepin candlepin > init.sql```
1. Add the following spec test to refresh_pools_spec.rb [1], also adding a line at the top to ```require 'json'```
1. Update the ```owner_export```, ```subs_file``` and ```truncate``` values to point to the export, upstream subscriptions and initial DB state files respectively
1. Run the spec test and verify the refresh pools job fails with a HibernateException indicating any of the following errors: row not found, entity does not exist, entity is not managed. Note that the spec test itself will always succeed; pass/fail is to be determined by checking candlepin.log
1. Checkout this branch and re-run from the above steps from the deploy step forward. Refresh pools should complete successfully this time (again, verify in candlepin.log)



[1]
```
  it 'attempts to reproduce refresh errors in hosted' do
    # Truncate all tables in the db first
    truncate = system('mysql -u candlepin candlepin -e "drop database candlepin; create database candlepin;" && mysql -u candlepin candlepin < /home/crog/devel/candlepin-origin/server/repro/refresh-row_not_found/init.sql')
    expect(truncate).to eq(true)

    # inject db state from that one thing
    owner_key = '11961316'
    owner_export = '/home/crog/devel/candlepin-origin/server/repro/refresh-row_not_found/export-11961316.zip'

    import_success = system("/home/crog/devel/candlepin-origin/server/bin/org_migrator.py --dbtype mysql --import --file #{owner_export}")
    expect(import_success).to eq(true)

    # Get the owner and inject it into the list of orgs to cleanup when we're done
    owner = @cp.get_owner(owner_key)

    # Get upstream subscription dump
    subs_file = File.read("/home/crog/devel/candlepin-origin/server/repro/refresh-row_not_found/11961316_upstream.json")
    subscriptions = JSON.parse(subs_file)

    subscriptions.each do |subscription|
      # update the subscription so it gets the new org info
      subscription['owner'] = { :id => owner['id'], :key => owner['key'] }
      subscription['ownerId'] = owner['id']

      puts "Creating upstream subscription: #{subscription['id']}"
      @cp.post('hostedtest/subscriptions', {}, subscription)
    end

    puts "Refreshing org: #{owner['key']}"
    output = @cp.refresh_pools(owner['key'], true)

    pp output
  end
```

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -1087,87 +984,25 @@ protected E lockAndLoadById(Class<E> entityClass, Serializable id) {
...
-                    .enableSessionCheck(false)
-                    .multiLoad(new ArrayList(idSet)));
-            }
+        // Fetch the entities from the DB...
+        if (idSet.size() > 0) {
+            return this.currentSession()
+                .byMultipleIds(entityClass)
+                .with(new LockOptions(LockMode.PESSIMISTIC_WRITE))
+                // .with(CacheMode.REFRESH)
+                // .enableSessionCheck(false)
```
nikosmoum: Remove commented lines

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -1003,64 +929,35 @@ public E lockAndLoadById(Serializable id) {
...
-            }
-        }
-
-        return entity;
+    protected E lockAndLoad(Class<E> entityClass, Serializable id) {
+        return this.currentSession()
+            .byId(entityClass)
+            .with(new LockOptions(LockMode.PESSIMISTIC_WRITE))
+            // .with(CacheMode.REFRESH)
+            // .enableSessionCheck(false)
```
nikosmoum: Remove commented lines

nikosmoum:
I run the test with/without your changes with expected results (the error is gone with your changes), but with your further suggestion of adding a sleep statement at the end, I was able to reproduce the error again.

nikosmoum:
I run the test with the sleep statement and the error is no longer there. I will merge since the spec failures are the usual suspects. I think it would be a good idea to create a splunk alert that looks for any of the potential errors we know of found in stage/prod with a daily notification or so, in order to immediately know of them once this change gets deployed (because it looks like there will be fixes needed like this latest one after this change).

PR #2303, opened 2019-04-11
[2.3] Improved hosted test adapter separation

Ceiu:
- The HostedTestSubscriptionResource and ServiceAdapter are now
  completely isolated from the locally managed products and
  content
- Updated the hosted test API to allow for directly updating
  products, content and subscriptions independently from their
  downstream counterparts
- Updated several spec tests to utilize the separation between
  up and downstream product/content definitions

### DIFF
--- a/server/client/ruby/hostedtest_api.rb
+++ b/server/client/ruby/hostedtest_api.rb
@@ -3,10 +3,14 @@ module HostedTest
   @@hosted_mode = nil
   @@hostedtest_alive = nil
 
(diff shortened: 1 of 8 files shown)


Ceiu:
This is a backport of the hosted test refactor components of the larger adapter refactor that was merged into future branches from PR #2067.

This is needed as a prerequisite to further attempts to fix the row-not-found issue, as we need the upstream and downstream separation to properly reproduce the issue in a dev environment.

PR #2309, opened 2019-04-17
Port ExportJob To The New Messaging Framework[ENT-849]

Januson:
- Migrate ExportJob and its test to the new Artemis framework

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -14,8 +14,13 @@
  */
 package org.candlepin.async;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.inject.Injector;
+import com.google.inject.persist.UnitOfWork;
 import org.candlepin.audit.EventSink;
 import org.candlepin.auth.Principal;
+import org.candlepin.auth.PrincipalData;
+import org.candlepin.auth.UserPrincipal;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -563,4 +579,17 @@ private AsyncJobStatus processJobFailure(AsyncJobStatus status, EventSink eventS
...
 
+
+    private void setupPrincipal(final JobDataMap jobData) {
+        // TODO remove this guard check once we have better way to share the principal
+        if (!jobData.containsKey("principal")) {
+            log.warn("Principal data are missing from the job data!");
+            return;
+        }
+        final String principalJson = jobData.getAsString(AsyncJobStatus.PRINCIPAL_KEY);
+        final PrincipalData principal = (PrincipalData) Util.fromJson(principalJson, PrincipalData.class);
```
Ceiu: We should be able to just dump the principal object as-is into this, and then pull it back out later. I believe the Hibernate serialization bits should handle all of it for us.

Ceiu: After investigating further, the current serialization is backed by Jackson which, indeed, returns LinkedHashMaps for any generic object, or any object for which it isn't provided an explicit class.

This should be fixed in the future, but for the interim, this workaround is acceptable.

PR #2311, opened 2019-04-24
1698528: Syspurpose support on activation keys

wottop:


server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyDTO.java
```diff
@@ -300,6 +303,68 @@ public ActivationKeyDTO setServiceLevel(String serviceLevel) {
...
+        return this;
+    }
+
+    /**
+     * Retrieves the add ons of this ActivationKeyDTO object.
+     *
+     * @return the add ons of this ActivationKeyDTO object.
+     */
+    public Set<String> getAddOns() {
+        return addOns;
```
Ceiu: Have this return a SetView of the addons collection

server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyDTO.java
```diff
@@ -300,6 +303,68 @@ public ActivationKeyDTO setServiceLevel(String serviceLevel) {
...
+
+    /**
+     * Sets the add ons of this ActivationKeyDTO object.
+     *
+     * @param addOns the add ons of this ActivationKeyDTO object.
+     *
+     * @return a reference to this DTO object.
+     */
+    public ActivationKeyDTO setAddOns(Set<String> addOns) {
+        this.addOns = addOns;
```
Ceiu: We should decouple from the inbound collection here. A simple ```this.addOns = addOns != null ? new HashSet<>(addOns) : null;``` will work.

PR #2312, opened 2019-04-26
1699386: Set support level of products, when service level is changed; ENT-1270

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1699386


barnabycourt:
The support level of a pool is intrinsic to the product that the pool is created from. The SLA attributes of a consumer have no effect on the entitlement certificates.

PR #2314, opened 2019-04-30
1709723: last check-in date is not updated for hypervisors reported by virt-who (candlepin changes) [ENT-1049]

Januson:
- Create a new endpoint POST /hypervisors/heartbeat/{owner}
- The new endpoint updates last checkin date of all consumers for the given reporter_id

### DIFF
--- a/server/spec/hypervisor_heartbeat_spec.rb
+++ b/server/spec/hypervisor_heartbeat_spec.rb
@@ -1,0 +1,40 @@
+require 'spec_helper'
+require 'candlepin_scenarios'
+require 'thread'
+
+describe 'Hypervisor Resource - Heartbeat Endpoint', :type => :virt do
+  include CandlepinMethods
+  include VirtHelper
+
+  before(:each) do
+    skip("candlepin running in standalone mode") if is_hosted?
+    @expected_host_hyp_id = random_string("host")
+    @expected_reporter_id = random_string("reporter")
+    @expected_host_name = random_string("name")
+    @expected_guest_ids = [@uuid1, @uuid2]
+
+    @owner = create_owner random_string('virt_owner')
+    @user = user_client(@owner, random_string('virt_user'))
+
+    # we must register the consumer to use it as a client
+    # hypervisor check in creation does not result in a client cert
+    @consumer = @user.register(@expected_host_name, :hypervisor, nil, {"test_fact" => "fact_value"},
+                              nil, @owner['key'], [], [], nil, [], @expected_host_hyp_id,
+                               [], nil, '1518341313', nil, nil, nil, 0, nil, nil, nil, nil, [], @expected_reporter_id)
+
+    @client = consumer_client(@user, random_string("consumer"))
+  end
+
+  it 'should update last checkin date of consumers of given reporter' do
+    job = @client.hypervisor_heartbeat_update(@owner['key'], @expected_reporter_id)
+
+    job.should_not be_nil
+    job['id'].should include("hypervisor_update")
+    wait_for_job(job['id'], 10)
+
+    consumer = @cp.get_consumer(@consumer.uuid)
+
+    Date.parse(consumer.lastCheckin).should == Date.today
+  end
+
+end
--- a/server/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/server/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -560,6 +560,40 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
             .executeUpdate();
     }
 
+    @Transactional
+    public void heartbeatUpdate(final String reporterId, final Date checkIn, final String ownerKey) {
+        final String query;
+        final String db = config.getProperty("jpa.config.hibernate.connection.driver_class");
+        if (db.contains("mysql")) {
+            query = "" +
+                "UPDATE cp_consumer a" +
+                " JOIN cp_consumer_hypervisor b on a.id = b.consumer_id " +
+                " JOIN cp_owner c on c.id = a.owner_id" +
+                " SET a.lastcheckin = :checkin" +
+                " WHERE b.reporter_id = :reporter" +
+                " AND c.account = :ownerKey";
+        }
+        else if (db.contains("postgresql")) {
+            query = "" +
+                "UPDATE cp_consumer" +
+                " SET lastcheckin = :checkin" +
+                " FROM cp_consumer a, cp_consumer_hypervisor b, cp_owner c" +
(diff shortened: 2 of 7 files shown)


server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -560,6 +560,21 @@ public void updateLastCheckin(Consumer consumer, Date checkinDate) {
...
     }
 
+    @Transactional
+    public void heartbeatUpdate(final String reporterId, final Date checkIn) {
+        final String hql = "" +
+            "UPDATE cp_consumer" +
+            " SET lastcheckin = :checkin" +
+            " WHERE id IN (SELECT consumer_id" +
+            " FROM cp_consumer_hypervisor" +
+            " WHERE reporter_id = :reporter)";
```
barnabycourt: This needs to be scoped to the current organization.

wottop: Don't need a subselect:

            "UPDATE cp_consumer a, cp_consumer_hypervisor b, cp_owner c" +
            " SET a.lastcheckin = :checkin" +
            " WHERE a.id = b.consumer_id" +
            " AND b.reporter_id = :reporter" +
            " AND a.owner_id = c.id" +
            " AND c.account = :orgKey;"

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorHeartbeatUpdateJob.java
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @param reporterId reporter for which to update lastCheckin dates
+     * @return a {@link JobDetail} that describes the job run
+     */
+    public static JobDetail from(final String reporterId) {
+        if (reporterId == null || reporterId.isEmpty()) {
+            throw new IllegalArgumentException("ReporterId is required!");
+        }
+        final JobDataMap map = new JobDataMap();
+        map.put(JobStatus.TARGET_TYPE, JobStatus.TargetType.OWNER);
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -297,6 +298,21 @@ public JobDetail hypervisorUpdateAsync(
...
+    @POST
+    @Produces(MediaType.APPLICATION_JSON)
+    @Transactional
+    @Path("/heartbeat/{owner}")
+    public JobDetail hypervisorHeartbeatUpdate(
+        @PathParam("owner")
+        @Verify(value = Owner.class, require = Access.READ_ONLY, subResource = SubResource.HYPERVISOR)
+        final String ownerKey,
+        @QueryParam("reporter_id") final String reporterId) {
+        return HypervisorHeartbeatUpdateJob.from(reporterId);
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorHeartbeatUpdateJob.java
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Creates a {@link JobDetail} that runs this job for the given reporter_id.
+     *
+     * @param reporterId reporter for which to update lastCheckin dates
+     * @return a {@link JobDetail} that describes the job run
+     */
+    public static JobDetail from(final String reporterId) {
+        if (reporterId == null || reporterId.isEmpty()) {
+            throw new IllegalArgumentException("ReporterId is required!");
+        }
```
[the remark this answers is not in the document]
Januson: Done.

server/spec/hypervisor_heartbeat_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+                              nil, nil, [], [], nil, [], @expected_host_hyp_id, @expected_reporter_id, last_checkin_date = '1518341313')
+
+    @client = consumer_client(@user, random_string("consumer"))
+  end
+
+  it 'should update last checkin date of consumers of given reporter' do
+    @client.hypervisor_heartbeat_update(@owner['key'], @expected_reporter_id)
+
+    # Wait for job to finish
+    sleep 1
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
Januson: Done.

server/src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -297,6 +298,29 @@ public JobDetail hypervisorUpdateAsync(
...
+    @ApiOperation(notes = "Updates last check in date of all consumers of the given reporterId.",
+        value = "hypervisorHeartbeatUpdate")
+    @ApiResponses({
+        @ApiResponse(code = 202, message = ""),
+        @ApiResponse(code = 400, message = "Illegal reporter ID was provided"),
+        @ApiResponse(code = 404, message = "Target owner not found.")})
+    @POST
+    @Produces(MediaType.APPLICATION_JSON)
+    @Transactional
+    @Path("/heartbeat/{owner}")
```
nikosmoum: Something I did not think about before is the design of the API itself.

- For the verb: I would say that this should be either a PATCH or a PUT instead of a POST (technically, it fits much more to use a PATCH, since we don't accept an entire entity, like a `ConsumerDTO` or a `HypervisorDTO` to be updated, but we only want to update a specific field of the entity, the last check in date).

- For the url: I think the `{owner}` should be at the start, since it is the top-level entity, which is consistent with all other endpoints we have. Also, maybe the `reporter_id` should also be in the URL, since this is what identifies the entity/ies being PATCHed. So the url should be one of: `/{owner}/heartbeat`,  `/{owner}/{reporter_id}/heartbeat` or `/{owner}/heartbeat/{reporter_id}`.

@wottop @Ceiu @Januson What are your thoughts?

wottop: My opinion is that it stays on the HypervisorResource. It mirrors the existing async endpoint for the hypervisor check in in the url and use of reporter id as a parameter.

barnabycourt: /{owner}/heartbeat assumes that there is only ever one type of thing that we are reporting a heartbeat for. Does it need to be scoped by the type of heartbeat?

Januson: Done.

PR #2318, opened 2019-05-06
[F] Added job constraints and the UniqueByArg constraint (ENT-844)

Ceiu:
- Added the UniqueByArg job queuing constraint and updated the
  JobManager and JobBuilder to specify and respect these
  constraints
- Added additional queries to the AsyncJobStatusCurator

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobConstraint.java
+++ b/server/src/main/java/org/candlepin/async/JobConstraint.java
@@ -1,0 +1,45 @@
+/**
+ * Copyright (c) 2009 - 2019 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.async;
+
+import org.candlepin.model.AsyncJobStatus;
+
+
+
+/**
+ * The JobConstraint interface provides a standard API for all job queuing constraints.
+ */
+@FunctionalInterface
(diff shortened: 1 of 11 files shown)


server/src/main/java/org/candlepin/model/AsyncJobStatusCurator.java
```diff
@@ -28,6 +37,54 @@ public AsyncJobStatusCurator() {
...
+        if (states != null && !states.isEmpty()) {
+            String jpql = "SELECT aj FROM AsyncJobStatus aj WHERE aj.state IN (:states)";
+
+            return this.getEntityManager()
+                .createQuery(jpql, AsyncJobStatus.class)
+                .setParameter("states", states)
+                .getResultList();
+        }
+
+        return new LinkedList<>();
```
nikosmoum: It's not really that important, but why is this method returning a `LinkedList`, while the overloaded of this an `ArrayList` in the case no jobs were found? It is probably better for both to just return `Collections.emptyList()` since that is more consistent and cost effective (not creating a new object).

server/src/main/java/org/candlepin/async/JobConstraint.java
```diff
@@ -0,0 +1,44 @@
...
+package org.candlepin.async;
+
+import org.candlepin.model.AsyncJobStatus;
+
+
+
+/**
+ * The JobConstraint interface provides a standard API for all job queuing constraints.
+ */
+public interface JobConstraint {
```
nikosmoum: It is good practice to add `@FunctionalInterface` to such interfaces.

Ceiu: Fixed.

server/src/main/java/org/candlepin/model/AsyncJobStatusCurator.java
Ceiu: I've made both return an ArrayList.

The reason I'm not using Collections.emptyList is due to the inconsistency we'd get with whether or not the list is mutable. If the list is returned from getResultList, then the list is mutable; otherwise it wouldn't be. I suppose we could wrap the query part in a Collections.unmodifiableList to make this method consistent, but then it will be the only curator that works this way. What're your thoughts here?

nikosmoum: No problem with not using emptyList since that would be inconsistent with other curators; the main issue was the ArrayList/LinkedList mismatch.

PR #2323, opened 2019-05-07
[F] Added the UniqueByArgument constraint to the async ExportJob (ENT-1265)

Ceiu:
- Added the UniqueByArgument constraint to ExportJob, preventing it
  from being queued if another is already active for the same
  consumer
- Changed a few job arguments to metadata or explicit job
  configuration as appropriate

### DIFF
--- a/server/src/main/java/org/candlepin/async/tasks/ExportJob.java
+++ b/server/src/main/java/org/candlepin/async/tasks/ExportJob.java
@@ -15,24 +15,25 @@
 package org.candlepin.async.tasks;
 
 import com.google.inject.Inject;
-import org.apache.log4j.MDC;
 import org.candlepin.async.AsyncJob;
 import org.candlepin.async.JobBuilder;
 import org.candlepin.async.JobDataMap;
 import org.candlepin.async.JobExecutionContext;
 import org.candlepin.async.JobExecutionException;
+import org.candlepin.async.JobConstraints;
 import org.candlepin.async.JobManager;
-import org.candlepin.common.filter.LoggingFilter;
 import org.candlepin.controller.ManifestManager;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.Owner;
-import org.candlepin.pinsetter.core.model.JobStatus;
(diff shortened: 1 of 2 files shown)


Ceiu:
Note that this PR doesn't completely finish the ExportJob task. We still need to wire it up, clean up the way the job is scheduled, and possibly fix the way the job key is registered. These will come after the AsyncJobStatus DTO is added.

server/src/main/java/org/candlepin/async/tasks/ExportJob.java
```diff
@@ -77,7 +80,7 @@ public ExportJob(ManifestManager manifestManager) {
     @Override
     public Object execute(final JobExecutionContext jdata) throws JobExecutionException {
         final JobDataMap map = jdata.getJobData();
-        final String consumerUuid = map.getAsString(JobStatus.TARGET_ID);
+        final String consumerUuid = map.getAsString(CONSUMER_KEY);
```
mstead: :+1:  Way more explicit.

server/src/main/java/org/candlepin/async/tasks/ExportJob.java
```diff
@@ -113,16 +116,14 @@ public static JobBuilder scheduleExport(
         final Map<String, String> extensionData
     ) {
         return JobBuilder.forJob(JOB_KEY)
```
mstead: **A NOTE, BUT NOT FOR THIS PR:**
Not overly worried if this gets changed, but, the method scheduleExport is a little misleading as it isn't really scheduling anything; it is just creating a builder. A better name... not sure. :)

mstead:
Looks good, waiting on tests.

server/src/main/java/org/candlepin/async/tasks/ExportJob.java
Ceiu: This is definitely on my radar for the second part of this. I was never terribly thrilled with scheduling a job by returning an object from a resource and letting a filter pick it up later.

Need to come up with something that's easy to use and a little more clear in the intent.

PR #2324, opened 2019-05-08
[F] Added the AsyncJobStatusDTO and associated translator (ENT-1321)

Ceiu:
- Added the AsyncJobStatusDTO and AsyncJobStatusTranslator
- Updated StandardTranslator accordingly

### DIFF
--- a/server/src/main/java/org/candlepin/dto/api/v1/AsyncJobStatusTranslator.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/AsyncJobStatusTranslator.java
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) 2009 - 2019 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.dto.api.v1;
+
+import org.candlepin.dto.ModelTranslator;
+import org.candlepin.dto.TimestampedEntityTranslator;
+import org.candlepin.model.AsyncJobStatus;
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/candlepin/dto/api/v1/AsyncJobStatusTranslator.java
```diff
@@ -0,0 +1,82 @@
...
+        destination.setEndTime(source.getEndTime());
+        destination.setResult(source.getJobResult());
+
+        JobState state = source.getState();
+        destination.setState(state != null ? state.name() : null);
+
+        JobState pstate = source.getPreviousState();
+        destination.setPreviousState(pstate != null ? pstate.name() : null);
+
+        return destination;
```
mstead: I noticed that the following AsyncJobStatus fields are not included in the DTO. I'm assuming that they aren't included for a reason, but just wanted to make sure that you didn't miss any. Some of them may be useful to someone viewing the details of a job... I've marked them as such. I'm not implying that they SHOULD be included, but mostly curious.

* jobKey - is this something that was missed?
* logExecutionDetails - not sure that this is important.
* metadata -- woudn't this contain info that someone may want to see via the API?
* maxAttempts -- was this missed?
* jobData -- Doesn't this contain the args for the job? Should this be included?

Ceiu: I used the Pinsetter job status instance as a loose guide for what to include. From that list:

- JobKey was explicitly excluded
- JobData did not exist, as it was stored in the Quartz job detail, rather than the job status itself
- logExecutionDetails, maxAttempts and metadata are net new, and the former two are simply configuration options for each job. I can see max attempts (and current attempts) being desirable to see on output for status, but execution detail logging and metadata didn't seem that useful to me.

What're your thoughts here?

mstead: Omit the jobKey.

As for the others, I'd say if they contain useful info that an API user would like to know, then add them in. Otherwise, you'll just be forcing folks to look up the details from the DB itself.

barnabycourt: JobStatus gets queried a lot, are these fields things we want to know every time or only some of the time? If they are not being accessed today, it might be best to wait until we have a requirement before adding more to the API DTO.

Ceiu: That's the route we're taking here. I've added the attempts and max attempts, but everything else will be left out until we have a need for it.

PR #2325, opened 2019-05-08
Use JUnit 5

awood:
This commit converts our tests to use JUnit 5.  It's not just running JUnit 4 tests under JUnit 5.  I actually found that didn't work all that well.

This PR
* Removes our use of `@Rule` which is a JUnit 4 thing.  We used some standard rules, but also had a custom one that set up the in-memory DB before each test.  There was no way to get JUnit 5 working with that rule that I could discern.
* Upgrades the version of Mockito we are using (which caused some headaches because in Mockito 2, `any()` doesn't match `null` anymore) so that we can use its JUnit 5 integration (which didn't exist in the old version of Mockito we were using).
* Uses the JUnit 5 native parameterized test functionality
* Converts a lot of the tests to pure JUnit 5 while still leaving a fair amount running under JUnit 4 that we can migrate over time.


awood:

Januson:
I don't know what is the difference from running these tests on Jenkins but running them from Vagrant results in quite a lot of failures. Mostly UnnecessaryStubbingException.

common/src/test/java/org/candlepin/common/guice/TestingScope.java
```diff
@@ -0,0 +1,93 @@
...
+
+        @Override
+        public <T> Provider<T> scope(final Key<T> key, final Provider<T> unscoped) {
+            return new Provider<T>() {
+                @SuppressWarnings("unchecked")
+                public T get() {
+
+                    Object o = backingMap.get(key);
+
+                    if (o == null) {
```
Januson: Just a detail. You could simplify this with computeIfAbsent.

server/src/test/resources/logback-test.xml
```diff
@@ -7,14 +7,14 @@
     </encoder>
   </appender>
   <logger name="org.candlepin.util.filters.DumpFilter" level="INFO"/>
-  <logger name="org.candlepin" level="DEBUG"/>
+  <logger name="org.candlepin" level="INFO"/>
```
Ceiu: We want debug output enabled by default during testing, right?

awood: It's awfully noisy.  I'm not opposed but maybe run it with DEBUG on your end and see what you think :)

Ceiu: You're not wrong about it being noisy, but do we have a way of running the tests in debug mode without having to manually update this file? IIRC, it used to be INFO as it is here, but then it became DEBUG due to someone (probably me) consciously changing it or forgetting to change it back and it slipped through a review. Also, with the sheer amount of spam from just the raw test results alone, is the extra/reduced noise even significant?

I suppose with debugging being easier with with the provided IntelliJ licenses this is less important as it was years ago, I just would like this to be easier to flip on demand.

Ceiu: This doesn't seem to make a difference with gradle. Disregard.

PR #2326, opened 2019-05-09
[2.4] 1704720: Synchronize java.util.Date serializations; ENT-1294

nikosmoum:


nikosmoum:
This is a backport of https://github.com/candlepin/candlepin/pull/2239

Ceiu:
ack

PR #2327, opened 2019-05-09
1694809: Reduce in operator block size

wottop:
Based on test of hypervisor check in with with 17K hosts and 6 guests each.
 Was able to execute with a block setting of 30K. Using half that for caution.


Ceiu:
What is the query that was triggering issues?

I believe this will work fine, but I want to be sure we're aware of the underlying issue and whether or not this will actually prevent the same type of problem occurring later. I've guessing I've hit a similar issue in other queries, so before we merge, I want to check if it's something we can address in a different way.

Ceiu:
Looking at the query raising this issue (getGuestIdMap) and the previous limits, the problem here was that we were running into the parameter limit in cases where our guest ID count was larger than 32767, as there was an additional parameter on top of the IN block (owner ID).

While this was the error I was expecting, fortunately the query in question doesn't use multiple IN blocks (or other such large parameter blocks), so dropping the IN block size to 15k will fix this.

PR #2328, opened 2019-05-13
1698614: Stop event persistence & deprecate event APIs

nikosmoum:
* Stop persisting events in the database.
* Change event & atom feed API endpoints to return empty lists
  and mark them as deprecated.
* Remove EventDTO/EventTranslator.
* Remove hibernate annotations from Event.
* Change tests accordingly.


server/src/main/java/org/candlepin/resource/EventResource.java
```diff
@@ -48,57 +43,57 @@
...
-            events.add(toReturn);
-            eventAdapter.addMessageText(events);
-
-            return this.translator.translate(toReturn, EventDTO.class);
-        }
-
-        throw new NotFoundException(i18n.tr("Event with ID \"{0}\" could not be found.", uuid));
+    public Event getEvent(@PathParam("uuid") String uuid) {
+        throw new BadRequestException(i18n.tr("This API endpoint is deprecated, and will be" +
+            " removed on the next major release.", uuid));
```
nikosmoum: Special attention to this endpoint change for reviewers: Is it an acceptable solution to throw a 400 here, where the user is fetching a single event (all other event/atom endpoints were changed to return empty lists), or is there a better approach (e.g. should we simply return a non-populated `new Event()`)?

Ceiu: To recap the discussion from IRC: use the NotFoundException with the new message and we should be covered here.

Ceiu:
This is good with the exception change

server/src/main/java/org/candlepin/resource/EventResource.java
nikosmoum: Fixed

barnabycourt:
Should we be dropping the cp_event table from the database as well?

nikosmoum:
@barnabycourt I believe during grooming of this task, we decided to split it in two, to better deal with hosted. From the jira task: The table removal needs to happen after a version of candlepin that doesn't depend on it is deployed.

Ceiu:
I've added an additional task to the backlog to complete the removal of the cp_event table and related curators and any other event storage bits that may still be lying around.

With that in mind, are there any other requests or additional work to be completed in this PR?

PR #2332, opened 2019-05-15
[F] Updated ConsumerResource to use the new async ExportJob task (ENT-1265)

Ceiu:
- ConsumerResource now uses the async/Artemis ExportJob task
  instead of the old Pinsetter version
- Added the AsyncJobStatusPermission to handle support for
  checking permissions on async jobs
- Updated the temp AsyncJobResource to use the DTOs and
  new permissions
- Added (temporary?) methods to the spec tests for fetching
  async jobs by ID
- Async job execution now uses the new JobPrincipal rather
  than a system principal or empty user principal
- Renamed JobBuilder to JobConfig to better match its actual usage
- JobConfig now has a .validate method which will check its
  settings and verify the minimum required fields are set
- ExportJob has been updated to use the new JobConfig and validator
- ExportJob.createJobConfig now returns a JobConfig to be filled
  in rather than taking six parameters and doing all the work
  directly
- Updated ExportTest accordingly

### DIFF
--- a/server/src/main/java/org/candlepin/async/tasks/ExportJob.java
+++ b/server/src/main/java/org/candlepin/async/tasks/ExportJob.java
@@ -16,7 +16,8 @@ package org.candlepin.async.tasks;
 
 import com.google.inject.Inject;
 import org.candlepin.async.AsyncJob;
-import org.candlepin.async.JobBuilder;
+import org.candlepin.async.JobConfig;
+import org.candlepin.async.JobConfigValidationException;
 import org.candlepin.async.JobDataMap;
 import org.candlepin.async.JobExecutionContext;
 import org.candlepin.async.JobExecutionException;
@@ -26,7 +27,6 @@ import org.candlepin.controller.ManifestManager;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.Owner;
 import org.candlepin.sync.ExportResult;
-import org.candlepin.util.Util;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -43,6 +43,7 @@ import java.util.Map;
  * @see ExportResult
  */
 public class ExportJob implements AsyncJob {
+    private static Logger log = LoggerFactory.getLogger(ExportJob.class);
 
     protected static final String OWNER_KEY = "org";
     protected static final String CONSUMER_KEY = "consumer_uuid";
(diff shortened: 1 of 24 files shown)


Ceiu:
Notes for reviewers:
- Part of this task is completed in ENT-1289, but is non-critical to this PR (adjusting how jobs are registered/job key stuff in ConfigProperties)
- I'm not entirely sold on the static factory methods for setting up the JobBuilder, but there's aside from a dedicated job factory or some such, there's no a lot more that can be done instead without requiring users of this particular job from needing to know what parameters the job needs to run. Perhaps this is something which should be addressed by better/more liberal use of the ArgumentConversionException or other such exceptions during job execution to allow it to fail more gracefully when misconfigured. Suggestions and ideas are welcome here, but I suspect we'll end up maintaining the pattern we're using today, for better or worse.

Ceiu:
Still seems like a few tests need to be updated to use the new/temp jobs endpoint.

server/src/main/java/org/candlepin/model/AsyncJobStatus.java
```diff
@@ -126,8 +126,8 @@ public boolean isTerminal() {
 
     @ElementCollection(fetch = FetchType.EAGER)
     @CollectionTable(name = "cp_async_job_metadata", joinColumns = @JoinColumn(name = "job_id"))
-    @MapKeyColumn(name = "key")
-    @Column(name = "value")
+    @MapKeyColumn(name = "mkey")
```
kahowell: Why this change (and mval)?

Ceiu: At least one appears to be a keyword in mysql, and Hibernate doesn't feel like quoting column names in the queries it generates. I'm pretty sure "value" might be fine, but I changed both just to be safe.

It probably won't end up mattering much as metadata is set to be serialized and the table removed in the other PR.

Ceiu: I've reverted this and, instead, added quotes to the column names in code (see AsyncJobStatus). Not sure why Hibernate doesn't do it by default if it has a standard syntax for specifying a column name should be quoted, but whatever.

server/src/main/java/org/candlepin/async/tasks/ExportJob.java
```diff
@@ -99,24 +98,44 @@ public Object execute(final JobExecutionContext jdata) throws JobExecutionExcept
...
+            throw new IllegalArgumentException("consumer is null");
+        }
+
+        if (owner == null) {
+            throw new IllegalArgumentException("owner is null");
+        }
+
         return JobBuilder.forJob(JOB_KEY)
-            .setJobName("export-" + Util.generateUUID())
+            .setJobName("export_manifest")
```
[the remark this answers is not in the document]
Ceiu: We already have two fields for that: correlated service ID and the job ID.

This old method was basically a holdover/hack from how we used to differentiate. Our log format for the line you're thinking of looks like this:
```<pattern>%d{ISO8601} [thread=%thread] [%X{requestType}=%X{requestUuid}, org=%X{org}, csid=%X{csid}] %-5p %c - %m%n</pattern>```

In Pinsetter, this was populated as follows:
```
MDC.put("requestType", "job");
MDC.put("requestUuid", context.getJobDetail().getKey().getName());
```
Note that we're setting the job name as the request UUID.

In JobManager, we're doing this instead:
```
MDC.put("requestType", "job");
MDC.put("requestUuid", status.getId());
```

If we were to keep the old-style log entries, I'd rather build the string on the fly using the name and ID, rather than forcing that to be done by the users creating the jobs. But, adjustment period aside, I think we'll be okay on this front.

**Comment by Ceiu** [comment]
@nikosmoum @kahowell 
I've played with the idea of the builder pattern a bit and this is where I've landed:

- I don't want the job builders being responsible for queuing the job -- that's something to be done explicitly by the callers
- I'm okay with moving away from methods with tons of parameters in favor of a fluent style reflecting that of the job builder's API natively
- If done properly, we can also hit another need for the job system by adding builder validation that can be performed in multiple places without multiple implementations

That said, this is what I came up with:
- Add a ```.validate()``` method to JobBuilder. The default validate method will perform the bare minimum checks for a job, such as ensuring the key has been set and... that's probably it.
- Jobs with complicated creation methods can extend JobBuilder to add concrete setters (syntactic sugar), and extend the ```.validate``` method with more explicit checks for that particular job. These can be returned from the job class itself like it is today (```ExportJob.getJobBuilder()```) or created manually (```new ExportJobBuilder()```). I'm leaning toward the former as it allows the builder to be fully enclosed in its containing job class instead of doubling up the files and distribution of code.
- Jobs with simple creation methods can opt to do nothing (as is the case in several Pinsetter jobs) or return a preconfigured, vanilla JobBuilder.

Using ExportJob as an example, scheduling it would look like this:

```
JobBuilder builder = ExportJob.getJobBuilder()
    .setConsumer(consumer)
    .setOwner(owner)
    .setCdnLabel(...)
    .setWebAppPrefix(...)
    .setApiUrl(...)
    .setExtensionData(...);

AsyncJobStatus job;
try {
    job = this.manager.queueJob(builder);
}
catch (JobValidationException e) {
    // If we need to translate the message into something friendlier for end-users, we can do so here.
}

return this.translator.translate(job, AsyncJobStatusDTO.class);
```

Thoughts?

**Comment by mstead** [comment]
> Using ExportJob as an example, scheduling it would look like this:
> 
> ```
> JobBuilder builder = ExportJob.getJobBuilder()
>     .setConsumer(consumer)
>     .setOwner(owner)
>     .setCdnLabel(...)
>     .setWebAppPrefix(...)
>     .setApiUrl(...)
>     .setExtensionData(...);
> 
> try {
>     AsyncJobStatus job = this.manager.queueJob(builder);
>     return this.translator.translate(job, AsyncJobStatusDTO.class);
> }
> catch (JobValidationException e) {
>     // If we need to translate the message into something friendlier
>     // for end-users, we can do so here.
>     throw e;
> }
> ```
> 
> Thoughts?

Architecturally, this seems fine. My only minor beef with this approach is that while building a job, it tells me nothing about what is required by the job; unless I first look at the validate method on the builder, or I run the code to find out that validation fails. That being said, I'm not a fan of methods that take a huge number of args and this seems like a way to mitigate that.

**Comment by kahowell** [comment]
> Architecturally, this seems fine. My only minor beef with this approach is that while building a job, it tells me nothing about what is required by the job; unless I first look at the validate method on the builder, or I run the code to find out that validation fails. That being said, I'm not a fan of methods that take a huge number of args and this seems like a way to mitigate that.

A possible alternative would be to create a parameter POJO that uses `@NotNull` to denote which properties are required (and then validation becomes implemented simply via `javax.validation.Validator#validate`).

However, I quite like the syntactical style of using the builder pattern. The only improvement I'd suggest is in order to make it more idiomatic, a `build()` method handling validation and returning a type of simply `Job`, and then modification of `JobManager`'s queue method to work with this type rather than `JobBuilder`. As written currently, `JobBuilder` feels like a misnomer; it's closer to `Job` with chain-able setters or fluent interface. Alternatively, JobBuilder could just be renamed to Job... Thoughts?

**Comment by Ceiu** [comment]
What would the Job object represent? What would it contain? Maybe the name of JobBuilder is a bit of a misnomer (JobConfigurator instead?), but I don't quite understand the idea behind adding a third object in the middle.

For validation, the constraints package will cover many/most of the cases, but I fear not all. We could start making our own annotations, but that's yet *more* classes for the soup. Also annotations. I hate annotations.

I don't have an answer for Mike's concern, and I'm not sure how to go about solving that one without falling back to the way we're doing it now. 

I'll think on this more a bit tonight and see if something clicks and we can find a middle ground.

**Comment by nikosmoum** [comment]
I agree that JobBuilder is currently a misnomer, so my initial suggestion was to make it more like the name suggests, which means it is a temporary utility for building a complex object in a readable/maintainable way, that also performs validation, and is a separate concern to queueing. So a JobBuilder can be used to build a AsyncJobStatus (at this point unmanaged, as if a DTO), which can then simply be passed into JobManager.queueJob() by the caller (JobBuilder itself shouldn't be queueing anything itself). 

I also agree that we shouldn't have an intermediate object like Job, which wouldn't provide any value (if we want to make the naming more clear we can rename JobBuilder to JobStatusBuilder or AsyncJobStatusBuilder). One of the things that felt awkward when reading the code was that this method `private AsyncJobStatus buildJobStatus(JobBuilder builder)` does not belong in the JobManager. Constructing objects (and validating them) is the job of a builder, and we already have a JobBuilder, so why not move that method there (extended JobBuilders' `build()` can then do their validation and then call `super.build()` to return an AsyncJobStatus)?

I'm also fine with going with the `.validate()` approach, and using an internally extended builder that is accessed with `ExportJob.getJobBuilder()` (as opposed to using a new file for each class that extends JobBuilder).

The value we get from both the `.build()` or `.validate()` approaches is readability, and separation between construction/validation and queueing (not having a bloated JobManager), and early validation (before queueing). Finally, maybe we are overengineering this, and using the old approach is fine - the reason I made my original comment was because it was a bit weird having a JobBuilder that doesn't build anything, and we can simply get away by renaming it to Job, as Kevin pointed out.

**Comment by Ceiu** [comment]
Okay, I've updated the branch with a mix of the above approaches. Let me know if this jives with everyone.

**Comment by kahowell** [comment]
> Okay, I've updated the branch with a mix of the above approaches. Let me know if this jives with everyone.

I dig it! :clap: :smile:

File: server/src/main/java/org/candlepin/resource/ConsumerResource.java [hunk 3]
```diff
@@ -2375,15 +2383,22 @@ public JobDetail exportDataAsync(
...
         Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
         ConsumerType ctype = this.consumerTypeCurator.getConsumerType(consumer);
 
+        Owner test = consumer.getOwner();
+
         Owner owner = ownerCurator.findOwnerById(consumer.getOwnerId());
 
-        return manifestManager.generateManifestAsync(consumerUuid, owner, cdnLabel, webAppPrefix,
-            apiUrl, getExtensionParamMap(extensionArgs));
+        log.info("TEST: {} => {}", test, owner);
```
nikosmoum: Better to remove this (and the added line above) since it is accidentally leftover test logging.

Ceiu: Done

nikosmoum:
Merging since the failing tests are the usual suspects that have been already removed from master for being useless.

PR #2333, opened 2019-05-17
1664481: match syspurpose attributes case insensitively [ENT-1284]

nikosmoum:
- During auto-attach and syspurpose status calculations, syspurpose
  attribute values are now compared with case insensitivity.

### DIFF
--- a/server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
+++ b/server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
@@ -20,9 +20,7 @@ import org.candlepin.model.ConsumerCurator;
 import org.candlepin.model.ConsumerType;
 import org.candlepin.model.ConsumerTypeCurator;
 import org.candlepin.model.Entitlement;
-import org.candlepin.model.EntitlementCurator;
 import org.candlepin.model.Product;
-import org.candlepin.policy.js.compliance.StatusReasonMessageGenerator;
 import org.candlepin.policy.js.compliance.hash.ComplianceStatusHasher;
 
 import com.google.inject.Inject;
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
```diff
@@ -168,8 +161,11 @@ public SystemPurposeComplianceStatus getStatus(Consumer consumer,
                     product.hasAttribute(Product.Attributes.ROLES)) {
                     List<String> roles =
                         Arrays.asList(product.getAttributeValue(Product.Attributes.ROLES).split("\\s*,\\s*"));
-                    if (roles.contains(consumer.getRole())) {
-                        status.addCompliantRole(consumer.getRole(), entitlement);
+
+                    final String unsatisfedRoleFinalCopy = unsatisfedRole;
```
Ceiu: Do we still need the final ref with lambdas? I thought I remembered reading that they fixed that part in Java9+.

nikosmoum: I am not aware of such a fix, but aren't we using java 8 still anyway?

Ceiu: Right... bit of a derp moment here for me.

I just tried googling it and nothing immediately came up. Maybe wishful thinking.

Ceiu: Scratch that. I just found that in Java8 they introduced something called "effectively final" -- or a variable that doesn't change after initialization. That might apply here, as the unsatisifiedRole doesn't change during the loop, and technically is redeclared on every iteration.

nikosmoum: Unfortunately it doesn't apply here. `unsatisifiedRole` is not effectively final because if we get a match for it, it is re-assigned to null (and this is required because after every iteration, we check for it: `StringUtils.isNotEmpty(unsatisfedRole)`. We still need to create/assign `unsatisfedRoleFinalCopy`, but we don't need the `final` identifier.

nikosmoum:
@Ceiu I've moved the trim calls to the product attribute list building, and also added trim calls to the consumer's side of attribute values. I added unit tests for these usecases in `SystemPurposeComplianceRulesTest`.

server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
Ceiu: Yeah. If we have to redeclare it for the sake of a lambda, I'm okay with leaving it final, as that's a pretty well-known pattern for such things.

PR #2335, opened 2019-05-21
1697518: Syspurpose status does not reflect system purpose matching status

Januson:
- Added missing @Transactional
- UpdateConsumer endpoint now updates consumer's system purpose status

### DIFF
--- a/server/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -1451,7 +1451,7 @@ public class ConsumerResource {
 
             // this should update compliance on toUpdate, but not call the curator
             complianceRules.getStatus(toUpdate, null, false, false);
-            systemPurposeComplianceRules.getStatus(toUpdate, toUpdate.getEntitlements(), null, false, false);
+            systemPurposeComplianceRules.getStatus(toUpdate, toUpdate.getEntitlements(), null, false, true);
 
             Event event = eventBuilder.setEventData(toUpdate).buildEvent();
             sink.queueEvent(event);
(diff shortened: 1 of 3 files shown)


Januson:
Is it ok to update system purpose in updateConsumer? There is a comment from ckozak stating: "this should update compliance on toUpdate, but not call the curator".

Ceiu:
I'll do some follow up with the people who worked on system purpose stuff tomorrow (wottop, specifically), but the comment you're referencing was written by someone who left the team a few years before system purpose was a thing. It's likely referring specifically to the line directly under it and not the syspurpose call updated in this PR.

However, my guess, based on that comment and the line above it, is that the intent is for the curator's merge/update call to be done later by the caller. Whether or not that's still true today and/or possible with how CP has changed around it since it was written is up for debate.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1451,7 +1451,7 @@ public boolean performConsumerUpdates(ConsumerDTO updated, Consumer toUpdate,
 
             // this should update compliance on toUpdate, but not call the curator
             complianceRules.getStatus(toUpdate, null, false, false);
-            systemPurposeComplianceRules.getStatus(toUpdate, toUpdate.getEntitlements(), null, false, false);
+            systemPurposeComplianceRules.getStatus(toUpdate, toUpdate.getEntitlements(), null, true, true);
```
Ceiu: We still want the updateConsumer parameter set to false here.

nikosmoum:
@Januson Since the issue does not occur in master, and we're only going to be applying this fix in 2.3.-HOTFIX, I've updated the bz to have target release 2.3, so the bugzilla-reference build should pass now.

server/src/main/java/org/candlepin/resource/ConsumerResource.java

PR #2338, opened 2019-05-22
1418996: Need string change for the message "Error subscribing: Owner has auto-bind disabled" (ENT 1340)

ShwethaKalyanaraman:


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2028,7 +2028,7 @@ public Response bind(
         if (owner.isAutobindDisabled()) {
             String caMessage = owner.isContentAccessEnabled() ?
                 " because of the content access mode setting" : "";
-            throw new BadRequestException(i18n.tr("Owner has autobind disabled" + caMessage + "."));
+            throw new BadRequestException(i18n.tr("Ignoring request to auto-attach. It is disabled for org 'admin'" + caMessage + "."));
```
nikosmoum: I'm not sure if this is the right wording. According to [comment 3](https://bugzilla.redhat.com/show_bug.cgi?id=1418996#c3) in the bz, which is the latest update of the discussion there, it should be 'Organization <ORG> has auto-attach disabled.'.

Also, we don't want to hardcode `'admin'` here, since that is just the name/key of the org used in the example. The string should be parameterised with the name of the current org being used (`owner.getKey()`).

nikosmoum:
There's some stylistic changes needed which the jenkins-checkstyle build is complaining about. Also, the 4 commits currently on the PR should be squashed into 1 with a descriptive message before we merge, since that helps keep the commit log clean (e.g. dwdjwhdn::n is not really helpful as a message, and also, there is no need to merge the candlepin master branch into your branch, since I don't expect there would be any conflicts with such a simple change).

ShwethaKalyanaraman:
Changes made as suggested. Please review it again.

PR #2339, opened 2019-05-29
ENT-1314: Modify the locking and transaction size for the hypervisor report

Januson:
- Split transaction into a series of bulk operations per 10 items

### DIFF
--- a/server/src/main/java/org/candlepin/guice/CandlepinModule.java
+++ b/server/src/main/java/org/candlepin/guice/CandlepinModule.java
@@ -153,6 +153,7 @@ import org.candlepin.resteasy.filter.VerifyAuthorizationFilter;
 import org.candlepin.resteasy.filter.VersionResponseFilter;
 import org.candlepin.service.UniqueIdGenerator;
 import org.candlepin.service.impl.DefaultUniqueIdGenerator;
+import org.candlepin.service.impl.HypervisorUpdateAction;
 import org.candlepin.swagger.CandlepinSwaggerModelConverter;
 import org.candlepin.sync.ConsumerExporter;
 import org.candlepin.sync.ConsumerTypeExporter;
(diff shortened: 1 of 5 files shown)


barnabycourt:
I see where the flush() is called by bulkUpdate() but the transaction is still open. Won't this still be holding locks for all the hypervisors for the duration of the processing?

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -622,6 +622,34 @@ private void validateFacts(Consumer consumer) {
         }
     }
 
+    /**
+     * @param consumers consumers to create
+     * @return created consumers
+     */
+    @Transactional
+    public Set<Consumer> bulkCreate(Set<Consumer> consumers) {
```
Ceiu: These methods already exist as saveAll and deleteAll in AbstractHibernateCurator

Januson: Done.

Januson:
> I see where the flush() is called by bulkUpdate() but the transaction is still open. Won't this still be holding locks for all the hypervisors for the duration of the processing?

I extracted a piece of Job's logic into a separate class so that the job itself doesn't have to be transactional. Since job is not transactional, transaction should end when the bulkUpdate method returns.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: :+1:

wottop:
Those spec failures need to be fixed. You are getting 404 errors on get_consumer.

Januson:
> The cause of the spec failures need to be fixed. You are getting 404 errors on get_consumer.

Done.

Ceiu:
@wottop Is this still needing review or are we good to merge with the fixed spec test ?

wottop:
@Ceiu That was my only concern. If you are good, go ahead.

PR #2340, opened 2019-05-29
Add task to generate candlepin-api jar

kahowell:


barnabycourt:
@kahowell Does anything else need to be done to ensure that regular CP builds push an API jar to nexus along with the war?

server/build.gradle
```diff
@@ -58,6 +58,22 @@ task generateConfig() {
     }
 }
 
+task apiJar(type: Jar) {
```
barnabycourt: Please add a comment with context as to why this exists here for others that come back to it in the future.

kahowell: Done!

PR #2341, opened 2019-05-30
[F] ENT-847: Port ImportJob To The New Messaging Framework

nikosmoum:
* Now the ImportJob is using the new artemis messaging framework.
* Made sure that in case of an ImportJob failure due to conflicts,
  the result data returned include the list of conflicts.
* Removed some unused constructor arguments from OwnerResource.

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -1086,7 +1086,6 @@ public class JobManager implements ModeChangeListener {
         // Rollback any unsent events
         eventSink.rollback();
 
-        // Only store the exception type and the message. Luckily for us, toString does exactly that.
         String result = throwable != null ? throwable.toString() : null;
 
         if (retry) {
--- a/server/src/main/java/org/candlepin/async/tasks/ExportJob.java
+++ b/server/src/main/java/org/candlepin/async/tasks/ExportJob.java
@@ -66,7 +66,8 @@ public class ExportJob implements AsyncJob {
         }
 
         /**
-         * Sets the consumer for this export job. The consumer must be
+         * Sets the consumer for this export job. The consumer is required, and also provides the
+         * context in which the job will be executed.
          *
          * @param consumer
          *  the consumer to set for this job
@@ -170,12 +171,12 @@ public class ExportJob implements AsyncJob {
                 String consumerUuid = arguments.getAsString(CONSUMER_KEY);
                 String cdnLabel = arguments.getAsString(CDN_LABEL);
 
-                if (!(consumerUuid instanceof String) || ((String) consumerUuid).isEmpty()) {
+                if (consumerUuid == null || consumerUuid.isEmpty()) {
                     String errmsg = "consumer has not been set, or the provided consumer lacks a UUID";
                     throw new JobConfigValidationException(errmsg);
                 }
 
-                if (!(cdnLabel instanceof String) || ((String) cdnLabel).isEmpty()) {
+                if (cdnLabel == null || cdnLabel.isEmpty()) {
                     String errmsg = "CDN label has not been set, or the provided label is empty";
                     throw new JobConfigValidationException(errmsg);
                 }
--- a/server/src/main/java/org/candlepin/async/tasks/ImportJob.java
+++ b/server/src/main/java/org/candlepin/async/tasks/ImportJob.java
@@ -1,0 +1,229 @@
+/**
+ * Copyright (c) 2009 - 2019 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.async.tasks;
+
+import com.google.inject.Inject;
+import org.candlepin.async.ArgumentConversionException;
+import org.candlepin.async.AsyncJob;
+import org.candlepin.async.ImportConflictJobException;
+import org.candlepin.async.JobArguments;
+import org.candlepin.async.JobConfig;
+import org.candlepin.async.JobConfigValidationException;
+import org.candlepin.async.JobConstraints;
+import org.candlepin.async.JobExecutionContext;
+import org.candlepin.async.JobExecutionException;
+import org.candlepin.common.exceptions.NotFoundException;
+import org.candlepin.controller.ManifestManager;
+import org.candlepin.model.ImportRecord;
+import org.candlepin.model.Owner;
+import org.candlepin.model.OwnerCurator;
+import org.candlepin.sync.ConflictOverrides;
+import org.candlepin.sync.ImportConflictException;
+import org.candlepin.sync.ImporterException;
+import org.candlepin.sync.file.ManifestFileService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Runs an asynchronous manifest import. This job expects that the manifest file was already uploaded.
+ */
+public class ImportJob implements AsyncJob {
+    private static Logger log = LoggerFactory.getLogger(ImportJob.class);
+
+    public static final String JOB_KEY = "IMPORT_JOB";
+    public static final String JOB_NAME = "import_manifest";
+
+    protected static final String OWNER_KEY = "org";
+    protected static final String STORED_FILE_ID = "stored_manifest_file_id";
+    protected static final String CONFLICT_OVERRIDES = "conflict_overrides";
+    protected static final String UPLOADED_FILE_NAME = "uploaded_file_name";
+
+    /**
+     * Job configuration object for the import job
+     */
+    public static class ImportJobConfig extends JobConfig {
+        public ImportJobConfig() {
+            this.setJobKey(JOB_KEY)
+                .setJobName(JOB_NAME)
+                .addConstraint(JobConstraints.uniqueByArgument(OWNER_KEY));
+        }
+
+        /**
+         * Sets the owner for this import job. The owner is required, and also provides the org
+         * context in which the job will be executed.
+         *
+         * @param owner
+         *  the owner to set for this job
+         *
+         * @return
+         *  a reference to this job config
+         */
+        public ImportJobConfig setOwner(Owner owner) {
+            if (owner == null) {
+                throw new IllegalArgumentException("owner is null");
+            }
+
+            // The owner is both part of metadata & arguments in this job.
+            this.setJobMetadata(OWNER_KEY, owner.getKey())
+                .setJobArgument(OWNER_KEY, owner.getKey())
+                .setLogLevel(owner.getLogLevel());
+
+            return this;
+        }
+
+        /**
+         * Sets the id of the stored file on the {@link ManifestFileService}.
+         *
+         * @param storedFileId
+         *  the stored file id to set for this job
+         *
+         * @return
+         *  a reference to this job config
+         */
+        public ImportJobConfig setStoredFileId(String storedFileId) {
+            this.setJobArgument(STORED_FILE_ID, storedFileId);
+            return this;
+        }
+
+        /**
+         * Sets the file name of the uploaded file on the {@link ManifestFileService}.
+         *
+         * @param uploadedFileName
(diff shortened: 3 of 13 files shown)


server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1086,8 +1087,18 @@ private AsyncJobStatus processJobFailure(AsyncJobStatus status, EventSink eventS
...
 
-        // Only store the exception type and the message. Luckily for us, toString does exactly that.
-        String result = throwable != null ? throwable.toString() : null;
+        Object result = null;
+        if (throwable != null) {
+            // CandlepinException subclasses can override the message() method to enhance the message from a
+            // simple String to something with more useful information, like ImportConflictException does
+            // to include specific import conflicts that happened.
+            if (throwable.getCause() instanceof CandlepinException) {
+                result = ((CandlepinException) throwable.getCause()).message();
```
nikosmoum: @Ceiu @kahowell Regarding your previous [comment](https://github.com/candlepin/candlepin/pull/2331#issuecomment-495802031) about reducing the exception result to a message string, and the potential for using a "failure" class, it turns out that we already have a mechanism for subclasses of `CandlepinException` to enhance their "message" to be more than a single string. This is currently used only by `ImportConflictException`, in order to add a list of "conflicts" that can happen during an import failure, which are surely useful to the client.

From my point of view, this is enough for now, until we need anything more complex in the future.

Ceiu: Maybe. For some context, CandlepinExceptions are generally used for exceptions to be passed out of the resources directly to the end-user via RESTeasy. 

I'll take a look at this a bit closer during review.

Ceiu: Looking at the replaced code, this may work, but it's likely going to be extra work to get the message. The replaced line which calls .toString will, by default, give us a string equivalent to the following:
```String msg = throwable.getClass().getName() + ": " + throwable.getLocalizedMessage()```

The .getLocalizedMessage call will call into .getMessage(), which in the case of CandlepinException, will return the same thing as .message().getDisplayMessage()/.message().toString(). Basically, we get similar (identical?) output without needing the additional checks here.

File: server/src/main/java/org/candlepin/async/tasks/ExportJob.java [hunk 2]
```diff
@@ -170,12 +171,12 @@ public void validate() throws JobConfigValidationException {
                 String consumerUuid = arguments.getAsString(CONSUMER_KEY);
                 String cdnLabel = arguments.getAsString(CDN_LABEL);
 
-                if (!(consumerUuid instanceof String) || ((String) consumerUuid).isEmpty()) {
+                if (consumerUuid == null || consumerUuid.isEmpty()) {
```
Ceiu: :+1:

The instanceof check was leftover from before the switch to the type-safe JobArguments. Thanks for updating this.

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
```diff
@@ -0,0 +1,226 @@
...
+        this.ownerCurator = ownerCurator;
+        this.manifestManager = manifestManager;
+    }
+
+    @Override
+    public Object execute(JobExecutionContext context) throws JobExecutionException {
+        JobArguments args = context.getJobArguments();
+
+        String ownerKey = args.getAsString(OWNER_KEY);
+        ConflictOverrides overrides = new ConflictOverrides(args.getAs(CONFLICT_OVERRIDES, String[].class));
```
Ceiu: You should be able to pull the conflict overrides directly without needing to rebuild from the string array. If that's not working, there's a problem in the new serializer paths that we need to address. Ideally the following line should work:
```ConflictOverrides overrides = args.getAs(CONFLICT_OVERRIDES, ConflictOverrides.class)```

If so, the setter above should also change accordingly.

File: server/src/main/java/org/candlepin/async/tasks/ImportJob.java [hunk 4]
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        manifestManager.recordImportFailure(targetOwner, caught, uploadedFileName);
+
+        // If an exception was thrown, the importer's transaction was rolled
+        // back. We want to make sure that the file gets deleted so that it
+        // doesn't take up disk space. It may be possible that the file was
+        // already deleted, but we attempt it anyway.
+        manifestManager.deleteStoredManifest(storedFileId);
+        throw new JobExecutionException(caught.getMessage(), caught, false);
```
Ceiu: This, along with the note above, probably requires a bit more thought. I imagine some exceptions will be non-terminal (DB-related ones or otherwise unexpected exceptions), whereas the SyncDataFormat and ImporterExceptions are probably terminal exceptions and should be flagged as such.

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
```diff
@@ -0,0 +1,226 @@
...
+            ImportRecord importRecord = manifestManager.importStoredManifest(targetOwner,
+                storedFileId, overrides, uploadedFileName);
+            log.info("Async import complete.");
+            return importRecord;
+        }
+        // Catch and handle SyncDataFormatException and ImporterExceptions
+        // as the OwnerResource.importManifest does which will provide a little more
+        // info about the exception that was thrown (CandlepinException).
+        catch (SyncDataFormatException e) {
+            caught = new BadRequestException(e.getMessage(), e);
```
Ceiu: I know this is how the Pinsetter version handled this, but is this what we want to continue doing? Something about the job system using exceptions intended for REST output gives me pause. Also, we have the terminal flag to juggle with some of these, so it should probably require a bit more attention (more on this below).

server/src/main/java/org/candlepin/async/JobManager.java
Ceiu: Also, if we ever want to standardize this handling a bit more, this gist [1] is something I had in the prototyping stage that was eventually cut due to it being a similar amount of boilerplate that throwable's default .toString already does for us.

[1] https://gist.github.com/Ceiu/b401efa25b1eda7637958b17eef8bf4b

nikosmoum: Unfortunately, the replaced code does not work, because the "message" returned by the `.toString()` method is a simple String which contains only the message of the exception (it resides on `Throwable.java` as such: `private String detailMessage;`), while CandlepinException has a completely different, custom `message` field which is a: `protected final ExceptionMessage message;`. In the case of the `ImportConflictException`, this is further extended as `ConflictExceptionMessage`, which, when serialized looks like this:
```
{
  "result" :
   {
    "displayMessage": "Import is the same as existing data",
    "requestUuid": "ff8080816b08d246016b08d2f35f0087",
    "conflicts": [ "MANIFEST_SAME", "DISTRIBUTOR_CONFLICT" ]
   }
}
```
, while the replaced code returns:
```
{
  "result" : "org.candlepin.sync.ImportConflictException: Import is the same as existing data"
}
```
CandlepinException does not can cannot override `getMessage()`, because the signature returns a simple String. This is why `message()` is used instead.

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
nikosmoum: This does not work; when setting, the object is not serialized properly.

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
nikosmoum: I agree the REST codes do not really fit in a job failure output.
I'm happy to use your `JobFailureResult` gist, which can then be extended.
e.g.:
* Instead of `SyncDataFormatException`, we can use a `JobFailureResult` with appropriate type/message. Similar for `ImporterException`, and every exception jobs currently throw that do not have any extra data on them.
* Instead of `ImportConflictException`, we can extend `JobFailureResult` with a `private Set<Importer.Conflict> conflicts;`. Similarly for every other exception jobs currently throw that needs to have extra data on them?

server/src/main/java/org/candlepin/async/JobManager.java
Ceiu: Argh... okay. And because we're setting the exception, the result is getting translated away from that. I wonder if the special handling of the CandlepinException is something we should be handling specifically in this job, since it seems to be unique to it.

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
Ceiu: Interesting. We set other objects and pull them out fine. I wonder what's broken with serialization this time...?

In any event, I don't really want to keep adding TODO tasks to the epic, so let me poke around a bit and see how complex it'll be to fix. If it's going to be a large undertaking, we can leave this as-is for now and fix it in another PR. Otherwise if it's a simple fix, I'd rather we do that here instead of working around the problem.

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
Ceiu: I guess what I'm getting at is that I'd rather exception handling be standardized, but even with the gist, I don't know how much value that will have. The dissonance we're seeing here is due to the special ExceptionMessage object CandlepinExceptions carry with them (which itself is kinda... eh... in my opinion). I'll take a quick glance at the other jobs; hopefully it's not something we're using elsewhere. And if that's the case, maybe the exception handling here should be updated to, at least, capture the CandlepinExceptions and disassembly manually, rather than propagating that up to the JobManager.

server/src/main/java/org/candlepin/async/JobManager.java
Ceiu: Took a quick look through the other pinsetter jobs. Only the following the HealEntireOrgJob and HypervisorUpdateJob use CandlepinExceptions, and even those are both BadRequestExceptions which can easily be replaced by a generic JobExecutionException. So, this logic seems to be something unique to the ImportJob. At least, that's all I saw with a surface-level glance through the jobs.

The difficult problem to address here is how to both set the result and throw an exception to indicate the job status in a way that lets us control both without some convoluted wrapper or output scheme. We used a special listener in Pinsetter to run code similar to what you've added to JobManager, but I feel like we can do better. Let me think on this a bit.

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
Ceiu: Looking into it a bit more, I see why it's serializing the way it is. I've written a lengthy email about this, so what we do here could vary. For now, assume we're going to leave it as written in this PR, and we'll address it as necessary depending on how discussions and planning go.

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
nikosmoum: I agree with this approach. I'll go with capturing CandlepinExceptions and propagating JobExecutionExceptions wrapping only the original non-CandlepinExceptions (e.g. SyncDataFormatException and ImporterException, but not BadRequestException or IseException). I've also commented on the other thread about details on how to deal with the unique nature of ImportConflictException,

server/src/main/java/org/candlepin/async/JobManager.java
nikosmoum: Since this is a unique case, this is what I've come up with:
* Leave the `JobManager` set the result as it originally was (`String result = throwable != null ? throwable.toString() : null;`). This works fine currently with all exceptions that jobs are throwing, except `ImportConflictException`.
* Have ImportJob transform the `ImportConflictException` to a new, [`ImportConflictJobException` (which extends `JobExecutionException`)](https://gist.github.com/nikosmoum/49dad9e7673245a42deedecb402b45ba) which provides the list of conflicts as a result of its `.toString()`. This way
  - We are being compatible with what `JobManager` does for all other exceptions.
  - We propagate a `JobExecutionException`, not a `CandlepinException`
  - We save the valuable part of the data (the list of conflicts and display message) in the result, without having the useless/unfit for purpose data(requestUuid & REST code).

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
nikosmoum: Fixed.

server/src/main/java/org/candlepin/async/JobManager.java
Ceiu: I'm okay with this for now.

We still have the interesting problem of resolving what to do when we want to set a result while throwing an exception. When I get back, I'll look into a kind of hybrid approach between what you have here and the ExceptionMessage idea from CandlepinException. That might work well enough to cover all of these type of cases in the future.

server/src/main/java/org/candlepin/async/tasks/ImportJob.java
Ceiu: ack

nikosmoum:
@Ceiu commits squashed

Ceiu:
ack

We can address the remaining oddities with future work next sprint.

PR #2342, opened 2019-05-30
1600588: [RFE] Add sorting by product to GET /owners/{owner_key}/pools  (ENT-745)

ShwethaKalyanaraman:
Added an additional endpoint GET /owners/{owner_key}/pools_ordered_by_product along with Junit tests.

### DIFF
--- a/server/src/main/java/org/candlepin/model/PoolCurator.java
+++ b/server/src/main/java/org/candlepin/model/PoolCurator.java
@@ -579,7 +579,8 @@ public class PoolCurator extends AbstractHibernateCurator<Pool> {
         if (poolIds != null && !poolIds.isEmpty()) {
             criteria = this.currentSession()
                 .createCriteria(Pool.class)
-                .add(CPRestrictions.in("id", poolIds));
+                .createAlias("product", "Product");
+            criteria.add(CPRestrictions.in("id", poolIds));
 
             return this.listByCriteria(criteria, pageRequest, postFilter);
         }
--- a/server/src/test/java/org/candlepin/model/PoolCuratorTest.java
+++ b/server/src/test/java/org/candlepin/model/PoolCuratorTest.java
@@ -23,6 +23,7 @@ import org.candlepin.common.paging.Page;
 import org.candlepin.common.paging.PageRequest;
 import org.candlepin.config.DatabaseConfigFactory;
 import org.candlepin.controller.CandlepinPoolManager;
+
 import org.candlepin.model.ConsumerType.ConsumerTypeEnum;
 import org.candlepin.model.activationkeys.ActivationKey;
 import org.candlepin.model.dto.Subscription;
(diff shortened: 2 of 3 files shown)


server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1528,6 +1528,145 @@ public int countConsumers(
...
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("{owner_key}/pools_ordered_by_product")
+    @SuppressWarnings("checkstyle:indentation")
+    @ApiOperation(notes = "Retrieves a list of Pools ordered by Product for an Owner", value = "List Pools")
+    @ApiResponses({
+            @ApiResponse(code = 404, message = "Owner not found"),
+            @ApiResponse(code = 400, message = "Invalid request")
+    })
+    public List<PoolDTO> listPoolsOrderedByProduct(
```
Ceiu: With the addition of the product alias, is this something where we need another end point, or does passing in "Product.name" in as the argument for the "sort_by" parameter on the base getPools end point work as one would expect?

server/src/test/java/org/candlepin/model/PoolCuratorTest.java
```diff
@@ -1464,6 +1462,63 @@ public void testGetPoolsByFilter() {
...
+
+        List<Pool> pools=page.getPageData();
+
+        List<Pool>results=new ArrayList<>();
+        results.add(p2);
+        results.add(p3);
+        results.add(p1);
+
+
+        //Checking for Descending
```
Ceiu: This should probably be a second test

server/src/main/java/org/candlepin/resource/OwnerResource.java
ShwethaKalyanaraman: We need the alias for the new end point to work

ShwethaKalyanaraman:
Made the changes as suggested.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -579,7 +579,8 @@ else if (!addFuture) {
         if (poolIds != null && !poolIds.isEmpty()) {
             criteria = this.currentSession()
                 .createCriteria(Pool.class)
-                .add(CPRestrictions.in("id", poolIds));
+                .createAlias("product", "Product");
+                criteria.add(CPRestrictions.in("id", poolIds));
```
Ceiu: This should be unindented, or we can drop the the criteria reference and move the .add to above the creation of the product alias.

server/src/main/java/org/candlepin/resource/OwnerResource.java
Ceiu: But we do not need a new endpoint for the functionality. Once the product alias is present in the criteria, we can use the existing paging infrastructure to provide the sorting functionality without needing a new endpoint.

For example, the following curl command demonstrates this in your branch:
```curl -k -u admin:admin "https://localhost:8443/candlepin/owners/admin/pools?include=id&include=productName&sort_by=Product.name&order=ASC"```

Whereas the same command on master results in:
```{
  "displayMessage" : "Runtime Error could not resolve property: Product of: org.candlepin.model.Pool at org.hibernate.persister.entity.AbstractPropertyMapping.propertyException:73",
  "requestUuid" : "e5e2ceb1-88fe-4db1-8a2f-b86ad78948ea"
}```

If there's a compelling reason to keep both endpoints, then at the very least this new method should simply call into the existing one after configuring the paging bits as it's doing now. But, again, if that's all it's doing, then there's really no reason to not have the users just specify the sorting as necessary.

File: server/src/test/java/org/candlepin/model/PoolCuratorTest.java [hunk 4]
```diff
@@ -1464,6 +1462,78 @@ public void testGetPoolsByFilter() {
...
+        this.poolCurator.create(p3);
+        this.poolCurator.create(p1);
+        this.poolCurator.create(p2);
+        PageRequest req1 = new PageRequest();
+        req1.setSortBy("Product.name");
+        req1.setOrder(PageRequest.Order.DESCENDING);
+
+        Date activeOn1 = TestUtil.createDate(2019, 2, 2);
+
+        Page<List<Pool>> page1 = poolManager.listAvailableEntitlementPools(null, null,  owner1.getId(), null, null, activeOn1, false, null,
```
Ceiu: This line (and the other matching one in the new test) is too long for the line limits.

The gradle linting tests appear to be broken, so for the interim, make sure you run the following command to check code style:
```buildr checkstyle```

File: server/src/test/java/org/candlepin/resource/OwnerResourceTest.java [hunk 5]
```diff
@@ -641,6 +641,28 @@ public void ownerAdminCannotAccessAnotherOwnersPools() {
...
+        List<PoolDTO> pools = ownerResource.listPoolsOrderedByProduct(owner.getKey(),
+                null, null, null, null, true, null, null, new ArrayList<>(), false, false, null, null,
+                principal, null);
+        assertEquals(2, pools.size());
+
+    }
+
+
+
+
```
Ceiu: Ton of extra whitespace here

server/src/test/java/org/candlepin/model/PoolCuratorTest.java
barnabycourt: ./gradlew check is showing the error regarding line length

barnabycourt: I was wrong, the cleanest way to ensure you see the issues is "./gradlew checkstyleTest checkstyleMain", I submitted a separate PR to update the the docker/jenkins script to run both targets.

ShwethaKalyanaraman:
Please review the changes. @Ceiu

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -579,7 +579,8 @@ else if (!addFuture) {
         if (poolIds != null && !poolIds.isEmpty()) {
             criteria = this.currentSession()
                 .createCriteria(Pool.class)
-                .add(CPRestrictions.in("id", poolIds));
+                    .createAlias("product", "Product");
```
Ceiu: Almost. :)

The .createAlias should be at the same indentation as the previous line, with the following line as it is now.

PR #2343, opened 2019-06-03
[F] ENT-850: Port RefreshPoolsForProductJob to new framework

Januson:


server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * Job configuration object for the export job
+     */
+    public static class RefreshPoolsForProductConfig extends JobConfig {
+
+        RefreshPoolsForProductConfig(final Product product, final boolean lazy) {
+            this.setJobKey(JOB_KEY)
+                .setJobName(JOB_NAME + Util.generateUUID())
```
Ceiu: UUID generation is no longer needed for the new job system

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -458,19 +463,23 @@ public void deleteProduct(
...
             return null;
         }
 
-        Owner owner = this.getOwnerByKey(ownerKey);
-        Product product = this.fetchProduct(owner, productId);
+        final Owner owner = this.getOwnerByKey(ownerKey);
+        final Product product = this.fetchProduct(owner, productId);
 
-        return RefreshPoolsForProductJob.forProduct(product, lazyRegen);
+        return this.translator.translate(
```
Ceiu: It'll probably be cleanest to declare each step independently. Our style requirements are silly when it comes to multiline indentation, and there's really no reason to cram everything into one line here.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -0,0 +1,147 @@
...
+                .setJobArgument(JobStatus.TARGET_TYPE, JobStatus.TargetType.PRODUCT)
+                .setJobArgument(JobStatus.TARGET_ID, product.getUuid())
+                .setJobArgument(LAZY_REGEN, lazy)
+                .setJobArgument(JobStatus.CORRELATION_ID, MDC.get(LoggingFilter.CSID_KEY));
```
Ceiu: This is no longer needed

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -0,0 +1,147 @@
...
+    public static JobConfig createJobConfig(final Product product, final boolean lazy) {
+        return new RefreshPoolsForProductConfig(product, lazy);
+    }
+
+    /**
+     * Job configuration object for the export job
+     */
+    public static class RefreshPoolsForProductConfig extends JobConfig {
+
+        RefreshPoolsForProductConfig(final Product product, final boolean lazy) {
```
Ceiu: This somewhat violates the fluent-style design we're going with for the per-job configuration objects. Ideally this is a default constructor and, (optionally,) provides setters for the required parameters. Take a look at the updated ExportJob and the PR for the ImportJob for examples.

Additionally, there's no validation on the inbound Product object before we go using it down on line 116.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -458,19 +463,23 @@ public void deleteProduct(
...
 
         if (config.getBoolean(ConfigProperties.STANDALONE)) {
             log.warn("Ignoring refresh pools request due to standalone config.");
             return null;
         }
 
-        Owner owner = this.getOwnerByKey(ownerKey);
-        Product product = this.fetchProduct(owner, productId);
+        final Owner owner = this.getOwnerByKey(ownerKey);
+        final Product product = this.fetchProduct(owner, productId);
```
Ceiu: Let's not add the final keyword to [existing] code unnecessarily. I'm begrudgingly accepting of it in net-new code, but let's not go adding churn to pre-existing stuff for it.

Januson: Ok. I can tune it a bit down. I like to be explicit with things like mutability. ... Why you don't like final?

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
Januson: Done.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
Januson: Ok. Done.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
Januson: Does it make sense to provide mandatory parameters through setters? It causes quite strong temporal coupling as it makes it easy to forget or accidentally remove required parameter and the only way to know that it belongs there is to look at the implementation.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -0,0 +1,150 @@
...
+     */
+    public static class RefreshPoolsForProductConfig extends JobConfig {
+
+        RefreshPoolsForProductConfig() {
+            this.setJobKey(JOB_KEY)
+                .setJobName(JOB_NAME)
+                .setJobArgument(JobStatus.TARGET_TYPE, JobStatus.TargetType.PRODUCT);
+        }
+
+        public RefreshPoolsForProductConfig setTarget(final Product product) {
```
nikosmoum: We shouldn't use the word `target` in method names / static flags any more. `setProduct` is a much more intuitive name. See the `ImportJob` and `ExportJob` examples.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -458,10 +463,10 @@ public void deleteProduct(
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes(MediaType.WILDCARD)
     @Transactional
-    public JobDetail refreshPoolsForProduct(
+    public AsyncJobStatusDTO refreshPoolsForProduct(
         @PathParam("owner_key") String ownerKey,
         @PathParam("product_id") String productId,
-        @QueryParam("lazy_regen") @DefaultValue("true") Boolean lazyRegen) {
+        @QueryParam("lazy_regen") @DefaultValue("true") boolean lazyRegen) throws JobException {
```
nikosmoum: Not sure it is a good idea to change to primitive here. I think we're using the wrappers for some reason (which I can't remember right now). @Ceiu might know.

server/src/test/java/org/candlepin/async/tasks/RefreshPoolsForProductJobTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        final Object actualResult = job.execute(context);
+
+        assertEquals(
+            "Unable to refresh pools for product \"null\": Could not find a product with the specified UUID",
+            actualResult);
+    }
+
+    @Test(expected = JobConfigValidationException.class)
+    public void productUuidMustBePresent() throws JobConfigValidationException {
```
nikosmoum: One more small test for the happy path of `.validate` would be good (e.g. that product with a proper uuid and lazy flag are both set and no exception thrown.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Job configuration object for the export job
+     */
+    public static class RefreshPoolsForProductConfig extends JobConfig {
+
+        RefreshPoolsForProductConfig() {
+            this.setJobKey(JOB_KEY)
+                .setJobName(JOB_NAME)
+                .setJobArgument(JobStatus.TARGET_TYPE, JobStatus.TargetType.PRODUCT);
```
nikosmoum: We have no need for target type arguments any more, so this line can be removed. The `setProduct` method will set the product uuid, which is all we need.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -0,0 +1,150 @@
...
+     *  a JobConfig instance configured to execute the export job
+     */
+    public static RefreshPoolsForProductConfig createJobConfig() {
+        return new RefreshPoolsForProductConfig();
+    }
+
+    /**
+     * Job configuration object for the export job
+     */
+    public static class RefreshPoolsForProductConfig extends JobConfig {
```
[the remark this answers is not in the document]
Januson: Yeah it is very long, but it is also tautological. Since RefreshPoolsForProductJobConfig is nested class we have to use name of its outer class. E.g.

> RefreshPoolsForProductJob.RefreshPoolsForProductJobConfig config

Since the job name is already there and has to be present we could remove it from the name of JobConfig. E.g.

> RefreshPoolsForProductJob.Config config

What do you think?

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
Januson: I switched it to primitive since it has a default value so it is always present.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -471,6 +476,16 @@ public JobDetail refreshPoolsForProduct(
         Owner owner = this.getOwnerByKey(ownerKey);
         Product product = this.fetchProduct(owner, productId);
 
-        return RefreshPoolsForProductJob.forProduct(product, lazyRegen);
+        if (product == null) {
+            throw new BadRequestException(
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -0,0 +1,150 @@
...
+
+        return result.toString();
+    }
+
+    /**
+     * Creates a JobConfig configured to execute the export job. Callers may further manipulate
+     * the JobConfig as necessary before queuing it.
+     *
+     * @return
+     *  a JobConfig instance configured to execute the export job
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -0,0 +1,150 @@
...
+     */
+    public static RefreshPoolsForProductConfig createJobConfig() {
+        return new RefreshPoolsForProductConfig();
+    }
+
+    /**
+     * Job configuration object for the export job
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -0,0 +1,150 @@
...
+        this.productCurator = Objects.requireNonNull(productCurator);
+        this.subAdapter = Objects.requireNonNull(subAdapter);
+        this.ownerAdapter = Objects.requireNonNull(ownerAdapter);
+    }
+
+    @Override
+    public String execute(final JobExecutionContext context) {
+        final JobArguments args = context.getJobArguments();
+
+        final String productUuid = args.getAsString(JobStatus.TARGET_ID);
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
Januson: Done.

server/src/test/java/org/candlepin/async/tasks/RefreshPoolsForProductJobTest.java
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
Januson: Done.

server/src/test/java/org/candlepin/resource/OwnerProductResourceTest.java
```diff
@@ -57,14 +58,15 @@
 public class OwnerProductResourceTest extends DatabaseTestFixture {
 
     @Inject protected ProductManager productManager;
+    @Inject protected JobManager jobManager;
```
nikosmoum: Something I forgot earlier, there is no need to create this field, since it is already present in the parent class `DatabaseTestFixture`.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
nikosmoum: I think it might have to do with being able to explicitly send a null value for that field, and using a primitive might fail doing that (I've not verified this though).

server/src/test/java/org/candlepin/resource/OwnerProductResourceTest.java
Januson: Done.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
Ceiu: In Java, final on non-static, scoped variables does very little. It functions like \*const does in C -- we can't change the reference, but we can still change the object we're referencing. And due to Java's pass-by-value calling convention, the final operator does little more than add six characters to a given variable declaration. In the case of these methods, the value will only change if another developer comes in to change the logic; and if they need to change the value (as is common when working with the entities), the devs will either just remove the modifier, or redeclare it as non-final, wasting another 4 bytes of stack space.

In C/C++ and other such languages, it has a lot more value beyond just some syntax (even though you can cast it off), but in Java, it does almost nothing for us.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
Ceiu: Yeah, that's a fair point that's been discussed a couple times. The options were basically parameter soup methods as we've had in the past, or the fluent-style syntax which conveys less information about requirements, but is easier to work with in code.

We're (currently) opting for the fluent-style due to the frequency of random usage of the jobs, but I'm definitely open to a 3rd option that moves us away from passing in a pile of arguments into a single method/constructor, but also is good at signaling its requirements/usage to those not immediately familiar with it.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
Ceiu: Yeah, I saw this in the review and thought what @Januson stated -- with a default value, it should be fine as a primitive. Admitted, I didn't consider trying to pass in null, but I don't think it's possible to do so via REST. I believe any value will be considered a string and converted accordingly.

PR #2345, opened 2019-06-03
[F] ENT-846: Port RefreshPoolsJob To The New Messaging Framework

nikosmoum:
* Now the RefreshPoolsJob is using the new artemis messaging
  framework.
* Updated OwnerResource.refreshPools() &
  ProductResource.refreshPoolsForProduct() to use the new job.
* Exposed the 'statusPath' field on AsyncJobStatusDTO.
* Added new 'getStatusAndDeleteIfFinished' endpoint to the temp
  async job resource.
* Updated the spec test api to expect new job states.
* Removed redundant field from Refresher.

### DIFF
--- a/server/src/main/java/org/candlepin/guice/CandlepinModule.java
+++ b/server/src/main/java/org/candlepin/guice/CandlepinModule.java
@@ -33,6 +33,7 @@ import org.candlepin.async.impl.ArtemisJobMessageDispatcher;
 import org.candlepin.async.tasks.ExportJob;
 import org.candlepin.async.tasks.ImportJob;
 import org.candlepin.async.tasks.RefreshPoolsForProductJob;
+import org.candlepin.async.tasks.RefreshPoolsJob;
 import org.candlepin.async.temp.AsyncJobResource;
 import org.candlepin.async.temp.TestJob1;
 import org.candlepin.audit.AMQPBusPublisher;
@@ -100,7 +101,6 @@ import org.candlepin.pinsetter.tasks.CertificateRevocationListTask;
 import org.candlepin.pinsetter.tasks.EntitlerJob;
 import org.candlepin.pinsetter.tasks.HypervisorUpdateJob;
 import org.candlepin.pinsetter.tasks.JobCleaner;
-import org.candlepin.pinsetter.tasks.RefreshPoolsJob;
 import org.candlepin.pinsetter.tasks.SweepBarJob;
 import org.candlepin.pinsetter.tasks.UnpauseJob;
 import org.candlepin.pki.impl.JSSPrivateKeyReader;
@@ -317,7 +317,6 @@ public class CandlepinModule extends AbstractModule {
         miscConfigurations();
 
         // Async Jobs
-        bind(RefreshPoolsJob.class);
         bind(EntitlerJob.class);
         requestStaticInjection(EntitlerJob.class);
         bind(HypervisorUpdateJob.class);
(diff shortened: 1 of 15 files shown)


server/src/main/java/org/candlepin/guice/CandlepinModule.java
```diff
@@ -418,6 +418,8 @@ private void configureAsyncJobs() {
         JobManager.registerJob(TestJob1.JOB_KEY, TestJob1.class);
         JobManager.registerJob(ExportJob.JOB_KEY, ExportJob.class);
         JobManager.registerJob(ImportJob.JOB_KEY, ImportJob.class);
+        JobManager.registerJob(org.candlepin.async.tasks.RefreshPoolsJob.JOB_KEY,
```
Ceiu: Should we import the class like we did for the others?

server/src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -289,26 +296,22 @@ public void deleteProduct(
...
-        //             return RefreshPoolsJob.forOwner(owner, lazy);
-        //         }
-        //     }
-        // );
+        List<AsyncJobStatus> statuses = new LinkedList<>();
+        for (JobConfig config : jobConfigs) {
+            statuses.add(this.jobManager.queueJob(config));
+        }
+
+        return statuses.stream().map(
```
Ceiu: random musings:
I wonder how we should handle the situation where one or more jobs are queued successfully and one or more jobs fail queuing. I don't think the pinsetter code handled this case, either, and it hasn't been a problem [yet/that we know of], but this seems like an awkward problem waiting to happen.

server/src/main/java/org/candlepin/guice/CandlepinModule.java
nikosmoum: The reason I used the full name is that the pinsetter RefreshPoolsJob is also imported in this file for a guice `bind()`. I suppose I can completely remove that since we no longer need it, and import the new class.

server/src/main/java/org/candlepin/resource/ProductResource.java
nikosmoum: I also thought about it but couldn't think of what would be a good way. The old job framework would call `scheduleJob()` in a loop for each `JobDetail` in the array, which simply throws a 503 error (so it didn't handle this case at all):
```
    private JobStatus scheduleJob(JobDetail detail) {
        try {
            return this.pinsetterKernel.scheduleSingleJob(detail);
        }
        catch (PinsetterException e) {
            throw new ServiceUnavailableException("Error scheduling refresh job.", e);
        }
    }
```

server/src/main/java/org/candlepin/guice/CandlepinModule.java
nikosmoum: Fixed.

server/src/main/java/org/candlepin/resource/ProductResource.java
nikosmoum: @Ceiu I've worked on handling this case, but I want to run it by you before pushing: http://pastebin.test.redhat.com/769504

Another approach to this would be to add a `Collection<AsyncJobStatus> queueJobs(Collection<JobConfig>)`  method to `JobManager`, which handles the exceptions of each queueing itself and returns a collection of statuses so that the caller doesn't have to do that work each time. (which sounds better than my initial approach).

Ceiu: It's not something critical to handle here, as the queueJob method handles the majority of actual issues we're concerned with. As far as valiadation failures go, I'm okay with the pastebin example except for the FAILED_VALIDATION state, as that's a reason for a failure state, not a state itself.

Anyway, splitting of hairs. If you want to add the extra exception handling and output of a generated AsyncJobStatus in a dead state, that's fine. But, again, no new state please. ;)

Ceiu:
ack

This is okay to merge now, or we can wait until an additional enhancement is added. I'm fine with either.

server/src/main/java/org/candlepin/resource/ProductResource.java
nikosmoum: I'll add the example from the pastebin (without new state), since we probably don't have many examples like this one to warrant a new method in JobManager.

nikosmoum:
@Ceiu Pushed the change from pastebin and fixed the conflict.

PR #2346, opened 2019-06-04
1673662: Better reason messages for Mismatched syspurpose status; ENT-1247

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1673662
* John requested some messages and it's IMHO reasonable
* PR created for subman (for testing):
  https://github.com/candlepin/subscription-manager/pull/2106


nikosmoum:
It also looks like there are some spec tests in `system_purpose_compliance_spec.rb` that asserted on those reason strings, which need to be changed now.

PR #2350, opened 2019-06-07
Remove Deprecated and unused annotations.

barnabycourt:
Candlepin now uses liquibase to generate the database schema. Because of this the ForeignKey and Index annotations are no longer necessary on the model objects. In addition the Hibernate.annotations.ForeignKey & Index annotations have been deprecated.

Rather than converting to the new JPA style annotations this change removes them entirely as they are not providing additional value on top of the liquibase schema generation.


Ceiu:
ack

PR #2351, opened 2019-06-07
Remove SELinux checks on gating-test docker image building

nikosmoum:
- The caller (jenkins node) will be responsible for turning
  SELinux enforcing off.


Lorquas:
These are built on isolated one off jenkins nodes, so I'm cool with it.

PR #2352, opened 2019-06-07
Update postgres dialect to minimum supported version of Postgres 9.2

barnabycourt:


barnabycourt:
Updated the commit for all the other places that referenced the dialect. Thanks for the catch in the build file, profiles, and cp-setup.

PR #2354, opened 2019-06-11
1502193: Periodic mysql exceptions when updating consumers in production

Januson:
- Add check to prevend updating hypervisor id to one already in use

### DIFF
--- a/server/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -1388,11 +1388,16 @@ public class ConsumerResource {
                 existing.setHypervisorId(null);
             }
             else {
+                Owner owner = ownerCurator.findOwnerById(existing.getOwnerId());
+                if (isUsedByOwner(incomingId, owner)) {
+                    throw new BadRequestException(i18n.tr(
+                        "Problem updating unit {0}. Hypervisor id: {1} is already used.",
+                        incoming, incomingId.getHypervisorId()));
+                }
                 if (existingId != null) {
                     if (existingId.getHypervisorId() != null &&
                         !existingId.getHypervisorId().equals(incomingId.getHypervisorId())) {
                         existingId.setHypervisorId(incomingId.getHypervisorId());
-                        Owner owner = ownerCurator.findOwnerById(existing.getOwnerId());
                         existingId.setOwner(owner);
                     }
                     else {
@@ -1401,7 +1406,6 @@ public class ConsumerResource {
                 }
                 else {
                     // Safer to build a new clean HypervisorId object
-                    Owner owner = ownerCurator.findOwnerById(existing.getOwnerId());
                     HypervisorId hypervisorId = new HypervisorId(incomingId.getHypervisorId());
                     hypervisorId.setOwner(owner);
                     existing.setHypervisorId(hypervisorId);
@@ -1410,6 +1414,13 @@ public class ConsumerResource {
             return true;
         }
         return false;
+    }
+
+    private boolean isUsedByOwner(HypervisorIdDTO hypervisor, Owner owner) {
+        if (hypervisor == null) {
+            return false;
+        }
+        return consumerCurator.getHypervisor(hypervisor.getHypervisorId(), owner) != null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1412,6 +1418,17 @@ private boolean checkForHypervisorIdUpdate(Consumer existing, ConsumerDTO incomi
         return false;
     }
 
+    private boolean isUsedByOwner(HypervisorIdDTO incomingId, Owner owner) {
+        if (incomingId == null) {
+            return false;
+        }
+        return owner.getConsumers().stream()
```
barnabycourt: What is the performance implication of this when an organization has 100k+ consumers?

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1402,6 +1403,11 @@ private boolean checkForHypervisorIdUpdate(Consumer existing, ConsumerDTO incomi
                 else {
                     // Safer to build a new clean HypervisorId object
                     Owner owner = ownerCurator.findOwnerById(existing.getOwnerId());
+                    if (isUsedByOwner(incomingId, owner)) {
```
barnabycourt: Can you help me understand under what condition this error is encountered?

server/src/main/java/org/candlepin/resource/ConsumerResource.java
barnabycourt: And why would you want to stream over all consumers vs getting a particular consumer with a more specific query?

server/src/main/java/org/candlepin/resource/ConsumerResource.java
Januson: There are a few cases where consumer update/creation precondition is not validated before hitting the database which is causing the MySQL error. In this case consumer should not be able to update its hypervisor id to one already in use by owner. You can check the error with the following spec test: "should not allow a consumer to update their hypervisorId to one in use by owner".

server/src/main/java/org/candlepin/resource/ConsumerResource.java
Januson: You are right. I didn't realize how many consumers there can be. I updated the query.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1412,6 +1418,16 @@ private boolean checkForHypervisorIdUpdate(Consumer existing, ConsumerDTO incomi
...
     }
 
+    private boolean isUsedByOwner(HypervisorIdDTO hypervisor, Owner owner) {
+        if (hypervisor == null) {
+            return false;
+        }
+        return consumerCurator.getHypervisorsBulk(
+            Collections.singletonList(hypervisor.getHypervisorId()), owner.getOwnerId())
+            .list()
+            .size() > 0;
```
nikosmoum: There already is a curator interface for querying with a single hypervisorId, without having to make a singleton list. The following should work:  `return this.consumerCurator.getHypervisor(hypervisor.getHypervisorId(), owner.getOwnerId()) != null`.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
barnabycourt: Ok, so to rephrase. We don't want to create two consumers in a single organization that represent the same hypervisor. In this particular case:
1) Two hypervisors consumers are created with different IDs
2) A change happens that causes the hypervisor IDs to converge.
3) Candlepin fails because it now has two records for what is effectively the same hypervisor. 

If the above is true, should we be merging the consumer records?

server/src/main/java/org/candlepin/resource/ConsumerResource.java

server/src/main/java/org/candlepin/resource/ConsumerResource.java
Januson: Does it make sense to merge two hypervisors? ID conflict could be accidental.

barnabycourt: @kahowell ^

barnabycourt: I'm ok with this as is for now.

kahowell: > Does it make sense to merge two hypervisors? ID conflict could be accidental.

I think we need to design some process for merging them with a script, or provide customer guidance on how to handle failing requests when they've gotten into this state. I agree with throwing the bad request, as this is a state we are trying to avoid generally.

PR #2356, opened 2019-06-12
[M] 1717903: syspurpose attributes now part of certificates [ENT-1369]

nikosmoum:
- syspurpose attributes are now part of the entitlement certificates.
- Bumped certificate version from 3.3 to 3.4.

### DIFF
--- a/server/src/test/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapterTest.java
+++ b/server/src/test/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapterTest.java
@@ -853,8 +853,9 @@ public class DefaultEntitlementCertServiceAdapterTest {
     public void ensureV3CertificateCreationOkWhenConsumerSupportsV3Dot1Certs()
         throws Exception {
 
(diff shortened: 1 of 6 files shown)


nikosmoum:
@jirihnidek PR is ready for testing

server/src/test/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapterTest.java
```diff
@@ -853,8 +853,9 @@ protected DefaultEntitlementCertServiceAdapter initCertServiceAdapter() {
     public void ensureV3CertificateCreationOkWhenConsumerSupportsV3Dot1Certs()
         throws Exception {
 
-        consumer.setFact("system.certificate_version", "3.3");
+        consumer.setFact("system.certificate_version", "3.4");
```
Januson: Certificate version is a good candidate for extraction. It is used at a lot of places so it will make a bit it easier to update the version in the future.

nikosmoum: Done.

PR #2359, opened 2019-06-18
[F] ENT-853: Port EntitleByProductsJob To The New Messaging Framework

Januson:


server/src/main/java/org/candlepin/model/AsyncJobStatusCurator.java
```diff
@@ -86,4 +89,30 @@ public AsyncJobStatusCurator() {
...
+        String hql = "update AsyncJobStatus j " +
+            "set j.state = :canceled " +
+            "where j.id = :jobid";
+        int updated = this.currentSession()
+            .createQuery(hql)
+            .setParameter("jobid", jobId)
+            .setInteger("canceled", JobState.CANCELED.ordinal())
+            .executeUpdate();
+        if (updated == 0) {
+            throw new NotFoundException("job not found");
```
Ceiu: NotFoundException is for REST, not to be thrown from the curators.

server/src/main/java/org/candlepin/model/AsyncJobStatusCurator.java
```diff
@@ -86,4 +89,30 @@ public AsyncJobStatusCurator() {
 
         return this.getJobsInState(states);
     }
+
+    @Transactional
+    public AsyncJobStatus cancel(String jobId) {
+        this.cancelNoReturn(jobId);
```
Ceiu: Given that we're actually returning the entity here, it probably makes sense to not hit the DB twice for this. Better off fetching, updating the status, persisting and then returning.

server/src/main/java/org/candlepin/auth/permissions/AsyncJobStatusPermission.java
```diff
@@ -59,6 +62,10 @@ public Criterion getCriteriaRestrictions(Class entityClass) {
 
     @Override
     public boolean canAccessTarget(AsyncJobStatus target, SubResource subResource, Access required) {
+        if (isOwnerInvalid(target)) {
```
Ceiu: We are (slowly) moving toward not doing this type of check in the permissions and, instead, checking before the job is created. Once the job is created, it's being assumed to have cleared all of these sorts of checks.

If this is to facilitate some kind of targeting for cancelling/fetching status, we should look at more robust solutions, and/or checks to do in the appropriate resource(s).

server/src/main/java/org/candlepin/async/temp/AsyncJobResource.java
```diff
@@ -140,4 +145,20 @@ public AsyncJobStatusDTO getStatusAndDeleteIfFinished(
...
     }
 
+    @ApiOperation(notes = "Cancels a Job Status", value = "cancel")
+    @ApiResponses({ @ApiResponse(code = 400, message = ""), @ApiResponse(code = 404, message = "") })
+    @DELETE
+    @Path("/{job_id}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public AsyncJobStatusDTO cancel(@PathParam("job_id") @Verify(AsyncJobStatus.class) String jobId) {
+        AsyncJobStatus status = this.jobCurator.get(jobId);
+        if (status.getState() == AsyncJobStatus.JobState.CANCELED) {
```
Ceiu: There's an argument to be made that this check should be performed at the curator level, but disregarding that, we also don't want to clobber other terminal states. This can be paired down to something like:

```
if (status.getState().isTerminal()) {
  // throw exception
}
```

We also need to check that the status actually exists.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
```diff
@@ -0,0 +1,182 @@
...
+                .setJobName(JOB_NAME)
+                .addConstraint(JobConstraints.uniqueByArgument(OWNER_KEY));
+        }
+
+        public EntitleByProductsJobConfig setOwner(final Owner owner) {
+            if (owner == null) {
+                throw new IllegalArgumentException("Owner is null");
+            }
+
+            this.setJobArgument(OWNER_KEY, owner.getKey())
```
Ceiu: Owner metadata is not being set here

server/src/main/java/org/candlepin/model/AsyncJobStatusCurator.java
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
Januson: Done.

server/src/main/java/org/candlepin/model/AsyncJobStatusCurator.java
Januson: Done.

server/src/main/java/org/candlepin/async/temp/AsyncJobResource.java
Januson: Done.

server/src/main/java/org/candlepin/auth/permissions/AsyncJobStatusPermission.java
Januson: You are right. I added the check for cancelling/fetching as according to the spec tests it was expected behavior. I created methods get(jobId) and cancel(jobId) in JobManager and moved the checks there.

Ceiu: For now let's just drop the check entirely. This is something that needs to be fleshed out more, and I really don't want jobs to be cancellable or such based on the arguments or metadata.

server/src/main/java/org/candlepin/model/AsyncJobStatusCurator.java
Ceiu: This is still doing a bit of unnecessary work, and introduces a bit of a race condition depending on the caching mode and innerworkings of Hibernate. It's fine for this specific PR, but will have to be completely reworked in the task that addresses the current DB vs messaging issue.

server/src/test/java/org/candlepin/async/tasks/EntitleByProductsJobTest.java
```diff
@@ -0,0 +1,173 @@
...
+        final List<String> fromPools = Collections.singletonList("pool_id1");
+
+        final JobConfig config = EntitleByProductsJob.createConfig()
+            .setOwner(owner)
+            .setConsumer(consumer)
+            .setProductIds(pids)
+            .setEntitleDate(new Date())
+            .setFromPools(fromPools);
+
+        assertDoesNotThrow(config::validate);
```
Ceiu: These tests are kind of okay, but testing for correctness based on the absence (or non-absence) of a value isn't complete. Also, this particular implementation relies on validate being correct, which should be tested on its own.

A better way to approach this would be to test that each method is setting the appropriate fields in isolation.

server/src/main/java/org/candlepin/auth/permissions/AsyncJobStatusPermission.java
Januson: Ok. I dropped the check and added skip to its spec test.

server/src/test/java/org/candlepin/async/tasks/EntitleByProductsJobTest.java
Januson: I am not testing those setters directly. In these tests I am using them to test validation. Whether values are retrievable is tested below when job tries to access those values.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1126,4 +1137,50 @@ public void modeChanged(Mode mode) {
...
+     * @param jobId
+     *  Id of the requested job
+     *
+     * @return
+     *  {@link AsyncJobStatus} of the requested job
+     *
+     * @throws JobAccessException.NotFound
+     *  When requested job was not found
+     */
+    public AsyncJobStatus get(final String jobId) throws JobAccessException {
```
Ceiu: The naming convention to follow is `<verb><noun><conjunction><modifiers>`. Keeping in step with the other methods, these should be getJob and cancelJob.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1126,4 +1137,50 @@ public void modeChanged(Mode mode) {
...
+     * @throws JobAccessException.NotFound
+     *  When requested job was not found
+     *
+     * @throws JobAccessException.InvalidState
+     *  When cancelled job is already terminated (CANCELLED, COMPLETED, ...)
+     */
+    public AsyncJobStatus cancel(final String jobId) throws JobAccessException {
+        final AsyncJobStatus status = this.get(jobId);
+        if (status.getState().isTerminal()) {
+            throw new JobAccessException.InvalidState(jobId, status.getState());
```
Ceiu: We can avoid a custom exception by just throwing an IllegalStateException instead.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1126,4 +1137,50 @@ public void modeChanged(Mode mode) {
...
+        }
+        status.setState(JobState.CANCELED);
+        this.jobCurator.saveAll(Collections.singletonList(status), false, false);
```
Ceiu: jobCurator.merge works here without needing to wrap the object in a list.

server/spec/job_status_spec.rb
```diff
@@ -161,64 +160,77 @@
...
+    system = consumer_client(@user, 'system7', :system,  nil,  {}, @owner['key'])
     job = system.consume_product(@monitoring.id, { :async => true })
-    status = system.get_job(job['id'])
-    system.cancel_job(job['id'])
+    status = system.get_async_job(job['id'])
+    system.cancel_async_job(job['id'])
     # wait for job to complete, or test clean up will conflict with the asynchronous job.
-    wait_for_job(status['id'], 15)
+    wait_for_async_job(status['id'], 15)
+    @cp.set_async_scheduler_status(true)
```
Ceiu: This is dangerous. If this test doesn't complete for any reason, the scheduler will be left in the paused state and will impact many other tests.

If we absolutely need this functionality for the test, we should do something similar to how created entities are cleaned up automatically at the end of every test to ensure the scheduler is turned back on at the end of a test, even if it fails.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1126,4 +1137,50 @@ public void modeChanged(Mode mode) {
...
+     *
+     * @throws JobAccessException.InvalidState
+     *  When cancelled job is already terminated (CANCELLED, COMPLETED, ...)
+     */
+    public AsyncJobStatus cancel(final String jobId) throws JobAccessException {
+        final AsyncJobStatus status = this.get(jobId);
+        if (status.getState().isTerminal()) {
+            throw new JobAccessException.InvalidState(jobId, status.getState());
+        }
+        status.setState(JobState.CANCELED);
```
Ceiu: This will work for this PR, but there's more work to be done here. There are a number of cases here which won't work as the operation implies, and those may or may not require attention. At the very least, they should be documented in the javadocs so we minimize confusion in the future.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
```diff
@@ -0,0 +1,183 @@
...
+            return this;
+        }
+
+        public EntitleByProductsJobConfig setEntitleDate(final Date entitleDate) {
+            this.setJobArgument(ENTITLE_DATE_KEY, entitleDate);
+
+            return this;
+        }
+
+        public EntitleByProductsJobConfig setFromPools(final Collection<String> fromPools) {
```
Ceiu: Careful on the method name here, this is just verbiage from the original author's variable names being carried forward. This should just be `setPools`.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1126,4 +1137,50 @@ public void modeChanged(Mode mode) {
...
+     * @return
+     *  {@link AsyncJobStatus} of the requested job
+     *
+     * @throws JobAccessException.NotFound
+     *  When requested job was not found
+     */
+    public AsyncJobStatus get(final String jobId) throws JobAccessException {
+        final AsyncJobStatus status = this.jobCurator.get(jobId);
+        if (status == null) {
+            throw new JobAccessException.NotFound(jobId);
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/JobManager.java
Januson: Done.

server/src/main/java/org/candlepin/async/JobManager.java
Januson: Done.

server/src/main/java/org/candlepin/async/JobManager.java
Januson: Done.

server/spec/job_status_spec.rb
Januson: Done. I added after(:each) to clean it up.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
Januson: Done.

server/src/main/java/org/candlepin/async/JobManager.java
Januson: What cases do you mean?

Ceiu: Off the top of my head:
- Job is currently executing, potentially on another node
- Created -> cancelled during message sending or failure will lead to created -> failed at the same time
- General race condition where another node (or even the current node) has started a transaction which will process the job message while this node is starting to cancel it (queued, failed_with_retry -> executing)

server/src/main/java/org/candlepin/async/temp/AsyncJobResource.java
```diff
@@ -172,4 +183,50 @@ public AsyncJobStatusDTO getStatusAndDeleteIfFinished(
...
     }
 
+    @ApiOperation(notes = "Cancels a Job Status", value = "cancel")
+    @ApiResponses({
+        @ApiResponse(code = 400, message = ""),
+        @ApiResponse(code = 404, message = "") })
+    @DELETE
+    @Path("/{job_id}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public AsyncJobStatusDTO cancel(@PathParam("job_id") @Verify(AsyncJobStatus.class) String jobId) {
```
nikosmoum: Let's remove this method for now.

server/src/main/java/org/candlepin/async/temp/AsyncJobResource.java
```diff
@@ -172,4 +183,50 @@ public AsyncJobStatusDTO getStatusAndDeleteIfFinished(
...
+                String.format("Could not cancel job with id: %s. Job was not found!", jobId));
+        }
+        return this.translator.translate(cancelled, AsyncJobStatusDTO.class);
+    }
+
+    @ApiOperation(notes = "Retrieves the Scheduler Status", value = "getSchedulerStatus")
+    @GET
+    @Path("/scheduler")
+    @Produces(MediaType.APPLICATION_JSON)
+    public SchedulerStatusDTO getSchedulerStatus() {
```
nikosmoum: Let's remove this method for now.

server/src/main/java/org/candlepin/async/temp/AsyncJobResource.java
```diff
@@ -172,4 +183,50 @@ public AsyncJobStatusDTO getStatusAndDeleteIfFinished(
...
+        return new SchedulerStatusDTO(jobManager.status());
+    }
+
+    @ApiOperation(notes = "Updates the Scheduler Status", value = "setSchedulerStatus")
+    @ApiResponses({ @ApiResponse(code = 500, message = "") })
+    @POST
+    @Path("/scheduler")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    public SchedulerStatusDTO setSchedulerStatus(boolean running) {
```
nikosmoum: Let's remove this method for now.

server/client/ruby/candlepin_api.rb
```diff
@@ -480,6 +480,10 @@ def set_scheduler_status(status)
     post("/jobs/scheduler", {}, status)
   end
 
+  def set_async_scheduler_status(status)
```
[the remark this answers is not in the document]
Januson: Done.

server/client/ruby/candlepin_api.rb
```diff
@@ -1375,6 +1379,10 @@ def cancel_job(job_id)
     delete "/jobs/#{job_id}"
   end
 
+  def cancel_async_job(job_id)
```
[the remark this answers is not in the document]
Januson: Done.

server/spec/job_status_spec.rb
```diff
@@ -13,6 +13,10 @@
     create_pool_and_subscription(@owner['key'], @monitoring.id, 4)
   end
 
+  after(:each) do
+    @cp.set_async_scheduler_status(true)
```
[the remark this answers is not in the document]
Januson: Done.

server/spec/job_status_spec.rb
```diff
@@ -161,64 +164,77 @@
   it 'should not allow user to cancel a job it did not initiate' do
     system = consumer_client(@user, 'system7')
     job = system.consume_product(@monitoring.id, { :async => true })
-    status = system.get_job(job['id'])
+    status = system.get_async_job(job['id'])
```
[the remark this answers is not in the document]
Januson: Done.

server/spec/job_status_spec.rb
```diff
@@ -161,64 +164,77 @@
...
     lambda do
-      @user.cancel_job(job['id'])
+      @user.cancel_async_job(job['id'])
     end.should raise_exception(RestClient::Forbidden)
   end
 
-  it 'should not allow user to view job status outside of managed org' do
+  # TODO Reenable after we decide how to check for access from outside of managed org
+  xit 'should not allow user to view job status outside of managed org' do
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1226,4 +1236,46 @@ public void modeChanged(Mode mode) {
...
+    /**
+     * Lookup job status by job id.
+     *
+     * @param jobId
+     *  Id of the requested job
+     *
+     * @return
+     *  {@link AsyncJobStatus} of the requested job or null
+     */
+    public AsyncJobStatus getJob(final String jobId) {
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1226,4 +1236,46 @@ public void modeChanged(Mode mode) {
...
+     *  Id of the cancelled job or null if cancelled job was not found
+     *
+     * @return
+     *  {@link AsyncJobStatus} of the cancelled job
+     *
+     * @throws IllegalStateException
+     *  When cancelled job is already terminated (CANCELLED, COMPLETED, ...)
+     */
+    @Transactional
+    public AsyncJobStatus cancelJob(final String jobId) {
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/async/temp/AsyncJobResource.java
Januson: Done.

server/src/main/java/org/candlepin/async/temp/AsyncJobResource.java
Januson: Done.

server/src/main/java/org/candlepin/async/temp/AsyncJobResource.java
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * Job configuration object for the entitle by products job
+     */
+    public static class EntitleByProductsJobConfig extends JobConfig {
+
+        public EntitleByProductsJobConfig() {
+            this.setJobKey(JOB_KEY)
+                .setJobName(JOB_NAME)
+                .addConstraint(JobConstraints.uniqueByArguments(OWNER_KEY));
```
nikosmoum: The original job does not seem to extend `UniqueByEntityJob`, or implement the `isSchedulable/scheduleJob` methods, so there is no need for constraints on this job.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        public EntitleByProductsJobConfig setOwner(final Owner owner) {
+            if (owner == null) {
+                throw new IllegalArgumentException("Owner is null");
+            }
+
+            this.setJobMetadata(OWNER_KEY, owner.getKey())
+                .setJobArgument(OWNER_KEY, owner.getKey())
```
nikosmoum: This job doesn't actually use the owner key in it's execute method, so it shouldn't be set as argument; we only need to set as metadata and log level.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        @Override
+        public void validate() throws JobConfigValidationException {
+            super.validate();
+
+            try {
+                final JobArguments arguments = this.getJobArguments();
+
+                final String ownerKey = arguments.getAsString(OWNER_KEY);
```
nikosmoum: The owner key is not an argument of this job's execution, so we don't need to validate its existence here.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/EntitleByProductsJob.java
Januson: Done.

server/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java
```diff
@@ -110,6 +112,27 @@ public static RefreshPoolsForProductJobConfig createJobConfig() {
...
+         * Sets the owner for this export job. The owner is not required, but provides the org
+         * context in which the job will be executed.
+         *
+         * @param owner
+         *  the owner to set for this job
+         *
+         * @return
+         *  a reference to this job config
+         */
+        public RefreshPoolsForProductJobConfig setOwner(Owner owner) {
```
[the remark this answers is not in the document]
Januson: Done.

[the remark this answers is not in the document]
Ceiu: Not sure if github has goofed or not, but the changes to this job are still showing up in the diff and locally no matter how many times I refresh/reset.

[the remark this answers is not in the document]
nikosmoum: I believe Ondrej accidentally removed the setOwner method from EntitleByProductsJobConfig, instead of the one from RefreshPoolsForProductJobConfig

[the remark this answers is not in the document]
Ceiu: Hah, oops.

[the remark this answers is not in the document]
Januson: Oh. Sorry. Now it should be fixed.

PR #2360, opened 2019-06-19
[F] Ported the UndoImportsJob to the Artemis job system (ENT-854)

Ceiu:
- Ported the UndoImportsJob to the Artemis job system
- Added a new owner-specific lockAndLoad variant that can
  lock and load an owner entity by key
- Added a constant to LoggingFilter to specify the key
  used in the packaged logback.xml configuration for
  referencing the org/owner key
- Updated all async tests to use the new
  LoggingFilter.OWNER_KEY constant when setting the
  owner key in the job metadata

### DIFF
--- a/server/src/main/java/org/candlepin/async/tasks/UndoImportsJob.java
+++ b/server/src/main/java/org/candlepin/async/tasks/UndoImportsJob.java
@@ -1,0 +1,227 @@
+/**
+ * Copyright (c) 2009 - 2012 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.async.tasks;
+
+import org.candlepin.async.ArgumentConversionException;
+import org.candlepin.async.AsyncJob;
+import org.candlepin.async.JobArguments;
+import org.candlepin.async.JobConfig;
+import org.candlepin.async.JobConfigValidationException;
+import org.candlepin.async.JobExecutionContext;
+import org.candlepin.async.JobExecutionException;
+import org.candlepin.async.JobConstraints;
+import org.candlepin.common.filter.LoggingFilter;
+import org.candlepin.controller.PoolManager;
+import org.candlepin.model.ExporterMetadata;
+import org.candlepin.model.ExporterMetadataCurator;
+import org.candlepin.model.ImportRecord;
+import org.candlepin.model.ImportRecordCurator;
+import org.candlepin.model.ImportUpstreamConsumer;
+import org.candlepin.model.Owner;
+import org.candlepin.model.OwnerCurator;
+import org.candlepin.model.Pool;
+import org.candlepin.model.UpstreamConsumer;
+import org.candlepin.service.SubscriptionServiceAdapter;
+
+import com.google.inject.persist.Transactional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.xnap.commons.i18n.I18n;
+
+import javax.inject.Inject;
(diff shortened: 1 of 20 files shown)


server/src/main/java/org/candlepin/pinsetter/tasks/UndoImportsJob.java
```diff
@@ -96,13 +96,15 @@ public void toExecute(JobExecutionContext context) throws JobExecutionException
         Principal principal = (Principal) map.get(PinsetterJobListener.PRINCIPAL_KEY);
 
         // TODO: Should we check the principal again here?
+        log.info("GRABBED OWNER: {}", owner);
```
nikosmoum: Leftover debug statement?

server/src/test/java/org/candlepin/async/tasks/UndoImportsJobTest.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.Locale;
+
+
+
+/**
+ * UndoImportsJobTest
+ */
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+public class UndoImportsJobTest extends DatabaseTestFixture {
```
nikosmoum: I noticed that the old `UndoImportsJobTest` has some re-fire scenario test cases. Those were basically testing the scenarios that the job itself was handling with catch clauses in the past, but then that logic then moved to the general KingpinJob re-fire logic. Do we need to bring that KingpinJob logic in the new framework (e.g. making sure we re-try when we find lock exceptions or runtime exceptions that wrap SQLExceptions), since it seems it is desirable?

common/pom.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<!--  vim: set expandtab sts=2 sw=2 ai:  -->
+<!-- **This file is auto-generated by Buildr.  Changes may be lost.** -->
```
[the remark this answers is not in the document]
Ceiu: Maybe leftover artifacts from the buildr era that my machine didn't clear? I can drop this from the PR and we can sort it out later.

server/src/main/java/org/candlepin/async/tasks/UndoImportsJob.java
```diff
@@ -0,0 +1,229 @@
...
+    @Override
+    @Transactional
+    public Object execute(JobExecutionContext context) throws JobExecutionException {
+        JobArguments args = context.getJobArguments();
+        String principalName = context.getPrincipalName();
+
+        String ownerKey = args.getAsString(OWNER_KEY);
+        Owner owner = this.ownerCurator.lockAndLoadByKey(ownerKey);
+
+        log.info("GRABBED OWNER: {}", owner);
```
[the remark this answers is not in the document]
Ceiu: Crap. Yep.

server/src/test/java/org/candlepin/async/tasks/UndoImportsJobTest.java
Ceiu: In the new framework, all jobs will refire on unexpected/uncaptured exceptions as the default behavior. We only need special configuration if we *don't* want that behavior.

Ceiu: As a side note, this is something that we can add to the job config if we need to fiddle with the default behavior for jobs wrt how they handle uncaught exceptions.

common/pom.xml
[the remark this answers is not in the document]
Ceiu: After reading what @awood mentioned in another PR, looks like this got generated automatically when I ran buildr at one point. Nothing more we need to do here once I axe the file entirely.

server/src/main/java/org/candlepin/pinsetter/tasks/UndoImportsJob.java
Ceiu: fixed

server/src/main/java/org/candlepin/async/tasks/UndoImportsJob.java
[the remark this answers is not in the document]
Ceiu: fixed

Ceiu:
I'm willing to bet it's the same race condition the other job PR is hitting, but it appears we no longer keep the logs around after runs on Jenkins. I'll see if I can trigger it locally here to verify.

PR #2362, opened 2019-06-20
1715838: Handle matching of system uuid regardless of casing [ENT-1365]

wottop:


common/pom.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<!--  vim: set expandtab sts=2 sw=2 ai:  -->
+<!-- **This file is auto-generated by Buildr.  Changes may be lost.** -->
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```
awood: This is not right.  I'm guessing you used Buildr to run some tests or do a build or something and it regenerated the POM file completely.  You need to remove this change from the PR.

Ceiu:
ACK
Hopefully @awood didn't have any other hangups. Looks like the pom.xml issue was sorted and everything else makes sense to me.

PR #2363, opened 2019-06-20
[F] Ported the HypervisorHeartbeatUpdateJob to the Artemis job framework (ENT-1390)

Ceiu:


server/src/main/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJob.java
```diff
@@ -0,0 +1,165 @@
...
+ * Asynchronous job for refreshing the last check-in time for a specific reporter ID
+ */
+public class HypervisorHeartbeatUpdateJob implements AsyncJob {
+    private static final Logger log = LoggerFactory.getLogger(HypervisorHeartbeatUpdateJob.class);
+
+    public static final String JOB_KEY = "HYPERVISOR_HEARTBEAT_UPDATE";
+    public static final String JOB_NAME = "hypervisor_heartbeat_update";
+
+    private static final String REPORTER_ID = "reporter_id";
+    private static final String OWNER_KEY = "owner_key";
```
nikosmoum: Maybe this should be `org` instead of `owner_key`, in order to be consistent with the other jobs?
It might be better if we factor out into public flags all the common flags used by jobs like `OWNER_KEY` and `CONSUMER_ID`? It might help in a potential future case where we're using debug SQL with job arguments being standardized instead of some jobs using 'org' and others 'owner_key' as a key.

server/src/main/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJob.java
```diff
@@ -0,0 +1,165 @@
...
+     * Job configuration object for the export job
+     */
+    public static class HypervisorHeartbeatUpdateJobConfig extends JobConfig {
+        public HypervisorHeartbeatUpdateJobConfig() {
+            this.setJobKey(JOB_KEY)
+                .setJobName(JOB_NAME);
+
+            // TODO: Should this be unique by (owner_key, reporter_id) or owner_key? If it's the
+            // former, we'll need to update the UniqueByArgConstraint to support examining multiple
+            // parameters
```
nikosmoum: Probably unique by (owner_key, reporter_id), we can add a card about the multi parameter checking.

server/src/test/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJobTest.java
```diff
@@ -0,0 +1,164 @@
...
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+import java.util.Date;
+import java.util.Map;
+
+
+
+/**
+ * UndoImportsJobTest
```
nikosmoum: Wrong comment

server/src/main/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJob.java
Ceiu: We could use org, but I feel that's more inheritance of the old pinsetter stuff, where it was called "org" specifically because of the parameter name from logback.xml (or the other way around). In this particular case, I want the variable to be clear that it's the org key and not ID or anything else. This is why the UndoImportsJob task includes the LoggingFilter.OWNER_KEY field (which has a value of "org" due to the afore mentioned config requirement) expressly for metadata concerns; we're no longer required to couple the name of our owner variable to what is needed for logging. Since I expect this PR to go in second, I didn't fully expand that variable out like I did in the other PR.

For jobs in general, I want the jobs to be free to use the most expressive argument names deemed necessary by the author -- especially since with the addition of per-job config objects, a lot/all of it is fully encapsulated anyway.

server/src/main/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJob.java
Ceiu: Alright. I assume we'll leave this as the TODO for now until the other card is completed?

server/src/test/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJobTest.java
Ceiu: Oops. I totally didn't just copy/paste this from the other PR. :P

Ceiu:
I've posted the fixes for the issues caught here, but I ran into a bigger problem that @mstead and I were worried about: there's a race condition between when job messages are sent and when the original transaction/request completes and the DB is updated. I've noticed in spec tests this tends to happen a lot with this particular job, but the proper fix is going to require a bit more attention that goes well outside the scope of this PR. If Jenkins is unable to complete the spec test (failing on the date with "no such job" in the logs), you can temporarily work around this by manually testing with a delay added to the job message.

Ceiu:
Rebased to fix conflicts

Ceiu:
The db/msg order race condition commonly present in the spec test failures here is addressed by #2367. This PR may need to wait for that PR to be merged for all tests to complete.

server/src/main/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJob.java
nikosmoum: If you think it's not much work, you can add the multiparameter checking in this PR, otherwise leave the TODO and create a card, up to you.

server/src/test/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJobTest.java
nikosmoum: I think this got overwritten by the rebase :smile:

server/src/main/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJob.java
Ceiu: Yeah, I'd rather address the multi-param constraint in another card.

server/src/test/java/org/candlepin/async/tasks/HypervisorHeartbeatUpdateJobTest.java
Ceiu: Argh... yup.

nikosmoum:

PR #2365, opened 2019-06-24
[M] Consumer creation will now respect syspurpose attributes (ENT-1362)

Ceiu:
- Added support for syspurpose attributes on consumer creation


wottop:
ack

PR #2366, opened 2019-06-25
Improved deploy script to allow usage of symlinks by tomcat

jirihnidek:
* Tomcat doesn't allow to use symbolic links be default due to
  security reasons. It is necessary to allow it in configuration
  file META-INF/context.xml
* Improved a little bit deploy script


Ceiu:
ack

PR #2367, opened 2019-06-25
[F] Added synchronization between the database and messaging bus

Ceiu:
- Added a synchronization operation between the database and
  messaging bus to ensure job messages aren't committed until
  the active transaction which would persist the backing job
  is complete
- Added an additional check to ensure a job class is registered
  for a given job key

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -34,7 +34,12 @@ import org.candlepin.util.Util;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 import com.google.inject.Injector;
+import com.google.inject.persist.Transactional;
 import com.google.inject.persist.UnitOfWork;
+
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.resource.transaction.spi.TransactionStatus;
 
 import org.quartz.CronTrigger;
 import org.quartz.CronScheduleBuilder;
@@ -70,6 +75,8 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import javax.transaction.Synchronization;
+import javax.transaction.Status;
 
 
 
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -857,6 +920,16 @@ public synchronized AsyncJobStatus executeJob(final JobMessage message) throws J
             this.setupLogging(status);
             this.setupPrincipal(status);
 
+            // Maybe in this case it'd be better to attempt to use the job key as the job class
+            // rather than failing directly. This would allow use of aliases and explicit job
+            // classes.
```
nikosmoum: I am not sure how that would work; isn't the job key a string like `EXPORT_JOB`, and not a class? Also, how would using the job key would help here, since it would render pointless the registration mechanism(?) since when `getJobClass` returns a null class, it's because it was not registered.

Ceiu: This is related to the e-mail I sent over vacation. I've never been fully sold on the concept of registration and whatnot, and wouldn't mind seeing a system where the job key is either an alias that requires resolution (as it is today) or a job class directly for firing jobs without registering them directly.

At this point it's kind of in-code musings, since it's not high enough priority to really matter.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -813,11 +859,28 @@ private AsyncJobStatus postJobStatusMessage(AsyncJobStatus status)
...
             JobMessage message = new JobMessage(status.getId(), status.getJobKey());
-            this.dispatcher.sendJobMessage(message);
+            this.dispatcher.postJobMessage(message);
 
             // Update the job's status
             status = this.updateJobStatus(status, JobState.QUEUED, null);
 
+            // Create and register a synchronization to commit or rollback the dispatcher based on whether
+            // or not the current DB transaction completes
+            final JobMessageDispatcher jmd = this.dispatcher;
```
Ceiu: Oops. Forgot to remove this from the PoC

server/src/main/java/org/candlepin/async/JobManager.java
nikosmoum: Ah I see. Changing this whole key/registration is fine with me, but I guess not in the scope for this PR.

nikosmoum:
@Ceiu hmmmm, it looks like the test that keeps failing intermittently isn't fixed by this change ('refreshes pools for orgs owning products') :(

Ceiu:
After many attempts, I got it to reproduce locally. The issue in question is an odd situation where for some reason a given job isn't leaving the CREATED state. There don't appear to be any failures anywhere, it's just not updating, or the result is being cached somewhere. When this happens, scheduling the next refresh task fails, but the failure object doesn't have an ID. The wait_for_async_job command passes the null value through, then builds the URL without the job ID; leading to the 404 we see when the test fails.

I'll see if I can figure out what is triggering this, but I fear that it has to do with the jobCurator not having strong control over the transactions. Also, not having a reliable reproducer makes this hard to trigger and harder to verify if I can fix it.

nikosmoum:
The latest changes look good to me and the annoying test looks to succeed! Any more changes required in this PR? @Ceiu

Ceiu:
Nah, this should be it for the short term unless something else fails or breaks.

PR #2369, opened 2019-06-26
[2.6] 1724115: Initialize the NSS db when loading a JSS provider [ENT-1388]

nikosmoum:
- Now, when creating a JSS provider on deployment, we also create
  a JSS CryptoManager, which initializes the NSS db.
- Downgraded JSS to 4.4.6 to be in line with the jars/libraries
  distributed in RHEL 7. This means with this fix, candlepin is
  not compatible with RHEL 8, which ships with JSS 4.5.x.
- Added https://barnabycourt.fedorapeople.org/repo/candlepin as
  a new, temporary, artifact repository for our builds, which holds
  the jss 4.4.6 dependency for now.

Co-authored-by: Alexander Scheel <ascheel@redhat.com>

### DIFF
--- a/buildfile
+++ b/buildfile
@@ -1,6 +1,7 @@
 # vi: set ft=ruby:
 
 ### Repositories
+repositories.remote << "https://barnabycourt.fedorapeople.org/repo/candlepin"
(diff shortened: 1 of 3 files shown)


Ceiu:
ack

PR #2372, opened 2019-06-28
1713420: Load JSS jar from file system

Januson:
- Make JSS a compile only dependency
- Add spec comand to include locally installed JSS into shared classpath

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -118,7 +118,8 @@ dependencies {
     implementation "org.ehcache:jcache:1.0.0"
     implementation "javax.cache:cache-api:1.0.0"
 
-    implementation "org.mozilla:jss:4.4.6"
+    compileOnly "org.mozilla:jss:4.4.6"
+    testImplementation "org.mozilla:jss:4.4.6"
     implementation "ldapjdk:ldapjdk:4.19"
     implementation "org.quartz-scheduler:quartz:2.2.1"
 
--- a/server/src/main/webapp/META-INF/context.xml
+++ b/server/src/main/webapp/META-INF/context.xml
@@ -1,2 +1,6 @@
 <!-- Candlepin needs symlink support to read its jars -->
-<Context path="/candlepin" allowLinking="true" />
+<Context path="/candlepin" allowLinking="true">
+    <Loader className="org.apache.catalina.loader.VirtualWebappLoader"
+            virtualClasspath="${basedir}/target/classes;${basedir}/target/${project.build.finalName}/WEB-INF/lib;/usr/lib/java/jss4.jar"/>
+    <JarScanner scanAllDirectories="true"/>
+</Context>
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -30,6 +30,18 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>org.mozilla</groupId>
+      <artifactId>jss</artifactId>
+      <version>4.4.6</version>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <artifactId>*</artifactId>
+          <groupId>*</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
       <groupId>javax.servlet</groupId>
       <artifactId>servlet-api</artifactId>
       <version>2.5</version>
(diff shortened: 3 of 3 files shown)


Ceiu:

server/candlepin.spec.tmpl
```diff
@@ -139,6 +139,13 @@ touch %{buildroot}/%{_sysconfdir}/%{name}/%{name}.conf
 %{__install} -d -m 755 %{buildroot}/%{_sysconfdir}/tomcat/
 %{__unzip} %{SOURCE1} -d %{buildroot}/%{_sharedstatedir}/tomcat/webapps/%{name}
 
+## Add JSS to the classpath
+if grep -xq "^shared.loader=$" /etc/tomcat/catalina.properties ; then
+    sudo sed -i "/^shared.loader=$/ s/$/\/usr\/lib64\/jss\/jss4.jar/" /etc/tomcat/catalina.properties
```
nikosmoum: Unfortunately, this isn't going to work. I didn't do a good job explaining that the %install part in the spec file isn't actually run during the installation of the rpm, but when we build the rpm. Any commands to install files here, should be copying/updating files in the virtual build root directory structure. Editing the /etc/tomcat/catalina.properties file here is going to edit it in the machine we're building the rpm, not in the machine that the rpm is going to be installed. This is why other similar commands reference %{buildroot}, etc.

awood: Even if you dropped this in a `%post` section in the spec file, I don't think we want to go editing `catalina.properties` since that file belongs to the tomcat RPM and modifying a file owned by something else would be a no-no.

Instead, what if we dropped a file into `/etc/tomcat/conf.d` that looked something like this

```
CATALINA_OPTS="$CATALINA_OPTS -Dshared.loader=/usr/lib64/jss/jss4.jar"
```

This may not be perfect because I think this method would override any shared.loader defined in `catalina.properties`.  Also the `/etc/tomcat/conf.d` directory only existed as of four or five Fedora distributions ago, so that may not be an option in older RHEL 7 releases.  We'd need to check that the directory exists on the RHEL 7 versions we support (it is there on the RHEL 7.7 box I have handy).

server/src/main/root/META-INF/context.xml
```diff
@@ -1,2 +1,6 @@
 <!-- Testing repositories have to contain symlinks, when golden ticket is used -->
-<Context override="true" allowLinking="true" />
+<Context override="true" allowLinking="true">
```
awood: Quick question about the file paths here.  I see we have it under `src/main/root` and `src/main/webapp`.  I've never seen `src/main/root` used before.  Would you mind elaborating a little on what this path does?

server/src/main/webapp/META-INF/context.xml
```diff
@@ -1,2 +1,6 @@
 <!-- Candlepin needs symlink support to read its jars -->
-<Context path="/candlepin" allowLinking="true" />
+<Context path="/candlepin" allowLinking="true">
+    <Loader className="org.apache.catalina.loader.VirtualWebappLoader"
+            virtualClasspath="${basedir}/target/classes;${basedir}/target/${project.build.finalName}/WEB-INF/lib;/usr/lib64/jss/jss4.jar"/>
+    <JarScanner scanAllDirectories="true"/>
+</Context>
```
awood: Please add a newline at the end of the file here.

server/src/main/root/META-INF/context.xml
Januson: I was not sure if this change is needed in both context files so I did a bit more testing. I seems `src/main/root` is not used for Candlepin but for setup of testing repository. Only change in webapp should be needed.

awood: Okay.  If you want to back this file out then, that's fine.

server/src/main/webapp/META-INF/context.xml
awood: Also, I have a small concern about `/usr/lib64/`.  In some systems (non-64 bit systems) the path is often `/usr/lib`.  If we are only supporting x86_64, no problem, but we need to check if that's true.  If we're supporting other arches, we'll need to figure something out.

server/build.gradle
```diff
@@ -118,7 +118,8 @@ dependencies {
     implementation "org.ehcache:jcache:1.0.0"
     implementation "javax.cache:cache-api:1.0.0"
 
-    implementation "org.mozilla:jss:4.4.6"
+    compileOnly "org.mozilla:jss:4.4.6"
+    testCompile "org.mozilla:jss:4.4.6"
```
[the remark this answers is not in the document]
Januson: This should work now. I changed the classpath modification to context.xml file which is packaged in .war.

server/candlepin.spec.tmpl
Januson: Removed this in favor of Tomcat's context.xml approach.

server/src/main/root/META-INF/context.xml
Januson: Reverted changes to this file.

server/src/main/webapp/META-INF/context.xml
Januson: Added newline. Changed path to /usr/lib/java/jss4.jar which seems consistent on RHEL 6-8 for both 64 and 32-bit versions.

PR #2375, opened 2019-07-02
[F] ENT-855: Port RegenEnvEntitlementCertsJob to the new job framework

nikosmoum:


Ceiu:
Looks like there are some spec tests to fix yet, but other than that, looks good. Some good maintenance changes as well.

nikosmoum:
@Ceiu tests fixed

PR #2378, opened 2019-07-03
[F] Ported the ExpiredPoolsJob to the Artemis job system (ENT-861)

Ceiu:
- Ported the ExpiredPoolsJob to the Artemis job system as the
  ExpiredPoolsCleanupJob

### DIFF
--- a/server/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/server/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -19,13 +19,13 @@ import static org.candlepin.common.config.ConfigurationPrefixes.*;
 
 import org.candlepin.async.tasks.ActiveEntitlementJob;
 import org.candlepin.async.tasks.CRLUpdateJob;
+import org.candlepin.async.tasks.ExpiredPoolsCleanupJob;
 import org.candlepin.async.tasks.ManifestCleanerJob;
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -459,7 +460,8 @@ public static String jobConfig(String jobKey, String cfgName) {
             // Async job defaults and scheduling
             this.put(ASYNC_JOBS_THREADS, "10");
 
-            this.put(jobConfig(ManifestCleanerJob.JOB_KEY, ASYNC_JOBS_JOB_ENABLED), "true");
+            this.put(jobConfig(ExpiredPoolsCleanupJob.JOB_KEY, ASYNC_JOBS_JOB_SCHEDULE),
```
nikosmoum: One thing I forgot: you didn't remove the `ExpiredPoolsJob.class.getName(),` entry from the `DEFAULT_TASK_LIST` in this file.

Ceiu:
Removed the Pinsetter ExpiredPoolsJob from the default task list and rebased against the latest feature branch.

Ceiu:
Looks like this task also has spec tests which trigger the job. We'll hold off on merging this until after #2379 is merged, which adds a means of manually triggering arbitrary async jobs.

PR #2381, opened 2019-07-04
[F] Port ImportRecordJob to the new job framework (ENT-865)

nikosmoum:
- Ported the ImportRecordJob to the Artemis job system
  as the ImportRecordCleanerJob

### DIFF
--- a/server/spec/content_versioning_spec.rb
+++ b/server/spec/content_versioning_spec.rb
@@ -279,7 +279,7 @@ describe 'Content Versioning' do
       end
 
       sleep 1
-      pp @cp.trigger_async_job("OrphanCleanupJob");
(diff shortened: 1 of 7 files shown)


Ceiu:
Like the other two pending PRs, this needs a spec test updated to trigger it. A rebase after #2379 is merged and a minor update to the job_schedule_spec should take care of it.

nikosmoum:
@Ceiu PR updated. I changed the triggering of OrhanCleaner jobs in a couple of spec tests to use the proper JOB_KEY :`      @cp.trigger_async_job("ORPHAN_CLEANUP")`. Apparently those spec tests pass regardless of that job running.

PR #2382, opened 2019-07-08
[F] Ported the ActiveEntitlementJob to the Artemis job system (ENT-858)

abhiskum:


Ceiu:
ack from me

@nikosmoum I'll leave hitting the button to you since you're assigned

PR #2383, opened 2019-07-09
[F] Ported the JobCleaner to the Artemis job system (ENT-862)

abhiskum:


server/src/main/java/org/candlepin/async/tasks/AsyncJobCleaner.java
```diff
@@ -0,0 +1,62 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Date;
+
+/**
+ * AsyncJobCleaner removes finished jobs older than yesterday, and failed
+ * jobs from 4 days ago.
+ */
+public class AsyncJobCleaner implements AsyncJob {
```
nikosmoum: Why change the name of the job? The 'Async' prefix is not consistent with the other job names. If you're worried about them having the same name with the old one, the old job classes will eventually get deleted before this feature branch gets merged into master.

server/src/main/java/org/candlepin/guice/CandlepinModule.java
```diff
@@ -448,6 +449,7 @@ private void configureAsyncJobs() {
         JobManager.registerJob(UnmappedGuestEntitlementCleanerJob.JOB_KEY,
             UnmappedGuestEntitlementCleanerJob.class);
         JobManager.registerJob(ActiveEntitlementJob.JOB_KEY, ActiveEntitlementJob.class);
+        JobManager.registerJob(AsyncJobCleaner.JOB_KEY, AsyncJobCleaner.class);
```
nikosmoum: We try to keep the list of registrations alphabetical here for readability, so it would be consistent to move this and the `ActiveEntitlementJob` above it to the start of the order.

server/src/main/java/org/candlepin/async/tasks/AsyncJobCleaner.java
abhiskum: As this cleaner job deals with async jobs hence I add Async prefix like AsyncJobResource. However, no issue to remove the async prefix. I will update PR.

abhiskum: Done.

server/src/main/java/org/candlepin/guice/CandlepinModule.java
abhiskum: Done.

PR #2385, opened 2019-07-10
(ENT-1377) Port logic from KeycloakOIDCServletFilter to KeycloakAuth

ShwethaKalyanaraman:
**PR for Integrating Keycloak on Candlepin side which caters to authentication through:**

- Bearer Token

- Refresh Token(Offline Token Based Authentication)

**Prerequisites:** Put keycloak.json in _/etc/candlepin/keycloak.json_ location and setup keycloak locally or use sso dev of redhat

**Details about each file:**
**Following files have been modified with reasons specified in bracket:**

- _build.gradle_ (Add dependencies to make keycloak work)

- _KeycloakAuth_ (Implements AuthProvider and mainly gets the principal either using bearer token method or refresh token method)

- _KeycloakOIDCFacade_ ( Handles request/response by implementing HttpFacade)

- _KeycloakRequestAuthenticator_ (KeycloakRequestAuthenticator is a customized RequestAuthenticator for Keycloak Integration)

- _ConfigProperties_ (Added a few final strings to enable/disable keycloak)

- _StatusDTO_ ( Implemented /status endpoint to add three extra parameters)

- _StatusDTOTest_ (Modified the test to incorporate changes made in StatusDTO)

- _CandlepinFilterModule_ (Used regex to make each endpoint work as expected)

- _StatusResource_ (A part of /status endpoint implementation)

- _Authentication Filter_ ( Injected another filter - KeycloakAuth)

- _token/index.html_ ( Prints refresh token from keycloak server)

- _keycloak.js_ ( Initializes and logs into keycloak server)


server/src/main/webapp/token/keycloak.js
```diff
@@ -0,0 +1,1424 @@
...
+    if ( typeof module === "object" && module && typeof module.exports === "object" ) {
+        module.exports = Keycloak;
+    } else {
+        window.Keycloak = Keycloak;
+
+        if ( typeof define === "function" && define.amd ) {
+            define( "keycloak", [], function () { return Keycloak; } );
+        }
+    }
+})( window );
```
Ceiu: Missing linebreak at the end of the file.

server/src/main/webapp/token/index.html
```diff
@@ -0,0 +1,65 @@
...
+                var clientId;
+
+                if (xhttp.readyState == 4 && xhttp.status == 200) {
+                        var status = JSON.parse(xhttp.responseText);
+                        realm= status['realm']
+                        authUrl = status['authUrl']
+                        clientId = status['resource']
+                }
+                else {
+                        console.log("Error", xhttp.statusText);
```
Ceiu: This branch will be hit on every state change outside of state 4; most of which will not be errors. This statement would belong in a branch unique to the status check:

```
if (xhttp.readyState == 4) {
    if (xhttp.status == 200) {
        ...
    }
    else {
        console.log(...);
    }
}
```

server/src/main/webapp/token/index.html
```diff
@@ -0,0 +1,65 @@
...
+                }
+
+                var keycloak = Keycloak({
+                        url: authUrl,
+                        realm: realm,
+                        clientId: clientId
+                });
+                keycloak.init().success(function(authenticated) {
+                        if(!authenticated)
+                        {
```
Ceiu: curly brace should move up to the end of the previous line; missing whitespace between if and the open paren.

server/src/main/webapp/token/index.html
```diff
@@ -0,0 +1,65 @@
...
+                                keycloak.login();
+                        } else {
```
Ceiu: Move the ```else {``` down to the next line.

server/src/main/webapp/token/index.html
```diff
@@ -0,0 +1,65 @@
...
+        var xhttp = new XMLHttpRequest();
+        xhttp.open('GET', "https://localhost:8443/candlepin/status", true);
+        xhttp.onreadystatechange = function() {
+                var realm;
+                var authUrl;
+                var clientId;
+
+                if (xhttp.readyState == 4 && xhttp.status == 200) {
+                        var status = JSON.parse(xhttp.responseText);
+                        realm= status['realm']
```
Ceiu: missing space between "realm" and '='

server/src/main/webapp/token/index.html
```diff
@@ -0,0 +1,65 @@
...
+                        authUrl = status['authUrl']
+                        clientId = status['resource']
+                }
+                else {
+                        console.log("Error", xhttp.statusText);
+                }
+
+                var keycloak = Keycloak({
```
Ceiu: This bit, and everything below is going to be run on every readyState and status combination. I'm assuming it's only intended to be run on readyState 4, in which case it belongs in the block above.

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
```diff
@@ -0,0 +1,166 @@
...
+                if (authArray[0].equals("Basic")) {
+                    return null;
+                }
+                else {
+                    Base64 base64Url = new Base64(true);
+                    String[] splitstring = authArray[1].split("\\.");
+                    String base64EncodedBody = splitstring[1];
+                    String body = new String(base64Url.decode(base64EncodedBody));
+                    JSONObject obj = new JSONObject(body);
+                    String tokenType = obj.getString("typ");
```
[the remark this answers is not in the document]
ShwethaKalyanaraman: It is typ only after parsing

ShwethaKalyanaraman:
@Ceiu , I will work on the changes

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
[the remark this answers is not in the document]
Ceiu: I don't understand. The key should be safe from any of the parsing being done here, as it would only be exposed after converting the decoded base64 to JSON. At which point, we shouldn't be doing any truncating or splitting on the data.

If the key is indeed "typ", is this something we have control over? Is "typ" an abbreviation for something? Why is the 'e' missing from the key name here?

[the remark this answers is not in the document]
ShwethaKalyanaraman: Honestly I dont know why its named typ and not type. 
Have a look at the picture which parses the Jwt token.
image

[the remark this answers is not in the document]
Ceiu: Yeesh... looks like they wanted to keep every key at three characters in length.

Alright, thanks for the details here. Disregard my request.

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
```diff
@@ -0,0 +1,166 @@
...
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.auth;
+
+
+import org.jboss.resteasy.spi.HttpRequest;
+import org.json.JSONObject;
```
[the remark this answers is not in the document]
ShwethaKalyanaraman: There is no specific reason. I just found it easier to use it this way. Let me know if it has to be changed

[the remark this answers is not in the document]
Ceiu: Unless there's a specific reason mandating we use an external dependency, we should try to avoid adding them. Especially in cases like this where we already have two other means of processing JSON (Jackson and native Java object).

server/src/main/webapp/token/index.html
```diff
@@ -0,0 +1,65 @@
...
+<html lang="en">
+<head>
+        <meta charset="utf-8"/>
+</head>
+<body>
+<script src="keycloak.js"></script>
+<script>
+        var status;
+        var xhttp = new XMLHttpRequest();
+        xhttp.open('GET', "https://localhost:8443/candlepin/status", true);
```
jirihnidek: This cannot be 'localhost', because it will not work in production. You something like this:

```javascript
var host = window.location.hostname;
xhttp.open('GET', "https://" + host + "8443/candlepin/status", true);
```

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
```diff
@@ -72,23 +90,20 @@ public KeycloakAuth(UserServiceAdapter userServiceAdapter, Provider<I18n> i18nPr
...
+        try {
             InputStream is = loadKeycloakConfigFile();
             kd = KeycloakDeploymentBuilder.build(is);
             deploymentContext = new AdapterDeploymentContext(kd);
-
         }
-        catch (Exception e) {
-
+        catch (FileNotFoundException e) {
+            log.error("{} :File not found", filepath);
```
Ceiu: It's good that we're logging this now, but what happens if this fails? What state are we left in?

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
```diff
@@ -72,23 +90,20 @@ public KeycloakAuth(UserServiceAdapter userServiceAdapter, Provider<I18n> i18nPr
...
         }
-
     }
 
     private InputStream loadKeycloakConfigFile() throws FileNotFoundException {
-        String filepath = "/etc/candlepin/keycloak.json";
+        filepath = ConfigProperties.KEYCLOAK_FILEPATH;
         return new FileInputStream(filepath);
```
Ceiu: Looking at this whole thing in depth, a few of things come to mind:

1. The keycloak file in question is a configuration file. Candlepin already has its own configuration file and format, and we could just as easily store everything there instead of introducing another file dependency.
2. If we want to stay with a secondary configuration file for this, we should still be loading it here in a way that lets us modify it as we need to, setting reasonable defaults and whatnot as necessary. Following this, I would argue passing around blackbox, unvalidated input streams is not the way to go.
3. This is still a hard-coded path, not a configurable one. 

In any case, we should be building the AdapterConfig ourselves and passing that in to KeycloakdeploymentBuilder.build, rather than throwing a file at it and hoping for the best. Additionally, I would argue the relevant configurations should be stored in CP's config rather than a separate config entirely.

To recap: At the absolute least, this needs to *not* be a hardcoded constant file. This should be a configurable reference, or values from the CP configuration, used to build an AdapterConfig that is passed into the KeycloakDeploymentBuilder.

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
```diff
@@ -104,20 +119,28 @@ public Principal getPrincipal(HttpRequest httpRequest) {
...
-                    if (tokenType.equals("Refresh")) {
-                        //Offline Token authentication
-                        handleRefreshToken(httpRequest, auth);
+                    try {
+                        Base64 base64Url = new Base64(true);
+                        String[] splitstring = authArray[1].split("\\.");
+                        String base64EncodedBody = splitstring[1];
+                        String body = new String(base64Url.decode(base64EncodedBody));
+                        ObjectMapper objectMapper = new ObjectMapper();
+                        JsonNode rootNode = objectMapper.readTree(body);
```
Ceiu: Digging into this a bit more, we don't need to be manually parsing these JSON structures. The Keycloak library provides DTOs which are annotated for JSON parsers like Jackson. In this particular case, we should be injecting the ObjectMapper (not creating it new here) and then using it to pull the appropriate request object from Keycloak's own representation package (see: JsonWebToken).

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
```diff
@@ -104,20 +119,28 @@ public Principal getPrincipal(HttpRequest httpRequest) {
...
+                        String tokenType = rootNode.get("typ").asText();
+                        Token token = Token.valueOf(tokenType.toUpperCase());
+                        switch(token) {
```
Ceiu: This is good, but as I noted above, these are already defined for us by the library in question. In such a case, we don't want to duplicate those definitions, as it allows for potential problems during version changes and future maintenance.

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
```diff
@@ -104,20 +119,28 @@ public Principal getPrincipal(HttpRequest httpRequest) {
...
+                                break;
+                            default:
+                                break;
+                        }
                     }
-                    else if (tokenType.equals("Bearer")) {
-                        //Bearer only authentication
-                        handleBearerToken(httpRequest);
+                    catch (UnsupportedEncodingException e) {
+                        throw new UnsupportedEncodingException(i18nProvider.get().tr("Decoding Failed"));
```
Ceiu: In a case like this where we're throwing the exact same exception that we just caught, we're better off not catching it at all. In this particular case, because the new exception doesn't retain the original as its cause, we lose the origin of the problem and all of the stack information with it.

server/src/main/java/org/candlepin/auth/KeycloakOIDCFacade.java
```diff
@@ -45,17 +45,17 @@
...
      */
-
     protected class RequestFacade implements OIDCHttpFacade.Request {
 
         private InputStream inputStream;
 
         @Override
         public String getFirstParam(String param) {
-            throw new RuntimeException("NOT IMPLEMENTED");
+            throw new UnsupportedOperationException("NOT IMPLEMENTED");
```
Ceiu: Inconsistent error message

server/src/main/java/org/candlepin/auth/KeycloakOIDCFacade.java
```diff
@@ -116,8 +114,8 @@ public InputStream getInputStream() {
 
         @Override
         public InputStream getInputStream(boolean buffered) {
-            if (inputStream != null) {
-                return inputStream;
+            if (this.inputStream != null) {
+                return this.inputStream;
```
Ceiu: Good, but this still is never used, so it serves no purpose. The input stream will always be fetched from the requestContext.

server/src/main/java/org/candlepin/auth/KeycloakOIDCFacade.java
```diff
@@ -131,19 +129,21 @@ public String getRemoteAddr() {
...
     }
+
     /**
      * ResponseFacade for Keycloak Responses
      */
     protected class ResponseFacade implements OIDCHttpFacade.Response {
 
-        private javax.ws.rs.core.Response.ResponseBuilder responseBuilder =
+        private ResponseBuilder responseBuilder =
             javax.ws.rs.core.Response.status(204);
```
Ceiu: Still fully qualifying the Response class here. Also, this all fits onto one line now.

server/src/main/java/org/candlepin/auth/KeycloakOIDCFacade.java
```diff
@@ -210,10 +209,10 @@ public Response getResponse() {
 
     @Override
     public X509Certificate[] getCertificateChain() {
-        throw new IllegalStateException("Not supported yet");
+        throw new UnsupportedOperationException("Not supported yet");
     }
 
     public boolean isResponseFinished() {
-        return responseFinished;
+        return true;
```
Ceiu: This is different now, but it doesn't address the problem cited: now we're always returning true, and we still have a useless variable in responseFinished that is never updated or (now) referenced.

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -277,7 +277,10 @@ private ConfigProperties() {
 
     public static final String SWAGGER_ENABLED = "candlepin.swagger.enabled";
     public static final String TOKEN_ENABLED = "candlepin.token.enabled";
-
+    /**
+     * Keycloak Filepath needed for authentication
+     */
+    public static final String KEYCLOAK_FILEPATH = "/etc/candlepin/keycloak.json";
```
Ceiu: The way this class works is the constant you're defining here is the config name that will appear in candlepin.conf. Like the above TOKEN_ENABLED, I would expect this to look like "candlepin.keycloak.config" or some such.

Once this is defined, we may also want to define a default, further down in the file (see how TOKEN_ENABLED is used for an example). However, whether or not we add a default depends on the feature in question. In this case, I'm not so sure.

server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -93,23 +87,19 @@ public StatusResource(RulesCurator rulesCurator, Configuration config, JsRunnerP
...
+        try (FileReader reader = new FileReader(ConfigProperties.KEYCLOAK_FILEPATH)) {
             //Read JSON file
-            JSONParser jsonParser = new JSONParser();
-            JSONObject jsonobj = (JSONObject) jsonParser.parse(reader);
-            realm = (String) jsonobj.get("realm");
-            authUrl = (String) jsonobj.get("auth-server-url");
-            resource = (String) jsonobj.get("resource");
-
+            ObjectMapper objectMapper = new ObjectMapper();
+            JsonNode rootNode = objectMapper.readTree(reader);
```
Ceiu: A couple important things here:

1. We don't want to roll our own parser if we don't need to. Keycloak already provides this in the form of AdapterConfig.
2. We should not be performing configuration parsing/loading at this point. It should already have been done and we should be referencing that already-parsed config at this point.

server/src/main/java/org/candlepin/auth/KeycloakOIDCFacade.java
ShwethaKalyanaraman: We cant shorten the Response class here as we need javax.ws.rs.core.Response.status because it conflicts with OIDCHttpFacade.Response

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
```diff
@@ -29,40 +31,56 @@
...
     private AccessTokenResponse response = null;
-
     private KeycloakDeployment kd;
+    private String filepath = null;
+    private RefreshableKeycloakSecurityContext securityContext = null;
+
+    /**
+     * Enum for Different types of Tokens
+     */
+    public enum Token {
```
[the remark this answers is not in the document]
ShwethaKalyanaraman: Yes, it is token type

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
kahowell: > 1. The keycloak file in question is a configuration file. Candlepin already has its own configuration file and format, and we could just as easily store everything there instead of introducing another file dependency.

The idiomatic way to integrate with Keycloak is to generate keycloak.json through the Keycloak admin interface, and then land this as part of the configuration of your application. In order to keep things simple and idiomatic, I want to ask that we keep keycloak.json separate from the rest of candlepin configuration.

> 2. If we want to stay with a secondary configuration file for this, we should still be loading it here in a way that lets us modify it as we need to, setting reasonable defaults and whatnot as necessary. Following this, I would argue passing around blackbox, unvalidated input streams is not the way to go.

Agree, it looks like we can use [`KeycloakDeployBuilder.loadAdapterConfig(InputStream)`](https://github.com/keycloak/keycloak/blob/ccc8e06f9a7d11db9c2848f5891bdb3f75c40e33/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/KeycloakDeploymentBuilder.java#L191), and then apply any overrides to the loaded AdapterConfig object. I don't think we have any overrides we need to make at the moment, so @ShwethaKalyanaraman please leave a comment that leaves a hint where we can apply overrides. something like:

```java
AdapterConfig config = KeycloakDeployBuilder.loadAdapterConfig(inputStream);
// NOTE: Candlepin-specific overrides to keycloak config can be made in this method.
return config;
```

> 3. This is still a hard-coded path, not a configurable one.

I'll suggest a name: `candlepin.keycloak.config_path`, with a default of `/etc/candlepin/keycloak.json`.

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
ShwethaKalyanaraman: You basically want me to add logs after the catch right?

Ceiu: Not quite.

If we hit either exception here, the module's state is in a dead/failed state. Keycloak will not work and, at best, it'll silently do nothing while Candlepin keeps chugging along. Worse, it may just cause more problems when other login/request processing occurs (due to the sequential processing nature I mentioned before).

This appears to me to be a fatal exception -- one that needs to propagate up and stop Candlepin from completing normal operation until it is fixed. I could be wrong here, but if I am, it would be nice to have some documentation, comments or indication of the state CP is in and why it's okay that it's just a one-line log entry.

ShwethaKalyanaraman:
> A number of questions to consider:
> 
> * This appears to be something that will always be enabled, yet may only have value within select environments. How would a user disable keycloak entirely if they don't need it?

_One can disable keycloak only through the conf file of candlepin by setting it to false. 
Setting it to false would print a message in subman stating - _Token Option is disabled__

> * What happens to Candlepin in the scenarios where it is enabled, but is misconfigured? What about other unexpected, fatal exceptions?

_Other unexpected fatal exceptions would be found in log file. The check in Authentication filter before injecting KeycloakAuth ensures misconfiguration doesnt happen. Incase it does, logs would help._

> * If it is disabled, does Candlepin still invoke the various adapters, interceptors and whatnot for it? Does it gracefully handle those things not being present?

_When keycloak is disabled, subman simply outputs as disabled(after checking status endpoint) and it never enters keycloak auth or any adapters._

Ceiu:
I'll schedule some time to look at this tomorrow with the a proper Keycloak setup, but in the interim, there are no unit tests for the new stuff, and no spec tests for the endpoint changes.

ShwethaKalyanaraman:
> I'll schedule some time to look at this tomorrow with the a proper Keycloak setup, but in the interim, there are no unit tests for the new stuff, and no spec tests for the endpoint changes.

Sure. I will work on it

server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -86,22 +83,11 @@ public StatusResource(RulesCurator rulesCurator, Configuration config, JsRunnerP
...
-        }
-        catch (FileNotFoundException e) {
-            log.error("Keycloak.json file not found", e);
-        }
-        catch (IOException e) {
-            log.error("IOException while using Keycloak.json", e);
+        if (keycloakAdapterConfiguration.getAdapterConfig() != null) {
+            realm = keycloakAdapterConfiguration.getAdapterConfig().getRealm();
+            authUrl = keycloakAdapterConfiguration.getAdapterConfig().getAuthServerUrl();
+            resource = keycloakAdapterConfiguration.getAdapterConfig().getResource();
```
Ceiu: A lot of repetition here. Grab the config once, reference it locally. This is also more efficient in cases where the getter isn't a simple one-liner the compiler will optimize out. And since the StatusResource object has no relation to the keycloak config, it should be treating all of the keycloak objects as black box.

```
AdapterConfig keycloakConfig = keycloakAdapterConfiguration.getAdapterConfig();
if (keycloakConfig != null) {
    this.realm = keycloakConfig.getRealm();
    this.authUrl = keycloakConfig.getAuthServerUrl();
    this.resource = keycloakConfig.getResource();
}
```

Also, whitespace is your friend. Separate logical blocks with a line of whitespace to separate them from other logic and make it clear at a glance that it is not connected to the previous block.

Ceiu:
Whelp... apparently I don't have access to keycloak, so the rest of the testing I wanted to do will have to wait, unfortunately. I'll try to get this sorted on Monday.

server/src/main/webapp/token/keycloak.js
```diff
@@ -0,0 +1,1420 @@
...
+                        if (options && options.prompt == 'none') {
+                            o += ',hidden=yes';
+                        }
+
+                        var loginUrl = kc.createLoginUrl(options);
+                        var ref = cordovaOpenWindowWrapper(loginUrl, '_blank', o);
+                        var completed = false;
+
+                        ref.addEventListener('loadstart', function(event) {
+                            if (event.url.indexOf('http://localhost') == 0) {
```
jirihnidek: Is it good idea to have here localhost?

server/src/main/webapp/token/index.html
```diff
@@ -0,0 +1,66 @@
...
+                                var status = JSON.parse(xhttp.responseText);
+                                realm = status['realm']
+                                authUrl = status['authUrl']
+                                clientId = status['resource']
+                        }
+                        var keycloak = Keycloak({
+                                url: authUrl,
+                                realm: realm,
+                                clientId: clientId
+                        });
```
jirihnidek: When xhttps.status is not 200, then `clientId` is undefined. BTW: Canldepin returned following result for `/candlepin/status`:

```json
{
  "mode" : "NORMAL",
  "modeReason" : "STARTUP",
  "modeChangeTime" : "2019-08-07T12:02:54+0000",
  "result" : true,
  "version" : "2.8.0",
  "rulesVersion" : "5.37",
  "release" : "1",
  "standalone" : true,
  "timeUTC" : "2019-08-07T12:21:54+0000",
  "realm" : null,
  "authUrl" : null,
  "resource" : null,
  "rulesSource" : "default",
  "managerCapabilities" : [ "instance_multiplier", "derived_product", "vcpu", "cert_v3", "hypervisors_heartbeat", "remove_by_pool_id", "syspurpose", "storage_band", "cores", "hypervisors_async", "org_level_content_access", "guest_limit", "ram", "batch_bind" ]
}
```

server/src/main/webapp/token/index.html
```diff
@@ -0,0 +1,66 @@
...
+<head>
+        <meta charset="utf-8"/>
+</head>
+<body>
+<script src="keycloak.js"></script>
+<script>
+        var status;
+        var xhttp = new XMLHttpRequest();
+        var host = window.location.hostname;
+        xhttp.open('GET', "https://" + host + ":8443/candlepin/status", true);
```
jirihnidek: Other values will be used in production. Port will be 443 and `/candlepin` will be replaced with `/subsription` or `/rhsm`

server/src/main/webapp/token/keycloak.js
ShwethaKalyanaraman: keycloak.js is an inbuilt file which I took from keycloak repository. 

> Hi, have you ever test it in vagrant box or is it necessary to run candlepin on localhost?

I have not tested it on vagrant box but only on localhost till now

server/src/main/webapp/token/index.html
ShwethaKalyanaraman: Did you enable keycloak auth in candlepin.conf?

server/src/main/webapp/token/keycloak.js
jirihnidek: OK. I will try it on localhost too. BTW: does keycloak require to run candlepin on machine with fqdn? Some SSOs have such requirement, but I didn't find such information about keycloak.

ShwethaKalyanaraman: Nothing as such about keycloak

Ceiu:
I'm *still* waiting on keycloak access, so rather than continuing to hold this up, perhaps someone can test the following scenarios and verify they are safe:

- Configure CP to use keycloak, but do not provide a valid path to a keycloak.conf
    - How does CP behave? Does it start at all? If so, does it respond to all/any requests?
- Configure CP to use keycloak, but provide invalid settings in keycloak.conf
    - Same questions as above
- Configure CP to not use keycloak at all
    - Does CP still load any keycloak classes? Are keycloak processes run on any/all requests? Does the response from the status endpoint still make sense? Is it intuitive for (potentially unaware) users?

Ceiu:
In addition to the above questions:

- What does the logging output look like in the failure scenarios?
- When a failure does occur (such as a timeout hitting the configured keycloak server), how does CP recover?

Basically, the key part I want to verify is whether or not CP fails completely early on, or gracefully late. Anything in between is something we want to avoid. Try to be as sure as possible that no matter what happens between CP and keycloak that CP is always in a known/expected state and can either recover gracefully or fail entirely.

ShwethaKalyanaraman:
> I'm _still_ waiting on keycloak access, so rather than continuing to hold this up, perhaps someone can test the following scenarios and verify they are safe:
-  Yes sure, someone with keycloak access can test it out. 
I will provide my inputs on these points. 
> * Configure CP to use keycloak, but do not provide a valid path to a keycloak.conf
>   * How does CP behave? Does it start at all? If so, does it respond to all/any requests?
 - When you give an invalid path to keycloak.conf . You get a warning in candlepin.log as -> "Keycloak.json file not found"
CP works fine. It responds to other requests.
> * Configure CP to use keycloak, but provide invalid settings in keycloak.conf
- On providing invalid settings, candlepin works fine and it logs the appropriate error in candlepin.conf
  There are two errors that has been taken care of :
   - Invalid Json format
   - Invalid keywords or not the right keywords basically
> * Configure CP to not use keycloak at all
>   * Does CP still load any keycloak classes? Are keycloak processes run on any/all requests? Does the response from the status endpoint still make sense? Is it intuitive for (potentially unaware) users?
- When you disable keycloak in CP by making changes to candlepin.conf file and setting keycloak.auth to false -> CP doesnt load any keycloak related classes . Status endpoint works fine but only the three parameters that is needed will be set to null as keycloak is disabled.

server/src/main/java/org/candlepin/auth/KeycloakAdapterConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+        return keycloakDeployment;
+    }
+
+    public AdapterConfig getAdapterConfig() {
+        return adapterConfig;
+    }
+
+    public RequestAuthenticator getRequestAuthenticator(HttpRequest httpRequest) {
+        KeycloakOIDCFacade keycloakOIDCFacade = new KeycloakOIDCFacade(httpRequest);
+        return new CandlepinKeycloakRequestAuthenticator(keycloakOIDCFacade, httpRequest, keycloakDeployment);
```
Ceiu: Is this something that should be built once and shared between all callers, or do we want to create a new instance on every invocation?

server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -161,7 +166,10 @@ public StatusDTO status() {
             .setModeReason(modeChangeReason != null ? modeChangeReason.toString() : null)
             .setModeChangeTime(modeChange.getChangeTime())
             .setManagerCapabilities(caps)
-            .setTimeUTC(new Date());
+            .setTimeUTC(new Date())
+            .setResource(resource)
```
Ceiu: These three additions introduce an interesting problem. In the vast majority of expected cases, it will have no value and will always be null, but we still want them to be available as necessary.

In this case, I would argue the solution here is to add a subclass of StatusDTO that provides the three additional fields. If we're operating with keycloak enabled, the subclass should be used and those fields set. Otherwise, we default to the "standard" StatusDTO that does not provide those fields.

The fields in the DTO should probably also be clearer. Maybe keycloakRealm or some such.

server/src/main/java/org/candlepin/resteasy/filter/AuthenticationFilter.java
```diff
@@ -91,6 +96,11 @@ public void setHttpServletRequest(HttpServletRequest request) {
     }
 
     public void setupAuthStrategies() {
+        // use keycloak authentication
+        if (config.getBoolean(ConfigProperties.KEYCLOAK_AUTHENTICATION) &&
```
Ceiu: Remember to default to something reasonable (false) in these cases so we can avoid crashing unnecessarily.

server/src/main/java/org/candlepin/auth/KeycloakAdapterConfiguration.java
ShwethaKalyanaraman: It should be built once and shared between all callers

server/src/main/java/org/candlepin/resteasy/filter/AuthenticationFilter.java
ShwethaKalyanaraman: Willl work on the suggestions!

server/src/main/java/org/candlepin/dto/api/v1/KeycloakStatusDTO.java
```diff
@@ -0,0 +1,121 @@
...
+        return keycloakRealm;
+    }
+
+    public StatusDTO setKeycloakRealm(String keycloakRealm) {
+        this.keycloakRealm = keycloakRealm;
+        return this;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
Ceiu: Most of this should be replaced by a call to the base equals check via super.equals(obj). Take a look at any of the existing DTOs that extend TimestampedCandlepinDTO for an example.

server/src/main/java/org/candlepin/dto/api/v1/KeycloakStatusDTO.java
```diff
@@ -0,0 +1,121 @@
...
+
+        return false;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int hashCode() {
+        HashCodeBuilder builder = new HashCodeBuilder(7, 17)
```
Ceiu: super.hashCode needs to be called here

server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -154,22 +156,30 @@ public StatusDTO status() {
...
-            .setMode(mode != null ? mode.toString() : null)
-            .setModeReason(modeChangeReason != null ? modeChangeReason.toString() : null)
-            .setModeChangeTime(modeChange.getChangeTime())
-            .setManagerCapabilities(caps)
-            .setTimeUTC(new Date())
-            .setResource(resource)
-            .setAuthUrl(authUrl)
-            .setRealm(realm);
+        if (adapterConfig != null) {
+            KeycloakStatusDTO keycloakstatus = new KeycloakStatusDTO();
```
Ceiu: You can avoid the redeclaration and extra reassignment here by using invocation chaining:

```
if (adapterConfig != null) {
    status = new KeycloakStatusDTO()
        .setKeycloakResource(...)
        .setKeycloakAuthUrl(...)
        ...
}
```

server/src/main/java/org/candlepin/auth/KeycloakAdapterConfiguration.java
```diff
@@ -42,18 +41,23 @@
...
             }
-            catch (FileNotFoundException e) {
-                log.warn("Keycloak.json file not found", e);
+            catch (IOException e) {
+                log.error("Unable to read keycloak configuration file: {}", configFile);
+                throw e;
             }
-            catch (RuntimeException e) {
-                log.warn("Unable to read keycloak.json", e);
+            catch (Exception e) {
```
Ceiu: One thing that needs to be retained that I didn't cover in my example was the RuntimeException handling, since that's what Keycloak uses for bad configuration exceptions.

server/src/main/java/org/candlepin/dto/api/v1/KeycloakStatusDTO.java
```diff
@@ -0,0 +1,121 @@
...
+    public StatusDTO setKeycloakAuthUrl(String keycloakAuthUrl) {
+        this.keycloakAuthUrl = keycloakAuthUrl;
+        return this;
+    }
+
+    public String getKeycloakRealm() {
+        return keycloakRealm;
+    }
+
+    public StatusDTO setKeycloakRealm(String keycloakRealm) {
```
Ceiu: This should be returning a KeycloakStatusDTO

server/src/main/java/org/candlepin/dto/api/v1/KeycloakStatusDTO.java
```diff
@@ -0,0 +1,121 @@
...
+    public KeycloakStatusDTO setKeycloakResource(String keycloakResource) {
+        this.keycloakResource = keycloakResource;
+        return this;
+    }
+
+    public String getKeycloakAuthUrl() {
+        return keycloakAuthUrl;
+    }
+
+    public StatusDTO setKeycloakAuthUrl(String keycloakAuthUrl) {
```
Ceiu: KeycloakStatusDTO

server/src/main/java/org/candlepin/dto/api/v1/KeycloakStatusDTO.java
```diff
@@ -0,0 +1,121 @@
...
+            .append(this.getTimeUTC())
+            .append(this.getManagerCapabilities())
+            .append(this.getKeycloakRealm())
+            .append(this.getKeycloakAuthUrl())
+            .append(this.getKeycloakResource());
+
+        return builder.toHashCode();
+    }
+
+
```
Ceiu: Extra whitespace between methods here.

server/src/main/java/org/candlepin/dto/api/v1/KeycloakStatusDTO.java
```diff
@@ -0,0 +1,121 @@
...
+    public KeycloakStatusDTO populate(KeycloakStatusDTO source) {
```
Ceiu: Missing javadoc block.

kahowell:
2f6712ed3..63f876239: rebase against master, squash

kahowell:
63f8762..d85d578: fixup (missed import with my rebase); doh!

server/src/main/java/org/candlepin/auth/KeycloakOIDCFacade.java
```diff
@@ -0,0 +1,213 @@
...
+        @Override
+        public InputStream getInputStream(boolean buffered) {
+            InputStream inputStream = requestContext.getInputStream();
+            return inputStream;
+        }
+
+        @Override
+        public String getRemoteAddr() {
+            // TODO: implement properly
+            return HostUtils.getIpAddress();
```
Ceiu: @kahowell This is actually fetching the local address, not the remote address.

server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -150,7 +149,19 @@ public StatusDTO status() {
 
         RulesSourceEnum rulesSource = jsProvider.getRulesSource();
 
-        StatusDTO status = new StatusDTO()
+        StatusDTO status;
+        if (adapterConfig != null) {
+            KeycloakStatusDTO keycloakStatus = new KeycloakStatusDTO()
```
Ceiu: @kahowell We can clean this up such that we skip an extra variable declaration:

```
status = new KeycloakStatusDTO()
    .SetKeycloakResource(...)
    ....();
```

server/src/main/java/org/candlepin/resteasy/filter/AuthenticationFilter.java
Ceiu: @kahowell Needs the default value set to avoid an exception.

server/src/main/java/org/candlepin/auth/KeycloakAdapterConfiguration.java
```diff
@@ -66,14 +71,30 @@ public KeycloakAdapterConfiguration(Configuration configuration) throws Exceptio
...
 
+    /**
+     * Create and return a new {@link RequestAuthenticator} for the given request.
+     *
+     * @param httpRequest the request to attempt authentication against
+     * @return an instance of {@link CandlepinKeycloakRequestAuthenticator}
+     */
     public RequestAuthenticator getRequestAuthenticator(HttpRequest httpRequest) {
         KeycloakOIDCFacade keycloakOIDCFacade = new KeycloakOIDCFacade(httpRequest);
         return new CandlepinKeycloakRequestAuthenticator(keycloakOIDCFacade, httpRequest, keycloakDeployment);
```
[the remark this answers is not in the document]
kahowell: Unfortunately, the name of the class - `RequestAuthenticator` - doesn't communicate effectively that an instance is specific to a given request. A given instance is only used once in `org.candlepin.auth.KeycloakAuth#handleBearerToken`, which should only be invoked once per request so we won't gain anything by caching it. What do you think of renaming the method to be more clear? `createAuthenticatorForRequest(HttpRequest)`?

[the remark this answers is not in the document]
Ceiu: I'm fine with whatever, I'm just basing my request on the response from my initial question asking what the intent was. If your assessment is different, then we can leave it as-is.

server/src/main/java/org/candlepin/resteasy/filter/AuthenticationFilter.java
kahowell: We added it as a default in https://github.com/candlepin/candlepin/blob/009b0ff1087d4ec2fb19b50f8d4f41adf8acc20d/server/src/test/java/org/candlepin/resteasy/filter/AuthenticationFilterTest.java#L120 ; that should be sufficient, no?

If not, I'll make another commit to make sure everywhere we use it has the default.

kahowell: See updated link if you already read.

server/src/main/java/org/candlepin/auth/KeycloakAuth.java
```diff
@@ -0,0 +1,163 @@
...
+                    String tokenType = TokenVerifier.create(authArray[1],
+                        JsonWebToken.class).getToken().getType();
+                    switch (tokenType) {
+                        case TokenUtil.TOKEN_TYPE_BEARER:
+                            handleBearerToken(httpRequest);
+                            break;
+                        case TokenUtil.TOKEN_TYPE_REFRESH:
+                            handleRefreshToken(httpRequest, auth);
+                            break;
+                        default:
```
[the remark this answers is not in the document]
kahowell: Good call!

kahowell:
I have one more piece of planned work: the capability. And then I'd like to squash my commits before merging.

kahowell:
@Ceiu please take another look when you get a chance.

server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -83,7 +84,14 @@ public StatusResource(RulesCurator rulesCurator, Configuration config, JsRunnerP
         if (config == null || !config.getBoolean(ConfigProperties.STANDALONE)) {
             standalone = false;
         }
+
         adapterConfig = keycloakAdapterConfiguration.getAdapterConfig();
+        if (adapterConfig != null && config != null &&
+            config.getBoolean(ConfigProperties.KEYCLOAK_AUTHENTICATION)) {
```
Ceiu: I think the comment went missing due to this line changing, as I'm unable to respond directly to it; but generally speaking when it comes to these types of checks, I tend to be very defensive and default to something that won't fail in a way that throws unexpected exceptions if we can avoid it. In this particular case, we're relying on the default being set outside of this object's influence which is something I tend to cover.

Not a huge deal, but in some sense a stylistic thing.

Ceiu:
The changes thus far look good. I'll take another look when the capability updating goes in.

server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -146,6 +154,9 @@ public StatusDTO status() {
         }
 
         CandlepinCapabilities caps = CandlepinCapabilities.getCapabilities();
+        if (keycloakEnabled) {
+            caps.add(CandlepinCapabilities.KEYCLOAK_AUTH_CAPBILITY);
```
Ceiu: One odd thing here is that we're updating the capability only once the status endpoint is called. This will be fine for callers, but if we ever examine it internally, it won't work as expected until that endpoint is hit. Also, we're repeatedly adding it on every iteration, which is just churn on the hashset.

Is there a place in the Keycloak module proper that would be a better fit for this?

server/src/main/java/org/candlepin/auth/KeycloakConfiguration.java
```diff
@@ -95,7 +94,7 @@ public AdapterConfig getAdapterConfig() {
      * @param httpRequest the request to attempt authentication against
      * @return an instance of {@link CandlepinKeycloakRequestAuthenticator}
      */
-    public RequestAuthenticator getRequestAuthenticator(HttpRequest httpRequest) {
+    public RequestAuthenticator createRequestAuthenticatorForRequest(HttpRequest httpRequest) {
```
Ceiu: Redundant name is redundant. :P

kahowell:
f388159..6c6157c: squash my changes

kahowell:
@Ceiu my changes are squashed. Would you like to merge this?

PR #2386, opened 2019-07-11
[F] Ported the HealEntireOrgJob to the Artemis job system (ENT-848)

abhiskum:


server/src/test/java/org/candlepin/async/tasks/HealEntireOrgJobTest.java
```diff
@@ -0,0 +1,230 @@
...
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.async.tasks;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
```
nikosmoum: We're trying to write all new unit tests (and switch any existing unit tests we touch/modify) in JUnit 5, so only using the `org.junit.jupiter` package from now on. If there are any problems like you mentioned before with the `@BeforeEach` which is the JUnit 5 equivalent of `@Before`, we can work through. I'll try to debug this problem now that I have some time.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
```diff
@@ -0,0 +1,187 @@
...
+
+        @Override
+        public void validate() throws JobConfigValidationException {
+            super.validate();
+
+            try {
+                JobArguments arguments = this.getJobArguments();
+
+                String ownerKey = arguments.getAsString(OWNER_KEY);
+
```
nikosmoum: In general, we try to not leave too many empty lines between statements, but rather group statements that are logically related together, and separate those groups with empty lines. This is obviously quite subjective, and more like a guidance that is up to you to change if you like, but too many empty lines can hurt readability.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
```diff
@@ -0,0 +1,187 @@
...
+    public Object execute(JobExecutionContext context) throws JobExecutionException {
+        try {
+
+            final JobArguments arguments = context.getJobArguments();
+            final String ownerKey = arguments.getAsString(OWNER_KEY);
+            Owner owner = ownerCurator.getByKey(ownerKey);
+            if (owner.isAutobindDisabled() || owner.isContentAccessEnabled()) {
+                String caMessage = owner.isContentAccessEnabled() ?
+                    " because of the content access mode setting" :
+                    "";
```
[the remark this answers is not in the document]
abhiskum: Done.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
```diff
@@ -0,0 +1,187 @@
...
+                // Perhaps add something to surface errors later
+                catch (Exception e) {
+                    log.debug("Healing failed for UUID \"{}\" with message: {}", uuid, e.getMessage());
+                    result.append("Healing failed for UUID: ").append(uuid).append("\n");
+                }
+            }
+            return result;
+        }
+        catch (Exception e) {
+            log.error("EntitlerJob encountered a problem.", e);
```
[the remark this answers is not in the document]
abhiskum: Copy paste from old job class. Fixed.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
```diff
@@ -0,0 +1,187 @@
...
+
+    /*
+     * Each consumer heal should be a separate transaction
+     */
+    @Transactional
+    private void healSingleConsumer(Consumer consumer, Owner owner, Date date)
+        throws AutobindDisabledForOwnerException {
+        List<Entitlement> ents = entitler.bindByProducts(AutobindData.create(consumer, owner).on(date), true);
+        entitler.sendEvents(ents);
+
```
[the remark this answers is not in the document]
abhiskum: Done.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**
+     * Job configuration object for the heal entire org job
+     */
+    public static class HealEntireOrgJobConfig extends JobConfig {
+
+        public HealEntireOrgJobConfig() {
+            this.setJobKey(JOB_KEY).setJobName(JOB_NAME)
+                .addConstraint(JobConstraints.uniqueByArguments(OWNER_KEY)); //TODO
```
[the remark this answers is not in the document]
abhiskum: I have added to confirm with you however forgot to remove same.

server/src/test/java/org/candlepin/async/tasks/HealEntireOrgJobTest.java
abhiskum: Done.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
abhiskum: Done.

server/src/test/java/org/candlepin/async/tasks/HealEntireOrgJobTest.java
```diff
@@ -0,0 +1,231 @@
...
+import org.candlepin.controller.Entitler;
+import org.candlepin.model.Consumer;
+import org.candlepin.model.ConsumerCurator;
+import org.candlepin.model.EmptyCandlepinQuery;
+import org.candlepin.model.Owner;
+import org.candlepin.model.OwnerCurator;
+import org.candlepin.test.TestUtil;
+
+import org.junit.Before;
+import org.junit.Test;
```
nikosmoum: These 2 lines should also be imported from the org.junit.jupiter package. I just tried locally with @BeforeEach instead of @Before and it worked without issues.

abhiskum: Done.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
```diff
@@ -0,0 +1,176 @@
...
+        this.ownerCurator = ownerCurator;
+        this.i18n = i18n;
+    }
+
+    @Override
+    public Object execute(JobExecutionContext context) throws JobExecutionException {
+        try {
+
+            final JobArguments arguments = context.getJobArguments();
+            final String ownerKey = arguments.getAsString(OWNER_KEY);
```
Ceiu: The final modifiers are extraneous here (and throughout the method).

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
```diff
@@ -0,0 +1,176 @@
...
+            Owner owner = ownerCurator.getByKey(ownerKey);
```
Ceiu: We should probably be checking that we actually received an owner here, and failing more gracefully (and terminally)

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
```diff
@@ -0,0 +1,176 @@
...
+            if (owner.isAutobindDisabled() || owner.isContentAccessEnabled()) {
+                String caMessage = owner.isContentAccessEnabled() ?
+                    " because of the content access mode setting" : "";
+                throw new BadRequestException(
```
Ceiu: Let's update this exception to something more suitable for the context (JobException), and ensure it's flagged as terminal so we don't attempt to retry the job.

Ceiu:
Since this just got merged as I put up my review, please post a new PR with the updates.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
abhiskum: If the owner does not exist then should we throw an exception or just simply terminate the method execution?

Ceiu: In this case, an exception is probably the way to go. I would assume that the process that fired off the job already validated the org existed at the time it was queued, so this would be a pretty nasty race condition if it happened.

Safest way to handle this is likely to throw a terminal JobException.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
abhiskum: This exception catch by outer catch block and re-through as JobException. Still, do we need to change the exception type here?

Ceiu: Probably should, as we're not in a request context here.

The rethrowing is fine-ish, we just need to be somewhat careful about whether or not the job should be retried. This particular failure is terminal, and the wrapper exception is set to non-terminal. Not a huge deal in the end; we'll just retry until the limit is hit, which will generate a bunch of log noise.

abhiskum: I am creating new PR with all these changes. Thanks for suggestions.

PR #2387, opened 2019-07-11
[F] HealEntireOrgJob review comment fix (ENT-848)

abhiskum:
@Ceiu Updated code changes as per your suggestions in https://github.com/candlepin/candlepin/pull/2386. 
Please review.

### DIFF
--- a/server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
+++ b/server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
@@ -35,7 +35,6 @@ import org.candlepin.resource.dto.AutobindData;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 
-import org.jboss.resteasy.spi.BadRequestException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xnap.commons.i18n.I18n;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
```diff
@@ -72,18 +71,23 @@ public HealEntireOrgJob(Entitler entitler, ConsumerCurator consumerCurator, Owne
...
+            if (owner == null) {
+                log.warn("Healing attempted against non-existent org id \"{}\"", ownerKey);
+                throw new JobExecutionException(
+                    i18n.tr("Healing attempted against non-existent org id \"{}\"", ownerKey));
+            }
             if (owner.isAutobindDisabled() || owner.isContentAccessEnabled()) {
                 String caMessage = owner.isContentAccessEnabled() ?
                     " because of the content access mode setting" : "";
-                throw new BadRequestException(
+                throw new JobExecutionException(
```
Ceiu: Remember to set the terminal flag to true here: 

```new JobExecutionException(i18n.tr("..."), true);```

File: server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java [hunk 2]
```diff
@@ -72,18 +71,23 @@ public HealEntireOrgJob(Entitler entitler, ConsumerCurator consumerCurator, Owne
...
 
-            final JobArguments arguments = context.getJobArguments();
-            final String ownerKey = arguments.getAsString(OWNER_KEY);
+            JobArguments arguments = context.getJobArguments();
+            String ownerKey = arguments.getAsString(OWNER_KEY);
             Owner owner = ownerCurator.getByKey(ownerKey);
+            if (owner == null) {
+                log.warn("Healing attempted against non-existent org id \"{}\"", ownerKey);
+                throw new JobExecutionException(
+                    i18n.tr("Healing attempted against non-existent org id \"{}\"", ownerKey));
```
Ceiu: Minor nit: the key and ID are different values, so this error message might be a bit confusing.

Also, need the terminal flag here as well.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
abhiskum: Done. I misunderstood your earlier comment.

server/src/main/java/org/candlepin/async/tasks/HealEntireOrgJob.java
abhiskum: Done.

Ceiu:
ack

PR #2388, opened 2019-07-12
1702383: prefer virtual pool and conusmer match over syspurpose null_…

wottop:
…rule

### DIFF
--- a/server/spec/autobind_spec.rb
+++ b/server/spec/autobind_spec.rb
@@ -724,4 +724,32 @@ describe 'Autobind On Owner' do
     status['compliantAddOns']['PROVIDED_ADDON'][0]['pool']['id'].should == pool_with_addon_only.id
     status['compliantRole']['PROVIDED_ROLE'][0]['pool']['id'].should == pool_with_role_only.id
   end
+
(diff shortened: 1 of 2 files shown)


wottop:
The weight for the null_rule_score was to much for virt_only to overcome. Reduced the null rule by a magnitude of 10.

server/spec/autobind_spec.rb
```diff
@@ -724,4 +724,31 @@
...
     status['compliantRole']['PROVIDED_ROLE'][0]['pool']['id'].should == pool_with_role_only.id
   end
+
+  it 'virtual system should bind to virt-only pool despite lacking syspurpose attribute matches' do
+    mkt_product1 = create_product(random_string('product'),
+                                  random_string('product'),
+                                  {:owner => owner_key})
+    mkt_product2 = create_product(random_string('product'),
+                                  random_string('product'),
+                                  {:attributes => {:virt_only => "true", :support_level => "provided_sla",:usage => "Production", :roles => "SP Server"},
```
nikosmoum: To make this test even more robust, I'd also add an addons attribute here e.g. `, :addons => "provided_addon"`. Even though addons are not used by production SKUs, this will boost p1's overall score more by adding another null rule score, and make the test more future-proof.

PR #2391, opened 2019-07-18
1722477: Remove storage of events and resources that access the store…

wottop:
…d events


Ceiu:
Looks fine to me. I think the only thing I would say we need to do is to repackage what's left of the event bits, chop off what we don't need from them and then drop the DTO stuff (since event itself is now a DTO).

But that's entirely tech-debt, and not something we need to do right this minute.

Ceiu:
So... I know I said it was for later, but what are everyone's thoughts on removing EventDTO and the translators and just outputting the Event object directly now?

Ceiu:
Some unused imports in the OwnerResource and ConsumerResource classes (Feed), but unless there is more event stuff hanging out somewhere that's not immediately jumping to mind, this should be good once all the tests pass.

PR #2392, opened 2019-07-18
[F] Reverted binary job constraint interface changes

Ceiu:
- The JobConstraint interface now accepts a collection of jobs for
  the existing jobs, rather than a singular job


Ceiu:
This is going to cause problems with two-way constraints eventually, but I guess we can address that if and when we need to.

Ceiu:
Ensuring support for things required/desired now vs things we *might* need in the future. The throttling constraint I'm not so worried about, but the hypervisor constraint where we want what is effectively a unique-by-arg within a limited time is another story.

Also, this was the first version of the design/PoC; so a lot of it was just copy/paste, cleanup and fixing of tests. Not a ton of work.

PR #2397, opened 2019-07-25
Cleanup of async job tests

Januson:
A lot of async job tests inherit from either BaseJobTest of DatabaseTestFixture even though only a few of them really require it. Jobs that require DatabaseTestFixture are ActiveEntitlementJobTest, ImportRecordCleanerJobTest, OrphanCleanupJobTest and UndoImportsJobTest. For the rest I removed the parent. With parent removed @MockitoSettings(strictness = Strictness.LENIENT) is redundant as unused mocking is no longer present. I also removed @ExtendWith(MockitoExtension.class) where tests do not use Mockito annotations.

- Replace try-catch with assertThrows
- Replace JUnit4 annotations with JUnit5
- Replace deprecated matchers
- Remove redundant annotations
- Remove redundant parents

### DIFF
--- a/server/src/test/java/org/candlepin/async/JobManagerTest.java
+++ b/server/src/test/java/org/candlepin/async/JobManagerTest.java
@@ -87,8 +87,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
-
-
 public class JobManagerTest {
 
     private static class StateCollectingStatus extends AsyncJobStatus {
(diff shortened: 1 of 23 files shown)


Ceiu:
UndoImportsJobTest still needs the lenient setting, or other changes to make it good.

Januson:
@Ceiu You are right. This was a mistake. I put this lenient back.

Ceiu:
ack

PR #2398, opened 2019-07-26
[ENT-288] Need a better way to identify pools as being 'custom'

wolfdale:
CUSTOM Pool Type

### DIFF
--- a/server/src/main/java/org/candlepin/resource/OwnerResource.java
+++ b/server/src/main/java/org/candlepin/resource/OwnerResource.java
@@ -1703,7 +1703,7 @@ public class OwnerResource {
         pool.setUpstreamPoolId(inputPoolDTO.getUpstreamPoolId());
         pool.setSubscriptionSubKey(inputPoolDTO.getSubscriptionSubKey());
         pool.setSubscriptionId(inputPoolDTO.getSubscriptionId());
-        pool.setAttribute(Pool.Attributes.CUSTOM_POOL, "true");
+        pool.setLocked(false);
 
         if (inputPoolDTO.getProductId() == null || inputPoolDTO.getProductId().isEmpty()) {
             throw new BadRequestException(i18n.tr("Pool product ID not specified"));
--- a/server/src/main/java/org/candlepin/resource/OwnerResource.java
+++ b/server/src/main/java/org/candlepin/resource/OwnerResource.java
@@ -1703,7 +1703,7 @@ public class OwnerResource {
         pool.setUpstreamPoolId(inputPoolDTO.getUpstreamPoolId());
         pool.setSubscriptionSubKey(inputPoolDTO.getSubscriptionSubKey());
         pool.setSubscriptionId(inputPoolDTO.getSubscriptionId());
-        pool.setAttribute(Pool.Attributes.CUSTOM_POOL,"true");
+        pool.setAttribute(Pool.Attributes.CUSTOM_POOL, "true");
 
         if (inputPoolDTO.getProductId() == null || inputPoolDTO.getProductId().isEmpty()) {
             throw new BadRequestException(i18n.tr("Pool product ID not specified"));
--- a/server/src/main/java/org/candlepin/resource/OwnerResource.java
+++ b/server/src/main/java/org/candlepin/resource/OwnerResource.java
@@ -1703,7 +1703,6 @@ public class OwnerResource {
         pool.setUpstreamPoolId(inputPoolDTO.getUpstreamPoolId());
         pool.setSubscriptionSubKey(inputPoolDTO.getSubscriptionSubKey());
         pool.setSubscriptionId(inputPoolDTO.getSubscriptionId());
-        pool.setLocked(false);
 
         if (inputPoolDTO.getProductId() == null || inputPoolDTO.getProductId().isEmpty()) {
             throw new BadRequestException(i18n.tr("Pool product ID not specified"));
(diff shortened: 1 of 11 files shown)


Ceiu:
This approach combines the ideas of using the pool type *and* using pool attributes, which is not the direction we should go here.

As you can probably see from the getPoolType method, there is nothing that would logically prevent a custom pool from being a derived pool or any other such type of pool -- none of them are really mutually exclusive (there's a fair argument to be made about the validity of the pool type as it exists today, but that's another topic entirely).  
For attributes, those are often used for flags coming from any of our three sources of data that are often then passed down to clients via Candlepin APIs and/or Subscription Manager.

The "custom" flag desired here is both a means to identify custom pools, a means to filter pool lookups, and a means to lock pools from being modified at the API level. Some of these functions will come in future cards/work rather than this PR, but this is the intent behind the flag. As such, it should closer mimic the "locked" flag on Products and Content. Given that we will be querying against it, and we don't need to be passing it down to clients (at the time of writing) or receiving it from upstream sources, a database column is probably going to be best here.

Again, this can largely be lifted from the "locked" functionality on products/content. If we want to maintain consistency with that naming convention, then a "locked" pool is logically equivalent to a non-"custom" pool. If we opt to call it "custom" then it is the inverse of locked -- set only at the API level.

Personally, I'm a fan of the locked method, as we already have full control over the path an upstream pool would come in, whereas "custom" needs to cover every API method that could result in a created pool. But, ultimately that comes down to which the implementer finds easiest relative to the values of being consistent and clear in naming/usage.

server/src/main/resources/db/changelog/20190729122218-add-locked-column-to-cp-pool.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20190729122218" author="prakgupt">
```
abhiskum: @wolfdale we need to append "-1" at the end of changeset id.

abhiskum:
These changes are fine for custom pools creation through API. However, the locked field should be set as true for non-custom pools. We should update non-custom pool creation logic to set this flag as true. Also, do we need to think of existing pools as part of this PR?

wolfdale:
`"locked (true)" pool is logically equivalent to a "non-custom" pool.
"un-locked (false)" pool is logically equivalent to a "custom" pool.`

@abhiskum  Currently As default all pools will be CUSTOM (Default value of Boolean valued column). which means every time a pool is instantiated it will be tagged as CUSTOM pool. 
Not sure if I make Boolean default as true (non-custom pool) or explicitly set values in code itself?
Do we require a database migration for tagging existing pools in production or stage ?

server/src/main/resources/db/changelog/20190729122218-add-locked-column-to-cp-pool.xml
```diff
@@ -0,0 +1,18 @@
...
+        <comment>Adding column locked to identify custom and non-custom pools.
+        </comment>
+
+        <addColumn tableName="cp_pool">
+            <column name="locked" type="int"/>
+        </addColumn>
+    </changeSet>
+
+</databaseChangeLog>
```
Ceiu: Add a line break to the end here

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1703,7 +1703,7 @@ public PoolDTO createPool(@PathParam("owner_key") @Verify(Owner.class) String ow
         pool.setUpstreamPoolId(inputPoolDTO.getUpstreamPoolId());
         pool.setSubscriptionSubKey(inputPoolDTO.getSubscriptionSubKey());
         pool.setSubscriptionId(inputPoolDTO.getSubscriptionId());
-        pool.setAttribute(Pool.Attributes.CUSTOM_POOL, "true");
+        pool.setLocked(false);
```
Ceiu: Since the default is false, this will basically no-op in the majority of cases. What we should do is, instead, set the locked flag to true during refresh (which also covers manifest import), as that's where non-custom pools are coming in anyway.

Ceiu:
If we leave everything as-is and then only lock pools that come in from manifest import and refresh, we don't need to worry about doing a migration (which itself may cause problems as our detection today isn't perfect). Things will stay exactly as they are now and slowly get corrected over time with the proper changes to refresh/import.

server/src/main/resources/db/changelog/20190729122218-add-locked-column-to-cp-pool.xml
wolfdale: Added.

server/src/main/resources/db/changelog/20190729122218-add-locked-column-to-cp-pool.xml
wolfdale: Fixed.

server/src/main/java/org/candlepin/resource/OwnerResource.java
wolfdale: Agreed.

Locked pools will come into play during Import manifest and Scheduled pool Job Flow.
Both involve creation of master pool that will be a common point to set the flag.
This will be taken care of in next commit.

wolfdale:
Agreed. Added/Modified unit test case in `PoolTranslatorTest` to verify that flag is being correctly copied to the DTO.

PR #2399, opened 2019-07-30
Handling for bad hypervisor json

wottop:
It previously resulted in a null pointer exception that gave
 no indication of the problem. The log now has an appropriate
 message.

### DIFF
--- a/server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
+++ b/server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
@@ -161,6 +161,9 @@ public class HypervisorUpdateJob extends KingpinJob {
             }
 
             final HypervisorList hypervisors = parsedHypervisors(map);
+            if (hypervisors.getHypervisors() == null) {
+                throw new Exception("The json report does not contain the proper structure.");
+            }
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -161,6 +161,9 @@ public void toExecute(final JobExecutionContext context) throws JobExecutionExce
             }
 
             final HypervisorList hypervisors = parsedHypervisors(map);
+            if (hypervisors.getHypervisors() == null) {
+                throw new Exception("The json report does not contain the proper structure.");
```
Ceiu: Do we want a more explicit exception type here? Also, is this something we should check and complain about at the resource level so we can throw a BadRequestException and not even schedule the job?

wottop: The problem with that is we would then parse the json synchronously. This would add a bunch of processing time on the front end to every request when the majority of them are fine.

Ceiu: Hrmm. Maybe something for the backlog; I feel we should be able to engineer something that does a basic syntax check on the tree without diving into everything in depth and killing the runtime of every request.

PR #2400, opened 2019-07-31
ENT-1408: Candlepin RHEL 8 support

nikosmoum:
- Now the candlepin rpm depends on tomcatjss instead of
  jss and tomcat/pki-servlet-container/pki-servlet-engine
  to avoid these name changes between RHEL 7, RHEL8 and
  RHEL 8.1, since that package is comprehensively taking
  care of that itself.
- The spec file and deploy script now update the
  META-INF/context.xml file to avoid breaking API changes
  between Tomcat 7 and 8+ when altering the classpath to
  include the jss jar from the filesystem.
- Updated cpsetup script:
    * to be python3 compatible.
    * with server.xml and CA cert/key changes for Tomcat 8+
      compatibility.
- JSSProvider/CryptoManager are now loaded once, during
  context initialization, and not during classloading of
  JSSProviderLoader.java.
- Due to the fact that we no longer package the jss jar,
  but we load it from the OS, and because of compile-
  breaking changes between jss version 4.4.X and 4.5.0,
  we now use dynamic classloading and reflection to
  initialize the CryptoManager.
- Added the *.fedorapeople.org repositories to parent
  pom to be able to do a maven build without having to
  download the artifacts they hold with buildr in the
  local .m2/repository.

### DIFF
--- a/server/src/main/webapp/META-INF/context_tomcat8.xml
+++ b/server/src/main/webapp/META-INF/context_tomcat8.xml
@@ -1,0 +1,11 @@
+<!-- context.xml file compatible from Tomcat version 8 and up -->
+<Context path="/candlepin">
+    <!-- Candlepin needs symlink support to read its jars. Also, we want to load the jss jar from the
+    filesystem, instead of having a hard dependency on a specific version. -->
+    <Resources allowLinking="true">
+        <JarResources className="org.apache.catalina.webresources.FileResourceSet"
+            base="/usr/lib/java/jss4.jar"
+            webAppMount="/WEB-INF/lib/jss4.jar">
+        </JarResources>
+    </Resources>
+</Context>
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,16 @@
                     <name>OAuth</name>
                     <url>http://oauth.googlecode.com/svn/code/maven/</url>
                 </repository>
+                <repository>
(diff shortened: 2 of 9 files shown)


nikosmoum:
This has been tested with:
- tito rpm build & install on RHEL 7.6, RHEL7.7 Beta, RHEL8.0, RHEL 8.1 Beta, Fedora 30.
- git checkout & ./server/bin/deploy on Fedora 30 and Centos 7.

server/src/main/webapp/META-INF/context_tomcat8.xml
```diff
@@ -0,0 +1,11 @@
+<!-- context.xml file compatible from Tomcat version 8 and up -->
+<Context path="/candlepin">
+    <!-- Candlepin needs symlink support to read its jars. Also, we want to load the jss jar from the
+    filesystem, instead of having a hard dependency on a specific version. -->
+    <Resources allowLinking="true">
+        <JarResources className="org.apache.catalina.webresources.FileResourceSet"
+            base="/usr/lib/java/jss4.jar"
+            webAppMount="/WEB-INF/lib/jss4.jar">
```
Ceiu: This covers the jss4 jar, but what about the rest of the file paths? Are they unnecessary in 8/9+?

Ceiu:
ack

PR #2401, opened 2019-08-01
[ENT-1530] 1732892: Registration messaging changed to "owner with key xxx no found"

wolfdale:
- Exception String change to be more descriptive


Ceiu:
NACK

This was changed several years ago to avoid exposing data about objects the user is not able to see. This change is one of those ones that should be CLOSED WONTFIX, as this was done with explicit intent.

wolfdale:
- Closing Pull Request.

PR #2407, opened 2019-08-13
1734521: Consumer returning duplicate guests [ENT-1535]

abhiskum:
The API returns duplicate guest ids in the response as the cp_consumer_guests table has duplicate guest entries for a consumer. As per my investigation, this situation occurs when update consumer request contains duplicate guest ids. Currently, there is no check to remove duplicate guest ids from the request.

To fix this, I removed duplicate guests from the response for existing records and also added code to remove duplicate entries from request to avoid duplicate entries in cp_consumer_guests table.

### DIFF
--- a/server/spec/consumer_resource_host_guest_spec.rb
+++ b/server/spec/consumer_resource_host_guest_spec.rb
@@ -195,6 +195,30 @@ describe 'Consumer Resource Host/Guest' do
     host2['uuid'].should == host_consumer2['uuid']
   end
 
+  it 'should ignore duplicate guest ids' do
+    uuid1 = random_string('system.uuid')
+    uuid2 = random_string('system.uuid')
+    guests = [{'guestId' => uuid1}, {'guestId' => uuid1}, {'guestId' => uuid2},  {'guestId' => uuid2}]
+
+    user_cp = user_client(@owner1, random_string('test-user'))
+    host_consumer = user_cp.register(random_string('host'), :system, nil,
+      {}, nil, nil, [], [])
+    guest_consumer1 = user_cp.register(random_string('guest'), :system, nil,
+      {'virt.uuid' => uuid1}, nil, nil, [], [])
+    guest_consumer2 = user_cp.register(random_string('guest'), :system, nil,
+      {'virt.uuid' => uuid2}, nil, nil, [], [])
+
+    consumer_client = Candlepin.new(nil, nil, host_consumer['idCert']['cert'], host_consumer['idCert']['key'])
+    consumer_client.update_consumer({:guestIds => guests})
+
+    guest_consumers = @cp.get_consumer_guests(host_consumer['uuid'])
+    expect(guest_consumers.length).to eq(2)
+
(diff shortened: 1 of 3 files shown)


barnabycourt:
Please add a spec test to ensure this doesn't come back in the future.

abhiskum:
@barnabycourt @Ceiu  Added spec test for duplicate guest ids.

server/spec/consumer_resource_host_guest_spec.rb
```diff
@@ -195,6 +195,25 @@
...
+      {}, nil, nil, [], [])
+    user_cp.register(random_string('guest'), :system, nil,
+      {'virt.uuid' => uuid1}, nil, nil, [], [])
+    user_cp.register(random_string('guest'), :system, nil,
+      {'virt.uuid' => uuid2}, nil, nil, [], [])
+
+    consumer_client = Candlepin.new(nil, nil, host_consumer['idCert']['cert'], host_consumer['idCert']['key'])
+    consumer_client.update_consumer({:guestIds => guests})
+
+    @cp.get_consumer_guests(host_consumer['uuid']).length.should == 2
```
Ceiu: Should also verify the UUIDs received are the two expected, remembering that the order isn't guaranteed.

abhiskum: Nice suggestion. I have added a check for UUIDs. 

Regarding commit message, Sure, I will try to add details about the fix.

server/spec/consumer_resource_host_guest_spec.rb
```diff
@@ -195,6 +195,28 @@
...
+      {'virt.uuid' => uuid1}, nil, nil, [], [])
+    guest_consumer2 = user_cp.register(random_string('guest'), :system, nil,
+      {'virt.uuid' => uuid2}, nil, nil, [], [])
+
+    consumer_client = Candlepin.new(nil, nil, host_consumer['idCert']['cert'], host_consumer['idCert']['key'])
+    consumer_client.update_consumer({:guestIds => guests})
+
+    guest_consumers = @cp.get_consumer_guests(host_consumer['uuid'])
+    guest_consumers.length.should == 2
+    [guest_consumers[0]['uuid'], guest_consumers[1]['uuid']].should =~ [guest_consumer1['uuid'],
```
Ceiu: This is clever, but a little obscure. I had to lookup that the old .should operation changes what =~ does with lists. While the obscure bit is usually not a problem for me, we're trying to move away from the injected ```.should``` tests in favor for the updated and more explicit ```expect(...).to``` syntax. With my limited Ruby knowledge, that ends up looking like this:

```
    guest_consumers = @cp.get_consumer_guests(host_consumer['uuid'])
    expect(guest_consumers.length).to eq(2)

    guest_uuids = guest_consumers.map { |guest| guest['uuid'] }
    expect(guest_uuids).to include(guest_consumer1['uuid'], guest_consumer2['uuid'])
```

abhiskum: @Ceiu Done.

PR #2408, opened 2019-08-13
[M] Moved testing HSQLDB instance to a file rather than memory

Ceiu:
- The HSQLDB instance used for general unit tests has moved from
  the in-memory DB, to a file-system DB in the system temp
  directory (/tmp)
- LiquibaseExtension now uses IF EXISTS when executing a
  DROP SCHEMA command
- LiquibaseExtension now drops any existing known/expected
  schema immediately after connecting to the DB, before
  attempting to create any new schemas
- LiquibaseCustomTask now implements Closeable and closing
  a custom task will forcefully close any open prepared statements
  created through the preparedStatement or executeQuery methods
- LiquibaseCustomTaskWrapper now uses the new .close method
  after executing any custom task to ensure opened statements
  are explicitly closed
- LiquibaseExtension now uses a direct Statement object and
  explicitly closes it, rather than offloading this behavior
  to the LiquibaseStatement object

### DIFF
--- a/server/src/main/java/org/candlepin/liquibase/LiquibaseCustomTaskWrapper.java
+++ b/server/src/main/java/org/candlepin/liquibase/LiquibaseCustomTaskWrapper.java
@@ -21,6 +21,8 @@ import liquibase.exception.SetupException;
 import liquibase.exception.ValidationErrors;
 import liquibase.resource.ResourceAccessor;
 
+import java.lang.reflect.InvocationTargetException;
+
 
 
 /**
@@ -40,6 +42,26 @@ public abstract class LiquibaseCustomTaskWrapper<T extends LiquibaseCustomTask> 
         }
 
         this.typeClass = typeClass;
+    }
+
+    /**
+     * Instantiates the actual task instance, providing the specified database and logger to the
+     * new instance.
+     *
+     * @param database
+     *  the database instance to provide to the new task
+     *
+     * @param logger
+     *  the custom task logger instance to provide to the new task
+     *
+     * @return
+     *  the new instance of the liquibase task
+     */
+    private T instantiateTask(Database database, CustomTaskLogger logger) throws NoSuchMethodException,
+        InstantiationException, IllegalAccessException, InvocationTargetException {
+
+        return this.typeClass.getConstructor(Database.class, CustomTaskLogger.class)
+            .newInstance(database, logger);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/candlepin/liquibase/LiquibaseCustomTaskWrapper.java
```diff
@@ -69,6 +69,7 @@ public void execute(Database database) throws CustomChangeException {
                 .newInstance(database, new LiquibaseCustomTaskLogger());
 
             task.execute();
+            task.close();
```
awood: Maybe put this in a `finally` block? Or you could use the new `try-with-resources` syntax.

server/src/test/java/org/candlepin/junit/LiquibaseExtension.java
```diff
@@ -99,32 +103,34 @@ public void runUpdate() throws LiquibaseException {
...
 
-    private void exec(String sql) {
-        SqlStatement s = new RawSqlStatement(sql);
+    private void executeUpdate(String sql) {
         try {
-            database.execute(new SqlStatement[] { s }, Collections.<SqlVisitor>emptyList());
+            Statement statement = this.connection.createStatement();
+            statement.executeUpdate(sql);
+
+            statement.close();
```
awood: `finally` here?

Ceiu:
Yes, surprisingly, HSQLDB will create the path necessary to drop its files, assuming it has permission to do so.

awood:
> Yes, surprisingly, HSQLDB will create the path necessary to drop its files, assuming it has permission to do so.

Does it do the delete too?  I mean it doesn't matter if there's a zero byte file sitting around in /tmp, but I'm just curious.

Ceiu:
Nope, no deletion. That's why it tries to clear out the DB right after connecting to it now. It ends up being around 200mb when all the tests complete.

I spent a bit of time trying to do some stuff like having a macro in the path in our persistence.xml URL (${DIR}) and manually creating a temp directory within Java that *would* be cleaned up, but persistence.xml is loaded multiple times, at least once by HSQLDB outside of what we have immediate control over and it interprets the macro literally, which leads to all sorts of problems.

Ceiu:
Code corrections made, but I have yet to find a solution that addresses the growing size of the lobs file. It *does* get overwritten by subsequent runs of the test suite, so it doesn't keep dropping multiple of these files.

awood:
Looks good to me.  The last thing to check is just to make sure that using a static filename isn't going to cause chaos on the Jenkins machines.

Ceiu:
I've added a fix that addresses the static filename issue *and* the persisting database files. The directory is now randomly generated per test suite, and the files are deleted after each test suite completes.

PR #2409, opened 2019-08-13
[M] 1740877: Added principal to consumer deletion records (ENT-1572)

Ceiu:
- DeletedConsumer now has a column to store the name of the principal
  that triggered the deletion of a consumer
- ConsumerCurator.delete now sets the name of the current principal
  on the deleted consumer record it creates


wottop:
Looks good. Waiting for jenkins to complete.

wottop:
ack

PR #2412, opened 2019-08-14
[M] Fixed incorrect wording in the deleted consumer changeset

Ceiu:
- Updated the comment in the deleted consumer changeset that adds
  a principal to the table


wottop:
ack

PR #2416, opened 2019-08-19
1466415 [RFE] Candlepin should allow setting of default auto-attach value for hypervisors

ShwethaKalyanaraman:
- Added a new column called autobindHyperVisorDisabled with type boolean in Owner Class
- Made changes in a condition statement for Entitler class to check if autobindHyperVisor is disabled
- Rest of the changes include changes for the new parameter and fixing tests to support new parameter

### DIFF
--- a/server/spec/autobind_disabled_for_owner_spec.rb
+++ b/server/spec/autobind_disabled_for_owner_spec.rb
@@ -9,6 +9,7 @@ describe 'Autobind Disabled On Owner' do
   before(:each) do
     owner = create_owner random_string('test_owner1')
     owner['autobindDisabled'] = true
+    owner['autobindHypervisorDisabled'] = true
     @cp.update_owner(owner['key'], owner)
     @owner = @cp.get_owner(owner['key'])
     @owner.should_not be_nil
--- a/server/src/main/java/org/candlepin/controller/Entitler.java
+++ b/server/src/main/java/org/candlepin/controller/Entitler.java
@@ -198,7 +198,8 @@ public class Entitler {
         Consumer consumer = data.getConsumer();
         Owner owner = data.getOwner();
 
-        if ((!consumer.isDev() && owner.isAutobindDisabled()) || owner.isContentAccessEnabled()) {
+        if ((!consumer.isDev() && owner.isAutobindDisabled()) &&
+            owner.isAutobindHypervisorDisabled() || owner.isContentAccessEnabled()) {
             String caMessage = owner.isContentAccessEnabled() ?
                 " because of the content access mode setting" : "";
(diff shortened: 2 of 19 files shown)


server/spec/autobind_disabled_for_owner_spec.rb
```diff
@@ -9,6 +9,7 @@
   before(:each) do
     owner = create_owner random_string('test_owner1')
     owner['autobindDisabled'] = true
+    owner['autobindHyperVisorDisabled'] = true
```
Ceiu: Hypervisor is one word; the V should be lowercase here, and in all other instances as well.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -199,7 +199,8 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
         Consumer consumer = data.getConsumer();
         Owner owner = data.getOwner();
 
-        if ((!consumer.isDev() && owner.isAutobindDisabled()) || owner.isContentAccessEnabled()) {
+        if ((!consumer.isDev() && owner.isAutobindDisabled()) &&
+            owner.isAutobindHyperVisorDisabled() || owner.isContentAccessEnabled()) {
```
Ceiu: This will catch all binds, not just those for hypervisors. We'll need to be more thorough in this check.

server/src/main/java/org/candlepin/keycloak-6.0.1/standalone/data/keycloak.lock.db
```diff
@@ -0,0 +1,6 @@
+#FileLock
```
Ceiu: Unnecessary file for this PR

Ceiu:
Also, do you happen to have the JIRA number associated with this task?

server/spec/autobind_disabled_for_owner_spec.rb
ShwethaKalyanaraman: Ok will make that change

server/src/main/java/org/candlepin/controller/Entitler.java
ShwethaKalyanaraman: Any suggestions to make it better?

ShwethaKalyanaraman:
> Also, do you happen to have the JIRA number associated with this task?

[Bug 1466415]- This is the only number I have

Ceiu:
Changes seem fine (though that if statement for the check is starting to get a bit unwieldy). Once the unit tests are fixed, this should be good to merge.

wottop:
I am working on this to complete the task.

wottop:
Replaced by 
https://github.com/candlepin/candlepin/pull/2475

PR #2417, opened 2019-08-20
[F]  Updated JobManager's scoping capabilities (ENT-1562)

Ceiu:
- Owner is no longer strongly referenced by the jobs table
- OwnerManager no longer performs any job cleanup on owner deletion
- JobManager no longer uses UnitOfWork when setting up the job
  environment
- JobManager now performs better transaction isolation and tracking
  during job initialization and execution
- JobManager will now attempt to automatically commit or rollback an
  open database transaction after job execution completes or fails
- JobManager now properly saves and restores MDC state before/after
  job execution
- Cancelled jobs no longer cause a JobInitializationException when
  its respective message is received
- AsyncJobStatus is now versioned and uses optimistic locking to
  prevent update collisions

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -31,12 +31,12 @@ import org.candlepin.model.AsyncJobStatusCurator;
 import org.candlepin.model.CandlepinModeChange;
 import org.candlepin.model.CandlepinModeChange.Mode;
 import org.candlepin.model.Owner;
+import org.candlepin.model.OwnerCurator;
 import org.candlepin.util.Util;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 import com.google.inject.Injector;
 import com.google.inject.persist.Transactional;
-import com.google.inject.persist.UnitOfWork;
 
 import org.hibernate.Session;
 import org.hibernate.Transaction;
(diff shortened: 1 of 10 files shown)


Ceiu:
This PR is blocked by #2404 and cannot be merged until that one is merged and this branch/PR is rebased against the feature branch. This PR should remain a draft until after the blocking PR is merged.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1440,6 +1457,96 @@ private AsyncJobStatus processJobFailure(AsyncJobStatus status, EventSink eventS
...
+     *  the updated job status associated with the canceled job, or null if no such job status
+     *  could be found
+     */
+    @Transactional
+    public synchronized AsyncJobStatus cancelJob(String jobId) {
+        if (jobId == null || jobId.isEmpty()) {
+            throw new IllegalArgumentException("jobId is null or empty");
+        }
+
+        AsyncJobStatus status = this.jobCurator.lockAndLoad(jobId);
```
barnabycourt: Do we need a lock? Could we use an optimistic "update where" & react appropriately if the update fails to update any rows?

Ceiu:
Looks like there's a problem with the versioning/optimistic locking in some cases. I'm not sure where it's going sideways, but it seems like something is doing an update to the job status silently somewhere.

PR #2420, opened 2019-08-23
[ENT-1478] Investigate & implement job constraint for EntitlerJob

Januson:
- Remove the throttling of EntitlerJob as it is no longer necessary
 - Remove scheduleJob/isSchedulable. Without custom constraint they are not needed


barnabycourt:
@Januson What are the use cases that this constraint is supporting? Do we have specific needs from customers, satellite or rhsm that are being addressed here?

PR #2422, opened 2019-08-28
[ENT-1478] Investigate & implement job constraint for EntitlerJob

Januson:
- Add throttling for the EntitlerJob

### DIFF
--- a/server/src/main/java/org/candlepin/async/tasks/EntitlerJob.java
+++ b/server/src/main/java/org/candlepin/async/tasks/EntitlerJob.java
@@ -118,8 +118,8 @@ public class EntitlerJob implements AsyncJob {
         return poolErrors;
     }
 
-    public static EntitlerJobConfig createConfig() {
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/candlepin/async/tasks/EntitlerJob.java
```diff
@@ -118,20 +118,19 @@ public Object execute(final JobExecutionContext context) throws JobExecutionExce
...
+        return new EntitlerJobConfig(limit);
     }
 
     /**
      * Job configuration object for the entitle by pool job
      */
     public static class EntitlerJobConfig extends JobConfig {
 
-        public EntitlerJobConfig() {
+        public EntitlerJobConfig(int limit) {
```
Ceiu: This is fine for now since we don't have a convenient means of updating the limit such that we can safely add it as another chained method; but this is something we should probably change in future work.

Ceiu:
ack

PR #2424, opened 2019-08-30
Make creating of test repositories more bullet proof

jirihnidek:
* When it is not possible to create or sign dummy RPM package,
  then deploy script is terminated
* When creating or signing of RPm is terminated, then some
  error is printed
* WIP: signing of RPM is executed with rebose=True option to see
  more details


wottop:
ack

PR #2425, opened 2019-08-30
[M] Added the @Verify annotation to ConsumerResource.consumerExists

Ceiu:
- Added the @Verify annotation to ConsumerResource.consumerExists
  to allow consumers to check for self-existence on a given
  Candlepin node without needing to fetch and serialize the entire
  Consumer object via ConsumerResource.getConsumer


blagae:
Thanks for doing this @Ceiu !

PR #2426, opened 2019-09-03
[ENT-1140] 1667952: candlepin refuses to start when having too many messages in ActiveMQ journal

wolfdale:
- Implement Artemis Queue Monitoring (Data Point - Pending Msg, Msg ID)
- Exception Handling when QPID throw RESOURCE LIMIT EXCEEDED


wolfdale:
This is not needed anymore.

PR #2427, opened 2019-09-03
Restricting Locked Pool Scope

wolfdale:
- Restricting Locked Pool Scope

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -730,8 +730,15 @@ public class CandlepinPoolManager implements PoolManager {
         List<Pool> pools = poolRules.createAndEnrichPools(pool, existingPools);
         log.debug("Creating {} pools: ", pools.size());
 
+        boolean isPoolLocked = pool.isLocked();
         for (Pool p : pools) {
             createPool(p);
+        }
+
+        for (Pool tempPool : pools) {
+            if (isPoolLocked) {
+                tempPool.setLocked(true);
+            }
         }
 
         return pool;
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -864,7 +864,6 @@ public class CandlepinPoolManager implements PoolManager {
         pool.setUpstreamPoolId(sub.getUpstreamPoolId());
         pool.setUpstreamEntitlementId(sub.getUpstreamEntitlementId());
         pool.setUpstreamConsumerId(sub.getUpstreamConsumerId());
-        pool.setLocked(true);
 
         // Resolve CDN
         if (sub.getCdn() != null) {
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -343,7 +343,7 @@ public class CandlepinPoolManager implements PoolManager {
 
         // These don't all necessarily belong to this owner
         List<Pool> subscriptionPools;
-
+        pool.setLocked(true);
         if (pool.getSubscriptionId() != null) {
             subscriptionPools = this.poolCurator.getPoolsBySubscriptionId(pool.getSubscriptionId()).list();
         }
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -343,7 +343,7 @@ void refreshPoolsForMasterPool(Pool pool, boolean updateStackDerived, boolean la
 
         // These don't all necessarily belong to this owner
         List<Pool> subscriptionPools;
-
+        pool.setLocked(true);
```
Ceiu: Try not to interlace logical blocks operations. This is placed right in the middle of the declaration for the sub pools and the code which populates it with no whitespace to separate it visually.

wolfdale: yes sure.

wolfdale:
@Ceiu Should we also lock the floating pools and pool that are being deleted for absent subscription in `refreshPoolsWithRegeneration` ?

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -726,10 +730,17 @@ public Pool createAndEnrichPools(Pool pool, List<Pool> existingPools) {
         List<Pool> pools = poolRules.createAndEnrichPools(pool, existingPools);
         log.debug("Creating {} pools: ", pools.size());
 
+        boolean isPoolLocked = pool.isLocked();
         for (Pool p : pools) {
             createPool(p);
         }
 
+        for (Pool tempPool : pools) {
+            if (isPoolLocked) {
```
Ceiu: You can skip the declaration above and just query the value here. Since the locked flag is a simple boolean that is returned from the Pool object, the call gets optimized down to basically what you're doing here without needing to write it out yourself.

There are cases where you'll want to explicitly cache stuff locally, but since we're not dealing with unboxing and it's a simple accessor on what is little more than a data store, this is not such a case.

server/src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -155,6 +155,7 @@ public void run() {
              * dirty, they will never get regenerated
              */
             Pool masterPool = poolManager.convertToMasterPool(subscription);
+            masterPool.setLocked(true);
```
Ceiu: Let the CPM handle this; the fewer places this is done explicitly, the better.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
wolfdale: Sure alright.

server/src/main/java/org/candlepin/controller/Refresher.java
wolfdale: Removing it.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -382,6 +383,9 @@ void refreshPoolsForMasterPool(Pool pool, boolean updateStackDerived, boolean la
                     subPool.setDerivedProduct(update);
                 }
             }
+            if (pool.isLocked()) {
```
Ceiu: Whitespace between this block and the previous block

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -730,6 +734,12 @@ public Pool createAndEnrichPools(Pool pool, List<Pool> existingPools) {
             createPool(p);
         }
 
+        for (Pool tempPool : pools) {
+            if (pool.isLocked()) {
```
Ceiu: This logic looks like it'll be fine, but we're at risk of losing the update to the locked state and we're doing a second iteration through the pools.

Conveniently, if we look through ```poolRules.createAndEnrichPools```, we get to a method called ```clonePool``` in ```PoolHelper```. I know I said before "the fewer places we do the locking the better," ```clonePool``` is one such place where we could use it. It also eliminates needing to set the locked flag here, which is somewhat random relative to the name of the method.

PoolHelper.java:255 right before the ```// Copy upstream fields``` comment would be a good place for the check + set logic used here.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
wolfdale: Fixed.

PR #2428, opened 2019-09-03
[M] 1742806: When deleting pools, delete entitlements using ids; ENT-1583

nikosmoum:
- Now, when deleting pools, the revocation of entitlements of those
  pools will use a delete query instead of a hibernate batch detach,
  in order to avoid memory overuse.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -2268,7 +2268,9 @@ public class CandlepinPoolManager implements PoolManager {
             // Revoke/delete entitlements
             if (!entitlements.isEmpty()) {
                 log.info("Revoking {} entitlements...", entitlements.size());
-                this.entitlementCurator.batchDelete(entitlements);
+                this.entitlementCurator.unlinkEntitlements(entitlements);
+                this.entitlementCertificateCurator.deleteByEntitlementIds(entitlementIds);
+                this.entitlementCurator.batchDeleteByIds(entitlementIds);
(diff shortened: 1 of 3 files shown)


nikosmoum:
Notes: Testing was performed using a mysql data dump and running the ExpiredPoolsJob job.

Previously, the job would fail with an OOM error, eating through 11GB of heap during the revoking of 37k entitlements, due to hibernate requiring enormous amounts of memory for the loaded objects.

With this change, the same action (revoking 37k entitlements), gets performed with the memory reaching around ~1.7GB, and the job overall finishes having reached a maximum utilization of 3.2GB of memory.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2268,7 +2268,8 @@ public void deletePools(Collection<Pool> pools, Collection<String> alreadyDelete
             // Revoke/delete entitlements
             if (!entitlements.isEmpty()) {
                 log.info("Revoking {} entitlements...", entitlements.size());
-                this.entitlementCurator.batchDelete(entitlements);
+                this.entitlementCertificateCurator.deleteByEntitlementIds(entitlementIds);
+                this.entitlementCurator.batchDeleteByIds(entitlementIds);
```
Ceiu: See the comment in #2429 for more details

Has there been any additional testing around skipping EntitlementCurator's extra behavior on delete?

nikosmoum: The extra runtime consistency code was added back.

PR #2429, opened 2019-09-03
[2.7] 1748406: When deleting pools, delete entitlements using ids; ENT-1583

nikosmoum:
- Now, when deleting pools, the revocation of entitlements of those
  pools will use a delete query instead of a hibernate batch detach,
  in order to avoid memory overuse.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -2268,7 +2268,9 @@ public class CandlepinPoolManager implements PoolManager {
             // Revoke/delete entitlements
             if (!entitlements.isEmpty()) {
                 log.info("Revoking {} entitlements...", entitlements.size());
-                this.entitlementCurator.batchDelete(entitlements);
+                this.entitlementCurator.unlinkEntitlements(entitlements);
+                this.entitlementCertificateCurator.deleteByEntitlementIds(entitlementIds);
+                this.entitlementCurator.batchDeleteByIds(entitlementIds);
                 this.entitlementCurator.flush();
                 this.entitlementCurator.batchDetach(entitlements);
                 log.info("Entitlements successfully revoked");
--- a/server/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/server/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -788,6 +788,43 @@ public class EntitlementCurator extends AbstractHibernateCurator<Entitlement> {
         }
     }
 
+    /**
+     * Deletes the entitlements with the given collection of ids, using a SQL delete.
+     *
+     * WARNING: This method does not maintain runtime consistency for any of the entitlements it is deleting,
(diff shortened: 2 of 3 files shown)


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2268,7 +2268,8 @@ public void deletePools(Collection<Pool> pools, Collection<String> alreadyDelete
             // Revoke/delete entitlements
             if (!entitlements.isEmpty()) {
                 log.info("Revoking {} entitlements...", entitlements.size());
-                this.entitlementCurator.batchDelete(entitlements);
+                this.entitlementCertificateCurator.deleteByEntitlementIds(entitlementIds);
+                this.entitlementCurator.batchDeleteByIds(entitlementIds);
```
Ceiu: The one thing that gives me pause here is that EntitlementCurator is one of a handful of curators that has extra behavior on delete. While I've no issue with doing direct SQL operations in general (though I prefer HQL due to Hibernate being Hibernate), I worry that our auto-commit behavior may cause problems if we don't update the entities still in memory.

Have you done any testing or such to verify that skipping the consistency assignments won't cause problems?

nikosmoum: Surprisingly, there were no issues when running this in a deployed candlepin with a data dump from prod, but it turns out the unit test failures were because of this omission. I have now added back the extra behavior in addition to using the SQL delete.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -788,6 +788,43 @@ public void batchDelete(Collection<Entitlement> entitlements) {
...
+    }
+
+    /**
+     * Maintains the runtime consistency of the given collection of entitlements, by removing references to
+     * themselves and their certificates. This method is supposed to be run before performing the deletion
+     * of the given list of entitlements.
+     *
+     * @param entitlements The collection of entitlements whose references are to be removed.
+     */
+    public void preDeleteOperations(Collection<Entitlement> entitlements) {
```
Ceiu: How's the memory usage with this step added back in? Are we getting back to the point where this is reintroducing the memory problems?

nikosmoum: Nope. The memory usage is almost exactly the same as without it. The memory issues where from the hibernate Session itself accumulating data of all the entities that are to be deleted before we flush. The entitlements and their consumers/pools that this method operates on are either already loaded, or will be loaded later on in the deletePools method anyway, so there is no impact in memory from it.

Ceiu: Classic Hibernate.

Ceiu:
Alright. The only hangup I have left is the method name; but I can't think of anything better/clearer off the top of my head (unlinkEntitlements, maybe?). If you can and feel like changing it, please do so. Otherwise, feel free to merge.

nikosmoum:
@Ceiu changed the method name to unlinkEntitlements!

PR #2431, opened 2019-09-09
[ENT-1530]1732892: Registration messaging changed to "owner with key xxx no found"

wolfdale:
- Modified Exception Message after product security review.

### DIFF
--- a/server/spec/owner_resource_spec.rb
+++ b/server/spec/owner_resource_spec.rb
@@ -185,7 +185,7 @@ describe 'Owner Resource' do
     #and this will fail
     lambda do
       ro_owner_client.register('systemFoo')
-    end.should raise_exception(RestClient::ResourceNotFound)
+    end.should raise_exception(RestClient::Forbidden)
   end
 
   it "does not let the owner key get updated" do
--- a/server/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -1113,7 +1113,8 @@ public class ConsumerResource {
             log.warn("User {} does not have access to create consumers in org {}",
                 principal.getPrincipalName(), owner.getKey());
 
-            throw new NotFoundException(i18n.tr("owner with key: {0} was not found.", owner.getKey()));
+            throw new ForbiddenException(i18n.tr("{0} is not authorized to register with " +
+                    "organization {1}", principal.getName(), owner.getKey()));
         }
 
         return owner;


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1113,7 +1113,8 @@ private Owner setupOwner(Principal principal, String ownerKey) {
             log.warn("User {} does not have access to create consumers in org {}",
                 principal.getPrincipalName(), owner.getKey());
 
-            throw new NotFoundException(i18n.tr("owner with key: {0} was not found.", owner.getKey()));
+            throw new ForbiddenException(i18n.tr("{0} is not authorized to register with " +
```
Ceiu: We still need to check if we're in the right condition to send this more explicit message. So far we only know that the principal is not allow to create consumers in the org, but we don't know if the user is part of the org.

Rather than replace the old message entirely, we should branch on whether or not the principal's owner list includes the owner in question (indicating the user is allowed to know about the org). The UserPrincipal has three methods for fetching the owners: ```getOwners```, ```getOwnerKeys```, and ```getOwnerIds```. The latter two simply iterate over the values returned by ```getOwners``` (which itself iterates over its permissions to build the list and does not check for uniqueness), so it's probably best to just use it directly and check if it contains any owner with the same ID as the owner we've fetched above on line 1104.

wolfdale: agreed. fixed it.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1113,8 +1113,15 @@ private Owner setupOwner(Principal principal, String ownerKey) {
             log.warn("User {} does not have access to create consumers in org {}",
                 principal.getPrincipalName(), owner.getKey());
 
-            throw new ForbiddenException(i18n.tr("{0} is not authorized to register with " +
+            List<Owner> listOfOwner = ((UserPrincipal) principal).getOwners();
+            boolean isOwnerContained =
+                listOfOwner.stream().anyMatch(t -> t.getOwnerId().equals(owner.getOwnerId()));
```
Ceiu: Couple minor things here:

- Since the principal implementation isn't entirely fixed to our own code (we may get a subclass from IT's adapters), it's a good idea to be a bit more defensive here. A null check on the list and a simple null check in the anyMatch makes this completely safe:

```boolean isOwnerContained = owners != null && owners.stream()....```

```anyMatch(t -> t != null && ...)```

- For formatting multi-line statements like this, generally break it up by operation:

```
boolean isOwnerContained = owners != null && owners.stream()
    .anyMatch(...);
```

This keeps the formatting consistent when doing method chaining, allows additional chains to be added without reformatting, and makes each individual operation clearly defined.

- Real minor/pedantic: grammatically the variable should be plural, as it contains multiple owners, e.g. listOfOwners. That can be further abbreviated to simply "owners" if the name is still available. However, keeping the "listOf" is a stylistic option if you'd prefer it.

File: server/spec/owner_resource_spec.rb [hunk 3]
```diff
@@ -185,7 +185,7 @@
     #and this will fail
     lambda do
       ro_owner_client.register('systemFoo')
-    end.should raise_exception(RestClient::ResourceNotFound)
+    end.should raise_exception(RestClient::Forbidden)
   end
```
Ceiu: Let's add another spec test to cover the different exception for the case where the user has no visibility to the owner.

PR #2434, opened 2019-09-13
[F] ENT-869: Use SSL when deploying Artemis remotely

nikosmoum:
- Also, now candlepin.conf is populated with remote artemis
  properties when running the deploy script with -b -a

### DIFF
--- a/server/conf/candlepin.conf.template
+++ b/server/conf/candlepin.conf.template
@@ -33,6 +33,10 @@ candlepin.importer.fail_on_unknown=false
 <% if (candlepin.external_broker) { %>\
 candlepin.audit.hornetq.embedded=false
 candlepin.audit.hornetq.broker_url=tcp://localhost:61617
+candlepin.audit.hornetq.truststore=/var/lib/artemis/candlepin/certs/artemis-client.ts
+candlepin.audit.hornetq.truststore_password=securepassword
+candlepin.audit.hornetq.keystore=/var/lib/artemis/candlepin/certs/artemis-client.ks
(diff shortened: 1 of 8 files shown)


nikosmoum:
Note for reviewers: Since the only additional thing required for Artemis to use SSL is generating certificates and configuration changes (in broker.xml and in candlepin.conf), this change is only doing that in the configure-artemis.py script, which is only called by the deploy script, which is not the way production deploys candlepin. In production, customers would have to do this themselves manually (as with the rest of what configure-artemis.py is doing, such as downloading and installing a standalone artemis instance/service).

server/conf/candlepin.conf.template
```diff
@@ -32,7 +32,7 @@ candlepin.importer.fail_on_unknown=false
 
 <% if (candlepin.external_broker) { %>\
 candlepin.audit.hornetq.embedded=false
-candlepin.audit.hornetq.broker_url=tcp://localhost:61617
+candlepin.audit.hornetq.broker_url=tcp://localhost:61617?sslEnabled=true&trustStorePath=/var/lib/artemis/candlepin/certs/artemis-client.ts&trustStorePassword=securepassword&keyStorePath=/var/lib/artemis/candlepin/certs/artemis-client.ks&keyStorePassword=securepassword
```
Ceiu: I feel like rather than the parameter soup on the URL we should build this internally using other config parameters. Maybe this will work for the short term (i.e. master with knowledge of the coming changes in the feature branch), but long-term it's kind of a mess.

nikosmoum: I've split these into multiple parameters.

PR #2437, opened 2019-09-18
[ENT-1140]1667952: candlepin refuses to start when having too many messages in ActiveMQ journal

wolfdale:
- Minor logging change to print stack trace

### DIFF
--- a/server/src/main/java/org/candlepin/audit/QpidEventMessageReceiver.java
+++ b/server/src/main/java/org/candlepin/audit/QpidEventMessageReceiver.java
@@ -70,12 +70,10 @@ public class QpidEventMessageReceiver extends MessageReceiver {
         }
         catch (Exception e) {
             log.debug("Message listener {} processed message: {} [{}]: FAILED", listener, msgId, origMsgId);
-            // Log a small message instead of a full stack trace to reduce log size.
-            String reason = (e.getCause() == null) ? e.getMessage() : e.getCause().getMessage();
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/candlepin/audit/QpidEventMessageReceiver.java
```diff
@@ -70,9 +70,7 @@ public void onMessage(ClientMessage msg) {
         }
         catch (Exception e) {
             log.debug("Message listener {} processed message: {} [{}]: FAILED", listener, msgId, origMsgId);
-            // Log a small message instead of a full stack trace to reduce log size.
-            String reason = (e.getCause() == null) ? e.getMessage() : e.getCause().getMessage();
-            log.error("Unable to process message {}[{}]: {}", msgId, origMsgId, reason);
+            log.error("Unable to process message {}[{}]: {}", msgId, origMsgId, e);
 
             // If debugging is enabled log a more in depth message.
             log.debug("Unable to process message. Rolling back client session.\n{}", body, e);
```
Ceiu: This extra debug output is largely redundant now. We should update it so that it no longer doubly prints the exception and duplicate message. This would probably be fine with just the body and no message (i.e. ```log.debug("{}", body)```).

wolfdale: Fixed it.

PR #2438, opened 2019-09-18
[2.7] 1752001: prefer virtual pool and conusmer match over syspurpose null_rule; ENT-1634

abhiskum:


abhiskum:
@nikosmoum updated correct bug id.

abhiskum:
@nikosmoum Fix check style issues.

PR #2439, opened 2019-09-18
[2.7] 1753201: Initialize the NSS db when loading a JSS provider; ENT-1656

nikosmoum:
- Now, when creating a JSS provider on deployment, we also create
  a JSS CryptoManager, which initializes the NSS db.
- Downgraded JSS to 4.4.6 to be in line with the jars/libraries
  distributed in RHEL 7. This means with this fix, candlepin is
  not compatible with RHEL 8, which ships with JSS 4.5.x.
- Added https://barnabycourt.fedorapeople.org/repo/candlepin as
  a new, temporary, artifact repository for our builds, which holds
  the jss 4.4.6 dependency for now.

Co-authored-by: Alexander Scheel ascheel@redhat.com

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -176,6 +176,8 @@ subprojects {
         maven { url "http://oauth.googlecode.com/svn/code/maven/" }
         // For LogDriver
         maven { url "http://awood.fedorapeople.org/ivy/candlepin/" }
+        // Temporary repo, which stores jss 4.4.6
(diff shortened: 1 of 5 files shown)


nikosmoum:
Note: The reason for porting this is that this bug is blocking development on branch 2.7 for developers that use the latest centos7 vagrant box, which has been upgraded to use Centos 7.7, in which this bug manifests itself (similar to how it manifested in RHEL 7.7 only, but not earlier versions, when in production sat environments).

PR #2442, opened 2019-09-19
[M] 1753750: Added a default value for Pool.locked

Ceiu:
- Added a default value for the Pool.locked field to properly
  handle pools which existed before the addition of this field


barnabycourt:
Merging because code review was provided & approved.

PR #2443, opened 2019-09-20
Update the version of org.hibernate.validator

barnabycourt:
Also updated the pom files which picked up some earlier library changes.


wottop:
ack

PR #2444, opened 2019-09-23
[ENT-290] Add Bulk consumer exists call

wolfdale:


server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -960,6 +960,23 @@ public boolean doesConsumerExist(String uuid) {
...
+    public List<String> getNonExistingConsumer(ArrayList<String> listOfUids) {
+        DetachedCriteria criteria = this.createSecureDetachedCriteria()
+            .add(Restrictions.in("uuid", listOfUids));
+
+        CandlepinQuery<Consumer> consumersFound = this.cpQueryFactory.
+            <Consumer>buildQuery(this.currentSession(), criteria);
+
+        List<String> collectFoundIds = new ArrayList<>();
+
+        for (Consumer co : consumersFound) {
```
Ceiu: This loop and the collection is unnecessary. If you're going to keep the general behavior of modifying the input list (which I would argue against), iterate through it and remove the elements you get back from the query.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -449,6 +449,26 @@ public void consumerExists(
         }
     }
 
+    @ApiOperation(
+        notes = "Checks for the existence of a Consumer in bulk. This API returns uuids of non-existing" +
+        "consumer",
+        value = "")
+    @ApiResponses({ @ApiResponse(code = 400, message = "Return 404 when empty or null array is passed"),
```
Ceiu: These response codes don't line up with the actual output in the code below, nor the outputs defined in the task.

According to the Jira task, we are looking for two things:
- If the list contains any consumer UUIDs which do not exist, this endpoint should return a 404 and a list containing those consumer UUIDs which did not exist.
- If the list contains only existing consumer UUIDs, this endpoint should return a 204 and an empty list. Note that an empty input list matches this case.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -449,6 +449,26 @@ public void consumerExists(
...
+        @ApiResponse(code = 200, message = "Return array of UUIDs of consumers that doesn't exist or " +
+        "cannot be accessed") })
+    @POST
+    @Consumes(MediaType.WILDCARD)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("exists")
+    public List<String> consumerExistsBulk(ArrayList<String> consumerUuid) throws BadRequestException {
```
Ceiu: Does this absolutely need to be an ArrayList, or will any list (or collection/iterable) work here?

Also, if it's a collection, the variable name should reflect that with a plural.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -449,6 +449,26 @@ public void consumerExists(
...
+        if (!CollectionUtils.isEmpty(consumerUuid)) {
+            return consumerCurator.getNonExistingConsumer(consumerUuid);
+        }
+        else {
+            throw new BadRequestException(i18n.tr("No Uuid's provided. Unable to process request."));
```
Ceiu: Grammar: 
- UUID is an acronym and should be entirely capitalized (or not capitalized at all depending on context and general usage throughout the application)
- An apostrophe signifies ownership or character omission; it is not applicable when pluralizing
- Generally follow the pattern of "Error: reason" or "event; details"; e.g. "Unable to process request: no UUIDs provided". However, in this case, the error being output to the end user will already imply that it was unable to process the request (that's what 4xx and 5xx codes are for), so in some contexts it's not necessary to include that part of the message.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -960,6 +960,23 @@ public boolean doesConsumerExist(String uuid) {
         return result != 0;
     }
 
+    public List<String> getNonExistingConsumer(ArrayList<String> listOfUids) {
```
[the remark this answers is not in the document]
wolfdale: Added Java Doc. 
Using Set to avoid duplicates

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -960,6 +960,23 @@ public boolean doesConsumerExist(String uuid) {
...
+        CandlepinQuery<Consumer> consumersFound = this.cpQueryFactory.
+            <Consumer>buildQuery(this.currentSession(), criteria);
+
+        List<String> collectFoundIds = new ArrayList<>();
+
+        for (Consumer co : consumersFound) {
+            collectFoundIds.add(co.getUuid());
+        }
+
+        listOfUids.removeAll(collectFoundIds);
```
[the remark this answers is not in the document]
wolfdale: Removing it from same Input Collection to avoid creating a copy of it. Will leave a comment on it.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
wolfdale: Yes. Fixing it. Removing it from same Input Collection to avoid creating a copy of it.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wolfdale: Fixing it. Using a Generic collection representation for I/O.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -960,6 +960,23 @@ public boolean doesConsumerExist(String uuid) {
         return result != 0;
     }
 
+    public List<String> getNonExistingConsumer(ArrayList<String> listOfUids) {
+        DetachedCriteria criteria = this.createSecureDetachedCriteria()
+            .add(Restrictions.in("uuid", listOfUids));
```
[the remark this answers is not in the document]
wolfdale: Partitioned it into block sizes of 1000.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wolfdale: Modifying it to "No UUIDs provided" as it is apparent from Bad Request Error code, that request cant be processed.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wolfdale: The Response to API does not add up to Jira, as I believe it was written according to HEAD API to be as a reference point.

wolfdale: At this point I am unsure if `BAD_REQUEST` is needed or not.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -960,6 +960,28 @@ public boolean doesConsumerExist(String uuid) {
         return result != 0;
     }
 
+    /**
+     * @param setOfUids of consumer UUIDs.
+     * @return set of consumer UUIDs that doesnt exists.
+     */
+    public Set<String> getNonExistingConsumer(Set<String> setOfUids) {
```
Ceiu: This is still very much being defined and implemented as a specific-purpose function, but at the curator level (which is general purpose). If this is the way you want this method to work, define it in the resource as a private function and call the standard lookup functions. If this is to stay in the curator, you need to attempt to generalize it so that it can be used elsewhere without needing to be concerned with the requirements of the endpoint that leaked into its implementation.

That is:
- It still uses a slow lookup of full Consumer instances (can be replicated by ```getConsumers```)
- It's still requiring a specific collection type (Set)
- It's still modifying that input collection, which prevents use of several proxy collections and unmodifiable/singleton collections
- The output is superfluous since the input collection is modified directly
- The behavior of the input collection modification is still undocumented and only apparent by reading the code
- The method name implies something other than the behavior implemented


As touched on above, it's also fetching entire consumer instances when all the logic cares about is whether or not a given consumer UUID represents an existing consumer or not. This being the case, the query should be fetching only consumer UUIDs (or absent UUIDs).

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -449,6 +449,33 @@ public void consumerExists(
...
+        "cannot be accessed"),
+        @ApiResponse(code = 204, message = "If the all consumer UUIDs exists and can be accessed"),
+        @ApiResponse(code = 400, message = "When no UUIDs are provided") })
+    @POST
+    @Consumes(MediaType.WILDCARD)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("exists")
+    public Response consumerExistsBulk(Set<String> consumerUuids) throws BadRequestException {
+        if (!CollectionUtils.isEmpty(consumerUuids)) {
+            Set<String> nonExistingUuids = consumerCurator.getNonExistingConsumer(consumerUuids);
```
Ceiu: Given the behavior of the getNonExistingConsumers method, this is very risky: you don't know the capabilities of the Set implementation you're passing through.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -449,6 +449,33 @@ public void consumerExists(
         }
     }
 
+    @ApiOperation(
+        notes = "Checks for the existence of a Consumer in bulk. This API return UUIDs of non-existing" +
+        "consumer",
+        value = "")
+    @ApiResponses({ @ApiResponse(code = 404, message = "Returns all consumer UUIDs that doesn't exist or " +
+        "cannot be accessed"),
+        @ApiResponse(code = 204, message = "If the all consumer UUIDs exists and can be accessed"),
```
Ceiu: Extraneous "the" in the message string.

wolfdale: removed

wolfdale:
@Ceiu Do we also need to secure this rest endpoint ?

server/src/main/java/org/candlepin/model/ConsumerCurator.java
wolfdale: Input collection is not being modified now. Criteria query is fixed to fetch UUIDs only.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wolfdale: Changed it to `Iterable`  to be more generic.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -960,6 +960,34 @@ public boolean doesConsumerExist(String uuid) {
         return result != 0;
     }
 
+    /**
+     * Given the Consumer UUIDs filter out & returns the non-existing & unique consumer UUIDs
+     *
+     * @param consumerUuids consumer UUIDs.
+     * @return set of consumer UUIDs that doesnt exists.
+     */
+    public Set<String> filterNonExistingConsumerIds(Iterable<String> consumerUuids) {
```
Ceiu: The name change makes more sense if you were still actually filtering the input collection.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -960,6 +960,34 @@ public boolean doesConsumerExist(String uuid) {
...
+        Set<String> nonExistingConsumerUuids = Collections.emptySet();
```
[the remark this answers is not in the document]
wolfdale: Alright.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
wolfdale: Sure.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -960,6 +960,34 @@ public boolean doesConsumerExist(String uuid) {
...
+
+        if (consumerUuids != null && consumerUuids.iterator().hasNext()) {
+            nonExistingConsumerUuids = new HashSet<>();
+            int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit());
+
+            consumerUuids.forEach(nonExistingConsumerUuids::add);
+
+            for (List<String> subList : Iterables.partition(consumerUuids, blockSize)) {
+                List existingUuids = this.createSecureCriteria()
```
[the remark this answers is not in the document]
wolfdale: I prefer JPA Criteria over JPQL, they are much more readable.

wolfdale:
This was not apparent, but makes sense. I would generally not prefer mixing Java and SQL in same method/place, and in much more favor of using hibernate criteria queries as it looks more elegant and readable. For now changes are done according to review comment.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -463,12 +463,18 @@ public void consumerExists(
     @Path("exists")
     public Response consumerExistsBulk(Iterable<String> consumerUuids) throws BadRequestException {
         if (consumerUuids != null && consumerUuids.iterator().hasNext()) {
-            Set<String> nonExistingUuids = consumerCurator.filterNonExistingConsumerIds(consumerUuids);
+            Set<String> existingUuids = consumerCurator.getExistingConsumerUuids(consumerUuids);
+
+            Set<String> nonExistingUuids = new HashSet<>();
+            consumerUuids.forEach(nonExistingUuids::add);
+            nonExistingUuids.removeAll(existingUuids);
```
Ceiu: Do we need the extra set here? You can just remove the received set from the input set and get the same result without doing a copy.

wolfdale: Here I don't have input set rather its `Iterable` type (it does not have remove method).  Even If I change it to `Collection` type I don't have a way to remove duplicate and return only unique IDs

Ceiu: Collection does have remove methods, but they're optional, so it's a bit dangerous to rely on them.

In this case since it's not a global or utility method, nor something that is commonly used elsewhere, it's not critical to be generic with your inputs\*. Also, we know RESTEasy is nice enough to give us whatever collection type we request.

\* My comment regarding that was in the context of libraries and shared classes. Sorry for the confusion.

wolfdale: Changed the Inputs to accept `SET`

wolfdale:
Squashed and ready for merge, after final review

Ceiu:
The ConsumerResourceTest is somewhat of an exception here, since it's so old and needs to be replaced; but next time make sure to write local unit tests for your new endpoint methods, rather than just spec tests.

Also try to focus on whitespace usage, even in tests. Big walls of text are generally bad.

PR #2445, opened 2019-09-23
[F] ENT-1610: Add Branding to Product model

nikosmoum:
- Now Product holds Branding information by means of
  the cp_product_branding table.
- ProductDTO supports BrandingDTO information in POST/GET
  product API requests.
- Updated/Created unit and spec tests
- Updated method javadocs in ProductManager.java

### DIFF
--- a/server/src/main/java/org/candlepin/model/ProductCurator.java
+++ b/server/src/main/java/org/candlepin/model/ProductCurator.java
@@ -330,6 +330,19 @@ public class ProductCurator extends AbstractHibernateCurator<Product> {
             }
         }
 
+        if (entity.getBranding() != null) {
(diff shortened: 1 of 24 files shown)


server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -330,6 +330,17 @@ protected Product validateProductReferences(Product entity) {
             }
         }
 
+        if (entity.getBranding() != null) {
```
Ceiu: I wonder if we're getting to a point where we should have an ```isComplete``` method on our entities to check this kind of thing without needing this kind of code in a number of places.

abhiskum:
Looks good to me.

PR #2449, opened 2019-09-25
Update to Gradle 5.6.2

barnabycourt:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-all.zip
```
Januson: PR should include the updated wrapper files as well. They will get downloaded by running './gradlew wrapper'.

Last time we updated the wrapper I added gradle wrapper task definition at the bottom of the main build.gradle. It has set the version for the wrapper to 5.3. We should either update the version there or remove it and supply version as an argument of the update command.

barnabycourt: I removed the version specification in the wrapper task def & re-ran with the wrapper plugin to update the versions. It appears that with that change the only file updated was the gradle-wrapper.properites.

PR #2450, opened 2019-09-25
Use more reliable method for detecting the hibernate dialect

wottop:
The Stage/Production deployments do not use the configuration entry that
 was in use


wottop:
Would adding the MariaDB dialect to the condition take care of it?

Ceiu:
For the short-term, I suppose. I just want to make sure we're not just kicking the can down the road, so to speak. Like I said at the start of the last post, the whole approach feels janky; but so long as we don't keep doing it, it's probably fine.

PR #2453, opened 2019-09-27
1702035 - [RFE] Set autoheal as part of consumer creation.

barnabycourt:
Allow setting the autoheal attribute of consumer during the rest call that creates the consumer.


barnabycourt:
Question for reviewer, should this add capability to CP specifically for this feature?

wottop:
I'd say no to the capability. As you could set it at update already, it was capable of setting this value.

ack. i'll merge if you agree with my logic.

barnabycourt:
@wottop makes sense to me. Sub man can set the value, and only make the 2nd call if the value is not persisted in the creation.

wottop:
ack

PR #2455, opened 2019-10-02
[F] ENT-1612, ENT-1617: Branding to Product on Refresh 

nikosmoum:
Now, during refresh, fetch branding from the upstream product
and not from the upstream subscription, and persist it on the
main product instead of the pool.

- Adapter API changes:
    - Add branding ProductInfo/ProductData.
    - Remove branding from SubscriptionInfo/Subscription.
    - Add utility method on BrandingInfo that returns a
      Comparator for objects of classes that implement it.

- REST API changes:
    - /owners/{owner_key}/subscriptions legacy APIs that use the
      Subscription object now expect branding on ProductData
      instead of Subscription.

- HostedTestSubscriptionServiceAdapter/HostedTestSubscriptionResource
  now expect branding on ProductData instead of Subscription.
- Add branding on manifest ProductDTO (but ignore it on
  serialization/deserialization), in order for EntitlementImporter
  to set imported branding to the product and not the pool.
- Populate PoolDTO branding from product on GET request (ENT-1617).
- ProductManager import now also takes into account Branding.
- Add branding to entitlement certificates from the main product
  now instead of the pool.
- PoolRules/PoolUpdate no longer track branding changes, since
  those will now be reflected in product changes.
- Add/Update relevant unit/spec tests.

### DIFF
--- a/server/spec/candlepin_scenarios.rb
+++ b/server/spec/candlepin_scenarios.rb
@@ -507,8 +507,6 @@ class StandardExporter < Exporter
     ]
 
     # the before(:each) is not initialized yet, call create_product sans wrapper
-    @products[:product1] = create_product(random_string('prod1'), random_string(),
-                              {:multiplier => 2, :branding => brandings})
     @products[:product2] = create_product(random_string('prod2'), random_string())
     @products[:virt_product] = create_product(random_string('virt_product'),
                                   random_string('virt_product'),
@@ -538,7 +536,21 @@ class StandardExporter < Exporter
                                    :required_tags => "TAG1,TAG2",
                                    :arches => "i386,x86_64"})
 
-    @cp.add_content_to_product(@owner['key'], @products[:product1].id, content.id)
+    # This is to avoid hibernate NonUniqueObjectExceptions in hosted mode during refresh, which happens later
+    # in this method, where the product is being updated because it is not locked. That exception is thrown
+    # due to the fact that the product's branding set is immutable. By creating the product upstream in the
+    # first place we avoid any such issues.
+    if is_hosted?
+      @products[:product1] = create_upstream_product(random_string('prod1'),
+                              {:multiplier => 2, :branding => brandings})
+      upstream_content_id = create_upstream_content(content.id, content)
+      add_content_to_product_upstream(@products[:product1].id, upstream_content_id.id)
+    else
+      @products[:product1] = create_product(random_string('prod1'), random_string(),
+                              {:multiplier => 2, :branding => brandings})
+      @cp.add_content_to_product(@owner['key'], @products[:product1].id, content.id)
+    end
+
     @cp.add_content_to_product(@owner['key'], @products[:product2].id, content.id)
     @cp.add_content_to_product(@owner['key'], @products[:product2].id, arch_content.id)
     @cp.add_content_to_product(@owner['key'], @products[:derived_product].id, content.id)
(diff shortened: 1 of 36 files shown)


server/src/main/java/org/candlepin/service/ProductServiceAdapter.java
```diff
@@ -37,7 +37,7 @@
      * When this method is called by candlepin, candlepin has already verified that
      * the specified owner is known to candlepin and should not be null. If an
      * implementation receives a null owner, a RuntimeException should be thrown.
-     * If the owner is unknown to the service, an empty list of products should
+     * If the owner is unknown to the service, an empty list of products shouldbrand
```
Ceiu: Typo

nikosmoum:
@Ceiu Do you mean that it could be reworked to use the *Info interfaces without changing them (e.g. adding getBranding() to the ProductInfo interface)? If so, in what way? Also, if we're, going to change ProductInfo to have getBranding(), that would break the Adapters, right? I'd be very happy with that, since it is much more elegant/simple, but the jira card epic description explicitly calls out not asking the adapters code to change, although I don't know the rationale (I guess we want to only ask them to do one large change later when we have the bulk of the product hierarchy changes?).

Ceiu:
One option we could pursue is adding the ```getBranding``` method to the ProductInfo interface as a default method that returns null, with the intent to make it *not* default in the near future. That would satisfy both the updating of the interface without breaking the existing adapter code; though I'm not a huge fan of the perversion of the definition of an interface with defaults.

But what I was getting at overall is that we have two critical paths of code that converge in CPM: manifest import and refresh. Both get their data from "upstream," where manifest import uses an adapter to simulate upstream from the manifest file, and refresh uses the IT adapters to pull it from their data source. As a result, the bulk of the code that uses or juggles branding within Candlepin will go through these paths, so it makes sense to me to ensure that these are designed with the data sources in mind. To me, that means making a decision as to whether or not the adapters (and thus, the Info interfaces) represent the data as it appears upstream and leave the juggling/massaging to CP, or if the adapters should act as a form of API for communicating with CP, which suggests we should update them accordingly.

nikosmoum:
@Ceiu I'd prefer to see it as the later (adapters should be an API).

Specifically about the *Info interfaces, I just checked the adapters code, and they don't use them to implement their own concrete objects. They still use our Subscription.java and ProductData.java concrete objects. So there is no literal "breaking" of the adapters (as in compile problems), but simply that they would be forced to update their code to populate ProductData.java objects with branding appropriately (and not populate branding on Subscription.java), so that we don't have to on our side.

@barnabycourt What do you think about this? What is the reason for the requirement of not changing the adapters API? Changing it now would avoid further refactoring in the future.

nikosmoum:
Moved to in progress since we got agreement from IT to change the adapters API.

nikosmoum:
@Ceiu Pushed an updated with rework to update the Adapters API

A couple of notes on things that I am concerned about:
- I used the branding model entity on ProductData, to avoid
  creating another DTO (e.g. BrandingData), since Subscription
  is already doing that (uses Owner, Cdn, etc.), and we already
  want to get rid of these *Data classes anyway. Do you think
  this is a good compromise, or could it cause any problems?
  We could use api.v1.BrandingDTO as an alternative, but that
  might not be right (to mix API types).

- Is it a problem that the /owners/{owner_key}/subscriptions 
  legacy APIs are going to change? Are they still used? If they
  are, then that is a problem, since we're using the same DTOs
  for them and for the Adapters, which is tying our hands (not
  being able to change the Adapter API).

PR #2460, opened 2019-10-04
ENT-1682: Update build process to use Fedora Zanata

Januson:
- Switched url to fedora zanata
- Removed deprecated locales block


Ceiu:
ack

PR #2464, opened 2019-10-10
[F] Added the Candlepin messaging package and the new CandlepinModeManager (ENT-1464)

Ceiu:
- Rewrote the mode manager as CandlepinModeManager and added additional
  handling for arbitrary reasons and classes of reasons, with provisions
  to ensure reason resolution doesn't trigger normal mode before all
  suspend mode reasons are resolved
- Moved all mode manager classes and functionality to the package
  org.candlepin.controller.mode
- Added the Candlepin messaging system in the package
  org.candlepin.messaging, providing an abstraction around typical
  broker-based messaging systems as they are used within Candlepin
  at the time of writing
- Added an Artemis-based implementation of the Candlepin messaging
  interfaces
- The state JobManager.PAUSED has been renamed to JobManager.SUSPENDED
  to better align with other subsystems which also perform state
  management
- The JobManager now uses the new Candlepin messaging abstraction
  rather than now-deprecated components of the audit system, or any
  specific messaging system/implementation
- Rewrote the JobMessageDispatcher and JobMessageReceiver in
  accordance of the change above. Both are no longer interfaces and
  instead implement proper message handling using the Candlepin
  messaging system
- Added the ability to suspend and resume the job system without
  risking erroneously resuming the job manager when juggling
  suspend mode management


Ceiu:
There may be more to be done here, or stuff that's sloppier than we may like. Off the top of my head, it makes little sense for the event and auditing system to continue using the Artemis and Qpid bits directly, and both should instead use the new CPM stuff.

Additionally, the configuration surrounding messaging is all over the place. However, that's a bit messier to get into now, as doing it "right" would mean adding backward-compatible lookups instead of just breaking everyone's configuration.

Thoughts on any of the above or any other such things contained within this PR are more than welcome.

Ceiu:
Do not merge this until it is squashed, and possibly rebased. But mostly squashed.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -289,19 +292,21 @@ public void beforeCompletion() {
...
     private final CandlepinRequestScope candlepinRequestScope;
     private final PrincipalProvider principalProvider;
     private final Injector injector;
 
     private ManagerState state;
+    private JobMessageSynchronizer synchronizer;
     private QuartzJobExecutor qrtzExecutor;
     private Scheduler scheduler;
     private ThreadLocal<Map<String, String>> mdcState;
+    private Set<Object> suspendKeys;
```
kahowell: Somewhere (doesn't have to be this line), it would be nice to see an explanation of why suspend keys are Object, and what concrete types we might expect to see for them.

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
```diff
@@ -14,13 +14,159 @@
...
+
+        this.config = Objects.requireNonNull(config);
+        this.cpmSessionFactory = Objects.requireNonNull(cpmSessionFactory);
+        this.objMapper = Objects.requireNonNull(objMapper);
+
+        this.initialized = false;
+    }
+
+    /**
+     * Initializes this message dispatcher, specfiying the job manager that will be dispatching
```
kahowell: Typo: s/specfiying/specifying/

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
```diff
@@ -14,13 +14,159 @@
...
+        return this.cpmSession;
+    }
+
+    /**
+     * Fetches the current CPM producer, creating a new one if necessary.
+     *
+     * @return
+     *  a CPMProducer instance
+     */
+    private CPMProducer getProducer() throws CPMException {
```
kahowell: Lots of other methods in this class are synchronized. Does this method need to be too?

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
```diff
@@ -14,13 +14,159 @@
...
+     * @param manager
+     *  the JobManager that will be dispatching messages
+     *
+     * @throws IllegalStateException
+     *  if this message receiver has already been initialized
+     *
+     * @throws IllegalArgumentException
+     *  if the provided job manager is null
+     */
+    public synchronized void initialize(JobManager manager) {
```
kahowell: This seems to not do anything. At first glance at least, it seems like this method, the `initialized` var, and `isInitialized` are no longer necessary?

server/src/main/java/org/candlepin/audit/ArtemisMessageSourceReceiverFactory.java
```diff
@@ -147,10 +146,10 @@ else if (blacklist.size() > 0) {
         int maxJobs = config.getInt(ConfigProperties.ASYNC_JOBS_THREADS);
         log.info("Adding {} job receivers with filter: {}", maxJobs, this.amqpFilter);
 
-        for (int i = 0; i < maxJobs; i++) {
-            messageReceivers.add(
-                new JobMessageReceiver(this.amqpFilter, this.manager, sessionFactory, mapper));
-        }
+        // for (int i = 0; i < maxJobs; i++) {
```
kahowell: Can we just remove this code now (instead of commented out)?

server/src/main/java/org/candlepin/messaging/CPMMessage.java
```diff
@@ -0,0 +1,150 @@
...
+
+    // Message properties
+    // CPMMessage setProperty(String key, boolean value);
+    // CPMMessage setProperty(String key, byte value);
+    // CPMMessage setProperty(String key, byte[] value);
+    // CPMMessage setProperty(String key, short value);
+    // CPMMessage setProperty(String key, int value);
+    // CPMMessage setProperty(String key, long value);
+    // CPMMessage setProperty(String key, float value);
+    // CPMMessage setProperty(String key, double value);
```
wottop: Plan?

server/src/main/java/org/candlepin/controller/mode/ModeChangeReason.java
```diff
@@ -0,0 +1,150 @@
...
+     *  A short string identifying the reason for entering suspend mode
+     *
+     * @param message
+     *  An optional, detailed message providing additional information for the reason Candlepin
+     *  operations are suspended
+     *
+     * @throws IllegalArgumentException
+     *  if the provided reason string is null or empty
+     */
+    public ModeChangeReason(String reasonClass, String reason, Date time, String message) {
```
kahowell: nitpick: reasonClass not documented in javadoc.

server/src/main/java/org/candlepin/messaging/CPMMessage.java
```diff
@@ -0,0 +1,150 @@
...
+     * Fetches the body of this message.
+     *
+     * @return
+     *  the body of the content of this message
+     */
+    String getBody();
+
+
+    // Message properties
+    // CPMMessage setProperty(String key, boolean value);
```
kahowell: Can we drop these, or leave a comment that other types for value could be added in the future?

server/src/main/java/org/candlepin/messaging/impl/artemis/ArtemisSessionFactory.java
```diff
@@ -0,0 +1,233 @@
...
+                this.sessionFactory.getConnection().addCloseListener(this);
+
+                log.debug("Created new Artemis client session factory: {}", this.sessionFactory);
+            }
+
+            return this.sessionFactory;
+        }
+
+        /**
+         * Fetches the current client session, creating and intializing a new instance as necessary.
```
wottop: The code appears to create only.

server/src/main/java/org/candlepin/async/JobManager.java
Ceiu: They're like locks -- it doesn't matter what the object passed in is, so long as the same (or an equal) object is used to resume it. I use strings internally for self-documenting purposes, but that is not strictly necessary. It's just as valid if we have multiple resource endpoints that could cause jobs to be suspended, that each would pass themselves in as the "lock" to ensure that they don't unlock each other's suspend action.

I can add more to the method documentation if it doesn't adequately cover this.

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
Ceiu: Yes

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
Ceiu: Maybe not. These were added before any actual code was written, so they largely live on due to the interface not changing after the fact and for consistency; though I suppose they could be dropped.

server/src/main/java/org/candlepin/audit/ArtemisMessageSourceReceiverFactory.java
Ceiu: Sure

server/src/main/java/org/candlepin/controller/mode/ModeChangeReason.java
Ceiu: fixed

server/src/main/java/org/candlepin/async/JobMessageReceiver.java
```diff
@@ -12,208 +12,485 @@
...
+public class JobMessageReceiver {
     private static Logger log = LoggerFactory.getLogger(JobMessageReceiver.class);
-    private static final String JOB_QUEUE_NAME = "jobs";
 
-    private JobManager manager;
-    private ClientSession session;
-    private ClientConsumer consumer;
-    private String msgFilter;
+    private static final String JOB_KEY_MESSAGE_PROPERTY = "job_key";
+    private static final String JOB_MESSAGE_ADDRESS = "job";
```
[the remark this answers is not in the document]
Ceiu: Removed

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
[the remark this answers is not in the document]
Ceiu: Part of me wants to keep it, but it does not have any immediate use with how the implementation ended up. I guess we can remove it for now.

server/src/main/java/org/candlepin/audit/ArtemisMessageSourceReceiverFactory.java
```diff
@@ -147,11 +146,6 @@ else if (blacklist.size() > 0) {
         int maxJobs = config.getInt(ConfigProperties.ASYNC_JOBS_THREADS);
         log.info("Adding {} job receivers with filter: {}", maxJobs, this.amqpFilter);
```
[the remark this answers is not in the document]
Ceiu: Fixed

server/src/main/java/org/candlepin/audit/EventSinkImpl.java
```diff
@@ -111,7 +115,10 @@ public void queueEvent(Event event) {
             return;
         }
 
-        modeManager.throwRestEasyExceptionIfInSuspendMode();
+        if (this.modeManager.getCurrentMode() != Mode.NORMAL) {
+            throw new IllegalStateException("Candlepin is in suspend mode");
```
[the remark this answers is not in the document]
Ceiu: We shouldn't be throwing REST exceptions from deep within a given subsystem. If we *really* need that exception, it should be handled by the appropriate endpoint and then converted to the proper REST exception.

In this case, since we have suspend mode handling already, the only time we should hit this is if something is misbehaving and allowing event-generating code to run while suspended.

server/src/main/java/org/candlepin/audit/ArtemisMessageSourceReceiverFactory.java
```diff
@@ -109,18 +108,18 @@ private String buildAMQPFilterExpression() {
             list.removeAll(blacklist);
 
             if (list.size() > 0) {
-                filter = String.format("%s IN ('%s')", ArtemisJobMessageDispatcher.JOB_KEY_MESSAGE_PROPERTY,
+                filter = String.format("%s IN ('%s')", JOB_KEY_MESSAGE_PROPERTY,
```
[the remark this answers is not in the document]
Ceiu: Removed

server/src/main/java/org/candlepin/messaging/CPMMessage.java
```diff
@@ -0,0 +1,132 @@
...
+     * Checks if this message is durable, or was sent durably.
+     *
+     * @return
+     *  true if this message is durable; false otherwise
+     */
+    boolean isDurable();
+
+    // CPMMessage setExpiration(long duration);
+
+    // boolean hasExpired();
```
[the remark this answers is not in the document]
Ceiu: These were leftover things I expected we'd need that we ended up not needing straight away. I guess we can remove it, but I'm slightly concerned that if we do end up needing them they won't get added in the right spot.

Removed for now.

server/src/main/java/org/candlepin/messaging/impl/artemis/ArtemisContextListener.java
```diff
@@ -0,0 +1,107 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void initialize(Injector injector) throws CPMException {
+        this.config = injector.getInstance(Configuration.class);
+
+        // Create the server if we're configured to do so
+        boolean embedded = this.config.getBoolean(ConfigProperties.ACTIVEMQ_EMBEDDED);
+
+        if (embedded) {
```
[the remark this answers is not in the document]
Ceiu: I don't feel this is the correct place for such a message. I've, instead, added the following log entry in the session factory:

```
log.info("Connecting to Artemis server at {}", brokerUrl);
```

This will indicate exactly where the server lives, in both the embedded and remote cases.

server/src/main/java/org/candlepin/guice/CandlepinModule.java
```diff
@@ -375,6 +381,21 @@ protected ValidatorFactory getValidationFactory(Provider<MessageInterpolator> in
             .buildValidatorFactory();
     }
 
+    protected void configureMessaging() {
+        String provider = this.config.getString(ConfigProperties.CPM_PROVIDER);
+
+        // TODO: Change this to a map lookup as we get more providers
+
+        if ("artemis".equalsIgnoreCase(provider)) {
```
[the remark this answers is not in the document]
Ceiu: Sure. Wasn't my intent, but makes sense.

server/src/main/java/org/candlepin/controller/ModeManagerImpl.java
```diff
@@ -1,122 +0,0 @@
...
- * mode where Candlepin stops serving requests and only responds to
- * Status resource.
- *
- * @author fnguyen
- */
-@Singleton
-public class ModeManagerImpl implements ModeManager {
-    private static Logger log = LoggerFactory.getLogger(ModeManagerImpl.class);
-    private CandlepinModeChange cpMode = new CandlepinModeChange(
-        new Date(), Mode.NORMAL, Reason.STARTUP);
```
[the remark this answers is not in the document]
Ceiu: I'm not really a fan of the idea of cycling the normal mode reasons, as it's not really useful for much beyond a "time since last incident" type of thing. As far as I can tell, the old mode reason thing was more an artifact of how the system worked rather than any by-design intent (especially considering it didn't work quite right when multiple suspend reasons got involved).

I could add the stack-based mode tracking, which would kind of address this, but the mode change reason would act more like a startup time than the old way of doing it. As such, I'm not sure if it's worth the effort to do unless someone actually checks those fields for normal mode state info and cared about the various reasons for restoring service.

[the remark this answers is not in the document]
nikosmoum: I'm fine with not adding this back, but I'd check if anyone (e.g. IT team) is using it for something.

Ceiu:
The skipped spec tests are no longer skipped and have been fixed.

server/src/main/java/org/candlepin/controller/mode/ModeChangeReason.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Date;
+
+
+
+/**
+ * Container class for a suspend mode reason and message pairing
+ */
+public class ModeChangeReason {
+    private final String reasonClass;
+    private final String reason;
```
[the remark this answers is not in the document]
Ceiu: Nah, since it'll vary from component to component that wants to suspend things. We can still use enums and pass those through (as is done in a couple places), but this is not mandated by the mode manager itself.

server/src/main/java/org/candlepin/messaging/CPMMessage.java
Ceiu: Already removed; but yes, these were originally planned until the implementation phase revealed we didn't need to implement them today, and laziness took over.

server/src/main/java/org/candlepin/audit/ArtemisMessageSourceReceiverFactory.java
```diff
@@ -147,10 +146,10 @@ else if (blacklist.size() > 0) {
...
         log.info("Adding {} job receivers with filter: {}", maxJobs, this.amqpFilter);
 
-        for (int i = 0; i < maxJobs; i++) {
-            messageReceivers.add(
-                new JobMessageReceiver(this.amqpFilter, this.manager, sessionFactory, mapper));
-        }
+        // for (int i = 0; i < maxJobs; i++) {
+        //     messageReceivers.add(
+        //         new JobMessageReceiver(this.amqpFilter, this.manager, sessionFactory, mapper));
+        // }
```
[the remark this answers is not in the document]
Ceiu: Already removed

server/src/main/java/org/candlepin/messaging/CPMMessage.java
```diff
@@ -0,0 +1,150 @@
...
+    CPMMessage setProperty(String key, String value);
+
+    // Boolean getPropertyAsBoolean(String key);
+    // Byte getPropertyAsByte(String key);
+    // byte[] getPropertyAsByteArray(String key);
+    // Short getPropertyAsShort(String key);
+    // Int getPropertyAsInt(String key);
+    // Long getPropertyAsLong(String key);
+    // Float getPropertyAsFloat(String key);
+    // Double getPropertyAsDouble(String key);
```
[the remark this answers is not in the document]
Ceiu: Removed

server/src/main/java/org/candlepin/messaging/CPMMessage.java
Ceiu: Removed

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
Ceiu: fixed

server/src/main/java/org/candlepin/controller/ModeManagerImpl.java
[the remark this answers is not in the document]
Ceiu: Preliminary check with IT and Sat suggests it is not needed today.

server/src/main/java/org/candlepin/async/JobManager.java
kahowell: :+1: Adding a statement such as "must implement equals" to suspend keys params would work.

Ceiu: But that's not strictly accurate. The only requirement would be that if equals or hashCode are overridden, they be implemented properly (or, at least, in a way that has repeatable and consistent output).

kahowell: Fair enough. The concerns I want to address are:
1) Try to avoid bugs introduced because someone passes in an instance of a custom class that *doesn't* implement `equals` properly.
2) Prevent folks from pestering you down the line with questions about what types work, etc.

It's really not a big deal though, so if you think of a line or two that helps, cool; otherwise, don't waste more time on it.

Ceiu:
I've added an additional note about it, and rebased+squashed the PR against the current feature branch.

PR #2465, opened 2019-10-11
ENT-797: Update ehcache version from 2.10.3 to 3.8.0

abhiskum:
- I did not find a clean way to configure cache in Java and use it into JPA initialization process. Hence continue using XML to configure cache properties.
- Updated net.sf.ehcache:ehcache:2.10.3 (ehcache 2) with org.ehcache:ehcache:3.8.0 (ehcache 3 latest version)
- Updated org.hibernate:hibernate-ehcache:5.3.8.Final with org.hibernate:hibernate-jcache:5.4.6.Final to support ehcache 3 and fix related to ehcache.xml file.
- Remove org.ehcache:jcache:1.0.0 as it is not required in ehcache 3.
- Updated ehcache.xml as per ehcahce 3 specifications.
- Updated persistence.xml to use ehcache 3 JCacheRegionFactory.
- Added hibernate.javax.cache.missing_cache_strategy = create to avoid warning related to missing cache definition in ehcache.xml.
- Updated JCacheManagerProvider to get CacheManager (created while JPA initialization) from CachingProvider
- In ehache 2.x, ManagementService is being used to register various mbeans like cachemanger, caches, cache configurations etc. However, this service is not available in ehcache 3.x. For now, updated CandlepinContextListener code to only register caches mbeans.
-  Move hibernate.javax.cache.uri property from persistence.xml to Candlepin.conf to reuse it in JCacheManagerProvider

### DIFF
--- a/server/src/main/java/org/candlepin/cache/JCacheManagerProvider.java
+++ b/server/src/main/java/org/candlepin/cache/JCacheManagerProvider.java
@@ -14,18 +14,33 @@
  */
 package org.candlepin.cache;
 
+import static org.candlepin.config.ConfigProperties.CACHE_CONFIG_FILE_URI;
+
+import org.candlepin.common.config.Configuration;
+
+import com.google.inject.Inject;
 import com.google.inject.Provider;
 
+import java.net.URISyntaxException;
+
+import javax.cache.CacheException;
 import javax.cache.CacheManager;
 import javax.cache.Caching;
 import javax.cache.spi.CachingProvider;
 
 /**
  * Provides object cache by indexed by String.
- * @author fnguyen
  *
+ * @author fnguyen
  */
 public class JCacheManagerProvider implements Provider<CacheManager> {
+
+    private Configuration config;
+
+    @Inject
+    JCacheManagerProvider(Configuration config) {
+        this.config = config;
+    }
 
     /**
      * It is safe to bind this as singleton, the CacheManager is supposed to
@@ -34,9 +49,20 @@ public class JCacheManagerProvider implements Provider<CacheManager> {
      */
     @Override
     public CacheManager get() {
-        CachingProvider cachingProvider = Caching.getCachingProvider();
-        CacheManager cacheManager = cachingProvider.getCacheManager();
-        return cacheManager;
-    }
+        CacheManager cacheManager = null;
 
+        CachingProvider cachingProvider = Caching.getCachingProvider();
+        String cacheManagerUri = config.getString(CACHE_CONFIG_FILE_URI);
+
+        try {
+            ClassLoader defaultClassLoader = cachingProvider.getDefaultClassLoader();
+            cacheManager = cachingProvider
+                .getCacheManager(defaultClassLoader.getResource(cacheManagerUri).toURI(), defaultClassLoader);
+
+            return cacheManager;
+        }
+        catch (URISyntaxException e) {
+            throw new CacheException("Couldn't load URI from " + cacheManagerUri, e);
+        }
+    }
 }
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -180,9 +176,13 @@ public void withInjector(Injector injector) {
         }
 
         if (config.getBoolean(ConfigProperties.CACHE_JMX_STATS)) {
-            MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
-            ManagementService.registerMBeans(CacheManager.getInstance(), mBeanServer,
-                true, true, true, true);
+            CacheManager cacheManager = injector.getInstance(CacheManager.class);
+            cacheManager.getCacheNames().forEach(name -> {
+                log.info("Enabling management and statistics for " + name + " cache");
```
Ceiu: Use the format syntax for log output:

```log.info("Enabling management and statistics for {} cache", name);```

File: server/src/main/java/org/candlepin/cache/JCacheManagerProvider.java [hunk 2]
```diff
@@ -16,17 +16,26 @@
...
 
 /**
  * Provides object cache by indexed by String.
- * @author fnguyen
  *
+ * @author fnguyen
  */
 public class JCacheManagerProvider implements Provider<CacheManager> {
 
+    private static final String CACHE_MANAGER_URI = "ehcache.xml";
```
Ceiu: If we're going to move the configuration file reference into code, we should follow the format we've done elsewhere and make it configurable via candlepin.conf.

server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
abhiskum: @Ceiu fixed.

server/src/main/java/org/candlepin/cache/JCacheManagerProvider.java
abhiskum: @Ceiu 
For now, I have moved ‘hibernate.javax.cache.uri’ property from persistence.xml and JCacheManagerProvider to Candlepin.conf file. However, I feel all related properties should be available in one place. Please suggest whether we should move the other properties related to caching to Candlepin.conf in the PR?

abhiskum:
@Ceiu 
Regarding the first line in the commit message, I wanted to mention this in PR comment instead of the commit message. I have removed it from the commit message. Thanks for noticing.

server/src/main/java/org/candlepin/cache/JCacheManagerProvider.java
Ceiu: I think for now, just the property being added is all we should focus on. Addressing config organization is a fairly large can of worms in itself that should be addressed at some point, but is substantial enough that it probably deserves its own discussion and PR.

How you've handled it seems fine to me, given that we do the same thing in other places. So long as the target file isn't hardcoded, but has a reasonable default in ConfigProperties, we're probably good for the foreseeable future.

That said, the default looks to be the only thing missing at this point. Just need to add that to the bottom blob in ConfigProperties and we should be good.

Ceiu:
That's what I assumed. Thanks for updating the commit message.

server/src/main/java/org/candlepin/cache/JCacheManagerProvider.java
abhiskum: @Ceiu Added default for the 'jpa.config.hibernate.javax.cache.uri' property.

PR #2469, opened 2019-10-18
[M] 1762815: Fixed NPE during system purpose status calculation (ENT-1705)

Ceiu:
- Fixed an issue that could cause a NullPointerException during
  system purpose status calculation


Ceiu:
Actually, upon thinking on it, I *really* don't like the multi-concat on the stream. Going to change that.

PR #2470, opened 2019-10-22
[ENT-1616] Remove branding property from pool and all direct references of this property 

wolfdale:
- Removed all branding property references from Pool


server/src/main/java/org/candlepin/dto/StandardTranslator.java
```diff
@@ -142,7 +140,7 @@ public StandardTranslator(ConsumerTypeCurator consumerTypeCurator,
         this.registerTranslator(
             new BrandingTranslator(), Branding.class, BrandingDTO.class);
         this.registerTranslator(
-            new ProductBrandingTranslator(), ProductBranding.class, BrandingDTO.class);
+            new BrandingTranslator(), Branding.class, BrandingDTO.class);
```
nikosmoum: This whole line can be removed, since the one above it does exactly the same thing. It was only temporarily added for the short period of the workaround of having 2 brandings (Branding and ProductBranding).

server/src/main/java/org/candlepin/dto/StandardTranslator.java
```diff
@@ -232,8 +230,8 @@ public StandardTranslator(ConsumerTypeCurator consumerTypeCurator,
             new org.candlepin.dto.manifest.v1.BrandingTranslator(),
             Branding.class, org.candlepin.dto.manifest.v1.BrandingDTO.class);
         this.registerTranslator(
-            new org.candlepin.dto.manifest.v1.ProductBrandingTranslator(),
-            ProductBranding.class, org.candlepin.dto.manifest.v1.BrandingDTO.class);
+            new org.candlepin.dto.manifest.v1.BrandingTranslator(),
```
nikosmoum: Same here, the previous line already registers the translator.

server/src/test/java/org/candlepin/dto/manifest/v1/BrandingTranslatorTest.java
```diff
@@ -66,8 +60,6 @@ protected void verifyOutput(Branding source, BrandingDTO dest, boolean childrenG
             assertEquals(source.getProductId(), dest.getProductId());
             assertEquals(source.getName(), dest.getName());
             assertEquals(source.getType(), dest.getType());
-            assertEquals(source.getUpdated(), dest.getUpdated());
-            assertEquals(source.getCreated(), dest.getCreated());
```
nikosmoum: The created/updated field testing should not be removed. They were accidentally not tested in the temporary ProductBrandingTranslatorTest.java file. In fact, they should also be added in the api/v1/BrandingTranslatorTest.java now that this was noticed.

server/src/main/java/org/candlepin/dto/StandardTranslator.java
wolfdale: good catch, Correcting it.

server/src/main/java/org/candlepin/dto/StandardTranslator.java
wolfdale: Correcting it.

PR #2472, opened 2019-10-23
Clarify large_msg_size artemis property usage

nikosmoum:
This is to enhance the documentation of the
candlepin.audit.hornetq.large_msg_size property by explaining
an issue that could occur if it is increased too much.


nikosmoum:
This is based on investigation for ENT-835

Ceiu:
ack

PR #2474, opened 2019-10-25
1691897: Candlepin should use cores as a default if vcpu is not set & cores is

Januson:
- Update rules so that pool uses cores for virtual consumers when the vcpu of pool is not set.


server/src/main/resources/rules/rules.js
```diff
@@ -317,6 +317,21 @@ function createPool(pool) {
         return poolSet;
     };
 
+    // When pool is missing vcpu and consumer is virtual, pool should use cores
+    // instead so we calculate calculate vcpu based on cores and sockets
+    pool.calculateVcpu = function (consumer) {
```
Ceiu: Does this function need to be attached to the specific pool?

server/src/main/resources/rules/rules.js
```diff
@@ -317,6 +317,21 @@ function createPool(pool) {
...
+        if (Utils.isGuest(consumer) && !this.hasAttribute(VCPU_ATTRIBUTE) && this.hasAttribute(CORES_ATTRIBUTE)) {
+            pool.attributes.push({
+                name: VCPU_ATTRIBUTE,
+                value: parseInt(this.getAttribute(CORES_ATTRIBUTE))
+            });
+        }
+    };
+
+    if (consumer != null) {
+        pool.calculateVcpu(consumer);
```
nikosmoum: Not something I feel strongly about, but if this is the only place this method is ever getting called from, then it doesn't have to be a method. You could simply put the method's code inline in this if statement.

server/src/main/resources/rules/rules.js
```diff
@@ -1,4 +1,4 @@
-// Version: 5.37
+// Version: 5.39
```
nikosmoum: Any reason 5.38 was skipped?

Januson: Just a typo.

server/src/main/resources/rules/rules.js
Januson: Done.

server/src/main/resources/rules/rules.js
Januson: No. I was originally calling it from other place and left it as a pool method. Inlined.

PR #2475, opened 2019-10-28
1466415: Candlepin should allow setting of default auto-attach value …

wottop:
…for hypervisors

### DIFF
--- a/server/src/main/java/org/candlepin/controller/Entitler.java
+++ b/server/src/main/java/org/candlepin/controller/Entitler.java
@@ -25,6 +25,8 @@ import org.candlepin.model.CandlepinQuery;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.ConsumerCurator;
 import org.candlepin.model.ConsumerInstalledProduct;
+import org.candlepin.model.ConsumerType;
+import org.candlepin.model.ConsumerTypeCurator;
 import org.candlepin.model.Content;
 import org.candlepin.model.Entitlement;
 import org.candlepin.model.EntitlementCurator;
@@ -74,6 +76,7 @@ public class Entitler {
 
     private Configuration config;
     private ConsumerCurator consumerCurator;
+    private ConsumerTypeCurator consumerTypeCurator;
     private ContentManager contentManager;
     private EventFactory evtFactory;
     private EventSink sink;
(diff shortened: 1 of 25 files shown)


server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -191,21 +199,35 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
     @Transactional
     public List<Entitlement> bindByProducts(AutobindData data, boolean force)
-        throws AutobindDisabledForOwnerException {
+        throws AutobindDisabledForOwnerException, AutobindHypervisorDisabledException {
         Consumer consumer = data.getConsumer();
         Owner owner = data.getOwner();
-
-        if ((!consumer.isDev() && owner.isAutobindDisabled()) || owner.isContentAccessEnabled()) {
+        ConsumerType type = this.consumerTypeCurator.getConsumerType(consumer);
+        if (!consumer.isDev() && (owner.isAutobindDisabled() ||
```
Ceiu: Is there any way we can pair down this beast of a conditional? Seems like we're doing a lot of very-specific context checking for this code.

If not, and if it's worth doing, we could add a simple utility function to ConsumerTypeEnum that could do the inverse check safely without needing to pretest for null:

```
public boolean isType(ConsumerType type) {
  return type != null && type.isType(this); // Alternatively, extend this out to not call into the type's isType method
}
```

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -191,21 +199,35 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+                " because of the hypervisor autobind setting" : "";
+            log.info("Skipping auto-attach for consumer '{}'. Auto-attach is disabled for owner {}{}{}",
+                consumer.getUuid(), owner.getKey(), caMessage, hypMessage);
+            if (!"".equals(hypMessage)) {
+                throw new AutobindHypervisorDisabledException(i18n.tr(
+                    "Auto-attach is disabled for owner \"{0}\"{1}.",
+                    owner.getKey(), hypMessage));
+            }
+            else {
+                throw new AutobindDisabledForOwnerException(i18n.tr(
```
Ceiu: We can save a few lines of code here by using a tertiary op:

```throw new AutobindHypervisorDisabledException(..., !hypMessage.isEmpty() ? hypMessage : caMessage);```

wottop: Those are different exceptions by design. Not sure we can use this. I'll change that condition.

Ceiu: Ahh, my brain autocorrected the exception names. Disregard.

server/src/main/java/org/candlepin/controller/Entitler.java
wottop: I made it easier to read. Good enough?

Ceiu: Yeah, that's much better. Thanks

PR #2479, opened 2019-10-31
ENT-497: Update CertificateSerial to no longer generate UUIDs on instantiation

abhiskum:
- Removed unique long id generation code from CertificateSerial constructor
 - Updated all the references to pass id while creating CertificateSerial object
 - Updated Junits related to this change

### DIFF
--- a/server/src/main/java/org/candlepin/model/CertificateSerial.java
+++ b/server/src/main/java/org/candlepin/model/CertificateSerial.java
@@ -14,17 +14,19 @@
  */
 package org.candlepin.model;
 
-import org.candlepin.util.Util;
 import org.candlepin.service.model.CertificateSerialInfo;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+
+import org.hibernate.annotations.GenericGenerator;
 
 import java.math.BigInteger;
 import java.util.Date;
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 import javax.persistence.Table;
 import javax.validation.constraints.NotNull;
@@ -43,6 +45,9 @@ public class CertificateSerial extends AbstractHibernateObject<CertificateSerial
 
     @Id
     @NotNull
+    @GeneratedValue(generator = "CertificateSerialIdGenerator")
+    @GenericGenerator(name = "CertificateSerialIdGenerator",
+        strategy = "org.candlepin.model.CertificateSerialIdGenerator")
     private Long id;
 
     @NotNull
@@ -61,11 +66,10 @@ public class CertificateSerial extends AbstractHibernateObject<CertificateSerial
      * Default constructor for serialization - DO NOT REMOVE!
      */
     public CertificateSerial() {
-        this.id = Util.generateUniqueLong();
+
     }
 
     public CertificateSerial(Date expiration) {
-        this();
         this.expiration = expiration;
     }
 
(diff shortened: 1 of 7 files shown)


Ceiu:
Have you explored ID auto generation here? It's a bit wonky compared to the typical classes due to the format of the ID and how we use it for two fields at the same time, but it might save on a ton of calls to ```Util.generateUniqueLong```.

abhiskum:
@Ceiu  I have not tried ID auto generator here. Are you suggesting to use ‘Util.generateUniqueLong’ as hibernate ID generator for this class or normal long ID generator?  Also, do you know, why are we generating long value in this way?

Ceiu:
Probably just standard long generator, so long as it uses SecureRandom or something equally unlikely to generate colliding longs.

abhiskum:
@Ceiu 
Default ID (IDENTITY, SEQUENCE) generator options are dependant on Database auto increment or sequence. i.e. need some additional changes. I tried with custom ID generator however few test cases are failing due to detached object related issues. 

I found another way to solve this issue using @PrePersist. It is working and also, we can reuse the same ID generation code. I have suggested the same to Sonali to use for Consumer UUID.   

Please share your thoughts on @PrePersist option. If you do not see any major issue then I will update the current fix to use @PrePersist option.

Ceiu:
The \@prepersist hook works for consumer, due to the column in question not being the primary key. For this class, we would be better served with a custom generator if Hibernate doesn't provide a suitable one for us.

```
public class CertSerialIdGenerator implements IdentifierGenerator {
  @Override
  public Serializable generate(SharedSessionContractImplementor session, Object object) {
    return Util.generateUniqueLong();
  }
}
```

```
@Id
@GeneratedValue(generator = "CertSerialIdGenerator")
@GenericGenerator(name = "CertSerialIdGenerator", strategy = "org.candlepin.package.CertSerialIdGenerator",
public Long id;
```

Ceiu:
Others seem to have a preference for the prepersist hook, so feel free to choose whichever implementation you prefer. @prepersist will probably be less code overall, since we won't be using this generation style anywhere else. I guess with that in mind, that probably makes more sense here.

abhiskum:
@Ceiu 
I agree ID generator is best for this use case. As I had mentioned in the previous comment, I tried the custom ID generator similar to you mentioned in the last comment. However, few tests are failing. I will try to fix this problem using an ID generator.

abhiskum:
@Ceiu 
Fixed this issue using ID generator approach. Please review now.

server/src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
```diff
@@ -449,6 +449,10 @@ public String cleanUpPrefix(String contentPrefix) throws IOException {
 
         log.debug("WE HAVE {} POOL QUANTITIES TO LOOP THROUGH", poolQuantities.size());
 
+        // Serials need to be saved to get generated ID.
+        log.debug("Persisting new certificate serials");
+        serialCurator.saveOrUpdateAll(serialMap.values(), false, false);
+
```
abhiskum: This change is required to fix one of Junit. I am not sure whether it is fine to change this Adapter class or not.

Ceiu: I stepped through the code and this change is fine. Amusingly, there is a comment a few lines up on 446 which calls out the original logic and why this wasn't done initially.

In addition to moving this block, that comment should be removed as well, as it is no longer accurate.

server/src/main/java/org/candlepin/model/CertificateSerial.java
```diff
@@ -61,7 +66,7 @@
      * Default constructor for serialization - DO NOT REMOVE!
      */
     public CertificateSerial() {
-        this.id = Util.generateUniqueLong();
+
```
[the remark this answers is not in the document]
abhiskum: Done.

server/src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
abhiskum: Done.

PR #2485, opened 2019-11-11
Move ActivationKeyRules to Java

nikosmoum:
- Move implementation of activation key rules from javascript
  to Java
- Move package org.candlepin.policy.js.activationkey one level up
  to org.candlepin.policy.activationkey
- Add Pool.getMergedProductAttribute/hasMergedProductAttribute
  methods which check for product attribute first, and pool
  attribute second
- ValidationResult now additionally supports the newly created
  RulesValidationError interface as error keys;
  This is a first step to transition all error/warning keys to Enums
  that implement that interface
- Add ActivationKeyRules.ErrorKeys enum to use as validation result
  instead of magic strings, for clarity/efficiency
- Rename methods in ActivationKeyRules.java for clarity and
  add javadoc comments

Note: this is the first of a series of changes to move all the
parts of rules.js to a Java implementation, in order to improve
maintainability/testability.


server/src/main/java/org/candlepin/policy/activationkey/ActivationKeyRules.java
```diff
@@ -0,0 +1,138 @@
...
+     * @return a result object which contains errors or warnings produced by the validation process.
+     *  In case the validation was successful, it will be empty.
+     */
+    public ValidationResult runPoolValidationForActivationKey(ActivationKey key, Pool pool, Long quantity) {
+        ValidationResult result = new ValidationResult();
+        if (quantity != null && quantity < 1) {
+            result.addError("rulefailed.invalid.quantity");
+        }
+
+        if (!"yes".equalsIgnoreCase(pool.getMergedProductAttribute(Pool.Attributes.MULTI_ENTITLEMENT))) {
```
Ceiu: I wonder if we shouldn't transition these magic strings to our stringified-boolean check that covers many affirmative string values (i.e. "yes", "y", "true", "t", "1", etc.). Such a thing would give us flexibility to standardize those values at some undetermined point in the future and remove a magic string from the code.

server/src/main/java/org/candlepin/policy/activationkey/ActivationKeyRules.java
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param quantity The quantity of the pool the key will consume
+     *
+     * @return a result object which contains errors or warnings produced by the validation process.
+     *  In case the validation was successful, it will be empty.
+     */
+    public ValidationResult runPoolValidationForActivationKey(ActivationKey key, Pool pool, Long quantity) {
+        ValidationResult result = new ValidationResult();
+        if (quantity != null && quantity < 1) {
+            result.addError("rulefailed.invalid.quantity");
```
Ceiu: Now that we're not passing the errors through the Java->JS layer, we should probably transition these to a proper enum. If we still need these error strings, we can include them as a property of the enum. As an added bonus, we'll get a tiny performance uplift in all the string checking when we build proper messages, as we could also attach *those* to the enums as like a "buildErrorMessage" method:
```
public enum ValidationErrors { // rename as appropriate
  ALREADY_EXISTS("rulefailed.already.exists", "Multi-entitlement not supported for pool \"{0}\"");
  // Add other errors here

  private final String code;
  private final String errmsg;

  ValidationError(String code, String errmsg) {
    this.code = code;
    this.errmsg = errmsg;
  }

  public String toString() {
    return this.code;
  }

  public String buildErrorMessage(I18n i18n, Object... args) {
    return i18n.tr(this.errmsg, args);
  }
}
```

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
```diff
@@ -227,8 +227,10 @@ public ActivationKeyDTO addPoolToKey(
         ActivationKey key = this.fetchActivationKey(activationKeyId);
         Pool pool = findPool(poolId);
 
-        // Throws a BadRequestException if adding pool to key is a bad idea
-        activationKeyRules.validatePoolForActKey(key, pool, quantity);
+        activationKeyRules.validatePoolForActivationKey(key, pool, quantity)
```
Ceiu: This is a very interesting way of handling this. I'm not quite sure how to feel about it yet, so let's leave it as is and see if it establishes a pattern for general output processing of the Java rules output. If so, keep them all; otherwise I would argue this is probably an expensive way of doing a null check.

server/src/main/resources/rules/rules.js
```diff
@@ -1491,10 +1491,14 @@ function isLevelExempt (level, exemptList) {
...
 
-/*
+/**
  * Code to validate pools before adding them to an activation key.
  * This way we can't create activation keys that can never successfully
  * register a consumer.
+ *
+ * DEPRECATED:
+ *  This functionality has been moved to ActivationKeyRules.java as of Candlepin 2.11.0. This method
+ *  will no longer be called by Candlepin and is considered deprecated.
```
Ceiu: :+1:

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
nikosmoum: Ondrej suggested it initially. After discussing with him again, we decided to revert it, since it doesn't gain us something in this case, while introducing a performance hit. And I wouldn't want for it to become a habit throughout the rest of the "move rules to Java".

server/src/main/java/org/candlepin/policy/activationkey/ActivationKeyRules.java
nikosmoum: Fixed.

server/src/main/java/org/candlepin/policy/activationkey/ActivationKeyRules.java
nikosmoum: Done. ValidationResult now will also accept Enums as error keys, and eventually, as more rules move to Java, we'll get rid of ValidationError/ValidationWarning classes altogether.

server/src/main/java/org/candlepin/policy/ValidationResult.java
```diff
@@ -27,13 +28,15 @@
 
     private List<ValidationError> errors;
     private List<ValidationWarning> warnings;
+    private List<Enum> errorKeys;
```
Ceiu: We could have our validation enums implement an interface here to keep it more limited than just any enum, and expose the buildErrorMessage method at this level.

server/src/main/java/org/candlepin/policy/ValidationResult.java
```diff
@@ -97,6 +117,14 @@ public boolean hasErrors() {
         return !errors.isEmpty();
```
Ceiu: This likely needs to return ```!errors.isEmpty() || !errorKeys.isEmpty();``` now, with another method to check for error strings. Though, I suppose if ```hasErrors``` returns true and ```hasErrorKeys``` returns false, we know that it has to be error strings.

nikosmoum: I actually envision the `errors` and `warnings` fields to eventually get removed from `ValidationResult` altogether, and errorKeys/warningKeys to be what stays on this class. The only reason they're not removed for now is because other rules use them. So I don't think we should mix logic of the two.

server/src/main/java/org/candlepin/policy/ValidationResult.java
nikosmoum: That is a good idea. I'll do that

nikosmoum: Done.

server/src/main/java/org/candlepin/policy/ValidationResult.java
nikosmoum: I've kept that separate for now (since ActivationKeyRules is only using the errorKeys, and other rules are only using the errors/warnings). If you feel strongly about not having that setup like this, I can create a new class that holds the enums (instead of highjacking ValidationResult).

Ceiu: I wonder if the future of the enum version is where the enum has a severity value associated with it, or a boolean to differentiate between warning/error. Or just a separate collection.

nikosmoum: I'd prefer separate collections, instead of always having to do validationError.isWarning() or validationError.getSeverity() on each of the items in the collection. They're used differently, so they should have different fields/methods.

PR #2486, opened 2019-11-12
Extract service layer controller from ActivationKeyResource

Januson:
Here is an example of how our Service Layer might look like. Please let me know what you think.

A few things I noticed during extraction:

 - We already have a package "controller" but the classes in it are our domain services. We might want some reorganization here so that we do not mix them. If we keep them separated we could theoretically enforce the direction of dependencies.

 - We do not have any resource specific tests. All of those tests target extracted logic. This is something we could use during SPEC test migration. We could have integration tests for stuff like response codes of our endpoints so that SPEC tests can focus on behavior.

 - Is CandlepinQuery a reasonable return type or do we want to keep it hidden and return collection instead?

- Our exceptions are coupled to the HTTP. They carry the Response Codes (Such as BadRequestException). Depending on how thorough we want to be with decoupling we might want to clean that up as well.


server/src/main/java/org/candlepin/controller/ActivationKeyController.java
```diff
@@ -0,0 +1,238 @@
...
+     */
+    private ActivationKey fetchActivationKey(String keyId) {
+        if (keyId == null || keyId.isEmpty()) {
+            throw new BadRequestException(i18n.tr("activation key ID is null or empty"));
+        }
+
+        ActivationKey key = this.activationKeyCurator.secureGet(keyId);
+
+        if (key == null) {
+            throw new BadRequestException(i18n.tr("ActivationKey with id {0} could not be found.", keyId));
```
Ceiu: The controllers shouldn't be throwing exceptions like this; that's up to the resource to translate.

Ceiu:
Also, do we want to merge into a feature branch first?

abhiskum:
@Januson 
Normally ‘Controller’ classes/package used to define REST layer API (like Resources classes/package in Candlepin) especially with respect to Spring web application or Spring boot. I would prefer to change ‘Controller’ with ‘Service’.

Ceiu:
I would argue #1 and 2 are a bit mixed. 

I feel the resources should do the initial validation + translation of the input data, hand off to the controller, then translate output (objects or exceptions) for the client.

nikosmoum:
@Ceiu I agree 1 and 2 can be mixed. That is why I'm asking, so we can clarify now how we want to separate the layers; we don't want to start implementing and introduce inconsistencies like that.
I am fine with the initial validation going either way (resource or controller), as long as it's consistent. You could differentiate the early/initial validation from the next task, not by using the no-db-access criterion, but by the fact that it is simple and not intertwined in the logic (e.g. the `fetchActivationKey` trying to resolve an existing activation key, and immediately throwing a REST exception with 404).

Also, what do you mean by 'translation of the input data'? Can you give an example?

Januson:
@Ceiu @nikosmoum @abhiskum 

- I would leave package reorg out for now. So that we do not have to deal with merge conflicts until we extract the classes.
- Created a task to deal with CandlepinQuery
- Reorganized exceptions into Candlepin domain exceptions and REST exceptions. Also added some handlers to translate domain exception into appropriate response
- I agree that we should do input validation before trying to reach the database. Currently we first check the input with @Verify and subsequently check it for null/empty. How do we want do approach this validation. Do we want to drop @Verify from resources?

Ceiu:
Even with @verify, we still need input validation since the current CP validation is little more than an existence/permissions check than one of proper input validation ala Spring validations. Whether or not we drop it is debatable. We would want to discuss this and how it'd impact our security model. If it's greenlit, I'd much rather see all the resources just accept the string IDs and do the lookup manually with a resource-standardized lookup (or even a resource-global/shared resolver tool).

@nikosmoum 
Sorry for the lengthy delay in answering this, but my vision of a resource following this model would look something like this:

```
@GET|PUT|POST
public EntityDTO doAThing(String entityId, <other args here>) {
  // Input validation bits here
  Entity entity = this.resolveEntity(entityId);
  // verify other args here as appropriate

  // Do the actual work via controller
  entity = this.entityController.doAThing(entity, args);

  // Translate the output to the API-specified model
  return this.modelTranslator.translate(entity, EntityDTO.class);
}
```

Now if the validation can eventually be performed via whatever auto-interface we end up using in the future, that's fine. However, I feel we'll still need to manually resolve the entities as they come in (security model concerns/restrictions not withstanding).

server/src/main/java/org/candlepin/controller/ActivationKeyController.java
```diff
@@ -0,0 +1,246 @@
...
+
+    public ActivationKey addPoolToKey(String activationKeyId, String poolId, Long quantity) {
+        ActivationKey key = this.fetchActivationKey(activationKeyId);
+        Pool pool = findPool(poolId);
+
+        // Throws a RuleValidationException if adding pool to key is a bad idea
+        String message = activationKeyRules.validatePoolForActivationKey(key, pool, quantity);
+        if (message != null) {
+            throw new RuleValidationException(message);
+        }
```
nikosmoum: The throwing of such an exception probably should be part of the rules API itself (thrown by `validatePoolForActivationKey`).

server/src/main/java/org/candlepin/util/ServiceLevelValidator.java
```diff
@@ -72,7 +71,7 @@ public void validate(String ownerId, Collection<String> serviceLevels) {
             String error = i18n.tr("Service level \"{0}\" is not available to units of organization {1}.",
                 StringUtils.join(invalidServiceLevels, ", "), owner.getKey());
 
-            throw new BadRequestException(error);
+            throw new RuleValidationException(error);
```
nikosmoum: This is a bit confusing place for this exception, since it doesn't come from any rules (`policy` package) execution. BadRequestException seemd a better fit.

server/src/main/java/org/candlepin/controller/ActivationKeyController.java
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.controller;
+
+import org.candlepin.common.exceptions.AlreadyRegisteredException;
+import org.candlepin.common.exceptions.NotFoundException;
+import org.candlepin.common.exceptions.RuleValidationException;
```
nikosmoum: It is not clear to me how, when the controller throws `AlreadyRegisteredException` and `NotFoundException`, they are handled? It should be the responsibility of the resource to turn those into resource exceptions, but ActivationKeyResource doesn't do that. Or at least a more general mechanism should exist for that conversion.

server/src/main/java/org/candlepin/async/tasks/HypervisorUpdateJob.java
```diff
@@ -23,7 +23,7 @@
 import org.candlepin.async.JobExecutionContext;
 import org.candlepin.async.JobExecutionException;
 import org.candlepin.auth.Principal;
-import org.candlepin.common.exceptions.BadRequestException;
+import org.candlepin.common.resource.exceptions.BadRequestException;
```
nikosmoum: Jobs and other non-resource parts of code shouldn't be throwing resource exceptions, but common ones. I assume this is a TODO right?

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
```diff
@@ -69,58 +58,23 @@
...
-        if (keyId == null || keyId.isEmpty()) {
-            throw new BadRequestException(i18n.tr("activation key ID is null or empty"));
-        }
-
-        ActivationKey key = this.activationKeyCurator.secureGet(keyId);
+    private static final Logger log = LoggerFactory.getLogger(ActivationKeyResource.class);
+    private static final String BAD_ACTIVATION_KEY_ID = "activation key ID is null or empty";
+    private static final String BAD_POOL_ID = "pool ID is null or empty";
+    private static final String BAD_PRODUCT_ID = "product ID is null or empty";
+    private static final String UPDATE_IS_NULL = "update is null";
```
Ceiu: Need to mark these as translatable for i18n: ```I18n.marktr```

common/src/main/java/org/candlepin/common/exceptions/ConcurrentModificationException.java
```diff
@@ -14,6 +14,8 @@
  */
 package org.candlepin.common.exceptions;
 
+import org.candlepin.common.resource.exceptions.ConflictException;
```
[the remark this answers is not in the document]
Januson: Done.

common/src/main/java/org/candlepin/common/exceptions/DeletedConsumerExceptionMessage.java
```diff
@@ -15,6 +15,8 @@
 package org.candlepin.common.exceptions;
 
 
+import org.candlepin.common.resource.exceptions.ExceptionMessage;
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/controller/ActivationKeyController.java
Januson: Do you want it just declared as throws or do you think it should be checked exception?

server/src/main/java/org/candlepin/util/ServiceLevelValidator.java
Januson: Changed to SlaValidationException.

server/src/main/java/org/candlepin/controller/ActivationKeyController.java
Januson: Exceptions are translated to appropriate responses using ExceptionMappers such as NotFoundExceptionMapper.

server/src/main/java/org/candlepin/async/tasks/HypervisorUpdateJob.java
Januson: Exactly. It should be done as we go over the resources, extracting the controllers.

server/src/main/java/org/candlepin/controller/ActivationKeyController.java
```diff
@@ -0,0 +1,246 @@
...
+    public ActivationKey getActivationKey(String activationKeyId) {
+        return this.fetchActivationKey(activationKeyId);
+    }
+
+    public Set<ActivationKeyPool> getActivationKeyPools(String activationKeyId) {
+        return this.fetchActivationKey(activationKeyId)
+            .getPools();
+    }
+
+    public ActivationKey updateActivationKey(String activationKeyId, ActivationKeyDTO update) {
```
barnabycourt: It seems strange to have DTOs which are presentation layer information mixed in with the Controller which is acting on internal models. I would expect the validation of the ActivationKeyDTO & the conversion to an ActivationKey to be taking place in the Resource while the update action for the model is using ActivationKey takes place in the controller.

server/src/test/java/org/candlepin/controller/ActivationKeyControllerTest.java
```diff
@@ -47,29 +46,27 @@
...
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 /**
- * ActivationKeyResourceTest
+ * ActivationKeyControllerTest
  */
-public class ActivationKeyResourceTest extends DatabaseTestFixture {
-    @Inject private ProductCurator productCurator;
+public class ActivationKeyControllerTest extends DatabaseTestFixture {
```
nikosmoum: As you have mentioned in comments above, we need proper Resource unit tests because our current *ResourceTests actually test the logic (which now resides in the controller). Having moved ActivationKeyResourceTest to ActivationKeyControllerTest, it is probably a good idea to create a proper ActivationKeyResourceTest.java now, that tests the resource stuff (for example, test the new validation `@NotEmpty, `@NotNull`, `@Positive` validation added).

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
```diff
@@ -321,32 +209,25 @@ public ActivationKeyDTO removeProductIdFromKey(
...
-            throw new BadRequestException(i18n.tr("Pool with id {0} could not be found.", poolId));
-        }
-
-        return pool;
+        this.activationKeyController.deleteActivationKey(activationKeyId);
     }
 
-    private Product confirmProduct(Owner o, String prodId) {
-        Product prod = this.ownerProductCurator.getProductById(o, prodId);
+    private ActivationKey toActivationKey(ActivationKeyDTO update) {
```
nikosmoum: Now that we have agreed to introduce automatic reverse-translation, this should be replaced by a new Translator class and call to that translator framework.

nikosmoum: Now that we have agreed to introduce automatic reverse-translation, this should be replaced by a new Translator class and call to that translator framework.

barnabycourt:
Added the "in progress" label back as this work is predicated on the spec-first DTO generation work due to all the translator updates needed.

wottop:
As this is going to need extensive changes before it is ready and relies on other work. We will close this for now but will reopen when it is updated and can be reviewed.

PR #2495, opened 2019-11-14
[F] Update job data serialization

Ceiu:
This is going to end up bigger than desired, so this is being broken up into several components. This PR begins addressing the serialization changes (which themselves are incomplete due to other issues pulling me in another direction for the moment).

The intended design behind these changes is as follows:
- Update the serialization for jobs such that (a) only primitives, strings, maps and collections are allowed to be serialized
- Remove the nested serialization, or change the individual components into discrete database columns which are individually serialized (probably faster overall) and deserialized on demand
- Update the job constraints to follow the registration + key lookup + stateless application model similar to jobs themselves, but without the per-invocation instantiation. When a constraint is checked, it should just be a matter of passing a few arguments to it and the list of active jobs.
- Add the per-job constraint configuration to the AsyncJobStatus entity somehow. At the time of writing, this is a map consisting of the constraint name mapped to a map of constraint arguments (Map<String, Map<String, Object>>). These should follow the same serialization rules.
- Add an additional table to the database: cp_blocked_jobs, with the following columns: blocked job id, blocking job id and blocking job state at the time of entry. When a job enters the BLOCKED state due to a constraint failure, a row should be inserted into this table for every job blocking it, along with their current state. Some diligence should be made to not consider jobs in terminal states to be blocking.
- Upon termination of any job (successful or otherwise), the cp_blocked_jobs table should be queried to find jobs that are (a) blocked on the job that just terminated and (b) are not blocked by any other job (that is, has no other rows in the table, or only has other rows in which the blocking job's state has changed since its entry). The jobs found should then have their constraints rechecked to determine whether or not they are unblocked. Unblocked jobs should be queued, though order is not guaranteed.

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobDataConverter.java
+++ b/server/src/main/java/org/candlepin/async/JobDataConverter.java
@@ -14,10 +14,28 @@
  */
 package org.candlepin.async;
 
+import org.candlepin.hibernate.AbstractJsonConverter;
+import org.candlepin.model.AsyncJobStatus.SerializedJobData;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.StringWriter;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
 
 
 
 public class JobDataConverter extends AbstractJsonConverter<SerializedJobData> {
+    private static Logger log = LoggerFactory.getLogger(JobDataConverter.class);
 
     private static final String METADATA_FIELD_NAME = "metadata";
     private static final String JOB_ARGUMENTS_FIELD_NAME = "job_arguments";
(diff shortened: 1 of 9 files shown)


barnabycourt:
@Ceiu Of the bullet items in the list, are all items included in this PR at this point or are some parts remaining to be done? i.e. If someone is reviewing the in progress work, what should they be focusing on?

Ceiu:
Serialization; and even that isn't 100%. 

As a note, the serialization change and the last bullet are going to make up the bulk of the work here. The rest is just simple bookkeeping and relatively simple work.

Ceiu:
Some things to consider on this:
- Do we still need JobArguments as a class? I'm finding myself going back and forth on it since its existence complicates things slightly (type/wrapper juggling), but it provides handy auto-casts and an entry point for on-demand deserialization when we get around to that in the future. If we opt to remove it, are we okay converting everything to just directly read from a generic map?
- Similar question for the constraints map, which is slightly more complex.

The real problem with these two maps comes from the general job lifecycle going from JobConfig to AsyncJobStatus. Passing the maps around means that the JobConfig.validate method has to work with one type, while (if they continue to exist), the job itself works with another. It also means extra work to handle the two types (wrapper and map) along the way.

server/src/main/java/org/candlepin/async/JobDataConverter.java
```diff
@@ -0,0 +1,358 @@
...
+
+
+
+public class JobDataConverter extends AbstractJsonConverter<SerializedJobData> {
+    private static Logger log = LoggerFactory.getLogger(JobDataConverter.class);
+
+    private static final String METADATA_FIELD_NAME = "metadata";
+    private static final String JOB_ARGUMENTS_FIELD_NAME = "job_arguments";
+    private static final String JOB_CONSTRAINTS_FIELD_NAME = "job_constraints";
+    private static final String JOB_OUTPUT_FIELD_NAME = "job_output";
```
nikosmoum: I'm generally opposed to job outputs having to be serialized/deserialized. Once a job is finished and its output is persisted, there is no more processing to be done on it (it is simply retrieved and sent to the client); isn't it a static piece of data whose structure we don't care from then on?

Ceiu: Yeah, probably. There's also some standardization that we could do with job output. With the exception of ImportJob and ExportJob, we don't use the output for much more than a sort of logging output to be sent with the job status. We could just as easily change the signature to be void and do what we used to do: setJobResult (though this has its own flow problems). At this point, we could apply the same type restrictions as everything else, serialize it on assignment, and then never touch it again.

Ceiu:
This is probably a dead PR, but I'll leave it up for another couple of working days to see if anyone has anything else they'd like to add or suggest. The serialization changes will probably be addressed in a later PR when time is less of a constraint.

PR #2497, opened 2019-11-15
[F] Removed pinsetter and pinsetter accessories (ENT-867)

Ceiu:


server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -180,21 +177,13 @@ private ConfigProperties() {
         DB_PASSWORD,
     };
 
-    // Pinsetter
-    public static final String TASKS = "pinsetter.tasks";
-    public static final String DEFAULT_TASKS = "pinsetter.default_tasks";
-    public static final String ENABLE_PINSETTER = "candlepin.pinsetter.enable";
```
nikosmoum: This property is also referenced in ./server/conf/candlepin.conf.template (and that is supposed to be set from the pinsetter_enabled property in ./server/build.gradle). So those entries should be removed. (Do we need to add entry for any of the new properties? It seems that we don't have a global equivalent property to enable/disable the asyn job system as a whole, and only job-specific, and white/black-list props?)

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -180,21 +177,13 @@ private ConfigProperties() {
...
-    public static final String PINSETTER_ASYNC_JOB_TIMEOUT = "pinsetter.waiting.timeout.seconds";
-    public static final String PINSETTER_MAX_RETRIES = "pinsetter.retries.max";
-    public static final int PINSETTER_MAX_RETRIES_DEFAULT = 10;
-
     public static final String[] DEFAULT_TASK_LIST = new String[] {};
 
     public static final String MANIFEST_CLEANER_JOB_MAX_AGE_IN_MINUTES =
```
nikosmoum: This is no longer used and can be removed . The property moved to `ManifestCleanerJob.CFG_MAX_AGE_IN_MINUTES` & `ManifestCleanerJob.DEFAULT_MAX_AGE_IN_MINUTES`, which is used by way of the whole job prefix/suffix system (e.g.  you'd have to write `candlepin.async.jobs.ManifestCleanerJob.max_age_in_minutes=1234` in the config to set this).

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -180,21 +177,13 @@ private ConfigProperties() {
...
 
-    // Pinsetter
-    public static final String TASKS = "pinsetter.tasks";
-    public static final String DEFAULT_TASKS = "pinsetter.default_tasks";
-    public static final String ENABLE_PINSETTER = "candlepin.pinsetter.enable";
-    public static final String PINSETTER_ASYNC_JOB_TIMEOUT = "pinsetter.waiting.timeout.seconds";
-    public static final String PINSETTER_MAX_RETRIES = "pinsetter.retries.max";
-    public static final int PINSETTER_MAX_RETRIES_DEFAULT = 10;
-
     public static final String[] DEFAULT_TASK_LIST = new String[] {};
```
nikosmoum: This can also be removed. It was the default task list of the old pinsetter jobs, which we removed one by one as we ported jobs.

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -180,21 +177,13 @@ private ConfigProperties() {
...
 
     public static final String MANIFEST_CLEANER_JOB_MAX_AGE_IN_MINUTES =
         "pinsetter.org.candlepin.pinsetter.tasks.ManifestCleanerJob.max_age_in_minutes";
 
     public static final String ENTITLER_JOB_THROTTLE =
-        "pinsetter." + EntitlerJob.class.getName() + ".throttle";
+        "pinsetter.org.candlepin.pinsetter.tasks.EntitlerJob.throttle";
```
nikosmoum: If we've already changed other job property names, then we no longer have to keep this legacy key. We can specify it using the prefix/suffix pattern (Similarly to what ManifestCleanerJob does; so it would become `candlepin.async.jobs.EntitlerJob.throttle`). For some reason we missed this when porting the EntitlerJob and its constraint.

nikosmoum:
Also, this PR looks like the work (or part of) of what has been described in ENT-867 and/or ENT-1698, so you might want to assign or alter those cards?

Ceiu:
The config stuff worries me a bit. I'm not sure if we *have* to maintain backward compatibility or not, but I feel like we do.

At the very least, I agree that the config templates should be updated accordingly and some of the other missed things should be axed. I'll have an update with those shortly.

server/src/main/java/org/candlepin/config/ConfigProperties.java
Ceiu: The "pinsetter enable" setting never actually disabled all of pinsetter, but rather only the scheduling components. I've added that back in via candlepin.async.scheduler.enabled and updated the template accordingly.

server/src/main/java/org/candlepin/config/ConfigProperties.java
Ceiu: done

server/src/main/java/org/candlepin/config/ConfigProperties.java
Ceiu: done

server/src/main/java/org/candlepin/config/ConfigProperties.java
Ceiu: done

PR #2499, opened 2019-11-19
1772194: use BYTES_TYPE for artemis messages

barnabycourt:
Previously we did not specify a type for artemis messages. This causes problems for non-native artemis clients. Specifically Katello is working on a move to use a STOMP based client to connect directly to Candlepin instead of requiring us to bounce the messages through QPID. When the default message type is used then the messages are prefixe with two Null characters and the STOMP client is unable to parse them. Moving to the BYTES_TYPE enables the STOMP client and does appear not have any effect on our serialization or deserialization.

### DIFF
--- a/server/src/main/java/org/candlepin/audit/EventSinkImpl.java
+++ b/server/src/main/java/org/candlepin/audit/EventSinkImpl.java
@@ -36,12 +36,11 @@ import org.apache.activemq.artemis.api.core.client.ClientSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.nio.charset.StandardCharsets;
 import java.util.LinkedList;
 import java.util.List;
 
 import javax.inject.Inject;
-
-import static org.apache.activemq.artemis.api.core.Message.BYTES_TYPE;
 
 /**
  * EventSink - Queues events to be sent after request/job completes, and handles actual
@@ -121,7 +120,7 @@ public class EventSinkImpl implements EventSink {
             if (messageSender == null) {
                 messageSender = new EventMessageSender(this.connection);
             }
-            messageSender.queueMessage(mapper.writeValueAsString(event));
+            messageSender.queueMessage(mapper.writeValueAsString(event), event.getType(), event.getTarget());
         }
         catch (Exception e) {
             log.error("Error while trying to send event", e);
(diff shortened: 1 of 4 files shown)


jturel:
Wth BYTES_TYPE there are preceding null characters that we have to strip out on the Katello side. I would rather not have to do any kind of modification to message bodies prior to converting them to JSON.

With TEXT_TYPE we don't receive any message body and I can not see the body in the management console which indicates to me that something is not quite right with this approach either. If the body were indeed text, why wouldn't the message be delivered correctly or display in the management UI?

Whatever the change ends up being in Candlepin I should be able to `JSON.parse(message.body)` on my side and for it to work as expected.

barnabycourt:
@jturel we are not injecting the extra characters into the JSON. This is not something that we control on the Candlepin side. you will have to parse these characters out on the Katello side.

server/src/main/java/org/candlepin/audit/EventSinkImpl.java
```diff
@@ -242,9 +241,17 @@ public EventMessageSender(ActiveMQConnection connection) {
...
-        public void queueMessage(String eventString) throws ActiveMQException {
-            ClientMessage message = session.createMessage(BYTES_TYPE, true);
-            message.getBodyBuffer().writeString(eventString);
+        public void queueMessage(String eventString, Event.Type type, Event.Target target) throws ActiveMQException {
+            ClientMessage message = session.createMessage(ClientMessage.BYTES_TYPE, true);
+            byte[] bytes = eventString.getBytes(StandardCharsets.UTF_8);
+            message.getBodyBuffer().writeBytes(bytes);
+            // Add Message type & target as message properties to allow better filtering/routing by consumers
+            if (type!= null) {
+                message.putStringProperty("MESSAGE_TYPE", type.name());
```
jturel: Beautiful!

jturel:
@bcourt since it's very likely for our first iteration we'll be using embedded artemis, what do think about adding the artemis-stomp-protocol as an included dependency? Very handy for us to not have to be landing extra jars anywhere

barnabycourt:
@jturel I'd be happy to add the jars for artemis-stomp but I believe that would be better suited to a 2nd PR as this one has already grown past the initial scope.

jturel:
> @jturel I'd be happy to add the jars for artemis-stomp but I believe that would be better suited to a 2nd PR as this one has already grown past the initial scope.

Makes sense. Want me to file something?

barnabycourt:
> > @jturel I'd be happy to add the jars for artemis-stomp but I believe that would be better suited to a 2nd PR as this one has already grown past the initial scope.
> 
> Makes sense. Want me to file something?

No need, I'll get that prepared separately.

jturel:
@barnabycourt @Ceiu 

I did an experiment to see how Core messages are received by a ruby STOMP and AMQP Client: https://github.com/jturel/artemis-interoperability

The nutshell is that `writeNullableSimpleString` / `readNullableSimpleString` seems to be the trick. The path here may be to switch use that, and to avoid breaking if there are messages in the queues on upgrade, check the message type:

```
if (msg.getType() == Message.DEFAULT_TYPE) {
  // use readString
} else { // it's a TEXT_TYPE
  // use readSimpleNullableString
}
```
```

Ceiu:
I've posted a different PR #2511 which addresses this issue using the logic from the simplestring PoC from @jturel above.

This may not be the best way to approach this situation, and perhaps a commit against this branch would have been better, but I was uncertain how to address it with outages affecting activity this week. I don't mean to step on any toes here with the PR change, so forgive me if that's how this comes across.

barnabycourt:
Closing in favor of PR #2511

PR #2500, opened 2019-11-19
[F] 1772194: use BYTES_TYPE for artemis messages

barnabycourt:
Previously we did not specify a type for artemis messages. This causes problems for non-native artemis clients. Specifically Katello is working on a move to use a STOMP based client to connect directly to Candlepin instead of requiring us to bounce the messages through QPID. When the default message type is used then the messages are prefixe with two Null characters and the STOMP client is unable to parse them. Moving to the BYTES_TYPE enables the STOMP client and does appear not have any effect on our serialization or deserialization.


Ceiu:
For this branch, we also need to add the ClientMessage.BYTES_TYPE constant to ArtemisSession.java:241

PR #2503, opened 2019-11-19
[F] Added the Transactional class and interfaces (ENT-1769)

Ceiu:
- Added the Transactional class and its nested interfaces for wrapping
  transaction-oriented logic
- Added a method to AbstractHibernateCurator to create transactional
  wrappers

### DIFF
--- a/server/src/main/java/org/candlepin/util/Transactional.java
+++ b/server/src/main/java/org/candlepin/util/Transactional.java
@@ -1,0 +1,323 @@
+/**
+ * Copyright (c) 2009 - 2017 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.util;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.LinkedList;
+import java.util.List;
+
+import javax.persistence.EntityManager;
+import javax.persistence.EntityTransaction;
+import javax.transaction.Status;
+
+
+
+/**
+ * The Transactional class represents a simple transaction wrapper around a given method or closure,
+ * as well as providing an interface for performing supplemental action upon completion of the
+ * transaction.
+ * <p></p>
+ * The wrapper can be reused within the context of a given operation if it is an operation that
+ * needs to be performed several times, perhaps with different inputs. To facilitate this, the
+ * arguments to pass into the action are provided on execution, rather than during configuration.
+ * This allows the following pattern:
+ *
+ * <pre>{@code
+ *  Transactional<String> transactional = this.curator.transactional(MyClass::some_action)
+ *      .onCommit(MyClass::commit_action)
+ *      .onRollback(MyClass::rollback_action)
+ *
+ *  for (String input : collection_of_inputs) {
+ *      String result = transactional.execute(input);
+ *  }
+ * }</pre>
+ *
+ * @param <O>
+ *  The output type of the action wrapped by this instance
+ */
+public class Transactional<O> {
+    private static Logger log = LoggerFactory.getLogger(Transactional.class);
+
+    /**
+     * An interface to execute an action within the bounds of a transaction
+     *
+     * @param <O>
+     *  the output type of this action
+     */
+    @FunctionalInterface
+    public interface Action<O> {
+
+        /**
+         * Called when a transaction this action is associated with has started.
+         *
+         * @param args
+         *  the arguments provided at the start of the transaction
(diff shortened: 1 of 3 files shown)


Ceiu:
While this is probably functional, at the time of writing there are no tests, the naming conventions could use some evaluation, the package may not be the best, and there are some improvements to be made around handling the end of a transaction (copying the Synchronization logic from JobManager to this).

Feel free to suggest improvements on any of the above.

server/src/main/java/org/candlepin/model/Transactional.java
```diff
@@ -0,0 +1,301 @@
...
+        EntityTransaction transaction = null;
+
+        try {
+            transaction = this.entityManager.getTransaction();
+
+            if (!transaction.isActive()) {
+                transaction.begin();
+            }
+            else {
+                log.warn("Transaction already started");
```
nikosmoum: Is warning here enough? Someone using this feature could assume that the action is happening in its own transaction, and not within a transaction that is being continued from before. I am not sure what the proper way to handle this is. We could try to commit the existing transaction here, before opening a new one, but that feels wrong. The cleanest would probably be to throw an exception. Another choice would be to make this a choice on the API (have a `reuseActiveTransaction` method, while the default behaviour would be to throw an exception instead). 

Whatever we choose, we should document this in Javadoc.

Ceiu: Yeah, I was leaning toward the latter, but opted to skip it for now due to wanting to get this done quickly. I'll add something here to control that a bit more.

server/src/main/java/org/candlepin/model/Transactional.java
```diff
@@ -0,0 +1,301 @@
...
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.model;
```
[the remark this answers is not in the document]
Ceiu: I don't predict this getting bigger or expanded upon once this PR goes through. Moving it to model.util is definitely better than leaving it in util though.

server/src/main/java/org/candlepin/model/Transactional.java
```diff
@@ -0,0 +1,301 @@
...
+ * needs to be performed several times, perhaps with different inputs. To facilitate this, the
+ * arguments to pass into the action are provided on execution, rather than during configuration.
+ * This allows the following pattern:
+ *
+ *  Transactional<String> transactional = this.curator.transactional(MyClass::some_action)
+ *      .onCommit(MyClass::commit_action)
+ *
+ *  for (String input : collection_of_inputs) {
+ *      String result = transactional.execute(input);
+ *  }
```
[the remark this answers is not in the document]
Ceiu: Done

server/src/main/java/org/candlepin/model/Transactional.java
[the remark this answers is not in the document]
nikosmoum: Actually, I meant to move it to `org.candlepin.util` (model.util was a typo).

Ceiu:
The Synchronization bits are the hooks into the transaction that let you handle commit/rollback asynchronously after it occurs. It's specific to the Transaction returned from Session (which has its own set of drawbacks), and it may or may not apply cleanly to this utility. I played with it a bit yesterday, and eventually dropped it since it didn't appear to add anything of real value.

Ceiu:
Alright, tests added and I added support for "exclusive" mode transaction handling (basically: the wrapper handles it or complains loudly).

I think I covered all of the usage cases, but if I missed something, feel free to call it out.

server/src/test/java/org/candlepin/util/TransactionalTest.java
```diff
@@ -0,0 +1,340 @@
...
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.util;
+
+import static org.hamcrest.MatcherAssert.*;
+import static org.hamcrest.Matchers.*;
```
nikosmoum: These 2 hamcrest imports are never used, but for some reason checkstyle doesn't complain (something to do with wildcard being used?).

server/src/main/java/org/candlepin/util/Transactional.java
```diff
@@ -0,0 +1,323 @@
...
+        catch (Exception e) {
+            if (!this.commitOnException) {
+                transaction.setRollbackOnly();
+            }
+
+            throw e;
+        }
+        finally {
+            if (!transaction.isActive()) {
+                String errmsg = "Transactional block completed without an active transaction";
```
nikosmoum: It would be good to add a test case that covers this block (e.g. when an action itself has already commited or rolled back the current transaction), just to be safe.

Ceiu: Added

server/src/test/java/org/candlepin/util/TransactionalTest.java
Ceiu: Yeah, I think the wildcards break the import checker (though the value of that over just using wildcards is questionable).

Anyway, removed.

PR #2504, opened 2019-11-20
ENT-1701; Containers now use gradle instead of buildr

nikosmoum:
- Building/Running the gating test containers will always use gradle
  from now on (those use the candlepin version used by Hosted, which
  will always have gradle from now on).
- Building the base container (used to run tests in jenkins) will
  now use gradle from now on. The scripts running within that
  container will keep using either buildr or gradle depending on
  the branch that is checked out.
- Fix bug in cpsetup that would misconfigure tomcat to only support
  TLSv1 (instead of additionally supporting TLSv1.1 and TLSv1.2).

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -130,7 +130,7 @@ cleanup() {
         trap - $signals
         trap "" ERR EXIT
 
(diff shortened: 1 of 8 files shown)


Ceiu:
ack

PR #2505, opened 2019-11-20
[ENT-1670] 1640822: [RFE] Change or switch content view from content host

wolfdale:
- Allowing updating consumer environment by env name. Modified consumer PUT update API.
- Added spec test for different scenarios.

### DIFF
--- a/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
+++ b/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
@@ -40,6 +40,7 @@ import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Projection;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
+
 import org.hibernate.internal.CriteriaImpl;
 import org.hibernate.internal.SessionImpl;
 import org.hibernate.metadata.ClassMetadata;
@@ -64,8 +65,9 @@ import javax.persistence.LockModeType;
 import javax.persistence.NonUniqueResultException;
 import javax.persistence.OptimisticLockException;
 import javax.persistence.TypedQuery;
-
-
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.Root;
+import javax.persistence.criteria.CriteriaQuery;
 
 /**
  * AbstractHibernateCurator base class for all Candlepin curators. Curators are
@@ -169,6 +171,52 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
     @Transactional
     public E get(Serializable id) {
         return id == null ? null : this.get(entityType, id);
+    }
+
+    /**
+     * Checks if entry exists in db.
+     * @param id primary key of entity.
+     * @return boolean value whether row exists or not
+     */
+    @Transactional
+    public boolean exists(Serializable id) {
+        boolean doesExists = false;
+        String columnName = getPrimaryKeyName();
+
+        if (columnName != null) {
+            CriteriaBuilder builder = this.getEntityManager().getCriteriaBuilder();
+            CriteriaQuery<Long> query = builder.createQuery(Long.class);
+            query.select(builder.count(query.from(entityType)));
+            Root<E> root = query.from(entityType);
+            query.where(builder.equal(root.get(columnName), id));
+
+            if (this.getEntityManager().createQuery(query).getSingleResult() > 0) {
+                doesExists = true;
+            }
+        }
+
+        return doesExists;
+    }
+
+    /**
+     * Returns name of primary key column
+     * @return Returns entity primary key column
+     */
+    public String getPrimaryKeyName() {
+        String primaryColumn = null;
+
+        try {
+            primaryColumn = this.getEntityManager()
+                .getMetamodel()
+                .entity(entityType)
+                .getId(String.class)
+                .getName();
+        }
+        catch (IllegalArgumentException e) {
+            log.debug("Unable to get primary key for entity {}", entityType);
+        }
+
+        return primaryColumn;
     }
 
     /**
--- a/server/src/main/java/org/candlepin/model/EnvironmentCurator.java
+++ b/server/src/main/java/org/candlepin/model/EnvironmentCurator.java
@@ -26,7 +26,7 @@ import java.util.Map;
 import java.util.Set;
 
 import javax.inject.Singleton;
-
+import javax.persistence.NoResultException;
 
 
 /**
@@ -171,5 +171,32 @@ public class EnvironmentCurator extends AbstractHibernateCurator<Environment> {
         }
 
         return count;
+    }
+
+    /**
+     * Get environment Id by owner id and environment name.
+     *
+     * @param ownerId owner Id.
+     * @param envName environment name.
+     * @return environment Id
+     */
+    public String getEnvironmentIdByName(String ownerId, String envName) {
+        String jpql = "SELECT e.id FROM Environment e WHERE e.owner.id = :owner_id and e.name = :env_name";
+        String envId = null;
+
+        if (ownerId != null && envName != null) {
+            try {
+                envId = this.getEntityManager()
+                    .createQuery(jpql, String.class)
+                    .setParameter("owner_id", ownerId)
+                    .setParameter("env_name", envName)
+                    .getSingleResult();
+            }
+            catch (NoResultException exp) {
+                log.trace("Unable to find environment Id by envName {}", envName);
+            }
+        }
+
+        return envId;
     }
 }
--- a/server/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -1316,27 +1316,7 @@ public class ConsumerResource {
 
         changesMade = updateSystemPurposeData(updated, toUpdate) || changesMade;
 
-        String environmentId = updated.getEnvironment() == null ? null : updated.getEnvironment().getId();
-        if (environmentId != null && (toUpdate.getEnvironmentId() == null ||
-            !toUpdate.getEnvironmentId().equals(environmentId))) {
-            Environment e = environmentCurator.get(environmentId);
-            if (e == null) {
-                throw new NotFoundException(i18n.tr(
-                    "Environment with ID \"{0}\" could not be found.", environmentId));
-            }
-            log.info("Updating environment to: {}", environmentId);
-            toUpdate.setEnvironment(e);
-
-            // reset content access cert
-            Owner owner = ownerCurator.findOwnerById(toUpdate.getOwnerId());
-            if (owner.isContentAccessEnabled()) {
-                toUpdate.setContentAccessCert(null);
-                contentAccessCertService.removeContentAccessCert(toUpdate);
-            }
-            // lazily regenerate certs, so the client can still work
-            poolManager.regenerateCertificatesOf(toUpdate, true);
-            changesMade = true;
-        }
+        changesMade = updateEnvironment(updated, toUpdate) || changesMade;
 
         // like the other values in an update, if consumer name is null, act as if
         // it should remain the same
@@ -1392,6 +1372,64 @@ public class ConsumerResource {
 
             Event event = eventBuilder.setEventData(toUpdate).buildEvent();
             sink.queueEvent(event);
+        }
+
+        return changesMade;
+    }
+
+    /**
+     * Updates consumer environment either by environment name or id, reset & re-generate CA certs.
+     * @return boolean status whether any updates are made or not.
+     */
+    private boolean updateEnvironment(ConsumerDTO updated, Consumer toUpdate) {
+        if (updated.getEnvironment() == null) {
+            return false;
+        }
+
+        boolean changesMade = false;
+        String environmentId = updated.getEnvironment() == null ? null : updated.getEnvironment().getId();
+        String environmentName = updated.getEnvironment() == null ? null : updated.getEnvironment().getName();
+        String validatedEnvId = null;
+
+        if (environmentId != null) {
+            if (toUpdate.getEnvironmentId() == null || !toUpdate.getEnvironmentId().equals(environmentId)) {
+
+                if (!environmentCurator.exists(environmentId)) {
+                    throw new NotFoundException(i18n.tr(
+                        "Environment with ID \"{0}\" could not be found.", environmentId));
+                }
+
+                validatedEnvId = environmentId;
+            }
+
+        }
+        else if (environmentName != null) {
+            String envId = this.environmentCurator.
+                getEnvironmentIdByName(toUpdate.getOwnerId(), environmentName);
+
+            if (envId == null) {
+                throw new NotFoundException(i18n.tr(
+                    "Environment with name \"{0}\" could not be found.", environmentName));
+            }
+
+            if (toUpdate.getEnvironmentId() == null || !envId.equals(toUpdate.getEnvironmentId())) {
+                validatedEnvId = envId;
+            }
+        }
+
+        if (validatedEnvId != null) {
+            toUpdate.setEnvironmentId(validatedEnvId);
+            // reset content access cert
+            Owner owner = ownerCurator.findOwnerById(toUpdate.getOwnerId());
+
+            if (owner.isContentAccessEnabled()) {
+                toUpdate.setContentAccessCert(null);
+                contentAccessCertService.removeContentAccessCert(toUpdate);
+            }
+
+            // lazily regenerate certs, so the client can still work
+            poolManager.regenerateCertificatesOf(toUpdate, true);
+            changesMade = true;
         }
 
         return changesMade;
--- a/server/spec/consumer_resource_spec.rb
+++ b/server/spec/consumer_resource_spec.rb
@@ -163,6 +163,58 @@ describe 'Consumer Resource' do
     results[@consumer1.uuid].should_not be_nil
   end
 
+  it 'should not let consumer update environment with incorrect env name' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+
+    expect{
+      consumer_client.update_consumer({:environment => {:name => "abc"}})
+    }.to raise_error(RestClient::ResourceNotFound)
+  end
+
+  it 'should let consumer update environment with valid env name only' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'],
+      consumer['idCert']['key'])
+    env = @cp.create_environment(@owner1['key'], random_string("env_id"), random_string("env_name"))
+
+    expect(consumer.environment).to be_nil
+
+    consumer_client.update_consumer({:environment => {:name => env.name}})
+    consumer = @cp.get_consumer(consumer['uuid'])
+
+    expect(consumer.environment).to_not be_nil
+  end
+
+  it 'should let consumer update environment with valid env id only' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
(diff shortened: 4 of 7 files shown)


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1397,6 +1377,63 @@ public boolean performConsumerUpdates(ConsumerDTO updated, Consumer toUpdate,
...
+                environmentName).list();
+
+            if (!environment.isEmpty() && environment.size() == 1) {
+                e = environment.get(0);
+                environmentId = environment.get(0).getId();
+                environmentFound = true;
+            }
+            else {
+                throw new NotFoundException(i18n.tr(
+                        "Environment with name \"{0}\" could not be found.", environmentName));
```
Ceiu: Indentation is too long here. If checkstyle is complaining, drop the i18n.tr down a line with it. Same with the identical line below.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1397,6 +1377,63 @@ public boolean performConsumerUpdates(ConsumerDTO updated, Consumer toUpdate,
...
+        boolean environmentFound = false;
+        Environment e = null;
+        String environmentId = updated.getEnvironment() == null ? null : updated.getEnvironment().getId();
+        String environmentName = updated.getEnvironment() == null ? null : updated.getEnvironment().getName();
+
+        if (environmentName != null && environmentId == null) {
+            List<Environment> environment = environmentCurator.listForOwnerByName(toUpdate.getOwner(),
+                environmentName).list();
+
+            if (!environment.isEmpty() && environment.size() == 1) {
```
Ceiu: This is a redundant check. If the environment collection size is 1, we know it's not empty. Since we're not differentiating between empty and more-than-one in the error output, we can simplify this down to just the size check.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1397,6 +1377,63 @@ public boolean performConsumerUpdates(ConsumerDTO updated, Consumer toUpdate,
...
+                throw new NotFoundException(i18n.tr(
+                        "Environment with name \"{0}\" could not be found.", environmentName));
+            }
+
+        }
+
+        if (environmentId != null && (toUpdate.getEnvironmentId() == null ||
+            !toUpdate.getEnvironmentId().equals(environmentId))) {
+
+            if (!environmentFound) {
```
Ceiu: environmentFound will never be true at this point, so this check doesn't do anything. If the intent is only to do this lookup if the environment has not yet been looked up (i.e. we already did that lookup by name), then this whole section needs a refactor around that idea.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1397,6 +1377,63 @@ public boolean performConsumerUpdates(ConsumerDTO updated, Consumer toUpdate,
...
+                            "Environment with ID \"{0}\" could not be found.", environmentId));
+                }
+            }
+
+            changesMade = true;
+
+            log.info("Updating environment to: {}", environmentId);
+        }
+
+        if (e != null && changesMade) {
```
Ceiu: I don't quite follow the intent behind changesMade here. It's set only in one of the lookup cases above, but it's not associated with any actual change. This seems like something that should be set only if an environment was found; but since we throw an exception in both lookup cases, this only serves the purpose of indicating if (a) no environment reference was provided or (b) a reference was provided *and* a change was made as a result. Following that, this should only be set in the block it's currently guarding.

Looking at the original code, that seems to be what it was doing before.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1397,6 +1377,63 @@ public boolean performConsumerUpdates(ConsumerDTO updated, Consumer toUpdate,
...
+
+            // reset content access cert
+            Owner owner = ownerCurator.findOwnerById(toUpdate.getOwnerId());
+            if (owner.isContentAccessEnabled()) {
+                toUpdate.setContentAccessCert(null);
+                contentAccessCertService.removeContentAccessCert(toUpdate);
+            }
+            // lazily regenerate certs, so the client can still work
+            poolManager.regenerateCertificatesOf(toUpdate, true);
+
```
Ceiu: Move this whitespace up two lines to logically separate the regeneration section and the CAC section.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1397,6 +1377,63 @@ public boolean performConsumerUpdates(ConsumerDTO updated, Consumer toUpdate,
...
+            }
+            else {
+                throw new NotFoundException(i18n.tr(
+                        "Environment with name \"{0}\" could not be found.", environmentName));
+            }
+
+        }
+
+        if (environmentId != null && (toUpdate.getEnvironmentId() == null ||
+            !toUpdate.getEnvironmentId().equals(environmentId))) {
```
Ceiu: We can pair this down a bit to drop an extraneous null check:

```if (environmentId != null && !environmentId.equals(toUpdate.getEnvironmentId())) {```

File: server/src/main/java/org/candlepin/resource/ConsumerResource.java (hunk 1 above)
Comment (wolfdale) [reply server/src/main/java/org/candlepin/resource/ConsumerResource.java]: No checkstyle complaints.

File: server/src/main/java/org/candlepin/resource/ConsumerResource.java (hunk 2 above)
Comment (wolfdale) [reply server/src/main/java/org/candlepin/resource/ConsumerResource.java]: yes sure.

File: server/src/main/java/org/candlepin/resource/ConsumerResource.java (hunk 5 above)
Comment (wolfdale) [reply server/src/main/java/org/candlepin/resource/ConsumerResource.java]: corrected.

**Comment by wolfdale** [comment]
Earlier the incoming envId is compared to existing envId (if any), if these both ids are different only then we set the env and update CAC & re-generate certs.
What I am trying to do is to save extra DB call & prevent unnecessary updates every time if API keep sending same environment name again & again. This is what I have in my mind.

```
Check if envName is not null && envId is null.
  if YES
    Get environment from DB.
    set environment Id , set environment object
    environmentFound = true (Preventing another DB call, if incoming envId is different from what we already have)

    Go on to next IF condition.

Check if envId is not null (This prevent when both envName and envId are null) && envId is different than what we already have.
  
  if YES
    
    if environmentFound = fasle
      Get ENV from DB
    
    Mark changesMade = true

if ChangesMade = true
  Update CAC and re-gen certs.
```

File: server/src/main/java/org/candlepin/resource/ConsumerResource.java (hunk 4 above)
Comment (wolfdale) [reply server/src/main/java/org/candlepin/resource/ConsumerResource.java]: I removed `e != null` check. It seems reasonable that we update only when changesMade flag is true.

File: server/src/main/java/org/candlepin/resource/ConsumerResource.java (hunk 1 above)
Comment (Ceiu) [reply server/src/main/java/org/candlepin/resource/ConsumerResource.java]: I meant if checkstyle complains about it not being indented enough. The indentation check is pretty spotty at times.

**Comment by Ceiu** [comment]
That's good that you're aware of the cost of doing needless database lookups, but let's step through the intent and code to clean this up a bit:

> Check if envName is not null && envId is null.

Straight away you're establishing a priority of attributes to use to lookup. By checking that envId is null, you're asserting that if it's *not* null, we should use that first, and fallback to name only if the ID isn't present. The priority here is important, as we need to establish some rules for the situation when both are provided. More on this later.

>   if YES
>    Get environment from DB.
>    set environment Id , set environment object
>    environmentFound = true (Preventing another DB call, if incoming envId is different from what we already have)
>
>    Go on to next IF condition.

We do the lookup, but then we store a bunch of redundant information -- even at the pseudocode level. If we have the ID to perform a lookup, there isn't much value in storing it in another variable; especially since it isn't expensive to fetch. Additionally, we don't need to store whether or not we found the environment, since the simple presence of an environment entity provides that naturally.

> Check if envId is not null (This prevent when both envName and envId are null) && envId is different than what we already have.

As hinted above, this is the interesting part here. We check the environment ID again to avoid a lookup, but we also verify that it's actually different to skip doing any work if the environment ID is unchanged. While this is fine, this gets a little tricky since we're bolting on a lookup ahead of it. This is the cornerstone of the method in both core functionality and optimization. However, the optimization may be premature here. More on this later.

>  if YES
>    if environmentFound = fasle
>      Get ENV from DB

Continuation of the resolution step, nothing major here other than using the variable we don't actually need. This part is fine in pseudocode, but in real code we can just use the environment object itself.

>    Mark changesMade = true
>
> if ChangesMade = true
>  Update CAC and re-gen certs.

This bit is a tad confusing. If you follow the code, it's understandable, but the naming convention chosen here is poor. We haven't actually made any changes yet, so by flagging it as such, anyone reading the code needs to understand that this is setting up a promise/contract, and that it's following the usage convention of the intended caller rather than the owning method. That last bit is backward in my opinion, and the caller of a utility method should have very little influence over the conventions applied by the method in question.

---

Alright, so with all of that out of the way, we can compare that to the code and what we actually *need* to do.

First, the bulk of this method surrounds environment resolution and validation. We're spending a great deal of effort here to avoid doing a full lookup if possible. That's good, but the optimizations chosen are starting to mix the operations. Moreover, the pseudocode does not make it clear that we have a specific order in which the attributes are ranked *and* that if both are provided we ignore the name -- even if it would lead to cases where the client has specified an invalid pairing.

Secondly, we perform the actual update. Looking at what we actually do with the bits we looked up, we can see that we don't actually need the whole environment, so the work we're doing is very heavy when all we want is a valid environment ID. Beyond that, the extra work we do is ancillary state management when the environment changes, so everything comes down to an ID comparison.

With all that said, we can rearrange the pseudocode as follows:

```
Check if env ID is present
Yes:
  check if specified ID exists
  throw REST exception if it does not
  
Check if we don't yet have a validated env ID AND env name is present
Yes:
  fetch environment ID by org+name
  throw REST exception if it does not exist

Check if we have a validated environment ID
Yes:
  compare to assigned environment ID
  if different:
    update consumer
    update CAC
    mark certs dirty
    set updated flag to true
  
return updated flag
```

This is very close to the work you had, but it rearranges everything in both priority and logical order without mixing the two during optimization. You'll note that we now require two additional DB operations, but that's not really a big deal -- one of them arguably should have already existed, and the other is just the cost of cleaning this code up.

Translated to code you'll get something along the lines of this:
```
public boolean updateEnvironment(ConsumerDTO update, Consumer toUpdate) {
  if (update.getEnvironment() == null) {
    // No updates being made here, don't bother attempting anything else
    return false;
  }

  boolean changesMade = false;
  String validatedEnvId = null;  
  String envId = update.getEnvironment().getId();
  String envName = update.getEnvironment().getName();
  
  if (envId != null) {
    if (!this.environmentCurator.exists(envId) {
      throw new NotFoundException(...);
    }
    
    validatedEnvId = envId;
  }
  
  if (validatedEnvId == null && envName != null) {
    validatedEnvId = this.environmentCurator.getEnvironmentIdByName(org, envName);
    
    if (validatedEnvId == null) {
      throw new NotFoundException(...);
    }
  }
  
  if (validatedEnvId != null && !validatedEnvId.equals(toUpdate.getEnvironmentId())) {
    ...
    
    changesMade = true;
  }
  
  return changesMade;
}
```

And if you *really* want to optimize out that existence check, you'd modify the first block as follows:

```
if (envId != null) {
  if (!envId.equals(toUpdate.getEnvironmentId()) && !this.environmentCurator.exists(envId) {
    throw new NotFoundException(...);
  }
  
  validatedEnvId = envId;
}
```

From here, the two DB calls would need to be added to AbstractHibernateCurator and EnvironmentCurator, respectively. Both can be done pretty quickly using a one-liner JPQL query.

**Comment by Ceiu** [comment]
One thing I forgot to mention that you should be aware of, and why the conflating of the two operations is kind of bad: right now the resource is acting as a resource and controller. As work continues to separate those out, the validation bits (where the REST exceptions originate) will be done in the resource and the actual update bits will occur in the controller. Exactly how we go about that will vary as we hit each resource, but we need to be mindful of that separation more than we have been in the past.

**Comment by wolfdale** [comment]
One corner cases is missed here, if the environment is to be updated for first time.
Added another check for it, at line 1395.
```
if (environmentId != null && (toUpdate.getEnvironmentId() == null ||
            !toUpdate.getEnvironmentId().equals(environmentId))) {
```
`This is very close to the work you had, but it rearranges everything in both priority and logical order without mixing the two during optimization. `
I did exact opposite of it, by prioritizing envName, which seems to work but was not cleaner approach. 

`You'll note that we now require two additional DB operations, but that's not really a big deal -- one of them arguably should have already existed, and the other is just the cost of cleaning this code up.
`
Earlier I used older Curator method just for code re-usability to avoid any redundant DB operations.

Also I have added a `exists` method at AbstractHibernateCurator. Notice that its a find operation, was trying with `entitymanager.getReference()` method but it return a proxy object and will not hit DB unit any GET method is called on it. Any suggestion here?

```
    public boolean exists(Serializable id) {
        return this.getEntityManager().find(entityType, id) != null;
    }
```
Also would request you to not post code it defeat the purpose to explore on different viewpoint.

**Comment by Ceiu** [comment]
> I did exact opposite of it, by prioritizing envName,
Something important to note, that while you had the name being *processed* first, because the processing was contingent on the ID being null. Thus, while the name was processed first in the code, it had lower logical priority than the ID.

For the exists method, we want to avoid fetching the object if at all possible, since we (a) don't need any of the data and (b) don't want to trigger secondary lookups for nested collections or whatever. Following that, we probably want to steer closer to doing a COUNT-based lookup. We can do this in one of two ways: JPQL or JPA criteria query. The latter is probably safest so we're not doing our own string concatenation, but either is acceptable.

The one nibble here is that at the AHC level, we don't know *exactly* what the ID column name is, so we'll have to fetch it from the entity metadata. Beyond that, this is just a simple expression of the SQL ```SELECT count(id) FROM cp_entity_table WHERE id = :id```.

File: server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java [hunk 7]
```diff
@@ -171,6 +170,49 @@ public E get(Serializable id) {
...
+        }
+
+        return doesExists;
+    }
+
+    /**
+     * Returns name of primary key column
+     * @return Returns entity primary key column
+     */
+    public String getPrimaryColumnName() {
```
Ceiu: We should follow Hibernate/JPA or SQL naming conventions here: either ```getIdentifierName``` or ```getPrimaryKeyName```.

wolfdale: done.

server/src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -172,4 +172,29 @@ public int deleteEnvironmentsForOwner(Owner owner) {
...
+        String envId = null;
+
+        if (ownerId != null && envName != null) {
+            try {
+                envId = this.getEntityManager().createQuery(jpql, String.class)
+                        .setParameter("owner_id", ownerId).setParameter("env_name", envName)
+                        .getSingleResult();
+            }
+            catch (NoResultException exp) {
+                log.debug("Unable to find environment Id by envName {}", envName);
```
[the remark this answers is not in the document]
wolfdale: Yes, agreed.

server/src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -172,4 +172,29 @@ public int deleteEnvironmentsForOwner(Owner owner) {
...
+     */
+    public String getEnvironmentIdByName(String ownerId, String envName) {
+        String jpql = "SELECT e.id FROM Environment e WHERE e.owner.id = :owner_id and e.name = :env_name";
+        String envId = null;
+
+        if (ownerId != null && envName != null) {
+            try {
+                envId = this.getEntityManager().createQuery(jpql, String.class)
+                        .setParameter("owner_id", ownerId).setParameter("env_name", envName)
+                        .getSingleResult();
```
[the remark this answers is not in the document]
wolfdale: Sure, I think this could be included in checkstyle.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1397,6 +1377,64 @@ public boolean performConsumerUpdates(ConsumerDTO updated, Consumer toUpdate,
...
+            }
+        }
+
+        if (validatedEnvId != null && !validatedEnvId.equals(toUpdate.getEnvironmentId())) {
+            e = environmentCurator.get(validatedEnvId);
+            changesMade = true;
+        }
+
+        if (changesMade) {
+            toUpdate.setEnvironment(e);
```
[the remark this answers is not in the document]
wolfdale: This was the part of old code. But it does make sense to update envId only. Fixed it.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -171,6 +170,49 @@ public E get(Serializable id) {
...
+     */
+    @Transactional
+    public boolean exists(Serializable id) {
+        String columnName = null;
+        boolean doesExists = false;
+        columnName = getPrimaryColumnName();
+
+        if (columnName != null) {
+            String jpql = "SELECT count(t) FROM " + entityType.getCanonicalName() +
+                " t WHERE " + columnName + " = :id";
```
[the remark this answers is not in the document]
wolfdale: Changing to JPA criteria.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1397,6 +1377,64 @@ public boolean performConsumerUpdates(ConsumerDTO updated, Consumer toUpdate,
...
+            return false;
+        }
+
+        boolean changesMade = false;
+        Environment e = null;
+        String environmentId = updated.getEnvironment() == null ? null : updated.getEnvironment().getId();
+        String environmentName = updated.getEnvironment() == null ? null : updated.getEnvironment().getName();
+        String validatedEnvId = null;
+
+        if (environmentId != null && (toUpdate.getEnvironmentId() == null ||
```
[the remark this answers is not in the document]
wolfdale: This is what we had in first iteration, IF-ELSE logic. its just I prioritized the envName, resulting in close coupling revolving around environment entity. Now that environment object is totally removed, making it make cleaner and easier to make sense.

My primary objective was to adjust environment name without touching old code, being cautious not to break things that I don't have good understanding on.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -171,6 +170,49 @@ public E get(Serializable id) {
...
+    }
+
+    /**
+     * Returns name of primary key column
+     * @return Returns entity primary key column
+     */
+    public String getPrimaryColumnName() {
+        String primaryColumn = null;
+
+        for (SingularAttribute attribute : this.getEntityManager().
```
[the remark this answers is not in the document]
wolfdale: Doesn't both approaches require to have entity objects. 
For
1)  We can get identifier from `EntityManagerFactory` this is supported only for JPA 2.0 onward, but it require entity object as parameter.    
```
Object key = this.getEntityManager()
            .getEntityManagerFactory()
            .getPersistenceUnitUtil()
            .getIdentifier(entityObj);
```
2) Hibernate session also require entity object. 
```
org.hibernate.Session session = (Session) this.currentSession().getDelegate();
Serializable key = session.getIdentifier(entity_obj);
```

wolfdale:
Added more test around AHC & EnvironmentCurator classes.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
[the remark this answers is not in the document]
Ceiu: You're not looking for the value of the identifier field, you're looking for the field name, which doesn't require an instance of the entity.

JPA:
```
String name = entityManager.getMetamodel()
  .entity(<entity class>)
  .getId()
  .getName();
```

Hibernate:
```
String name = ((SessionImpl) session)
  .getSessionFactory()
  .getClassMetadata(<entity class>)
  .getIdentifierPropertyName();
```

server/src/main/java/org/candlepin/resource/ConsumerResource.java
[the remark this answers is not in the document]
Ceiu: Sometimes that's fine, but for situations like this where the order fields are processed and the way certain actions are done, it's best to redo the whole block so the input semantics make sense and there's no confusing side effects.

server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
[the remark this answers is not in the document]
wolfdale: Regarding composite keys, JPA `getId()` throws `IllegalArgumentException` exception, tested with `OwnerProduct` entity which has `OwnerProductKey` as key. For now I decided to suppress the exception and return null. 
Also added a test case with composite key in AHC class.

PR #2506, opened 2019-11-20
1774640: Provide configuration option to disable UUID uniqueness cons…

wottop:
…traint in Satellite 6

### DIFF
--- a/server/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/server/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -221,6 +221,7 @@ public class ConfigProperties {
 
     public static final String STANDALONE = "candlepin.standalone";
     public static final String ENV_CONTENT_FILTERING = "candlepin.environment_content_filtering";
+    public static final String USE_SYSTEM_UUID_FOR_MATCHING = "candlepin.use_system_uuid_for_matching";
 
     public static final String CONSUMER_SYSTEM_NAME_PATTERN = "candlepin.consumer_system_name_pattern";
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/service/impl/HypervisorUpdateAction.java
```diff
@@ -204,6 +202,21 @@ else if (jobReporterId == null) {
         return new Result(result, hypervisorKnownConsumersMap);
     }
 
+    private Consumer reconcileByUuid(Consumer knownHost, Consumer incoming, Map<String,
+        Consumer> systemUuidKnownConsumersMap) {
```
nikosmoum: Probably better for readability to not break the `Map<String,Consumer>` in half with line break (I'm actually surprised checkstyle didn't complain)

server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -221,6 +221,7 @@ private ConfigProperties() {
 
     public static final String STANDALONE = "candlepin.standalone";
     public static final String ENV_CONTENT_FILTERING = "candlepin.environment_content_filtering";
+    public static final String USE_SYSTEM_UUID = "candlepin.use_system_uuid";
```
nikosmoum: I can't help but think that this property isn't descriptive enough for what it does, but can't think of an alternative that would be better, so I suppose it's fine as is.

jturel:
Does this also handle the general consumer registration case? It appears to only address hypervisor update. In Satellite we want all-or-nothing (hypervisor and ordinary systems alike)

PR #2511, opened 2019-11-25
[M] 1772194: Updated event messaging to use TEXT_TYPE and simple strings

Ceiu:
- Changed the event messaging format to use an explicit TEXT type,
  and updated the method used to write the message body to avoid
  extraneous sizing values
- Added two new properties to event messages: EVENT_TYPE and
  EVENT_TARGET, representing the event type and target, respectively
- Updated the EventSinkImpl and EventMessageReceiver tests to use
  JUnit 5
- Removed a couple tests that were no longer needed from the
  EventSinkImpl test suite

### DIFF
--- a/server/src/test/resources/logback-test.xml
+++ b/server/src/test/resources/logback-test.xml
@@ -7,7 +7,7 @@
     </encoder>
   </appender>
   <logger name="org.candlepin.util.filters.DumpFilter" level="INFO"/>
-  <logger name="org.candlepin" level="INFO"/>
+  <logger name="org.candlepin" level="DEBUG"/>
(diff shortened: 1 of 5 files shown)


Ceiu:
This is an updated version of #2499 that uses TEXT_TYPE and simple strings rather than bytes.

server/src/test/resources/logback-test.xml
```diff
@@ -7,7 +7,7 @@
     </encoder>
   </appender>
   <logger name="org.candlepin.util.filters.DumpFilter" level="INFO"/>
-  <logger name="org.candlepin" level="INFO"/>
+  <logger name="org.candlepin" level="DEBUG"/>
```
nikosmoum: :+1:

PR #2512, opened 2019-11-26
[F] Updated event messaging to use TEXT_TYPE and simple strings

Ceiu:
- Changed the event messaging format to use an explicit TEXT type,
  and updated the method used to write the message body to avoid
  extraneous sizing values
- Added two new properties to event messages: EVENT_TYPE and
  EVENT_TARGET, representing the event type and target, respectively
- Updated the EventSinkImpl and EventMessageReceiver tests to use
  JUnit 5
- Removed a couple tests that were no longer needed from the
  EventSinkImpl test suite

### DIFF
--- a/server/src/main/java/org/candlepin/audit/QpidEventMessageReceiver.java
+++ b/server/src/main/java/org/candlepin/audit/QpidEventMessageReceiver.java
@@ -16,6 +16,7 @@ package org.candlepin.audit;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
+import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.api.core.client.ClientMessage;
 import org.candlepin.async.impl.ActiveMQSessionFactory;
 import org.slf4j.Logger;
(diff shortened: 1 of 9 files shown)


Ceiu:

server/src/main/java/org/candlepin/audit/QpidEventMessageReceiver.java
```diff
@@ -57,7 +58,16 @@ public void onMessage(ClientMessage msg) {
             }
 
             // Process the message via our EventListener framework.
-            body = msg.getBodyBuffer().readString();
+            if (msg.getType() == ClientMessage.TEXT_TYPE) {
```
nikosmoum: Any reason that in this PR the text type handling was added to QpidEventMessageReceiver, but on the master PR it wasn't?

Ceiu: That'd be an oversight on my part. I only noticed it in this branch because of the hierarchy change and for whatever reason I didn't think to go back and update master.

Leave this PR up over the weekend as a reminder and I'll get something posted on Monday for master.

Ceiu:
Okay, this is ready. #2514 has been posted to address the oversight in master.

PR #2513, opened 2019-11-27
Modify hypervisor report processing to lock hypervisors individually …

wottop:
…[ENT-1477]


server/src/main/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJob.java
```diff
@@ -167,23 +164,6 @@ public void toExecute(final JobExecutionContext context) throws JobExecutionExce
...
-                    consumer, HypervisorConsumerDTO.class);
-                if (result.wasCreated(translated)) {
-                    created.add(consumer);
-                }
-                else {
-                    updated.add(consumer);
-                }
-            }
-
-            doInBulk(created, consumers -> consumerCurator.saveAll(consumers, false, false));
```
Januson: We should remove doInBulk as it's unused. There are also some unused dependencies.

PR #2515, opened 2019-12-02
[M] Added required module for Java 11 support (ENT-1027)

Ceiu:
- Added the javax.xml.bind module for continued JAXB functionality


nikosmoum:
@Ceiu I know this is already merged, but the dependency should also be added in server/pom.xml or else the release build will fail

PR #2524, opened 2019-12-10
fixed qpid_proton avilable version and add keycloak dependecy

zhongtianxie:
try qpid_proton version in Gemfile*, only version 0.28.0
can be used
lack of keycloak build dependency in buildfile will block
compiling, need be added

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,5 @@
-source 'https://rubygems.org'
+#source 'https://rubygems.org'
+source 'https://gems.ruby-china.com/'
 
 gem 'buildr', '1.5.3'
 # Not a typo - we use both buildr and builder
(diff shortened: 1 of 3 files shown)


barnabycourt:
This changes the source for ruby gems and we are also moving away from using the buildfile for compiling and are strictly using gradlew for compiling the project at this point.

zhongtianxie:
> This changes the source for ruby gems and we are also moving away from using the buildfile for compiling and are strictly using gradlew for compiling the project at this point.
ok,thanks  a lot; where i can find some deploy guide; 
https://www.candlepinproject.org/docs/candlepin/developer_deployment.html

zhongtianxie:
> > This changes the source for ruby gems and we are also moving away from using the buildfile for compiling and are strictly using gradlew for compiling the project at this point.
> > ok,thanks  a lot; where i can find some deploy guide;
> > https://www.candlepinproject.org/docs/candlepin/developer_deployment.html

sorry my fault, i get it ...

PR #2525, opened 2019-12-10
Clean up unused files

barnabycourt:
Since buildr is no longer being used, we should remove the buildfile & the associated tasks


barnabycourt:
retest this please

PR #2526, opened 2019-12-10
[M] Restored the rvm+ruby setup step during container construction (ENT-1816)

Ceiu:
- Docker containers will install and configure rvm and Ruby again
- The test script now provides an option to skip pulling images
  from the repo in favor of using local images

### DIFF
--- a/docker/candlepin-base/setup-devel-env.sh
+++ b/docker/candlepin-base/setup-devel-env.sh
@@ -66,6 +66,24 @@ BASHRC
 git clone https://github.com/candlepin/candlepin.git /candlepin
 cd /candlepin
 
+# Setup and install rvm, ruby and pals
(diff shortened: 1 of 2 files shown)


Ceiu:
Reviewers: Disregard the spec failures. This exists outside of what those test and serves to address the problems they're currently facing.

To manually test this, run the commands locally from the docker directory. A full test probably involves building the images and then running some manual invocations of cp-test via test script. For instance:

```
$ sudo ./build-images
$ sudo ./test -l -p -c '/usr/bin/cp-test -r'
```

With the test command specifically, the presence of the ```-l``` flag above should prevent docker-compose from pulling the images. A verification can be done by looking at the image hash via ```docker images``` after running the ```build-images``` command but before invoking ```test```, then verifying that after the test run, the image hash has not changed. Removing the ```-l``` argument should trigger a pull on test invocation, which should cause the hash to change (as well as breaking rspec within the container).

Lorquas:
* chicken and egg problem, these failing tests won't pass until after merge and/or pushing up this ruby-ized container first

PR #2527, opened 2019-12-11
[F] Move EntitlementRules to Java

nikosmoum:
It follows some of the patterns introduced in the move of [ActivationKeyRules to java](https://github.com/candlepin/candlepin/pull/2485).

- Remove dead code from EntitlementRules.java
- Move package org.candlepin.policy.js.entitlement
- Create PoolValidator enum which holds the validation code
  for pre-entitlement rules (equivalent of rules.js'
  do_pre_* methods).
- EntitlementRules.java now uses the PoolValidator instead of
  calling the javascript rules.
- Add utility methods on Consumer and Pool.
- Add FactValueCalculator utility class.
- Transform EntitlementRulesTranslator validation
  errors/warnings from Strings to Enums.
- Update/Add/Remove unit tests as needed.


server/src/main/java/org/candlepin/policy/Util.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * Returns true of the consumer has the specified capability, or false otherwise.
+     *
+     * @param consumer the consumer who's capability is checked.
+     * @param capability the capability we want to check for.
+     * @return true of the consumer has the capability, or false otherwise.
+     */
+    public static boolean isCapable(Consumer consumer, String capability) {
```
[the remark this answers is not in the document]
nikosmoum: That is a good idea and it makes a lot of sense!

server/src/main/java/org/candlepin/policy/Util.java
```diff
@@ -0,0 +1,99 @@
...
+                if (consumerCapability.getName().equals(capability)) {
+                    isCapable = true;
+                    break;
+                }
+            }
+        }
+        return isCapable;
+    }
+
+    public static boolean isMultiEnt(Pool pool) {
```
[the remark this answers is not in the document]
nikosmoum: Yep, same here. Makes sense.

server/src/test/java/org/candlepin/policy/entitlement/PoolValidatorTest.java
```diff
@@ -0,0 +1,311 @@
...
+            I18nFactory.FALLBACK);
+        Product product = new Product();
+        product.setId("prod_id_1");
+        assertEquals(
+            "This system has already had the subscription matching pool ID \"prod_id_1\" attached.",
+            EntitlementRules.ErrorKeys.ALREADY_ATTACHED.buildErrorMessage(i18n, product.getId()));
+    }
+
+    @Test
+    public void testGlobalValidateNoErrorsWhenConsumerIsManifest() {
```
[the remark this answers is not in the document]
nikosmoum: Didn't know about test nesting; it looks interesting. Even though in this case the separate before(setup) isn't needed, the grouping helps for readability, so I'll try it.

server/src/main/java/org/candlepin/policy/Util.java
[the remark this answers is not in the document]
nikosmoum: Done!

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -822,6 +829,36 @@ public void setCapabilities(Set<ConsumerCapability> capabilities) {
...
+    /**
+     * Was this consumer created in the last 24 hours?
+     *
+     * @return boolean was created in the last 24 hours
+     */
+    public boolean isNewborn() {
+        if (getCreated() == null) {
+            return false;
+        }
+        Date now = new Date();
```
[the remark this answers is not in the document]
nikosmoum: Done

server/src/main/java/org/candlepin/policy/FactValueCalculator.java
```diff
@@ -0,0 +1,91 @@
...
+        // private & empty by design
+    }
+
+    /*
+     * The default calculator used if there is no custom calculator
+     * defined for the product attribute. The RAW consumer fact value
+     * is returned, or otherwise 1.
+     */
+    private static int defaultCalculator(String productAttribute, Consumer consumer) {
+        return consumer.getFact(ATTRIBUTES_TO_FACTS.get(productAttribute)) != null ?
```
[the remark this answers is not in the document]
nikosmoum: Done

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -481,8 +481,8 @@ protected void verifyDevProducts(Consumer consumer, String expectedSku, Develope
                 log.debug("consumer {} dry-run errors:", consumer.getUuid());
                 for (Entry<String, ValidationResult> entry : e.getResults().entrySet()) {
                     log.debug("errors for pool id: {}", entry.getKey());
-                    for (ValidationError error : entry.getValue().getErrors()) {
-                        log.debug(error.getResourceKey());
+                    for (RulesValidationError error : entry.getValue().getErrors()) {
+                        log.debug("{}", error);
```
Ceiu: :+1: 
Clever way to avoid a .toString call if debug logging is disabled.

server/src/main/java/org/candlepin/policy/Util.java
```diff
@@ -0,0 +1,71 @@
...
+        String consumerTypeLabel) {
+        // Non-system consumers without an architecture fact can pass this rule
+        // regardless what arch the product requires.
+        if ((consumerArch == null || consumerArch.isEmpty()) &&
+            !ConsumerType.ConsumerTypeEnum.SYSTEM.getLabel().equals(consumerTypeLabel)) {
+            return true;
+        }
+
+        if (productArch != null) {
+            List<String> supportedArches = new ArrayList<>(
```
Ceiu: Change this to a set to avoid a bunch of iteration on the .contains calls.

server/src/main/java/org/candlepin/policy/entitlement/EntitlementRulesTranslator.java
```diff
@@ -0,0 +1,245 @@
...
+    // Add documentation for the intended circumstances for these error keys
+
+    /**
+     * Keys that represent the various warnings that can be produced by the EntitlementRules.
+     * Warning messages may use the following variable(s):
+     * {0} - pool id
+     */
+    public enum WarningKeys implements RulesValidationWarning {
+        ARCHITECTURE_MISMATCH(
+            "The entitlement's product architecture does not match with the consumer's."),
```
Ceiu: You'll need to mark these with ```I18n.marktr``` so the translation stuff can pick it up. Otherwise the strings will be passthrough when i18n.tr is called and won't ever get translated.

server/src/main/java/org/candlepin/policy/entitlement/EntitlementRulesTranslator.java
```diff
@@ -0,0 +1,245 @@
...
+     * {2} - Pool restricted username
+     * {3} - Host consumer uuid
+     * {4} - Pool product (sku) name
+     * {5} - Product multiplier
+     * {6} - Pool's product id
+     * {7} - Pool end date
+     */
+    public enum ErrorKeys implements RulesValidationError {
+        DERIVED_PRODUCT_UNSUPPORTED_BY_CONSUMER(
+            "Unit does not support derived products data required by pool \"{1}\""),
```
Ceiu: More ```I18n.marktr``` here.

server/src/main/java/org/candlepin/policy/entitlement/EntitlementRulesTranslator.java
```diff
@@ -0,0 +1,245 @@
...
+     * Indicates the context that the error happened in.
+     */
+    private enum ErrorMessageTypes {
+        POOL_ERROR,
+        ENTITLEMENT_ERROR,
+        PRODUCT_ERROR
+    }
+
+    private static final String DEFAULT_PRODUCT_ERROR_MESSAGE =
+        "Unable to attach subscription for the product \"{1}\": {0}.";
```
Ceiu: ```I18n.marktr```s here

server/src/main/java/org/candlepin/policy/Util.java
nikosmoum: Done

server/src/main/java/org/candlepin/policy/entitlement/EntitlementRulesTranslator.java
nikosmoum: Done

server/src/main/java/org/candlepin/policy/entitlement/EntitlementRulesTranslator.java
nikosmoum: Done

server/src/main/java/org/candlepin/policy/entitlement/EntitlementRulesTranslator.java
nikosmoum: Done

nikosmoum:
@Ceiu I've fixed the issues and rebased against master

server/src/main/java/org/candlepin/policy/Util.java
```diff
@@ -0,0 +1,71 @@
...
+        // Non-system consumers without an architecture fact can pass this rule
+        // regardless what arch the product requires.
+        if ((consumerArch == null || consumerArch.isEmpty()) &&
+            !ConsumerType.ConsumerTypeEnum.SYSTEM.getLabel().equals(consumerTypeLabel)) {
+            return true;
+        }
+
+        if (productArch != null) {
+            Set<String> supportedArches = new HashSet<>(
+                Arrays.asList(productArch.toUpperCase().trim().split("\\s*,\\s*")));
```
Ceiu: Sorry I didn't think of this the first time around, but we can pair this down a bit by using either ```Set.of``` or our own ```Util.asSet``` to save the declaration and ```Arrays.asList``` call.

nikosmoum: Done. I used `Util.asSet`, since `Set.of` is a Java 9 feature, so not yet available.

Ceiu:
Is there anything else that should be considered before merging this? Looks like it's going into a feature branch, so I don't expect there to be anything holding it up, but I want to be sure before slamming that merge button.

As an aside, one thing that stands out with the FQCN usage in the policy Util, and something I should have asked about before, is whether or not we want such things to live in the Util class or if we want to continue adding the helper methods to the entities directly as we did with some of the new things on consumer. Is there a plan, expectation, or pattern for Util beyond what's there now?

nikosmoum:
@Ceiu This was basically a PoC or "let's see how this can be done", and see what issues might be encountered, and try to get feedback in general (such as suggestions of patterns like the one you just asked about), but also I implemented as if it is production-ready (full test coverage, etc), so unless you see anything bad, it can get merged in the feature branch.

Generally, I have a few notes of patterns we want to use already for the rest of the tasks, and will write up a small doc with them that will be reviewed as a task of the epic I'm about to create as part of this card.

Specifically about the Util class: my thought is that any "utility" code of small or moderate size that fits conceptually as an instance method, should go on the entity (like the `isNewborn()/isCapable()` methods), but anything that doesn't fit that concept should be either 1) put into it's own utility class (e.g. `FactValueCalculator`), or 2) the `policy/Util` class. The reason I left the `architectureMatches` method in the Util class is that it that it uses data from 3 entities (pool, consumer, consumerType), but thinking about it now, it might be better to re-structure it as a method on the consumer instance and move the consumerType check to the caller method.

nikosmoum:
@Ceiu I've removed the Util class and added the architectureMatches method on the consumer class.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -822,6 +831,65 @@ public void setCapabilities(Set<ConsumerCapability> capabilities) {
...
+     * Returns true if the consumer's architecture matches that of the input.
+     *
+     * @param productArch The architecture attribute value of a product.
+     * @return true if the architecture matches, false otherwise.
+     */
+    public boolean architectureMatches(String productArch) {
+
+        if (productArch != null) {
+            Set<String> supportedArches =
+                org.candlepin.util.Util.asSet(productArch.toUpperCase().trim().split("\\s*,\\s*"));
```
Ceiu: This doesn't need to be fully qualified if this lives here

nikosmoum: You're right, forgot about that! Done

PR #2529, opened 2019-12-12
[M] Merge Async Jobs Feature (ENT-1723)

Ceiu:


server/src/main/java/org/candlepin/pinsetter/tasks/HealEntireOrgJob.java
```diff
@@ -122,7 +122,7 @@ public static JobDetail healEntireOrg(Owner owner, Date entitleDate) {
         map.put(JobStatus.TARGET_TYPE, JobStatus.TargetType.OWNER);
```
nikosmoum: This file shouldn't be here at all, since the whole pinsetter package should be gone.

server/src/test/java/org/candlepin/pinsetter/tasks/HypervisorUpdateJobTest.java
```diff
@@ -18,6 +18,8 @@
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
```
nikosmoum: Same here. pinsetter should be gone

PR #2530, opened 2019-12-13
[ENT-84]:Decide whether to reorganize the Certificate DTOs so they follow the same hierarchy the model objects do.

wolfdale:
- Re-organised the ProductCertificateDTO to inherit from AbstractCertificateDTO. 
- Getter method for serial certificate dto are overridden to return null.
- Skipped testing `serial` (inherited) field for ProductCertificateDTO in AbstractDTOTest

### DIFF
--- a/server/src/test/java/org/candlepin/dto/AbstractDTOTest.java
+++ b/server/src/test/java/org/candlepin/dto/AbstractDTOTest.java
@@ -64,7 +64,8 @@ public abstract class AbstractDTOTest<T extends CandlepinDTO<T>> {
         for (Method method : this.dtoClass.getMethods()) {
             try {
                 Matcher matcher = ACCESSOR_NAME_REGEX.matcher(method.getName());
-                if (matcher.matches() && method.getParameterTypes().length == 0) {
+                if (matcher.matches() && method.getParameterTypes().length == 0 &&
+                    !skipMethod(method.getName())) {
                     String fieldName = matcher.group(1);
                     Method mutator = this.dtoClass.getMethod("set" + fieldName, method.getReturnType());
                     fields.put(fieldName, new Method[] { method, mutator });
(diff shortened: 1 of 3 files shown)


wolfdale:

server/src/test/java/org/candlepin/dto/AbstractDTOTest.java
```diff
@@ -88,6 +89,18 @@ public AbstractDTOTest(Class<T> dtoClass) {
...
 
+    /**
+     * Method to skip testing any field in any DTO class by
+     * not including their accessor method for testing.
+     * Defaults to false
+     *
+     * @param methodName getter method of field.
+     * @return boolean value
+     */
+    protected boolean skipMethod(String methodName) {
```
Ceiu: This shouldn't be a thing. If a field isn't implemented or otherwise won't be valuable, we should be verifying that it always returns null or, as is the task here, determining whether or not the model is valid.

wolfdale: Actually AbstractDTOTest is a generic class which perform test for all fields in inheritance hierarchy, for ProductCertificateDTO.class which now override the Getter/Setter method for `CertificateSerialDTO aka serial` to return null for Get method and do nothing for Set method. This lead to failure of few Junit due to assert calls on DTO for test like EqualsForIndividualFields. I can bypass this by adding a dummy value for serial in ProductCeritficateDTOTest, but ProductCertificateDTO has nothing to do with Serial CertificateSerialDTO and doesn't make sense.

wolfdale:
@Ceiu I have added few approaches in Jira.

server/src/test/java/org/candlepin/dto/AbstractDTOTest.java
Ceiu: The point here is that if we have a method, we should be testing it -- especially when the work for that test is literally just adding a one-liner to set the expected values.

Moreover, you're touching on the reason the task exists: the model may be incorrect here and it might be worth either completely separating them or refactoring them and making them more connected. This could occur on the entity side, the DTO side, both or neither. Perhaps some of the details of the task were lost in the conversion from Trello to Jira, but that's the objective here. This was supposed to be an investigative task first and foremost.

Ceiu:
After further discussions on this, and with the looming DTO changes coming from the spec-first epic, the output we want from this task is not going to be code, but a design (of sorts) that outlines what we want the models (both entity and DTO) look like, so they can be properly implemented as part of the other tasks.

As such, no further code changes will be needed to the existing DTOs, and any entity changes should come in a separate PR after the design is vetted.

PR #2531, opened 2019-12-13
container refactor

Lorquas:
TODO:

- [x] Create test enviornment image
  - [x] add ruby to the user's path
- [x] Create testenv script for easy local dev
  - [ ] bring up a pod when running rspec tests
- [ ] Create candlepin image
- [ ] Create Jenkinsfile
  - [ ] update candlepin-jobs repo
- [ ] Deprecate docker/ folder
- [ ] Write README


Ceiu:
Looks good so far. Seems like it'll always run as the Jenkins user, but I don't think that will be a problem for running it locally.

Ceiu:
I spent some of the day trying to sort out a few problems I'm hitting with the current state of this and looked into more of the nested container stuff.

I've found this article which describes exactly what I was describing initially. Since this seems to be part of the container logic, perhaps this is something we can work into testenv rather than exposing it to the host environment.
https://www.docker.com/blog/docker-can-now-run-within-docker/

That aside, I've been fighting with some selinux errors when building the testenv container, and I'm unable to create a container from the completed image either manually or with the script. This might come down to user error or something busted on my system, but it's currently preventing me from making progress.

I'll continue plugging away at it, but it looks like we have a viable path for doing nested containers if the docker method described above translates. Hopefully I can get to the point of running it locally again to find out.

Ceiu:
Copy/paste from some thoughts late on the day before break:

```
15:24:40 < crog> khowell: when we're back from break, I want to talk through the value of these containers again with regard to not being able to do containers in containers. 
15:25:14 < crog> Given that it means we have to build outside of the testenv container, we still have to build and do basically everything natively; so it doesn't seem like we actually gain anything from doing it that way.
15:25:36 < crog> I guess you could argue that we can move rspec and all the ruby deps and that mess into a container, but that's about it.
15:26:56 < crog> In light of that, it might make sense to refocus everything there: one container for running rspec and one for doing whatever jenkins needs to do.
```

Lorquas:
Adding this as a note: I also found this since with podman, this area specifically behaves differently than docker.  There is a buildah container used for making other images: https://developers.redhat.com/blog/2019/08/14/best-practices-for-running-buildah-in-a-container/

However following the steps I have not been able to get this to work; the steps around allocating subuid/subgid don't seem to work since sudo is required for installing the ruby bits (even though it should?).  I would add some brainstorming around this to that meeting.

Lorquas:
squashed commits and added rootless ruby so I can try the buildah thing...

wottop:
This PR is not far enough along to keep on the list. When there is content that needs to be reviewed, or it is closer to the end, then we can reopen it.

PR #2534, opened 2019-12-16
[M] Adjusted bundler version and skip container gem system update

Ceiu:
- Skip the system update for ruby gems, as this forces bundler 2.1
  to be used as the default bundler version
- Manually specify bundler v1.16 to be installed

### DIFF
--- a/docker/candlepin-base/setup-devel-env.sh
+++ b/docker/candlepin-base/setup-devel-env.sh
@@ -80,8 +80,7 @@ rvm use --default 2.5.3
 set -v
 
 # Install all ruby deps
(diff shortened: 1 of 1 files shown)


Ceiu:
This is kinda janky, but I've been trying variations of versions all day and none of them work nicely. Latest ruby + latest bundler + latest rspec still fails. In fact, it seems any combination where bundler goes beyond 1.16 breaks. Unfortunately this means we can't update the system gems as this will get us a permanent version of bundler 2.1.0 which doesn't work.

Ceiu:

PR #2543, opened 2019-12-20
Fake insights-client in testing repository

jirihnidek:
* It is possible to include some files in testing RPMS
  - Files should not be platform specific
  - Files should not require any other dependency
* Added one more product, content and repo containing RPM
  with fake insights-client, because this RPM is not available
  in any repo for CentOS or Fedora
* It is useful for testing cockpit plugin, e.g. this PR: https://github.com/candlepin/subscription-manager/pull/2180


Ceiu:
Seems fine to me; doesn't appear to break anything on Jenkins or my machine. ack

PR #2547, opened 2020-01-02
1785446: Hypervisor update fails when consumer name is null [M]

wottop:


server/src/main/java/org/candlepin/service/impl/HypervisorUpdateAction.java
```diff
@@ -182,8 +182,9 @@ public Consumer reconcileHost(Owner owner, Consumer incomingHost, HypervisorUpda
             boolean hypervisorIdUpdated = updateHypervisorId(resultHost, owner, jobReporterId,
                 hypervisorId);
 
-            boolean nameUpdated = resultHost.getName() == null ||
-                !resultHost.getName().equals(incomingHost.getName());
+            boolean nameUpdated = incomingHost.getName() != null &&
```
Ceiu: Looking at the BZ, it seems like we're hitting an exception if the new name is null or empty. This catches the null condition, but not the empty condition.

PR #2548, opened 2020-01-03
Add spec-first build changes

kahowell:
After this is in place, the process for porting a given resource or DTO to spec-first is:

1. Add the resource definition to `api/candlepin-api-spec.yaml` ([editor.swagger.io](https://editor.swagger.io/) works well for this).
2. Add any DTOs to `api/candlepin-api-spec.yaml`.
3. Change the resource to implement the generated API interface.
4. Remove all `javax.ws.rs` annotations from the resource implementation.
5. Remove all `io.swagger` annotations from the resource implementation.
6. Remove old DTOs from the codebase (refactor to use the generated ones instead).
7. Fix any broken code or tests.
8. Ensure unit tests still pass and the resource still works.
9. Ensure swagger.json is still generated and is equivalent.

Note: After building, you can view generated code in `api/build/generated/src/gen/java`.

### DIFF
--- a/api/candlepin-api-config.json
+++ b/api/candlepin-api-config.json
@@ -1,0 +1,8 @@
+{
+  "modelPackage": "org.candlepin.dto.api.v1",
+  "apiPackage": "org.candlepin.resource",
+  "invokerPackage": "org.candlepin",
+  "groupId": "org.candlepin",
+  "artifactId": "candlepin-api",
+  "java8": true
+}
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -1,0 +1,28 @@
+openapi: 3.0.2
+info:
+  title: Candlepin
+  description: Candlepin is a subscription management server written in Java. It helps with management of software subscriptions.
+  version: 3.0.0-draft
+
+servers:
+  - url: /candlepin
+
+security:
+  - basicAuth: []
+
+paths: {}
+
+components:
+  schemas:
+    ExceptionMessage:
+      description: An exception occurred
+      properties:
+        displayMessage:
+          type: string
+        requestUuid:
+          type: string
+
+  securitySchemes:
+    basicAuth:
+      type: http
+      scheme: basic
--- a/build.gradle
+++ b/build.gradle
@@ -9,10 +9,12 @@ buildscript {
     }
     dependencies {
         classpath group: 'org.yaml', name: 'snakeyaml', version: '1.19'
+        classpath "org.openapitools:openapi-generator-gradle-plugin:4.0.1"
     }
 }
 
 plugins {
+    id "io.spring.dependency-management" version "1.0.8.RELEASE"
     id "nebula.lint" version "11.5.0"
     id "checkstyle"
 }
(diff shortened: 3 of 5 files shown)


kahowell:
@awood since I ported this over from rhsm-conduit/rhsm-subscriptions, I was hoping you could give it a once over for anything I missed, or anything that doesn't make sense after porting to candlepin build.

@abhiskum @sonalidhome @sonalidhome @Januson @Ceiu I would like everyone to see the changes here and have a rough understanding of how it works, and I'd like at least one of you to formally approve/reject (whoever feels comfortable).

kahowell:
#2549 illustrates what follow-on work looks like, and may be helpful in reviewing this PR.

api/candlepin-api-config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "modelPackage": "org.candlepin.dto.api.v1",
```
Ceiu: What are the implications of the package changing with API version differences? That is, how painful is it going to be if we change this to a v2 or some such in the future?

api/candlepin-api-spec.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+security:
+  - basicAuth: []
+
+paths: {}
+
+components:
+  schemas:
+    ExceptionMessage:
+      description: An exception occurred
```
Ceiu: Is this for a generic, unhandled exception? Do more specific exceptions get their own entry in this area?

api/candlepin-api-config.json
awood: We could always have two API specification files that generate to different package names.  That wouldn't be awesome, but that's one way we could manage a change to v2.

build.gradle
```diff
@@ -194,6 +211,74 @@ subprojects {
...
+    }
+
+    openApiGenerate {
+      generatorName = "jaxrs-spec"
+      inputSpec = api_spec_path
+      configFile = config_file
+      outputDir = "$buildDir/generated"
+      configOptions = [
+          interfaceOnly: true,
+          generatePom: false
```
awood:
```suggestion
          generatePom: false,
          dateLibrary: "java8"
```

Not mandatory or anything, but helpful for future proofing

api/candlepin-api-spec.yaml
kahowell: This essentially a port of `org.candlepin.common.exceptions.ExceptionMessage` to the swagger YAML. As resources require it we could have additional schemas (e.g. `org.candlepin.common.exceptions.DeletedConsumerExceptionMessage`). That looks like the only other place we've done this.

build.gradle
kahowell: :+1:

kahowell:
@awood I made the suggested change to specify java8 date library. Any other feedback?
@Ceiu are you happy with my answers? Any other feedback?

api/candlepin-api-spec.yaml
Ceiu: That sounds fine, and we can discuss specifics if it comes up during actual implementation.

awood:
@kahowell Looks good!

PR #2549, opened 2020-01-03
Refactor root resource to spec-first

kahowell:
Until #2548 is merged, this is just a draft, but can serve as an example to help understand the process outlined in that PR.


api/candlepin-api-spec.yaml
```diff
@@ -10,7 +10,29 @@ servers:
...
 
-paths: {}
+paths:
+  /:
+    get:
+      description: Retrieves a list of links corresponding to Root resources
+      tags:
+        - root
+      operationId: getRootResources
+      security: []  # this resource doesn't require auth
```
Ceiu: This is probably a bit late to be asking this, but what is the intent for security here? Do we just throw our existing verify annotations in here for now (where appropriate) and then change this when we get around to updating our security model?

api/candlepin-api-spec.yaml
```diff
@@ -22,6 +44,14 @@ components:
         requestUuid:
           type: string
 
+    Link:
+      description: Used to just represent a link to a top level resource in Candlepin
```
Ceiu: Grammar police: remove "just"

api/candlepin-api-spec.yaml
kahowell: Yes, I think we can rework our security model in the future. Going ahead and specifying auth or noauth (by adding the line above) conveys intent in the spec, but we should keep our custom security annotations on the resource impl for now.

kahowell:
23ca10734..6fb4f9084: grammar fix

Note this needs #2575 before tests will pass.

nikosmoum:
Now that #2575 has been merged, this can be rebased right?

PR #2563, opened 2020-01-14
[F] [ENT-1831][ENT-1830] update pool response and manifest export flow

abhiskum:
- Update org/candlepin/dto/api/v1/PoolTranslator.java to populate provided products into PoolDTO from Pool.Product.ProvidedProducts
 - Updated org/candlepin/sync/Exporter.java to copy provided products from product and derived product
 - Updated product creation logic in rspec
 - Updated org/candlepin/dto/manifest/v1/PoolTranslator.java to populate provided products into PoolDTO from pool.product.providedProducts
 - Updated rspec and junit as per code changes
- Removed spec tests related to accumulating provided products in case of stacked entitlements as marketing product and stack id has one to one relationship i.e. different marketing product will not have the same stacking id and all the required provided products will be available in marketing product itself from the adapter.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -974,6 +974,11 @@ public class CandlepinPoolManager implements PoolManager {
             }
 
             pool.setProvidedProducts(products);
+            // TODO: workaround to pass import spec tests. we will revisit and update this in import and
+            // refresh code changes
+            if (pool.getProduct() != null) {
+                pool.getProduct().setProvidedProducts(products);
+            }
         }
 
         if (sub.getDerivedProvidedProducts() != null) {
@@ -993,6 +998,11 @@ public class CandlepinPoolManager implements PoolManager {
             }
 
             pool.setDerivedProvidedProducts(products);
+            // TODO: workaround to pass import spec tests. we will revisit and update this in import and
+            // refresh code changes
+            if (pool.getDerivedProduct() != null) {
+                pool.getDerivedProduct().setProvidedProducts(products);
+            }
         }
 
         return pool;
--- a/server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
@@ -25,6 +25,7 @@ import org.candlepin.model.SubscriptionsCertificate;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
 
 
@@ -109,8 +110,8 @@ public class PoolTranslator extends TimestampedEntityTranslator<Pool, PoolDTO> {
             dest.setSourceEntitlement(sourceEntitlement != null ?
                 modelTranslator.translate(sourceEntitlement, EntitlementDTO.class) : null);
 
-            Collection<Branding> branding = source.getProduct() != null ?
-                source.getProduct().getBranding() : null;
+            Collection<Branding> branding =
+                source.getProduct() != null ? source.getProduct().getBranding() : null;
             if (branding != null && !branding.isEmpty()) {
                 for (Branding brand : branding) {
                     if (brand != null) {
@@ -122,33 +123,32 @@ public class PoolTranslator extends TimestampedEntityTranslator<Pool, PoolDTO> {
                 dest.setBranding(Collections.emptySet());
             }
 
-            Set<Product> products = source.getProvidedProducts();
-            if (products != null && !products.isEmpty()) {
-                for (Product prod : products) {
-                    if (prod != null) {
-                        dest.addProvidedProduct(
-                            new PoolDTO.ProvidedProductDTO(prod.getId(), prod.getName()));
-                    }
-                }
-            }
-            else {
-                dest.setProvidedProducts(Collections.<PoolDTO.ProvidedProductDTO>emptySet());
-            }
+            Collection<Product> products =
+                source.getProduct() != null ? source.getProduct().getProvidedProducts() : null;
+            Set<PoolDTO.ProvidedProductDTO> providedProductDTOs = new HashSet<>();
+            addProvidedProducts(products, providedProductDTOs);
+            dest.setProvidedProducts(providedProductDTOs);
 
-            Set<Product> derivedProducts = source.getDerivedProvidedProducts();
-            if (derivedProducts != null && !derivedProducts.isEmpty()) {
-                for (Product derivedProd : derivedProducts) {
-                    if (derivedProd != null) {
-                        dest.addDerivedProvidedProduct(
-                            new PoolDTO.ProvidedProductDTO(derivedProd.getId(), derivedProd.getName()));
-                    }
-                }
-            }
-            else {
-                dest.setDerivedProvidedProducts(Collections.<PoolDTO.ProvidedProductDTO>emptySet());
-            }
+            Collection<Product> derivedProducts =
+                source.getDerivedProduct() != null ? source.getDerivedProduct().getProvidedProducts() : null;
+            Set<PoolDTO.ProvidedProductDTO> derivedProvidedProductDTOs = new HashSet<>();
+            addProvidedProducts(derivedProducts, derivedProvidedProductDTOs);
+            dest.setDerivedProvidedProducts(derivedProvidedProductDTOs);
         }
 
         return dest;
+    }
+
+    private void addProvidedProducts(Collection<Product> providedProducts,
+        Set<PoolDTO.ProvidedProductDTO> providedProductDTOs) {
+        if (providedProducts == null || providedProducts.isEmpty()) {
+            return;
+        }
+        for (Product product : providedProducts) {
+            if (product != null) {
+                providedProductDTOs.add(new PoolDTO.ProvidedProductDTO(product.getId(), product.getName()));
+                addProvidedProducts(product.getProvidedProducts(), providedProductDTOs);
+            }
+        }
     }
 }
--- a/server/src/main/java/org/candlepin/dto/manifest/v1/PoolTranslator.java
+++ b/server/src/main/java/org/candlepin/dto/manifest/v1/PoolTranslator.java
@@ -25,6 +25,7 @@ import org.candlepin.model.SubscriptionsCertificate;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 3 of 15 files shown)


server/src/main/java/org/candlepin/sync/Exporter.java
```diff
@@ -492,12 +492,18 @@ private void exportProducts(File baseDir, Consumer consumer) throws IOException
             Product product = pool.getProduct();
             products.put(product.getId(), product);
 
+            for (Product providedProduct : product.getProvidedProducts()) {
```
Ceiu: Product model translation needs to be n-tier aware during the conversion to a flat list here and below.

server/src/main/java/org/candlepin/dto/manifest/v1/PoolTranslator.java
```diff
@@ -150,8 +162,47 @@ public PoolDTO populate(ModelTranslator modelTranslator, Pool source, PoolDTO de
             else {
                 dest.setDerivedProvidedProducts(Collections.emptySet());
             }
+
+            // accumulate provided products for stacked entitlements
+            if (PoolType.STACK_DERIVED.equals(source.getType())) {
+                Collection<Product> providedProducts = accumulateStackDerivedPoolProvidedProducts(source);
+                if (source.getDerivedProduct() != null) {
+                    for (Product providedProduct : providedProducts) {
```
Ceiu: These need to be n-tier aware translations

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
```diff
@@ -147,8 +159,48 @@ public PoolDTO populate(ModelTranslator modelTranslator, Pool source, PoolDTO de
             else {
                 dest.setDerivedProvidedProducts(Collections.<PoolDTO.ProvidedProductDTO>emptySet());
             }
+
+            // accumulate provided products for stacked entitlements
+            if (PoolType.STACK_DERIVED.equals(source.getType())) {
+                Collection<Product> providedProducts = accumulateStackDerivedPoolProvidedProducts(source);
+                if (source.getDerivedProduct() != null) {
+                    for (Product providedProduct : providedProducts) {
```
Ceiu: These need to be n-tier aware translations

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
```diff
@@ -122,20 +132,22 @@ public PoolDTO populate(ModelTranslator modelTranslator, Pool source, PoolDTO de
...
                 }
+
             }
             else {
                 dest.setProvidedProducts(Collections.<PoolDTO.ProvidedProductDTO>emptySet());
             }
 
-            Set<Product> derivedProducts = source.getDerivedProvidedProducts();
+            Collection<Product> derivedProducts =
+                source.getDerivedProduct() != null ? source.getDerivedProduct().getProvidedProducts() : null;
```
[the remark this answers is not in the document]
abhiskum: fixed.

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
abhiskum: fixed.

server/src/test/java/org/candlepin/resource/util/InstalledProductStatusCalculatorTest.java
```diff
@@ -117,7 +118,7 @@ public void setUp() {
 
         translator = new StandardTranslator(this.consumerTypeCurator,
             this.environmentCurator,
-            this.ownerCurator);
+            this.ownerCurator, this.entitlementCurator);
```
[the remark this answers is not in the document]
abhiskum: Fixed.

server/src/test/java/org/candlepin/sync/ImporterTest.java
```diff
@@ -163,8 +164,11 @@ public void init() throws URISyntaxException, IOException {
         this.mockSubReconciler = Mockito.mock(SubscriptionReconciler.class);
         this.consumerTypeCurator = Mockito.mock(ConsumerTypeCurator.class);
         oc = mock(OwnerCurator.class);
+        entitlementCurator = mock(EntitlementCurator.class);
 
-        this.translator = new StandardTranslator(this.consumerTypeCurator, this.environmentCurator, this.oc);
+
+        this.translator = new StandardTranslator(this.consumerTypeCurator, this.environmentCurator, this.oc
+            , this.entitlementCurator);
```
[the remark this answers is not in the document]
abhiskum: Fixed.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -974,6 +974,11 @@ private Pool convertToMasterPoolImpl(SubscriptionInfo sub, Owner owner, Map<Stri
             }
 
             pool.setProvidedProducts(products);
+            //TODO: workaround to pass import spec tests. we will revisit and update this in import and
```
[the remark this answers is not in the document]
abhiskum: Fixed.

server/src/main/java/org/candlepin/sync/Exporter.java
abhiskum: Fixed.

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
abhiskum: Done.

server/src/main/java/org/candlepin/dto/manifest/v1/PoolTranslator.java
abhiskum: Done.

abhiskum:
@Ceiu Updated code as per your review comments. Please review it again.

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
```diff
@@ -122,33 +132,74 @@ public PoolDTO populate(ModelTranslator modelTranslator, Pool source, PoolDTO de
...
+            dest.setProvidedProducts(providedProductDTOs);
+
+            Collection<Product> derivedProducts =
+                source.getDerivedProduct() != null ? source.getDerivedProduct().getProvidedProducts() : null;
+            Set<PoolDTO.ProvidedProductDTO> derivedProvidedProductDTOs = new HashSet<>();
+            addProvidedProducts(derivedProducts, derivedProvidedProductDTOs);
+            dest.setDerivedProvidedProducts(derivedProvidedProductDTOs);
+
+            // accumulate provided products for stacked entitlements
+            if (PoolType.STACK_DERIVED.equals(source.getType())) {
```
nikosmoum: This is something we discussed for a bit before, and it didn't seem right to me. I wasn't familiar with the code in StackedSubPoolValueAccumulator and how it's used, and it's hard to go through, but I tried to look into it a bit, and from what I understand this is what happens: 

When we want to update sub-pools (stack derived pools) of a master pool, we find the entitlements from the same stack of that sub-pool, then find the pools off of those entitlements, and see if any of their provided products changed, and if they did change, we set those provided products on that sub-pool [here](https://github.com/candlepin/candlepin/blob/93ff253c81fe5618eeaa823e5e0a24531f34a469/server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java#L577), and eventually persist it.

I assume the problem that arose that prompted you to add this change in the Translator is that those updated provided products are not reflected in the pool's marketing product, so we add them in the DTO here due to possible spec test failures? Would it not be preferable to change the code in PoolRules instead, to add those updated provided products on the marketing product directly and make sure our model data is consistent? Then we wouldn't need to do this on the DTO?

 @abhiskum Maybe I'm misunderstanding something since I'm not familiar with this part of the application? @Ceiu do you happen to know more about this?

server/src/main/java/org/candlepin/dto/manifest/v1/PoolTranslator.java
```diff
@@ -125,33 +135,73 @@ public PoolDTO populate(ModelTranslator modelTranslator, Pool source, PoolDTO de
...
-        return dest;
+    private void addProvidedProducts(Collection<Product> providedProducts,
+        Set<PoolDTO.ProvidedProductDTO> providedProductDTOs) {
+        if (providedProducts == null || providedProducts.isEmpty()) {
+            return;
+        }
+        for (Product product : providedProducts) {
+            if (product != null) {
+                providedProductDTOs.add(new PoolDTO.ProvidedProductDTO(product.getId(), product.getName()));
+                addProvidedProducts(product.getProvidedProducts(), providedProductDTOs);
```
nikosmoum: Are these recursive calls what is meant by "n-tier aware"? Are we sure that this is how we're going to be handling this in the future; as a marketing product that has provided products and each provided product has its own provided products etc.? Last I checked we were not going to be doing the N-tier stuff, because we don't yet know how the product management people are going to design that stuff.
@abhiskum @Ceiu

Ceiu: "N-tier" the project name and n-tier as the thing I've been saying here are two different things (I think).

I've stated this in another PR already, but for the sake of github comment history, I'll reiterate here:
I posed this question during a post scrum, as I wasn't sure if we should be ignoring the provided products of provided products or not. The discussion ended with "yes, we should handle those" and that's what lead to my (perhaps poorly summarized) comment about n-tier handling.

Whether or not that's something that is or will be used on the business side... who knows. Personally, I feel the effort to write it now and not worry about whether or not they'll use it is so low that I see no reason to skip it, and if it never gets utilized, it's almost free as far as processing time goes. We've probably spent more effort discussing the value than just doing it straight away.

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
Ceiu: This probably isn't something we want to change internally in pool rules. As far as I understood the intent of this particular block, it's that we want to maintain the old flat representation of provided products on a given pool at the API level. But apparently stacks share provided products? That bit seems wrong to me, but will require some investigation to verify. If that's truly how stacks work, then this logic is fine (though maybe some other details need to be sorted out).

nikosmoum: This is something we should discuss. @abhiskum will set up a meeting.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -993,6 +998,11 @@ private Pool convertToMasterPoolImpl(SubscriptionInfo sub, Owner owner, Map<Stri
             }
 
             pool.setDerivedProvidedProducts(products);
+            //TODO: workaround to pass import spec tests. we will revisit and update this in import and
```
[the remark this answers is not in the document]
abhiskum: Fixed.

server/src/test/java/org/candlepin/sync/EntitlementImporterTest.java
```diff
@@ -101,7 +102,7 @@ public void init() {
 
         this.translator = new StandardTranslator(mockConsumerTypeCurator,
             mockEnvironmentCurator,
-            ownerCurator);
+            ownerCurator, entitlementCurator);
```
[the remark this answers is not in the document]
abhiskum: Fixed.

server/src/main/java/org/candlepin/dto/api/v1/PoolTranslator.java
abhiskum: As per discussion with @barnabycourt, @Ceiu, @nikosmoum and @wolfdale, we came up with the conclusion that we do not need to accumulate provided products in case of stacked entitlement as marketing product and stack id has one to one relationship i.e. different marketing product will not have the same stacking id and all the required provided products will be available in marketing product itself from adapter.

server/src/main/java/org/candlepin/dto/manifest/v1/PoolTranslator.java
nikosmoum: While re-reviewing this earlier today, I noticed that the way we handle n-tier here, is that we set all provided products of the hierarchy to the top marketing object.
So if ProductA has ProvidedProductB, and ProvidedProductB itself has ProvidedProductC, then we end up with this DTO structure:
```
ProductA {
    providedProducts: [ ProvidedProductB, ProvidedProductC ]
} 
```
Shouldn't it instead be like this?:
```
ProductA {
    providedProducts: [ 
        ProvidedProductB { 
            providedProducts: [ ProvidedProductC ]
        }
    ]
}
```

Again, sorry if this was already discussed in more detail and I missed it!

Ceiu: That should be correct, since this whole remapping is done to simulate the old flat structure we had before.

nikosmoum: I see, disregard my comment then!

abhiskum:
@nikosmoum Updated commit message as per your recommendation.

abhiskum:
@Ceiu We have resolved all the change requests. Please review and merge?

PR #2569, opened 2020-01-16
[F] ENT-1911 Add method to get provided products from ProductInfo

wolfdale:
- Added method to get provided products from ProductInfo. 
(Note: `getProvidedProducts` and `getDerivedProvidedProducts` will get removed from subscriptionInfo in later PR's)

PS : 
- This is done to isolate it from refresh code flow.
- This will help to fix rspec issues related to #2565 changes.

Merge order :
This or #2563

### DIFF
--- a/server/src/main/java/org/candlepin/dto/api/v1/ProductDTO.java
+++ b/server/src/main/java/org/candlepin/dto/api/v1/ProductDTO.java
@@ -41,6 +41,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.xml.bind.annotation.XmlRootElement;
 import javax.xml.bind.annotation.XmlTransient;
@@ -1080,7 +1081,13 @@ public class ProductDTO extends TimestampedCandlepinDTO<ProductDTO> implements P
         copy.setProductContent(this.getProductContent());
         copy.setDependentProductIds(this.getDependentProductIds());
         copy.setBranding(this.getBranding());
-        copy.setProvidedProducts(this.getProvidedProducts());
+
+        if (this.getProvidedProducts() != null) {
+            copy.providedProducts = new HashSet<>();
+            copy.providedProducts.addAll(this.getProvidedProducts().stream()
+                .map(prodDTO -> prodDTO.clone())
+                .collect(Collectors.toSet()));
+        }
 
         return copy;
     }
@@ -1112,7 +1119,12 @@ public class ProductDTO extends TimestampedCandlepinDTO<ProductDTO> implements P
         this.setHref(source.getHref());
         this.setLocked(source.isLocked());
         this.setBranding(source.getBranding());
-        this.setProvidedProducts(source.getProvidedProducts());
+
+        if (source.getProvidedProducts() != null) {
+            this.setProvidedProducts(source.getProvidedProducts().stream()
+                .map(prod -> new ProductDTO(prod))
+                .collect(Collectors.toSet()));
+        }
 
         return this;
     }
--- a/server/src/main/java/org/candlepin/dto/manifest/v1/ProductDTO.java
+++ b/server/src/main/java/org/candlepin/dto/manifest/v1/ProductDTO.java
@@ -38,6 +38,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.xml.bind.annotation.XmlRootElement;
 
@@ -128,6 +129,7 @@ public class ProductDTO extends TimestampedCandlepinDTO<ProductDTO> implements P
     protected Boolean locked;
 
     protected Set<BrandingDTO> branding;
+    protected Set<ProductDTO> providedProducts;
 
     /**
      * Initializes a new ProductDTO instance with null values.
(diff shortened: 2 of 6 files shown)


server/src/main/java/org/candlepin/dto/manifest/v1/ProductDTO.java
```diff
@@ -846,6 +847,92 @@ private boolean isNullOrIncomplete(BrandingDTO branding) {
...
+    public ProductDTO setProvidedProducts(Set<ProductDTO> providedProducts) {
+        if (providedProducts != null) {
+            if (this.providedProducts == null) {
+                this.providedProducts = new HashSet<>();
+            }
+            else {
+                this.providedProducts.clear();
+            }
+
+            for (ProductDTO dto : providedProducts) {
```
barnabycourt: Would it make sense to do the validation before clearing the existing list?

wolfdale: At DTO level it wouldn't make much difference as we want a fresh DTO to be returned every time with all its properties. There is still a possibility others can throw exception resulting in failure of whole operation as there are no try-catch guarding setter method invocation anywhere. Also there are multiple such methods in execution chain and failure of even one method can lead to failure of operation, which is expected.

server/src/main/java/org/candlepin/model/dto/ProductData.java
```diff
@@ -1118,6 +1189,12 @@ public ProductData populate(Product source) {
         this.setDependentProductIds(source.getDependentProductIds());
         this.setBranding(source.getBranding());
 
+        if (source.getProvidedProducts() != null) {
+            this.setProvidedProducts(source.getProvidedProducts().stream()
+                .map(prod -> this.populate(prod))
```
nikosmoum: I'm a bit confused by this. Why are we calling the populate method recursively here, which will overwrite all the data on the marketing ProductData (this) we're populating already? If it is in order to turn the Product into a ProductData, `new ProductData(Product product)` is what we should be using instead.

server/src/main/java/org/candlepin/dto/manifest/v1/ProductDTO.java
```diff
@@ -846,6 +847,92 @@ private boolean isNullOrIncomplete(BrandingDTO branding) {
...
+     * If the provided products have not yet been defined, this method returns null.
+     *
+     * Note that the collection returned by this method is a view of the collection backing this
+     * set of provided products. Elements cannot be added to the collection, but elements may be removed.
+     * Changes made to the collection will be reflected by this product DTO instance.
+     *
+     * @return
+     *  The provided products associated with this key, or null if they have not yet been defined
+     */
+    @JsonIgnore
```
nikosmoum: I know you've added the ignore annotations here, but the truth is that since the manifest API should not change (provided products should still stay in the manifest PoolDTO), there is no need to add this field on this DTO at all.
I know you've added it because the ProductInfo interface dictates it, but this is another case of asking: why does this DTO need that interface??? @Ceiu since I see you added that, what was the reason? Our manifest API is clearly different than our model, our REST API, and our adapters API, so do we need that interface here?

server/src/main/java/org/candlepin/model/dto/ProductData.java
Ceiu: Are we still using ProductData? If we can avoid it, we should.

nikosmoum: Subscription, and in extent, the adapters code is definitely using it.

Ceiu: Gross. We really need to spend some time to revisit the refactor tasks around it and finally get rid of all the PoC and v1 DTO layers. Especially with the spec-first stuff driving what is effectively a v3 layer.

server/src/main/java/org/candlepin/dto/manifest/v1/ProductDTO.java
Ceiu: Something that seems to be getting lost or misunderstood is what role the various models play in conjunction with each other, and a bit of Candlepin's development history.

First, the model we care about primarily is our data model (the entities). That drives how we store data and how we work with data at the logic level. The DTOs and various DTO-like things (adapter \*Info interfaces) are views into that model or data. We try to keep them as a subset of the data layer, but that isn't always possible due to preexisting conditions, usages, or just some legacy behavior we cannot remove or update for some reason. In any event, the data model is what drives most of the design of everything around it, not the views.

In the case of the adapter interfaces, they are a stopgap implementation to get adapter consumers (IT) to stop using our internal model objects and start using their own, for both maintenance and performance reasons. However, as soon as the adapter's model changed, we had to address the fallout of how the adapters were designed and intended to be used vs how they ended up actually being used; coupled with the fact we now had code that was reliant on multiple versions of the model (data, manifest view and adapter view). So, we were left with a few options, one of which was rewrite every bit of code that surrounded the adapters (i.e. most of the manifest processing, chunks of refresh, etc.), which, as I'm sure you can imagine, was deemed unviable given the sheer volume of such work. After several discussions we landed on the interface + existing DTO reuse we have today. @kahowell might remember and have more to add here if he cares enough.

Now, should the adapters have their own DTO implementations that we use? Maybe. Given how we seem to *want* to use the adapters, probably not, though. However, both answers to this question come with plenty of baggage and need to be properly understood before a decision is made. Perhaps the DTO layers should all be interfaces and we can implement them in whatever means we choose?

So then... getting back to the original question of why do we need the JsonIgnore annotation. Short answer is: we don't. Having extra fields is a waste of data and processing time, but it's not strictly required that we avoid it if the change is a net addition that doesn't violate any security concerns. In this specific case, I'd be curious to understand why we don't want the product model

We have other potential concerns here, such as the one @nikosmoum is seemingly alluding to here and one I brought up in the design discussion yesterday in that we should be avoiding having a view dictate the model implementation. That kind of applies here, but perhaps should be better framed as we should be working to separate the shared implementation of the two interfaces if the overlap is no longer maintainable.

Ceiu: I can go more in-depth on some of this stuff if necessary, but it's getting hard to keep up with given how messy the history and design behind all of this is; and the research involved just in writing this part took a good 90 minutes of my morning.

server/src/main/java/org/candlepin/model/dto/ProductData.java
wolfdale: @nikosmoum there are few more places where DTO methods need to be "n-tier" aware, all are now taken care of and fixed. 
@Ceiu Are we planning for ProductData to be removed?

server/src/main/java/org/candlepin/dto/manifest/v1/ProductDTO.java
wolfdale: From what I understand by looking at manifest ProductDTO which is closely tied with ProductInfo interface at multiple places involving refresh flow and manifest processing.
Are we planing to refactor this by any chance ? @Ceiu @nikosmoum 
Removing JsonIgnore as it is not strictly needed.

nikosmoum: @Ceiu Thanks for the detailed background. I suppose removing the jsonignore annotation is fine here, since net additions aren't a direct problem. Maybe it would be good to document it in the comment for future reference. Adding new fields to our APIs just because we can, but without actually using them is not something I'm a fan of (We should be more intentional with our API changes, but this is I guess a longer discussion that relates our overall API/versioning strategy).

> but perhaps should be better framed as we should be working to separate the shared implementation of the two interfaces if the overlap is no longer maintainable.

Yes, ideally, just like a view shouldn't dictate our model, a view (adapters API) should not dictate another view (manifest API) in my opinion.

server/src/main/java/org/candlepin/model/dto/ProductData.java
Ceiu: Yes. ProductData was one of the first attempts to start standardizing a DTO layer around the time that Subscription was being removed from the model. It's been almost entirely replaced by ProductDTO and the general DTO layer for manifests. However, since Subscription is still used heavily in manifests, we haven't gotten around to a proper refactor.

server/src/main/java/org/candlepin/dto/manifest/v1/ProductDTO.java
Ceiu: Something that's being glossed over in all of this is that manifests use two models simultaneously. If we spend the time to separate that out, or otherwise standardize the manifest DTO layer such that it lines up with the adapter layer (which is kind of what happens when ProductDTO inherits ProductInfo), then this mostly goes away on its own.

There are a couple ways of doing that:
- Write a translator and copy the data between the models (again)
- Decouple the manifest code from the refresh path which would drop the need for the adapter models in manifest.
- Decouple refresh from the adapter models

And so on and so forth. Lots of options to do it, but none of them are easy or efficient; and despite many pleas on my behalf, we haven't found/had time to dedicate to giving manifest the rewrite/refactor it deserves.

server/src/main/java/org/candlepin/dto/api/v1/ProductDTO.java
```diff
@@ -1035,6 +1035,8 @@ public boolean equals(Object obj) {
 
             // Check product content
             equals = equals && Util.collectionsAreEqual(this.getProductContent(), that.getProductContent());
+            equals = equals &&
+                Util.collectionsAreEqual(this.getProvidedProducts(), that.getProvidedProducts());
```
nikosmoum: Isn't the use of the `Util.collectionsAreEqual` unnecessary here and on all other DTOs (manifest ProductDTO, ProductData)? @Ceiu am I right in thinking we only need to use this method in hibernate entity classes, because hibernate collections don't implement .equals properly?

wolfdale: Removed it.

server/src/main/java/org/candlepin/dto/api/v1/ProductDTO.java
```diff
@@ -1055,6 +1056,15 @@ public int hashCode() {
...
 
+        int ppHashCode = 0;
+        Collection<ProductDTO> providedProducts = this.getProvidedProducts();
+
+        if (providedProducts != null) {
+            for (ProductDTO dto : providedProducts) {
+                ppHashCode = 31 * ppHashCode + (dto != null ? dto.hashCode() : 0);
+            }
+        }
+
```
nikosmoum: Similarly to the equals method, where we don't need to check equality in a custom way, we also don't need to calculate hashcode in a custom way. You can just add `this.getProvidedProducts()` in the HashCodeBuilder below.

server/src/main/java/org/candlepin/dto/manifest/v1/ProductDTO.java
```diff
@@ -908,7 +1011,9 @@ public int hashCode() {
             .append(this.getAttributes())
             .append(this.getDependentProductIds())
             .append(this.getBranding())
-            .append(pcHashCode);
+            .append(this.getProvidedProducts())
+            .append(pcHashCode)
+            .append(ppHashCode);
```
nikosmoum: Same here, no need to calculate hashcode above and set it here. This line is enough: `.append(this.getProvidedProducts())`.

server/src/main/java/org/candlepin/model/dto/ProductData.java
```diff
@@ -1001,7 +1072,8 @@ public int hashCode() {
             .append(this.content)
             .append(this.dependentProductIds)
             .append(this.branding)
-            .append(this.locked);
+            .append(this.locked)
+            .append(ppHashCode);
```
nikosmoum: Same here, no need for calculating hashcode and setting it. Just `append.(this.getProvidedProducts())`

PR #2572, opened 2020-01-16
Add custom openapi vendor extension to override collection type

kahowell:
Add custom openapi vendor extension to override collection type

This makes it so that if for an operation you specify a line such as:

```yaml
x-java-response-container-type: java.util.stream.Stream
```

on an operation with a response return type of

```yaml
responses:
  200:
    description: OK
    content:
      application/json:
        schema:
          type: array
          items:
            $ref: '#/components/schemas/Link'
```

then the generated return type becomes `java.util.stream.Stream<Link>`

Specifically, this goes at the same level as `operationId`, because there is not a easy way to get the vendor extension for a response incorporated into the return type.

### DIFF
--- a/buildSrc/src/main/resources/templates/returnTypeInterface.mustache
+++ b/buildSrc/src/main/resources/templates/returnTypeInterface.mustache
@@ -1,0 +1,1 @@
+{{#vendorExtensions.x-java-response-container-type}}{{vendorExtensions.x-java-response-container-type}}<{{returnBaseType}}>{{/vendorExtensions.x-java-response-container-type}}{{^vendorExtensions.x-java-response-container-type}}{{{returnType}}}{{/vendorExtensions.x-java-response-container-type}}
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/resources/templates/returnTypeInterface.mustache
```diff
@@ -0,0 +1 @@
+{{#vendorExtensions.x-java-response-stream-type}}java.util.stream.Stream<{{vendorExtensions.x-java-response-stream-type}}>{{/vendorExtensions.x-java-response-stream-type}}{{^vendorExtensions.x-java-response-stream-type}}{{{returnType}}}{{/vendorExtensions.x-java-response-stream-type}}
```
Ceiu: We can do something like:

```{{container_type}}{{^container_type}}java.util.List{{/container_type}}```

to have a default value of List so the standard behavior is retained, while still allowing us to provide our own containers for arrays.

kahowell: @Ceiu I reworked a bit, now `x-java-response-container-type` specifies the container type (leave off to get `List`.

PR #2573, opened 2020-01-16
[M] Converted ConsumerTypeResource to the OpenAPI specification (ENT-1868)

Ceiu:
- Created new specifications for the consumer type resource and
  API DTO
- Updated ConsumerTypeResource to adhere to the generated API
  specification
- Removed the existing API ConsumerTypeDTO


Ceiu:
A few notes about this PR:

1. First and foremost, this PR is contingent on #2549, and cannot be merged until that PR is merged and this PR is rebased against the updated master branch.
2. At the time of writing, the getConsumerTypes method returns a stream, and does a bit of extra work to convert the CandlepinQuery into a stream object. This will probably work (though requires more testing), but comes with some concerns that the interceptor addressed. I suspect in some cases, the DB session will be closed or otherwise preoccupied and we won't get the expect result. It'll probably be cleaner and better for the short term to just output the CandlepinQuery as we're doing today instead of bothering to convert it, and then deal with the conversion when we get around to removing it entirely.
3. The manually-created ConsumerTypeDTO still exists in this branch, yet doesn't seem to cause any problems. We may have some instrumentation problems here, or it's just a quirk in how Java loads classes and doesn't care if the same FQCN exists in multiple places, so long as they don't both load at the same time. Either way, this should be something we look into a bit closer.
4. The generated DTOs have a reeeeeeal stupid toString method that deviates heavily from what we've used up to now. Probably not a big deal, but still a gripe.
5. We should look into a way to break up that yaml file into individual files, either via self-rolled pre-processor, or some other OpenAPI/Swagger tooling. Having one massive file is not only a maintenance nightmare, but it's going to get in the way of doing parallel updates as we work through this task. I'm not looking forward to the merge conflicts we're going to run into on this front.

Ceiu:
PR updated. The manual DTO has been removed, and everything surrounding it updated or removed as appropriate. I've also used spec inheritance using allOf for the timestamp bits, but that may end up needing to change pending discussion around that particular area.

server/src/main/java/org/candlepin/dto/api/v1/ImportUpstreamConsumerDTO.java
```diff
@@ -205,7 +205,17 @@ public ImportUpstreamConsumerDTO clone() {
         ImportUpstreamConsumerDTO copy = super.clone();
 
         ConsumerTypeDTO type = this.getConsumerType();
-        copy.setConsumerType(type != null ? (ConsumerTypeDTO) type.clone() : null);
+        if (type != null) {
+            copy.setConsumerType(new ConsumerTypeDTO()
```
nikosmoum: So, we don't get a clone method??? This looks tedious.

Ceiu: Yes

Ceiu:
One more update to add a deserializer for ConsumerType that handles both bare labels as types and the proper object type.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
```diff
@@ -788,6 +789,7 @@ public ConsumerTypeDTO getType() {
      *
      * @return a reference to this DTO object.
      */
+    @JsonDeserialize(using = ConsumerTypeDeserializer.class)
```
nikosmoum: What happens when ConsumerDTO.java gets deleted and gets defined in the yaml file? Is there a way to specify custom field se/deserializers for openapi-defined entities?

server/src/main/java/org/candlepin/jackson/ConsumerTypeDeserializer.java
```diff
@@ -0,0 +1,96 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+
+
+/**
+ * The CandlepinAttributeDeserializer handles the deserialization of attributes, processing both
```
nikosmoum: The comment should get updated to reflect the new class

server/src/main/java/org/candlepin/jackson/ConsumerTypeDeserializer.java
```diff
@@ -0,0 +1,96 @@
...
+        super(valueClass);
+    }
+
+    @Override
+    public ConsumerTypeDTO deserialize(JsonParser parser, DeserializationContext context)
+        throws IOException, JsonProcessingException {
+
+        // TreeNode node = parser.readValueAsTree();
+
+        // log.debug("DESERIALIZING CONSUMER TYPE. TREENODE: {}", node);
```
nikosmoum: These should be removed probably

server/src/main/java/org/candlepin/dto/api/v1/ConsumerDTO.java
Ceiu: That's indeed a problem that needs to be solved; however is not one that is unique to this addition. We have other such annotations and custom format handlers we'll need to figure out how to add into the (de)serializers.

To some extend this is kind of kicking the can down the road for someone else to sort out, but since we have other areas that use it and need sorting, we can revisit it then.

server/src/main/java/org/candlepin/jackson/ConsumerTypeDeserializer.java
Ceiu: Fixed

server/src/main/java/org/candlepin/jackson/ConsumerTypeDeserializer.java
Ceiu: Thought I removed these already. Must have ctrl+z'd one too many times.

PR #2574, opened 2020-01-17
Always rebuild gating test docker images from scratch

nikosmoum:
- Use --no-cache when building docker images for gating tests
  to avoid using stale images, and causing build errors when
  hosted has upgraded to a newer version of candlepin.


nikosmoum:
Note: This PR, in combination with https://github.com/candlepin/candlepin-jobs/pull/53, should fix the candlepin-docker-images Jenkins build.

PR #2575, opened 2020-01-17
Lookup annotations on concrete resource implementations

kahowell:


server/src/main/java/org/candlepin/resteasy/AnnotationLocator.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Inject
+    public AnnotationLocator(Injector injector) {
+        // Maintain the insertion order for nice output in debug statement
+        internalMap = new LinkedHashMap<>();
+        this.injector = injector;
+    }
+
+    @SuppressWarnings("rawtypes")
+    public void init() {
```
Ceiu: Is there any chance this can be called multiple times by multiple threads for a given instance? If so, we should probably have the method be synchronized so we don't run the (minute?) risk of two simultaneous calls getting around the lock. If not, disregard this comment.

server/src/main/java/org/candlepin/resteasy/AnnotationLocator.java
```diff
@@ -0,0 +1,141 @@
...
+        if (hasBeenInitialized) {
+            throw new IllegalStateException("This map has already been initialized.");
+        }
+
+        List<Binding<?>> rootResourceBindings = new ArrayList<>();
+        for (final Binding<?> binding : injector.getBindings().values()) {
+            final Type type = binding.getKey().getTypeLiteral().getType();
+            if (type instanceof Class) {
+                final Class<?> beanClass = (Class) type;
```
Ceiu: Why are these all final? Is there a lambda I'm overlooking or some such?

server/src/main/java/org/candlepin/resteasy/AnnotationLocator.java
```diff
@@ -0,0 +1,141 @@
...
+        if (concreteMethod != null) {
+            T annotation = concreteMethod.getAnnotation(annotationClass);
+            if (annotation != null) {
+                return annotation;
+            }
+        }
+        return method.getAnnotation(annotationClass);
+    }
+
+    public synchronized void lock() {
```
Ceiu: I imagine the ```synchronized``` keyword is going to move up to init and then this will become private. Is there any situation where we want this to be locked externally?

server/src/main/java/org/candlepin/resteasy/AnnotationLocator.java
kahowell: This was just copy-paste from `ResourceLocatorMap`. Will remove, because it's really not providing value there.

server/src/main/java/org/candlepin/resteasy/AnnotationLocator.java
kahowell: I will move to synchronized from `lock` to `init`; that seems better overall. Another quirk copied from `ResourceLocatorMap`.

server/src/main/java/org/candlepin/resteasy/AnnotationLocator.java
kahowell: Nope, I'll push an update shortly.

Ceiu:
ack

PR #2577, opened 2020-01-22
[F] Convert DeletedConsumerResource to the OpenAPI specification (ENT-1871)

abhiskum:
- Created new specifications for the deleted consumer resource and deleted consumer DTO
 - Updated DeletedConsumerResource to implement the generated API specification
 - Deleted the existing API DeletedConsumerDTO
 - Deleted DeletedConsumerDTOTest test
 - Updated DeletedConsumerTranslator as per generated DTO


abhiskum:
This PR cannot be merged until #2549 PR is merged and this PR is rebased against the updated master branch.

server/src/main/java/org/candlepin/resource/DeletedConsumerResource.java
```diff
@@ -51,17 +40,11 @@ public DeletedConsumerResource(DeletedConsumerCurator deletedConsumerCurator,
...
-    public CandlepinQuery<DeletedConsumerDTO> listByDate(@QueryParam("date") String dateStr) {
-        return this.translator.translateQuery(dateStr != null ?
+    @Override
+    public Stream<DeletedConsumerDTO> listByDate(String dateStr) {
+        CandlepinQuery<DeletedConsumerDTO> query = this.translator.translateQuery(dateStr != null ?
             this.deletedConsumerCurator.findByDate(ResourceDateParser.parseDateString(dateStr)) :
-            this.deletedConsumerCurator.listAll(),
-            DeletedConsumerDTO.class);
+            this.deletedConsumerCurator.listAll(), DeletedConsumerDTO.class);
+        return StreamSupport.stream(query.spliterator(), false);
```
Ceiu: This doesn't actually work in all cases. Since the processing of the stream occurs later, the connection backing the CPQ can close and it'll trigger an exception instead of returning the result.

For now, just return the CandlepinQuery object as normal, and set the container type to "Iterable" in the yaml.

server/src/main/java/org/candlepin/dto/api/v1/DeletedConsumerTranslator.java
```diff
@@ -15,14 +15,16 @@
...
 
+import java.time.ZoneOffset;
+import java.util.Date;
+
 /**
  * This translator provides translation from DeletedConsumer model objects to DeletedConsumerDTOs
  */
-public class DeletedConsumerTranslator extends
-    TimestampedEntityTranslator<DeletedConsumer, DeletedConsumerDTO> {
+public class DeletedConsumerTranslator implements ObjectTranslator<DeletedConsumer, DeletedConsumerDTO> {
```
nikosmoum: Is this the trend we're following to solve the problem of "DeletedConsumerDTO is not within its bound; should extend `org.candlepin.dto.TimestampedCandlepinDTO`"? Turning the `extends TimestampedEntityTranslator` to `implements ObjectTranslator`?

server/src/main/java/org/candlepin/resource/DeletedConsumerResource.java
abhiskum: @Ceiu 
I had tested this API locally and it was working. For now, I have updated code as per your suggestion i.e. use Iterable instead of Stream. 
 
If we use Stream then would it fail in a larger response? Could you please share the scenarios in which this will throw an exception in the case of Stream.

Ceiu: The easier thing to do would be to take a look at CandlepinQueryInterceptor.

There are two risks to dropping the interceptor and just using a stream directly:
1) We give up our control over when the stream is processed vs when the connection backing it is closed. We've had issues in the past with race conditions where the connection would close and cause various exceptions when the stream is processed. I would have to dip into the git history to give you concrete examples.
2) Nested objects straight up fail in many cases. This was the driving feature here that caused the CandlepinQueryInterceptor to start using its own connection.

I think the long-term goal is to remove CPQ entirely and start using properly designed streams, but that is not something we're willing/able to address in these PRs. For the immediate term, we can move into the spec-first world by setting the container type to be Iterable, but then returning the CandlepinQuery as-is and letting the interceptor pick it up and process it safely. We can even leverage Java's covariant return typing to let the generated API interface define it as an Iterable, but our class define it as a CandlepinQuery as it does today.

server/src/main/java/org/candlepin/dto/api/v1/DeletedConsumerTranslator.java
Ceiu: I believe so. I've done a bit of tinkering now, and I can't get the API generation to do any kind of inheritance that's worth considering, which leaves is us with composition; and that, unfortunately, means manually doing all the shared things like timestamping.

server/src/main/java/org/candlepin/resource/DeletedConsumerResource.java
abhiskum: Thanks @Ceiu for details.

abhiskum:
@kahowell 
When I removed the description from the response section, openApiGenerate task failed with description required message.

api/candlepin-api-spec.yaml
```diff
@@ -193,6 +193,50 @@ paths:
...
+      tags:
+        - Consumer
+        - DeletedConsumer
+      operationId: listByDate
+      parameters:
+        - in: query
+          name: date
+          schema:
+            type: string
+          description: ''
```
kahowell: Maybe something like "filter deleted consumers to those on or after this date"?

kahowell:
@abhiskum For the response descriptions, the simple HTTP descriptions would work (e.g. `OK`, `Not Found`) if there's not already something more descriptive present.

api/candlepin-api-spec.yaml
abhiskum: @kahowell 
Added description value in the spec. Please re-review.

nikosmoum:

PR #2578, opened 2020-01-22
[2.5] 1794041: Fixed NPE during ExpiredPoolsJob execution

Januson:
- Fixed an issue that could cause a NullPointerException during
  system purpose status calculation


barnabycourt:
This is not targetting the correct BZ

PR #2580, opened 2020-01-22
[2.9] 1794043: Fixed NPE during ExpiredPoolsJob execution

Januson:
- Fixed an issue that could cause a NullPointerException during system purpose status calculation


barnabycourt:
Merging as this has been approved and is passing CI tests

PR #2581, opened 2020-01-23
[F] Convert StatusResource to the OpenAPI specification (ENT-1884)

abhiskum:
- Created new specifications for the status resource and status DTO
   - Updated StatusResource to implement the generated API specification
   - Deleted existing StatusDTO and KeycloakStatusDTO
   - Deleted existing StatusDTOTest test
   - Merge keycloak status fields into status DTO
   - Updated StatusCache to use new DTO


abhiskum:
This PR cannot be merged until #2549 PR is merged and this PR is rebased against the updated master branch.

api/candlepin-api-spec.yaml
```diff
@@ -22,6 +42,44 @@ components:
...
+        rulesSource:
+          type: string
+        rulesVersion:
+          type: string
+          example: 5.8
+        capabilities:
+          type: array
+          items:
+            type: string
+        keycloakRealm:
```
Ceiu: Something of note here: when we added the keycloak bits, we didn't want the values to be present in status if it was disabled (which it would be in most places); hence the separated DTOs. We should probably retain that.

abhiskum: @Ceiu 
Are you suggesting separated DTO with just keycloak fields or with status DTO fields plus keycloak fields?

Ceiu: An extension of StatusDTO like we had when it was manual. allOf with an inline object should work here.

abhiskum: @Ceiu 
I was under the impression that 'allOf' will generate different classes instead of inheritance. However, after exploring open API documentation, It appears we can generate inheritance using 'discriminator'.  I have updated to code to generate StatusDTO and KeycloakStatusDTO classes. However, after adding this, one additional property (discriminator property value)  coming in the response. I need to explore to avoid this additional property in the response.

Ceiu: Disregard my request here. As you noted, what we should be doing is not possible with OpenAPI short of making substantial modifications to it.

What you have here that combines the DTOs is what we're going to be doing going forward.

api/candlepin-api-spec.yaml
```diff
@@ -22,6 +42,53 @@ components:
...
+      discriminator:
+        propertyName: type
+      properties:
+        mode:
+          type: string
+        modeReason:
+          type: string
+        modeChangeTime:
+          type: string
+          format: date-time
```
Ceiu: Depending on where we land with the other date things, these may need to become a date format.

api/candlepin-api-spec.yaml
abhiskum: @Ceiu 
I had one additional query, I have already made changes to create separate DTOs using ‘discriminator’ and pushed in this PR. 

To clarify, Is your approval for combining both StatusDTO and KeycloakStatusDTO into one or for separate DTOs using ‘discriminator’?

Ceiu: Combined until we find something better.

server/src/main/java/org/candlepin/resource/StatusResource.java
```diff
@@ -157,30 +146,26 @@ public StatusDTO status() {
...
+            .version(version)
+            .release(release)
+            .standalone(standalone)
+            .rulesVersion(jsProvider.getRulesVersion())
+            .rulesSource(rulesSource != null ? rulesSource.toString() : null)
+            .mode(mode != null ? mode.toString() : null)
+            .modeReason(mcr != null ? mcr.toString() : null)
+            .modeChangeTime(mcr != null ? mcr.getTime().toInstant().atOffset(ZoneOffset.UTC) : null)
+            .managerCapabilities(caps.stream().collect(Collectors.toList()))
+            .timeUTC(new Date().toInstant().atOffset(ZoneOffset.UTC));
```
nikosmoum: There is no need to create a Date instance first and then convert it to OffsetDateTime here. `OffsetDateTime.now()` will do.

server/src/test/java/org/candlepin/cache/StatusCacheTest.java
```diff
@@ -46,8 +47,8 @@ public void setAndGet() {
         StatusCache cache = new StatusCache();
 
         StatusDTO status = new StatusDTO()
-            .setResult(true)
-            .setTimeUTC(new Date());
+            .result(true)
+            .timeUTC(new Date().toInstant().atOffset(ZoneOffset.UTC));
```
nikosmoum: Same for all the changes in this file, just use `OffsetDateTime.now()` directly.

api/candlepin-api-spec.yaml
```diff
@@ -193,6 +193,27 @@ paths:
...
+  /status:
+    get:
+      description: Returns status of the server
+      tags:
+        - Status
+      operationId: status
+      security: []
+      responses:
+        200:
+          description: ''
```
kahowell: Please remove this line

server/src/main/java/org/candlepin/resource/StatusResource.java
abhiskum: Done.

server/src/test/java/org/candlepin/cache/StatusCacheTest.java
abhiskum: Done.

api/candlepin-api-spec.yaml
abhiskum: @kahowell 
I am getting the following error after removing description from response section. 

`Execution failed for task ':api:openApiGenerate'.
> There were issues with the specification. The option can be disabled via validateSpec (Maven/Gradle) or --skip-validate-spec (CLI).
   | Error count: 1, Warning count: 0
  Errors: 
        -attribute paths.'/status'(get).responses.200.description is missing
`

kahowell: Ah, I should have looked closer! We'll need to fill in something for description to make the spec valid as you already figured out.

kahowell: For a 200, a simple description of "OK" would be fine.

nikosmoum: After some discussion, it would be better to include some description (here and in other PRs with empty descriptions like https://github.com/candlepin/candlepin/pull/2577) so that we can use this as an opportunity to improve our API documentation.

It doesn't have to be a lot, a verbose to-the-point description will do. This document has some nice guidelines/examples: https://swagger.io/docs/specification/describing-responses/

abhiskum: Done.

nikosmoum:

kahowell:
@abhiskum this needs a rebase, since I merged #2577.

abhiskum:
@kahowell Rebased with the feature branch. Please review and merge.

nikosmoum:
Needs rebase due to the docker registry migration change, which will solve jenkins.

api/candlepin-api-spec.yaml
```diff
@@ -236,6 +236,27 @@ paths:
...
```
nikosmoum: With the latest rebase, the description got accidentally reverted back to an empty string it seems. Once it's added back it should be good to merge

PR #2583, opened 2020-01-24
[M] Added the AnnotationLocator to the context listener and Candlepin module

Ceiu:
- AnnotationLocator is now bound in CandlepinModule as a singleton
- AnnotationLocator is now initialized in the context listener


Ceiu:
This PR unblocks #2549

PR #2584, opened 2020-01-24
[F] ENT-1855: Port CertificateSerialDTO to openapi spec

nikosmoum:
- Add custom deserializer for OffsetDateTime that
  supports ISO 8601 dates with optional offset, millis, or time.
- Add custom serializer for OffsetDateTime that cuts off millis and
  converts to UTC using the offset.


server/src/main/java/org/candlepin/dto/api/v1/CertificateSerialTranslator.java
```diff
@@ -56,14 +61,26 @@ public CertificateSerialDTO populate(CertificateSerial source, CertificateSerial
...
+            throw new IllegalArgumentException("dest is null");
+        }
 
-        dest.setId(source.getId() == null ? null : source.getId().toString());
-        dest.setSerial(source.getSerial() == null ? null : source.getSerial().toString());
-        dest.setExpiration(source.getExpiration());
-        dest.setCollected(source.isCollected());
-        dest.setRevoked(source.isRevoked());
+        Date created = source.getCreated();
+        dest.setCreated(created != null ?  Instant.ofEpochMilli(created.getTime()).atZone(ZoneId.systemDefault()).toLocalDate(): null);
```
Ceiu: I wonder if it'd be worth the effort (and if it's even possible) to update the entities to use LocalDate as well. While this translation works, I don't know how keen I am to be doing that everywhere.

api/candlepin-api-config.json
```diff
@@ -4,5 +4,6 @@
   "invokerPackage": "org.candlepin",
   "groupId": "org.candlepin",
   "artifactId": "candlepin-api",
-  "java8": true
+  "java8": true,
+  "dateLibrary": "java8"
```
abhiskum: @kahowell @nikosmoum 
If I remember correctly @kahowell mentioned that dateLibrary value set at openApiGenerate section in build.gradle. 
Do we need this here also?

nikosmoum: You are right, this is not needed here, so I'll remove it.

nikosmoum:
@Ceiu @kahowell @abhiskum It turns out LocalDate is a bad choice after all.

- Currently: We have been using Date in our APIs so far, which includes date and time, but no timezone information at all (it is by default UTC), BUT, jackson does read the offset information during deserializing, and creates a Date object with the time converted based on that timezone. For example, sending '2021-01-24T13:30:30+0100' as a date will result in a Date object like this: '2021-01-24T12:30:30+0000' (converted to UTC). Also, when we output dates from our API, they are like that by default (UTC).

- LocalDate ONLY includes date (no time, or timezone information), so it is out of the question because we lose all that information.

- OffsetDateTime includes date, time, and offset, which is good, but we need to work on some formats jackson seemingly doesn't support for it (with Date, everything was handled somehow).
  It supports:
  * '2021-01-24T13:30:30+05:00'
  * '2021-01-24T13:30:30.382+01:00'
  * '2021-01-24T13:30:30Z'

  But does not support dates without timezone or without time, or with offset that doesn't have colon:
  * '2021-01-24T13:30:30'
  * '2021-01-24T13:30:30.382'
  * '2021-01-24'
  * '2021-01-24T13:30:30+0500'

The resources pointed to by @kahowell ([here](https://github.com/RedHatInsights/rhsm-subscriptions/blob/f1841c2e0337f22c64535b9c3c1e73d3e38172fd/src/main/java/org/candlepin/subscriptions/resteasy/OffsetDateTimeParamConverterProvider.java) and [here](https://github.com/RedHatInsights/rhsm-subscriptions/blob/f1841c2e0337f22c64535b9c3c1e73d3e38172fd/src/main/java/org/candlepin/subscriptions/jackson/ObjectMapperContextResolver.java#L47-L48)) do not cover those issues, so I'm currently trying to make it by using OffsetDateTime and configuring a deserializer with optional time and offset using DateTimeFormatterBuilder.

server/src/main/java/org/candlepin/dto/api/v1/CertificateSerialTranslator.java
nikosmoum: I tried looking into changing CertificateSerial.java's expiration field from Date to OffsetDateTime.
We would need to change our CertificateSerialInfo Adapter inferface to support OffsetDateTime too, which would also force us to change the manifest CertificateSerialDTO to use it too, and this whole thing suddenly cascades wildly. Basically, if we change our model Dates to OffsetDateTimes, the Dates on the *Info would all also need to change. Not only that, but Date is basically used all across the codebase, so if we start changing those on the models, we'd have to change them everywhere (including model unit tests).

Compared to all that, doing these conversions in translators only (and translator tests), is a good price to pay. So I'd vote to not do that. If we _were_ going to do it though, it would probably have to be a massive effort, which I'd prefer be done separately alltogether from this epic (probably another epic).

Ceiu: Yikes, alright then.

We probably should switch to the java.time package eventually, but if it's going to be such a massive undertaking, then I agree that this PR is not the place for it.

nikosmoum:
@abhiskum @Ceiu I've updated the PR with pom.xml changes I had forgotten, and also found a way to properly parse the date in different formats without having to resort to catching exceptions.

Ceiu:
The new parsing method looks much cleaner. Just need to get Jenkins happy and I'm alright with this.

PR #2588, opened 2020-01-29
[F] Fetch provided products from pool.product (ENT-2026)

abhiskum:
- Updated Product.java to cache provided products
  - Updated Pool.java to populate providedProductDtos from product.providedProducts
  - Updated code to fetch provided products from pool.product.providedProducts
  - Updated Junit and rspec based on these changes

### DIFF
--- a/server/spec/sku_level_enable_override_spec.rb
+++ b/server/spec/sku_level_enable_override_spec.rb
@@ -21,9 +21,11 @@ describe 'SKU Level Enable Override' do
     consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'],
       consumer['idCert']['key'])
 
-    product = create_product(nil, nil,  {:attributes => { :content_override_enabled => @content_list }})
     providedProduct = create_product
     @cp.add_content_to_product(@owner['key'], providedProduct['id'], @content1['id'], false)
+    product = create_product(nil, nil,  {:attributes => { :content_override_enabled => @content_list },
+      :providedProducts => [providedProduct.id]})
+
     pool = create_pool_and_subscription(@owner['key'], product['id'], 10, [providedProduct.id])
     ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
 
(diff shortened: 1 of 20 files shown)


server/src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
```diff
@@ -461,7 +461,7 @@ public String cleanUpPrefix(String contentPrefix) throws IOException {
                 ent.getQuantity(),
                 ent.getId());
 
-            Set<Product> products = new HashSet<>(productCurator.getPoolProvidedProductsCached(pool));
+            Set<Product> products = new HashSet<>(pool.getProduct().getProvidedProducts());
```
nikosmoum: I'm trying to find out if this change is enough or not. From what I can find out, the code in the `productCurator.getPoolProvidedProductsCached()` method was added in a0986b46f076be684fc742a761d4082ca916f437 as an attempt for better performance by caching provided products in JCache, and to use these methods on the curator instead of the `pool.getProvidedProducts/derivedProvidedProducts` getters. This was later changed by @barnabycourt to use Hibernate L2 cache in 404af5066f49f9aac350cfac86454e40d5f677c7.

My assumption is that the original mechanism with the productCurator methods is no longer needed, because Product is already immutable and cachable in hibernate L2, and the only thing we'd need to do is mark the `providedProducts` collection on Product with `@Cachable` and `@Cache(usage = CacheConcurrencyStrategy.READ_ONLY)`, just like the other collections, and caching will work implicitly (without using the productCurator workarounds).

@Ceiu @barnabycourt Are my assumptions correct, or am I missing something?

Ceiu: That is my assumption as well

server/spec/sku_level_enable_override_spec.rb
```diff
@@ -165,4 +175,4 @@
     extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
     extensions_hash["1.3.6.1.4.1.2312.9.2.#{@content3['id']}.1.8"].should == '..1'
   end
-end
+end
```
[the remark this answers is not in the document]
abhiskum: Fixed.

server/src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
abhiskum: @nikosmoum Added caching annotation on providedProducts and also, remove reference of `productCurator.getPoolProvidedProductsCached()` from other places. Only 3 more places where this method still being used and these will be updated in refresh PR.

PR #2590, opened 2020-01-29
[M] 1469259: validate hypervisor update json (ENT-530)

abhiskum:


server/spec/hypervisor_check_in_spec.rb
```diff
@@ -1201,18 +1208,10 @@ def create_virtwho_client(user)
...
-    job_detail["state"].should == "FAILED"
-
-    # this is the correct version of an empy list of hypervisors
-    report = {hypervisors:[]}
-    job_detail = send_host_guest_mapping(owner, user, report.to_json())
-    job_detail["state"].should == "FINISHED"
+    virtwho = create_virtwho_client(@user)
+    lambda do
+      virtwho.hypervisor_update(@owner['key'], report.to_json())
+    end.should raise_exception(RestClient::BadRequest)
```
nikosmoum: Replacing the first assertion with this is ok, but we should probably not remove the other 2 assertions (with report `{}` and `{hypervisors:[]}`.

abhiskum: Done.

PR #2592, opened 2020-01-30
[F] ENT-2027: Refactoring products to support hierarchy of provided products. 

wolfdale:
- Changes made to Product Manager to support one level hierarchy of provided products. 
- Product hierarchy is being processed by transforming it into a flat map of products and processing them separately inside transaction boundary.

### DIFF
--- a/server/src/main/java/org/candlepin/controller/ProductManager.java
+++ b/server/src/main/java/org/candlepin/controller/ProductManager.java
@@ -48,6 +48,8 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -282,6 +284,7 @@ public class ProductManager {
      * @return
      *  A mapping of Red Hat content ID to content entities representing the imported content
      */
+    @SuppressWarnings("checkstyle:methodlength")
     @Transactional
     @Traceable
     public ImportResult<Product> importProducts(@TraceableParam("owner") Owner owner,
@@ -305,6 +308,20 @@ public class ProductManager {
             return importResult;
         }
 
+        HashMap<String, ProductInfo> orderedProductMap = new LinkedHashMap<>();
+        HashMap<String, ProductInfo> productsWithProvidedProducts = new HashMap<>();
+
+        for (ProductInfo product : productData.values()) {
+            if (product.getProvidedProducts() == null) {
+                orderedProductMap.put(product.getId(), product);
+            }
+            else {
+                productsWithProvidedProducts.put(product.getId(), product);
+            }
+        }
+
+        orderedProductMap.putAll(productsWithProvidedProducts);
+
         Map<String, Product> skippedProducts = importResult.getSkippedEntities();
         Map<String, Product> createdProducts = importResult.getCreatedEntities();
         Map<String, Product> updatedProducts = importResult.getUpdatedEntities();
@@ -313,6 +330,7 @@ public class ProductManager {
         Map<String, Product> sourceProducts = new HashMap<>();
         Map<String, List<Product>> existingVersions = new HashMap<>();
         List<OwnerProduct> ownerProductBuffer = new LinkedList<>();
+        Map<String, Product> providedEntityMap = new HashMap<>();
 
         // - Divide imported products into sets of updates and creates
         log.debug("Fetching existing products for update...");
@@ -330,6 +348,17 @@ public class ProductManager {
 
             // Prevent this product from being changed by our API
             product.setLocked(true);
+
+            if (update.getProvidedProducts() == null) {
+                providedEntityMap.put(product.getId(), product);
+            }
+            else {
+                product.getProvidedProducts().clear();
+                for (ProductInfo pInfo : update.getProvidedProducts()) {
+                    Product providedProd = providedEntityMap.get(pInfo.getId());
+                    product.addProvidedProduct(providedProd);
+                }
+            }
 
             updatedProducts.put(product.getId(), product);
             productVersions.put(product.getId(), product.getEntityVersion());
@@ -350,6 +379,17 @@ public class ProductManager {
 
                 // Prevent this product from being changed by our API
                 product.setLocked(true);
+
+                if (update.getProvidedProducts() == null) {
+                    providedEntityMap.put(product.getId(), product);
+                }
+                else {
+                    product.getProvidedProducts().clear();
+                    for (ProductInfo pInfo : update.getProvidedProducts()) {
+                        Product providedProd = providedEntityMap.get(pInfo.getId());
+                        product.addProvidedProduct(providedProd);
+                    }
+                }
 
                 createdProducts.put(product.getId(), product);
                 productVersions.put(product.getId(), product.getEntityVersion());
@@ -426,12 +466,44 @@ public class ProductManager {
             stagedEntities.put(updated.getId(), updated);
         }
 
+        // Set the correct version and fully loaded object of provided product to product.
+        for (Map.Entry<String, Product> prod : stagedEntities.entrySet()) {
+            if (orderedProductMap.containsKey(prod.getValue().getId())) {
+                ProductInfo pInfo = orderedProductMap.get(prod.getValue().getId());
+                Product product = prod.getValue();
+
+                if (pInfo.getProvidedProducts() != null) {
+                    product.getProvidedProducts().clear();
+
+                    for (ProductInfo pp : pInfo.getProvidedProducts()) {
+                        Product providedProduct = null;
+
+                        if (createdProducts.get(pp.getId()) != null) {
+                            providedProduct = createdProducts.get(pp.getId());
+                        }
+
+                        if (skippedProducts.get(pp.getId()) != null) {
+                            providedProduct = skippedProducts.get(pp.getId());
+                        }
+
+                        if (updatedProducts.get(pp.getId()) != null) {
+                            providedProduct = updatedProducts.get(pp.getId());
+                        }
+
+                        if (providedProduct != null) {
+                            product.addProvidedProduct(providedProduct);
+                        }
+                    }
+                }
+            }
+        }
+
         // Persist our staged entities
         // We probably don't want to evict the products yet, as they'll appear as unmanaged if
         // they're used later. However, the join objects can be evicted safely since they're only
         // really used here.
         log.debug("Persisting product changes...");
-        this.productCurator.saveAll(stagedEntities.values(), true, false);
+        persistInOrder(stagedEntities.values());
         this.ownerProductCurator.saveAll(ownerProductBuffer, true, true);
 
         // Perform bulk reference update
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/controller/ProductManager.java
```diff
@@ -262,6 +262,69 @@ public Product updateProduct(ProductDTO update, Owner owner, boolean regenerateE
...
+     * @param importedContent
+     *  A mapping of Red Hat content ID to content instances to use to lookup and resolve content
+     *  references on the provided product DTOs.
+     *
+     * @return
+     *  A mapping of Red Hat content ID to content entities representing the imported content
+     */
+    @Transactional
+    @Traceable
+    public ImportResult<Product> importProducts(@TraceableParam("owner") Owner owner,
```
nikosmoum: What exactly are we achieving with this? the original `productData` map that enters this method is already a flat map (contains marketing and provided products equally, as created by the `ImportedEntityCompiler`). This method is simply splitting them up into two maps, and calls the same method for both of them, then merging them together again. So we do the same processing for both maps, with the behaviour unchanged. Is the PR a WIP to get N-tier working?

wolfdale: As discussed N-Tier support for multi-level hierarchy is being scope creep at this point, which need to be taken care of
in separate task.
Currently in this PR we are supporting one level of product hierarchy (ie. products having provided products).
In this PR I am splitting the incoming product Map into two maps of products having provided products and product
with no provided products (`productsWithNoProvidedProducts` & `productsWithProvidedProducts`). These two maps are then processed separately, This is necessary to make sure that the child product (eng product) are persisted into DB (regardless of ordering in the input map) before the main product (mkt product).

nikosmoum: Ah, I see, so this is a preemptive change to make sure the order is correct (eng. products first, then marketing), which otherwise could have caused issues?

wolfdale: Yes, the ordering of the products inside `productData` is unpredictable, this change make sure the eng. products are processed and persisted in DB before the marketing products.

wolfdale: There is another solution where we set the dummy provided product, in product Update/Create loops. This dummy product is then replaced with original provided products which can be found in one of three lists (ie created, updated, skipped lists) at the time of staging entities. 

This allow us to process `productData` as it is (without worrying about the levels of hierarchy), 
also doesn't require any hibernate cascade to persist the nested entities. This can be easily extended to support N-tier in future where we can process the staged entities in order of depth (bottom to top).

server/src/main/java/org/candlepin/controller/ProductManager.java
```diff
@@ -646,6 +721,41 @@ private Product applyProvidedProductChanges(Product entity, ProductDTO update, O
...
+     *  The updated product entity.
+     */
+    private Product applyProvidedProductChanges(Product entity, ProductInfo update, Owner owner) {
+
+        if (update.getProvidedProducts() != null && !update.getProvidedProducts().isEmpty()) {
+            entity.getProvidedProducts().clear();
+
+            for (ProductInfo pInfo : update.getProvidedProducts()) {
+                if (pInfo != null && pInfo.getId() != null) {
+                    Product newProd = this.ownerProductCurator.getProductById(owner,
```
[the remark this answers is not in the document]
wolfdale: This is not needed now.

nikosmoum:

wolfdale:
@candlepin-bot  retest this please

Ceiu:
nack on the bulk of this.

I dedicated some time to playing with this and exploring some of the things we discussed in the meeting last week, and it seems the complications you were hitting were due to flushing partial product update states, rather than doing a full flush at the end. Additionally, a lot of the product collection was being done in the wrong spot, and only 1-tier, which was both more work today, and additional work in the future.

By updating the ImportedEntityProvider to consider provided products when adding new products, we can avoid doing the collection in ProductManager.importProducts. Granted, there's a fair argument to be made here that importProducts should import everything associated with a given product, but there's also a debate to be made about whether or not nested entities should be reflected in that. Anyway, this change ends up being a one-liner since we already have a recursive structure for us:

ImportedEntityCollector.java:165
```
                // Add this product's provided products...
                this.addProducts(product.getProvidedProducts());
```

Secondly, the entirety of the changes in ProductManager end up being just a change to how flushing is done:

ProductManager.java:506
```
        this.productCurator.saveAll(products, false, false);
        this.productCurator.flush();
```

And that's it. We don't need to worry about doing ordering or properly traversing the tree or whatever. Hibernate will do that for us so long as we aren't flushing early. The only concern we have at this point is memory usage; but that will depend on the average number of products we expect to be refreshing in a high-load system. If there is a scenario where this falls apart or otherwise doesn't complete as expected, then we should call that out *and* add a test to both expose it and verify when it is fixed.

One thing we still haven't sorted out here is the versioning issues, and how we need to cascade back up the product stack as provided products are updated (make sure that's being done properly). I've sent an e-mail to the list to discuss a couple of ways to go about this; but either way, that's a very major area we need to keep an eye on.

As an aside there are a few cleanup things we should be doing in this branch based on the amount of repeat code and structuring I've been seeing:

For n-tier fetching of provided products as a flat map, rather than have a bunch of utility methods in every area that needs it, this is something that could easily live on the product itself:

```
    /**
     * Retrieves a flat set consisting of all products provided by this product or any
     * product it provides. If this product does not provide any products, this method
     * returns an empty set.
     *
     * @return
     *  the set of all products in the tree of products provided by this product.
     */
    public Set<Product> getFlattenedProvidedProduts() {
        Set<Product> provided = new HashSet<>();

        this.buildFlattenedProvidedProductsSet(provided);

        return provided;
    }

    /**
     * Recursive provided products set builder
     */
    private void buildFlattenedProvidedProductsSet(Set<Product> products) {
        if (this.providedProducts != null) {
            for (Product provided : this.providedProducts) {
                products.add(provided);
                provided.buildFlattenedProvidedProductsSet(products);
            }
        }
    }
```

Now this could probably be improved (both in name and probably performance), but this would provide a one-liner for all the various places that need to just fetch the entire tree of provided products as a flat set, and would clean up a ton of areas where we do this logic by hand.

wolfdale:
I have tried your approach as it is. it does not seem to work, since here we did not apply any provided products inside marketing product during processing in Update/Create loop, and it also does not take care of product version replacement.
Complication is not only with the flushing of partial product update or persisting the hierarchy, rather it is building the correct set of products entities for persistence and there is more to it.
Let us suppose we have p1 and p2 provided products and m1 marketing product having provided product p1 & p2 inside it.Assuming everything in the product data map is in random order.

To summarize complications -
1) When we process m1 how do we get the p1 and p2 references to be set in m1's provided product ?
Hence my approach was to have a map of processed entities, which contains the already processed provided products p1 and p2. (Question - How do I already know that I have processed P1 and p2 ? I have ordered the incoming productData map) So we can use it while processing m1.
```
if (update.getProvidedProducts() == null) {
  providedEntityMap.put(product.getId(), product);
}
else {
  product.getProvidedProducts().clear();
  for (ProductInfo pInfo : update.getProvidedProducts()) {
    Product providedProd = providedEntityMap.get(pInfo.getId());
    product.addProvidedProduct(providedProd);
  }
}
```
2) The product version replacement - suppose there is Px version found in DB for p2, now as the code is written it directly remove the p2 from staged entities and replace it with Px. There is a code comment on the loops for product version processing.
```
        // Process the created group...
        // Check our created set for existing versions:
        //  - If there's an existing version, we'll remove the staged entity from the creation
        //    set, and stage an owner-product mapping for the existing version
        //  - Otherwise, we'll stage the new entity for persistence by leaving it in the created
        //    set, and stage an owner-product mapping to the new entity
``` 

Now the problem is earlier while processing the m1 we have placed object reference of p2, which is now also need to be replaced with px inside m1. 
```
for (Map.Entry<String, Product> prod : stagedEntities.entrySet()) {
            if (orderedProductMap.containsKey(prod.getValue().getId())) {
                ProductInfo pInfo = orderedProductMap.get(prod.getValue().getId());
                Product product = prod.getValue();

                if (pInfo.getProvidedProducts() != null) {
                    product.getProvidedProducts().clear();

                    for (ProductInfo pp : pInfo.getProvidedProducts()) {
                        Product providedProduct = null;

                        if (createdProducts.get(pp.getId()) != null) {
                            providedProduct = createdProducts.get(pp.getId());
                        }

                        if (skippedProducts.get(pp.getId()) != null) {
                            providedProduct = skippedProducts.get(pp.getId());
                        }

                        if (updatedProducts.get(pp.getId()) != null) {
                            providedProduct = updatedProducts.get(pp.getId());
                        }

                        if (providedProduct != null) {
                            product.addProvidedProduct(providedProduct);
                        }
                    }
                }
            }
        }
```
The reason behind this is that the product get replaced if we have existing version of it in DB, it does not replace the product if it is present in mkt product hierarchy to do so we would require a lookup to identify which products needs replacement in hierarchy. To overcome this we finally set the provided product (inside a product hierarchy), after all product entities are staged, this is done by looking up what kind of hierarchy we had in incoming ordered map.

As we still haven't sorted out the versioning issues around products, but logically every occurrence of p2 is needed to be replaced with Px wherever it occurs.
There are bits and pieces that are currently not that polished in code right now, but the issues that are highlighted here could be the starting point.

Ceiu:
Before I jump into individual or focused comments, are there any tests around this behavior? If not, that is mandatory before this gets merged, as if it's a point of contention, there needs to be a constant way to test and verify it.

Ceiu:
Alright, so for the bulk of this, one thing that needs to be made very clear is that the versioned lookups *have* to be considered untouchable. That is, if two products match according to the version lookup, then they match, and no modification should be made from there. As soon as that is violated, we're no longer in a safe place.

The bulk of this code being updated was designed with three core tasks in mind:
- Identify which products are net-new, which are being updated, and which are entirely unchanged.
- For products that are new or being updated, do we already have an existing product that matches the final state the product will be in? If so, we should use the existing instance, and discard the pending reference.
- Store the actual updated or new products, and in the case of changed products, update references accordingly.

...and so far nothing in the problem set indicates we should be changing that. The only extra piece I see -- since we've decided to plow ahead with the tree-style versioning -- is that we're going to need a recursive step that can walk back up the tree to update products referencing products we've just imported, as much as I dislike running through this step multiple times.

Much like ContentManager's importContent, I feel this is going to end up calling into itself a few times per import. Given how critical this portion is, I think it's time to pump the brakes and formalize a design for how the import bits work. It's imperative we get this right straight away, as fixing it later is going to be causing downtime for lots of teams both in and outside of RH.

nikosmoum:
I agree with formalized design on this, because its a critical piece of code that we can't get wrong.
Also, btw, the current getEntityVersion on this branch is not recursively calling itself for provided products; instead [hashCode is called](https://github.com/candlepin/candlepin/blob/08124b0003b5e5240ab8f99ff5e3a97661728469/server/src/main/java/org/candlepin/model/Product.java#L1289). We should fix that.

wolfdale:
@nikosmoum yes I am already aware of this error. I too found this while going through entity versioning calculation.

wolfdale:
Closing this PR as discussed.

PR #2594, opened 2020-01-30
1792529: Virt-who is not able to upload the Hypervisor list if the ma…

wottop:
…nifest has content access mode as org_environment

This was a designed behavior originally. The requirements have now changed.

### DIFF
--- a/server/src/main/java/org/candlepin/async/tasks/HypervisorUpdateJob.java
+++ b/server/src/main/java/org/candlepin/async/tasks/HypervisorUpdateJob.java
@@ -23,7 +23,6 @@ import org.candlepin.async.JobConstraints;
 import org.candlepin.async.JobExecutionContext;
 import org.candlepin.async.JobExecutionException;
 import org.candlepin.auth.Principal;
-import org.candlepin.common.exceptions.BadRequestException;
 import org.candlepin.dto.ModelTranslator;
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/async/tasks/HypervisorUpdateJob.java
```diff
@@ -120,16 +119,6 @@ public Object execute(final JobExecutionContext context) throws JobExecutionExce
                 return "Nothing to do. Owner does not exist";
             }
 
-            if (owner.isAutobindDisabled() || owner.isContentAccessEnabled()) {
```
nikosmoum: Are we sure we want to completely remove the 'owner autobind disabled' functionality? I see that the change (43c11bac9095d0913414c77af5b7235947b87e36) that introduced the golden ticket restriction only added the second part of the if statement, while the `owner.isAutobindDisabled()` was still there.

wottop: For the sake of creating the hypervisor/guest relationship, it makes no sense to block for autobind disabled.

This is left behind from when we handled guest migrations differently.

nikosmoum: ack

wottop:
buildbot: retest this please

PR #2596, opened 2020-01-30
ENT-1872: Port DistributorVersionResource to spec-first

Januson:
- Add yaml definitions of resource and DTOs
 - Update Translator
 - Fix broken tests


server/src/main/java/org/candlepin/dto/api/v1/DistributorVersionTranslator.java
```diff
@@ -60,28 +66,43 @@ public DistributorVersionDTO populate(DistributorVersion source, DistributorVers
...
-                for (DistributorVersionCapability capability : capabilities) {
-                    if (capability != null) {
-                        dest.addCapability(new DistributorVersionDTO.DistributorVersionCapabilityDTO(
-                            capability.getId(), capability.getName()));
-                    }
-                }
-            }
-            else {
-                dest.setCapabilities(Collections.emptySet());
+        if (modelTranslator != null) {
```
abhiskum: @Januson 
`modelTranslator` never used in this code. Any specific reason for this null check?

server/src/main/java/org/candlepin/resource/DistributorVersionResource.java
```diff
@@ -139,26 +111,16 @@ else if (!StringUtils.isBlank(capability)) {
             this.translator.getStreamMapper(DistributorVersion.class, DistributorVersionDTO.class));
     }
 
-    @ApiOperation(notes = "Deletes a Distributor Version", value = "delete")
-    @ApiResponses({ @ApiResponse(code = 400, message = ""), @ApiResponse(code = 404, message = "") })
-    @DELETE
-    @Produces(MediaType.WILDCARD)
-    @Path("/{id}")
-    public void delete(@PathParam("id") String id,
-        @Context Principal principal) {
```
abhiskum: @Januson What was to the purpose of `@Context Principal principal` in this API?

server/src/main/java/org/candlepin/dto/api/v1/DistributorVersionTranslator.java
Januson: No reason. Just forgot to remove it. Done.

server/src/main/java/org/candlepin/resource/DistributorVersionResource.java
Januson: Principal is unused so it should be save to remove. Just including it should have any side effects.

api/candlepin-api-spec.yaml
```diff
@@ -193,6 +193,143 @@ paths:
...
+              $ref: '#/components/schemas/DistributorVersionDTO'
+      responses:
+        200:
+          description: A successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DistributorVersionDTO'
+        409:
+          description: A distributor version already exists
```
nikosmoum: In this case, the code throws a `BadRequestException`, which is 400, not 409.

api/candlepin-api-spec.yaml
```diff
@@ -263,6 +400,40 @@ components:
...
+          properties:
+            id:
+              type: string
+            name:
+              type: string
+            displayName:
+              type: string
+            capabilities:
+              type: array
+              uniqueItems: true
```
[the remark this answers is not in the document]
Ceiu: To some extent, maybe this is an area where we should evaluate our logic and clean it up ourselves. However, that aside, we could always use the container typing as a short-term (heh) patch to deal with this.

It would be nice if we could stay completely within the bounds of the OpenAPI spec, but this could be another shortcoming.

server/src/main/java/org/candlepin/dto/api/v1/DistributorVersionTranslator.java
```diff
@@ -60,28 +67,43 @@ public DistributorVersionDTO populate(DistributorVersion source, DistributorVers
...
-                    if (capability != null) {
-                        dest.addCapability(new DistributorVersionDTO.DistributorVersionCapabilityDTO(
-                            capability.getId(), capability.getName()));
-                    }
-                }
-            }
-            else {
-                dest.setCapabilities(Collections.emptySet());
-            }
+    private OffsetDateTime toDateTime(Supplier<Date> source) {
```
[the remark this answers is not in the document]
Ceiu: If we have a use of the utility such that throwing a supplier at it is easier or cleaner than just using the utility method as a mapper, then it makes sense to do so. However, I would bet that most, if not all, of our uses are going to be just raw one-off conversions for DTO translation.

nikosmoum:
PR marked as revisit due to another task needed to be done first (ENT-2060).

api/candlepin-api-spec.yaml
```diff
@@ -263,6 +400,40 @@ components:
...
+              items:
+                $ref: '#/definitions/DistributorVersionCapabilityDTO'
+
+    DistributorVersionCapabilityDTO:
+      description: A DTO representation of the DistributorVersionCapability
+      type: object
+      required:
+        - id
+        - name
```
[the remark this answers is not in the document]
nikosmoum: As we've discussed before this is one area that the current dto does perform validation (name must not be null or empty), so we should keep it as such. The id though does not get validated, so there is no need to include it.

So with that, this section would become:
```
      required:
        - name
      properties:
        id:
          type: string
        name:
          type: string
          minLength: 1
```
The PR I just posted (https://github.com/candlepin/candlepin/pull/2630) introduces a DTOValidator which can be used on resources to validate incoming DTOs against the constraints generated by the `required` (adds `@NotNull` annotation), and `minLength` (adds `@Size(min=1)` annotation to make sure the string is not empty). So I'd suggest to rebase and merge this PR after that one, in order to apply the DTO validation on the `DistributorVersionResource.create/update` methods.

api/candlepin-api-spec.yaml
```diff
@@ -263,6 +400,40 @@ components:
...
 
+    DistributorVersionDTO:
+      description: A DTO representation of the DistributorVersion
+      allOf:
+        - $ref: '#/components/schemas/TimestampedEntity'
+        - type: object
+          required:
+            - displayName
+            - id
+            - name
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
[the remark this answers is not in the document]

server/src/main/java/org/candlepin/dto/api/v1/DistributorVersionTranslator.java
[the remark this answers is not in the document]
Januson: Done. Supplier was a leftover from experiment I forgot to rollback. Util function expects Date.

api/candlepin-api-spec.yaml

api/candlepin-api-spec.yaml
```diff
@@ -500,6 +636,39 @@ components:
...
+        - $ref: '#/components/schemas/TimestampedEntity'
+        - type: object
+          required:
+            - name
+          properties:
+            id:
+              type: string
+            name:
+              type: string
+              minLength: 1
```
nikosmoum: Same here, only the capability name field is validated for being empty, not the distributor verison name

server/src/main/java/org/candlepin/dto/api/v1/DistributorVersionTranslator.java
```diff
@@ -60,28 +63,38 @@ public DistributorVersionDTO populate(DistributorVersion source, DistributorVers
...
-        dest.setDisplayName(source.getDisplayName());
+        if (dest == null) {
+            throw new IllegalArgumentException("dest is null");
+        }
 
-        if (modelTranslator != null) {
+        dest.id(source.getId())
+            .name(source.getName())
+            .displayName(source.getDisplayName())
+            .created(Util.toDateTime(source.getUpdated()))
```
nikosmoum: This is setting the source's updated date to the dest's created date. it should be setting the created date

api/candlepin-api-spec.yaml
```diff
@@ -500,6 +636,39 @@ components:
...
+            displayName:
+              type: string
+            capabilities:
+              type: array
+              uniqueItems: true
+              items:
+                $ref: '#/definitions/DistributorVersionCapabilityDTO'
+
+    DistributorVersionCapabilityDTO:
```
nikosmoum: One more thing I forgot to mention. Now that our DTOs don't extend CandlepinDTO, which had a `@JsonFilter` annotation that ensures the dynamic property filter is used during their serialization, we need to indirectly add that annotation to all DTOs we port to openapi.

So both DistributorVersionDTO and DistributorVersionCapabilityDTO should have a mixin added in the `JsonProvider.addMixInAnnotationsForDTOs` method just like all other DTOs

api/candlepin-api-spec.yaml
```diff
@@ -257,6 +257,142 @@ paths:
...
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: A successful operation
+          content: {}
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -500,6 +636,39 @@ components:
             principalName:
               type: string
 
+    DistributorVersionDTO:
+      description: A DTO representation of the DistributorVersion
+      allOf:
+        - $ref: '#/components/schemas/TimestampedEntity'
+        - type: object
+          required:
+            - name
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml

server/src/main/java/org/candlepin/dto/api/v1/DistributorVersionTranslator.java
Januson: Done.

api/candlepin-api-spec.yaml

server/src/main/java/org/candlepin/resource/DistributorVersionResource.java
```diff
@@ -170,14 +132,8 @@ public DistributorVersionDTO create(
...
-    @ApiOperation(notes = "Updates a Distributor Version", value = "update")
-    @PUT
-    @Consumes(MediaType.APPLICATION_JSON)
-    @Produces(MediaType.APPLICATION_JSON)
-    @Path("/{id}")
-    public DistributorVersionDTO update(@PathParam("id") String id,
-        @ApiParam(name = "distributorVersion", required = true) DistributorVersionDTO dto,
-        @Context Principal principal) {
+    @Override
+    public DistributorVersionDTO update(String id, DistributorVersionDTO dto) {
```
nikosmoum: One more thing I forgot (sorry, I know I'm asking for changes multiple times instead of all at once, but it's hard to keep in mind everything that needs to be in line. I'll try to take some notes of a complete list of things to look for in each openapi port PR review from now on!): 

Now that we've marked the DistributorVersionCapabilityDTO's name field as `required: true` and `minLength: 1`, this is going to generate some validation annotations, but we still need to manually run the validation (with `DTOValidator`) on the `created/update` endpoints with: `validator.validateConstraints(dto);`. Also, the old DTO is validating that none of the capability DTOs set by the setCapabilities setter is null, so we also have to check this ourselves using `validator.validateCollectionElementsNotNull(dto::getCapabilities);`

Januson: Done.

PR #2600, opened 2020-02-03
[M] Upgraded Resteasy to 4.4.2 (ENT-1918)

Ceiu:
- Resteasy has been upgraded to version 4.4.2
- Updated the VerifyAuthorizationFilter to no longer use internal
  Resteasy interfaces, but instead pull argument details from the
  standardized request context


Ceiu:
There may be some spec tests that need to be addressed here, depending on the import tests that have been flagged elsewhere for being flaky.

server/src/main/java/org/candlepin/resteasy/filter/VerifyAuthorizationFilter.java
```diff
@@ -65,6 +65,7 @@
 @Priority(Priorities.AUTHORIZATION)
 public class VerifyAuthorizationFilter extends AbstractAuthorizationFilter {
     private static final Logger log = LoggerFactory.getLogger(VerifyAuthorizationFilter.class);
+
     private StoreFactory storeFactory;
     private ResourceLocatorMap locatorMap;
```
Januson: locatorMap is no longer used.

server/src/main/java/org/candlepin/resteasy/filter/CandlepinQueryInterceptor.java
```diff
@@ -64,45 +65,78 @@ public CandlepinQueryInterceptor(
...
                     session.close();
                 }
+
                 throw e;
             }
+
         }
     }
 
+    // @Override
```
Januson: Forgotten commented code.

server/src/main/java/org/candlepin/resteasy/filter/VerifyAuthorizationFilter.java
```diff
@@ -111,39 +110,79 @@ public void runFilter(ContainerRequestContext requestContext) {
...
-            throw new IseException("Method " + method.getName() + " not registered as RESTful.");
-        }
+        Map<String, List<String>> pathParams = null;
+        Map<String, List<String>> queryParams = null;
+        Map<String, List<String>> headers = null;
 
-        MethodInjector methodInjector = injectorFactory.createMethodInjector(locator, resourceFactory);
-        HttpResponse response = ResteasyProviderFactory.popContextData(HttpResponse.class);
-        Object[] args = methodInjector.injectArguments(request, response);
+        for (int i = 0; i < annotations.length; ++i) {
```
Januson: This could be enhanced for

server/src/main/java/org/candlepin/resteasy/filter/VerifyAuthorizationFilter.java
Ceiu: Fixed. Removed this line and the entire ResourceLocatorMap file.

server/src/main/java/org/candlepin/resteasy/filter/CandlepinQueryInterceptor.java
Ceiu: Removed

server/src/main/java/org/candlepin/resteasy/filter/VerifyAuthorizationFilter.java
Ceiu: I suppose, but this was something that was pre-existing that would get (immeasurably) slower by a change, so I opted to just leave it as-is.

Ceiu:
Rebased against master to hopefully address the issues we were seeing with checkstyle failing.

Ceiu:

Ceiu:

nikosmoum:

server/pom.xml
```diff
@@ -197,6 +197,18 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.candlepin</groupId>
+      <artifactId>api</artifactId>
+      <version>unspecified</version>
```
nikosmoum: This is more relevant to the spec-first branch/feature than to this, but shouldn't this have a version here? Do/Should we declare the api as a maven module?

PR #2612, opened 2020-02-13
[M] Use dependencyManagement plugin to mange versions (ENT-2029)

abhiskum:
- Define dependencies with version in the dependencyManagement
  - Remove dependency version value from build.gradle files (root, candlepin and checks projects)


Ceiu:
I have no immediate problem with what I saw (though I'm fairly inexperienced with Gradle and Maven), however I would like to request that we do not merge this until after the RE4.4.2 upgrade goes in, and then we update this PR around that set of changes.

barnabycourt:
Is the goal to support importing maven BOMs? If so, what are the benefits of this method vs migrating to Gradle 6 & the "platform" support that is provided there: https://docs.gradle.org/6.0/userguide/platforms.html#sub:bom_import

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                // JCL for example.
+                "org.slf4j:jcl-over-slf4j",
+                "org.slf4j:log4j-over-slf4j",
+                "org.slf4j:slf4j-api",
+                "net.logstash.logback:logstash-logback-encoder",
+        ],
+        javax                 : [
+                "org.hibernate.javax.persistence:hibernate-jpa-2.1-api",
+                "javax.transaction:jta",
+                "javax.persistence:javax.persistence-api",
```
kahowell: Please remove, pulled in transitively by hibernate-core.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "javax.xml.bind:jaxb-api",
```
kahowell: Please remove, pulled in transitively by hibernate-core.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "commons-io:commons-io",
+                "commons-lang:commons-lang",
+        ],
+        guice                 : [
+                "com.google.inject.extensions:guice-assistedinject",
+                "com.google.inject.extensions:guice-multibindings",
+                "com.google.inject.extensions:guice-servlet",
+                "com.google.inject.extensions:guice-throwingproviders",
+                "com.google.inject.extensions:guice-persist",
+                "com.google.inject:guice",
```
kahowell: Please remove, pulled in transitively by various deps.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "aopalliance:aopalliance",
+                "javax.inject:javax.inject",
```
kahowell: Please remove, pulled in transitively by guice.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+        ],
+        liquibase             : "org.liquibase:liquibase-core",
+        liquibase_slf4j       : "com.mattbertolini:liquibase-slf4j",
+        oauth                 : [
+                "net.oauth.core:oauth",
```
kahowell: Please remove, pulled in transitively by oauth-provider.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "net.oauth.core:oauth-provider",
+        ],
+        collections           : "com.google.guava:guava:25.1-jre",
+        hibernate_validator_ap: "org.hibernate.validator:hibernate-validator-annotation-processor",
+        checkstyle            : [
+                "com.puppycrawl.tools:checkstyle",
+                "com.github.sevntu-checkstyle:sevntu-checks"
```
kahowell: Please remove, pulled in transitively by checkstyle.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "org.keycloak:keycloak-adapter-core",
+                "org.keycloak:keycloak-common",
+                "org.keycloak:keycloak-core",
+                "org.keycloak:keycloak-adapter-spi",
+                "org.keycloak:keycloak-servlet-adapter-spi",
+                "org.apache.httpcomponents:httpcore",
+                "org.bouncycastle:bcpkix-jdk15on",
+                "org.bouncycastle:bcprov-jdk15on",
+                "org.keycloak:keycloak-jaxrs-oauth-client",
+                "org.keycloak:keycloak-authz-client",
```
kahowell: Please remove all but necessary. I suspect only keycloak-servlet-filter-adapter is necessary. I suggest starting with the obviously transitive deps and then removing and seeing if you can still build, until you reach a minimum set.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "org.hibernate.common:hibernate-commons-annotations",
+                "net.bytebuddy:byte-buddy",
+                "org.javassist:javassist",
+                "com.fasterxml:classmate",
+                "com.mchange:c3p0",
+                "com.mchange:mchange-commons-java",
+                "org.dom4j:dom4j",
+                "org.jboss:jandex",
+                "org.jboss.logging:jboss-logging",
+                "javax.activation:javax.activation-api",
```
kahowell: Please only leave hibernate-c3po, the rest should be pulled in transitively.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "org.apache.activemq:artemis-commons",
+                "org.apache.activemq:artemis-selector",
+                "org.apache.activemq:artemis-stomp-protocol",
+                "org.apache.activemq:artemis-journal",
+                "io.netty:netty-all",
+                "commons-beanutils:commons-beanutils",
+                "org.jctools:jctools-core",
+                "org.jgroups:jgroups",
+                "org.apache.geronimo.specs:geronimo-json_1.1_spec",
+                "org.apache.johnzon:johnzon-core",
```
kahowell: Please leave only artemis-server and artemis-stomp-protocol. The rest should be pulled in transitively.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+        ],
+        amqp                  : [
+                "org.apache.qpid:qpid-common",
+                "org.apache.qpid:qpid-client",
+                "org.apache.mina:mina-core",
+                "org.apache.mina:mina-filter-ssl",
+                "geronimo-spec:geronimo-spec-jms",
```
kahowell: Please remove all except for qpid-client. The rest should be transitive.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "org.jboss.resteasy:resteasy-core-spi",
+                "org.jboss.resteasy:resteasy-jaxb-provider",
+                "org.jboss.resteasy:resteasy-guice",
+                "org.jboss.resteasy:resteasy-atom-provider",
+                "org.jboss.resteasy:resteasy-multipart-provider",
+                "org.scannotation:scannotation",
+                "org.apache.httpcomponents:httpclient",
+                "org.apache.james:apache-mime4j-core",
+                "org.apache.james:apache-mime4j-dom",
+                "org.apache.james:apache-mime4j-storage",
```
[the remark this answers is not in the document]
abhiskum: Done.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "javax.mail:mail",
```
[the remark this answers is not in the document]
abhiskum: Done.

server/build.gradle
```diff
@@ -160,92 +160,42 @@ dependencies {
...
-    testImplementation "org.bouncycastle:bcprov-jdk15on:1.60"
+    testImplementation "org.bouncycastle:bcpkix-jdk15on"
+    testImplementation "org.bouncycastle:bcprov-jdk15on"
 
     // Testing DB Drivers
-    testRuntime "org.hsqldb:hsqldb:2.3.2"
+    testRuntime "org.hsqldb:hsqldb"
 }
 
 task pom {
```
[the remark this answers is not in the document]
abhiskum: Done.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
-    javax_validation: "javax.validation:validation-api:2.0.1.Final",
-    jmock: [
-        "org.jmock:jmock:2.5.1",
-        "org.jmock:jmock-junit4:2.5.1",
-    ],
-    validator: [
-        "org.hibernate.validator:hibernate-validator:6.1.0.Final",
-        "org.hibernate.validator:hibernate-validator-annotation-processor:6.1.0.Final",
-    ],
+        antlr                 : "antlr:antlr",
```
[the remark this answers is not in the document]
abhiskum: Done.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+        resteasy              : [
+
+                "org.jboss.resteasy:resteasy-core",
+                "org.jboss.resteasy:resteasy-core-spi",
```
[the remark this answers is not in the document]
abhiskum: Done.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "org.jboss.resteasy:resteasy-jaxb-provider",
+                "org.jboss.resteasy:resteasy-guice",
+                "org.jboss.resteasy:resteasy-atom-provider",
+                "org.jboss.resteasy:resteasy-multipart-provider",
+                "org.scannotation:scannotation",
```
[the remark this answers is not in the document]
abhiskum: Done.

build.gradle
```diff
@@ -21,137 +21,185 @@ plugins {
...
+                "org.apache.httpcomponents:httpclient",
```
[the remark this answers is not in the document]
abhiskum: Done.

build.gradle
abhiskum: After removing "com.github.sevntu-checkstyle:sevntu-checks",  checkstyleMain and checkstyleTest tasks are failing.

build.gradle
abhiskum: Done.

build.gradle
abhiskum: Removed all except "com.mchange:c3p0" as it is required for Junit.

build.gradle
abhiskum: Done.

build.gradle
abhiskum: Removed all except "geronimo-spec:geronimo-spec-jms" as it is required for classes in javax.jms package.

abhiskum:
@kahowell Thanks for suggesting to remove following from build.gradle. 

` configurations.matching { !it.name.startsWith("checkstyle")}.all { 
     dependencies.matching { !it.name.startsWith("junit")}.all { 
         transitive = false 
     } 
 } `

I was trying to remove transitive dependencies without this change and build was failing. 

I have updated PR as per your suggestion. Please review.

kahowell:
Needs:
 - [ ] }`.withXml { dependencyManagement.pomConfigurer.configurePom(asNode()) }` applied to both common/build.gradle and server/build.gradle (I suspect you may already have this locally?)
 - [ ] `git fetch; git rebase origin/master`
 - [ ] `./gradlew pom`

and then it builds properly for me once again.

kahowell:
I made another pass through. Please remove all references to the following because they are pulled in transitively by other dependencies:
- [ ] org.apache.james:apache-mime4j-core
- [ ] org.apache.james:apache-mime4j-dom
- [ ] org.apache.james:apache-mime4j-storage
- [ ] io.smallrye:smallrye-config
- [ ] io.smallrye:smallrye-config-common
- [ ] org.reactivestreams:reactive-streams
- [ ] io.swagger:swagger-core
- [ ] io.swagger:swagger-models
- [ ] org.reflections:reflections
- [ ] org.objenesis:objenesis

Because we probably still want to control the versions of these (but can let the deps be transitively included, please remove the entries from ext.libraries for:
- [ ] com.fasterxml.jackson.dataformat:jackson-dataformat-yaml
- [ ] com.fasterxml.jackson.dataformat:jackson-dataformat-xml
- [ ] org.jboss.resteasy:resteasy-core
- [ ] org.slf4j:slf4j-api
- [ ] ch.qos.logback:logback-core

Please remove the following; they appear to be no longer needed:
- [ ] javax.mail:mail
- [ ] javax.json.bind:javax.json.bind-api
- [ ] javax.annotation:javax.annotation-api

Note we only need one of `org.jboss.spec.javax.ws.rs:jboss-jaxrs-api_2.1_spec` and `javax.ws.rs:javax.ws.rs-api`; please:
- [ ] Replace `org.jboss.spec.javax.ws.rs:jboss-jaxrs-api_2.1_spec` with `javax.ws.rs:javax.ws.rs-api` in `api` project.
- [ ] Remove remaining references to `org.jboss.spec.javax.ws.rs:jboss-jaxrs-api_2.1_spec`

Also please:
- [ ] Remove the two bouncycastle deps as testImplementation, as they are actually dependencies for keycloak (can just remove the two lines and the comment).

To verify both before and after, I examined the war files generated by `mvn clean package` are similar enough (unsurprisingly, turning back on transitive deps added some more deps, but I'm not concerned). I used a diff of the output of `unzip -l server/target/candlepin-3.1.6.war | awk '{print $4}' | sort` to verify. As long as the spec tests are fine, I'm not concerned.

For what it's worth, the same process can be followed for gradle (`./gradlew clean assemble; unzip -l build/libs/candlepin-3.1.6.war | awk '{print $4}' | sort`). I did so and saw similar changes to the maven build.

abhiskum:
@kahowell Updated PR as per your last comment.  Please review. 

Also, please share process/steps find transitive dependencies. I tried `./gradlew candlepin:dependencies` however, I did not get these details. I must be missing something in my setup.

Ceiu:
@nikosmoum 
When I run the tests locally, I also get failures in the import tests on mysql, but they're not related to any class mismatches or anything. Definitely something going sideways here, though. Maybe try clearing out the gradle cache to pull down the new transient deps?

nikosmoum:
@abhiskum @Ceiu @kahowell I've looked into this a bit more. I've cleared my gradle cache and rebuild it for both this temp branch, and master branch and saw that:
- In master, I have only one source of the NewProxyPreparedStatement class: com.mchange/c3p0-0.9.5.4.jar
- In this branch, I have two sources of it: the same c3p0-0.9.5.4 (same as above), but also from c3p0-0.9.1.1.

I believe I'm hitting this: https://stackoverflow.com/questions/14771399/hibernate-4-calls-setbinarystream-that-is-not-supported-in-c3p0
because the 0.9.1.1 version is being used, which has this method signature only:
`setBinaryStream(int a, InputStream b, int c)` and this one which exists in 0.9.5.4: `setBinaryStream(int a, InputStream b, long c)`. So something is smuggling in that old c3p0-0.9.1.1 jar.

If we're moving to transitive dependencies, we better have a way of finding out how/why a dependency got in, so a gradle tool to print the dependency tree is needed I would say (but I don't have the gradle experience of how this can be accomplished).

Ceiu:
@nikosmoum 
That's odd. I would expect that the resolution of dependencies would not pull different versions of the same thing, unless they were sufficiently isolated from the application's classloader. If what you're hitting is consistent, that means the result we get is going to be based on which version of the class is found by the classloader first, which likely means it's based on very specific details about where the files live on the filesystem. In short, not something we ever want to deal with.

I fully agree with the request for a command or tool that outputs a nice way of looking through the deps. A simple tree-like printout would be sufficient and would make debugging this process very easy.

nikosmoum:
It's definitely consistent. Adding the following line in `CandlepinContextListener.contextInitialized`: `log.info(NewProxyPreparedStatement.class.getResource("/" + NewProxyPreparedStatement.class.getName().replace('.', '/') +
        ".class").toString());`, in master it shows that the class is loaded from the 0.9.5.4 jar, while for this branch it's loaded from the 0.9.1.1 jar.

nikosmoum:
@Ceiu @abhiskum OK, after running the gradle candlepin:dependencies task, I can see c3p0 specified thrice. Once explicitly at the top level (0.9.5.4), then by hibernate-c3p0 (0.9.5.3, but gradle auto-resolves that by using the latest that is specified elsewhere, which is 0.9.5.4), and then by quarts.

The problem is that the one pulled from quartz is so old that gradle can't auto-resolve it, because the group name has changed (used to be `c3p0`, now its `com.mchange`), so gradle thinks it's a different dependency.
```
+--- org.hibernate:hibernate-c3p0 -> 5.4.9.Final
|    +--- org.jboss.logging:jboss-logging:3.3.2.Final -> 3.4.0.Final
|    +--- org.hibernate:hibernate-core:5.4.9.Final
|    |    ...
|    \--- com.mchange:c3p0:0.9.5.3 -> 0.9.5.4
|         \--- com.mchange:mchange-commons-java:0.2.15
+--- com.mchange:c3p0 -> 0.9.5.4 (*)
...
+--- org.quartz-scheduler:quartz -> 2.2.2
|    +--- c3p0:c3p0:0.9.1.1
|    \--- org.slf4j:slf4j-api:1.7.7 -> 1.7.25
...

```

abhiskum:
Thank you @nikosmoum for finding root cause. After upgrading Quartz to 2.3.2, spec tests are passing in my local. 

@barnabycourt @nikosmoum There is a PR https://github.com/candlepin/candlepin/pull/2631 for upgrading Quartz to 2.3.2. 

Should I wait for https://github.com/candlepin/candlepin/pull/2631 PR to merge or close https://github.com/candlepin/candlepin/pull/2631  and upgrade Quartz in this PR itself?

abhiskum:
@kahowell Please review.

kahowell:
> I can't really comment specifically on the gradle stuff, but it looks like that now all children pom.xml files (that are autogenerated) (server/common/api/checks) contain an identical <dependencyManagement> tag that contains what seems to me a superset of all dependencies we have. Is this intended / supposed to be like this? It seems a bit redundant.

Since these poms are generated, it's fine. Having extra pinning in a dependencyManagement section doesn't cause any harm, and this way we are managing the deps globally a single time in `build.gradle`.

server/build.gradle
```diff
@@ -160,98 +159,46 @@ dependencies {
...
 
     testImplementation libraries.core_testing
     testImplementation libraries.liquibase_slf4j
     testImplementation libraries.jmock
 
     // Only include bouncycastle for testing purposes (compare against jss)
-    testImplementation "org.bouncycastle:bcpkix-jdk15on:1.60"
-    testImplementation "org.bouncycastle:bcprov-jdk15on:1.60"
+    testImplementation "org.bouncycastle:bcpkix-jdk15on"
+    testImplementation "org.bouncycastle:bcprov-jdk15on"
```
kahowell: Please remove these lines, otherwise changes look good to me.

abhiskum: @kahowell Done.

PR #2613, opened 2020-02-14
[F] Port CertificateDTO to  openapi spec (ENT-1854)

abhiskum:
- Added CertificateDTO schema in spec file
  - Updated code to use generated CertificateDTO
  - Removed existing CertificateDTO and CertificateDTOTest
  - Removed CertificateDTO.clone method calls from CdnDTO, UpstreamConsumerDTO, ConsumerDTO and poolDTO


Ceiu:
ack

Comment for other reviewers:
The straight removal for clone does technically break things, but (a) it only breaks clone, which is unused, (b) those DTOs which are going to be replaced in this feature branch, and (c) because this is a feature branch, we can tolerate temporary breakages like this.

PR #2614, opened 2020-02-14
[M] ENT-2107: Make jobs run in parallel

nikosmoum:
- Currently, jobs cannot run in parallel because we synchronize on
  the JobManager singleton object on the execute method.
  Once an artemis thread starts running a job, no other job can run
  in parallel because the other artemis threads are waiting for the
  lock. Not only that, but we also can't queue jobs at the same
  time that a job is running, because the queue method is also
  synchronized. This fix removes synchronization from these two
  methods (and a couple more that don't need it).
- Added synchronization on ArtemisProducer & ArtemisSession to avoid
  an artemis client warning when we try to queue jobs (AMQ212051:
  Invalid concurrent session usage. Sessions are not supposed to be
  used by more than one thread concurrently). This appeared once
  the queueJob method stopped being synchronized.

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -982,8 +982,7 @@ public class JobManager implements ModeChangeListener {
      *  an AsyncJobStatus instance representing the queued job's status, or the status of the
      *  existing job if it already exists
      */
-    public synchronized AsyncJobStatus queueJob(JobConfig config) throws JobException {
-
+    public AsyncJobStatus queueJob(JobConfig config) throws JobException {
         ManagerState state = this.getManagerState();
         if (state != ManagerState.RUNNING) {
             // Check if we're paused. If so, and if the "queue while paused" config is not set,
@@ -1177,10 +1176,9 @@ public class JobManager implements ModeChangeListener {
      * @return
      *  a JobStatus instance representing the job's status
      */
-    public synchronized AsyncJobStatus executeJob(JobMessage message) throws JobException {
-
+    public AsyncJobStatus executeJob(JobMessage message) throws JobException {
         ManagerState state = this.getManagerState();
-        if (this.getManagerState() != ManagerState.RUNNING) {
+        if (state != ManagerState.RUNNING) {
             String msg = String.format("Jobs cannot be executed while the manager is in the %s state", state);
             throw new IllegalStateException(msg);
         }
@@ -1602,7 +1600,7 @@ public class JobManager implements ModeChangeListener {
      *  could be found
      */
     @Transactional
-    public synchronized AsyncJobStatus cancelJob(String jobId) {
+    public AsyncJobStatus cancelJob(String jobId) {
         if (jobId == null || jobId.isEmpty()) {
             throw new IllegalArgumentException("jobId is null or empty");
         }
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -982,19 +982,20 @@ public AsyncJobStatus findJob(String jobId) {
...
-            // throw our usual ISE
-            if (state != ManagerState.SUSPENDED ||
-                !this.configuration.getBoolean(ConfigProperties.ASYNC_JOBS_QUEUE_WHILE_SUSPENDED)) {
-
-                String msg = String.format("Jobs cannot be queued while the manager is in the %s state",
-                    state);
-
-                throw new IllegalStateException(msg);
+    public AsyncJobStatus queueJob(JobConfig config) throws JobException {
+        synchronized (this) {
```
abhiskum: @nikosmoum 
If the same JobManager instance shared with multiple Artemis threads and the intention is to run `queueJob` and `executeJob` methods in parallel then `synchronized (this)` will not help here as both blocks are synchronized over `this`. 

I would recommend using 2 different locks in these methods. You can try using 2 different `ReentrantLock` here.  

Also, I did not get how the synchronized block will help to run jobs in parallel.

nikosmoum: @Ceiu The `synchronized (this)` block does not span the whole queueJob and executeJob methods. That would be pointless because it would be the same as leaving the synchronized keyword on the method signature itself. It only synchronizes the little bit of code that checks the manager state. when they get to the `result = job.execute(status);` bit, there is no synchronization. But now I realize that the `getManagerState()` itself is synchronized, so there is no point in even doing that.

@abhiskum I am aware that they will be synchronized over the same object, and it does make sense in this case. Instead of synchronizing the whole method, only this small bit that checks the manager state is now synchronized (and we want it to be synchronized over the same object, because we're checking the state of the JobManager, which is a Singleton), not the whole of execute and queue methods. But as I said before, I realize this is pointless because that method we're calling is already synchronized, so I'll remove this.

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
```diff
@@ -138,7 +138,7 @@ private synchronized CPMProducer getProducer() throws CPMException {
      * @throws JobMessageDispatchException
      *  if the message cannot be posted for any reason
      */
-    public void postJobMessage(JobMessage jobMessage) throws JobMessageDispatchException {
+    public synchronized void postJobMessage(JobMessage jobMessage) throws JobMessageDispatchException {
```
Ceiu: Since this is a restriction of Artemis, and not something we care about at this level, we should try adding the synchronized flag onto the Artemis implementations of the messaging system. In this case, ArtemisProducer.send, ArtemisSession.commit and .rollback (probably .close, too)

server/src/main/java/org/candlepin/async/JobManager.java
abhiskum: Thank you, @nikosmoum for details.

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
nikosmoum: Done

PR #2618, opened 2020-02-18
Adjust pom generation to handle spec-first changes

kahowell:
Adds a pom task to the api sub-project, which does the following:

- uses openapi generator maven plugin to generate code
- uses dependencyManagement gradle plugin to include dependencyManagement in the pom.

Applies the spec version plugin to the API sub-project, so that the API version ends up with a proper version.

Updates generated pom.xml files.

To test:

```
./gradlew pom; mvn package
```

Without this change, the above fails; with it, it succeeds.

NOTE: While reviewing #2612, I encountered issues trying to verify that Maven builds are equivalent, and I found the need for this PR.


build.gradle
```diff
@@ -262,6 +263,55 @@ project(":api") {
...
+                        version "2.0"
+                        relativePath ".."
+                    }
+                    packaging "jar"
+                    build([:]) {
+                        plugins {
+                            plugin {
+                                groupId 'org.openapitools'
+                                artifactId 'openapi-generator-maven-plugin'
+                                version '4.0.2'
```
barnabycourt: Current master is using version 4.2.3

kahowell: Ack, addressed by specifying a var in the buildscript section for openapi_version, and then referencing in buildscript and pom generation section. Also, rebased against master.

PR #2627, opened 2020-02-28
ENT-2060: Handle Set in openapi generated DTOs

Januson:
- Override java codegen to generate set as a workaround until Openapi merges our contribution to support sets


Ceiu:
This looks like we're modifying a version of their class and including it correctly? If so, can you point out the bits that are added by us?

Januson:
@Ceiu @nikosmoum I separated the changes into two commits so that our modifications are visible.

PR #2630, opened 2020-03-02
[F] ENT-1846: Port ActivationKeyDTO to openapi spec

nikosmoum:
- Replace use of current ReleaseVersionWrapSerializer and
  ReleaseVersionWrapDeserializer on ActivationKeyDTO
  with a new ReleaseVerDTO and ReleaseVersionWrapDeserializer.
- Replace use of current ProductWrapSerializer and
  ProductWrapDeserializer on ActivationKeyDTO with a new
  ActivationKeyProductDTO.
- Replace use of current OwnerDTO which was serialized with the
  assistance of Hateoas annotations when being nested, with a
  cut-down version called NestedOwnerDTO, to be used by all
  DTOs that have reference a nested OwnerDTO, eliminating the
  need of hateoas filters.
- Add non-null and non-empty field restrictions
  (required, minLength) to ActivationKeyDTO and
  existing ContentOverrideDTO and StatusDTO specs
  for validation.
- Introduce DTOValidator utility class to be used on Resource
  endpoints for performing validation that previously happened
  on DTO setters/constructors.


nikosmoum:
Note: This is marked in progress because the original ActivationKeyDTO contains `Set` fields, so it will need to be altered & rebased for that once https://github.com/candlepin/candlepin/pull/2627 is merged.

nikosmoum:
Rebased and updated to use HashSets.

api/candlepin-api-spec.yaml
```diff
@@ -284,6 +284,86 @@ components:
...
+    ActivationKeyDTO:
+      description: A token that allows registration without username/password and sets subscription and content related properties at registration.
+      allOf:
+        - $ref: '#/components/schemas/TimestampedEntity'
+        - type: object
+          properties:
+            id:
+              type: string
+            name:
+              type: string
```
Januson: Just an idea. We could add pattern: "^[a-zA-Z0-9_-]+$" so that we don't have to validate the name manually.

Ceiu: Does the pattern method basically add a regex validator on this field? Does it also handle nulls?

api/candlepin-api-spec.yaml
```diff
@@ -284,6 +284,86 @@ components:
...
+        quantity:
+          type: integer
+          format: int64
+
+    ReleaseVerDTO:
+      properties:
+        releaseVer:
+          type: string
+
+    NestedOwnerDTO:
```
Ceiu: I assume this is the replacement for owner + hateoas filtering?

nikosmoum: Correct. Other DTOs that reference OwnerDTO will be using this

api/candlepin-api-spec.yaml
nikosmoum: I just realized that the resource is also doing this:
```
        if (StringUtils.isBlank(dto.getName())) {
            throw new BadRequestException(i18n.tr("Must provide a name for activation key."));
        }
```
so I can simply add `name` to the list of required properties (which will generate a `@NotNull` annotation) and will validate automatically

server/src/main/java/org/candlepin/jackson/ReleaseVersionWrapDeserializer.java
```diff
@@ -0,0 +1,84 @@
...
+
+            return parseValueNode(node);
+        }
+        else if (node.isObject()) {
+            log.debug("Processing {} as a containing object node.", fieldName);
+
+            TreeNode valueNode = node.path(fieldName);
+            if (valueNode.isMissingNode()) {
+                throw new CandlepinJsonProcessingException(
+                        "The field " + fieldName + " is missing from: " + node.asToken(),
```
Ceiu: Indentation

server/src/main/java/org/candlepin/resource/validation/DTOValidator.java
```diff
@@ -0,0 +1,239 @@
...
+        if (isCollectionInvalid(collection1) ||
+            isCollectionInvalid(collection2)) {
+            throw new IllegalArgumentException("collection contains null elements");
+        }
+    }
+
+    /**
+     * Works the same as {@link #validateCollectionElementsNotNull(Supplier, Supplier, Supplier, Supplier...)}
+     */
+    public void validateCollectionElementsNotNull(Supplier<? extends Collection<?>> getter1,
```
Ceiu: I'm mostly alright with the microoptimization here around the avoidance of the implicit array creation, however, I feel these should be structured with code minimization in mind -- especially since we're offloading the work to a worker method to begin with.

Anyway, since we have a private method which validates a collection, and the single-param version does nothing more than unpack the collection from its supplier, it feels like the cleanest way to do all of this is to declare the single-param version as the base, and then have the various variadic overloads call it with each supplier. For example:

```
    public void validateCollectionElementsNotNull(Supplier<? extends Collection<?>> getter) {
        Collection<?> collection = getter.get();
        if (collection == null || collection.stream().anyMatch(Objects::isNull)) {
            throw new IllegalArgumentException("collection contains null elements");
        }
    }

    public void validateCollectionElementsNotNull(Supplier<? extends Collection<?>> getter1,
        Supplier<? extends Collection<?>> getter2) {

        this.validateCollectionElementsNonNull(getter1);
        this.validateCollectionElementsNonNull(getter2);
    }

    // and so on
```

server/src/main/java/org/candlepin/resource/validation/DTOValidator.java
```diff
@@ -0,0 +1,239 @@
...
+     * object that happens during the vararg method call. There are currently no DTOs that contain more than
+     * three maps that require null element validation, so this optimisation would cover 100% of cases
+     * for now, while allowing us to validate more than 3 maps at a time in the future.
+     * ======================================================================================================
+     */
+
+    /**
+     * Works the same as {@link #validateMapElementsNotNull(Supplier, Supplier, Supplier, Supplier...)}
+     */
+    public void validateMapElementsNotNull(Supplier<? extends Map<?, ?>> getter) {
```
Ceiu: Same deal here: use this version as the base, and have the extended versions just call into it with each arg.

api/candlepin-api-spec.yaml
Januson: Exactly. It just adds @Pattern annotation. @Pattern is not validated if the value is null.

server/src/main/java/org/candlepin/resource/validation/DTOValidator.java
```diff
@@ -0,0 +1,239 @@
...
+        if (isMapInvalid(map1) ||
+            isMapInvalid(map2) ||
+            isMapInvalid(map3)) {
+            throw new IllegalArgumentException("map contains null elements");
+        }
+    }
+
+    private boolean isMapInvalid(Map<?, ?> map) {
+        return map != null &&
+            (map.values().stream().anyMatch(Objects::isNull) ||
```
Januson: Map has methods `map.containsKey(null)` and `map.containsValue(null)`. We could use those.

server/src/test/java/org/candlepin/resource/validation/DTOValidatorTest.java
```diff
@@ -0,0 +1,350 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+
+/**
+ * Test suite for the DTOValidator class
+ */
+public class DTOValidatorTest extends DatabaseTestFixture {
+
+    @Inject DTOValidator validator;
```
Januson: We could use `Validation.buildDefaultValidatorFactory()` instead of extending the whole DB fixture.

server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyTranslator.java
```diff
@@ -101,27 +124,37 @@ public ActivationKeyDTO populate(ModelTranslator modelTranslator,
         }
         dest.setAddOns(addOns);
 
-        // Process nested DTO objects if we have a model translator to use to the translation...
-        if (modelTranslator != null) {
-            dest.setOwner(modelTranslator.translate(source.getOwner(), OwnerDTO.class));
+        Owner owner = source.getOwner();
+        if (source.getOwner() != null) {
+            NestedOwnerDTO ownerDTO = new NestedOwnerDTO()
```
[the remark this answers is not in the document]
nikosmoum: Ah, I forgot to use the translator here because initially wrote this as if we might not need it.

server/src/main/java/org/candlepin/resource/validation/DTOValidator.java
nikosmoum: Good idea

server/src/main/java/org/candlepin/resource/validation/DTOValidator.java
nikosmoum: Apparently, some implementations of `Map` such as `ConcurrentHashMap` will throw an NPE if you pass null to `containsValue`, so I'll keep this as is

server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyTranslator.java
```diff
@@ -83,16 +98,24 @@ public ActivationKeyDTO populate(ModelTranslator modelTranslator,
                     productIds.add(prod.getId());
                 }
             }
-
-            dest.setProductIds(productIds);
+            Set<ActivationKeyProductDTO> productIdObjects = productIds.stream().map(productId ->  {
+                ActivationKeyProductDTO newProduct =
+                    new ActivationKeyProductDTO();
```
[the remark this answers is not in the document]
nikosmoum: Done

server/src/main/java/org/candlepin/dto/api/v1/ActivationKeyTranslator.java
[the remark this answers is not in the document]
nikosmoum: Done

server/src/main/java/org/candlepin/jackson/ReleaseVersionWrapDeserializer.java
nikosmoum: Done

server/src/main/java/org/candlepin/resource/validation/DTOValidator.java
nikosmoum: Done

server/src/main/java/org/candlepin/resource/validation/DTOValidator.java
nikosmoum: Done

server/src/test/java/org/candlepin/resource/validation/DTOValidatorTest.java
nikosmoum: Done. I had to add the javax.el library, but it saves us ~8 seconds of execution

api/candlepin-api-spec.yaml
nikosmoum: I decided not to do this because it will change the current error message we send back to [this one](https://github.com/candlepin/candlepin/blob/32258d638e79e90eba5cf4cb5b3efc12354ca24b/common/src/main/java/org/candlepin/common/validation/CandlepinMessageInterpolator.java#L69), which is less user friendly.

PR #2631, opened 2020-03-03
Update Quartz to 2.3.2

barnabycourt:


barnabycourt:
@nikosmoum Thanks for the reminder about updating the server pom.

barnabycourt:
Closing in favor of #2612 as it includes this change and we can avoid forcing another rebase in that PR.

PR #2638, opened 2020-03-11
Fix import for set param

Januson:


abhiskum:
@nikosmoum This fix is working for method arguments. But as per your comment, it seems still has an issue with the return type.

nikosmoum:
Looks like there's a merge conflict needs resolving. Also, there's no need to include the ActivationKeyDTO commit in this PR, since that has already been merged in the feature/spec_first branch

PR #2639, opened 2020-03-13
[F] ENT-2105: Apply dynamic property filtering to openapi DTOs

nikosmoum:
- Old DTOs now use the "LegacyDTOFilter" filter name for hateoas and
  dynamic property filtering.
- New DTOs (ported to openapi spec) now use the "DTOFilter" filter
  name for dynamic property filtering, by indirectly applying a
  @JsonFilter annotation to them using Jackson Mix-Ins.
- Stop adding the following obsolete filters to all ObjectMappers,
  and remove all applied annotations for them: PoolAttributeFilter,
  ProductAttributeFilter, ProvidedProductFilter.
- Stop adding the following obsolete filters to RulesObjectMapper
  and remove applied annotations for them on rules DTOs:
  ConsumerFilter, EntitlementFilter, OwnerFilter, GuestFilter.
  The rules DTOs are already directly filtered-down versions of
  their entity counterparts. Delete obsolete RulesObjectMapper unit
  tests.
- Remove irrelevant annotations from manifest DTOs: OwnerFilter,
  PoolFilter.
- Stop adding the EntitlementFilter to the SyncUtils ObjectMapper;
  instead, directly remove the consumer field from the manifest
  EntitlementDTO which was being filtered anyway.
- Add unit tests for openapi-generated DTO property filtering.
- Introduce GuestIdDTOArrayElement & related translator, which is
  a version of GuestIdDTO that does not include the attributes
  field, used in cases where the @HateoasArrayExclude annotation
  filtering was previously used.


server/src/test/java/org/candlepin/resteasy/JsonProviderTest.java
```diff
@@ -142,4 +148,203 @@ private boolean isEnabled(JsonProvider provider, SerializationFeature feature) {
...
+        JsonNode akNode = null;
+        try {
+            akNode = mapper.readTree(serializedKey);
+        }
+        catch (JsonProcessingException e) {
+            fail("Parsing serialized ActivationKeyDTO failed!");
+        }
+        assertEquals(1, akNode.size());
+        assertTrue(akNode.has("owner"), "The 'owner' field should have been included!");
+        assertTrue(akNode.get("owner").has("id"), "The 'owner.id' field should have been included!");
```
Ceiu: For these tests where we have nested nodes, we should also verify that the children nodes have the correct amount of elements.

Applies below as well

server/src/main/java/org/candlepin/resteasy/JsonProvider.java
```diff
@@ -107,6 +124,11 @@ private void configureHateoasObjectMapper(ObjectMapper mapper, boolean indentJso
         }
 
         SimpleFilterProvider filterProvider = new SimpleFilterProvider();
+        filterProvider = filterProvider.addFilter("DTOFilter", new DynamicPropertyFilter());
+
+        // TODO: Once all DTOs are ported to the openapi spec, stop adding the following
+        //  filters: OldDTOFilter, ConsumerFilter, EntitlementFilter, OwnerFilter, GuestFilter.
+        filterProvider = filterProvider.addFilter("OldDTOFilter", new DynamicPropertyFilter());
```
Ceiu: This is probably the lowest priority thing I've ever requested, but could we call this LegacyDTOFilter rather than Old? I don't know why, but "old" feels very rough to me.

api/candlepin-api-spec.yaml
```diff
@@ -515,6 +515,17 @@ components:
               additionalProperties:
                 type: string
 
+    ArrayExcludeGuestIdDTO:
```
Ceiu: The name seems odd to me, and probably more references the original HATEOAS usage rather than what the DTO itself *is*. I feel that something like AttributelessGuestIdDTO would be clearer going forward; assuming we also have a GuestIdDTO that *does* have the attributes associated with it.

nikosmoum: Yes we do have GuestIdDTO that does have the attributes, already in the spec file. The reason I just used the original hateoas usage was because we'll have a couple more DTOs that will need to have cut-down versions because they use that annotation (specifically, ConsumerDTO and UpstreamConsumerDTO, which I've updated the relevant jira cards to make sure they get created), and wanted to use a standard naming scheme for all. For example, ConsumerDTO has that annotation for the `idCert` and `facts` fields, and naming it `IdCertLessAndFactLessConsumerDTO` is not a good idea.

Maybe a better format for all 3 of them would be `ArrayElement*DTO` or `*DTOArrayElement`, because those DTOs are only going to be used as references in an array?

server/src/main/java/org/candlepin/resteasy/JsonProvider.java
nikosmoum: No problem, `LegacyDTOFilter` sounds good

api/candlepin-api-spec.yaml
Ceiu: The *DTOArrayElement for standardization sounds good to me.

nikosmoum:
@Ceiu 
- I removed consumer from manifest EntitlementDTO, because the objectmapper used by export was configured to [filter it](https://github.com/candlepin/candlepin/blob/bac6a37df96ebcd4a962103877a348c51dbd53a2/server/src/main/java/org/candlepin/sync/SyncUtils.java#L84) out anyway, so there was no need to populated it on the DTO in the first place (so both the field and the filtering were removed).
- About jackson mix-ins: I'll update the class to an interface because it is actually better practice, but it does not make any functional difference. Jackson isn't compiling them together, it's simply reading the Jackson annotations added on the mix-in class/interface and applying them on the target class for purposes of de/serialization.

server/src/test/java/org/candlepin/resteasy/JsonProviderTest.java
nikosmoum: Done

server/src/main/java/org/candlepin/resteasy/JsonProvider.java
nikosmoum: Done

api/candlepin-api-spec.yaml
nikosmoum: Done

Ceiu:
ack

PR #2644, opened 2020-03-19
1692411: Added additional product ids into consumer installed products (ENT-1285)

abhiskum:
Current rule code is only to consider consumer installed products to calculate the pool's average priority. Hence added any additional products into consumer’s installed products to correctly calculate pool average priority.


abhiskum:
@nikosmoum  Added bug id at the start of the commit message.

PR #2647, opened 2020-03-20
ENT-1851: Port CdnDTO to openapi spec

wolfdale:
Porting CdnDTO to openapi spec
- Added CdnDTO schema in spec file.
- Removed existing CdnDTO and CdnDTOTest files.
- Updated the populate method of CDN translator.
- Added the mixing for CdnDTO to enable filtering.


api/candlepin-api-spec.yaml
```diff
@@ -428,6 +428,23 @@ components:
...
               type: string
 
+    CdnDTO:
+      description: DTO representing a Cdn
+      allOf:
+        - $ref: '#/components/schemas/TimestampedEntity'
+        - type: object
+          properties:
+            id:
+              type: string
```
nikosmoum: Because the current CdnDTO has @JsonProperty/@JsonIgnore set to getId/setId respectively, we should also add the `readOnly: true` property to this field. Currently it doesn't seem like it does anything at all, but I'm investigating if anything can be done.

nikosmoum: Alright, since we're dropping the use of @JsonIgnore entirely, lets just leave this as is

wolfdale:
Right. Added the mixin for the CdnDTO for filtering.

PR #2655, opened 2020-03-20
[M] Add workflow for auto-assign reviewer action

nikosmoum:


nikosmoum:
Note: This is adding a github "workflow" on the candlepin repository, which runs a github "action" I wrote: https://github.com/nikosmoum/auto-assign-reviewer-team 

The action is auto-assigning the specified team. In this case, our workflow passes in the `candlepin-server` team name. 

At that point, the workflow/action is finished. The setting we've set on the repository to automatically assign one of the members of the `candlepin-server` team, is removing the team and adding one of the team members as a reviewer in a round-robin fashion.

Ceiu:
I see nothing horrible here; though I'm not a huge fan of adding stuff like this to the mainline repo.

I did have two questions though:
- The flow on this PR looks like it requested the team, and then me, and then unrequested the team. Is that "normal" for this, something that will be sorted out later, or a one-off due to how it's not part of the repo yet?
- I assume the node.js stuff from your repo will be running on the github side of things and will never actually leave your repo and find its way here in any capacity. Is that correct?

nikosmoum:
@Ceiu 
- The flow is going to stay the same. I believe this is how the team auto-assignment you set up before [works](https://help.github.com/en/github/setting-up-and-managing-organizations-and-teams/managing-code-review-assignment-for-your-team).
- The nodejs stuff on my repo can be whenever we want them. I put them in my repo because It seemed the easiest/fastest way to test it is working because most examples online do it this way (standalone github action in a repo, published in the github marketplace). If we wanted, we could either 1) create the same repo under the candlepin org and publish it on the github marketplace from there, or 2) put the whole code within this repository (under .github/actions/), so that both the workflow and action are in one place. I can do either if you feel that's better

Ceiu:
I'd prefer to *not* have the node.js stuff make its way here, so it living in your repo is fine. I just wanted to understand the long-term intent for it.

PR #2661, opened 2020-03-23
[F] Fix oauth spec test

nikosmoum:
oauth spec tests are failing with the newly created vagrant box (destroy and up). As per my investigation, either we need to pass only #{uri} or URL object instead of #{site}#{uri} in Net::HTTP::Get.new.


Ceiu:
ack

Ceiu:
Looks like Jenkins had a bit of a network hiccup:

```
Unable to update GitHub status (commit: 829fb303eeedef8017be14a16b73e328911fe010, status: SUCCESS, context: jenkins-pipeline)
Finished: SUCCESS
```

Merging without rerunning.

PR #2663, opened 2020-03-24
[F] ENT-2177: Use concrete method to locate ResourceLocator

abhiskum:
- Added new class MethodLocator to hold interface method and concrete method mapping
  - Updated ResourceLocatorMap class to use MethodLocator
  - Update AnnotationLocator class to use MethodLocator instead of internal

### DIFF
--- a/server/src/main/java/org/candlepin/resteasy/MethodLocator.java
+++ b/server/src/main/java/org/candlepin/resteasy/MethodLocator.java
@@ -1,0 +1,129 @@
+/**
+ * Copyright (c) 2009 - 2020 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.resteasy;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.inject.Binding;
+import com.google.inject.Injector;
+
+import org.jboss.resteasy.spi.metadata.ResourceBuilder;
+import org.jboss.resteasy.spi.metadata.ResourceClass;
+import org.jboss.resteasy.spi.metadata.ResourceLocator;
+import org.jboss.resteasy.spi.metadata.ResourceMethod;
+import org.jboss.resteasy.util.GetRestful;
+import org.slf4j.Logger;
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/candlepin/resteasy/MethodLocator.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Inject
+    public MethodLocator(Injector injector) {
+        // Maintain the insertion order for nice output in debug statement
+        internalMap = new LinkedHashMap<>();
+        this.injector = injector;
+    }
+
+    @SuppressWarnings("rawtypes")
+    public synchronized void init() {
```
Ceiu: It's not strictly critical to change for this PR, but in the future I would like to see these type of methods return a self-ref so we can do some chaining in the cases where our flow is to just instantiate and then initialize.

server/src/test/java/org/candlepin/test/DatabaseTestFixture.java
```diff
@@ -211,8 +213,9 @@ public void init(boolean beginTransaction) throws Exception {
         locatorMap = this.injector.getInstance(ResourceLocatorMap.class);
```
Ceiu: This ResourceLocatorMap instantiation is incorrect with the method locator being instantiated as it is below.

abhiskum: Fixed.

server/src/main/java/org/candlepin/resteasy/MethodLocator.java
abhiskum: Done.

server/src/test/java/org/candlepin/test/DatabaseTestFixture.java
abhiskum: Fixed.

abhiskum: Done.

server/src/main/java/org/candlepin/resteasy/ResourceLocatorMap.java
```diff
@@ -284,7 +290,7 @@ protected void logLocators() {
         }
     }
 
-    protected void registerLocators(ResourceClass resourceClass) {
+    protected void registerLocators(ResourceClass resourceClass, Class<?> concreteClass) {
```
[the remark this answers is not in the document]
abhiskum: @nikosmoum Initially I added method mapping in this class and then moved to separate class however forgot to remove this change.

abhiskum:
@nikosmoum Updated commit message.

server/src/test/java/org/candlepin/test/DatabaseTestFixture.java
```diff
@@ -157,6 +158,7 @@
     @Inject protected PermissionFactory permissionFactory;
 
     @Inject protected ResourceLocatorMap locatorMap;
+    @Inject protected MethodLocator methodLocator;
```
Ceiu: The changes below are good, but mean that these should not be injected (since we create them ourselves during ```init```).

abhiskum: Done.

PR #2665, opened 2020-03-24
[F] Add back user id that was accidentally removed

nikosmoum:


server/src/main/java/org/candlepin/dto/api/v1/UserInfoTranslator.java
```diff
@@ -62,14 +62,16 @@ public UserDTO populate(ModelTranslator translator, UserInfo source, UserDTO des
         if (dest == null) {
             throw new IllegalArgumentException("dest is null");
         }
+       // We don't have an ID from the adapters, so we'll just null it out.
```
Ceiu: Extraneous comment

Ceiu:
Looks good; just an extra comment to remove and this is okay with me. Going to leave it to Prakhar to merge.

As an aside, I want to make sure we're clear on the spec-first front that the *Info interfaces are not affecting the design of the spec-first DTOs. If we need to keep a separate set of DTOs for them, that's an issue we can address on a case-by-case basis.

nikosmoum:
@Ceiu @wolfdale I've also fixed the type of a field from the invalid 'long' to 'int64'

PR #2671, opened 2020-03-30
[M] ENT-2215: truncate correlation id to 40 characters

abhiskum:
- Truncate correlation id to 40 characters instead of dropping it.

### DIFF
--- a/common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
+++ b/common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
@@ -47,7 +47,7 @@ public class LoggingFilter implements Filter {
 
     public static final String CSID_KEY = "csid";
     private static final int CSID_MAX_LENGTH = 40;
-    private static final Pattern CSID_REGEX = Pattern.compile("^([a-zA-Z0-9-]){1," + CSID_MAX_LENGTH + "}$");
+    private static final Pattern CSID_REGEX = Pattern.compile("^([a-zA-Z0-9-]){1,}$");
 
     /** The metadata key used to display the owner's key or display name */
     public static final String OWNER_KEY = "org"; // This value must match that set in logback.xml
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
```diff
@@ -46,8 +46,7 @@
     private static Logger log = LoggerFactory.getLogger(LoggingFilter.class);
 
     public static final String CSID_KEY = "csid";
-    private static final int CSID_MAX_LENGTH = 40;
```
wottop: Please keep and use the constant.

common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
```diff
@@ -95,11 +94,11 @@ public void doFilter(ServletRequest request, ServletResponse response,
 
             if (correlationId != null) {
                 if (CSID_REGEX.matcher(correlationId).matches()) {
-                    MDC.put(CSID_KEY, correlationId);
+                    MDC.put(CSID_KEY,
```
Ceiu: I really don't feel truncating is the right approach here, but if that's what we're going to do, we should still output a log message indicating that we've truncated the original value down, and what that value was so we can at least track and figure out what happened in the event that it happens or when there's a collision caused by it.

common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
abhiskum: Fixed.

common/src/main/java/org/candlepin/common/filter/LoggingFilter.java
abhiskum: Fixed.

wottop:
ack

PR #2673, opened 2020-04-07
[M] 1792278: use autoheal value during registration (ENT-2187)

abhiskum:
Leverage auto-heal value available in consumer during registration with auto-attach enable activation key.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/util/ConsumerBindUtil.java
+++ b/server/src/main/java/org/candlepin/resource/util/ConsumerBindUtil.java
@@ -40,6 +40,7 @@ import org.candlepin.version.CertVersionConflictException;
 
 import com.google.inject.Inject;
 
+import org.apache.commons.lang.BooleanUtils;
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/resource/util/ConsumerBindUtil.java
```diff
@@ -95,12 +97,18 @@ public void handleActivationKeys(Consumer consumer, List<ActivationKey> keys,
...
+                if (autoattachDisabledForOwner || key.getOwner().isContentAccessEnabled() || !isAutoheal) {
                     String caMessage = "";
                     if (key.getOwner().isContentAccessEnabled()) {
                         caMessage = " because of the content access mode setting";
                         isCAModeEnabledForAny = true;
                     }
+                    if (!isAutoheal) {
+                        caMessage = StringUtils.isBlank(caMessage) ?
+                            " because of the consumer autoheal value" :
+                            " and consumer autoheal value";
```
nikosmoum, line 109: I know this is a bit nit-picky, but good log messages help people figure out what's happened without having to attach a debugger. Using the same warning log output message and appending either of these two would result in misleading warning messages:

On the scenario where contentAccessEnabled is true, this is a bit misleading:
`Auto-attach is disabled for owner because of the content access mode setting and consumer autoheal value. Skipping...`
It should be more like:
`Auto-attach is disabled for owner because of the content access mode setting. Also, the consumer autoheal value is false. Skipping...`

On the scenario where contentAccessEnabled is false, this is very misleading:
`Auto-attach is disabled for owner because of the consumer autoheal value. Skipping...`
The consumer autoheal value has no effect on the owner's auto-attach setting, so auto-attach cannot be disabled for the owner. This should be more like:
`The consumer autoheal value is false. Skipping...`

abhiskum: @nikosmoum thanks for catching this, earlier it was absolutely incorrect. I have fixed it. Please re-review.

PR #2674, opened 2020-04-07
ENT-1886: Port SubscriptionResource to spec-first

Januson:
- Add SubscriptionResource definition
- Add missing DTO definitions
- Add translators
- Add AttributesDeserializer

Note: This PR needs to pull in the changes from ContentDTO PR. I pulled older version as ProductDto needed to use ContentDto.


server/src/main/java/org/candlepin/jackson/CandlepinAttributesDeserializer.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+
+/**
+ * The CandlepinAttributeDeserializer handles the deserialization of attributes, processing both
+ * the current attribute mapping style and obsoleted formats.
+ */
+public class CandlepinAttributesDeserializer extends StdDeserializer<AttributesDTO> {
```
Ceiu, line 38: Does this have value when we're explicitly defining the structure via objects (and inheriting oddities such as the ```.getAttributes().getAttributes()``` as above?)

server/src/main/java/org/candlepin/resource/CdnResource.java
```diff
@@ -51,7 +51,6 @@ public CdnResource(I18n i18n, CdnCurator curator, CdnManager manager, ModelTrans
         this.cdnManager = manager;
         this.translator = translator;
     }
-
```
Ceiu, line 54: Keep this whitespace

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -346,14 +351,36 @@ public ProductDTO addBatchContent(
...
+        content.setContent(dto);
+        content.setEnabled(enabled);
+
+        boolean changed = product.getProductContent()
+            .stream()
+            .filter(contentDTO -> contentDTO.getContent().getId().equals(content.getContent().getId()))
+            .noneMatch(contentDTO -> contentDTO.equals(content));
+
+        if (changed) {
+            product.getProductContent().add(content);
```
Ceiu, line 380: encapsulation

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -196,7 +201,7 @@ protected Content fetchContent(Owner owner, String contentId) {
         response = Product.class, responseContainer = "list")
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    public CandlepinQuery<ProductDTO> listProducts(
+    public CandlepinQuery<ProductLegacyDTO> listProducts(
```
Ceiu, line 204: We seem to be heavily using the "Legacy" product here and aren't using the base product at all, which implies it's not actually legacy.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -407,30 +434,44 @@ public ProductDTO removeBatchContent(
...
+        }
+        if (contentIds.isEmpty()) {
+            return false;
+        }
+
+        Set<ProductContentDTO> toRemove = product.getProductContent().stream()
+            .filter(content -> contentIds.contains(content.getContent().getId()))
+            .collect(Collectors.toSet());
+
+        product.getProductContent().removeAll(toRemove);
```
Ceiu, line 458: encapsulation

server/src/main/java/org/candlepin/util/Util.java
```diff
@@ -143,6 +147,13 @@ public static Date toDate(String dt) {
...
     }
 
+    public static Date toDate(OffsetDateTime dt) {
+        if (dt == null) {
+            return null;
+        }
+        return Date.from(dt.toInstant());
+    }
+
     public static <T> T assertNotNull(T value, String message) {
```
Ceiu, line 157: This is unrelated to your PR, but can we put a TODO on this to axe this entirely? I don't think anyone knows it exists, and it doesn't provide any real value other than saving two lines of code.

server/src/main/java/org/candlepin/util/Util.java
```diff
@@ -547,4 +558,37 @@ public static String getHostname() {
...
+    }
+
+    /**
+     * Takes a key value pair and returns a map with a single entry.
+     *
+     * @param key a key of the single entry
+     * @param value a value of the single entry
+     * @return a map with a single entry
+     */
+    public static Map<String, String> singletonMap(String key, String value) {
```
Ceiu, line 588: This method is already defined in ```java.util.Collections``` with the exact same name and arguments, but is generically typed.

api/candlepin-api-spec.yaml
```diff
@@ -1133,7 +1296,7 @@ components:
               type: boolean
 
     ContentDTO:
-      description: Represents the content data exposed to the API.
+      description: Represents the content data exposed to the API and adapter layers.
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1517,90 @@ components:
...
+      required:
+        - id
+      properties:
+        content:
+          $ref: "#/components/schemas/ContentDTO"
+        enabled:
+          type: boolean
+
+    ProductLegacyDTO:
+      description: Version and Status information about running Candlepin server
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1517,90 @@ components:
...
+              items:
+                $ref: "#/components/schemas/BrandingDTO"
+            href:
+              type: string
+              example: /products/ff808081554a3e4101554a3e9033005d
+            locked:
+              type: boolean
+
+    ProductDTO:
+      description: Version and Status information about running Candlepin server
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1517,90 @@ components:
...
+        - id
+      allOf:
+        - $ref: '#/components/schemas/ProductLegacyDTO'
+        - type: object
+          properties:
+            attributes:
+              $ref: '#/components/schemas/AttributesDTO'
+
+    AttributesDTO:
+      description: Version and Status information about running Candlepin server
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1517,90 @@ components:
             upstreamConsumer:
               $ref: '#/components/schemas/UpstreamConsumerDTO'
 
+    ProductContentDTO:
+      description: Product content mapping exposed to the API.
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1411,6 +1669,80 @@ components:
...
+        upstreamEntitlementId:
+          type: string
+        upstreamConsumerId:
+          type: string
+        cert:
+          $ref: "#/components/schemas/CertificateDTO"
+        cdn:
+          $ref: "#/components/schemas/CdnDTO"
+
+    SubscriptionDTO:
```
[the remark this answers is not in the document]

server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -318,18 +320,32 @@ public Content updateContent(ContentDTO update, Owner owner, boolean regenerateE
...
+    private void addContent(ProductDTO product, ContentDTO dto, boolean enabled) {
+        if (dto == null || dto.getId() == null) {
+            throw new IllegalArgumentException("dto references incomplete content");
+        }
+
+        ProductContentDTO content = new ProductContentDTO();
+        content.setContent(dto);
+        content.setEnabled(enabled);
+
+        product.getProductContent().add(content);
```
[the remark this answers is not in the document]
Januson: Done. Removed the original before insert.

server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -257,13 +257,15 @@ public Content updateContent(ContentDTO update, Owner owner, boolean regenerateE
                     log.debug("Updating affected product: {}", product);
                     ProductDTO pdto = this.modelTranslator.translate(product, ProductDTO.class);
 
-                    ProductContentDTO pcdto = pdto.getProductContent(cdto.getId());
-                    if (pcdto != null) {
-                        pdto.addContent(cdto, pcdto.isEnabled());
+                    pdto.getProductContent().stream()
```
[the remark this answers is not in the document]
Januson: You are right, that stream was a bit sloppy. I refactored it a bit.

server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -318,18 +320,32 @@ public Content updateContent(ContentDTO update, Owner owner, boolean regenerateE
             log.debug("Updating affected product: {}", product);
             ProductDTO pdto = this.modelTranslator.translate(product, ProductDTO.class);
 
-            ProductContentDTO pcdto = pdto.getProductContent(cdto.getId());
-            if (pcdto != null) {
-                pdto.addContent(cdto, pcdto.isEnabled());
+            pdto.getProductContent().stream()
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/controller/ProductManager.java
```diff
@@ -817,17 +823,17 @@ public static boolean isChangedBy(Product entity, ProductDTO dto) {
             };
 
             if (!Util.collectionsAreEqual((Collection) entity.getProductContent(),
-                (Collection) productContent, comparator)) {
+                (Collection) dto.getProductContent(), comparator)) {
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/controller/ProductManager.java
```diff
@@ -783,12 +786,15 @@ public static boolean isChangedBy(Product entity, ProductDTO dto) {
...
+        Map<String, String> attributes = null;
+        if (dto.getAttributes() != null) {
+            attributes = dto.getAttributes().getAttributes();
+        }
         if (attributes != null && !attributes.equals(entity.getAttributes())) {
             return true;
         }
 
-        Collection<ProductContentDTO> productContent = dto.getProductContent();
+        Set<ProductContentDTO> productContent = dto.getProductContent();
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/controller/ProductManager.java
```diff
@@ -817,17 +823,17 @@ public static boolean isChangedBy(Product entity, ProductDTO dto) {
...
             if (!Util.collectionsAreEqual((Collection) entity.getProductContent(),
-                (Collection) productContent, comparator)) {
+                (Collection) dto.getProductContent(), comparator)) {
 
                 return true;
             }
         }
 
-        Collection<BrandingDTO> brandingDTOs = dto.getBranding();
+        Collection<BrandingInfo> brandingDTOs = getBranding(dto);
```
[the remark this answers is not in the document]
Januson: Done. Replaced the use of adapter with a new comparator.

server/src/main/java/org/candlepin/dto/api/v1/ProductTranslator.java
```diff
@@ -104,11 +118,37 @@ public ProductDTO populate(ModelTranslator modelTranslator, Product source, Prod
...
+        }
+        return dto;
+    }
+
+    private void addBranding(ProductDTO product, BrandingDTO branding) {
+        if (isNullOrIncomplete(branding)) {
+            throw new IllegalArgumentException("branding is null or incomplete");
+        }
+
+        product.getBranding().add(branding);
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/dto/api/v1/ProductTranslator.java
```diff
@@ -61,41 +63,53 @@ public ProductDTO populate(Product source, ProductDTO destination) {
...
+            else {
+                destination.productContent(Collections.emptySet());
             }
 
             Collection<Branding> branding = source.getBranding();
             if (branding != null && !branding.isEmpty()) {
                 for (Branding brand : branding) {
                     if (brand != null) {
-                        destination.addBranding(modelTranslator.translate(brand, BrandingDTO.class));
+                        addBranding(destination, modelTranslator.translate(brand, BrandingDTO.class));
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/dto/shim/ProductDataTranslator.java
```diff
@@ -70,40 +74,43 @@ public ProductDTO populate(ModelTranslator modelTranslator, ProductData source,
...
+                dest.productContent(null);
+            }
 
             Collection<Branding> productBrandings = source.getBranding();
             dest.setBranding(null);
             if (productBrandings != null) {
                 for (Branding brand : productBrandings) {
                     if (brand != null) {
-                        dest.addBranding(modelTranslator.translate(brand, BrandingDTO.class));
+                        addBranding(dest, modelTranslator.translate(brand, BrandingDTO.class));
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/dto/shim/ProductDataTranslator.java
```diff
@@ -70,40 +74,43 @@ public ProductDTO populate(ModelTranslator modelTranslator, ProductData source,
...
 
-                for (ProductContentData pcd : productContentData) {
-                    if (pcd != null && pcd.getContent() != null) {
-                        ContentDTO dto = contentTranslator.translate(modelTranslator, pcd.getContent());
-                        dest.addContent(dto, pcd.isEnabled());
+                for (ProductContentData productContent : productContentData) {
+                    if (productContent != null && productContent.getContent() != null) {
+                        ContentDTO dto = contentTranslator
+                            .translate(modelTranslator, productContent.getContent());
+                        addContent(dest, dto, productContent.isEnabled());
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/dto/shim/ProductDataTranslator.java
```diff
@@ -114,4 +121,42 @@ public ProductDTO populate(ModelTranslator modelTranslator, ProductData source,
...
+    private void addContent(ProductLegacyDTO product, ContentDTO dto, boolean enabled) {
+        if (dto == null || dto.getId() == null) {
+            throw new IllegalArgumentException("dto references incomplete content");
+        }
+
+        ProductContentDTO content = new ProductContentDTO();
+        content.setContent(dto);
+        content.setEnabled(enabled);
+
+        product.getProductContent().add(content);
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/dto/shim/ProductDataTranslator.java
```diff
@@ -114,4 +121,42 @@ public ProductDTO populate(ModelTranslator modelTranslator, ProductData source,
...
+    private boolean addBranding(ProductLegacyDTO product, BrandingDTO branding) {
+        if (isNullOrIncomplete(branding)) {
+            throw new IllegalArgumentException("branding is null or incomplete");
+        }
+
+        if (product.getBranding() == null) {
+            product.setBranding(new HashSet<>());
+        }
+
+        return product.getBranding().add(branding);
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/hostedtest/BrandingInfoAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.hostedtest;
+
+import org.candlepin.dto.api.v1.BrandingDTO;
+import org.candlepin.service.model.BrandingInfo;
+
+public class BrandingInfoAdapter implements BrandingInfo {
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/resource/CdnResource.java
Januson: Done.

server/src/main/java/org/candlepin/util/Util.java
Januson: Done.

server/src/main/java/org/candlepin/util/Util.java
Januson: Done.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
Januson: We are using both. ProductLegacyDTO and CandlepinAttributesDeserializer are my attempt to preserve current functionallity of @JsonSerialize/Deserialize we use for attributes.

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1517,90 @@ components:
...
+              uniqueItems: true
+              items:
+                $ref: "#/components/schemas/BrandingDTO"
+            href:
+              type: string
+              example: /products/ff808081554a3e4101554a3e9033005d
+            locked:
+              type: boolean
+
+    ProductDTO:
```
nikosmoum, line 1578: I am very confused here. I assume this is the main ProductDTO, and that the ProductLegacyDTO is supposed to be the DTO for ProductData. If this is supposed to be the original ProductDTO, then it definitly should not be inheriting from the ProductLegacyDTO, and ProductLegacyDTO should definitely NOT be used throughout our codebase as a replacement for ProductDTO. ProductLegacyDTO should be used only for the member fields on the SubscriptionDTO (since this is a legacy dto structure and should go away eventually).

I feel we should not be blending/inheriting these two. We should do either of:
-  There should be a ProductDTO specified fully based on what our current ProductDTO looks like, and then a ProductLegacyDTO that is separate and fully based on what ProductData look like now. 
- If we chose to do inheritance (even though I think we shouldn't), then it would have to go the other way: Define ProductDTO based on what the current ProductDTO looks like, and then ProductLegacyDTO can inherit from that (if needed).
- Ideally, we could specify ONLY ProductDTO, which will be used for both purposes, since the structure is almost the same for our current ProductDTO and ProductData.

@Januson @Ceiu what do you think?

Also I'm not sure what's happening with the AttributesDTO, why is it added here, while there's already an attributes field on the legacyDTO.

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1517,90 @@ components:
...
+            multiplier:
+              type: integer
+              format: int64
+              example: 1
+            attributes:
+              type: array
+              items:
+                type: object
+                additionalProperties:
+                  type: string
```
nikosmoum, line 1556: Why does this not just use AttributesDTO instead?

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1517,90 @@ components:
             upstreamConsumer:
               $ref: '#/components/schemas/UpstreamConsumerDTO'
 
+    ProductContentDTO:
+      description: Product content mapping exposed to the API
+      required:
+        - id
```
nikosmoum, line 1523: this DTO does not have an id field, so the required section should be removed

server/src/main/java/org/candlepin/dto/api/v1/ProductContentTranslator.java
```diff
@@ -0,0 +1,74 @@
...
+package org.candlepin.dto.api.v1;
+
+import org.candlepin.dto.ModelTranslator;
+import org.candlepin.dto.ObjectTranslator;
+import org.candlepin.model.ProductContent;
+
+
+/**
+ * The ContentTranslator provides translation from Content model objects to
+ * ContentDTOs
```
nikosmoum, line 24: Wrong comment

server/src/main/java/org/candlepin/dto/api/v1/ProductTranslator.java
```diff
@@ -61,54 +64,94 @@ public ProductDTO populate(Product source, ProductDTO destination) {
...
 
             Collection<Branding> branding = source.getBranding();
             if (branding != null && !branding.isEmpty()) {
+                Set<BrandingDTO> dtos = new HashSet<>();
                 for (Branding brand : branding) {
                     if (brand != null) {
-                        destination.addBranding(modelTranslator.translate(brand, BrandingDTO.class));
+                        BrandingDTO dto = modelTranslator.translate(brand, BrandingDTO.class);
+                        if (isNullOrIncomplete(dto)) {
+                            throw new IllegalArgumentException("branding is null or incomplete");
```
nikosmoum, line 116: I don't think we should be throwing exceptions in translators. This kind of thing is for when we receive a DTO in a POST/PUT request and we have to validate the client data. Here, we're just translating the data we know is already valid (Branding) to be output back, so we don't need any validation

server/src/main/java/org/candlepin/dto/api/v1/SubscriptionTranslator.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+
+/**
+ * The RoleTranslator provides translation from Role model objects to
+ * RoleDTOs
```
nikosmoum, line 35: wrong comment

server/src/main/java/org/candlepin/dto/shim/ProductDataTranslator.java
```diff
@@ -70,42 +79,51 @@ public ProductDTO populate(ModelTranslator modelTranslator, ProductData source,
...
-            dest.setBranding(null);
-            if (productBrandings != null) {
+            if (productBrandings != null && !productBrandings.isEmpty()) {
+                Set<BrandingDTO> dtos = new HashSet<>();
                 for (Branding brand : productBrandings) {
                     if (brand != null) {
-                        dest.addBranding(modelTranslator.translate(brand, BrandingDTO.class));
+                        BrandingDTO dto = modelTranslator.translate(brand, BrandingDTO.class);
+                        if (isNullOrIncomplete(dto)) {
+                            throw new IllegalArgumentException("branding is null or incomplete");
```
nikosmoum, line 121: Same here: no need for validation at this point

server/src/main/java/org/candlepin/resource/SubscriptionResource.java
```diff
@@ -63,34 +65,37 @@
...
+        this.subService = Objects.requireNonNull(subService);
+        this.consumerCurator = Objects.requireNonNull(consumerCurator);
+        this.poolManager = Objects.requireNonNull(poolManager);
+        this.i18n = Objects.requireNonNull(i18n);
+        this.translator = Objects.requireNonNull(translator);
     }
 
     @ApiOperation(notes = "Retrieves a list of Subscriptions", value = "getSubscriptions")
     @GET
     @Produces(MediaType.APPLICATION_JSON)
```
nikosmoum, line 91: We're moving this resource to the spec file, which means all these annotations should be removed from all endpoints here (since they will be auto-generated in SubscriptionApi)

api/candlepin-api-spec.yaml
```diff
@@ -1070,21 +1229,25 @@ components:
...
+        - $ref: '#/components/schemas/TimestampedEntity'
+        - type: object
+          required:
+            - productId
+            - name
+            - type
+          properties:
+            productId:
+              type: string
+              maxLength: 255
```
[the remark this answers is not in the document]
Januson: I took the maxLength validation from Branding entity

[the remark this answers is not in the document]
nikosmoum: Ah, I see. Then we can leave the maxLength as they are and just add the minLength: 1

server/src/main/java/org/candlepin/util/Util.java
```diff
@@ -547,4 +550,24 @@ public static String getHostname() {
...
         return date != null ? date.toInstant().atOffset(ZoneOffset.UTC) : null;
     }
+
+    /**
+     * Takes a map and splits it into a list of maps each with one item from the given map.
+     *
+     * @param map a map to be split
+     * @return a list of single item maps
+     */
+    public static List<Map<String, String>> split(Map<String, String> map) {
```
nikosmoum, line 560: Where exactly is this needed? I can't find where this is used in your changes?

server/src/test/java/org/candlepin/dto/api/v1/ProductContentTranslatorTest.java
```diff
@@ -0,0 +1,109 @@
...
+        content.setName("test_value");
+        content.setVendor("test_value");
+        content.setContentUrl("test_value");
+        content.setRequiredTags("test_value");
+        content.setReleaseVersion("test_value");
+        content.setGpgUrl("test_value");
+        content.setMetadataExpiration(1234L);
+        content.setModifiedProductIds(Arrays.asList("1", "2", "3"));
+        content.setArches("test_value");
+        content.setLocked(Boolean.TRUE);
```
nikosmoum, line 64: Code duplication: This can be replaced by instantiating a `ContentTranslatorTest` and calling `initSourceObject()`

server/src/test/java/org/candlepin/dto/api/v1/ProductContentTranslatorTest.java
```diff
@@ -0,0 +1,109 @@
...
+                assertEquals(content.getLabel(), contentDto.getLabel());
+                assertEquals(content.getName(), contentDto.getName());
+                assertEquals(content.getVendor(), contentDto.getVendor());
+                assertEquals(content.getContentUrl(), contentDto.getContentUrl());
+                assertEquals(content.getRequiredTags(), contentDto.getRequiredTags());
+                assertEquals(content.getReleaseVersion(), contentDto.getReleaseVer());
+                assertEquals(content.getGpgUrl(), contentDto.getGpgUrl());
+                assertEquals(content.getMetadataExpiration(), contentDto.getMetadataExpire());
+                assertEquals(content.getModifiedProductIds(), contentDto.getModifiedProductIds());
+                assertEquals(content.getArches(), contentDto.getArches());
```
nikosmoum, line 102: Code duplication: This can be replaced by calling `ContentTranslatorTest`'s `verifyOutput` method

server/src/test/java/org/candlepin/dto/api/v1/SubscriptionTranslatorTest.java
```diff
@@ -0,0 +1,169 @@
...
+    @Override
+    protected Subscription initSourceObject() {
+        Subscription source = new Subscription();
+
+        source.setId("test_id");
+        source.setOwner(this.createOwner());
+        source.setProduct(this.createProduct());
+        source.setDerivedProduct(this.createProduct());
+        source.setProvidedProducts(this.createProducts());
+        source.setDerivedProvidedProducts(this.createProducts());
```
nikosmoum, line 71: There is no use for creating new util methods to create ProductData instances. You can create a `ProductDataTranslatorTest` instance and call `initSourceObject()` instead

server/src/test/java/org/candlepin/dto/api/v1/SubscriptionTranslatorTest.java
```diff
@@ -0,0 +1,169 @@
...
+        source.setEndDate(new Date());
+        source.setContractNumber("test_contact");
+        source.setAccountNumber("test_acc_num");
+        source.setModified(new Date());
+        source.setOrderNumber("test_order_num");
+        source.setUpstreamPoolId("test_pool_id");
+        source.setUpstreamEntitlementId("test_ent_id");
+        source.setUpstreamConsumerId("test_cons_id");
+        source.setCertificate(this.createCert());
+        source.setCdn(this.createCdn());
```
nikosmoum, line 83: These 2 can be replaced by using `CertificateTranslatorTest.initSourceObject()` and `CdnTranslatorTest.initSourceObject()`

server/src/test/java/org/candlepin/dto/api/v1/SubscriptionTranslatorTest.java
```diff
@@ -0,0 +1,169 @@
...
+    protected SubscriptionTranslator initObjectTranslator() {
+        return this.translator;
+    }
+
+    @Override
+    protected Subscription initSourceObject() {
+        Subscription source = new Subscription();
+
+        source.setId("test_id");
+        source.setOwner(this.createOwner());
```
nikosmoum, line 67: `NestedOwnerTranslatorTest.initSourceObject()`

server/src/test/java/org/candlepin/dto/api/v1/SubscriptionTranslatorTest.java
```diff
@@ -0,0 +1,169 @@
...
+            else {
+
+                assertTrue(isEmpty(dto.getProvidedProducts()));
+                assertTrue(isEmpty(dto.getDerivedProvidedProducts()));
+                assertNull(dto.getOwner());
+                assertNull(dto.getProduct());
+                assertNull(dto.getDerivedProduct());
+                assertNull(dto.getCert());
+                assertNull(dto.getCdn());
+            }
```
nikosmoum, line 145: The various `verifyOutput` methods from the *TranslatorTest.java instances can be used here instead

Januson: Done.

server/src/test/java/org/candlepin/dto/api/v1/SubscriptionTranslatorTest.java
Januson: Done.

server/src/test/java/org/candlepin/dto/api/v1/SubscriptionTranslatorTest.java
Januson: Done.

server/src/test/java/org/candlepin/dto/api/v1/SubscriptionTranslatorTest.java
Januson: Done.

server/src/test/java/org/candlepin/dto/api/v1/ProductContentTranslatorTest.java
Januson: Done.

server/src/test/java/org/candlepin/dto/api/v1/ProductContentTranslatorTest.java
Januson: Done.

server/src/main/java/org/candlepin/util/Util.java
Januson: Done.

server/src/main/java/org/candlepin/resource/SubscriptionResource.java
Januson: Done.

server/src/main/java/org/candlepin/dto/shim/ProductDataTranslator.java
Januson: Done.

server/src/main/java/org/candlepin/dto/api/v1/SubscriptionTranslator.java
Januson: Done.

server/src/main/java/org/candlepin/dto/api/v1/ProductTranslator.java
Januson: Done.

server/src/main/java/org/candlepin/dto/api/v1/ProductContentTranslator.java
Januson: Done.

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1525,67 @@ components:
...
+            branding:
+              type: array
+              uniqueItems: true
+              items:
+                $ref: "#/components/schemas/BrandingDTO"
+            href:
+              type: string
+              example: /products/ff808081554a3e4101554a3e9033005d
+            locked:
+              type: boolean
```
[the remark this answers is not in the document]

server/src/main/java/org/candlepin/controller/ProductManager.java
```diff
@@ -825,16 +834,39 @@ public static boolean isChangedBy(Product entity, ProductDTO dto) {
...
         }
 
         return false;
     }
 
+
+    /**
+     * Utility method that returns a Comparator for objects that implement BrandingInfo.
+     *
+     * @return A comparator for BrandingInfo objects.
```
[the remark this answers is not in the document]
Januson: Done.

api/candlepin-api-spec.yaml
```diff
@@ -918,6 +930,152 @@ paths:
...
+      security: []
+      parameters:
+        - name: id
+          in: path
+          description: The ID of the subscription to retrieve
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -918,6 +930,152 @@ paths:
...
+      parameters:
+        - name: id
+          in: path
+          description: The ID of the subscription certificate to retrieve
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: Status information about running Candlepin server
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -918,6 +930,152 @@ paths:
...
+        404:
+          description: Subscription could not be found using the provided ID
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ExceptionMessage'
+        default:
+          $ref: '#/components/responses/default'
+
+  /subscriptions/{id}/cert:
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -918,6 +930,152 @@ paths:
...
+            type: string
+        - name: emailLocale
+          in: query
+          description: The ID of the subscription to retrieve
+          required: true
+          schema:
+            type: string
+      responses:
+        202:
+          description: A list of subscriptions
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -918,6 +930,152 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: email
+          in: query
+          description: The ID of the subscription to retrieve
+          required: true
+          schema:
+            type: string
+        - name: emailLocale
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1525,67 @@ components:
...
+              type: array
+              uniqueItems: true
+              items:
+                $ref: "#/components/schemas/ProductContentDTO"
+            dependentProductIds:
+              type: array
+              uniqueItems: true
+              items:
+                type: string
+            branding:
```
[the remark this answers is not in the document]
Januson: Done.

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1525,67 @@ components:
             upstreamConsumer:
               $ref: '#/components/schemas/UpstreamConsumerDTO'
 
+    ProductContentDTO:
+      description: Product content mapping exposed to the API
+      required:
+        - id
+      properties:
+        content:
+          $ref: "#/components/schemas/ContentDTO"
```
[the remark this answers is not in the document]
Januson: Done.

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1525,67 @@ components:
...
+        enabled:
+          type: boolean
+
+    ProductDTO:
+      description: DTO representing the product data exposed to the API
+      required:
+        - id
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1070,22 +1229,34 @@ components:
...
-          type: string
-        displayName:
+        name:
           type: string
-        href:
+        value:
           type: string
 
+    BrandingDTO:
+      description: Represents the branding
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
Januson: Done.

api/candlepin-api-spec.yaml
Januson: Done. *LegacyDTO dtos were removed

api/candlepin-api-spec.yaml
Januson: Done. *LegacyDTO dtos were removed

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1517,90 @@ components:
...
+    ProductLegacyDTO:
+      description: DTO representing the product data exposed to the API
+      required:
+        - id
+      allOf:
+        - $ref: '#/components/schemas/TimestampedEntity'
+        - type: object
+          properties:
+            id:
+              type: string
```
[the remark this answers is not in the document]
Januson: Done.

api/candlepin-api-spec.yaml
```diff
@@ -1343,6 +1517,90 @@ components:
...
+        - $ref: '#/components/schemas/ProductLegacyDTO'
+        - type: object
+          properties:
+            attributes:
+              $ref: '#/components/schemas/AttributesDTO'
+
+    AttributesDTO:
+      description: Version and Status information about running Candlepin server
+      required:
+        - id
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1411,6 +1669,80 @@ components:
         keycloakResource:
           type: string
 
+    SubscriptionLegacyDTO:
+      description: Version and Status information about running Candlepin server
+      properties:
+        id:
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1411,6 +1669,80 @@ components:
         keycloakResource:
           type: string
 
+    SubscriptionLegacyDTO:
+      description: Version and Status information about running Candlepin server
+      properties:
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -1411,6 +1669,80 @@ components:
         keycloakResource:
           type: string
 
+    SubscriptionLegacyDTO:
+      description: Version and Status information about running Candlepin server
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/resteasy/JsonProvider.java
```diff
@@ -113,6 +115,7 @@ public JsonProvider(boolean indentJson, ProductCachedSerializationModule product
         customModule.addSerializer(Date.class, new DateSerializer());
         // Ensure we handle releaseVer fields properly
         customModule.addDeserializer(ReleaseVerDTO.class, new ReleaseVersionWrapDeserializer());
+        customModule.addDeserializer(AttributesDTO.class, new CandlepinAttributesDeserializer());
         mapper.registerModule(customModule);
```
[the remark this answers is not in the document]
Januson: Done.

server/spec/autobind_disabled_for_owner_spec.rb
```diff
@@ -37,7 +37,6 @@
     @owner['autobindHypervisorDisabled'] = true
     @cp.update_owner(@owner['key'], @owner)
     @owner = @cp.get_owner(@owner['key'])
-    puts (@owner.inspect())
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
Januson: Removed *LegacyDTOs

server/src/main/java/org/candlepin/jackson/CandlepinAttributesDeserializer.java
Januson: Done. Not needed anymore after removal of ProductLegacyDTO.

server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -306,28 +303,56 @@ public Content updateContent(ContentDTO update, Owner owner, boolean regenerateE
...
+        return null;
+    }
 
-        return updated;
+    private void addContent(ProductDTO product, ContentDTO update) {
+        if (update == null || update.getId() == null) {
+            throw new IllegalArgumentException("update references incomplete content");
+        }
+
+        Set<ProductContentDTO> productContent = product.getProductContent();
```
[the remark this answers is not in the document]
Januson: That is a problem with our current use of dtos. DTO is breaking encapsulation by definition. Using them as a makeshift domain objects is only making it worse. We are modifying the all over the place which we should not do.

api/candlepin-api-spec.yaml
```diff
@@ -918,6 +930,152 @@ paths:
...
+    post:
+      description: Activates a Subscription
+      tags:
+        - Subscription
+      operationId: activateSubscription
+      security: []
+      x-java-response:
+        type: javax.ws.rs.core.Response
+      parameters:
+        - name: consumerUuid
```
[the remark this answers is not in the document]

server/src/main/java/org/candlepin/dto/api/v1/ProductCertificateDTO.java
```diff
@@ -35,7 +35,7 @@
     protected String id;
     protected String key;
     protected String cert;
-    private ProductDTO product;
+    private ProductLegacyDTO product;
```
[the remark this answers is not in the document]
Januson: Done. Removed ProductLegacyDTO

server/src/main/java/org/candlepin/resteasy/JsonProvider.java
```diff
@@ -174,13 +182,17 @@ private void addMixInAnnotationsForDTOs(ObjectMapper mapper) {
...
         mapper.addMixIn(RoleDTO.class, DynamicPropertyFilterMixIn.class);
         mapper.addMixIn(StatusDTO.class, DynamicPropertyFilterMixIn.class);
         mapper.addMixIn(TimestampedEntity.class, DynamicPropertyFilterMixIn.class);
         mapper.addMixIn(UserDTO.class, DynamicPropertyFilterMixIn.class);
         mapper.addMixIn(OwnerDTO.class, DynamicPropertyFilterMixIn.class);
+        mapper.addMixIn(SubscriptionDTO.class, DynamicPropertyFilterMixIn.class);
         mapper.addMixIn(UpstreamConsumerDTO.class, DynamicPropertyFilterMixIn.class);
         //TODO: Add more mix-ins here as more DTOs are ported to openapi spec
+
+        mapper.addMixIn(ProductDTO.class, ProductAttributesMixIn.class);
```
nikosmoum, line 196: It looks like adding a second mixin for the same target class doesn't work (it overrides the first one). So the filtering that is provided by the DynamicPropertyFilterMixIn addition doesn't work right not (I've created the following spec test to test dynamic filtering: http://pastebin.test.redhat.com/857369).

To fix it, I tried making ProductAttributesMixIn implement DynamicPropertyFilterMixIn, but that didn't work. Then I found [this](https://github.com/FasterXML/jackson-docs/wiki/JacksonMixInAnnotations#usage-notes), which says that it should work using inheritance, so I turned DynamicPropertyFilterMixIn from an interface to an abstract class, and had ProductAttributesMixIn extend it, and it did work, so we should do that

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
Januson: Done.

server/src/main/java/org/candlepin/resource/OwnerProductResource.java
Januson: Done.

server/src/main/java/org/candlepin/controller/ContentManager.java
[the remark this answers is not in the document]
Ceiu: We've already run into problems a number of times before by doing this. Making assumptions about how the DTOs map their data isn't a good practice here. As soon as you do, you're coupling the internal logic of two completely separate things -- one of which we have very little control over now.

Should ContentManager be using DTOs these days? Absolutely not. It's an area that's very deserving of a refactor. But that's not reason enough to exacerbate the problem.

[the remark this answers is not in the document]
Januson: We are apparently agreeing about the problem. What do you suggest w should do? I refactored the code so that the collection are at least not modified in place, but coupling is still there.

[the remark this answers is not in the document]
nikosmoum: In the interest of getting this PR merged, and not adding additional work, I think the solution
for now is good. But we should find a better way of doing this in another task.

I spent sometime trying to understand why we even use DTOs in this way in the ContentManager.
Firstly, using DTOs in Product/ContentManager is no problem in itself, if what we're doing is
simply accepting a DTO as an update from a resource endpoint and only call getters on it.
And I agree with both of you, we shouldn't be altering the data on the DTO.
The reason (which seems uniquely affecting ContentManager but not ProductManager) looks to me to be:

What we want to do:
- We're fetching entities from the DB (Content and Product) and want to process them (involves altering the data on them).
- Then we want to pass on that data to the ProductManager (because they're intrinsically connected), BUT, the ProductManager
  accepts DTOs as input, so we have to translate the entities into DTOs and then pass them to ProductManager.

What we're actually doing:
1. Fetching entities from the DB (alternate versions of content, affected products etc.)
2. We're explicitly using a translator and turning them into DTOs (_ONLY because we want to later send them to ProductManager, which accepts DTOs_)
3. We're processing/altering them (here is the current problem).
4. Sending them off to ProductManager.

What I think we should be doing:
1. Fetching entities from DB.
2. We're processing/altering them. **
3. We're translating them to DTOs.
4. Sending them to ProductManager.

** If we have hibernate issues in step 2 here, as some comments in the code imply,
where altering collections on entities is a problem, then we have some alternatives:
a. Up until now we used DTOs (but now this is a problem, so something else should be used)
b. We can clone the entities (so that they are not hibernate-managed, and alter freely)
c. We can create new data containers (you could call them DTOs) that provide the encapsulation
we need.

I think a is not an option, and c is overkill (because we'd still need to re-translate into api DTOs), and prefer option b.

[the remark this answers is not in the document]
Ceiu: @Januson 
We should be avoiding directly modifying the collection and assuming it will update the DTO's state. That last part is what makes this class dependent on the inner workings of the DTO. We should, instead, create a new collection of the content we want, and assign that to the DTO.

@nikosmoum 
Some history on why DTOs were ever used in the Product/Content manager:
Originally, it only accepted model objects, as the expectation was that it would take a modified product, fetch a new instance from the DB exactly how it's represented on the backend, and do all its calculations based on the modified local (potentially attached) instance, and the detached instance it fetched representing the current DB state.

However, due to how we've configured Hibernate (primarily, auto-commit/flush), there are a number of operations which cause Hibernate to be "helpful" and persist some changes to managed objects, even if we haven't explicitly called .merge or .update. The first version had a number of weird bugs where collections on products and content would not be reflected in version comparisons, because Hibernate had persisted the change for us as soon as we did *any* select on the DB against the table owning the object -- even if it was for a different entity/instance.

This led to a number of choices, and after weighing the pros/cons of them, we, perhaps erroneously, opted to use DTOs at this layer to allow us to pass in the changes without risking putting a managed object in an unknown state.

Since then, we've made a number of other changes, such as making product and content immutable, which may fix the issue entirely by means of telling Hibernate to never change the object; we've added the service adapter model, which is how most of these changes are made today; and I'm currently working on a refresh refactor which pulls much of the work out of these managers.

It's entirely probable that in the very near future, we'll have a clear view of what these managers need to do, and how we should go about using them. The resource refactor should be the final nail in the coffin of passing the DTOs into the managers, and will likely be the optimal point for addressing the remaining components of the managers in question.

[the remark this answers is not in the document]
nikosmoum: @Ceiu @Januson Alright, we can address the long term impact after those changes. For the short-term changes in this PR then, the change should be: iterate the DTO collection items and add them to an entirely new collection, modify the new collection as needed, then add it back to the DTO, right?

api/candlepin-api-spec.yaml
```diff
@@ -918,6 +930,90 @@ paths:
...
+            type: string
+      responses:
+        202:
+          description: A Subscription is being activated
+        400:
+          description: Invalid request such as missing email/locale or when the consumer with the given uuid was not found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ExceptionMessage'
```
wolfdale, line 987: Can we please add `example` property here. Something like in https://github.com/candlepin/candlepin/pull/2684
it will allow YAML file to be consistent.

api/candlepin-api-spec.yaml
```diff
@@ -918,6 +930,90 @@ paths:
...
+            type: string
+      responses:
+        204:
+          description: Subscription was successfully deleted
+        404:
+          description: Subscription could not be found using the provided ID
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ExceptionMessage'
```
wolfdale, line 1013: Can we please add `example` property here.  Something like in https://github.com/candlepin/candlepin/pull/2684
it will allow YAML file to be consistent.

Januson: Done

api/candlepin-api-spec.yaml
Januson: Done

PR #2676, opened 2020-04-07
[M] Reverted the renaming of content modified products

Ceiu:
- Reverted the method rename on "modified products" from the new
  "required products" naming convention


Ceiu:
Alright

PR #2677, opened 2020-04-09
[M] Remove non-ASCII character that breaks tito rpm build

nikosmoum:
- This fixes the following error thrown by the Cheetah template
  engine used by tito, when building the rpm:

  File "/usr/lib64/python2.7/site-packages/Cheetah/Compiler.py", line 1711, in `__init__`
    source = unicode(source)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 8104: ordinal not in range(128)

### DIFF
--- a/server/candlepin.spec.tmpl
+++ b/server/candlepin.spec.tmpl
@@ -234,7 +234,7 @@ fi
 - 1692411: Added additional product ids into consumer installed products
   Current rule code is only to consider consumer installed products to
   calculate the pool's average priority. Hence added any additional products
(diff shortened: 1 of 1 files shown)


nikosmoum:
To test:
- Install tito (instructions [here](https://github.com/rpm-software-management/tito#install))
- cd into `<candlepin_root>/server/`
- Run `tito build --rpm --test --verbose --debug`

candlepin test rpms should be built and you should not get the error mentioned in the commit message.

PR #2678, opened 2020-04-09
1812110: Expose artemis port with SELinux boolean; ENT-2190

nikosmoum:


server/candlepin.spec.tmpl
```diff
@@ -170,6 +172,7 @@ for selinuxvariant in %{selinux_variants}
 do
   /usr/sbin/semodule -s $selinuxvariant -i %{_datadir}/selinux/$selinuxvariant/%{modulename}.pp &> /dev/null || :
 done
+/usr/sbin/semanage port -a -t candlepin_activemq_port_t -p tcp 61613 &> /dev/null || :
```
Ceiu, line 175: Do we want to define the port as a variable somewhere so it's not hard coded? Is that even a thing we can do?

server/selinux/candlepin.te
```diff
@@ -1,10 +1,19 @@
...
 # Declarations
 #
 
+## <desc>
+##      <p>
+##	    Allow tomcat to bind to the port used
+##	    by candlepin's embedded activemq artemis,
+##	    labeled candlepin_activemq_port_t
+##	    </p>
+## </desc>
```
Ceiu, line 14: Indentation is a bit wonky here

server/selinux/candlepin.te
```diff
@@ -103,3 +115,7 @@ allow tomcat_t candlepin_var_lib_t:file { manage_file_perms };
 allow tomcat_t candlepin_var_log_t:dir { manage_dir_perms };
 allow tomcat_t candlepin_var_log_t:file { manage_file_perms };
 allow tomcat_t postgresql_port_t:tcp_socket name_connect;
+
+tunable_policy(`candlepin_can_bind_activemq_port',`
+	allow tomcat_t candlepin_activemq_port_t:tcp_socket name_bind;
```
Ceiu, line 120: Indentation again. Looks like a tab instead of spaces?

server/candlepin.spec.tmpl
nikosmoum: Given that this is in the spec file, I don't know where/how you would create a variable for it, except the spec file itself, but it would be only for readability/aesthetic reasons, not functional.

Ceiu: Oh yeah... missed that. It looks like bash scripting; though I can't quite figure out exactly what that double pipe is doing at the end. Is that just to suppress any error output from the command?

nikosmoum: It is running bash commands, since this is inside a `#raw` block: https://stackoverflow.com/questions/16357755/meaning-of-colon-in-bash-after-a-double-pipe#comment23436601_16357803

server/selinux/candlepin.te
nikosmoum: fixed

server/selinux/candlepin.te
nikosmoum: fixed

kahowell:
@jturel (and anyone else), some quick notes for getting this installed (extrapolated from the spec file):

1. `sudo yum install selinux-policy-devel -y`
2. From this branch: `cd server/selinux`
3. `make NAME=targeted -f /usr/share/selinux/devel/Makefile`
4. `sudo semodule -i candlepin.pp` (this will take a minute or so)
5. See that the new bool exists:
```
$ getsebool -a | grep candlepin
candlepin_can_bind_activemq_port --> off
```
6. Toggle the bool as desired via `sudo setsebool candlepin_can_bind_activemq_port on` or `sudo setsebool candlepin_can_bind_activemq_port off`
7. Label the desired port as the candlepin activemq port (the one we're writing into the spec is below):
`sudo semanage port -a -t candlepin_activemq_port_t -p tcp 61613`

Note: If you want to downgrade/remove the module, you should unlabel the port as follows (extracted from this PR):
```
sudo semanage port -d 61613 -p tcp
```

jturel:
Thanks for the steps @kahowell - I was able to test!

@nikosmoum this change looks good to me. One thing I do see in the audit.log is this:

```
type=AVC msg=audit(1586825398.050:6940): avc:  denied  { execute } for  pid=8953 comm="localhost-start" path="/var/cache/tomcat/temp/libnetty_transport_native_epoll_x86_645599493681718427018
.so" dev="vda1" ino=34619671 scontext=system_u:system_r:tomcat_t:s0 tcontext=system_u:object_r:tomcat_cache_t:s0 tclass=file permissive=0
```

This only shows up when I've enabled my stomp acceptor in broker.xml:

```
<acceptor name="stomp">tcp://localhost:61613?protocols=STOMP;sslEnabled=true;needClientAuth=true;keyStorePath=/etc/candlepin/certs/server-side-keystore.jks;keyStorePassword=changeme;trustStorePath=/etc/candlepin/certs/server-side-truststore.jks;trustStorePassword=changeme</acceptor>```

I'm not sure what the implications are when that temp dir cannot be written to but it might be worth addressing. Let me know what you think.

nikosmoum:
@jturel that denial is for executing a native library (it's an .so file), so I'd be surprised if it had no effect. Apparently this is something that happens according to [this](https://noobient.com/2017/10/02/selinux-woes-with-tomcat-on-centos-74/) explanation (see "Native libraries" section). I'd be worry to add an allowance for tomcat to execute things in the cache, because it really sounds like it shouldn't (and it's not even that we could allow that single file to be executed, since it's name looks to be dynamically generated). Can you check what the implications are if this is not addressed? (e.g. can you use the STOMP acceptor?)

PR #2680, opened 2020-04-13
[M] Added support for JAAS in the embedded ActiveMQ broker (ENT-2154)

Ceiu:
- Added support for the using the ActiveMQJAASSecurityManager when
  creating the embedded ActiveMQ broker
- Added new configurations for modifying which entries are used
  in a JAAS config file
- Updated the broker.xml with details about using external
  connections to the embedded Artemis instance

### DIFF
--- a/server/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/server/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -62,7 +62,8 @@ public class ConfigProperties {
 
     /*
      * XXX The actual property key refers to HornetQ which was ActiveMQ's ancestor.  We have to keep the
-     * key unchanged for compatibility reasons.
+     * key unchanged for compatibility reasons. These are deprecated, however, and should be replaced by
+     * newer configuration values as features are added.
      */
     public static final String ACTIVEMQ_ENABLED = "candlepin.audit.hornetq.enable";
 
@@ -150,8 +151,14 @@ public class ConfigProperties {
     // TODO:
     // Clean up all the messaging configuration. We have all sorts of prefixes and definitions for
     // common broker options, and stuff which is unique to specific brokers. We should be defining
-    // them as "candlepin.<broker type>.<setting>" instead of the various sections we have now.
+    // them as "candlepin.messaging.<broker type>.<setting>" instead of the various sections we
+    // have now.
+    public static final String ACTIVEMQ_EMBEDDED_BROKER = "candlepin.messaging.activemq.embedded.enabled";
 
+    public static final String ACTIVEMQ_JAAS_INVM_LOGIN_NAME =
+        "candlepin.messaging.activemq.embedded.jaas_invm_login_name";
+    public static final String ACTIVEMQ_JAAS_CERTIFICATE_LOGIN_NAME =
+        "candlepin.messaging.activemq.embedded.jaas_certificate_login_name";
 
     // AMQP stuff
     public static final String AMQP_INTEGRATION_ENABLED = "candlepin.amqp.enable";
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -362,9 +369,19 @@ public static String jobConfig(String jobKey, String cfgName) {
...
 
             this.put(ACTIVEMQ_ENABLED, "true");
+
             this.put(ACTIVEMQ_EMBEDDED, "true");
+
+            // TODO: Delete the above configuration and only use EMBEDDED_BROKER going forward.
+            // This is currently blocked by complexity in supporting moving configuration paths,
+            // and should coincide with an update to the Configuration object
+            this.put(ACTIVEMQ_EMBEDDED_BROKER, "true");
+            this.put(ACTIVEMQ_JAAS_PROPERTIES_LOGIN_NAME, "PropertiesLogin");
```
jturel, line 379: Maybe the default should be "InVmLogin" or "GuestLogin" to be more precise since we're not actually using the PropertiesLogin module

server/src/main/resources/broker.xml
```diff
@@ -19,16 +19,88 @@
...
+        <!-- set this to true if using external clients -->
         <security-enabled>false</security-enabled>
 
+        <security-settings>
+            <!--
+                Enable this when using the embedded broker with the security manager enabled.
+                Remember sure to update the role name to match the role given to the "guest"
+                in-vm user, following the login details provided above.
+
+            <security-setting match="#">
```
jturel, line 89: I have a suspicion that these permissions are broader than what's actually needed. I experimented with this, and it seems to work but it's likely that i'm not testing all the execution paths:

```
                <security-setting match="job.jobs">
                        <permission type="createDurableQueue" roles="invm-role"/>
                        <permission type="deleteDurableQueue" roles="invm-role"/>
                        <permission type="send" roles="invm-role"/>
                        <permission type="consume" roles="invm-role"/>
                </security-setting>
                <security-setting match="event.default.#">
                        <permission type="createDurableQueue" roles="invm-role"/>
                        <permission type="deleteDurableQueue" roles="invm-role"/>
                        <permission type="send" roles="invm-role"/>
                        <permission type="consume" roles="invm-role"/>
                </security-setting>
```

jturel:
Also verified that my existing configuration (security-mode=false) works in a backward compatible way :+1:

server/src/main/resources/broker.xml
Ceiu: It's definitely a sledgehammer, but I don't want to limit or break in-vm functionality every time we add something going forward. I'm pretty sure we need the manage side of things today, and if not, it'll be coming soon with the rework of the status monitor.

I'm somewhat torn here in that I understand the concern for security, but I also feel that the connection Candlepin makes to its own embedded broker shouldn't be encumbered at all. That it even is subject to the security manager seems somewhat silly.

In any event, what is the update flow look like from your end if we have to update the permissions on a release?

server/src/main/java/org/candlepin/config/ConfigProperties.java
Ceiu: Ahh, I thought that was the convention for the first config name in the JAAS module, but I was mistaken. I've updated this to use InVMLogin as suggested.

server/src/main/resources/broker.xml
jturel: That's fair to say w.r.t limiting or breaking functionality. Something like that could go easily missed on our side, too. We'd need to update our installer where we will have the broker.xml puppetized. Let's leave it as you have it.

jturel:
@Ceiu what are the chances of getting this merged and in a build by the end of the week? we've wrapped up installer changes on our side and need new build upstream to merge. we're excited!!

PR #2681, opened 2020-04-15
ENT-2252: Upgrade openapi-generator to latest version

wolfdale:


wolfdale:
Closing this PR.
Due to a bug in openAPI generator (v4.3.0), its not a worth while implementing a workaround. Bug is intended to be fixed in next release of OpenAPITools. So holding off upgrading until it is fixed.
More details on JIRA card about bug and issue tracker.

PR #2682, opened 2020-04-17
ENT-1859: Port EnvironmentDTO to openapi spec

wolfdale:
- Updated candlepin YAML file.
- Removed EnvironmentDTO & EnvironmentDTOTest files.
- Updated EnvironmentTranslator & other related tests.


api/candlepin-api-spec.yaml
```diff
@@ -1265,6 +1265,35 @@ components:
...
+              type: string
+            name:
+              type: string
+            description:
+              type: string
+            owner:
+              $ref: "#/components/schemas/NestedOwnerDTO"
+            environmentContent:
+              type: array
+              uniqueItems: true
```
sonalidhome, line 1284: I hope this will not change the API schema defination. Previously it was not unique.

nikosmoum: It should be fine, because even thougn now it is a Collection<>, before the DTO was created we were using the Environment.java model entity, which uses a Set, so we should be fine

wolfdale: No I don't think it will impact the APIs.
Earlier `environmentContent` property was backed by a HashMap (I assume this was done to gain performance with quick equality check while adding a content), now it is changed with a Set representation, this is much more aligned with the entity class `Environment` which also follows a Set representation for EnvironmentContent.

api/candlepin-api-spec.yaml
```diff
@@ -1265,6 +1265,35 @@ components:
           type: string
           minLength: 1
 
+    EnvironmentDTO:
+      description: DTO representing an environment
```
nikosmoum, line 1269: Since we're trying to have this spec also be a good source of documentation, we should provide better descriptions.

A good guideline for all descriptions is: Try to use information from the model entity's class javadoc comment (in this case Environment.java), instead of from the DTO, because our DTOs don't have useful documentation (it's always something like "A DTO representation of X entity"). Then, take that and try to distill it to not have any implementation details, and be concise. In this case it could be something like: "Represents an environment within an org used to enable/disable/promote content in specific places"

server/src/test/java/org/candlepin/dto/api/v1/ConsumerDTOTest.java
```diff
@@ -92,6 +92,22 @@ public ConsumerDTOTest() {
 
         NestedOwnerDTO owner = new NestedOwnerDTO().id("OwnerId").displayName("Name").key("12345");
 
+        ContentDTO cDTO = new ContentDTO()
+            .id("ContentId")
+            .name("ContentName")
+            .modifiedProductIds(new HashSet<>(Arrays.asList("ID1", "ID2")));
+
+        Set<EnvironmentContentDTO> envContentDTO = new HashSet<>(Arrays.asList(new EnvironmentContentDTO()
```
nikosmoum, line 100: Nitpick: No need for Arrays.asList here if only 1 element is being added, you can create empty HashSet and then add it

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentTranslator.java
```diff
@@ -64,35 +62,48 @@ public EnvironmentDTO populate(Environment source, EnvironmentDTO dest) {
...
+        }
 
-        dest.setId(source.getId());
-        dest.setName(source.getName());
-        dest.setDescription(source.getDescription());
+        Date created = source.getCreated();
+        dest.created(created != null ? created.toInstant().atOffset(ZoneOffset.UTC) : null);
+
+        Date updated = source.getUpdated();
+        dest.updated(updated != null ? updated.toInstant().atOffset(ZoneOffset.UTC) : null);
```
nikosmoum, line 77: One more minor change, that we could apply in general from now on: Use org.candlepin.util.Util.toDateTime() to as a utility converter from Date to OffsetDateTime

api/candlepin-api-spec.yaml
Ceiu: While true, one thing we want to be careful about is adding restrictions that will inadvertently break current clients (even if they're doing silly things).

When it comes to API or communication design, I like to follow Postel's law as much as possible. Basically: be conservative in what you send, be liberal in what you accept.

server/src/test/java/org/candlepin/dto/api/v1/ConsumerDTOTest.java
Ceiu: There's also ```Collections.singleton``` and ```Util.asSet```

api/candlepin-api-spec.yaml
wolfdale: Agreed. Added description for `EnvironmentDTO` & `EnvironmentContentDTO`

server/src/test/java/org/candlepin/dto/api/v1/ConsumerDTOTest.java
wolfdale: Done.

server/src/main/java/org/candlepin/dto/api/v1/EnvironmentTranslator.java
wolfdale: Fixed.

PR #2683, opened 2020-04-20
[F]ENT-2291: Create adapter methods for DTOs implementing *Info Interfaces

sonalidhome:
- Added a class InfoAdapter to provide the implementation of info interfaces.
 - Removed the below translators and their test suites:
   - UserDTOTranslator
   - RoleDTOTranslator
   - PermissionBluePrintDTOTranslator


sonalidhome:
A new class is created to provide the implementation of the info interfaces. Idea is to provide implementations at a single place. If this design is agreed then we can move the other methods implemented in `HostedTestSubscriptionResource` class.

server/src/main/java/org/candlepin/dto/api/v1/InfoAdapter.java
```diff
@@ -0,0 +1,230 @@
...
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.dto.api.v1;
```
nikosmoum, line 16: I would not put this in this package, since it isn't part of our API. It's a utility class for internal transformations of data. I'd say `org.candlepin.resource.util` is more appropriate

server/src/main/java/org/candlepin/dto/api/v1/InfoAdapter.java
```diff
@@ -0,0 +1,230 @@
...
+            public Boolean isSuperAdmin() {
+                return source.getSuperAdmin();
+            }
+
+            /**
+             * {@inheritDoc}
+             */
+            @Override
+            public Collection<? extends RoleInfo> getRoles() {
+                return null;
```
nikosmoum, line 198: why null?

server/src/main/java/org/candlepin/dto/api/v1/InfoAdapter.java
```diff
@@ -0,0 +1,230 @@
...
+    public static OwnerInfo ownerInfoAdapter(NestedOwnerDTO source) {
+        return new OwnerInfo() {
+            @Override
+            public String getKey() {
+                return source.getKey();
+            }
+
+            @Override
+            public Date getCreated() {
+                return null;
```
nikosmoum, line 220: why null?

server/src/main/java/org/candlepin/dto/api/v1/InfoAdapter.java
```diff
@@ -0,0 +1,230 @@
...
+            }
+
+            @Override
+            public Date getUpdated() {
+                return null;
```
nikosmoum, line 225: why null?

server/src/main/java/org/candlepin/dto/api/v1/InfoAdapter.java
sonalidhome: `NestedOwnerDTO` does not have properties Created/Updated. Hence returned null by default.

server/src/main/java/org/candlepin/dto/api/v1/InfoAdapter.java
sonalidhome: `NestedOwnerDTO` does not have properties Created/Updated. Hence returned null by default.

server/src/main/java/org/candlepin/dto/api/v1/InfoAdapter.java
sonalidhome: `UserDTO` does not have property `roles`

server/src/main/java/org/candlepin/dto/api/v1/InfoAdapter.java
Ceiu: Util is probably best if we're going to have a separate class for it. However, we'll need to be careful and ensure that any place which uses the functions are heavy scrutinized so we don't end up using this in places that should be updated in a different way.

Ceiu:
Seems generally fine to me. Though I'll be interested to see how it ends up working in practice since this version is converting only the filtered User DTO rather than the full DTO we should be expecting as input.

server/src/main/java/org/candlepin/dto/api/v1/InfoAdapter.java
sonalidhome: Class relocated to `org.candlepin.resource.util` package.

sonalidhome:
Moved the adapter methods from the `HostedTestSubscriptionResource` class.

PR #2688, opened 2020-04-23
[2.9] 1826545: Added an additional null check during entitlement revocation (ENT-2320)

Ceiu:
- Added an additional null check during entitlement revocation
  to avoid a null-pointer exception
- Removed an unnecessary list copy during entitlement revocation


Ceiu:
rebased

PR #2691, opened 2020-04-27
ENT-1870: Port ContentResource to spec-first

Januson:


api/candlepin-api-spec.yaml
```diff
@@ -489,6 +489,60 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/ContentDTO'
+        default:
+          $ref: '#/components/responses/default'
+
+  /content/content_uuid:
```
wolfdale, line 514: Minor issue - Use `{}` here for path parameter.

Januson: Good catch. Removed the spec test and fixed those braces.

PR #2693, opened 2020-04-30
ENT-1885: Port RoleResource to spec-first

wolfdale:
- Updated candlepin.yaml schema file with API definition.
- Updated RoleResouce file with API implementation.

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -1124,6 +1124,390 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /roles:
+    get:
+      description: Retrieves a list of Roles
+      tags:
+        - Roles
+      operationId: getRoles
+      x-java-response:
+        type: java.util.stream.Stream
+        isContainer: true
+      security: []
+      responses:
+        200:
+          description: A list of roles
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/RoleDTO'
+        default:
+          $ref: '#/components/responses/default'
+
+    post:
+      tags:
+        - Roles
+      description: Creates a Role
+      operationId: createRole
+      security: []
+      requestBody:
+        description: A role to be created
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/RoleDTO'
+      responses:
+        200:
+          description: Role successfully created.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RoleDTO'
+        400:
+          description: Role is null or empty or role name not specified
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ExceptionMessage'
+              example:
+                displayMessage: Role is null or empty or role name not specified
+                requestUuid: c4347004-8792-41fe-a4d8-fccaa0d3898a
+        409:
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/resource/RoleResource.java
```diff
@@ -30,38 +30,16 @@
...
- *
+ * API implementation for Role operations
  */
-@Path("/roles")
-@Api(value = "roles", authorizations = { @Authorization("basic") })
-public class RoleResource {
+public class RoleResource implements RolesApi {
 
     private UserServiceAdapter userService;
     private OwnerCurator ownerCurator;
```
Januson, line 45: Not part of your changes but ownerCurator and permissionCurator are not used anymore so we could get rid of them.

api/candlepin-api-spec.yaml
```diff
@@ -1124,6 +1124,390 @@ paths:
...
+      parameters:
+        - in: path
+          name: role_name
+          description: Role name
+          required: true
+          schema:
+            type: string
+      security: []
+      responses:
+        200:
```
Januson, line 1296: We should return 204 as we are not returning any data.

wolfdale: Corrected.

server/src/main/java/org/candlepin/resource/RoleResource.java
wolfdale: Fixed it.

api/candlepin-api-spec.yaml
```diff
@@ -1124,6 +1124,390 @@ paths:
...
+              $ref: '#/components/schemas/PermissionBlueprintDTO'
+      responses:
+        200:
+          description: Permission to a Role is added
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RoleDTO'
+        400:
+          description: Access type NONE not supported
```
nikosmoum, line 1443: There is one more reason that this could return a 400: when the role_name is null or empty, so this could be: `Access type NONE not supported, or role name is null or empty`

server/src/main/java/org/candlepin/resource/RoleResource.java
```diff
@@ -283,14 +239,15 @@ public RoleDTO deleteUserFromRole(@PathParam("role_name") String roleName,
...
+     * {@InheritDoc}
+     */
+    @Override
+    public Iterable<RoleDTO> getRoles() {
         // TODO: Add in filter options
 
         Collection<? extends RoleInfo> roles = this.userService.listRoles();
-        return roles.stream().map(this.modelTranslator.getStreamMapper(RoleInfo.class, RoleDTO.class));
+        return roles.stream().map(this.modelTranslator.getStreamMapper(RoleInfo.class, RoleDTO.class))
+            .collect(Collectors.toSet());
```
nikosmoum, line 251: There's no reason to change this into returning a Set. We can still return a Stream here, by the x-java-response: type: to `java.util.stream.Stream` from `Iterable` (check other examples in the spec file)

wolfdale: Fixed it.

api/candlepin-api-spec.yaml
wolfdale: Done.

PR #2697, opened 2020-05-12
ENT-2316: Introduce Swatch Profile Timestamp

sonalidhome:
- Added `rh_cloud_profile_modified` column in consumer
 - Updated the ConsumerResource, HypervisorResource and HypervisorUpdateAction
   to add modified timestamp on certain conditions
 - Added Enum for specific facts
 - Test suites added for the respective changes

### DIFF
--- a/server/src/main/java/org/candlepin/model/CloudProfileFacts.java
+++ b/server/src/main/java/org/candlepin/model/CloudProfileFacts.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright (c) 2009 - 2020 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.model;
+
+/**
+ * The subscription profile time stamp should also get updated
+ * on the specific Facts updates.
+ * This Enum helps to configure more facts if required in the future.
+ */
+public enum CloudProfileFacts {
+
+    CPU_CORES_PERSOCKET("cpu.core(s)_per_socket"),
+    CPU_SOCKETS("cpu.cpu_socket(s)"),
+    DMI_BIOS_VENDOR("dmi.bios.vendor"),
+    DMI_BIOS_VERSION("dmi.bios.version"),
+    DMI_CHASSIS_ASSET_TAG("dmi.chassis.asset_tag"),
+    DMI_SYSTEM_MANUFACTURER("dmi.system.manufacturer"),
+    DMI_SYSTEM_UUID("dmi.system.uuid"),
+    MEMORY_MEMTOTAL("memory.memtotal"),
+    OCM_UNITS("ocm.units"),
+    UNAME_MACHINE("uname.machine"),
+    VIRT_IS_GUEST("virt.is_guest");
+
+    private final String fact;
+
+    CloudProfileFacts(String fact) {
+        this.fact = fact;
+    }
+
+    public String getFact() {
+        return this.fact;
+    }
+
+    @Override
+    public String toString() {
+        return fact;
+    }
+
+    public static boolean containsFact(String incomingFact) {
+        for (CloudProfileFacts fact : CloudProfileFacts.values()) {
+            if (fact.getFact().equals(incomingFact)) {
+                return true;
+            }
+        }
+        return false;
+    }
+}
--- a/server/src/main/java/org/candlepin/model/Consumer.java
+++ b/server/src/main/java/org/candlepin/model/Consumer.java
@@ -261,6 +261,9 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
     @JoinColumn(name = "owner_id", insertable = false, updatable = false)
     private Owner owner;
 
+    @Column(name = "rh_cloud_profile_modified")
+    private Date rhCloudProfileModified;
+
     public Consumer(String name, String userName, Owner owner, ConsumerType type) {
         this();
 
@@ -390,6 +393,7 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
      */
     public void setTypeId(String typeId) {
         this.typeId = typeId;
+        this.updateRHCloudProfileModified();
     }
 
     /**
@@ -402,8 +406,8 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
         if (type == null || type.getId() == null) {
             throw new IllegalArgumentException("type is null or has not been persisted");
         }
-
         this.typeId = type.getId();
+        this.updateRHCloudProfileModified();
     }
 
     /**
@@ -473,6 +477,9 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
      * @param factsIn facts about this consumer.
      */
     public void setFacts(Map<String, String> factsIn) {
+        if (this.checkForCloudProfileFacts(factsIn)) {
+            this.updateRHCloudProfileModified();
+        }
         facts = factsIn;
     }
 
@@ -519,9 +526,16 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
      * @param value to set
      */
     public void setFact(String name, String value) {
-        if (facts == null) {
-            facts = new HashMap<>();
+        if (this.facts == null) {
+            this.facts = new HashMap<>();
         }
+
+        HashMap<String, String> fact = new HashMap<>();
+        fact.put(name, value);
+        if (this.checkForCloudProfileFacts(fact)) {
+            this.updateRHCloudProfileModified();
+        }
+
         this.facts.put(name, value);
     }
 
@@ -631,6 +645,7 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
 
     public void setInstalledProducts(Set<ConsumerInstalledProduct> installedProducts) {
         this.installedProducts = installedProducts;
+        this.updateRHCloudProfileModified();
     }
 
     public void addInstalledProduct(ConsumerInstalledProduct installed) {
@@ -640,6 +655,7 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
 
         installed.setConsumer(this);
         installedProducts.add(installed);
+        this.updateRHCloudProfileModified();
     }
 
     public Boolean isAutoheal() {
@@ -656,6 +672,7 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
     @JsonProperty
     public void setGuestIds(List<GuestId> guests) {
         this.guestIds = guests;
+        this.updateRHCloudProfileModified();
     }
 
     /**
@@ -672,6 +689,7 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
         }
         guestId.setConsumer(this);
         guestIds.add(guestId);
+        this.updateRHCloudProfileModified();
     }
 
     public String getEntitlementStatus() {
@@ -688,6 +706,7 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
 
     public void setServiceLevel(String level) {
         this.serviceLevel = level;
+        this.updateRHCloudProfileModified();
     }
 
     public String getRole() {
@@ -696,6 +715,7 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
 
     public void setRole(String role) {
         this.role = role;
+        this.updateRHCloudProfileModified();
     }
 
     public String getUsage() {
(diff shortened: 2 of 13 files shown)


Ceiu:
While this will work, I feel like there's a safer way to cover every possible place the change to those values could be made: the Consumer object itself.

Any time we're calling a mutator for one of the target values, we should update the date on the new field. You could even get crazy and do it as a persist hook, but that's probably unnecessarily complex for this task.

nikosmoum:
@Ceiu I am not a fan of the persist hook, but the mutator approach might work.

sonalidhome:
The updated code is with the `Consumer` entity mutators approach. 
Setters are updating the current timestamp for relevant fields.

server/src/test/java/org/candlepin/resource/HypervisorCloudProfileTest.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+import javax.inject.Inject;
+
```
wolfdale, line 56: Please add class description here.

server/src/main/java/org/candlepin/model/CloudProfileFacts.java
```diff
@@ -0,0 +1,55 @@
...
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.model;
+
```
wolfdale, line 17: Please add description here.

server/src/test/java/org/candlepin/resource/HypervisorCloudProfileTest.java
sonalidhome: Added.

server/src/main/java/org/candlepin/model/CloudProfileFacts.java
sonalidhome: Added.

server/src/test/java/org/candlepin/resource/ConsumerResourceIntegrationTest.java
```diff
@@ -686,4 +690,87 @@ public void testContentAccessExpireRegen() {
...
+
+        updatedConsumerDTO = new ConsumerDTO();
+        updatedConsumerDTO.setFact(CloudProfileFacts.CPU_CORES_PERSOCKET.getFact(), "1");
+        consumerResource.updateConsumer(consumer.getUuid(), updatedConsumerDTO, principal);
+        Date modifiedTSOnNecessaryFactUpdate = consumerCurator.get(consumer.getId())
+            .getRHCloudProfileModified();
+
+        assertNotNull(consumerCurator.get(consumer.getId()).getRHCloudProfileModified());
+        assertNotEquals(modifiedTSOnUnnecessaryFactUpdate, modifiedTSOnNecessaryFactUpdate);
+    }
```
nikosmoum, line 775: It would probably be good to include tests in this class for at least a couple of the fields that are not supposed to update the timestamp, and verify that the timestamp stays the same after the update. Also, same for a couple of facts that are not cloud-profile related

server/src/test/java/org/candlepin/resource/HypervisorCloudProfileTest.java
```diff
@@ -0,0 +1,194 @@
...
+            .getRHCloudProfileModified();
+
+        guestIdResource.updateGuests(consumer.getUuid(), guestIds);
+        Date afterUpdateTimestamp = consumerCurator.findByUuid(consumer.getUuid())
+            .getRHCloudProfileModified();
+
+        assertNotNull(afterUpdateTimestamp);
+        assertNotEquals(beforeUpdateTimestamp, afterUpdateTimestamp);
+    }
+
```
nikosmoum, line 193: Same for this class, a couple tests for each of HypervisorResource and GuestIdResource where we verify the profile timestamp is not updated because it is a field we don't care about

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -893,13 +915,45 @@ public void setContentAccessMode(String contentAccessMode) {
...
+        this.setRHCloudProfileModified(new Date());
     }
 
     public void addAddOn(String addOn) {
         this.addOns.add(addOn);
+        this.setRHCloudProfileModified(new Date());
     }
 
     public void removeAddOn(String addOn) {
         this.addOns.remove(addOn);
```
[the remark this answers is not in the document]
sonalidhome: Added.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -519,9 +526,15 @@ else if (!myVal.equals(otherVal)) {
      * @param value to set
      */
     public void setFact(String name, String value) {
-        if (facts == null) {
-            facts = new HashMap<>();
+        if (this.facts == null) {
+            this.facts = new HashMap<>();
+        }
+
+        HashMap<String, String> fact = new HashMap<String, String>() {{ put(name, value); }};
```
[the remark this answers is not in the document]
sonalidhome: Gotcha! Updated.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -893,13 +915,45 @@ public void setContentAccessMode(String contentAccessMode) {
...
     public void addAddOn(String addOn) {
         this.addOns.add(addOn);
+        this.setRHCloudProfileModified(new Date());
     }
 
     public void removeAddOn(String addOn) {
         this.addOns.remove(addOn);
     }
+
+    public void setRHCloudProfileModified(Date rhCloudProfileModified) {
```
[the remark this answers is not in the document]
sonalidhome: Agreed, updated.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -712,6 +732,7 @@ public String getSystemPurposeStatus() {
 
     public void setSystemPurposeStatus(String systemPurposeStatus) {
         this.systemPurposeStatus = systemPurposeStatus;
+        this.setRHCloudProfileModified(new Date());
```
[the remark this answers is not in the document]
sonalidhome: This is not needed here, It got added mistakenly by the name system purpose. Now removed it.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -893,13 +915,45 @@ public void setContentAccessMode(String contentAccessMode) {
...
+
+    /**
+     * Check if the consumers facts have changed and contains the cloud profile facts.
+     * It returns true if incoming facts are cloud profile facts
+     *
+     * @param existing existing consumer
+     * @param incomingFacts incoming facts
+     * @return a boolean
+     */
+    public boolean checkForCloudProfileFacts(Consumer existing, Map<String, String> incomingFacts) {
```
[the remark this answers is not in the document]
sonalidhome: Agreed, removed the Consumer parameter.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -893,13 +915,45 @@ public void setContentAccessMode(String contentAccessMode) {
...
+            return false;
+        }
+        else if (!existing.factsAreEqual(incomingFacts)) {
+            for (String fact : incomingFacts.keySet()) {
+                if (CloudProfileFacts.containsFact(fact)) {
+                    return true;
+                }
+            }
+        }
+        return false;
```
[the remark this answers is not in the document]
sonalidhome: Thanks for noticing this. 
I have now revised the method as per the suggestion. Also, added few more tests to validate this method.

server/src/main/resources/db/changelog/20200430172317-add-column-rh-cloud-profile-modified.xml
```diff
@@ -0,0 +1,18 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20200430172317-1" author="sdhome">
+        <comment>Adding column rh_cloud_profile_modified for Swatch Profile Timestamp feature.
+        </comment>
+
+        <addColumn tableName="cp_consumer">
+            <column name="rh_cloud_profile_modified" type="${timestamp.type}"/>
```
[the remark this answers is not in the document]
sonalidhome: Somehow this worked in a development environment. But, it's good to have it added before any consequences.

server/src/test/java/org/candlepin/resource/ConsumerResourceIntegrationTest.java
sonalidhome: Added.

server/src/test/java/org/candlepin/resource/HypervisorCloudProfileTest.java
sonalidhome: Added.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -893,13 +915,63 @@ public void setContentAccessMode(String contentAccessMode) {
...
+     * @param incomingFacts incoming facts
+     * @return a boolean
+     */
+    public boolean checkForCloudProfileFacts(Map<String, String> incomingFacts) {
+
+        if (incomingFacts == null) {
+            return false;
+        }
+
+        if (this.facts != null && incomingFacts != null && this.facts.equals(incomingFacts)) {
```
nikosmoum, line 956: we already know `incomingFacts` is not null here, since if it was we would have returned false, so there is no need for this check

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -893,13 +915,63 @@ public void setContentAccessMode(String contentAccessMode) {
...
+            return false;
+        }
+
+        for (Entry<String, String> entry : incomingFacts.entrySet()) {
+            if (CloudProfileFacts.containsFact(entry.getKey())) {
```
nikosmoum, line 961: This will work, but it will have to iterate over all of the incoming facts (which usually are 200+) and check for each of them if it is a cloud-profile fact. A more performant loop would be to loop over the CloudProfileFacts.values() (which right now are only 11), then check if each of them is included in the incoming facts (and then continue the rest of the checks)

server/src/test/java/org/candlepin/resource/ConsumerResourceIntegrationTest.java
```diff
@@ -686,4 +694,179 @@ public void testContentAccessExpireRegen() {
...
+
+    @Test
+    public void testCloudProfileFactDidNotChange() {
```
nikosmoum, line 782: This, and the 4 following tests which don't call into ConsumerResource, should probably moved into the `ConsumerTest.java` file, since that seems more relevant there

sonalidhome: Agreed, Moved these tests to `ConsmerTest.java`.

server/src/main/java/org/candlepin/model/Consumer.java
sonalidhome: Loop is now revised as suggested.

server/src/main/java/org/candlepin/model/Consumer.java
sonalidhome: done.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -893,13 +915,71 @@ public void setContentAccessMode(String contentAccessMode) {
...
+
+        for (CloudProfileFacts profileFact : CloudProfileFacts.values()) {
+
+            if (incomingFacts.containsKey(profileFact.getFact())) {
+
+                Map<String, String> filteredIncomingFact = incomingFacts.entrySet().stream()
+                    .filter(key -> key.getKey().equals(profileFact.getFact()))
+                    .collect(HashMap::new, (m, v)->m.put(v.getKey(), v.getValue()), HashMap::putAll);
+
+                for (Entry<String, String> entry : filteredIncomingFact.entrySet()) {
```
nikosmoum, line 968: I'm not sure why either the filtering (which will still do a 200+ iterations loop) or this second loop is needed?  At this point we already know the key we are interested is a cloud-profile fact and is contained in the incoming facts, and only need to check the Consumer's side.

Also, the whole purpose of using a HashMap (or HashSet), is that you get O(1) retrieval time, so you only need to call `containsKey` to check if a key is there, or `.get('key')` to retrieve the value. Looping over a HashMap is almost never needed. 

Thus, this whole method could be simplified to this:
```
    public boolean checkForCloudProfileFacts(Map<String, String> incomingFacts) {
        if (incomingFacts == null) {
            return false;
        }

        if (this.facts != null && this.facts.equals(incomingFacts)) {
            return false;
        }

        for (CloudProfileFacts profileFact : CloudProfileFacts.values()) {
            if (incomingFacts.containsKey(profileFact.getFact())) {
                if (this.facts == null ||
                    !this.facts.containsKey(profileFact.getFact()) ||
                    !this.facts.get(profileFact.getFact()).equals(incomingFacts.get(profileFact.getFact()))) {
                    return true;
                }
            }
        }

        return false;
    }
```

sonalidhome: I misunderstood the behavior of incoming facts. Thanks for explaining. 
I have corrected it as suggested.

PR #2698, opened 2020-05-12
[F] Removed unnecessary code and added missing refresh tests (ENT-2207)

Ceiu:
- Added versioning tests for the full RefreshWorker execute
  code path
- Removed ContentManager.importContent, as it is entirely
  obsoleted by the RefreshWorker framework
- Deprecated the ProductManager and ContentManager classes
  as a whole

### DIFF
--- a/server/src/main/java/org/candlepin/controller/ContentManager.java
+++ b/server/src/main/java/org/candlepin/controller/ContentManager.java
@@ -21,17 +21,13 @@ import org.candlepin.dto.api.v1.ProductDTO.ProductContentDTO;
 import org.candlepin.model.Content;
 import org.candlepin.model.ContentCurator;
 import org.candlepin.model.Owner;
-import org.candlepin.model.OwnerContent;
 import org.candlepin.model.OwnerContentCurator;
 import org.candlepin.model.Product;
-import org.candlepin.model.ProductContent;
 import org.candlepin.model.ProductCurator;
 import org.candlepin.model.dto.ContentData;
 import org.candlepin.model.dto.ProductContentData;
 import org.candlepin.model.dto.ProductData;
 import org.candlepin.service.model.ContentInfo;
-import org.candlepin.util.Traceable;
-import org.candlepin.util.TraceableParam;
 import org.candlepin.util.Util;
 
 import com.google.inject.Inject;
@@ -46,7 +42,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/candlepin/controller/ProductManager.java
```diff
@@ -62,7 +62,13 @@
  * <p></p>
  * The methods provided by this class are the prefered methods to use for CRUD operations on
  * products, to ensure product versioning and linking is handled properly.
+ *
+ * @deprecated
+ *  This class's primary purpose has been replaced by the RefreshWorker and its various components.
+ *  New code should avoid using this class and its methods if at all possible, instead opting to
+ *  use the RefreshWorker.
  */
+@Deprecated
```
nikosmoum, line 71: Same question as ContentManager

server/src/test/java/org/candlepin/controller/PoolManagerTest.java
```diff
@@ -579,76 +579,6 @@ public RefreshResult answer(InvocationOnMock iom) throws Throwable {
         .when(this.refreshWorker).execute(eq(owner));
     }
 
-    // private void mockProductImport(Owner owner, final Map<String, Product> products) {
```
nikosmoum, line 582: This file also has commented out calls of these methods throughout many tests like the following, which should be removed:
```
// this.mockProductImport(owner, product);
// this.mockContentImport(owner, new Content[] {});
```

server/src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -60,7 +55,13 @@
  * <p></p>
  * The methods provided by this class are the prefered methods to use for CRUD operations on
  * content, to ensure content versioning and linking is handled properly.
+ *
+ * @deprecated
+ *  This class's primary purpose has been replaced by the RefreshWorker and its various components.
+ *  New code should avoid using this class and its methods if at all possible, instead opting to
+ *  use the RefreshWorker.
```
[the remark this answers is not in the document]
Ceiu: There are a few more improvements overall to do (such as adding deletion functionality), but aside from that, the CRUD operations on products and content is basically "change an entity and then update everything above it", most of which is just duplicating what the RefreshWorker does.

At the end of the card I created for this, I noted that if we opt to keep them around, they should be nothing more than wrappers around the RefreshWorker setup and execution.

server/src/test/java/org/candlepin/controller/PoolManagerTest.java
Ceiu: Fixed

server/src/main/java/org/candlepin/controller/ContentManager.java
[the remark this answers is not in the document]
Ceiu: I've removed the deprecation annotation from ContentManager and moved ProductManager's onto importProducts, but this is something that needs to be cleaned up eventually. It's almost entirely dead code at this point.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -376,47 +382,67 @@ private DeveloperProducts getDeveloperPoolProducts(Owner owner, String sku) {
...
-                });
-            }
-        }
+            // Do a refresh so we're all up to date here
+            log.debug("Importing products for dev pool resolution...");
 
-        log.debug("Importing products for dev pool resolution...");
-        ImportedEntityCompiler compiler = new ImportedEntityCompiler();
+            RefreshWorker refresher = this.refreshWorkerProvider.get();
+            refresher.addProducts(this.productAdapter.getProductsByIds(owner.getKey(), devProductIds));
```
nikosmoum, line 402: This is not directly related to your changes, but rather the previous changes in this method.
If the initial call of `productAdapter.getProductsByIds(owner.getKey(), Arrays.asList(sku))` returns the marketing product we want, I assume that when Teamnado are going to implement our current ProductInfo interface which has the new `getProvidedProducts()` method on it, then that means we already have the whole product hierarchy fetched (well, for now a single mkt product on level 0 and some eng products on level 1). Why do we need to call the adapter a second time for all the children of `sku`?

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -376,47 +382,67 @@ private DeveloperProducts getDeveloperPoolProducts(Owner owner, String sku) {
...
+            RefreshResult refreshResult = refresher.execute(owner);
 
-        compiler.addProducts(this.productAdapter.getProductsByIds(owner.getKey(), devProductIds));
+            // Step through the items we refreshed and add the resulting products to our map
+            for (String pid : refresher.getProducts().keySet()) {
+                Product product = refreshResult.getProduct(pid);
+                if (product != null) {
+                    devProductMap.put(product.getId(), product);
```
nikosmoum, line 409: You mentioned yesterday that because of how the RefreshWorker works, we're adding all the org's products to the `devProductMap`. Can we not do some filtering here, and only add the refreshed products that are part of `sku` (and its hierarchy below it)?

server/src/main/java/org/candlepin/controller/Entitler.java
Ceiu: That's a very valid point.

Today this path is still necessary to get the overall code working, since our default product service adapter hasn't been ported to the new model yet. But it would be a lot nicer to not need to do a fetch->collect->fetch step.

server/src/main/java/org/candlepin/controller/Entitler.java
Ceiu: I don't fully understand the question.

This block is what is doing that filtering. Previously we were relying on the poorly named .getImportedProducts to grab everything, but with the new system, we need to be a little more selective to get the correct result.

Part of the issue here is how DeveloperProduct works, and still expects to be working with a flat map. I feel there are going to be many places like this where we have to make a decision to either update a larger area of code to be more n-tier aware, or have patches like this which allow those places to continue working as they do.

nikosmoum: Nevermind, I misread the code!

PR #2701, opened 2020-05-14
ENT-1863: Port ProductCertificateDTO to openapi spec

sonalidhome:
- Added specification in yaml file
 - Translator and its test changes


api/candlepin-api-spec.yaml
```diff
@@ -2501,6 +2501,17 @@ components:
         enabled:
           type: boolean
 
+    ProductCertificateDTO:
+      description: DTO representing product certificates presented to the API
+      allOf:
+        - $ref: '#/components/schemas/TimestampedEntity'
+        - type: object
+          properties:
```
wolfdale, line 2509: Not sure why `ProductDTO` & `Id` are not included in properties.

sonalidhome: Getter and Setter of Id have ignored with `@JsonIgnore` annotation. Hence removed as it is not exposed over API.
Similarly, the Product is also not exposed over API. Maybe I can keep `writeOnly` for Product in properties to set the Product.

sonalidhome: As per the discussion, ignoring `ProductDTO` in properties.

sonalidhome:
Thanks for noticing the missing mixin entry. Added now.

PR #2702, opened 2020-05-15
ENT-1847: Port ActivationKeyResource to spec-first

wolfdale:
- Updated Candlepin Schema file
- Updated ActivationKeyResource API

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -30,6 +30,325 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /activation_keys:
+    get:
+      description: Retrieves a list of activation keys
+      tags:
+        - Activation Key
+      operationId: findActivationKey
+      x-java-response:
+        type: Iterable
+        isContainer: true
+      security: []
+      responses:
+        200:
+          description: List of activation keys
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/ActivationKeyDTO'
+        default:
+          $ref: '#/components/responses/default'
+
+  /activation_keys/{activation_key_id}:
+    get:
+      description: Retrieves a single activation key
+      tags:
+        - Activation Key
+      operationId: getActivationKey
+      security: []
+      parameters:
+        - name: activation_key_id
+          in: path
+          description: The ID of the activation key to retrieve
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: Activation key
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ActivationKeyDTO'
+        400:
+          description: Activation key ID is null, empty or could not be found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ExceptionMessage'
+              example:
+                displayMessage: Activation key ID is null, empty or could not be found
+                requestUuid: c4347004-8792-41fe-a4d8-fccaa0d3898a
+        default:
+          $ref: '#/components/responses/default'
+
+    put:
+      description: Updates an Activation Key
+      tags:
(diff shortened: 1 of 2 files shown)


api/candlepin-api-spec.yaml
```diff
@@ -30,6 +30,336 @@ paths:
...
+    get:
+      description: Retrieves a single activation key
+      tags:
+        - Activation Key
+      operationId: getActivationKey
+      security: []
+      parameters:
+        - name: activation_key_id
+          in: path
+          description: The Id of activation key to retrieve
```
Ceiu, line 65: ID should be capitalized when used in a description or display message.

api/candlepin-api-spec.yaml
```diff
@@ -30,6 +30,336 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/ActivationKeyDTO'
+        400:
+          description: Activation key Id is null or empty or could not be found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ExceptionMessage'
+              example:
+                displayMessage: Activation key Id is null or empty or could not be found
```
Ceiu, line 83: There should be a comma somewhere in that list at the end. Any of the following would work, depending on your preference for the Oxford comma, and the intended coupling of "null" and "empty":

Activation key ID is null, empty, or could not be found
Activation key ID is null, empty or could not be found
Activation key ID is null or empty, or could not be found

api/candlepin-api-spec.yaml
```diff
@@ -30,6 +30,336 @@ paths:
...
+              $ref: '#/components/schemas/ActivationKeyDTO'
+      responses:
+        200:
+          description: Updated activation key
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ActivationKeyDTO'
+        400:
+          description: Validation failure, Null or Empty or Unknown Activation key Id. Either service level
```
Ceiu, line 116: Lots of unnecessary capitalization here: "Null," "Empty," "Unknown," and "Activation key" are not proper nouns and should not be capitalized. The service level/activation key name validation failure at the end is also a bit confusing, as it seems like it should be a separate message entirely with the way it's written.

I'm assuming this is intending to collect information about every possible validation failure on the key as a whole, but that's likely going to date itself eventually. We should probably just focus on expressing that the validation failed somehow, and leave the details to the exception.

api/candlepin-api-spec.yaml
```diff
@@ -30,6 +30,336 @@ paths:
...
+            format: int64
+      responses:
+        200:
+          description: Pool added to activation key
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ActivationKeyDTO'
+        400:
+          description: Activation key Id is null or empty or either pool or activation key could not be
```
Ceiu, line 228: This feels like another collection of all possible BadRequestExceptions. In this case it's better to be less explicit about any possible error (since that doesn't tell you anything anyway), and just focus on the type of issue it could be: invalid input or the pool could not be added.

Though, honestly, this is a case where maybe we consider changing the exception type and http code for the rules exception, as those really aren't bad requests; but that's a separate discussion that would involve Katello.

api/candlepin-api-spec.yaml
wolfdale: Done.

api/candlepin-api-spec.yaml
wolfdale: Done.

api/candlepin-api-spec.yaml
wolfdale: Simplified the description.

api/candlepin-api-spec.yaml
wolfdale: Done

PR #2703, opened 2020-05-18
[M] Corrected case in activation key resource

AmitMandliya:


Ceiu:
ok to test

server/src/main/java/org/candlepin/resource/ActivationKeyResource.java
```diff
@@ -111,13 +111,13 @@ public ActivationKeyResource(ActivationKeyCurator activationKeyCurator, I18n i18
...
         if (keyId == null || keyId.isEmpty()) {
-            throw new BadRequestException(i18n.tr("activation key ID is null or empty"));
+            throw new BadRequestException(i18n.tr("Activation key ID is null or empty"));
         }
 
         ActivationKey key = this.activationKeyCurator.secureGet(keyId);
 
         if (key == null) {
-            throw new BadRequestException(i18n.tr("ActivationKey with id {0} could not be found.", keyId));
+            throw new BadRequestException(i18n.tr("Activation key with id {0} could not be found.", keyId));
```
Ceiu, line 120: Let's also correct the capitalization on ID here, and properly quote the parameter:

```Activation key with ID \"{0}\" could not be found```

PR #2707, opened 2020-05-22
1836973 - Listing pools available for an activation_key lists pools

sonalidhome:
already in the key.

 - Fixed the condition to add pool into filtered pools
 - Added a spec test to validate the activation key filter is working

### DIFF
--- a/server/spec/owner_resource_spec.rb
+++ b/server/spec/owner_resource_spec.rb
@@ -793,6 +793,17 @@ describe 'Owner Resource Pool Filter Tests' do
     expect(poolIds).to include(pools[1].id)
   end
 
+  it "lets owners filter pools by activation_key" do
+    activation_key = @cp.create_activation_key(@owner['key'], random_string('test_key'))
+
+    @cp.add_pool_to_key(activation_key['id'], @pools[0].id)
+    @cp.add_pool_to_key(activation_key['id'], @pools[1].id)
+
+    # ignores the pools already with the activation key
+    pools = @cp.list_owner_pools(@owner['key'],{ :activation_key => activation_key['name']})
+    pools.length.should == 1
+  end
+
 end
 
 describe 'Owner Resource Consumer Fact Filter Tests' do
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2605,7 +2605,7 @@ public void updatePoolsFromStackWithoutDeletingStack(Consumer consumer, List<Poo
         List<Pool> filteredPools = new LinkedList<>();
         for (Pool p : pools) {
             ValidationResult result = activationKeyRules.runPoolValidationForActivationKey(key, p, null);
-            if (result.isSuccessful() && (!result.hasWarnings() || includeWarnings)) {
+            if (!result.hasErrorKeys() && (!result.hasWarnings() || includeWarnings)) {
```
nikosmoum, line 2608: The issue here was that a new piece of data (`errorKeys`) was added to ValidationResult in addition to `errors`, which is not taken into account by the `isSuccessful` method. This will work, but I think to future-proof the use of the `isSuccessful` method (which is used elsewhere, and in the future when we move more code from rules.js to Java, the same bug will appear there too), we should leave this check as it was and simply alter ValidationResult.isSuccessful to more correctly check both errors and errorKeys as such: `return !hasErrors() && !hasErrorKeys();`

sonalidhome: Done.

Ceiu: Is there a reason we don't just fix ```isSuccessful``` and any places that break as a consequence of that fix? Seems like a trap to leave a method like that around that doesn't check what its name implies.

sonalidhome: The method `isSuccessful()` is being updated with `return !hasErrors() && !hasErrorKeys();`. Also the above change in `CandlepinPoolManager.java` is reverted. Is there anything else that needs to be done?

PR #2708, opened 2020-05-22
1836988: Fix for datetime parsing error with bad TZ offset (ENT-2379)

wolfdale:
Fixes 
- Resteasy 4.2.2 now throws `javax.ws.rs.NotFoundException` which needed to be mapped to BAD_REQUEST via candlepin NotFoundExceptionMapper.


wolfdale:
Thanks, it is much simpler & elegant.
Candlepin 3.1.3 or below use to throw `"displayMessage": "Unable to parse date parameter"` I went through all the trouble to keep the exception message consistent with older candlepin.

nikosmoum:
Yes I thought so, but the important part is to keep the HTTP code consistent. Also the message that shows which value is wrong for what field is clearer and since it was already part of the code and thrown for other, non-date fields, it is better to use it

PR #2712, opened 2020-05-27
ENT-1880: Port OwnerProductResource to spec-first

wolfdale:
- Updated schema file
- Updated OwnerProductResource API


wolfdale:
@nikosmoum  @Ceiu @Januson  There is one issue, while porting `OwnerContentResource` & `OwnerProductResource` `(& OwnerResource)` the generated `OwnersApi` interface contains all the APIs (since all of them starts with `/owner/`) of `OwnerContentResource & OwnerProductResource (& OwnerResource)` making it difficult to implement the interface to respective classes as it enforces the rule to implement all the methods. Currently looking into alternatives to handle this.

nikosmoum:
It doesn't look like openapi supports splitting the url paths into more than one generated interface, so the best course of action would be to merge the 3 *Resource files into one (OwnerResource.java), which would be preferable to another custom codegen or template change.

Ceiu:
At the very least, we should probably put in an RFE to the OpenAPI project to allow specifying the interface for a given endpoint. I'm not a huge fan of those getting rolled into the owners interface, as it's already pretty bloated as it is.

wolfdale:
@nikosmoum Made some new changes to the PR. 
Can you please take a look at this PR again.

wolfdale:
Re-based the PR to resolve conflicts.

PR #2714, opened 2020-05-27
[M] 1840850: Fixed parameter overflow in bulk entitlement deletion (ENT-2437)

Ceiu:
- Fixed an issue during entitlement deletion that could cause a
  query to be generated that attempts to use more parameters than
  the backend database supports
- Flagged the bulkSQLDelete method as deprecated, as it lacks
  protection for the database parameter limit

### DIFF
--- a/server/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/server/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -796,11 +796,25 @@ public class EntitlementCurator extends AbstractHibernateCurator<Entitlement> {
      * preferably by calling {@link #unlinkEntitlements} first.
      *
      * @param entitlementIds The collection of ids of the entitlements to be deleted.
+     *
+     * @return
+     *  the number of entitlements deleted as a result of this operation
      */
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -796,11 +796,25 @@ public void batchDelete(Collection<Entitlement> entitlements) {
...
      * @param entitlementIds The collection of ids of the entitlements to be deleted.
+     *
+     * @return
+     *  the number of entitlements deleted as a result of this operation
      */
-    public void batchDeleteByIds(Collection<String> entitlementIds) {
-        Map<String, Object> criteria = new HashMap<>();
-        criteria.put("id", entitlementIds);
-        this.bulkSQLDelete(Entitlement.DB_TABLE, criteria);
+    public int batchDeleteByIds(Collection<String> entitlementIds) {
```
Januson, line 803: Could you add a unit-test for this change? I think it would be helpful to know that we do not re-introduce the issue in case we need to make changes in the future.

Ceiu: Done

PR #2715, opened 2020-05-28
1834762: crl file candlepin-crl.crl is not found on the candlepin server

Januson:
- Added a check to initialize the CRL file if it is missing during the Candlepin startup

### DIFF
--- a/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
+++ b/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2009 - 2012 Red Hat, Inc.
+ * Copyright (c) 2009 - 2020 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -38,6 +38,7 @@ import org.candlepin.messaging.CPMContextListener;
 import org.candlepin.pki.impl.JSSProviderLoader;
 import org.candlepin.resteasy.AnnotationLocator;
 import org.candlepin.swagger.CandlepinSwaggerModelConverter;
+import org.candlepin.util.CrlFileUtil;
 import org.candlepin.util.Util;
 
 import com.google.inject.AbstractModule;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -224,6 +229,20 @@ private void initializeSubsystems(Injector injector) throws Exception {
...
+        // Init CRL file
+        this.crlFileUtil = injector.getInstance(CrlFileUtil.class);
+        String filePath = getCrlFilePath();
+        File crlFile = new File(filePath);
+        if (!crlFile.exists() || crlFile.length() == 0) {
+            try {
+                this.crlFileUtil.initializeCRLFile(crlFile, Collections.emptyList());
+            }
+            catch (IOException e) {
+                e.printStackTrace();
```
Ceiu, line 241: We should write these to the log and probably rethrow the exception as well.

server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -382,4 +401,15 @@ private void insertValidationEventListeners(Injector injector) {
         registry.getEventListenerGroup(EventType.PRE_UPDATE).appendListener(listenerProvider.get());
         registry.getEventListenerGroup(EventType.PRE_DELETE).appendListener(listenerProvider.get());
     }
+
+
+    private String getCrlFilePath() {
+        String filePath = config.getString(ConfigProperties.CRL_FILE_PATH);
+
+        if (filePath == null) {
+            throw new IseException("CRL file path not defined in config file");
```
Ceiu, line 410: IseException is a REST exception. We should throw something like a ConfigurationException here instead.

Januson: Done.

server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
Januson: Done.

server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -224,6 +228,21 @@ private void initializeSubsystems(Injector injector) throws Exception {
...
+        this.crlFileUtil = injector.getInstance(CrlFileUtil.class);
+        String filePath = getCrlFilePath();
+        File crlFile = new File(filePath);
+        if (!crlFile.exists() || crlFile.length() == 0) {
+            try {
+                this.crlFileUtil.initializeCRLFile(crlFile, Collections.emptyList());
+            }
+            catch (IOException e) {
+                log.error("Error occurred during initialization of CRL file!", e);
+                throw new ConfigurationException("Error occurred during initialization of CRL file!", e);
```
Ceiu, line 241: This one should stay as an IOException, just rethrown.

Januson: Done.

PR #2723, opened 2020-06-08
ENT-2440: Map internal async job states to a subset of client-supported states

Januson:


server/src/main/java/org/candlepin/dto/api/v1/AsyncJobStatusTranslator.java
```diff
@@ -72,14 +71,54 @@ public AsyncJobStatusDTO populate(ModelTranslator translator, AsyncJobStatus sou
...
+            case QUEUED:
+            case RUNNING:
+            case SCHEDULED:
+            case WAITING:
+                publicJobState = PublicJobState.RUNNING;
+                break;
+            case ABORTED:
+            case CANCELED:
+            case FAILED:
+            case FAILED_WITH_RETRY:
```
Ceiu, line 108: I'm torn on this one. I can see a temporary "FAILED" output being useful, but it's still kind of "RUNNING" since it'll get picked up again unless something breaks internally.

server/src/main/java/org/candlepin/dto/api/v1/AsyncJobStatusTranslator.java
```diff
@@ -72,14 +71,54 @@ public AsyncJobStatusDTO populate(ModelTranslator translator, AsyncJobStatus sou
...
+        }
+        PublicJobState publicJobState;
+        switch (state) {
+            case FINISHED:
+                publicJobState = PublicJobState.FINISHED;
+                break;
+            case CREATED:
+                publicJobState = PublicJobState.CREATED;
+                break;
+            case QUEUED:
```
Ceiu, line 99: Should QUEUED be part of RUNNING or CREATED? This is another one where I can see arguments for either case.

server/spec/job_status_spec.rb
```diff
@@ -123,18 +123,18 @@
       job = @user.autoheal_org(@owner['key'])
       #make sure we see a job waiting to go
       joblist = @cp.list_jobs(@owner['key'])
-      expect(joblist.find { |j| j['id'] == job['id'] }['state']).to eq('QUEUED')
+      expect(joblist.find { |j| j['id'] == job['id'] }['state']).to eq('RUNNING')
 
       @user.cancel_job(job['id'])
       #make sure we see a job canceled
       joblist = @cp.list_jobs(@owner['key'])
-      expect(joblist.find { |j| j['id'] == job['id'] }['state']).to eq('CANCELED')
```
Ceiu, line 131: As mentioned in IRC, let's hang on to the CANCELED state, since it doesn't fit into either of the two completion states from the request, and it was preexisting so we won't break clients.

server/src/main/java/org/candlepin/dto/api/v1/AsyncJobStatusTranslator.java
Januson: I haven't thought of that. It makes sense to map it to failed only once it reaches terminal state. Done.

server/src/main/java/org/candlepin/dto/api/v1/AsyncJobStatusTranslator.java
Januson: I guess you are right. Queued job is not running yet so we can merge the state into the CREATED instead of RUNNING. Done.

server/spec/job_status_spec.rb
Januson: Done.

PR #2728, opened 2020-06-18
[Bug 1844258]: Job Manager Logging Format Changes for Splunk (ENT-2480)

wolfdale:
- Added the job key to the log context (MDC)
- Changes related to log format.

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -93,6 +93,7 @@ public class JobManager implements ModeChangeListener {
 
     private static final String MDC_REQUEST_TYPE_KEY = "requestType";
     private static final String MDC_REQUEST_UUID_KEY = "requestUuid";
+    private static final String MDC_JOB_KEY_KEY = "jobKey";
     private static final String MDC_LOG_LEVEL_KEY = "logLevel";
 
     private static final String QRTZ_GROUP_CONFIG = "cp_async_config";
@@ -1191,28 +1192,28 @@ public class JobManager implements ModeChangeListener {
             return status;
         }
 
-        final Class<? extends AsyncJob> jobClass = getJobClass(status.getJobKey());
-
-        // Maybe in this case it'd be better to attempt to use the job key as the job class
-        // rather than failing directly. This would allow use of aliases and explicit job
-        // classes.
-        if (jobClass == null) {
-            String errmsg = String.format("No registered job class for job: %s", status.getJobKey());
-
-            this.updateJobStatus(status, JobState.FAILED, errmsg);
-
-            log.error(errmsg);
-            throw new JobInitializationException(errmsg, true);
-        }
-
         try {
             this.setupJobRuntimeEnvironment(status);
+
+            // Maybe in this case it'd be better to attempt to use the job key as the job class
+            // rather than failing directly. This would allow use of aliases and explicit job
+            // classes.
+            Class<? extends AsyncJob> jobClass = getJobClass(status.getJobKey());
+
+            if (jobClass == null) {
+                String errmsg = String.format("No registered job class for job: %s", status.getJobKey());
+
+                this.updateJobStatus(status, JobState.FAILED, errmsg);
+
+                log.error(errmsg);
+                throw new JobInitializationException(errmsg, true);
+            }
 
             EventSink eventSink = injector.getInstance(EventSink.class);
             AsyncJob job = injector.getInstance(jobClass);
 
             if (job == null) {
-                String errmsg = String.format("Unable to instantiate job class \"{}\" for job: {}",
+                String errmsg = String.format("Unable to instantiate job class \"%s\" for job: %s",
                     jobClass.getName(), status.getJobKey());
 
                 log.error(errmsg);
@@ -1292,6 +1293,7 @@ public class JobManager implements ModeChangeListener {
 
         MDC.put(MDC_REQUEST_TYPE_KEY, "job");
         MDC.put(MDC_REQUEST_UUID_KEY, status.getId());
+        MDC.put(MDC_JOB_KEY_KEY, status.getJobKey());
 
         // Inject all of our metadata...
         for (Map.Entry<String, String> entry : status.getMetadata().entrySet()) {
--- a/server/src/main/resources/logback.xml
+++ b/server/src/main/resources/logback.xml
@@ -20,7 +20,7 @@
     <appender name="CandlepinAppender" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${LOG_DIR}/candlepin.log</file>
         <encoder>
-            <pattern>%d{ISO8601} [thread=%thread] [%X{requestType}=%X{requestUuid}, org=%X{org}, csid=%X{csid}] %-5p %c - %m%n</pattern>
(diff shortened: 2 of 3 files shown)


server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1293,6 +1295,9 @@ private void setupJobRuntimeEnvironment(AsyncJobStatus status) {
         MDC.put(MDC_REQUEST_TYPE_KEY, "job");
         MDC.put(MDC_REQUEST_UUID_KEY, status.getId());
 
+        MDC.put(MDC_REQUEST_SUB_TYPE_KEY, "jobKey");
```
Ceiu, line 1298: For the displayed value, we want to use snake case, so "job_key" here is more appropriate.

server/src/main/resources/logback.xml
```diff
@@ -20,7 +20,7 @@
     <appender name="CandlepinAppender" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${LOG_DIR}/candlepin.log</file>
         <encoder>
-            <pattern>%d{ISO8601} [thread=%thread] [%X{requestType}=%X{requestUuid}, org=%X{org}, csid=%X{csid}] %-5p %c - %m%n</pattern>
+            <pattern>%d{ISO8601} [thread=%thread] [%X{requestType}=%X{requestUuid}, %X{requestSubType}=%X{requestName}, org=%X{org}, csid=%X{csid}] %-5p %c - %m%n</pattern>
```
Ceiu, line 23: This is going to result in a floating "=, " for the majority of log entries. If we're going to use this method for omitting the entry when it's not present, we also need to do that for the separators as well, which is a bit hack.

We should look for a better option here, such that we can avoid printing the whole thing if the target variable is empty. All else failing, we can update the logging format here to specify the value separator and the keypair separator, and then inject them ourselves as suggested above.

Ceiu: Looks like ```%equals{job_key=%X{jobKey}, }{job_key=, }{}``` would omit the job key printing if the value is not present in the MDC. This would also allow us to drop the request name bit as well, unless there's some other intended use for that.

Ceiu: Scratch that; wrong logging format. This works, however:
```%replace(job_key=%X{jobKey}, ){'job_key=, ', ''}```

File: server/src/main/java/org/candlepin/async/JobManager.java (hunk 1 above)
Comment (wolfdale) [reply server/src/main/java/org/candlepin/async/JobManager.java:1298]: Dropped this key from here.

File: server/src/main/resources/logback.xml (hunk 2 above)
Comment (wolfdale) [reply server/src/main/resources/logback.xml]: Replacement is working correctly for job key field. 
Leaving other as it is (`org=, csid=`).

PS: Even a slightest mistake in Logback patterns will make your logs disappears. Learnt this the hard way :(

File: server/src/main/java/org/candlepin/async/JobManager.java [hunk 3]
```diff
@@ -1197,7 +1198,7 @@ public AsyncJobStatus executeJob(JobMessage message) throws JobException {
         // rather than failing directly. This would allow use of aliases and explicit job
         // classes.
         if (jobClass == null) {
-            String errmsg = String.format("No registered job class for job: %s", status.getJobKey());
+            String errmsg = String.format("No registered job class for jobKey=%s", status.getJobKey());
```
Ceiu, line 1201: Looks like a refactor artifact here.

server/src/main/resources/logback.xml
```diff
@@ -20,7 +20,7 @@
     <appender name="CandlepinAppender" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${LOG_DIR}/candlepin.log</file>
         <encoder>
-            <pattern>%d{ISO8601} [thread=%thread] [%X{requestType}=%X{requestUuid}, org=%X{org}, csid=%X{csid}] %-5p %c - %m%n</pattern>
+            <pattern>%d{ISO8601} [thread=%thread] [%X{requestType}=%X{requestUuid}, %replace(job_key=%X{jobKey}, ){'job_key=, ', ''} org=%X{org}, csid=%X{csid}] %-5p %c - %m%n</pattern>
```
Ceiu, line 23: No space between the final ```}``` and the ```org=```. If we have it in there, you get a double space between org and the preceding value.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -93,6 +93,7 @@
 
     private static final String MDC_REQUEST_TYPE_KEY = "requestType";
     private static final String MDC_REQUEST_UUID_KEY = "requestUuid";
+    private static final String MDC_REQUEST_JOB_KEY = "jobKey";
```
Ceiu, line 96: ```MDC_JOB_KEY_KEY```

The name sounds a bit silly and redundant, but it is the key for the job key.

server/src/main/resources/logback.xml
Ceiu: Yeah, leaving the others is fine for now. I'm poking people on the value of conditionally removing them, and we can make a decision on that front later.

server/src/main/java/org/candlepin/async/JobManager.java
wolfdale: Changed it to `job_key`.

server/src/main/resources/logback.xml
wolfdale: fixed.

server/src/main/java/org/candlepin/async/JobManager.java
wolfdale: Done.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1197,7 +1198,7 @@ public AsyncJobStatus executeJob(JobMessage message) throws JobException {
         // rather than failing directly. This would allow use of aliases and explicit job
         // classes.
         if (jobClass == null) {
-            String errmsg = String.format("No registered job class for job: %s", status.getJobKey());
+            String errmsg = String.format("No registered job class for job_key=%s", status.getJobKey());
```
Ceiu, line 1201: I'm guessing the intent here is to make this particular line searchable as part of the same log query? If so, changing the message isn't the way to go here. This block can safely live in the try block below immediately following the ```setupJobRuntimeEnvironment``` call (with proper whitespace).

I also noticed a bug unrelated to your code in the log message below regarding job instantiation. We can fix that here as well:

```
        try {
            this.setupJobRuntimeEnvironment(status);

            // Maybe in this case it'd be better to attempt to use the job key as the job class
            // rather than failing directly. This would allow use of aliases and explicit job
            // classes.
            Class<? extends AsyncJob> jobClass = getJobClass(status.getJobKey());
            if (jobClass == null) {
                String errmsg = String.format("No registered job class for job: %s", status.getJobKey());

                this.updateJobStatus(status, JobState.FAILED, errmsg);

                log.error(errmsg);
                throw new JobInitializationException(errmsg, true);
            }

            EventSink eventSink = injector.getInstance(EventSink.class);
            AsyncJob job = injector.getInstance(jobClass);

            if (job == null) {
                String errmsg = String.format("Unable to instantiate job class \"%s\" for job: %s",
                    jobClass.getName(), status.getJobKey());

                log.error(errmsg);
                throw new JobInitializationException(errmsg);
            }
```

wolfdale: Done.

PR #2730, opened 2020-06-22
[M] 1844243: Added Quartz trigger misfire logging (ENT-2479)

Ceiu:
- The JobManager now logs Quartz trigger misfire events as they occur

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -53,6 +53,7 @@ import org.quartz.SchedulerFactory;
 import org.quartz.Trigger;
 import org.quartz.TriggerBuilder;
 import org.quartz.TriggerKey;
+import org.quartz.TriggerListener;
 import org.quartz.impl.matchers.GroupMatcher;
 import org.quartz.spi.JobFactory;
 import org.quartz.spi.TriggerFiredBundle;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -168,6 +169,37 @@ public void execute(org.quartz.JobExecutionContext context) throws org.quartz.Jo
...
+
+            log.warn("Trigger misfired for job: {} [start: {}, end: {}, next fire time: {}, " +
+                "final fire time: {}, priority: {}, misfire instruction: {}]",
+                jobKey, trigger.getStartTime(), trigger.getEndTime(), trigger.getNextFireTime(),
+                trigger.getFinalFireTime(), trigger.getPriority(), trigger.getMisfireInstruction());
+        }
+
+        @Override
+        public boolean vetoJobExecution(Trigger trigger, org.quartz.JobExecutionContext context) {
+            return false;
```
wolfdale, line 201: Do you think we should add a condition to check if Candlepin is in SUSPEND mode and not allow job to be executed.

Ceiu: Nah. Suspend mode is already handled by pausing both the scheduler and the message receiver. We don't need explicit handling here.

PR #2732, opened 2020-06-25
ENT-1856: Port ConsumerDTO to openapi spec

sonalidhome:
- Udpated specifications for CapabilityDTO, ConsumerDTO,
   ConsumerInstalledProductDTO, HypervisorIdDTO
 - Added ConsumerDTOArrayElement to exclude the properties
   mentioned as HateoasArrayExclude
 - Updated the ConsumerTypeDeserializer to get appropriate
   consumer type based on incoming object/string


buildSrc/src/main/resources/templates/pojo.mustache
```diff
@@ -86,6 +86,14 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}} {{#seriali
 
     return this;
   }
+
+  public {{items.datatypeWithEnum}} get{{nameInCamelCase}}Value({{{items.datatypeWithEnum}}} {{name}}Key) {
```
Ceiu, line 90: Since this will be recognized as a getter, will this show up as an additional field to the serialized DTOs?

api/candlepin-api-spec.yaml
```diff
@@ -3992,6 +3992,196 @@ components:
...
+                type: string
+              type:
+                $ref: '#/components/schemas/ConsumerTypeDTO'
+              idCert:
+                allOf:
+                  - $ref: '#/components/schemas/CertificateDTO'
+              guestIds:
+                type: array
+                items:
+                  $ref: '#/components/schemas/GuestIdDTO'
```
nikosmoum, line 4081: I believe, since this GuestIdDTO is within a list, we should instead use the GuestIdDTOArrayElement here.
I am not 100% sure though, so before doing that, you could check what current master does (e.g. does a GET /consumer/{consumer_uuid} for a consumer that has guest ids that themselves have attributes return the attributes too?)

api/candlepin-api-spec.yaml
```diff
@@ -3992,6 +3992,196 @@ components:
...
+              href:
+                type: string
+
+    ConsumerDTOArrayElement:
+      description: Represents a consumer
```
nikosmoum, line 4086: should be something like `Represents a reduced view of a consumer (excluding data like facts and identify certificate)`

api/candlepin-api-spec.yaml
```diff
@@ -3992,6 +3992,196 @@ components:
...
+                type: boolean
+              annotations:
+                type: string
+              contentAccessMode:
+                type: string
+              type:
+                $ref: '#/components/schemas/ConsumerTypeDTO'
+              idCert:
+                allOf:
+                  - $ref: '#/components/schemas/CertificateDTO'
```
nikosmoum, line 4077: was `allOf` added here by accident?

server/src/main/java/org/candlepin/jackson/ConsumerTypeDeserializer.java
```diff
@@ -16,77 +16,69 @@
...
-                    parser.readValueAsTree(), parser.getCurrentLocation());
+            return parseValueNode(node);
+        }
+        else if (node.isObject()) {
+            log.debug("Processing {} as a containing object node.", fieldName);
+
+            TreeNode valueNode = node.path(fieldName);
+            if (valueNode.isMissingNode()) {
+                throw new CandlepinJsonProcessingException(
+                    "The field " + fieldName + " is missing from: " + node.asToken(),
```
nikosmoum, line 58: Why are we substituting the `fieldName` variable in all of these strings? It's a static variable with the value 'label' that never gets reassigned as far as I can see, so there's not point. Specifically about the exception messages, we should try to keep them the same (or similar) to before, since 'Unexpected consumer type format:' is much clearer than 'The field label is missing from: '

buildSrc/src/main/resources/templates/pojo.mustache
sonalidhome: This method will be generated for all auto-generated DTOs.

api/candlepin-api-spec.yaml
sonalidhome: We are setting empty guest details for every `Consumer` translation to `ConsumerDTO`. https://github.com/candlepin/candlepin/blob/3c74b22ea291221b8c33da0bd9a1ab4197d956ad/server/src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java#L176
Hence there is no need to update the `GuestIdDTOArrayElement` here.

api/candlepin-api-spec.yaml
sonalidhome: Updated.

api/candlepin-api-spec.yaml
sonalidhome: Yes, I was initially trying out an option `writeOnly: true`. Later removed but `allOf` remained to delete.

server/src/main/java/org/candlepin/jackson/ConsumerTypeDeserializer.java
sonalidhome: Agreed, Updated the exception messages.

server/src/main/java/org/candlepin/dto/api/v1/HypervisorIdTranslator.java
```diff
@@ -55,13 +55,17 @@ public HypervisorIdDTO populate(HypervisorId source, HypervisorIdDTO dest) {
...
 
-        dest.setId(source.getId() != null ? source.getId().toString() : null);
-        dest.setHypervisorId(source.getHypervisorId());
-        dest.setReporterId(source.getReporterId());
+        if (dest == null) {
+            throw new IllegalArgumentException("dest is null");
+        }
 
-        return dest;
+        return dest.id(source.getId() != null ? source.getId().toString() : null)
```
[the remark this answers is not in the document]
sonalidhome: Updated.

server/src/main/java/org/candlepin/dto/api/v1/ConsumerInstalledProductTranslator.java
```diff
@@ -58,16 +59,22 @@ public ConsumerInstalledProductDTO populate(ConsumerInstalledProduct source,
...
-        dest.setVersion(source.getVersion());
-        dest.setArch(source.getArch());
-        dest.setStatus(source.getStatus());
-        dest.setStartDate(source.getStartDate());
-        dest.setEndDate(source.getEndDate());
+        if (dest == null) {
+            throw new IllegalArgumentException("dest is null");
+        }
+
+        dest.id(source.getId())
```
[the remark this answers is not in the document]
sonalidhome: Corrected.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -408,7 +409,7 @@ public void sanitizeConsumerFacts(Consumer consumer) {
     @Produces(MediaType.APPLICATION_JSON)
     @Wrapped(element = "consumers")
     @SuppressWarnings("checkstyle:indentation")
-    public CandlepinQuery<ConsumerDTO> list(@QueryParam("username") String userName,
+    public CandlepinQuery<ConsumerDTOArrayElement> list(@QueryParam("username") String userName,
```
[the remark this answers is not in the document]
sonalidhome: Right, Updated the methods.

buildSrc/src/main/resources/templates/pojo.mustache
nikosmoum: I don't think it will be recognized as a getter, since it has accepts an argument

sonalidhome: hmm, I'll revert the template changes rather create a new method inside `ConsumerResource.java`. Template changes could difficult to keep track on. We'll have to update this in `OpenAPIGenerator`.

api/candlepin-api-spec.yaml
```diff
@@ -3992,6 +3992,195 @@ components:
             serial:
               $ref: '#/components/schemas/CertificateSerialDTO'
 
+    CapabilityDTO:
+      description: Represents a consumer capability
+      properties:
+        id:
+          type: string
+        name:
```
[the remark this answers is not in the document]
sonalidhome: Added.

api/candlepin-api-spec.yaml
```diff
@@ -3992,6 +3992,195 @@ components:
...
+
+    ConsumerInstalledProductDTO:
+      description: Represents consumer installed product details
+      allOf:
+          - $ref: '#/components/schemas/TimestampedEntity'
+          - type: object
+            properties:
+              id:
+                type: string
+              productId:
```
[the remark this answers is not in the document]
sonalidhome: Added.

api/candlepin-api-spec.yaml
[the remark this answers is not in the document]
sonalidhome: Added.

api/candlepin-api-spec.yaml
```diff
@@ -3992,6 +3992,195 @@ components:
...
+              contentAccessMode:
+                type: string
+              type:
+                $ref: '#/components/schemas/ConsumerTypeDTO'
+              idCert:
+                $ref: '#/components/schemas/CertificateDTO'
+              guestIds:
+                type: array
+                items:
+                  $ref: '#/components/schemas/GuestIdDTO'
```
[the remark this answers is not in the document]
sonalidhome: Added.

sonalidhome:
Added validations as suggested.

api/candlepin-api-spec.yaml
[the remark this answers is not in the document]
nikosmoum: It's not the `id` field that needs to be in `required:`, but the `guestId` one

api/candlepin-api-spec.yaml
```diff
@@ -3992,6 +3992,201 @@ components:
...
+              type: array
+              items:
+                $ref: '#/components/schemas/GuestIdDTOArrayElement'
+            href:
+              type: string
+
+    ConsumerInstalledProductDTO:
+      description: Represents consumer installed product details
+      required:
+        - productId
```
nikosmoum, line 4165: One more thing: the `required:` property is not recognized in this place, it has to be under the `allOf:` property (at the same level as `properties:`) for the annotations to be generated

api/candlepin-api-spec.yaml
```diff
@@ -4202,6 +4397,8 @@ components:
 
     GuestIdDTO:
       description: Represents a guest ID running on a virt host consumer
+      required:
```
nikosmoum, line 4400: Same here, `required:` needs to be at the same level as `properties:` is

api/candlepin-api-spec.yaml
```diff
@@ -4217,6 +4414,8 @@ components:
 
     GuestIdDTOArrayElement:
       description: Represents a guest ID running on a virt host consumer. Does not include the attributes field
+      required:
```
nikosmoum, line 4417: Same here, `required:` needs to be at the same level as `properties:` is

api/candlepin-api-spec.yaml

api/candlepin-api-spec.yaml

api/candlepin-api-spec.yaml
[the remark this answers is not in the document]
sonalidhome: corrected.

PR #2736, opened 2020-06-25
1847879: Artemis errors while trying to send events during hypervisor update job (ENT-2529)

wolfdale:
Changes -
- Added a check to open the session on-demand if it is closed for while queuing messages.
- Hardened `sendMessages` & `cancelMessages` methods to allow execution only if AMQ session is open.

### DIFF
--- a/server/src/main/java/org/candlepin/audit/EventSinkImpl.java
+++ b/server/src/main/java/org/candlepin/audit/EventSinkImpl.java
@@ -43,8 +43,6 @@ import java.util.List;
 
 import javax.inject.Inject;
 
-
-
 /**
  * EventSink - Queues events to be sent after request/job completes, and handles actual
  * sending of events on successful job or API request, as well as rollback if either fails.
@@ -249,6 +247,7 @@ public class EventSinkImpl implements EventSink {
                  * messages safely and the session is then ready to start over the next time
                  * the thread is used.
                  */
+                this.sessionFactory = sessionFactory;
                 session = sessionFactory.getEgressSession(true);
                 producer = session.createProducer(MessageAddress.DEFAULT_EVENT_MESSAGE_ADDRESS);
             }
@@ -260,6 +259,16 @@ public class EventSinkImpl implements EventSink {
 
         public void queueMessage(String eventString, Event.Type type, Event.Target target)
             throws ActiveMQException {
+            if (session.isClosed()) {
+                try {
+                    session = sessionFactory.getEgressSession(true);
+                    producer = session.createProducer(MessageAddress.DEFAULT_EVENT_MESSAGE_ADDRESS);
+                }
+                catch (Exception e) {
+                    log.error("Unable to open session while queuing messages", e);
+                    throw new RuntimeException(e);
+                }
+            }
 
             ClientMessage message = session.createMessage(ClientMessage.TEXT_TYPE, true);
             message.getBodyBuffer().writeNullableSimpleString(SimpleString.toSimpleString(eventString));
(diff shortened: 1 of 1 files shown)


Ceiu:
There are a couple of things to tackle in this area of the code:

1. This is still using Artemis (ActiveMQ) directly, which we shouldn't be doing anymore. This might be a good time to move it over to the CPM framework so we're not directly using Artemis anymore. There will be one hiccup with the getQueueInfo call, which will mandate some larger changes (additions to the CPM framework), so it might make sense to do that in a separate task.

1. While the problem was correctly identified, the solution of putting the onus on the caller to deal with the sessions is kind of a problem. Given the goal of the the event sender abstraction is to automatically handle all the session stuff, it seems like pushing that off to the callers is not the way to go. A better solution here for usability might be to keep the autoclosing behavior (as it does our session cleanup without needing explicit hooks), but have the queueMessage method open the session on-demand if it is either null or closed. The ```JobMessageDispatcher``` class uses this behavior to avoid the issue you're addressing.

1. This is somewhat personal preference, but when we're adding optional, binary functionality like this, I generally prefer to overload the method and expose the boolean. That way the method names don't change (making refactoring a tad easier), but you still get the new functionality in cases where you want it.

server/src/main/java/org/candlepin/audit/EventSinkImpl.java
```diff
@@ -127,8 +127,8 @@ public void queueEvent(Event event) {
 
         try {
             // Lazily initialize the message sender when the first
-            // message gets queued.
-            if (messageSender == null) {
+            // message gets queued or initialize session if it's closed.
+            if (messageSender == null || messageSender.session.isClosed()) {
```
Ceiu, line 131: This works, but is still breaking the encapsulation open to accomplish the goal. The purpose of the ```EventMessageSender``` is to be our abstraction layer for managing the session, so we should avoid spinning up a new one to do the work it should be doing. From the other change in this code, it appears you've already spotted the downside in doing it this way: you need to do the close check in every operation; making the session management exist in three places, rather than one.

However, this does highlight another minor problem with the existing code: EventMessageSender isn't static. The wrapper does not care about the state of its outer class, nor should it ever use it. We don't need the baggage an inner class inherits in such a case, and we should be declaring it statically.

wolfdale: Right, made some adjustments to it. 
Checks on `sendMessages` & `cancelMessages` methods are to prevent doing anything if we don't have session opened. With [this line](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/async/JobManager.java#L1286) inside `JobManager` which make sure that events are always sent out or rolled backed (in case of exception). Now with transnational wrapper specifically in HypervisorUpdateJob the AMQ session get closed, which leads to problem when control goes back to JobManager.

Ceiu:
I'd like to see slightly more abstracted session handling, but this will get the job done.

The last thing to do here is to throw it at a profiler to ensure the repeated opening and closing of sessions isn't going to add a noticeable amount of load to the job processing. If it is, we'll have to revert back to your original implementation. I don't predict it will be, but we'll let the data determine that.

There's a script in ```server/bin``` named ```gen_fake_report.py``` which can be used to generate a report file, and then the cpc utility can be used to fire off an update job:

```./client/ruby/cpc hypervisor_update_file snowwhite test_report.json true 'reporter_id'```

I'll pick this up again in the morning, but feel free to profile it yourself so we have multiple sets of data to verify it.

wolfdale:
I did the profiling with 5000 hypervisor with 100 guests (not sure about upper cap that we usually use to do such things) which took about 11 minutes to complete. I didn't noticed much of deviation from expectation. There are only minute differences in Heap memory & CPU utilization. Garbage collection & pause time was also of similar pattern in both cases. 

I noticed slightly more activity with ActiveMQ client threads that were busy to open/close sessions, but this was expected and was under control.

Ceiu:
That result is inline with what I'm seeing. The sendMessage and commitMessage occupy around 4% of the total runtime, but the time spent in the session opening and closing of those operations isn't even registering in the profiler.

PR #2737, opened 2020-06-26
[F] WIP: RefreshWorker now prunes "unused" entities during refresh (ENT-2352)

Ceiu:
DONE:
- Added the PoolNodeBuilder, PoolNodeVisitor, and general handling
  of pool nodes. Pool entities are still not processed by the
  refresh worker, but the framework is now present.
- NodeVisitor now has a pruneNode operation
- NodeProcessor now performs two passes through the trees: once to
  prune nodes/entities, and once to process/merge remaining entities
- ContentNodeVisitor and ProductNodeVisitor now will remove
  "locked" entities which exist locally but are not present upstream
  during a refresh (suggesting they are no longer used by the org)
- NodeProcessor.processNodes now returns a RefreshResult instance
  directly instead of requiring an explicit call to compileResults
- NodeFactory is now responsible for building all nodes, rather
  than requiring explicit, individual calls to NodeFactory.buildNode
- Updated OwnerContentCurator.removeOwnerContentReferences to no
  longer use the unsafe bulkSQLDelete, and instead performs properly
  partitioned JPQL deletes
- Updated OwnerContentCurator.updateOwnerContentReferences to use
  partitioned data sets, to avoid running over the parameter limit
  when processing certain sets of data
- Added the AbstractHibernateCurator.partitionMap utility method
- Removed EntityNode.visited() and any backing state management for
  it, as storing the visited state on the nodes does not work very
  well for performing multiple passes through the tree(s)
- Removed other EntityNode utility methods for determining the
  type of entity operation performed during refresh, as they were
  not very accurate, and were also unused
- Added EntityNode.NodeState for better definining the resultant
  state of the entity represented by the node. This field may also
  double as an impromptu "visited" flag in certain circumstances

TODO:
- Rewrite/Fix a bunch of tests
- Fix OwnerProductCurator to be in line with the changes made to
  OwnerContentCurator
- Updated RefreshResult to better collect, store, and provide
  its entities. This should be done with an EntityState enum,
  and by storing a Pair<EntityState, Entity> rather than the
  raw entity in individual maps. This will prevent an entity
  from existing twice in the EntityStore and will make fetching
  entire maps much easier with some fancy use of streams,
  filters, and collectors.
- Remove the new code from EnvironmentContentCurator if it's not
  used in the final implementation
- Add a NodeState.OMITTED so we don't run into issues when
  building a refresh result that has any pools. Since they're
  being skipped entirely, the result will almost always be invalid
  and lead to NPEs at best, and undefined runtime behavior at worst.
- Add details for other work that's been done that I haven't yet
  captured in this commit message.
- Potentially remove the flat entity maps in favor of just building
  the nodes directly. This may not work given the order of
  operations, but it seems like having n+1 flatmaps is not very
  memory efficient, nor required.


Ceiu:
For reviewers:
The bulk of the core logic is done here, but there are piles of tests and some cleanup to be done (as indicated in the TODO above).

The one bit of logic I'm struggling to deal with is the order-of-operations with regards to pool updating and product removal. That is, at the point we've identified that a product is unused, it is possible that a local pool still references it and, thus, has a link to it in the DB and in its local managed entity. I've temporarily worked around it by lifting the restriction of verifying a product has a pool count of 0 before removal, but this allows us to enter a state in which a pool can link to a product which is not linked to the org if misused. I suppose that could simply be one of those things we check during review, but that's a tad scary.

This problem sorts itself out when pool refresh is moved into this framework if we flip the order of pruning and processing around in NodeProcessor (such that processing is done before pruning), but so long as we do pool processing as an external step well after all of this, we have a potential risk that must be either ignored or somehow worked around.

server/src/main/java/org/candlepin/async/tasks/OrphanCleanupJob.java
```diff
@@ -88,23 +91,29 @@ private int deleteOrphanedContent() {
...
+        List<String> orphanedProductUuids = this.ownerProductCurator.getOrphanedProductUuids();
 
-        for (Product product : productQuery) {
-            this.productCurator.delete(product);
-            ++count;
+        Set<Pair<String, String>> activePoolProducts = this.productCurator
+            .getProductsWithPools(orphanedProductUuids);
+
+        if (activePoolProducts != null && !activePoolProducts.isEmpty()) {
+            log.warn("Found {} pools referencing orphaned products:");
```
Januson, line 107: Log argument is missing.

server/src/main/java/org/candlepin/controller/refresher/builders/NodeFactory.java
```diff
@@ -67,23 +71,45 @@ public NodeFactory addBuilder(NodeBuilder builder) {
...
+     *  the mapper to add to this factory
      *
      * @throws IllegalArgumentException
      *  if the provided mapper is null
      *
      * @return
      *  a reference to this node factory
      */
-    public NodeFactory setNodeMapper(NodeMapper mapper) {
+    public NodeFactory addMapper(EntityMapper mapper) {
```
Januson, line 86: Could we do something about our use of raw types? It's starting to feel a bit excessive. We could at least use the wildcards.

server/src/main/java/org/candlepin/controller/refresher/visitors/ContentNodeVisitor.java
```diff
@@ -87,32 +89,38 @@ public ContentNodeVisitor(ContentCurator contentCurator, OwnerContentCurator own
...
-                    contentUuidMap = new HashMap<>();
-                    this.ownerContentUuidMap.put(node.getOwner(), contentUuidMap);
+            if (importedEntity != null) {
+                if (ContentManager.isChangedBy(existingEntity, importedEntity)) {
+                    Content mergedEntity = this.createEntity(mapper, node);
+                    node.setMergedEntity(mergedEntity);
+
+                    // Store the mapping to be updated later
+                    Map<String, String> contentUuidMap = this.ownerContentUuidMap.get(node.getOwner());
+                    if (contentUuidMap == null) {
```
Januson, line 116: We could you Map.computeIfAbsent here.

server/src/main/java/org/candlepin/async/tasks/OrphanCleanupJob.java
Ceiu: Fixed

server/src/main/java/org/candlepin/controller/refresher/visitors/ContentNodeVisitor.java
Ceiu: Ahh, nice. Turns out it's ever-so-slightly faster, too. Fixed here and in a few other places.

server/src/main/java/org/candlepin/controller/refresher/builders/NodeFactory.java
Ceiu: I've added wildcards where appropriate, but this doesn't really buy us much outside of one weird compiler error that required a random cast to the same type.

Januson (approved):

PR #2738, opened 2020-06-30
[M] Restored omitted return statement in the hypervisor update job (ENT-2577)

Ceiu:
- Restored a return statement which was erroneously removed during
  the transition to the updated JobExecutionContext


barnabycourt:
Looks good

PR #2746, opened 2020-07-06
ENT-1882: Port PoolResource to spec-first

sonalidhome:
- Added Pool reseaource specification
 - Upated related files

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -2849,6 +2849,324 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /pools:
+    get:
+      description: |
+        Retrieves a list of pools @deprecated Use the method on /owners.
+        This endpoint supports paging with query parameters.
+        For more details please visit https://www.candlepinproject.org/docs/candlepin/pagination.html#paginating-results-from-candlepin.
+      tags:
+        - Pools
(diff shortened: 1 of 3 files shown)


api/candlepin-api-spec.yaml
```diff
@@ -2849,6 +2849,305 @@ paths:
...
+        - Certificates
+      operationId: getSubCert
+      parameters:
+        - in: path
+          name: pool_id
+          description: Pool ID
+          required: true
+          schema:
+            type: string
+      responses:
```
wolfdale, line 3138: Missing API response (HTTP 404) when "Pool ID or cert not found".

api/candlepin-api-spec.yaml
```diff
@@ -2849,6 +2849,305 @@ paths:
...
+        200:
+          description: Retrives a single pool by its ID
```
wolfdale, line 2997: Incorrect description for delete API.

Januson, line 2997: Wrong description and response code should be 204.

api/candlepin-api-spec.yaml
sonalidhome: Added.

sonalidhome:
Added break lines in API description.

PR #2749, opened 2020-07-09
ENT-2455: Fix too many params for Entitlements query

Januson:
Changes I needed to prevent "too many parameters" error while testing with a large DB dump.

Note: I am not sure how to deal with paging. With partitioned parameters we will be loading a lot of unnecessary data.


server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -437,27 +440,69 @@ private String sanitizeMatchesFilter(String matches) {
...
-                entitlementsPage.setPageData(Collections.<Entitlement>emptyList());
+                entitlementsPage.setPageData(Collections.emptyList());
                 entitlementsPage.setMaxRecords(0);
             }
         }
-
         return entitlementsPage;
     }
 
+    private Set<Entitlement> findEntitlements(List<String> entitlementIds) {
```
Ceiu, line 454: If we're going to be using this lookup for paging, we need the query itself to be ordered, and we cannot use a set (as its ordering is not guaranteed either).

The entire criteria usage itself likely needs to be axed here.

Januson: Done.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -437,27 +437,61 @@ private String sanitizeMatchesFilter(String matches) {
...
+            result.addAll(query.setParameter("entitlement_ids", block).getResultList());
+        }
+        return result;
+    }
+
+    private Page<List<Entitlement>> createPage(List<Entitlement> entitlements, PageRequest pageRequest) {
+        Page<List<Entitlement>> page = new Page<>();
+        if (pageRequest != null) {
+            int start = (pageRequest.getPage() - 1) * pageRequest.getPerPage();
+            List<Entitlement> currentPage = entitlements.stream()
```
Ceiu, line 475: If we're requiring a list as input, we could use .sublist instead of streaming and collecting.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -437,27 +437,61 @@ private String sanitizeMatchesFilter(String matches) {
             List<String> entitlementIds = criteria.list();
 
             if (entitlementIds != null && !entitlementIds.isEmpty()) {
-                criteria = this.currentSession()
-                    .createCriteria(Entitlement.class)
-                    .add(CPRestrictions.in("id", entitlementIds));
+                entitlementsPage = createPage(this.findEntitlements(entitlementIds), pageRequest);
```
Ceiu, line 440: We're still broken here (and, for the record, we were broken before): as soon as we re-query here, we lose any sorting from the pageRequest unless we pass the same sorting params through.

Additionally, if we're going to keep the multi-query behavior, we should use this as an opportunity to not query the entire set of entitlements, and only grab the entitlements from the block of IDs we're going to fetch.

Ceiu:
The changes, in a vacuum, are fine. But, the problem I have here overall is that this change highlights a bigger problem with the ```listFilteredPages``` method; namely that it doesn't work. Paged requests will not fetch the requested info, as our paging and sorting gets thrown out in favor of ordering by entitlement ID in the second query. This is better than it was before (being both unreliable and unrepeatable), but it's still fundamentally busted.

Really, the whole method needs to be rewritten to look something closer to the dynamic filtering queries from the ```AsyncJobStatusCurator```; but with proper gating on the parameter and block sizes. The use of old Hibernate criteria should be tossed out, and the whole method rewritten from the ground up.

Ceiu:
Since the metaclasses are generated, I don't think those should be committed as part of the repo. Should probably follow the rest of the generated classes and just be a compile-time thing.

Other than that, my only remaining gripes are that of organization.

nikosmoum (commented):
I know this is already reviewed, but just commenting for future reference: Please squash commits before merging, and when adding/removing/updating a dependency in build.gradle, please regenerate the pom file by running `./gradlew pom` so that it will also get the poms updated, or otherwise this holds up our release automation.

PR #2760, opened 2020-07-22
[M] 1854221: Added additional job cleanup to JobManager and JobCleaner (ENT-2632, ENT-2630, ENT-2586)

Ceiu:
- JobCleaner will now abort non-terminal, non-running jobs older
  than three days (by default; configurable)
- JobCleaner will now abort running jobs older than two days
  (by default; configurable)
- JobManager will now attempt to "rewind" the job state for
  abandoned jobs during startup, where an "abandoned" job is
  defined as one in the RUNNING state with the same executor
  as the node which is performing the check

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -63,6 +63,7 @@ import org.slf4j.MDC;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -73,6 +74,7 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
@@ -471,6 +473,10 @@ public class JobManager implements ModeChangeListener {
                 if (this.modeManager.getCurrentMode() == Mode.SUSPEND) {
                     this.suspendKeys.add(SUSPEND_KEY_TRIGGERED);
                 }
+
+                // Attempt to restore any jobs which were running on this node but did not get
+                // to gracefully shutdown.
+                this.recoverAbandonedJobs();
             }
 
             log.info("Job manager initialization complete");
@@ -492,6 +498,30 @@ public class JobManager implements ModeChangeListener {
             }
 
             throw new StateManagementException(this.state, ManagerState.INITIALIZED, errmsg, e);
+        }
+    }
+
+    /**
+     * Attempts to recover all jobs that were abandoned by this node if it shutdown abnormally
+     * while executing jobs. Any jobs in the RUNNING state with an executor set to this node
+     * will have their state forcefully rewound to QUEUED in an attempt to allow the task to be
+     * rerun.
+     */
+    private void recoverAbandonedJobs() {
+        AsyncJobStatusQueryBuilder queryBuilder = new AsyncJobStatusQueryBuilder()
+            .setJobStates(Collections.singleton(JobState.RUNNING))
+            .setExecutors(Collections.singleton(Util.getHostname()));
+
+        // Impl note: this violates normal state transitions, but we're trying to recover from an
+        // ungraceful shutdown in which our process was terminated while a job was running. We'll
+        // set the job's state to QUEUED, which will allow us to pick up the job and run it again
+        // next time the message is received. If the message has been lost, then the job cleaner
+        // will eventually nuke this job as part of its non-terminal job aborting step.
+        for (AsyncJobStatus job : this.jobCurator.findJobs(queryBuilder)) {
+            log.warn("Recovering abandoned job: {}", job);
+
+            job.setState(JobState.QUEUED);
+            this.jobCurator.merge(job);
         }
     }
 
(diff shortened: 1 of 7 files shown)


Ceiu:
For reviewers and those doing Jira bookkeeping:
This PR addresses ENT-2632, supersedes ENT-2630, and partially overlaps with the goals of ENT-2586. Further discussion will be necessary to determine the final scoping of ENT-2586 in light of these changes, but the former two tasks will be addressed here.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1717,11 +1746,59 @@ public int cleanupJobs(AsyncJobStatusQueryBuilder queryBuilder) {
...
+
+        // Add any other sanity restrictions deemed necessary here
+
+        List<AsyncJobStatus> jobs = this.jobCurator.findJobs(queryBuilder);
+
+        // Step through each job and update the job state
+        for (AsyncJobStatus job : jobs) {
+            this.setJobState(job, JobState.ABORTED);
+            this.jobCurator.merge(job);
+        }
```
nikosmoum, line 1797: This could be a performance hit. Why do we need to fetch the actual objects before changing their state?
If the only reason is to utilize the transition validation the `setJobState` does, then maybe we should be doing that validation without fetching the job objects into memory (we have the collection of JobState, we can just validate against JobState.ABORTED).

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -495,6 +500,30 @@ public synchronized void initialize() throws StateManagementException {
...
+        // Impl note: this violates normal state transitions, but we're trying to recover from an
+        // ungraceful shutdown in which our process was terminated while a job was running. We'll
+        // set the job's state to QUEUED, which will allow us to pick up the job and run it again
+        // next time the message is received. If the message has been lost, then the job cleaner
+        // will eventually nuke this job as part of its non-terminal job aborting step.
+        for (AsyncJobStatus job : this.jobCurator.findJobs(queryBuilder)) {
+            log.warn("Recovering abandoned job: {}", job);
+
+            job.setState(JobState.QUEUED);
+            this.jobCurator.merge(job);
```
nikosmoum, line 523: Same question here: why fetch job objects into memory? we don't even do transition validation here, so an update query would be much faster

server/src/main/java/org/candlepin/async/tasks/JobCleaner.java
```diff
@@ -63,36 +66,66 @@ public JobCleaner(Configuration config, JobManager jobManager) {
...
             .collect(Collectors.toSet());
 
         // Build the query builder with our config
         AsyncJobStatusQueryBuilder queryBuilder = new AsyncJobStatusQueryBuilder()
             .setJobStates(jobStates)
             .setEndDate(cutoff);
 
-        int deleted = this.jobManager.cleanupJobs(queryBuilder);
+        int removed = this.jobManager.cleanupJobs(queryBuilder);
+        log.info("Removed %d terminal jobs older than %2$tF %2$tT%2$tz", removed, cutoff);
```
nikosmoum, line 110: This logging doesn't work. When the JobCleaner runs the logs print the literal string `Removed %d terminal jobs older than %2$tF %2$tT%2$tz`. The intellij tooltip says "More arguments provided (2) than placeholders specified (0)"

server/src/main/java/org/candlepin/async/tasks/JobCleaner.java
```diff
@@ -63,36 +66,66 @@ public JobCleaner(Configuration config, JobManager jobManager) {
...
+
+        // Build the query builder with our config
+        AsyncJobStatusQueryBuilder queryBuilder = new AsyncJobStatusQueryBuilder()
+            .setJobStates(jobStates)
+            .setEndDate(cutoff);
 
-        String result = String.format("Removed %d terminal jobs older than %2$tF %2$tT%2$tz",
-            deleted, cutoff);
+        int aborted = this.jobManager.abortNonTerminalJobs(queryBuilder);
+        log.info("Aborted %d non-running jobs older than %2$tF %2$tT%2$tz", aborted, cutoff);
```
nikosmoum, line 127: Same here, logging doesn't work

server/src/main/java/org/candlepin/async/JobManager.java
Ceiu: The number here *should* be no more than the number of job executor threads configured (default is 10), so it shouldn't be a *huge* hit compared to a raw query. The important thing I want here, though, is the logging.

server/src/main/java/org/candlepin/async/tasks/JobCleaner.java
Ceiu: Oh, whoops. That's a copy/paste mistake on my part.

server/src/main/java/org/candlepin/async/JobManager.java
Ceiu: The original intent was:
- Validate the state transition (since this is a public utility method that may get exposed via endpoint soon)
- Update the job result with some kind of abort message
- The number of jobs we're terminating this way should never be large

I ended up not doing the second one there, but I don't remember if I made a decision not to for some reason, or just plain forgot to do it. But, an update query would not prevent this behavior anyway.

An update query would, without question, be faster and more efficient here, but at the expense of the transition validation.

Ceiu: I trailed off a bit, but what are you thoughts there?

server/src/main/java/org/candlepin/async/tasks/JobCleaner.java
Ceiu: Fixed

server/src/main/java/org/candlepin/async/tasks/JobCleaner.java
Ceiu: Fixed

Ceiu:
After more discussion and investigation today, I wonder if it makes sense to add a 3rd option to the JobCleaner for "max job runtime" to catch running jobs that seem stuck. I think that starts getting into danger territory though, so I'm not entirely convinced. Thoughts?

server/src/main/java/org/candlepin/async/JobManager.java
nikosmoum: You're right. 10 jobs should be negligible

server/src/main/java/org/candlepin/async/JobManager.java
nikosmoum: I agree that the number of jobs this will target _should_ never be large, but with the current situation in Hosted, I'd expect this to be fetching a large amount at least the first couple of times. Also, I don't think the transition validation would be an expense. We can still do that without fetching the jobs, since we already have the static list of states we're fetching for, and if the validation passes for all states, then we do the update query.

nikosmoum:
Do we even need to differentiate between non-running non-terminal and running non-terminal? In the past the JobCleaner was not just aborting, but straight up deleting everything older than 4 days: https://github.com/candlepin/candlepin/blob/candlepin-2.9-HOTFIX/server/src/main/java/org/candlepin/model/JobCurator.java#L94-L100

server/src/main/java/org/candlepin/async/JobManager.java
Ceiu: Oof... I hadn't even consider that. I had myself stuck in the idea that every job needed state transition validation individually. I'll work on a query for bulk updating a job's state.

Ceiu: Query added

Ceiu:
I've added an additional cycle to remove running jobs older than 2 days by default.

PR #2765, opened 2020-07-27
Fix for params in Link header URLs when paging

sonalidhome:
- Getting queryParamsStartOffset = -1 if the URI does not
have query parameters. Altered this condition to set the offset
as request-URI length.


nikosmoum:
Closing this PR, because it looks like the tests are what need to be fixed, not the code changed, and more importantly, Jenkins isn't actually running the unit tests in the common package.. which is why we didn't see those tests (in `LinkHeaderResponseFilterTest.java` ) failing.

PR #2766, opened 2020-07-27
[M] 1854221: Graceful shutdown of Artemis job threads; ENT-2586

nikosmoum:
Now, when tomcat is shut down gracefully:
- First wait for Artemis client threads to finish before
  closing the sessions, to make sure that commits happen
  instead of potentially finishing a job and the broker
  not getting a commit back.
- Don't use ActiveMQClient.clearThreadPools() to quiesce
  the threads, because it:
   * Waits only 10 seconds for jobs to finish before force
   shutting them down.
   * Uses ExecutorService.shutdownNow() which will interrupt
   threads, causing an InterruptedException and not wait for
   the jobs to finish.
- Use ExecutorService.shutdown() which will gracefully wait
  for the threads to finish their work.
- Introduce new config "candlepin.async.thread.shutdown.timeout"
  to specify how long (in seconds) to wait for job threads to
  finish before forcefully shutting them down.

### DIFF
--- a/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
+++ b/server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
@@ -14,9 +14,7 @@
  */
 package org.candlepin.guice;
 
-import static org.candlepin.config.ConfigProperties.ACTIVEMQ_ENABLED;
-import static org.candlepin.config.ConfigProperties.ENCRYPTED_PROPERTIES;
-import static org.candlepin.config.ConfigProperties.PASSPHRASE_SECRET_FILE;
+import static org.candlepin.config.ConfigProperties.*;
 
 import org.candlepin.async.JobManager;
 import org.candlepin.audit.AMQPBusPublisher;
@@ -259,6 +257,9 @@ public class CandlepinContextListener extends GuiceResteasyBootstrapServletConte
     }
 
     private void destroySubsystems() throws Exception {
+        // Perform graceful shutdown operations before the job system's final destruction
+        this.cpmContextListener.shutdown();
+
         // Tear down the job system
         this.jobManager.shutdown();
 
--- a/server/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/server/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -315,6 +315,9 @@ public class ConfigProperties {
         UnmappedGuestEntitlementCleanerJob.JOB_KEY
     };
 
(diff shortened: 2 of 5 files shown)


nikosmoum:
Testing instructions:

- Make sure to add a "TimeoutStopSec=" long enough in
  /usr/lib/systemd/system/tomcat.service because by default
  'systemctl stop' will only wait the service to shutdown
  in 90 seconds before forcing.
- Choose a job type and make sure it will run long (e.g. 4-5 mins)
  by any means possible (e.g. introduce a sleep statement).
- Run the job (e.g. by running a custom rspec test)
- Run 'systemctl stop tomcat' immediately after
- Expected: Tomcat should shutdown after however minutes that
  job's sleep statement was (more or less). Checking the database
  for that job record should show it is FINISHED with a
  result column populated accordingly.

server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -291,6 +294,35 @@ private void destroySubsystems() throws Exception {
         this.loggerListener.contextDestroyed();
     }
 
+    private void waitForArtemisJobThreadsToFinish() {
```
Ceiu, line 297: If this is specifically for the job system, we should move this to somewhere that makes a bit more sense. Either in jobManager.shutdown or, and probably best in my opinion, ArtemisContextListener.shutdown (which gets hit automatically via cpmContextListener.shutdown above)

server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -259,6 +260,8 @@ public void contextDestroyed(ServletContextEvent event) {
     }
 
     private void destroySubsystems() throws Exception {
+        this.waitForArtemisJobThreadsToFinish();
```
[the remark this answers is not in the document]
nikosmoum: It was deliberately added before the JobManager.shutdown, because that method calls into these:
```
                this.dispatcher.shutdown();
                this.receiver.shutdown();
```
and the `receiver.shutdown()` is actually closing the sessions that these threads use. 
We don't want to close sessions of threads that are still running, but first wait for the threads to finish

server/src/main/java/org/candlepin/guice/CandlepinContextListener.java
nikosmoum: Same as above answer, by the time we get to `ArtemisContextListener.shutdown`, the sessions are already closed. We want the sessions to keep open for the message commits to happen first

PR #2767, opened 2020-07-28
[M] Fixed a race condition with messaging session committal in the job system (ENT-2679)

Ceiu:
- Fixed a race condition with messaging session committal that could
  occur if two threads queue a job at the same time: the first thread
  to complete its database transaction would commit the messaging
  session shared by both, causing both job messages to be sent and
  potentially picked up before the second thread's database session
  had finished persisting the job status

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
+++ b/server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
@@ -14,7 +14,6 @@
  */
 package org.candlepin.async;
 
-import org.candlepin.common.config.Configuration;
 import org.candlepin.messaging.CPMException;
 import org.candlepin.messaging.CPMMessage;
 import org.candlepin.messaging.CPMProducer;
@@ -28,6 +27,11 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.lang.ref.Reference;
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


server/src/test/java/org/candlepin/async/JobMessageDispatcherTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+        CPMSession session = this.mockCPMSession();
+        CPMProducer producer = mock(CPMProducer.class);
+
+        doReturn(session).when(this.sessionFactory).createSession(any(CPMSessionConfig.class));
+        doReturn(producer).when(session).createProducer(any(CPMProducerConfig.class));
+
+        TaskExecutor executor = new TaskExecutor();
+
+        try {
```
Januson, line 241: Redundant try-catch

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
```diff
@@ -44,12 +49,158 @@
...
+         */
+        @Override
+        public int hashCode() {
+            return this.hashCode;
+        }
+    }
+
+
+
     private final Configuration config;
```
Januson, line 198: Not a part of your changes but there is an unused dependency.

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
```diff
@@ -44,12 +49,158 @@
...
+                return true;
+            }
+
+            if (cmp instanceof ThreadReference) {
+                Thread lhs = this.get();
+                Thread rhs = ((ThreadReference) cmp).get();
+
+                return lhs != null ?
+                    lhs.equals(rhs) :
+                    rhs == null && this.hashCode() == ((ThreadReference) cmp).hashCode();
```
Januson, line 181: Redundant cast

server/src/test/java/org/candlepin/async/JobMessageDispatcherTest.java
Ceiu: Fixed

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
Ceiu: Fixed

server/src/main/java/org/candlepin/async/JobMessageDispatcher.java
Ceiu: Fixed

PR #2770, opened 2020-08-04
ENT-2703: Jenkins isn't running unit tests of common package

sonalidhome:
- The project directory path is set to the server package.
   This leads to execute the test only from the server package.
   Hence added the path of commons package to execute
   the test of commons.

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -45,7 +45,7 @@ build_lint() {
 build_unittest() {
     UTCMD=$1
     if [ -f ../gradlew ]; then
-        ../gradlew test
+        cd $CP_HOME; ./gradlew test; cd $PROJECT_DIR
     else
         buildr $UTCMD
     fi


docker/candlepin-base/cp-test
```diff
@@ -45,6 +45,13 @@ build_lint() {
 build_unittest() {
     UTCMD=$1
     if [ -f ../gradlew ]; then
+
+        # set project to run units from the candlepin-common project
+        cd ../common/
+        ../gradlew test
+
+        # set project back to server
+        cd ../server/
```
nikosmoum, line 54: Instead of doing `cd` everywhere, we should simplify things, and just run `.gradlew test` from the root directory.
So here this if block would become:
```
    if [ -f ./gradlew ]; then
        ./gradlew test
```
and later on (right before we run the unit tests), we don't cd into server/ , but cd into the root dir (so `cd $PROJECT_DIR` turns into `cd $CP_HOME`. And we only do a `cd $PROJECT_DIR` later, once the unit tests are finished and before we setup qpid

sonalidhome: Agreed, simplified the code.

PR #2772, opened 2020-08-04
1865906: Add "insights_auto_register" capability

cnsnyder:
This adds the string "insights_auto_register" to the capabilities returned from the status response.


nikosmoum:
Merging since the 2 unit test failures are irrelevant and will be fixed by #2771

PR #2774, opened 2020-08-11
[M] 1866525: Updated SCA certificate content prefixes and paths (ENT-2717)

Ceiu:
- Certificates generated for Simple Content Access (SCA) will
  use a placeholder value for the ENTITLEMENT_DATA certificate
  extension when generated from a Candlepin instance not running
  in standalone mode
- Content prefixes are no longer generated while Candlepin is not
  running in standalone mode

### DIFF
--- a/server/src/main/java/org/candlepin/controller/ContentAccessManager.java
+++ b/server/src/main/java/org/candlepin/controller/ContentAccessManager.java
@@ -15,6 +15,8 @@
 package org.candlepin.controller;
 
 import org.candlepin.audit.EventSink;
+import org.candlepin.common.config.Configuration;
+import org.candlepin.config.ConfigProperties;
 import org.candlepin.model.CandlepinQuery;
 import org.candlepin.model.CertificateSerial;
 import org.candlepin.model.CertificateSerialCurator;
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -406,27 +419,72 @@ else if (type.isType(ConsumerTypeEnum.HYPERVISOR)) {
...
+     * @param environment
+     *  the environment to which the consumer receiving the entitlement belongs
+     *
+     * @return
+     *  the container content path for the given owner and environment
+     */
+    private String getContainerContentPath(Owner owner, Environment environment) throws IOException {
+        // Fix for BZ 1866525:
+        // - In hosted, SCA entitlement content path needs to use a dummy value to prevent existing
+        //   clients from breaking, while still being clear the path is present for SCA
```
nikosmoum, line 438: Just curious why/who having a dummy value prevents existing clients from breaking?

Ceiu: The big one is subman/rhsmlib's certificate2.py doesn't check that the extension properly exists before attempting to use the value, so the result you get is an exception rather than a false from ```check_paths```, or empty array of paths in ```provided_paths```.

PR #2776, opened 2020-08-18
[2.6] 1869582 - Satellite does not have "Layered" SLA option for system purpose

sonalidhome:
- Added a condition in compliance rules do not match the system
   purpose status
 - Added rspec to test this scenario
 - Updated rules.js to ignore calculating priority of subscriptions
   to attach during auto-attach

### DIFF
--- a/server/spec/autobind_spec.rb
+++ b/server/spec/autobind_spec.rb
@@ -682,4 +682,56 @@ describe 'Autobind On Owner' do
     entitlements = @cp.list_entitlements(:uuid => consumer.uuid)
     entitlements.size.should == 1
   end
+
+  it 'should be matched with consumer on auto attach when SLA of current entitlement is layered exempt' do
+    product1 = create_product(random_string('product'),
+                              random_string('product'),
+                              {:attributes => {:support_level => 'Standard'},
+                               :owner => owner_key})
+    product2 = create_product(random_string('product'),
+                              random_string('product'),
+                              {:attributes => {:support_level => 'Layered',
+                                               :support_level_exempt => 'true'},
+                               :owner => owner_key})
+
+    @cp.create_pool(owner_key, product1.id)
+    @cp.create_pool(owner_key, product2.id)
+
+    installed = [
+        {'productId' => product1.id, 'productName' => product1['name']},
+        {'productId' => product2.id, 'productName' => product2['name']}]
+
+    consumer = @cp.register(
+        random_string('system'), :system, nil, {}, nil, owner_key, [], installed, nil, [],
+        nil, [], nil, nil, nil, nil, nil, 0, nil, 'Standard')
+
+    @cp.consume_product(product2['id'], {:uuid => consumer.uuid})
+
+    status = @cp.get_purpose_compliance(consumer['uuid'])
+    expect(status['status']).to eq('mismatched')
+
+    entitlements = @cp.list_entitlements(:uuid => consumer.uuid)
+
+    # Trying to bind the without SLA exempt.
+    # Then system purpose status should be match.
+
+    entitlements.each do |ent|
+      @cp.unbind_entitlement(ent.id, {:uuid => consumer.uuid})
+    end
+
+    product3 = create_product(random_string('product'),
+                              random_string('product'),
+                              {:attributes => {:support_level => 'Standard',
+                                               :support_level_exempt => 'false'},
+                               :owner => owner_key})
+
+    @cp.create_pool(owner_key, product3.id)
+
+    @cp.consume_product(product3['id'], {:uuid => consumer.uuid})
+
+    status = @cp.get_purpose_compliance(consumer['uuid'])
+    expect(status['status']).to eq('matched')
+
+  end
+
 end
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
```diff
@@ -198,7 +198,12 @@ public SystemPurposeComplianceStatus getStatus(Consumer consumer,
                 if (StringUtils.isNotEmpty(preferredSla) &&
                     product.hasAttribute(Product.Attributes.SUPPORT_LEVEL)) {
                     String sla = product.getAttributeValue(Product.Attributes.SUPPORT_LEVEL);
-                    if (sla.equalsIgnoreCase(preferredSla)) {
+                    //TODO: This is a short term fix for bug 1858167. Added a condition to match
+                    // Syspurpose status for entitlement which has SLA as Layered and SLA exempt as true
+                    boolean slaExempt = Boolean.parseBoolean(product
+                        .getAttributeValue(Product.Attributes.SUPPORT_LEVEL_EXEMPT));
+                    if ((sla.equalsIgnoreCase("Layered") && slaExempt) ||
```
nikosmoum, line 205: If we are indeed going to apply this solution, then we should only check the `slaExempt` attribute, and not if the sla is `Layered`. From what I understand, currently only Layered SLAs are considered exempt for now, but you never know if in the future we might have another SLA type that is also considered exempt.

server/spec/autobind_spec.rb
```diff
@@ -682,4 +682,56 @@
...
+        {'productId' => product2.id, 'productName' => product2['name']}]
+
+    consumer = @cp.register(
+        random_string('system'), :system, nil, {}, nil, owner_key, [], installed, nil, [],
+        nil, [], nil, nil, nil, nil, nil, 0, nil, 'Standard')
+
+    @cp.consume_product(product2['id'], {:uuid => consumer.uuid})
+
+    status = @cp.get_purpose_compliance(consumer['uuid'])
+    status['status'].should == 'matched'
```
nikosmoum, line 711: We should avoid the `.should ==` because it is deprecated. instead, the `expect(variable).to eq(value)` should be used

server/spec/autobind_spec.rb
```diff
@@ -682,4 +682,56 @@
...
+                              {:attributes => {:support_level => 'Layered',
+                                               :support_level_exempt => 'false'},
+                               :owner => owner_key})
+
+    pool3 = create_pool_and_subscription(owner_key, product3.id)
+
+    @cp.consume_product(product3['id'], {:uuid => consumer.uuid})
+
+    status = @cp.get_purpose_compliance(consumer['uuid'])
+    status['status'].should == 'mismatched'
```
nikosmoum, line 733: same: use `expect` instead

server/spec/autobind_spec.rb
```diff
@@ -682,4 +682,56 @@
...
+                              {:attributes => {:support_level => 'Standard'},
+                               :owner => owner_key})
+    product2 = create_product(random_string('product'),
+                              random_string('product'),
+                              {:attributes => {:support_level => 'Layered',
+                                               :support_level_exempt => 'true'},
+                               :owner => owner_key})
+
+    pool1 = create_pool_and_subscription(owner_key, product1.id)
+    pool2 = create_pool_and_subscription(owner_key, product2.id)
```
nikosmoum, line 698: The method is deprecated. We should create pools directly using create_pool method from candlepin_api.rb

server/spec/autobind_spec.rb
```diff
@@ -682,4 +682,56 @@
...
+      @cp.unbind_entitlement(ent.id, {:uuid => consumer.uuid})
+    end
+
+    product3 = create_product(random_string('product'),
+                              random_string('product'),
+                              {:attributes => {:support_level => 'Layered',
+                                               :support_level_exempt => 'false'},
+                               :owner => owner_key})
+
+    pool3 = create_pool_and_subscription(owner_key, product3.id)
```
nikosmoum, line 728: same: use create_pool instead

nikosmoum:
Also, I forgot to mention: the buzilla-reference jenkins build is failing because the commit/PR summary is referring to a Satellite bz id. Instead, they should be referring to the candlepin clone bz id.

server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
jsefler: I agree with Nikos.   There should be no hard-coded reference to 'Layered'; slaExempt is what is imprtant.

server/spec/autobind_spec.rb
sonalidhome: Updated.

server/spec/autobind_spec.rb
sonalidhome: Yeah, updated to new method.

server/spec/autobind_spec.rb
sonalidhome: updated

server/spec/autobind_spec.rb
sonalidhome: corrected

server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
sonalidhome: Agreed,  updated the condition to check only `slaExempt`.

nikosmoum:
Marking as in-progress because there's more discussion to be had around this first

nikosmoum:
Closing this, since it's become stale and we have not reached a definite way to resolve the use cases around this. Will revisit in the future.

PR #2778, opened 2020-08-19
[M] 1591315: Registration with activation keys no longer fails in SCA mode (ENT-2747)

Ceiu:
- Clients registering with activation keys to organizations using
  simple content access (SCA) mode will no longer fail if a key
  cannot be fully entitled


Januson (approved):
Changes look good.

PR #2781, opened 2020-08-25
org_migrator to python 3 and pipenv

Tomboyo:
I set up org_migrator yesterday in order to help research an incident. In the process I captured the dependencies of the script in a pipenv file and updated two lines so that this was python 3 compatible. I figured I would create a PR if anyone would like these changes.

Let me know if I should squash or anything, happy to do so.

Thanks!


nikosmoum:
ok to test

Ceiu:
Changes look alright to me, but I would prefer the commits to be squashed.

Tomboyo:
@Ceiu Can do! Commits are now squished together.

server/bin/org_migrator/org_migrator.py
```diff
@@ -64,7 +64,7 @@ def list_orgs(db):
 
 def jsonify(data):
     def data_converter(obj):
-        if isinstance(obj, bytearray) or isinstance(obj, buffer):
+        if isinstance(obj, (bytearray, memoryview)):
```
Ceiu, line 67: In line with the other Python2 compat change, can we keep the buffer type present? I've also added some warning output and a handler in the event the type isn't present:

```
    def data_converter(obj):
        if isinstance(obj, (bytearray, buffer, memoryview)):
            return binascii.b2a_base64(obj)
        if isinstance(obj, datetime.datetime):
            return str(obj)

        log.warn("Unexpected type received; attempting to convert to string: %s" % (type(obj)))
        return str(obj)
```

Ceiu:
I want to keep the pipenv stuff, but the other bits will need to be changed as well. In the interest of time, I'll merge this as-is and put in the fixes for the other bits in another PR.

PR #2783, opened 2020-08-25
[M] Standardized string keys used for modifying the logging context

Ceiu:
- Added the LoggingUtil class in the common package to hold
  constants for the various MDC keys used throughout Candlepin
- All instances of MDC.set and MDC.get now use the constants
  defined in the LoggingUtil class rather than hard-coded
  strings or locally-defined constants


wolfdale (approved):
ACK

PR #2785, opened 2020-08-26
[M] Added the upstream_subscription_loader debugging tool

Ceiu:
- Added the new upstream_subscription_loader.rb tool for assisting
  in reproducing and debugging issues during refresh and other
  adapter-backed operations reliant on upstream data
- Removed some unnecessary logging code in CandlepinPoolManager
- Added a missing existence check in HostedTestSubscriptionResource

### DIFF
--- a/server/bin/upstream_subscription_loader.rb
+++ b/server/bin/upstream_subscription_loader.rb
@@ -1,0 +1,138 @@
+#!/usr/bin/env ruby
+# Upstream subscription loader
+# Loads subscription information into the Candlepin HostedTest adapter for testing.
+#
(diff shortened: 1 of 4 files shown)


Januson (approved):
Looks good to me.

nikosmoum (approved):
This looks like it does the same thing with this other script in the provided_products branch: https://github.com/candlepin/candlepin/blob/feature/move_provided_product/server/bin/upstream_subscription_injector.rb
If they do, ideally we should reconcile the them (keep one of the two, and merge any changes that make sense from both into it) once (or right before) the provided_products branch gets merged.

PR #2792, opened 2020-08-31
1872357: [RFE] Filter syspurpose SLA values available to an Owner if the SLA is exempt

Januson:


server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -2106,6 +2106,11 @@ public SystemPurposeAttributesDTO getSyspurpose(
 
         for (Product p : products) {
             for (SystemPurposeAttributeType type : SystemPurposeAttributeType.values()) {
+                boolean slaExempt = Boolean.parseBoolean(
+                    p.getAttributeValue(Product.Attributes.SUPPORT_LEVEL_EXEMPT));
+                if (slaExempt) {
+                    continue;
+                }
```
nikosmoum, line 2113: This will make it so that every single syspurpose attribute of this product will get ignored when the product has Product.Attributes.SUPPORT_LEVEL_EXEMPT set to true. We only want the SLA attribute to be ignored, so this if statement should be `if (type == SystemPurposeAttributeType.SERVICE_LEVEL && slaExempt)`

nikosmoum (changes requested):
One issue to be fixed, and we can merge

server/src/main/java/org/candlepin/resource/OwnerResource.java
Januson: Done.

PR #2795, opened 2020-09-03
ENT-2817 : Fix candlepin docker image Jenkins build

wolfdale:
Docker build occasionally fails while dumping the database. 
Changes 
- Increased timeout for dumping database, proper cleanup of docker swam & stack in case of failure.
- Upgraded Tito to newer version (0.6.15) which is available with the fixes needed by Candlepin, therefore no need to manually install older version.


nikosmoum (approved):
wow, so the issue was only that the waiting time wasn't enough? ACK

wolfdale:
It happens only sometimes. I was able to reproduce this issue only once or twice, the recent docker build was passed even without increasing the timeout.

PR #2796, opened 2020-09-04
[F] Use RESTEasy's MultipartInput for multipart/form-data

nikosmoum:
- Add the default api/formParams mustache templates that ship
  with openapi-generator for the JavaJaxRS/spec/ generator type.

- The openapi-generator type we use (pure JAX-RS spec) has no
  knowledge of our JAX-RS implementation (RESTEasy) and by
  default generates '@FormParam InputStream' as the input type
  when declaring to consume 'multipart/form-data' for file
  uploads.

  This change alters the default templates used for code
  generation to use RESTEasy's 'MultipartInput' utility class
  instead of '@FormParam InputStream' for file uploads.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -443,6 +443,7 @@ project(":api") {
         compile 'javax.validation:validation-api'
         compile 'javax.ws.rs:javax.ws.rs-api'
         compile 'io.swagger:swagger-annotations'
+        compile 'org.jboss.resteasy:resteasy-multipart-provider'
     }
 
     sourceSets.main.java.srcDirs = ["${buildDir}/generated/src/gen/java"]
--- a/buildSrc/src/main/resources/templates/api.mustache
+++ b/buildSrc/src/main/resources/templates/api.mustache
@@ -1,0 +1,30 @@
+package {{package}};
+
+{{#imports}}import {{import}};
+{{/imports}}
+
+import javax.ws.rs.*;
+import javax.ws.rs.core.Response;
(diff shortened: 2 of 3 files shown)


nikosmoum:
Note: This is to fix the issues with the OwnerResource.importManifest/importManifestAsync endpoints in the ENT-1848 task, which use multipart/form-data.

barnabycourt:
@nikosmoum Is there a corresponding PR to fix this upstream?

nikosmoum:
@barnabycourt No, because the openapi generator type we use is based on the JAX-RS spec (see [here](https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources/JavaJaxRS/spec)), and not on any specific implementation (resteasy/jersey/etc), so applying this fix that is resteasy-specific would not be acceptable. 

They do have another generator type that is specific to the resteasy implementation (see [here](https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources/JavaJaxRS/resteasy)), which already works in it's own way (it generates `MultipartFormDataInput`) and doesn't need a fix, but for us to switch to it would be a massive task (it generates code in a very different way, so we'd have to change all of the resource code we've already ported again, and who knows what other issues we might find).

sonalidhome (approved):
Changes look good to me. It's working for `importManifest` to consume `multipart/form-data`.

PR #2799, opened 2020-09-08
ENT-2806: Only add content from active pools on the SCA certificate

wolfdale:
Changes 
- Updated ContentAccessManager to only add content from active pools on the SCA certificate.
- Added Spec and unit test cases

### DIFF
--- a/server/spec/content_access_spec.rb
+++ b/server/spec/content_access_spec.rb
@@ -36,6 +36,7 @@ describe 'Content Access' do
 
     @product = create_product('test-product', 'some product')
     @cp.add_content_to_product(@owner['key'], @product['id'], @content_id)
+    @cp.create_pool(@owner['key'], @product['id'], {:quantity => 10})
   end
 
   it "does allow addition of the content access level" do
@@ -537,11 +538,12 @@ describe 'Content Access' do
   it 'should not auto-attach when org_environment is set for owner' do
     mkt_product1 = create_product(random_string('product'), random_string('product'), {:owner => @owner['key']})
     eng_product = create_product(random_string('product'), random_string('product'), {:owner => @owner['key']})
-    p1 = create_pool_and_subscription(@owner['key'], mkt_product1.id, 10, [eng_product.id])
+    @cp.create_pool(@owner['key'], mkt_product1['id'], {:quantity => 10})
 
     installed = [{'productId' => eng_product.id, 'productName' => eng_product['name']}]
 
-    consumer_cp = consumer_client(@user, @consumername, type=:system, username=nil, facts= {'system.certificate_version' => '3.3'})
+    consumer_cp = consumer_client(@user, @consumername, type=:system, username=nil,
+        facts= {'system.certificate_version' => '3.3'})
     consumer_cp.update_consumer({:installedProducts => installed})
 
     lambda do
@@ -578,7 +580,10 @@ describe 'Content Access' do
         @owner['key'], "cname-c2", 'test-content-c2', random_string("clabel"), "ctype", "cvendor",
         {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
     @cp.add_content_to_product(@owner['key'], product['id'], content_c2['id'], false)
-    @consumer = consumer_client(@user, @consumername, type=:system, username=nil, facts= {'system.certificate_version' => '3.3'})
+
+    @cp.create_pool(@owner['key'], product['id'], {:quantity => 10})
+    @consumer = consumer_client(@user, @consumername, type=:system, username=nil,
+        facts= {'system.certificate_version' => '3.3'})
     certs = @consumer.list_certificates
 
     expect(certs.length).to eq(1)
@@ -591,10 +596,10 @@ describe 'Content Access' do
     # Check content status
     json_body['products'][0]['content'].each do |content|
       if content.id == content_c1.id
-        expect(content.enabled).to be_nil
+          expect(content.enabled).to be_nil
       end
       if content.id == content_c2.id
-        expect(content.enabled).to eq(false)
+          expect(content.enabled).to eq(false)
       end
     end
   end
@@ -602,7 +607,7 @@ describe 'Content Access' do
   it 'filter out content not promoted to environment when owner is in SCA mode' do
     @env = @org_admin.create_environment(@owner['key'], 'testenv1', "My Test Env 1", "For test systems only.")
     consumer = @org_admin.register(random_string('testsystem'), :system, nil,
-      {'system.certificate_version' => '3.1'}, nil, nil, [], [], @env['id'])
+        {'system.certificate_version' => '3.1'}, nil, nil, [], [], @env['id'])
 
     expect(consumer['environment']).to_not be_nil
 
@@ -617,13 +622,10 @@ describe 'Content Access' do
 
     # Promote content with enabled false
     job = @org_admin.promote_content(@env['id'],
-    [{
-      :contentId => content['id'],
-      :enabled => false ,
-     }])
+    [{:contentId => content['id'], :enabled => false , }])
     wait_for_job(job['id'], 15)
 
-    pool = create_pool_and_subscription(@owner['key'], product['id'], 10)
+    pool = @cp.create_pool(@owner['key'], product['id'], {:quantity => 10})
     ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
     value = extension_from_cert(ent['certificates'][0]['cert'], "1.3.6.1.4.1.2312.9.6")
 
@@ -664,7 +666,10 @@ describe 'Content Access' do
     @cp.add_content_to_product(@owner['key'], product_1['id'], content_c3['id'], false)
     @cp.add_content_to_product(@owner['key'], product_2['id'], content_c3['id'], false)
 
-    @consumer = consumer_client(@user, @consumername, type=:system, username=nil, facts= {'system.certificate_version' => '3.3'})
+    @cp.create_pool(@owner['key'], product_1['id'], {:quantity => 10})
+    @cp.create_pool(@owner['key'], product_2['id'], {:quantity => 10})
+    @consumer = consumer_client(@user, @consumername, type=:system, username=nil,
+        facts= {'system.certificate_version' => '3.3'})
     certs = @consumer.list_certificates
 
     expect(certs.length).to eq(1)
@@ -676,13 +681,13 @@ describe 'Content Access' do
 
     json_body['products'][0]['content'].each do |content|
       if content.id == content_c1.id
-        expect(content.enabled).to be_nil()
+          expect(content.enabled).to be_nil()
       end
       if content.id == content_c2.id
-        expect(content.enabled).to be_nil()
+          expect(content.enabled).to be_nil()
       end
       if content.id == content_c3.id
-        expect(content.enabled).to eq(false)
+          expect(content.enabled).to eq(false)
       end
     end
   end
@@ -697,7 +702,7 @@ describe 'Content Access' do
     @cp.add_content_to_product(@owner['key'], product['id'], content_c1['id'], true)
     consumer = consumer_client(@user, @consumername, type=:system, username=nil,
         facts= {'system.certificate_version' => '3.3'})
-    pool = create_pool_and_subscription(@owner['key'], product['id'], 10)
+    pool = @cp.create_pool(@owner['key'], product['id'], {:quantity => 10})
     consumer.consume_pool(pool['id'], {:quantity => 1})[0]
 
     cert_export = StandardExporter.new
@@ -722,4 +727,96 @@ describe 'Content Access' do
 
     cert_export.cleanup
   end
+
+  it 'should only add content from active pools on the SCA certificate' do
+    product_1 = create_product('test-product-p1', 'some product-p1')
+    product_2 = create_product('test-product-p2', 'some product-p2')
+
+    content_c1 = @cp.create_content(
+        @owner['key'], "cname-c1", 'test-content-c1', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], product_1['id'], content_c1['id'], true)
+
+    content_c2 = @cp.create_content(
+        @owner['key'], "cname-c2", 'test-content-c2', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], product_2['id'], content_c2['id'], true)
+
+    @cp.create_pool(@owner['key'], product_2['id'], {:quantity => 10})
+
+    @consumer = consumer_client(@user, @consumername, type=:system, username=nil,
+        facts= {'system.certificate_version' => '3.3'})
+    certs = @consumer.list_certificates
+
+    expect(certs.length).to eq(1)
+
+    cert = certs[0]['cert']
+    json_body = extract_payload(cert)
+    expect(json_body['products'][0]['content'].length).to eq(2)
+
+    # Make sure that content c1 is not present in cert,
+    # since product_1 does not have active pool
+    json_body['products'][0]['content'].each do |content|
+        expect(content.id).to_not eq(content_c1.id)
+    end
+  end
+
+  it 'should include content from all products associated with active pool to SCA cert' do
+    mkt_product1 = create_product(random_string('productp1'), random_string('product'),
+        {:owner => @owner['key']})
+    eng_product = create_product(random_string('productp2'), random_string('product'),
+        {:owner => @owner['key']})
+    derived_product = create_product(random_string('productp4'), random_string('product'),
+        {:owner => @owner['key']})
+    dev_eng_product = create_product(random_string('productp3'), random_string('product'),
+        {:owner => @owner['key']})
+
+    # Content enabled = true
+    content_c1 = @cp.create_content(
+        @owner['key'], "content_c1", 'test-content-c1', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], eng_product['id'], content_c1['id'], true)
+
+    content_c2 = @cp.create_content(
+        @owner['key'], "content_c2", 'test-content-c2', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], mkt_product1['id'], content_c2['id'], true)
+
+    content_c3 = @cp.create_content(
+        @owner['key'], "content_c3", 'test-content-c3', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], derived_product['id'], content_c3['id'], true)
+
+    content_c4 = @cp.create_content(
+        @owner['key'], "content_c4", 'test-content-c4', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], dev_eng_product['id'], content_c4['id'], true)
+
+    @cp.create_pool(@owner['key'], mkt_product1['id'],
+        {:quantity => 10, :provided_products => [eng_product['id']],
+        :derived_product_id => derived_product['id'], :derived_provided_products => [dev_eng_product['id']]
+    })
+
+    @consumer = consumer_client(@user, @consumername, type=:system, username=nil,
+        facts= {'system.certificate_version' => '3.3'})
+    certs = @consumer.list_certificates
+
+    expect(certs.length).to eq(1)
+
+    cert = certs[0]['cert']
+    json_body = extract_payload(cert)
+
+    expect(json_body['products'][0]['content'].length).to eq(5)
+
+    returned_uuids = []
+    json_body['products'][0]['content'].each do |content|
+        returned_uuids << content['id']
+    end
+
+    expect(returned_uuids).to include(content_c1.id)
+    expect(returned_uuids).to include(content_c2.id)
+    expect(returned_uuids).to include(content_c3.id)
+    expect(returned_uuids).to include(content_c4.id)
+  end
+
 end
--- a/server/src/main/java/org/candlepin/controller/ContentAccessManager.java
+++ b/server/src/main/java/org/candlepin/controller/ContentAccessManager.java
@@ -17,7 +17,6 @@ package org.candlepin.controller;
 import org.candlepin.audit.EventSink;
 import org.candlepin.common.config.Configuration;
 import org.candlepin.config.ConfigProperties;
-import org.candlepin.model.CandlepinQuery;
 import org.candlepin.model.CertificateSerial;
 import org.candlepin.model.CertificateSerialCurator;
 import org.candlepin.model.Consumer;
@@ -34,13 +33,12 @@ import org.candlepin.model.EnvironmentContent;
 import org.candlepin.model.EnvironmentCurator;
 import org.candlepin.model.KeyPairCurator;
 import org.candlepin.model.Owner;
+import org.candlepin.model.OwnerContentCurator;
 import org.candlepin.model.OwnerCurator;
 import org.candlepin.model.OwnerEnvContentAccess;
 import org.candlepin.model.OwnerEnvContentAccessCurator;
-import org.candlepin.model.OwnerProductCurator;
 import org.candlepin.model.Pool;
 import org.candlepin.model.Product;
-import org.candlepin.model.ProductContent;
 import org.candlepin.pki.PKIUtility;
 import org.candlepin.pki.X509ByteExtensionWrapper;
 import org.candlepin.pki.X509ExtensionWrapper;
@@ -189,6 +187,7 @@ public class ContentAccessManager {
     private KeyPairCurator keyPairCurator;
     private CertificateSerialCurator serialCurator;
     private OwnerCurator ownerCurator;
+    private OwnerContentCurator ownerContentCurator;
     private ContentAccessCertificateCurator contentAccessCertificateCurator;
     private X509V3ExtensionUtil v3extensionUtil;
     private OwnerEnvContentAccessCurator ownerEnvContentAccessCurator;
@@ -196,7 +195,6 @@ public class ContentAccessManager {
     private ConsumerTypeCurator consumerTypeCurator;
     private EnvironmentCurator environmentCurator;
     private ContentAccessCertificateCurator contentAccessCertCurator;
-    private OwnerProductCurator ownerProductCurator;
     private EventSink eventSink;
 
     private boolean standalone;
@@ -210,12 +208,12 @@ public class ContentAccessManager {
         KeyPairCurator keyPairCurator,
         CertificateSerialCurator serialCurator,
         OwnerCurator ownerCurator,
+        OwnerContentCurator ownerContentCurator,
         OwnerEnvContentAccessCurator ownerEnvContentAccessCurator,
         ConsumerCurator consumerCurator,
         ConsumerTypeCurator consumerTypeCurator,
         EnvironmentCurator environmentCurator,
         ContentAccessCertificateCurator contentAccessCertCurator,
-        OwnerProductCurator ownerProductCurator,
         EventSink eventSink) {
 
         this.config = Objects.requireNonNull(config);
@@ -225,13 +223,13 @@ public class ContentAccessManager {
         this.serialCurator = Objects.requireNonNull(serialCurator);
         this.v3extensionUtil = Objects.requireNonNull(v3extensionUtil);
         this.ownerCurator = Objects.requireNonNull(ownerCurator);
+        this.ownerContentCurator = Objects.requireNonNull(ownerContentCurator);
         this.ownerEnvContentAccessCurator = Objects.requireNonNull(ownerEnvContentAccessCurator);
         this.consumerCurator = Objects.requireNonNull(consumerCurator);
         this.consumerTypeCurator = Objects.requireNonNull(consumerTypeCurator);
         this.environmentCurator = Objects.requireNonNull(environmentCurator);
         this.contentAccessCertCurator = Objects.requireNonNull(contentAccessCertCurator);
         this.eventSink = Objects.requireNonNull(eventSink);
-        this.ownerProductCurator = Objects.requireNonNull(ownerProductCurator);
         this.standalone = this.config.getBoolean(ConfigProperties.STANDALONE, true);
     }
 
@@ -509,7 +507,6 @@ public class ContentAccessManager {
     private byte[] createContentAccessDataPayload(Owner owner, Environment environment) throws IOException {
         // fake a product dto as a container for the org content
         Set<Product> containerSet = new HashSet<>();
-        CandlepinQuery<Product> ownerProduct = ownerProductCurator.getProductsByOwner(owner);
         Set<String> entitledProductIds = new HashSet<>();
         List<org.candlepin.model.dto.Product> productModels = new ArrayList<>();
         Map<String, EnvironmentContent> promotedContent = getPromotedContent(environment);
@@ -524,21 +521,9 @@ public class ContentAccessManager {
         containerSet.add(container);
         container.setId("content_access");
         container.setName(" Content Access");
-        Map<String, Boolean> contentEnabledMap = new HashMap<>();
 
-        for (Product product : ownerProduct) {
-            for (ProductContent pc : product.getProductContent()) {
-                if (!contentEnabledMap.containsKey(pc.getContent().getUuid())) {
-                    container.addContent(pc.getContent(), pc.isEnabled());
-                    contentEnabledMap.put(pc.getContent().getUuid(), pc.isEnabled());
-                }
-                else if (pc.isEnabled() && !contentEnabledMap.get(pc.getContent().getUuid())) {
-                    // Overwrite the content
-                    container.addContent(pc.getContent(), pc.isEnabled());
-                    contentEnabledMap.put(pc.getContent().getUuid(), pc.isEnabled());
-                }
-            }
-        }
+        this.ownerContentCurator.getActiveContentByOwner(owner.getId())
+            .forEach((content, enabled) -> container.addContent(content, enabled));
 
         emptyConsumer.setEnvironment(environment);
         emptyEnt.setPool(emptyPool);
--- a/server/src/main/java/org/candlepin/model/OwnerContentCurator.java
+++ b/server/src/main/java/org/candlepin/model/OwnerContentCurator.java
@@ -14,6 +14,7 @@
  */
 package org.candlepin.model;
 
+import com.google.common.collect.Iterables;
 import com.google.inject.persist.Transactional;
 
 import org.hibernate.Criteria;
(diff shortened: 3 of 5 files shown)


barnabycourt:
Some products (virtual datacenter, instance based) can have different content on the derived product from the master product. In this case we need to accumulate the content from both the product on the master pool & any derived provided products as well.

wolfdale:
Changes made to accumulate the content from both the product/derived product on the master pool & any derived or provided products.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -2006,4 +2006,217 @@ public void clearPoolSourceEntitlementRefs(Iterable<String> poolIds) {
...
+            for (Object[] cols : (List<Object[]>) query.getResultList()) {
+                Set<String> providedProducts = providedProductMap.get((String) cols[0]);
+
+                if (providedProducts == null) {
+                    providedProducts = new HashSet<>();
+                    providedProductMap.put((String) cols[0], providedProducts);
+                }
+
+                providedProducts.add((String) cols[1]);
+            }
```
Januson, line 2065: This logic is repeated across multiple methods. We should refactor this into reusable pieces.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -2006,4 +2006,217 @@ public void clearPoolSourceEntitlementRefs(Iterable<String> poolIds) {
...
+            else {
+                builder.append(" p.id IN (:pids)");
+                query = this.getEntityManager().createQuery(builder.toString())
+                    .setParameter("pids", poolIds);
+            }
+
+            for (Object[] cols : (List<Object[]>) query.getResultList()) {
+                Set<String> providedProducts = providedProductMap.get((String) cols[0]);
+
+                if (providedProducts == null) {
```
Januson, line 2059: This could be Map.computeIfAbsent.

Januson (changes requested):
Overall the changes look good. I have just one remark regarding the duplication in PoolCurator.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Refactored the logic.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Done.

Januson (approved):

server/spec/content_access_spec.rb
```diff
@@ -682,4 +687,95 @@ def read_json_file(filename)
...
+      expect(content.id).to_not eq(content_c1.id)
+    end
+  end
+
+  it 'should include content from all products associated with active pool to SCA cert' do
+    mkt_product1 = create_product(random_string('productp1'), random_string('product'),
+      {:owner => @owner['key']})
+    eng_product = create_product(random_string('productp2'), random_string('product'),
+      {:owner => @owner['key']})
+    derived_product =create_product(random_string('productp4'), random_string('product'),
```
Ceiu, line 728: Missing whitespace after the =

server/spec/content_access_spec.rb
```diff
@@ -682,4 +687,95 @@ def read_json_file(filename)
...
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], derived_product['id'], content_c3['id'], true)
+
+    content_c4 = @cp.create_content(
+        @owner['key'], "content_c4", 'test-content-c4', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], dev_eng_product['id'], content_c4['id'], true)
+
+
+    create_pool_and_subscription(@owner['key'], mkt_product1.id,
```
Ceiu, line 755: We should try to be consistent with the way we do line continuation and indentation. This statement here trails off early and is partially indented on the third line of the statement, whereas the statement on line 759 below is heavily indented immediately, and the statements above on line 739, 744, and 749 are all indented with 4 spaces on each line.

This kind of inconsistency destroys auto-alignment detection in IDEs, and makes it harder to read at a glance. Let's pick one level of indentation (either 4 or 2 spaces), and then follow a single pattern for indentation and line continuation.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1684,26 +1685,40 @@ public Boolean providesDerived(Pool pool, String derivedProvidedProductId) {
...
-     * Fetches a mapping of pool IDs to sets of product IDs representing the provided products of
+     * Fetches a mapping of pool IDs to sets of product IDs or UUIDs representing the provided products of
      * the given pool. The returned map will only contain mappings for pools specified in the given
      * collection of pool IDs.
      *
      * @param poolIds
-     *  A collection of pool IDs for which to fetch provided product IDs
+     *  A collection of pool IDs for which to fetch provided product IDs or UUIDs
+     *
+     * @param fetchMode
```
Ceiu, line 1699: Nack on this.

This completely changes the typing on our output, which is *very* bad. Consistency is more important than code reuse.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1735,14 +1750,10 @@ public Boolean providesDerived(Pool pool, String derivedProvidedProductId) {
...
             for (Object[] cols : (List<Object[]>) query.getResultList()) {
-                Set<String> providedProducts = providedProductMap.get((String) cols[0]);
-
-                if (providedProducts == null) {
-                    providedProducts = new HashSet<>();
-                    providedProductMap.put((String) cols[0], providedProducts);
-                }
-
-                providedProducts.add((String) cols[1]);
+                providedProductMap.computeIfAbsent((String) cols[0],
```
Ceiu, line 1753: Incorrect use of ```computeIfAbsent```. The method will generate a new value if not present *and* return whatever ends up in the field in either case. This saves us an interaction with the map by making the fetch, generation, and storage operations all atomic and packaged in a single call. You can then use method chaining and the knowledge that you'll never get a null value out of the map to eliminate the need for explicit declaration. It should look like this:

```
providedProductMap.computeIfAbsent((String) cols[0], key -> new HashSet<>())
    .add((String) cols[1]);
```

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1772,26 +1783,39 @@ public Boolean providesDerived(Pool pool, String derivedProvidedProductId) {
...
+     *  Boolean field.
+     *  True  -> Returns mapping of Pool IDs & provided product IDs
+     *  False -> Returns mapping of Pool IDs & provided product UUIDs
      *
      * @return
-     *  A mapping of pool IDs to provided product IDs
+     *  A mapping of pool IDs to provided product IDs or UUIDs
      */
-    public Map<String, Set<String>> getDerivedProvidedProductIdsByPoolIds(Collection<String> poolIds) {
+    public Map<String, Set<String>> getDerivedProvidedProductByPoolIds(Collection<String> poolIds,
```
Ceiu, line 1805: Again, nack on these changes -- we don't want to completely swing a method's operation and output on a single boolean input. Rewrite the method and make changes as necessary.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -2006,4 +2026,93 @@ public void clearPoolSourceEntitlementRefs(Iterable<String> poolIds) {
...
+     * Fetches a collection of active pool IDs mapped to a given owner. If no such pool
+     * are present, an empty collection will be returned.
+     *
+     * @param ownerId
+     *  The ID of the owner for which to fetch active Pool IDs
+     *
+     * @return
+     *  A collection of active Pool IDs belonging to the given owner.
+     */
+    public Collection<String> getActivePoolByOwnerId(String ownerId) {
```
Ceiu, line 2040: This should be named "getActivePool*Ids*ByOwnerId", as it is fetching multiple pool IDs, not a single pool.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -2006,4 +2026,93 @@ public void clearPoolSourceEntitlementRefs(Iterable<String> poolIds) {
...
+        Date currentDate = new Date();
+        String jpql = "SELECT p.id FROM Pool p WHERE p.owner.id = " +
+            " :owner_id AND p.endDate >= :endDate AND p.startDate <= :startDate";
+
+        List<String> uuids = this.getEntityManager()
```
Ceiu, line 2045: This should be "ids" rather than "uuids"

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -2006,4 +2026,93 @@ public void clearPoolSourceEntitlementRefs(Iterable<String> poolIds) {
...
+     *
+     * @param ownerId
+     *  The ID of the owner for which to fetch active Pool IDs
+     *
+     * @return
+     *  A collection of active Pool IDs belonging to the given owner.
+     */
+    public Collection<String> getActivePoolByOwnerId(String ownerId) {
+        Date currentDate = new Date();
+        String jpql = "SELECT p.id FROM Pool p WHERE p.owner.id = " +
```
Ceiu, line 2042: Try not to put line breaks for continuations in the middle of a statement or clause. There's enough space for the :owner_id part of the first WHERE clause, or the WHERE portion should move down to the line with it.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -2006,4 +2026,93 @@ public void clearPoolSourceEntitlementRefs(Iterable<String> poolIds) {
...
+     * the given pool. The returned map will only contain mappings for pools specified in the given
+     * collection of pool IDs.
+     *
+     * @param poolIds
+     *  A collection of pool IDs for which to fetch product & derived product UUIDs.
+     *
+     * @return
+     *  A mapping of pool IDs to product & derived product UUIDs
+     */
+    public Map<String, Set<String>> getProductAndDerivedProductUuidsByPoolIds(Collection<String> poolIds) {
```
Ceiu, line 2066: Let's try to find a better way to write this. This method looks heavily derived from the bulkSQL[update|delete] methods from AbstractHibernateCurator which should be flagged as deprecated if they're not already. Building raw SQL can be dangerous, and runs into potential problems with parameter limits (as this implementation does here).

server/src/main/java/org/candlepin/model/ProductCurator.java
```diff
@@ -505,4 +505,25 @@ public boolean productHasSubscriptions(Owner owner, Product product) {
...
+    /**
+     * Fetches products by product UUIDs.
+     *
+     * @param productUuids
+     *  The Collection of product UUIDs
+     *
+     * @return
+     *  Query collection of Products
+     */
+    public CandlepinQuery<Product> getProductsByProductUuids(Collection<String> productUuids) {
```
Ceiu, line 518: This is a functional duplication of ```listAllByUuid```

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -509,7 +518,35 @@ private String createDN(Consumer consumer, Owner owner) {
...
+            if (derivedProvidedProducts.get(poolId) != null) {
+                productUuids.addAll(derivedProvidedProducts.get(poolId));
+            }
+
+            if (products.get(poolId) != null) {
+                productUuids.addAll(products.get(poolId));
+            }
+        }
+
+        CandlepinQuery<Product> productsToConsider =
```
Ceiu, line 547: CandlepinQuery is deprecated, we shouldn't introduce new code that uses it.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -509,7 +518,35 @@ private String createDN(Consumer consumer, Owner owner) {
     private byte[] createContentAccessDataPayload(Owner owner, Environment environment) throws IOException {
         // fake a product dto as a container for the org content
         Set<Product> containerSet = new HashSet<>();
-        CandlepinQuery<Product> ownerProduct = ownerProductCurator.getProductsByOwner(owner);
+        Set<String> productUuids = new HashSet<>();
+
+        Collection<String> activePoolIds = poolCurator.getActivePoolByOwnerId(owner.getId());
```
Ceiu, line 523: This is an expensive way to fetch the product information for active pools. If the goal here is to fetch products for active pools, we should do exactly that: fetch active pools and then pull the products from them in code (functionally identical to what we're doing here), or write a new curator method that fetches only products from active entitlements of a given organization.

Ceiu (changes requested):

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
wolfdale: Removed.

server/src/main/java/org/candlepin/model/ProductCurator.java
wolfdale: Removed.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Noted. Changed the logic here.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Noted

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Removed.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Removed.

server/spec/content_access_spec.rb
wolfdale: Corrected.

server/spec/content_access_spec.rb
wolfdale: Noted. Corrected the indentation.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Removed this piece of code from here.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Removed this piece of code from here.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Removed this piece of code from here.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
wolfdale: Changed the logic to pull products & it's content directly in the code from active pools.

Not sure if we can use active entitlement to pull product's contents here, since active pool can exists without any active entitlement.

nikosmoum:

nikosmoum:

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -2006,4 +2008,115 @@ public void clearPoolSourceEntitlementRefs(Iterable<String> poolIds) {
...
+        predicates.add(criteriaBuilder.lessThanOrEqualTo(pool.get("startDate"), currentDate));
+        Predicate[] predicateArray = new Predicate[predicates.size()];
+        query.where(predicates.toArray(predicateArray));
+        List<javax.persistence.criteria.Order> orderList = new ArrayList();
+        orderList.add(criteriaBuilder.desc(pool.get("startDate")));
+        query.orderBy(orderList);
+        int offset = 0;
+
+        if (blockCount > 1) {
+            while (blockCount > 0) {
```
Ceiu, line 2046: The blocking here probably isn't critical in our general case, but it also doesn't hurt anything.

However, the code goes out of its way to have special handling for the 1 block case, when that's unnecessary -- even with only a single block, setting the max results and first result in the query won't change the performance of the query enough to justify having separate branches for each case.

Additionally, the block count bit is largely unnecessary. Once you have the element count, block size, and keep track of the offset, the counting of blocks becomes extraneous -- just check that your offset is lower than your element count. Even without the element count, you can achieve the same level of performance by just querying until you get a batch that has zero elements.

That said, fetching the element count is basically the same query here except without returning the pool information -- the DB work is effectively the same. Moreover, the collection processing below runs into not one, but two n+1 problems of doing a ton of extra queries to fetch the product and content collections. I would bet this is overall slower than just fetching all of the "active" pool IDs, then using those to get the related product UUIDs, and then using that to get the content objects. You'd get the whole thing done in three queries, rather than the 2n^p+b+1 queries we have here.

Ceiu (changes requested):
It's a step in the right direction, but we need to get that active content query hammered out.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Yes it is slower, firing extra queries under the hood to fetch product & contents, this is what I was afraid of hence I chose to execute queries to get Product UUIDs.

> If the goal here is to fetch products for active pools, we should do exactly that: fetch active pools and then pull the products from them in code.

From your previous comment I thought you meant to get the products (and content) of active pool right from the code. 
Maybe we can go back to previous approach of having direct queries to fetch all associated products UUIDs of active pools and then fetch the product's content. What are your thoughts on that ? Also do we care if fetch these UUIDs in batch or not ?

Ceiu: Sorry, Github isn't showing me the old code, so I'm going from memory here...

Anyway, IIRC, the old code did pool iteration as a one-off in the method itself, so my assumption was that it would be augmented to simply pull the objects off it there and we'd drop the custom SQL builder code derived from a deprecated method that had a very dangerous oversight in it. My overall stance is that if we're going to add a method to do this operation at the curator level, it's something that should be generalized and as efficient as reasonably possible since it will be available to the rest of the code base at that point, and we might attempt to use it elsewhere in the future.

Once we are working at the curator level, the three queries I mentioned above can be used to fetch the information quicker than object iteration, and with minimal fetching of DB data. One of them will require raw SQL, but we should avoid constructing SQL statements via string concatenation without careful consideration of all the problems that introduces.

Regarding the batching question, there are two cases we want to do requests in batches:
1) if the number of input parameters would exceed the limits of either an IN operator or the overall query parameter limit (this is the most common scenario)
2) optimizing lookups when fetching lazy collections off of certain entities (rare thus far in Candlepin)

I believe in this case, once we move the operation to the curator level, we will need to batch each query due to the nature of the operation (N pools, M products, O content). The easiest way to do this is with the partition method provided by AbstractHibernateCurator, or you can use the partitioning functionality directly. There are other ways to batch, but I would recommend them only in specific cases where your data store(s) or operations don't lend themselves nicely to the partitioning functionality.

Hope that helps, and my apologies for seemingly going back and forth here.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -2006,4 +2008,115 @@ public void clearPoolSourceEntitlementRefs(Iterable<String> poolIds) {
...
+                List<Pool> pools = this.getEntityManager()
+                    .createQuery(query)
+                    .setMaxResults(blockCount)
+                    .setFirstResult(offset)
+                    .setMaxResults(blockSize)
```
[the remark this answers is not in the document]
wolfdale: Not needed anymore.

server/src/main/java/org/candlepin/model/PoolCurator.java
wolfdale: Changes made to cut back the complex batching logic & to simplify the SQL queries with JPA criteria queries. I broke the steps into multiple parts of pulling the active PoolIDs, then product & derived product UUIDs, provided product UUIDs and finally derived provided product UUIDs. In between these queries we process for product content so that we can avoid holding everything in-memory.

Ceiu:
There's still a bunch of extra work being done here and a number of extra queries being performed while also separating this logic across three or four different classes.

The goal here is to answer a single question: what content is currently active for a given owner?

We can do that in a single curator (OwnerContentCurator, as the question pertains specifically to owner-specific content), and collect the entirety of the results in three queries + partitioning on result size:

```
    public Set<Content> getActiveContentByOwner(String ownerId) {
        EntityManager entityManager = this.getEntityManager();
        Date now = new Date();

        String jpql = "SELECT p.id FROM Pool p " +
            "WHERE p.owner.id = :owner_id AND p.startDate <= :start_date AND p.endDate >= :end_date";

        List<String> poolIds = entityManager.createQuery(jpql, String.class)
            .setParameter("owner_id", ownerId)
            .setParameter("start_date", now)
            .setParameter("end_date", now)
            .getResultList();

        // Use the pool IDs to select all related product UUIDs
        // Note: we have to start partitioning now to safely handle any number of active pools

        String sql = "SELECT pool.product_uuid FROM cp_pool pool WHERE pool.id IN (:pool_ids) " +
            "UNION " +
            "SELECT pool.derived_product_uuid FROM cp_pool pool WHERE pool.id IN (:pool_ids) " +
            "UNION " +
            "SELECT pp.product_uuid FROM cp2_pool_provided_products pp WHERE pp.pool_id IN (:pool_ids) " +
            "UNION " +
            "SELECT dpp.product_uuid FROM cp2_pool_derprov_products dpp WHERE dpp.pool_id IN (:pool_ids) ";

        Query prodQuery = entityManager.createNativeQuery(sql);
        Set<String> prodIds = new HashSet<>();

        for (List<String> block : this.partition(poolIds)) {
            prodIds.addAll(prodQuery.setParameter("pool_ids", block)
                .getResultList());
        }

        // Use the product UUIDs to select all related enabled content

        jpql = "SELECT pc.content FROM ProductContent pc " +
            "WHERE pc.product.uuid IN (:product_uuids) AND pc.enabled = true";

        TypedQuery<Content> contentQuery = entityManager.createQuery(jpql, Content.class);
        HashSet<Content> activeContent = new HashSet<>();

        for (List<String> block : this.partition(prodIds)) {
            activeContent.addAll(contentQuery.setParameter("product_uuids", block)
                .getResultList());
        }

        return activeContent;
    }
```

This reduces the code in the ContentAccessManager to a simple:
```
    .forEach(content -> container.addContent(content));
```

The overall additional code is under 50 lines, we limit our base query count to three, our logic is encapsulated in a single location, and we're not juggling all the extra mappings and whatnot.

wolfdale:
@Ceiu  Can you please take a look again. Quick note - goal is to fetch all the products content out of active pools regardless of content is enabled or not. So I removed ``pc.enabled = true`` condition from 3rd query.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -559,6 +543,32 @@ else if (pc.isEnabled() && !contentEnabledMap.get(pc.getContent().getUuid())) {
...
+     * To process product content and added it to container product.
+     *
+     * @param container
+     *  Product container
+     *
+     * @param ownerId
+     *  Id of an owner
+     *
+     */
+    private void processOwnerContent(Product container, String ownerId) {
```
Ceiu, line 556: Ah ha. I remember this discussion now.

So we need to include all "active" content for a given org, but we need to flag it as enabled properly, resolving any conflicts to enabled.

That complicates things a bit, but is something we can still solve efficiently.

Since ProductContent is a join object, we don't want to fetch it directly if we can help it -- especially since the object we *actually* care about is the Content instance, which we have to fetch even if we try to use the ProductContent mutators of Product directly. Using ProductContent like this results in another N+1 issue, where examining the content field results in another query. Running the spec test for this method, for instance, results in the following queries:

```
select pool0_.id as col_0_0_ from cp_pool pool0_ where pool0_.owner_id=? and pool0_.startDate<=? and pool0_.endDate>=?

SELECT pool.product_uuid FROM cp_pool pool WHERE pool.id IN (?, ?) UNION SELECT pool.derived_product_uuid FROM cp_pool pool WHERE pool.id IN (?, ?) UNION SELECT pp.product_uuid FROM cp2_pool_provided_products pp WHERE pp.pool_id IN (?, ?) UNION SELECT dpp.product_uuid FROM cp2_pool_derprov_products dpp WHERE dpp.pool_id IN (?, ?) 

select productcon0_.id as id1_12_, productcon0_.created as created2_12_, productcon0_.updated as updated3_12_, productcon0_.content_uuid as content_5_12_, productcon0_.enabled as enabled4_12_, productcon0_.product_uuid as product_6_12_ from cp2_product_content productcon0_ where productcon0_.product_uuid in (? , ? , ?)

select content0_.uuid as uuid1_1_0_, content0_.created as created2_1_0_, content0_.updated as updated3_1_0_, content0_.arches as arches4_1_0_, content0_.contentUrl as contentU5_1_0_, content0_.entity_version as entity_v6_1_0_, content0_.gpgUrl as gpgUrl7_1_0_, content0_.content_id as content_8_1_0_, content0_.label as label9_1_0_, content0_.locked as locked10_1_0_, content0_.metadataExpire as metadat11_1_0_, content0_.name as name12_1_0_, content0_.releaseVer as release13_1_0_, content0_.requiredTags as require14_1_0_, content0_.type as type15_1_0_, content0_.vendor as vendor16_1_0_ from cp2_content content0_ where content0_.uuid=?

select content0_.uuid as uuid1_1_0_, content0_.created as created2_1_0_, content0_.updated as updated3_1_0_, content0_.arches as arches4_1_0_, content0_.contentUrl as contentU5_1_0_, content0_.entity_version as entity_v6_1_0_, content0_.gpgUrl as gpgUrl7_1_0_, content0_.content_id as content_8_1_0_, content0_.label as label9_1_0_, content0_.locked as locked10_1_0_, content0_.metadataExpire as metadat11_1_0_, content0_.name as name12_1_0_, content0_.releaseVer as release13_1_0_, content0_.requiredTags as require14_1_0_, content0_.type as type15_1_0_, content0_.vendor as vendor16_1_0_ from cp2_content content0_ where content0_.uuid=?
```

Note the two content queries at the end -- that's invoked once for each ProductContent fetched by the getProductContentsOfActivePoolsByOwner\* method when getContent is invoked. Side note: that method name is a bit wordy -- getActiveProductContentByOwner would suffice, and even that's a tad long due to specifying the return type in the method name.

So, that means we need to solve this in the curator method.

My first thought was to use a DISTINCT with ORDER BY on the enabled field in ascending order, and then collect the content objects in order. This would put all the enabled content at the end, and would overwrite any disabled entries with an enabled one:

```
jpql = "SELECT DISTINCT pc.content, pc.enabled FROM ProductContent pc WHERE pc.product.uuid IN (:product_uuids) ORDER BY pc.enabled ASC";
```

This would work perfectly if it wasn't for one small hiccup: since we're partitioning, we have to assume we'll be running this query at least twice; and if we're running it twice, our order is no longer guaranteed between queries, and we run the risk of clobbering our "enabled" entries from the end of previous query with "disabled" entries in the next. We can work around this with some logic similar to what is outlined in the method this comment hangs on, but at that point, the ordering does literally nothing for us -- so we can omit it entirely.

```
    // OwnerContentCurator.java
    public Map<Content, Boolean> getActiveContentByOwner(String ownerId) {
        EntityManager entityManager = this.getEntityManager();
        Date now = new Date();

        // Get all the active pools

        String jpql = "SELECT p.id FROM Pool p " +
            "WHERE p.owner.id = :owner_id AND p.startDate <= :start_date AND p.endDate >= :end_date";

        List<String> poolIds = entityManager.createQuery(jpql, String.class)
            .setParameter("owner_id", ownerId)
            .setParameter("start_date", now)
            .setParameter("end_date", now)
            .getResultList();

        // Use the pool IDs to select all related product UUIDs
        // Note: we have to start partitioning now to safely handle any number of active pools

        String sql = "SELECT pool.product_uuid FROM cp_pool pool WHERE pool.id IN (:pool_ids) " +
            "UNION " +
            "SELECT pool.derived_product_uuid FROM cp_pool pool WHERE pool.id IN (:pool_ids) " +
            "UNION " +
            "SELECT pp.product_uuid FROM cp2_pool_provided_products pp WHERE pp.pool_id IN (:pool_ids) " +
            "UNION " +
            "SELECT dpp.product_uuid FROM cp2_pool_derprov_products dpp WHERE dpp.pool_id IN (:pool_ids) ";

        Query prodQuery = entityManager.createNativeQuery(sql);
        Set<String> prodIds = new HashSet<>();

        int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit() / 4);

        for (List<String> block : Iterables.partition(poolIds, blockSize)) {
            prodIds.addAll(prodQuery.setParameter("pool_ids", block).getResultList());
        }

        // Use the product UUIDs to select all related enabled content
        jpql = "SELECT DISTINCT pc.content, pc.enabled FROM ProductContent pc " +
            "WHERE pc.product.uuid IN (:product_uuids)";

        Query contentQuery = entityManager.createQuery(jpql);

        HashMap<Content, Boolean> activeContent = new HashMap<>();

        for (List<String> block : this.partition(prodIds)) {
            contentQuery.setParameter("product_uuids", block)
                .getResultList()
                .forEach(col -> {
                    Content content = (Content) ((Object[]) col)[0];
                    Boolean enabled = (Boolean) ((Object[]) col)[1];

                    activeContent.merge(content, enabled, (v1, v2) ->
                        v1 != null && v1.booleanValue() ? v1 : v2);
                });
        }

        return activeContent;
    }

    ...

    // ContentAccessManager.java
        .forEach((content, enabled) -> container.addContent(content, enabled));

```

This avoids the N+1 problem entirely, as we can verify by checking the output during the same spec test:

```
select pool0_.id as col_0_0_ from cp_pool pool0_ where pool0_.owner_id=? and pool0_.startDate<=? and pool0_.endDate>=?

SELECT pool.product_uuid FROM cp_pool pool WHERE pool.id IN (?, ?) UNION SELECT pool.derived_product_uuid FROM cp_pool pool WHERE pool.id IN (?, ?) UNION SELECT pp.product_uuid FROM cp2_pool_provided_products pp WHERE pp.pool_id IN (?, ?) UNION SELECT dpp.product_uuid FROM cp2_pool_derprov_products dpp WHERE dpp.pool_id IN (?, ?) 

select distinct productcon0_.content_uuid as col_0_0_, productcon0_.enabled as col_1_0_, content1_.uuid as uuid1_1_, content1_.created as created2_1_, content1_.updated as updated3_1_, content1_.arches as arches4_1_, content1_.contentUrl as contentU5_1_, content1_.entity_version as entity_v6_1_, content1_.gpgUrl as gpgUrl7_1_, content1_.content_id as content_8_1_, content1_.label as label9_1_, content1_.locked as locked10_1_, content1_.metadataExpire as metadat11_1_, content1_.name as name12_1_, content1_.releaseVer as release13_1_, content1_.requiredTags as require14_1_, content1_.type as type15_1_, content1_.vendor as vendor16_1_ from cp2_product_content productcon0_ inner join cp2_content content1_ on productcon0_.content_uuid=content1_.uuid where productcon0_.product_uuid in (? , ? , ?)
```

Note that this also keeps the logic for this problem in one place.

server/spec/content_access_spec.rb
```diff
@@ -578,7 +580,10 @@ def read_json_file(filename)
         @owner['key'], "cname-c2", 'test-content-c2', random_string("clabel"), "ctype", "cvendor",
         {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
     @cp.add_content_to_product(@owner['key'], product['id'], content_c2['id'], false)
-    @consumer = consumer_client(@user, @consumername, type=:system, username=nil, facts= {'system.certificate_version' => '3.3'})
+
+    create_pool_and_subscription(@owner['key'], product['id'], 10)
```
Ceiu, line 584: Another one here

server/spec/content_access_spec.rb
```diff
@@ -664,7 +666,10 @@ def read_json_file(filename)
     @cp.add_content_to_product(@owner['key'], product_1['id'], content_c3['id'], false)
     @cp.add_content_to_product(@owner['key'], product_2['id'], content_c3['id'], false)
 
-    @consumer = consumer_client(@user, @consumername, type=:system, username=nil, facts= {'system.certificate_version' => '3.3'})
+    create_pool_and_subscription(@owner['key'], product_1['id'], 10)
+    create_pool_and_subscription(@owner['key'], product_2['id'], 10)
```
Ceiu, line 670: Two more uses of this function

server/spec/content_access_spec.rb
```diff
@@ -722,4 +727,96 @@ def read_json_file(filename)
...
+        @owner['key'], "cname-c1", 'test-content-c1', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], product_1['id'], content_c1['id'], true)
+
+    content_c2 = @cp.create_content(
+        @owner['key'], "cname-c2", 'test-content-c2', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], product_2['id'], content_c2['id'], true)
+
+    create_pool_and_subscription(@owner['key'], product_2['id'], 10)
```
Ceiu, line 745: Here

server/spec/content_access_spec.rb
```diff
@@ -722,4 +727,96 @@ def read_json_file(filename)
...
+        @owner['key'], "content_c3", 'test-content-c3', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], derived_product['id'], content_c3['id'], true)
+
+    content_c4 = @cp.create_content(
+        @owner['key'], "content_c4", 'test-content-c4', random_string("clabel"), "ctype", "cvendor",
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+    @cp.add_content_to_product(@owner['key'], dev_eng_product['id'], content_c4['id'], true)
+
+    create_pool_and_subscription(@owner['key'], mkt_product1.id, 10, [eng_product.id],
```
Ceiu, line 795: and here

Ceiu (changes requested):
Getting there

server/src/main/java/org/candlepin/model/OwnerContentCurator.java
```diff
@@ -548,4 +554,60 @@ public void removeOwnerContentReferences(Owner owner, Collection<String> content
...
+            "SELECT pool.derived_product_uuid FROM cp_pool pool WHERE pool.id IN (:pool_ids) " +
+            "UNION " +
+            "SELECT pp.product_uuid FROM cp2_pool_provided_products pp WHERE pp.pool_id IN (:pool_ids) " +
+            "UNION " +
+            "SELECT dpp.product_uuid FROM cp2_pool_derprov_products dpp WHERE dpp.pool_id IN (:pool_ids) ";
+
+        Query prodQuery = entityManager.createNativeQuery(sql);
+        Set<String> prodIds = new HashSet<>();
+
+        for (List<String> block : this.partition(poolIds)) {
```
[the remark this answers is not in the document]
wolfdale: This is superb fix. I was not aware of this. 
``int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit() / 4);``

server/spec/content_access_spec.rb
```diff
@@ -36,6 +36,7 @@
 
     @product = create_product('test-product', 'some product')
     @cp.add_content_to_product(@owner['key'], @product['id'], @content_id)
+    create_pool_and_subscription(@owner['key'], @product['id'], 10)
```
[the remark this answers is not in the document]
wolfdale: Sure, Done.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
wolfdale: This look very good and efficient. This certainty avoids the N+1 problem. 
thank for help me out with this.

server/spec/content_access_spec.rb
wolfdale: Done.

server/spec/content_access_spec.rb
wolfdale: Done

server/spec/content_access_spec.rb
wolfdale: Done.

server/spec/content_access_spec.rb
wolfdale: Done.

Ceiu (approved):

PR #2800, opened 2020-09-09
ENT-2789: Port RulesResource to spec-first

wolfdale:
Changes made to port RulesResource to OpenAPI


Januson (approved):
Changes look good.

PR #2801, opened 2020-09-10
1874531: When delete the hypervisor from satellite, the manifest will be broken

Januson:


nikosmoum (commented):
The change seems good. I just wanted to ask if this covers both cases that the problem can be reproduced, as noted by a comment on the bz: `Can be reproduced not only when deleting the hypervisor host, but also when removing the attached SKU from hypervisor host`?

Also, we'll need to port this to the 2.9 branch too

Januson:
Yes. I tested with both scenarios. Even tried to craft it into a spec tests but couldn't get anything reliable.

Backport created.

PR #2692, opened 2020-04-28
ENT-1876: Port HypervisorResource to spec-first

sonalidhome:
- Added API specifications in yaml file
 - Ported HypervisorUpdateResultDTO and HypervisorConsumerDTO
 - Update the resource and test classes
 - Added exception examples for GuestIdResource

PR #2802, opened 2020-09-10
ENT-1848: Port OwnerResource to spec-first

sonalidhome:
- Added endpoints definitions
 - Added specifications for following DTOs
   1. AbstractCertificateDTO
   2. ConsumptionTypeCountsDTO
   3. ImportRecordDTO
   4. ImportUpstreamConsumerDTO
   5. OwnerInfoDTO
   6. SystemPurposeAttributesDTO
   7. UeberCertificateDTO
   8. UpstreamConsumerArrayElement
 - Created/updated translators for DTOs


server/spec/owner_resource_spec.rb
```diff
@@ -30,6 +30,18 @@
...
   end
 
+  it 'consumer cannot list other consumers for an owner' do
+    owner = create_owner random_string('test_owner')
+    username = random_string("test-user")
+    consumer_name = random_string("test-consumer")
+    user = user_client(owner, username)
+    consumer = consumer_client(user, consumer_name)
+
+    lambda do
```
wolfdale, line 40: Can we use ``exepect { }`` here to test exceptions. Something like -
```
expect {
      consumer.list_owner_consumers(owner['key'])
}.to raise_error(RestClient::Forbidden)
```

server/spec/owner_resource_spec.rb
```diff
@@ -690,6 +702,59 @@
     # and only include one instance of each unique value.
     res["systemPurposeAttributes"]["roles"].length.should == 1
   end
+
```
wolfdale, line 705: > Moved some of the access related tests from Junit to Rspec.
> Junits did not invoke the VerifyAuthorizationFilter
> when the endpoint is ported into OpenAPI.

Having some doubts about these changes of converting Junits to functional Spec tests. Since we are making changes with ``Principle`` which is necessary to port resource class to OpenAPI, not sure if we are tampering with older functionality unknowingly. 
Any thoughts ? @Ceiu @nikosmoum @Januson

api/candlepin-api-spec.yaml
```diff
@@ -2192,6 +2192,309 @@ paths:
...
+          description: Boolean value to forcefully delete an owner. This is not in use.
+          schema:
+            type: boolean
+            default: false
+      security: []
+      responses:
+        200:
+          description: Owner successfully deleted
+        409:
+          description: Occured conflict while owner deletion
```
wolfdale, line 2375: You might consider changing the description to ``` Conflict occurred while deleting an owner ```

api/candlepin-api-spec.yaml
```diff
@@ -2432,55 +2735,1142 @@ paths:
...
+          $ref: '#/components/responses/default'
+
+  /owners/{owner_key}/consumers/count:
+    get:
+      description: Retrieves a count of consumers for the owner
+      tags:
+        - Owner
+        - Consumer
+      operationId: countConsumers
+      parameters:
```
wolfdale, line 2848: Parameters descriptions are incorrect at multiple places in this endpoint.

api/candlepin-api-spec.yaml
```diff
@@ -2432,55 +2735,1142 @@ paths:
...
+          schema:
+            type: string
+        - name: product
+          in: query
+          description: The product id
+          schema:
+            type: string
+        - name: matches
+          in: query
+          description: Matches
```
wolfdale, line 2980: Please add parameter description here.

api/candlepin-api-spec.yaml
```diff
@@ -2432,55 +2735,1142 @@ paths:
...
+      parameters:
+        - name: owner_key
+          in: path
+          description: The owner key
+          required: true
+          schema:
+            type: string
+      security: []
+      responses:
+        200:
```
wolfdale, line 3533: Please change the response code to ```204``` instead of 200.

api/candlepin-api-spec.yaml
```diff
@@ -2691,347 +4315,435 @@ paths:
...
+      parameters:
+        - name: owner_key
+          in: path
+          description: The key of the owner
+          required: true
+          schema:
+            type: string
+        - name: auto_create_owner
+          in: query
+          description: Auto create owner?
```
wolfdale, line 4550: Please remove ```?``` special character from description here & add meaningful description.

wolfdale (changes requested):

server/spec/owner_resource_spec.rb
sonalidhome: Agreed. Updated the code.

api/candlepin-api-spec.yaml
sonalidhome: Updated.

api/candlepin-api-spec.yaml
sonalidhome: Added descriptions.

api/candlepin-api-spec.yaml
sonalidhome: Added description.

api/candlepin-api-spec.yaml
sonalidhome: Changed the response code to 204.

api/candlepin-api-spec.yaml
sonalidhome: Updated the description.

server/spec/owner_resource_spec.rb
Januson: I agree. Rather than moving tests to RSpec we should have proper integration tests for our filters and other RestEasy stuff.

I would propose to create a card to create said tests and instead of deleting the broken tests to disable them with a card tag.

Ceiu: I'm a bit late to respond here, but yeah... let's not move JUnit tests to rspec if we can avoid it. Last time we discussed it, the long-term plan is to get rid of rspec (which is a large part of why we're doing the switch to OpenAPI stuff to begin with), so it'll just create more work for us down the line.

nikosmoum: I agree that whenever possible, we should have unit tests instead of spec tests. Spec tests take much longer to run, so if a functionality can be tested at the unit test level, we should keep it at that.

sonalidhome: Agreed. After more digging into this, it came down to be a **Guice AOP [issue](https://github.com/google/guice/issues/201).** 
Currently, we override the methods in resource classes from the interface. So few of the annotation in the resource class method is `@Override`. And other HTTP Annotations are in API interface methods. So in `TestingModules`, we bind authorization interceptor, we use `HttpMethodMatcher`. But It does not find the superclass HTTP annotations and fails to invoke the authorization filter for each method.

To overcome this particular scenario, I have added a code to explicitly add interceptor for OwnerResource :
` bindInterceptor(Matchers.subclassesOf(OwnerResource.class),
                Matchers.any(), authMethodInterceptor);`
This will simply work for any method present in OwnerResource. And it serves the current purpose. Also, we need to look again when we port `ConsumerResource`.

server/src/main/java/org/candlepin/dto/api/v1/OwnerInfoDTOTranslator.java
```diff
@@ -0,0 +1,91 @@
...
+import org.candlepin.model.OwnerInfo;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * The OwnerInfoDTOTranslator provides translation from
+ * OwnerInfo model objects to OwnerInfoDTO
+ */
+public class OwnerInfoDTOTranslator implements ObjectTranslator<OwnerInfo, OwnerInfoDTO> {
```
Ceiu, line 29: Let's use this as an opportunity to rename OwnerInfo to *not* that. OwnerStatistics would be a better fit here

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1655,125 +1367,28 @@ public int countConsumers(
...
+        List<SubscriptionDTO> subscriptions = new LinkedList<>();
 
         for (Pool pool : this.poolManager.listPoolsByOwner(owner).list()) {
-            SourceSubscription srcsub = pool.getSourceSubscription();
 
+            SourceSubscription srcsub = pool.getSourceSubscription();
             if (srcsub != null && "master".equalsIgnoreCase(srcsub.getSubscriptionSubKey())) {
-                subscriptions.add(this.poolManager.fabricateSubscriptionFromPool(pool));
+                subscriptions.add(this.translator.translate(
+                    this.poolManager.fabricateSubscriptionFromPool(pool),
```
Ceiu, line 1381: I'd have to go back and look, but did we add a DTO for Subscription? Subscription is already a DTO.

Additionally, rather than converting twice, since we only use ```fabricateSubscriptionFromPool``` in the resources to convert a pool to a subscription, we could/should just write a translator that explicitly handles the translation from Pool to Subscription[DTO] and use that instead.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1794,30 +1409,23 @@ public AsyncJobStatusDTO refreshPools(
...
+        AsyncJobStatus job;
+
+        try {
+            job = this.jobManager.queueJob(config);
+        }
+        catch (JobException e) {
+            String errmsg = this.i18n.tr("An unexpected exception occurred while scheduling job \"{0}\"",
+                config.getJobKey());
+            log.error(errmsg, e);
+            throw new IseException(errmsg, e);
```
Ceiu, line 1421: Our mappers already handle converting exceptions to the generic "internal server error" if they're not caught. If we don't have explicit error handling to perform, or some other behavior to improve the output for clients. Additionally, the declaration and translation of the job can only happen if the queuing of the job succeeds, so there is no reason to declare the job or perform translation outside the try/catch block.

Is there some behavior change with OpenAPI that I'm not aware of? What's the intent here?

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -2002,42 +1559,28 @@ public AsyncJobStatusDTO undoImports(
         JobConfig config = UndoImportsJob.createJobConfig()
             .setOwner(owner);
 
-        AsyncJobStatus job = this.jobManager.queueJob(config);
+        AsyncJobStatus job;
```
Ceiu, line 1562: Same questions here

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -2121,7 +1637,17 @@ public AsyncJobStatusDTO importManifestAsync(
             JobConfig config =  manifestManager.importManifestAsync(owner, fileData.getData(),
                 fileData.getUploadedFilename(), overrides);
 
-            AsyncJobStatus job = this.jobManager.queueJob(config);
+            AsyncJobStatus job;
```
Ceiu, line 1640: Same questions here

Ceiu (changes requested):

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+      security: []
+      parameters:
+        - name: owner_key
+          in: path
+          description: Owner key
+          required: true
+          schema:
+            type: string
+      requestBody:
+        description: A pool to be created
```
nikosmoum, line 4107: updated instead of created

nikosmoum (changes requested):
A few things found on first pass, mostly minor. I'll take another look tomorrow

server/src/main/java/org/candlepin/resource/OwnerResource.java
nikosmoum: We added a SubscriptionDTO when porting the SubscriptionResource to OpenApi (and it looks like we also do the double conversion there). We can't help not redefining it since everything that is part of the API has to be in defined in the yaml file. I agree with creating a Pool to SubscriptionDTO translator, and getting rid of the SubscriptionTranslator and the fabricateSubscriptionFromPool method.

api/candlepin-api-spec.yaml
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+                $ref: '#/components/schemas/ExceptionMessage'
+              example:
+                displayMessage: |
+                  Invalid inputs, pools cannot be listed.
+                  Reasons could be,
+                  An owner could not be found using the provided key
+                  or the consumer specified does not belong to owner on path
+                  or activationKey could not be found with provided key
+                  or the flags add_future and only_future cannot be used at the same time
+                  or the flags add_future and only_future cannot be used with the parameter after
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+                requestUuid: c4347004-8792-41fe-a4d8-fccaa0d3898a
+        404:
+          description: |
+            Invalid inputs, pools cannot be listed.
+            Reasons could be,
+            An owner could not be found using the provided key
+            or the consumer specified does not belong to owner on path
+            or activationKey could not be found with provided key
+            or the flags add_future and only_future cannot be used at the same time
+            or the flags add_future and only_future cannot be used with the parameter after
```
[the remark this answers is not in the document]

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+            One or more pool IDs to use to filter the output; only pools with IDs matching
+            those provided will be returned; may be specified multiple times
+          schema:
+            type: array
+            items:
+              type: string
+      security: []
+      responses:
+        200:
+          description: Consumers successfully retrieved
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2524,6 +2524,309 @@ paths:
...
+            default: true
+        - name: force
+          in: query
+          description: Boolean value to forcefully delete an owner. This is not in use.
+          schema:
+            type: boolean
+            default: false
+      security: []
+      responses:
+        200:
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+        - name: owner_key
+          in: path
+          description: The owner key
+          required: true
+          schema:
+            type: string
+      security: []
+      responses:
+        204:
+          description: Owner successfully deleted
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+        - name: owner_key
+          in: path
+          description: The key of the owner
+          required: true
+          schema:
+            type: string
+        - name: level
+          in: query
+          description: The log level
+          required: true
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+          $ref: '#/components/responses/default'
+
+    post:
+      description: |
+        Imports a manifest zip file for the given organization.
+        This will bring in any products, content, and subscriptions that were
+        assigned to the distributor who generated the manifest.
+      tags:
+        - Owner Import
+      operationId: importManifest
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+        - name: owner_key
+          in: path
+          description: The key of the owner
+          required: true
+          schema:
+            type: string
+        - name: hypervisor_id
+          in: query
+          description: The list of hypervisor Ids
+          required: true
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+            type: string
+        - name: name
+          in: query
+          description: The name of the environment
+          schema:
+            type: string
+      security: []
+      responses:
+        200:
+          description: Activation keys successfully retrieved
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2764,55 +3067,1142 @@ paths:
...
+        - name: owner_key
+          in: path
+          description: The key of the owner
+          required: true
+          schema:
+            type: string
+      security: []
+      responses:
+        200:
+          description: Successfully operation
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2524,6 +2524,309 @@ paths:
...
+          in: query
+          description: Boolean value to forcefully delete an owner. This is not in use.
+          schema:
+            type: boolean
+            default: false
+      security: []
+      responses:
+        200:
+          description: Owner successfully deleted
+        409:
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2524,6 +2524,309 @@ paths:
...
+          description: Owner successfully updated
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/OwnerDTO'
+        400:
+          description: |
+            Invalid inputs, Owner cannot be updated.
+            Reasons could be, invalid key
+            or the default service level is specified
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2524,6 +2524,309 @@ paths:
...
+            or the owner content access mode and content access mode list cannot be set directly in standalone mode
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ExceptionMessage'
+              example:
+                displayMessage: |
+                  Invalid inputs, Owner cannot be updated.
+                  Reasons could be, invalid key
+                  or the default service level is specified
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2524,6 +2524,309 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /owners:
+    get:
+      description: Retrieves a list of owners
```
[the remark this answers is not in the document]
sonalidhome: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2524,6 +2524,309 @@ paths:
...
+      tags:
+        - Owner
+      operationId: listOwners
+      x-java-response:
+        type: Iterable
+        isContainer: true
+      parameters:
+        - name: key
+          in: query
+          description: The owner keys
```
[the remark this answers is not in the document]
sonalidhome: Done.

server/src/main/java/org/candlepin/resource/OwnerResource.java
sonalidhome: The auto-generated method does not throws an exception.

server/src/main/java/org/candlepin/resource/OwnerResource.java
sonalidhome: The auto-generated method does not throws an exception.

server/src/main/java/org/candlepin/resource/OwnerResource.java
sonalidhome: As already discussed in PR: https://github.com/candlepin/candlepin/pull/2692#discussion_r421507021. We need to explicitly throw an exception because the auto-generated method does not throws an exception.

server/src/main/java/org/candlepin/resource/OwnerResource.java
sonalidhome: This was updated as per the previous changes done in PR: https://github.com/candlepin/candlepin/pull/2674.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1498,86 +1248,48 @@ public void deleteLogLevel(@PathParam("owner_key") String ownerKey) {
...
-     * @httpcode 400
-     * @httpcode 404
-     * @httpcode 200
-     */
+    @Override
     @Transactional
-    @GET
-    @Produces(MediaType.APPLICATION_JSON)
-    @Path("{owner_key}/pools")
     @SuppressWarnings("checkstyle:indentation")
```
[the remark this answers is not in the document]
sonalidhome: Removed.

server/src/main/java/org/candlepin/dto/api/v1/UeberCertificateTranslator.java
```diff
@@ -58,14 +59,26 @@ public UeberCertificateDTO populate(UeberCertificate source, UeberCertificateDTO
...
-        dest = super.populate(translator, source, dest);
+        if (source == null) {
+            throw new IllegalArgumentException("source is null");
+        }
+
+        if (dest == null) {
+            throw new IllegalArgumentException("dest is null");
+        }
+
+        dest.id(source.getId())
```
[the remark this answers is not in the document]
sonalidhome: Added these fields.

server/src/main/java/org/candlepin/dto/api/v1/OwnerInfoDTOTranslator.java
nikosmoum: It looks like there's only 1 endpoint that uses OwnerInfo as output, and ideally we wouldn't need a translator (we'd just change `OwnerInfoCurator.getByOwner` to populate the OwnerInfoDTO that is defined in yaml), but the problem is that OwnerInfo is not _just_ a DTO, and has a lot of `add*()` methods that contain logic (and yaml-defined DTOs can't be easily altered to contain additional methods with extra logic).

One solution about eliminating the translator entirely would be to use the builder pattern here, but creating a `OwnerInfoBuilder` that will own all these `add*()` methods and accumulate the data, and when we're done we'd call `.build()` which would return an OwnerInfoDTO.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1498,86 +1248,46 @@ public void deleteLogLevel(@PathParam("owner_key") String ownerKey) {
...
+        List<String> poolIds) {
+
+        Principal principal = this.principalProvider.get();
+        PageRequest pageRequest = ResteasyContext.getContextData(PageRequest.class);
 
         Owner owner = findOwnerByKey(ownerKey);
 
+        // set default to current date
+        Date activeOnDate = activeOn != null ? Util.toDate(activeOn) : new Date();
+        Date afterDate = after != null ? Util.toDate(after) : null;
```
nikosmoum, line 1289: For this line, you don't need to check for null. `Util.toDate()` will just return null of the arg is null

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1213,58 +1091,29 @@ public void deleteOwner(@PathParam("owner_key") String ownerKey,
...
         Owner owner = findOwnerByKey(ownerKey);
         JobConfig config = HealEntireOrgJob.createJobConfig().setOwner(owner).setEntitleDate(new Date());
 
-        AsyncJobStatus job = this.jobManager.queueJob(config);
+        AsyncJobStatus job = null;
+        try {
+            job = this.jobManager.queueJob(config);
+        }
+        catch (JobException e) {
+            e.printStackTrace();
```
nikosmoum, line 1106: We should be doing what other endpoints do here, and use proper logging (`log.error()`) and then throwing an IseException

nikosmoum: Also, like the place that crog mentioned: we don't have to declare the job variable and do the translation outside the try block. It can just be:
```
        try {
            AsyncJobStatus job = this.jobManager.queueJob(config);
             return this.translator.translate(job, AsyncJobStatusDTO.class);
        }
        catch (JobException e) {
```
since there's no other work being done after the catch anyway

server/src/test/java/org/candlepin/dto/api/v1/OwnerInfoDTOTranslatorTest.java
```diff
@@ -0,0 +1,86 @@
...
+            assertEquals(source.getEntitlementsConsumedByFamily().size(),
+                dest.getEntitlementsConsumedByFamily().size());
+
+            //TODO: Add family check here
+            /*boolean isConsumed;
+            if(!source.getEntitlementsConsumedByFamily().isEmpty()) {
+                isConsumed = source.getEntitlementsConsumedByFamily().entrySet().stream()
+                    .allMatch(e -> e.equals(dest.getEntitlementsConsumedByFamily().get(e.getKey())));
+                assertTrue(isConsumed);
+            }*/
```
nikosmoum, line 79: If we end up removing the OwnerInfoTranslator class, then this won't matter, but if we don't, we should either uncomment/update this block or remove it

sonalidhome: Removed OwnerInfoTranslator.

server/src/main/java/org/candlepin/resource/OwnerResource.java
sonalidhome: Updated.

server/src/main/java/org/candlepin/resource/OwnerResource.java
sonalidhome: Done.

server/src/main/java/org/candlepin/resource/OwnerResource.java
sonalidhome: Updated the code to translate from Pool to SubscriptionDTO.

server/src/main/java/org/candlepin/dto/api/v1/OwnerInfoDTOTranslator.java
sonalidhome: Updated the code to build OwnerInfo from OwnerInfoBuilder.

server/src/main/java/org/candlepin/model/OwnerInfoBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        // just ignore the default family if we have nothing to put in it.
+        if (physical > 0 || virtual > 0) {
+            addToEntitlementsConsumedByFamily("none", physical, virtual);
+        }
+    }
+
+    public void addDefaultEnabledConsumerTypeCount(int activePools) {
+        for (String key : ownerInfo.getEntitlementsConsumedByType().keySet()) {
```
nikosmoum, line 98: This should be iterating over `getConsumerTypeCountByPool`

server/src/main/java/org/candlepin/model/OwnerInfoBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+
+public class OwnerInfoBuilder {
+
+    private OwnerInfo ownerInfo;
+
+    public static final String PHYSICAL = "physical";
+    public static final String GUEST = "guest";
+
+    public OwnerInfoBuilder() {
+        ownerInfo = new OwnerInfo();
```
nikosmoum, line 35: The original OwnerInfo constructor was doing some data initialization:
```
        consumerGuestCounts.put(GUEST, 0);
        consumerGuestCounts.put(PHYSICAL, 0);
```
We should also add that here, and add a unit test for it if one does not already exist

server/src/main/java/org/candlepin/dto/api/v1/OwnerInfoDTOTranslator.java
nikosmoum: ack from me

server/src/main/java/org/candlepin/model/OwnerInfoBuilder.java
sonalidhome: Corrected.

server/src/main/java/org/candlepin/model/OwnerInfoBuilder.java
sonalidhome: Consumer count is set in `setConsumerGuestCounts` from `OwnerInfoCurator`. And this method is getting called before `info.build()`. Hence the Owner with no consumer is working as excepted. Still, I have added these into initialization as per the request.

server/src/main/java/org/candlepin/dto/api/v1/PoolToSubscriptionTranslator.java
```diff
@@ -74,13 +74,13 @@ public SubscriptionDTO populate(
             throw new IllegalArgumentException("dest is null");
         }
 
-        dest.id(source.getId())
+        dest.id(source.getSubscriptionId())
             .quantity(source.getQuantity())
```
nikosmoum, line 78: Unfortunately, it is not as simple as changing the source on the translator from `Subscription` to `Pool`.
When the fabricate method creates a new Subscription object, it uses this constructor: `public Subscription populate(Pool source, ProductCurator productCurator)` which itself does A LOT of work that involves logic, and the ProductCurator. For example, we can't just assign the quantity, but calculate it first with `Subscription.setQuantityFromPool`.

We should move all that logic from the `Subscription.populate(source, productCurator)` method here because it is basically a translator method hidden within the Subscription class, and entirely remove the following methods, because they are will no longer be needed:
`Subscription.Subscription(Pool source, ProductCurator productCurator)`
`Subscription.populate(Pool source, ProductCurator productCurator)`
`Subscription.setQuantityFromPool(Pool pool)`
`CandlepinPoolManager.fabricateSubscriptionFromPool(Pool pool)`
`PoolManager.fabricateSubscriptionFromPool(Pool pool)`

sonalidhome: Updated the changes as per the `Subscription.populate` method. Also removed the not required methods.

server/src/test/java/org/candlepin/controller/PoolManagerTest.java
```diff
@@ -1703,117 +1703,6 @@ public void createPoolsForPoolBonusExist() {
...
-        assertEquals(updated, fabricated.getModified());
-        assertEquals(subscriptionId, fabricated.getId());
-    }
-
-
-    /**
-     * See BZ1292283
-     */
-    @Test
-    public void testFabricateSubWithMultiplier() {
```
nikosmoum, line 1763: One last thing to mention:

I think we should preserve these 3 unit tests (this one, along with `testFabricateSubWithZeroInstanceMultiplier` and `testFabricateSubWithMultiplierAndInstanceMultiplier`) because they test the quantity translation logic. We can port them directly into `PoolToSubscriptionTranslatorTest.java` and change the fabricateSubscriptionFromPool call to `translator.translate(pool, SubscriptionDTO)`

sonalidhome: Added these tests under `PoolToSubscriptionTranslatorTest.java`

nikosmoum (approved):
ACK from me. I'll let @Ceiu to re-review & merge

api/candlepin-api-spec.yaml
```diff
@@ -5564,6 +7371,47 @@ components:
             upstreamConsumer:
               $ref: '#/components/schemas/UpstreamConsumerDTO'
 
+    OwnerInfo:
+      description: Represents an owner information
```
Ceiu, line 7375: Minor grammatical mistake here. Needs either a noun or to drop the "an":

"Represents owner information", or "Represents an owner-information container"

Ceiu (approved):

Ceiu:
Just one last minor correction and then I'm happy.

api/candlepin-api-spec.yaml
sonalidhome: Done.

PR #2803, opened 2020-09-11
[M] 1877340: Remove stale Quartz data on upgrade; ENT-2859

nikosmoum:
- Add liquibase task to truncate all Quartz tables in order
  to avoid stale data from the old Pinsetter job framework
  causing startup errors when the new Artemis-based job
  framework is initialized after an upgrade to Candlepin 3


sonalidhome (approved):
This change meets the workaround mentioned in the bug description.

Ceiu:
We'll want to revisit this sooner than later. This fix is contingent on time of application and still leaves room for a similar class of error in the future.

PR #2807, opened 2020-09-15
[M] 1877340: JobManager now removes dead jobs from the scheduler (ENT-2859)

Ceiu:
- Jobs which cannot be loaded for any reason will now be removed
  from the scheduler when the JobManager is initialized


Januson (approved):

PR #2808, opened 2020-09-16
ENT-2792: Port ConsumerContentOverrideResource to spec-first

Januson:


api/candlepin-api-spec.yaml
```diff
@@ -509,6 +509,166 @@ paths:
...
+      responses:
+        200:
+          description: List of content overrides of the Consumer
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/ContentOverrideDTO'
+        400:
```
wolfdale, line 539: One minor change here as I think it should be ```404``` originating from ```verifyAndLookupConsumer``` method of consumer curator class.

wolfdale (changes requested):
All changes looks good to me.

api/candlepin-api-spec.yaml

wolfdale (approved):
ACK

PR #2809, opened 2020-09-17
ENT-2794: @ApiResponse annotation with message

sonalidhome:
'An unexpected exception has occurred' and status code 200
 is added to API endpoints.
 - Overridden status code to 500 as default response code.
   Initially, it is set as 0 but fails with regular expression
   in openapi schema json. It needs 3 occurrences of a number.
   Also, it cannot be set a "default" string because the ApiResponse
   class has an int type code. Hence keeping the default status code to 500.


nikosmoum (changes requested):
ack on the change, but we just need to update the commit message to have an empty line between the summary and the details / bullet points

sonalidhome:
ack, updated the commit message.

nikosmoum (approved):

PR #2810, opened 2020-09-17
[M] 1879494: Allow consumer registration with SCA mode; ENT-3062

nikosmoum:
- Add back negation operator that was accidentally removed
  which resulted in failure to register manifest consumer
  when the provided content access mode was in fact allowed by
  the owner.


sonalidhome (approved):
Changes look good. Working as expected.

PR #2811, opened 2020-09-18
[M] 1854221: Validate early if executing job in RUNNING state

nikosmoum:
- In a clustered environment, if somehow a job that is
  already in the RUNNING state would get picked up for
  execution, we would update the executor field for that record
  (before eventually failing), causing the original node that
  previously executed it to never recover that job.
- This change makes it so that we error out early when trying to
  execute a job that is in the RUNNING state, and don't end up
  updating it's record.
- In the special case that the job is in the RUNNING state, but
  was previously executed on the current node, we don't fail at
  all, but move the job to QUEUED, so that it will be executed
  as normal.

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -1539,6 +1539,21 @@ public class JobManager implements ModeChangeListener {
             log.error(errmsg);
             throw new JobInitializationException(errmsg, true);
         }
+        else if (jobState == JobState.RUNNING) {
+            if (Util.getHostname().equals(status.getExecutor())) {
+                // This goes against valid state transitions, but this node was the previous executor,
+                // so we should be fine to go on executing it again here.
+                status.setState(JobState.QUEUED);
+            }
+            else {
+                String errmsg = String.format(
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -1539,6 +1539,14 @@ private AsyncJobStatus fetchJobStatus(JobMessage message)
             log.error(errmsg);
             throw new JobInitializationException(errmsg, true);
         }
+        else if (jobState == JobState.RUNNING) {
+            String errmsg =
+                String.format("Job \"%s\" (%s) is in the %s state, which is non-terminal and invalid",
```
Ceiu, line 1544: Can we change the error message to something that gives more information on the situation? Something along the lines of:

```"Job \"%s\" (%s) is already running on host \"%s\"; ignoring execution request"```

We could also add a test here where if the host equals the current host, we just manually set the job state to QUEUED and continue execution as normal.

nikosmoum: Done

Ceiu:
ack

PR #2812, opened 2020-09-21
[2.9] 1881181: Fixed an NPE during refresh when adding new derived products

Ceiu:
- Fixed a null pointer exception (NPE) that can occur during refresh
  if an existing pool which does not define a derived product is
  updated to have a derived product


sonalidhome (approved):
ACK.

PR #2813, opened 2020-09-23
RHSMQE-2273: Adding Jenkins pipeline

Lorquas:
This will run all the existing tests (with minimal changes to them) as part of a single Jenkins pipeline vs the many individual jobs we have today.  It will require changes to candlepin-jobs which can be worked on after https://github.com/candlepin/candlepin-jobs/pull/61 is merged.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -1,0 +1,64 @@
+pipeline {
+    agent { label 'docker' }
+    stages {
+        stage('Test') {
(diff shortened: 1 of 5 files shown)


Lorquas:
I've opened this PR extremely early in the dev process because I kind of need it to test itself.  No review needed until the in progress tag is removed.

nikosmoum (approved):
ACK

PR #2814, opened 2020-09-25
1841591: Manifest delete takes 24 hours or more in large environment

Januson:


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -2114,9 +2114,13 @@ public void deletePool(Pool pool) {
...
         // Must do a full revoke for all entitlements:
-        for (Entitlement e : poolCurator.entitlementsIn(pool)) {
+        List<Entitlement> entitlements = poolCurator.entitlementsIn(pool);
+        for (Entitlement e : entitlements) {
             e.setDeletedFromPool(true);
-            revokeEntitlement(e);
+        }
+
+        for (List<Entitlement> partition : Iterables.partition(entitlements, 1000)) {
+            revokeEntitlements(partition);
```
nikosmoum, line 2123: As per our conversation, there's more things to discuss here. The ideal solution is to not fetch the entitlement objects into memory at all (they're 60k+ after all), but instead do a batch delete by entitlement IDs like is done elsewhere.
To do that though, we first have to make sure we do anything else that needs to be done (revokeEntitlements does a lot of stuff, and we need to find out if we even need to do most of those in this specific use case (UndoImportsJob)).

server/src/main/java/org/candlepin/async/tasks/UndoImportsJob.java
```diff
@@ -164,11 +167,12 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
...
-        for (Pool pool : pools) {
-            if (this.poolManager.isManaged(pool)) {
-                this.poolManager.deletePool(pool);
-            }
-        }
+        this.poolManager.deletePools(
+            pools.stream()
+                .filter(this.poolManager::isManaged)
+                .collect(Collectors.toList()),
+            new ArrayList<>()
```
nikosmoum, line 174: No need to pass another argument here, the deletePools method has an overloaded version which only accepts a single list of pools. Other than that, this looks good

Januson: Done.

Ceiu:
*This* was responsible for 8 hours of unnecessary churn? Crazy how code that was once absolutely fine becomes a bottleneck over the years.

Good find; ack from me.

PR #2816, opened 2020-10-06
ENT-2788: Port AdminResource to spec-first

sonalidhome:
- Added specifications for QueueStatus DTO and admin resource APIs


Januson (approved):

PR #2817, opened 2020-10-06
ENT-2790: Port CrlResource to spec-first

Januson:


api/candlepin-api-spec.yaml
```diff
@@ -1215,6 +1215,64 @@ paths:
...
+              example:
+                displayMessage: CRL file path not defined in config file
+                requestUuid: c4347004-8792-41fe-a4d8-fccaa0d3898a
+        default:
+          $ref: '#/components/responses/default'
+
+    delete:
+      tags:
+        - Certificate Revocation List
+      description: Deletes a Certificate from the Revocation List
```
sonalidhome, line 1250: The same applies here. Let's avoid capitalization of the display message.

api/candlepin-api-spec.yaml
```diff
@@ -1215,6 +1215,64 @@ paths:
...
+        - Certificate Revocation List
+      description: Retrieves the Certificate Revocation List
+      operationId: getCurrentCrl
+      security: []
+      x-java-response:
+        type: javax.ws.rs.core.Response
+        isContainer: false
+      responses:
+        200:
+          description: Certificate Revocation List successfully retrieved
```
sonalidhome, line 1230: Let's avoid capitalization of the display message.

server/src/main/java/org/candlepin/resource/CrlResource.java
```diff
@@ -26,55 +25,44 @@
...
-import javax.ws.rs.QueryParam;
-import javax.ws.rs.core.Context;
-import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
 /**
  * CrlResource
  */
 @Path("/crl")
 @Api(value = "crl", authorizations = { @Authorization("basic") })
```
sonalidhome, line 46: `@Path` and `@Api` annotations can be removed.

sonalidhome (changes requested):

server/src/main/java/org/candlepin/resource/CrlResource.java
Januson: Done.

api/candlepin-api-spec.yaml

api/candlepin-api-spec.yaml

sonalidhome (approved):

PR #2818, opened 2020-10-06
ENT-1675: Upgrade candlepin to use Java 11

wolfdale:
Changes 
- Dependencies upgraded to support Java 11

How to test/verify this PR locally ?
1. For Local Gradle build 
- Install Java-11
```java-11-openjdk-devel```
- Make it a default
```update-alternatives --config java```
 - Run gradle build
  
2. For Vagrant (Dependent on https://github.com/candlepin/ansible-role-candlepin/pull/14)
Update [requirement.yml](https://github.com/candlepin/candlepin/blob/master/vagrant/requirements.yml) file to checkout the above PR's branch.

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -157,6 +157,16 @@ cleanup() {
     fi
 }
 
+# This function converts candlepin version to a int value with (Zero) padding
+# which helps in comparing versions.
+# Examples -
+# 2.5.6  -> 2005006000
+# 3.2.0  -> 3002000000
+# 2.9.6  -> 2009006000
+get_simplified_version() {
+	echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }';
+}
+
 usage() {
     cat <<HELP
 usage: cp-test [options]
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: docker/candlepin-base/cp-test [hunk 1]
```diff
@@ -246,6 +254,14 @@ fi
 # Set our project to test
 PROJECT=${PROJECT:-server}
 
+# Update tomcat config to use Java 11 &
+# explicitly set Java 11 as default
+if [ "$JAVA_VERSION" == "11" ]; then
+    echo "Overriding tomcat.conf to use Java 11"
+    update-alternatives --set java /usr/lib/jvm/java-$JAVA_VERSION-openjdk-$JAVA_VERSION*/bin/java
+    sed -i 's/JAVA_HOME=.*/JAVA_HOME="\/usr\/lib\/jvm\/jre-11"/' /etc/tomcat/tomcat.conf
```
nikosmoum, line 262: Why are we doing this twice? This is already done above

nikosmoum (changes requested):
Looks good. Only some duplicate code

docker/candlepin-base/cp-test
wolfdale: Logic changed. Not need anymore.

wolfdale:
We are not going with the approach to pass the "-j 11" flag in Jenkin build, instead we will detect which version of java to use by identifying Candlepin version. We will use Java11 for candlepin 3.2.0 and onwards. 
To do this we have 2 options.

Option 1) Create 2 PR's, First PR for Docker changes to setup Java11 environment inside images. Second PR for Java11 related dependencies. In this case PR, 1st PR needs to be merged, to verify the second PR.

Option 2) Push the docker related changes (Java11 related changes) from the local system to the Candlepin docker registry, (In this case if something wrong happens, we can just trigger the old docker build to revert back to the previous state of docker images). This will allow us to verify the current PR (all tests should be green) and then we can merge it (along with docker changes).  

We are more inclined to follow Option-2.

nikosmoum (approved):
ACK. Only question I have is, has the rpm building, installation and deployment been tested and works?

wolfdale:
> Only question I have is, has the rpm building, installation and deployment been tested and works?

@nikosmoum Do we need to consider `candlepin-rhel7` docker images to be compliant with Java11 as well. Right now in this PR, I have not made any changes to `candlepin-rhel7` and `gating-test-images`. I thought we are going to leave them as it is. 

I did verification with `gating-test-images` (by making Java11 changes) to check if rpm is correctly build, installed and deployed, since I realized that `<candlepin>/docker/gating-test-images/cp-latest-stage/startup.sh` uses `cpsetup`. Not sure how valid is this. Any more ideas ?

nikosmoum:
@wolfdale there's no need for more container changes for now. The rpm test within gating-test-images is not exactly valid, since building & running in a centos7 container is not the same as rhel7, but we should be fine

PR #2829, opened 2020-10-22
ENT-2594: Mechanism to determine if SCA can be enabled for a consumer

Januson:
- Add a new endpoint GET /consumers/{consumer_uuid}/content_access
 - Add a new endpoint GET /owners/{owner_key}/content_access

### DIFF
--- a/server/src/main/java/org/candlepin/controller/OwnerContentAccess.java
+++ b/server/src/main/java/org/candlepin/controller/OwnerContentAccess.java
@@ -1,0 +1,63 @@
+/**
+ * Copyright (c) 2009 - 2020 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.controller;
+
+import java.util.Objects;
+
+public class OwnerContentAccess {
+
+    private final String contentAccessMode;
+    private final String contentAccessModeList;
+
+    public OwnerContentAccess(String contentAccessMode, String contentAccessModeList) {
+        this.contentAccessMode = contentAccessMode;
+        this.contentAccessModeList = contentAccessModeList;
+    }
+
+    public String getContentAccessMode() {
+        return contentAccessMode;
+    }
+
+    public String getContentAccessModeList() {
+        return contentAccessModeList;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        OwnerContentAccess that = (OwnerContentAccess) o;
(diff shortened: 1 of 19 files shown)


jturel:
hey @Januson is there any way you can get us a scratch build to test this change against in Katello? Our plan is to code to the API, and handle the HTTP 404 case appropriately. The idea being that once this reaches Hosted it will start working in the way we intend to use it.

server/src/main/java/org/candlepin/controller/ContentAccessService.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import javax.inject.Singleton;
+
+/**
+ * Service responsible for handling queries for content access data.
+ */
+@Singleton
+public class ContentAccessService {
```
Ceiu, line 38: I feel like the logic here could live in the ContentAccessManager instead of adding another class for handling business logic.

Ceiu (changes requested):
I feel like the logic here could live in the ContentAccessManager instead of adding another class for handling business logic. And, oddly, there's not really business logic here -- it's just a matter of fetching the consumer or owner object, and then building a DTO to return following the rules outlined in the design doc.

Following the current Candlepin design, this type of thing should live entirely in the resource layer, and then when we have a service layer for the owner and consumer endpoints, that moves there (though we still have some details to work out like who handles translation to and from DTOs, etc.)

A lot of this can be boiled down to two very tiny resource methods:

For owner content access:
```
    public ContentAccessDTO getContentAccess(@PathParam("owner_key") @Verify(Owner.class) String key) {
        Owner owner = this.findOwnerByKey(key);

        String caMode = Util.firstOf(
            owner.getContentAccessMode(),
            ContentAccessMode.getDefault().toDatabaseValue());

        String caList = Util.firstOf(
            owner.getContentAccessModeList(),
            ContentAccessMode.getDefault().toDatabaseValue());

        return new ContentAccessDTO()
            .setContentAccessMode(caMode)
            .setContentAccessModeList(caList);
    }
```

and for consumers:

```
    public ContentAccessDTO getContentAccess(@PathParam("consumer_uuid") @Verify(Consumer.class) String uuid) {
        Consumer consumer = consumerCurator.verifyAndLookupConsumer(uuid);

        String caMode = Util.firstOf(
            consumer.getContentAccessMode(),
            consumer.getOwner().getContentAccessMode(),
            ContentAccessMode.getDefault().toDatabaseValue());

        String caList = Util.firstOf(
            consumer.getOwner().getContentAccessModeList(),
            ContentAccessMode.getDefault().toDatabaseValue());

        return new ContentAccessDTO()
            .setContentAccessMode(caMode)
            .setContentAccessModeList(caList);
    }
```

Where ```Util.firstOf``` is defined as:

```
    public static String firstOf(String... values) {
        for (T value : values) {
            if (value != null && !value.isEmpty()) {
                return value;
            }
        }

        return null;
    }
```

These could be further shrunk, optimized, and made slightly more generic around what we need with a bit of work. But mostly they use pre-existing validation steps (even if the Consumer resolver/validator is in the wrong spot itself), and minimizes the overall work needed. 

This is almost purely API-level logic, so I'm not sure we need to get fancier than necessary here.

Ceiu:
The only remaining issue I have is a non-blocking issue that we can correct later if it comes up. acking.

server/src/main/java/org/candlepin/controller/OwnerContentAccess.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(contentAccessMode, contentAccessModeList);
+    }
+
+    @Override
+    public String toString() {
+        return "OwnerContentAccess{" +
```
Ceiu, line 58: For these, we should follow the format we use in the other objects: ClassName [prop=value, prop2=value2, ...]

Ceiu (approved):

PR #2831, opened 2020-11-03
[3.1] 1893836: Fixed encoding of environment name in SCA certificates (ENT-3211)

Ceiu:
- Content paths in certificates for simple content access (SCA)
  no longer encode the forward slash of any containing environment
  name

### DIFF
--- a/server/src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
+++ b/server/src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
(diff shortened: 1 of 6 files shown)


server/src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20201103112757-1" author="crog">
```
nikosmoum, line 9: Should we not add `dbms="postgresql"` here, to only run this on satellites? (since it is not needed for hosted)?

PR #2833, opened 2020-11-04
[2.9] 1894455: Manifest delete takes 24 hours or more in large environment (60k consumers in owner)

Januson:


sonalidhome (approved):
ACK

PR #2838, opened 2020-11-10
1884410: Fix for misleading exception message in auto-attach (ENT-3128)

wolfdale:
Changes/Fix 
- There was ambiguity in handling exception for conditions, where auto-attach is disabled for org & when org in SCA mode. This was leading to misleading exception messages.
- Added functional tests.


sonalidhome (approved):
Changes look good to me.

PR #2840, opened 2020-11-16
[M] Update rpm spec and cpsetup/cpdb for Java 11; ENT-3255

nikosmoum:
- Update spec dependency to 'java-11', since 'java' is
  used for the default in RHEL which is 8
- Set JAVA_HOME to 11 for liquibase migration call
- Set JAVA_HOME to 11 in tomcat.conf

### DIFF
--- a/server/code/setup/cpsetup
+++ b/server/code/setup/cpsetup
@@ -92,6 +92,10 @@ class TomcatSetup(object):
                     TLS_ECDH_anon_WITH_AES_128_CBC_SHA,
                     TLS_ECDH_anon_WITH_AES_256_CBC_SHA"
                truststorePass="%s" />""" % (keystorepwd, keystorepwd)
+        self.main_conf = '/etc/tomcat/tomcat.conf'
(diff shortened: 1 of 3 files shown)


server/code/setup/cpsetup
```diff
@@ -118,6 +131,21 @@ class TomcatSetup(object):
...
+        original_main = original_main_file.read()
+        original_main_file.close()
+
+        if re.search(self.main_existing_pattern, original_main, re.MULTILINE):
+            updated_main = self._replace_current_main(original_main)
+        elif re.search(self.main_comment_pattern, original_main, re.MULTILINE):
+            updated_main = self._replace_commented_main(original_main)
+        else:
+            updated_main = original_main + os.linesep + self.new_java_home
+        main_config = open(self.main_conf, 'w')
```
Januson, line 145: You could use Python's with keyword to close the file automatically.
```
with open(self.main_conf, 'w') as main_config:
    main_config.write(updated_main)
```

nikosmoum: updated

Januson (approved):

PR #2844, opened 2020-11-19
using vendored docker images

Lorquas:
Using copies of the database from our own repos


nikosmoum (approved):
ACK

The pipeline only fails when re-run, due to permission issues. That's going to get resolved by Kevin's PR.

PR #2845, opened 2020-11-19
[M] Added support for cloud registration (ENT-3213) (ENT-3214)

Ceiu:
- Added support for automated cloud registration via JWT tokens
- Added the CloudRegistrationAdapter and CloudRegistrationInfo
  adapter and interface for allowing service adapters to implement
  the organization resolution logic for cloud registration details
- Added new JWT configurations for controlling the issuer name and
  TTL
- Updated the mutators in Owner to return itself for fluent-style
  method chaining
- Updated KeycloakAuth to no longer end authentication on
  unsupported auth types, or failed authentication in general

### DIFF
--- a/common/src/main/java/org/candlepin/common/exceptions/NotImplementedException.java
+++ b/common/src/main/java/org/candlepin/common/exceptions/NotImplementedException.java
@@ -1,0 +1,32 @@
+/**
+ * Copyright (c) 2009 - 2020 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.common.exceptions;
+
+import javax.ws.rs.core.Response.Status;
+
+
+/**
+ * NotImplementedException
+ * Represents a Not Implemented (HTTP 501) error.
+ */
+public class NotImplementedException extends CandlepinException {
+    public NotImplementedException(String message) {
+        super(Status.NOT_IMPLEMENTED, message);
+    }
+
+    public NotImplementedException(String message, Throwable cause) {
+        super(Status.NOT_IMPLEMENTED, message, cause);
+    }
+}
--- a/server/src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
+++ b/server/src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
@@ -1,0 +1,316 @@
+/**
+ * Copyright (c) 2009 - 2020 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.auth;
+
+import org.candlepin.auth.permissions.OwnerPermission;
+import org.candlepin.auth.permissions.Permission;
+import org.candlepin.common.config.Configuration;
+import org.candlepin.common.config.ConversionException;
+import org.candlepin.common.resteasy.auth.AuthUtil;
+import org.candlepin.config.ConfigProperties;
+import org.candlepin.model.Owner;
+import org.candlepin.model.OwnerCurator;
+import org.candlepin.pki.CertificateReader;
+import org.candlepin.service.CloudRegistrationAdapter;
+import org.candlepin.service.exception.CloudRegistrationAuthorizationException;
+import org.candlepin.service.exception.MalformedCloudRegistrationException;
+import org.candlepin.service.model.CloudRegistrationInfo;
+import org.candlepin.util.Util;
+
+import org.jboss.resteasy.spi.HttpRequest;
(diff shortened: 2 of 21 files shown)


Ceiu:
The only oddity here that probably deserves some discussion, is the assignment of the "type" field on the JWT object itself. This *shouldn't* matter, and we can pull it if we want, it just allows us to have control over some unnecessary processing or potential collisions between keycloak and cloud auth, as both use the same transport and try to process each other's tokens without some additional check.

common/src/main/java/org/candlepin/common/exceptions/NotImplementedException.java
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) 2009 - 2012 Red Hat, Inc.
```
wolfdale, line 2: Copyright header year.

server/src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -0,0 +1,316 @@
+/**
+ * Copyright (c) 2009 - 2012 Red Hat, Inc.
```
wolfdale, line 2: Incorrect copyright year.

server/src/main/java/org/candlepin/dto/api/v1/CloudRegistrationDTO.java
```diff
@@ -0,0 +1,180 @@
+/**
+ * Copyright (c) 2009 - 2017 Red Hat, Inc.
```
wolfdale, line 2: Incorrect copyright year.

server/src/main/java/org/candlepin/service/model/CloudRegistrationInfo.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Fetches the metadata for the cloud provider, such as the user's account identifiers.
+     *
+     * @return
+     *  the cloud provider metadata, or null of the metadata has not been set
+     */
+    String getMetadata();
+
+    /**
+     * Fetches the signature to use for verifying the identy of the cloud provider.
```
wolfdale, line 46: Typo. It should be ``identity``

wolfdale (changes requested):
Overall changes looks good to me.

common/src/main/java/org/candlepin/common/exceptions/NotImplementedException.java
Ceiu: Fixed

server/src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
Ceiu: Fixed

server/src/main/java/org/candlepin/dto/api/v1/CloudRegistrationDTO.java
Ceiu: Fixed

server/src/main/java/org/candlepin/service/model/CloudRegistrationInfo.java
Ceiu: Fixed

nikosmoum (changes requested):
A couple of things I noticed; will take another look next week.

server/src/main/java/org/candlepin/service/CloudRegistrationAdapter.java
```diff
@@ -0,0 +1,46 @@
...
+package org.candlepin.service;
+
+import org.candlepin.service.exception.CloudRegistrationAuthorizationException;
+import org.candlepin.service.exception.MalformedCloudRegistrationException;
+import org.candlepin.service.model.CloudRegistrationInfo;
+
+
+
+/**
+ * Interface to the Certificate Service.
```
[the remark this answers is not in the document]
Ceiu: Fixed

server/src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
+
+
+
+/**
+ * Endpoints for cloud-features features and authentication
```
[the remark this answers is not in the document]
Ceiu: Fixed

server/src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -0,0 +1,316 @@
...
+     *
+     * @return
+     *  an encrypted JWT token string
+     */
+    private String buildRegistrationToken(Principal principal, String ownerKey) {
+        String keyId = KeyUtils.createKeyId(this.publicKey);
+
+        // Try to use the username present in the principal; otherwise use the default
+        String username = principal.getUsername();
+        if (username == null) {
```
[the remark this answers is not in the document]
Ceiu: Not always; it depends on the headers provided in the request. More on this below.

server/src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -0,0 +1,80 @@
...
+        this.cloudRegistrationAuth = Objects.requireNonNull(cloudRegistrationAuth);
+    }
+
+    @POST
+    @Path("authorize")
+    @Produces(MediaType.TEXT_PLAIN)
+    @Consumes(MediaType.APPLICATION_JSON)
+    @SecurityHole(noAuth = true)
+    public String authorize(CloudRegistrationDTO cloudRegDTO,
+        @Context Principal principal) {
```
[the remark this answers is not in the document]
Ceiu: The intent here is to allow a principal, if present, to be (partially) carried forward. In general, at least today, I agree that this will virtually always going to resolve to the default.

This is one of those things where I feel we should be doing *something* to track who is doing what. Especially since in the current design, the token usage *is not* limited to one use, since this is all stateless. We could use the correlation ID and make sure that matches during the auth step.

Given that we require a rather destructive permission, we should probably do something.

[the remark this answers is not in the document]
nikosmoum: I suppose we can keep the principal here for now, although it's not directly linked to using the correlation ID (I would imagine that would be set on the token itself, not having anything to do with the principal)

nikosmoum (approved):
ACK for now, but not merging, since as you said maybe it's a good idea to use the correlation ID, and probably better to do it sooner than later

Ceiu:
@nikosmoum 
After further discussion, we're not going to implement correlation ID tracking yet. Fortunately the pieces are in place to do any further scoping without requiring client changes, so we can change our mind later.

PR #2846, opened 2020-11-20
Figure out candlepin version through spec file

nikosmoum:
- Don't try to find out candlepin version through git tag
  because we might be in a detached HEAD


nikosmoum:
This needs to be rebased once https://github.com/candlepin/candlepin/pull/2843 gets merged, for the jenkins pipeline to pass.

Januson (approved):

PR #2849, opened 2020-11-26
ENT-3109: Create Jenkins job for translation extraction

sonalidhome:
(gettext/msgattrib)
 - Added a parameter "-g" in cp-test to get text to
   extract strings for translation from source java files

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -49,6 +49,29 @@ validate_translation() {
     fi
 }
 
+gettext() {
+    if [ -f ./gradlew ]; then
+        ./gradlew gettext
+    fi
+
+    #Code to commit the updated files
+    CHANGED_FILE=`git diff --name-only HEAD`
+    ADD_COUNT=0
+
+    for file in $(echo $CHANGED_FILE | tr "\n" "\n")
+    do
+      echo "Performing: git add $file."
+      git add $file
+      ((ADD_COUNT++))
+    done
+
+    if [ $ADD_COUNT -gt 0 ]; then
+      echo "Committing and pushing the above mentioned files."
+      git commit -m "keys.pot file is updated"
+      git push
+    fi
+}
+
 build_unittest() {
     UTCMD=$1
     if [ -f ../gradlew ]; then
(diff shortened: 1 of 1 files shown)


sonalidhome:
This PR needs to be reviewed and merged before PR:https://github.com/candlepin/candlepin/pull/2848. Because cp-test changes with the newly added parameter "-g" should reflect to execute the gettext job task.

docker/candlepin-base/cp-test
```diff
@@ -49,6 +49,29 @@ validate_translation() {
...
+        ./gradlew gettext
+    fi
+
+    #Code to commit the updated files
+    CHANGED_FILE=`git diff --name-only HEAD`
+    ADD_COUNT=0
+
+    for file in $(echo $CHANGED_FILE | tr "\n" "\n")
+    do
+      echo "Performing: git add $file."
```
nikosmoum, line 63: There's no need to loop and add each individual file here a `git add common/po` would suffice, since gettext only changes the keys.pot file.

docker/candlepin-base/cp-test
```diff
@@ -49,6 +49,29 @@ validate_translation() {
...
+      git add $file
+      ((ADD_COUNT++))
+    done
+
+    if [ $ADD_COUNT -gt 0 ]; then
+      echo "Committing and pushing the above mentioned files."
+      git commit -m "keys.pot file is updated"
```
nikosmoum, line 70: In addition to the gettext changes, we should also be running msgattrib right after it in this job (which will be removing strings from the .po files, based on them being removed from the .pot file). That's the 'Replace the second zanata pull with' step from the design document.

For how that is done, we have a couple of options. In shell script it could be done as such:
```
for pofile in common/po/*.po; do
    msgattrib --set-obsolete --ignore-file=common/po/keys.pot -o $pofile $pofile;
    msgattrib --no-obsolete -o $pofile $pofile;
done;
```
But maybe a more consistent way would be to do add it as another gradle task in the Gettext.groovy file we have (similar to the existing gradle `gettext` and `msgmerge` commands we have).

docker/candlepin-base/cp-test
```diff
@@ -195,11 +218,12 @@ OPTIONS:
   -j <version>     use a specific Java version instead of the auto-detected default
   -v               enable verbose/debug output
   -i               internationalization - validate translation files
+  -g               internationalization - gettext to extract strings for translation from source java files
```
nikosmoum, line 221: Also, similarly to the other PR comment, we probably don't have to add this to the cp-test script, since it doesn't have to run in a docker container.

nikosmoum (changes requested):
A couple of comments around the missing msgattrib work, and also how we want this to work overall (probably not in a docker container)

docker/candlepin-base/cp-test
sonalidhome: Agreed, raised a PR https://github.com/candlepin/candlepin/pull/2850 to add new gradlew task `msgattrib`.

sonalidhome:
Closing this PR because these changes are not required.

PR #2850, opened 2020-12-01
ENT-3109: Create Jenkins job for translation extraction

sonalidhome:
(gettext/msgattrib)
 - Added a Gradle task to perform msgattrib on po files.

### DIFF
--- a/buildSrc/src/main/groovy/Gettext.groovy
+++ b/buildSrc/src/main/groovy/Gettext.groovy
@@ -108,5 +108,27 @@ class Gettext implements Plugin<Project> {
                 }
             }
         }
+
+        def msgattrib_task = project.task('msgattrib') {
+            description = 'Use msgattrib to remove obsolete strings (that were already removed from the source code & template file) from translation files.'
+            group = 'build'
+            doLast {
+                def po_files = new FileNameFinder().getFileNames("${extension.keys_project_dir}/po/", '*.po')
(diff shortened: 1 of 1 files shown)


sonalidhome:
Note: Once this gets approved and merged, we need to cherry-pick this into candlepin-3.1-HOTFIX and candlepin-2.9-HOTFIX branches.

buildSrc/src/main/groovy/Gettext.groovy
```diff
@@ -108,5 +108,27 @@ class Gettext implements Plugin<Project> {
                 }
             }
         }
+
+        def msgattrib_task = project.task('msgattrib') {
+            description = 'The msgattrib program filters the messages of a translation catalog according to their attributes, and manipulates the attributes.'
```
nikosmoum, line 113: Even though this is the description of what msgattrib is doing in general, we're using msgattrib in a very specific way here, so let's change this description to reflect that. Something like `Use msgattrib to remove obsolete strings (that were already removed from the source code & template file) from translation files. ` should be good

nikosmoum (approved):
ACK (only 1 minor change needed)

buildSrc/src/main/groovy/Gettext.groovy
sonalidhome: Updated.

PR #2852, opened 2020-12-01
1884047: Unable to unschedule trigger while deleting job crash during startup (ENT-3129)

wolfdale:
Since original bug was harder to reproduce locally, judging form the exception trace the crash happens around ```deleteJob()``` area.
This PR has minor update to efficiently unschedule (& delete) quartz jobs using Trigger-keys. Internally ```unscheduleJobs()``` is a bulk operation which faster & efficient. We can leverage this as all of our Quartz job are non-durable which means job gets automatically deleted from the scheduler once there are no longer any active triggers associated with it.

### DIFF
--- a/server/src/main/java/org/candlepin/async/JobManager.java
+++ b/server/src/main/java/org/candlepin/async/JobManager.java
@@ -912,7 +912,18 @@ public class JobManager implements ModeChangeListener {
 
         // Unschedule dead/invalid jobs
         for (JobKey key : unschedule) {
-            this.scheduler.deleteJob(key);
+            List<Trigger> jobTriggers = (List<Trigger>) this.scheduler.getTriggersOfJob(key);
+
+            if (!jobTriggers.isEmpty()) {
+                this.scheduler.unscheduleJobs(jobTriggers
+                    .stream()
+                    .map(Trigger::getKey)
+                    .collect(Collectors.toList()));
+            }
+            else {
+                this.scheduler.deleteJob(key);
+            }
+
             log.info("Removed existing schedule for job: {}", key.getName());
         }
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -912,8 +912,16 @@ private void synchronizeJobSchedule() throws SchedulerException {
 
         // Unschedule dead/invalid jobs
         for (JobKey key : unschedule) {
-            this.scheduler.deleteJob(key);
-            log.info("Removed existing schedule for job: {}", key.getName());
+            List<Trigger> jobTrigger = (List<Trigger>) this.scheduler.getTriggersOfJob(key);
```
Ceiu, line 915: jobTrigger should be plural, since it's a collection.

server/src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -912,8 +912,16 @@ private void synchronizeJobSchedule() throws SchedulerException {
...
+
+            if (!jobTrigger.isEmpty()) {
+                this.scheduler.unscheduleJobs(jobTrigger
+                    .stream()
+                    .map(Trigger::getKey)
+                    .collect(Collectors.toList()));
+                log.info("Removed existing schedule for job: {}", key.getName());
+            }
```
Ceiu, line 923: We should have a contingency here in the event jobTriggers is empty. In such a case, we'll have a job key associated with nothing, and that should get cleaned up appropriately.

Ceiu (changes requested):
Looks good, just need to handle one more case to ensure everything is cleaned up no matter how bad the data is.

server/src/main/java/org/candlepin/async/JobManager.java
wolfdale: Fixed.

server/src/main/java/org/candlepin/async/JobManager.java
wolfdale: Added case to delete the job for proper cleanup (assuming we might have stale data). 
Although since all Candlepin quartz jobs are non-durable, it is automatically deleted from the scheduler once there are no longer any active triggers associated with it.

Ceiu (approved):

PR #2854, opened 2020-12-04
Update build.gradle

jdhrh:
Add controller package to the apijar included sources which makes our internal build easier.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -129,6 +129,7 @@ task apiJar(type: Jar) {
     includes = [
         'auth',
         'config',
+        'controller',
         'jackson',
         'model',
         'pki',


Ceiu:
Some details as to why this is needed today:
Current adapter code makes use of the Owner and OwnerCurator, which themselves reference the ContentAccessManager and OwnerContentAccess from the controller package, respectively.

As a short-term solution, this may be the cleanest way to unbreak the adapter build process. Other options include moving the referenced classes out of controller, or scoping the inclusion to only the classes needed. These may cause more headaches down the line, though.

@nikosmoum @kahowell thoughts on this?

nikosmoum (approved):
ACK as a short term solution, but for the long-term, it would be good to have a list of classes/methods the adapters use currently, along with how/why, so we can instead create a clear API that includes everything they need to use and only export that, instead of the 'include all the things' strategy

PR #2855, opened 2020-12-07
[M] Fixed hash collision resolution bug with branding (ENT-3304)

Ceiu:
- Branding.equals no longer checks the product UUID reference
  as part of the equality check
- Fixed a bug in the product and content version lookups that
  would cause JDBC failures if too many versions were matched
  or requested


Ceiu:
retest this please

nikosmoum (approved):
ACK

PR #2856, opened 2020-12-07
[M] 1905199: Fixed bug with state filtering validation during job cleanup (ENT-3330)

Ceiu:
- Fixed a bug that would cause job cleanup to match all jobs if
  only non-terminal states were provided to the job cleanup
  operation


sonalidhome (approved):
ACK.

PR #2859, opened 2020-12-10
[M] Set the default consumer window size to zero in broker.xml (ENT-3066)

Ceiu:
- Set the default-consumer-window-size setting to zero for the job
  address in broker.xml to prevent a single consumer from grabbing
  multiple job messages while other threads are waiting for work


sonalidhome (approved):
ack.

PR #2860, opened 2020-12-10
[M] Added an option to cp-test for generating test repos on deploy (ENT-3088)

Ceiu:
- Added the -R option to cp-test to trigger generation of test
  repos during Candlepin deployment
- Added the -a option to cp-test to pass through any custom
  argument to the Candlepin deploy script that is not otherwise
  supported by cp-test
- Fixed an issue with build-images that would prevent it from
  working properly on python 3
- Removed the -p option from cp-test for selecting which
  subproject to build, as Candlepin no longer has subprojects

### DIFF
--- a/docker/build-images
+++ b/docker/build-images
@@ -62,7 +62,7 @@ if [ "$1" != "" ]; then
 fi
 
 get_image_names() {
-  python -c 'import yaml,sys;y=yaml.safe_load(sys.stdin); print " ".join(y["services"].keys())' < $SCRIPT_HOME/docker-compose-build.yml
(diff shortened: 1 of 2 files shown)


sonalidhome (approved):

nikosmoum:
Unfortunately this doesn't address the needs for what the DNF team is looking for. This makes changes to the candlepin-base container which we use for our Jenkins tests, and the way we use it, will run the deploy script, then run any tests we want and exit.
A more appropriate change to achieve what's needed would be in the gating-test-images; if we could generate the test repos dump them to a tmp directory in the same way we dump the sql, and then have that same directory mounted to  /var/lib/tomcat/webapps/ROOT/ when we run the main pod.

PR #2861, opened 2020-12-11
ENT-3317: Drop the cp_job table

sonalidhome:
- Added a xml file to drop the table
 - Removed the `batch-bind-performance.rb` script because it
   is no longer useful in latest version. Table `cp_job` used
   to check consumer_compliance. Also, the functionality
   `consume_pools` does not exists.


nikosmoum (approved):
ACK

PR #2867, opened 2020-12-16
ENT-3286: Manifest delete takes 24 hours or more in large environment (60k consumers in owner)

Januson:
- Fix too many params issue in manifest import
 - Replace batch delete of entitlements with batch delete by ids
 - Refactor updatePoolsFromStack to avoid calling deletePools n times
 - Some cleanup as I went through the code (replace finally with try with resources, remove redundant type params, etc..)

### DIFF
--- a/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -40,7 +40,6 @@ import org.candlepin.model.Entitlement;
 import org.candlepin.model.EntitlementCertificateCurator;
 import org.candlepin.model.EntitlementCurator;
 import org.candlepin.model.Owner;
-import org.candlepin.model.OwnerContentCurator;
 import org.candlepin.model.OwnerCurator;
 import org.candlepin.model.OwnerProductCurator;
 import org.candlepin.model.Pool;
@@ -103,6 +102,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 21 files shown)


Ceiu (commented):
I haven't given this a super in-depth review since it's still a draft, but generally this looks to be taking things in the right direction. The churn from some of the maintenance cleanup bits is going to make rebasing incredibly painful, but a lot of it needs to be done anyway, so whatever.

The two things I do not like, though, are the vanishing of the scope modifiers and the expansion of the static imports in the test suites, as neither do anything for us, and are just git churn (and more unnecessary pain in the afore mentioned rebasing).

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -568,41 +569,51 @@ public void updateOwnerProductReferences(Owner owner, Map<String, String> produc
 
         Session session = this.currentSession();
 
-        Map<String, Object> criteria = new HashMap<>();
+        Map<String, Object> ownerProductCriteria = new HashMap<>();
```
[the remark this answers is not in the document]
Januson: This change was not related to performance. This is the place where manifest import fails with too many params. I did some renaming to better see what is happening with all those reassignments. I can revert those to to prevent conflicts but partitioning needs to stay otherwise import will fail.

server/src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -997,10 +976,27 @@ public Entitlement replicate(Entitlement ent) {
             .createAlias("ent_pool.sourceStack", "ss", org.hibernate.sql.JoinType.LEFT_OUTER_JOIN)
             .add(Restrictions.isNull("ss.id"));
 
-        return this.cpQueryFactory.<Entitlement>buildQuery(this.currentSession(), criteria);
+        return criteria.list();
     }
 
     @SuppressWarnings("unchecked")
+    public List<Entitlement> findByStackIds(Collection<String> stackIds) {
```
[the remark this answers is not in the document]
Januson: Done.

server/src/test/java/org/candlepin/controller/PoolManagerFunctionalTest.java
```diff
@@ -71,25 +73,25 @@
...
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 import javax.persistence.EntityNotFoundException;
 
 
+class PoolManagerFunctionalTest extends DatabaseTestFixture {
```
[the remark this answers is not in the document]
Januson: I tried to apply suggestion from JUnit5 to use default visibility and forgot to revert the commit. Modifiers are back now.

server/src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1145,6 +1147,21 @@ public void batchDelete(Collection<Pool> pools, Collection<String> alreadyDelete
         return (List<Pool>) getPools.list();
     }
 
+    @SuppressWarnings("unchecked")
+    public List<Pool> getSubPoolForStackIds(Collection<String> stackIds) {
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1937,49 +1938,57 @@ public void revokeEntitlements(List<Entitlement> entsToRevoke, Set<String> alrea
 
         log.info("Starting batch delete of pools");
         poolCurator.batchDelete(poolsToDelete, alreadyDeletedPools);
-        for (Pool pool : poolsToDelete) {
-            this.sink.queueEvent(this.eventFactory.poolDeleted(pool));
-        }
+        firePoolDeletedEvents(poolsToDelete);
         log.info("Starting batch delete of entitlements");
         entitlementCurator.batchDelete(entsToRevoke);
```
nikosmoum, line 1943: I'm confused. Your comment says "Replace batch delete of entitlements with batch delete by ids", but this is unchanged (it's still deleting by entity)?

Januson: That is the issue I mentioned during scrum. With delete by ids I am hitting issues with detached entities. Not sure why that happens so I removed it from PR until it is fixed.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -595,15 +594,19 @@ public void updateOwnerProductReferences(Owner owner, Map<String, String> produc
...
-            log.debug("{} provided products updated", count);
-
-            count = this.bulkSQLUpdate("cp2_pool_derprov_products", "product_uuid", uuidMap, criteria);
-            log.debug("{} derived provided products updated", count);
+            for (List<String> poolIdBlock : this.partition(ids, getBatchBlockSize())) {
+                Map<String, Object> updateCriteria = new HashMap<>();
+                updateCriteria.put("product_uuid", productUuidMap.keySet());
+                updateCriteria.put("pool_id", poolIdBlock);
+
+                count = this.bulkSQLUpdate("cp2_pool_provided_products", "product_uuid",
```
Ceiu, line 602: This will make debugging a pain. We should keep separate counts and then output the totals when we're done partitioning.

Ceiu (changes requested):
I only spotted a couple of minor things. Looks good otherwise, I think.

Due to its size and constantly being pulled back into other things, I would like a second set of eyes to make sure I didn't overlook anything.

server/src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -595,15 +594,19 @@ public void updateOwnerProductReferences(Owner owner, Map<String, String> produc
...
+                updateCriteria.put("product_uuid", productUuidMap.keySet());
+                updateCriteria.put("pool_id", poolIdBlock);
+
+                int providedProductsUpdated = this.bulkSQLUpdate("cp2_pool_provided_products",
+                    "product_uuid", uuidMap, updateCriteria);
+                log.debug("{} provided products updated", providedProductsUpdated);
+
+                int derivedProductsUpdated = this.bulkSQLUpdate("cp2_pool_derprov_products", "product_uuid",
+                    uuidMap, updateCriteria);
+                log.debug("{} derived provided products updated", derivedProductsUpdated);
```
Ceiu, line 608: Sorry, I guess I wasn't clear on this last time. I meant moving the output statements to outside the loop rather than a bunch of individual updates that will elongate output and require manual math during debugging.

Januson: Done.

Ceiu (approved):

PR #2868, opened 2020-12-16
[F] Added support for updating development pools and products during refresh (ENT-1910)

Ceiu:
- Refresh will now update development products (and, by extension,
  pools) for the org
- Refactored the hosted test logic to separate the adapter
  implementations and data storage
- Changed how unused entity deletion works, such that entities which
  are technically unused will no longer be deleted if they are still
  present in the imported data during refresh


wolfdale (approved):
Looks good, ack.

PR #2871, opened 2020-12-18
ENT-3088: Fix container repo generation & add detached mode

nikosmoum:
- Add packages required for the generation of test content & rpms
- Add flag for detached mode to the test script, which will run
  containers detached from the docker/test script and will not
  stop/remove them once it exits. It will also expose ports 8443,
  8080 and 22 to the host machine


sonalidhome (approved):
ack.

PR #2872, opened 2020-12-18
[M] Added missing query modifiers to the entity version lookup queries

Ceiu:
- Removed the unnecessary ordering and added the missing distinct
  keyword to the product and content lookup by version queries
- Fixed a bug that was causing the orphan cleanup job to output
  the product and content counts backward


sonalidhome (approved):
ACK.

PR #2874, opened 2021-01-04
Fix date in tests broken by new year

Januson:


Ceiu:
Wouldn't be a proper new year if we didn't have a hard-coded-date-related test failure. ACK

PR #2876, opened 2021-01-06
[3.1] Increased the Artemis disk usage limit (ENT-3337)

Ceiu:
- Increased the max-disk-usage Artemis setting from the default 90
  to 99, to help avoid erroneously hitting the limit before
  running out of disk space on larger, modern disks/drives


wolfdale (approved):
ack

PR #2879, opened 2021-01-11
ENT-3390: Implement trusting of jenkins pipelines

sonalidhome:
- Added a library `fh-pipeline-library` for
   trusting Jenkins pipeline.


nikosmoum (changes requested):
Aside from not needing to manually trust members of our github org, we should probably not rely on the external library from feedhenry, and instead fork it under the Candlepin org

PR #2884, opened 2021-01-14
[3.1]ENT-3286: Manifest delete takes 24 hours or more in large environment (60k consumers in owner)

Januson:
- Fix too many params issue in manifest import
 - Refactor updatePoolsFromStack to avoid calling deletePools n times


nikosmoum (approved):
ACK

PR #2887, opened 2021-01-19
ENT-2348: Show activation key name when consumer has been registered via activation key

wolfdale:
Changes made to show activation key name when consumer has been registered via activation key.


Januson (approved):

sonalidhome (approved):
ack.

PR #2888, opened 2021-01-19
ENT-3297: Port Cloud VM autoregistration endpoints & DTOs to OpenApi

Januson:


server/src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -55,16 +46,13 @@ public CloudRegistrationResource(I18n i18n, CloudRegistrationAuth cloudRegistrat
...
-    @POST
-    @Path("authorize")
-    @Produces(MediaType.TEXT_PLAIN)
-    @Consumes(MediaType.APPLICATION_JSON)
+    @Override
     @SecurityHole(noAuth = true)
-    public String authorize(CloudRegistrationDTO cloudRegDTO,
-        @Context Principal principal) {
-
+    public String cloudAuthorize(CloudRegistrationDTO cloudRegistrationDTO) {
```
Ceiu, line 51: Why did the name of the method change here?

Januson: I used cloudAuthorize as an operation id and it generates method names from the operation id. It has to be unique but we don't have such an endpoint yet so I can change it to authorize if you want.

api/candlepin-api-spec.yaml
```diff
@@ -555,6 +555,57 @@ paths:
...
+          description: Cloud provider or account details could not be resolved to an organization
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ExceptionMessage'
+              example:
+                displayMessage: Cloud provider or account details could not be resolved to an organization
+                requestUuid: c4347004-8792-41fe-a4d8-fccaa0d3898a
+        501:
+          description: Cloud provider or account details could not be resolved to an organization
```
sonalidhome, line 598: An error message is not updated here after pasting the error block. The example shows the correct error message.

sonalidhome (commented):
Minor error message change required. Other than that changes look good to me.

api/candlepin-api-spec.yaml

server/src/main/java/org/candlepin/resource/CloudRegistrationResource.java
Ceiu: Nah, that's not critical. Just wanted understanding as to why it changed.

api/candlepin-api-spec.yaml
```diff
@@ -555,6 +555,57 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /cloud:
+    post:
```
nikosmoum, line 559: A note here: this was not accurately ported. The original code had `@Path("/cloud")` on the class and `@Path("authorize")` on the method, which would effectively be a `POST /cloud/authorize`, while the way you ported would result in a `POST /cloud`

PR #2894, opened 2021-01-20
[2.6] ENT-2725: Update system purpose status to NOT consider Layered SLA

sonalidhome:
- Updated the `rules.js` to ignore pool in priority calculation
 - Updated `SystemPurposeComplianceRules` to ignore

### DIFF
--- a/server/spec/autobind_spec.rb
+++ b/server/spec/autobind_spec.rb
@@ -682,4 +682,56 @@ describe 'Autobind On Owner' do
     entitlements = @cp.list_entitlements(:uuid => consumer.uuid)
     entitlements.size.should == 1
   end
+
+  it 'should be not specified with consumer on auto attach when SLA of current entitlement is layered exempt' do
+    product1 = create_product(random_string('product'),
+                              random_string('product'),
+                              {:attributes => {:support_level => 'Standard'},
+                               :owner => owner_key})
+    product2 = create_product(random_string('product'),
+                              random_string('product'),
+                              {:attributes => {:support_level => 'Layered',
+                                               :support_level_exempt => 'true'},
+                               :owner => owner_key})
+
+    @cp.create_pool(owner_key, product1.id)
+    @cp.create_pool(owner_key, product2.id)
+
+    installed = [
+        {'productId' => product1.id, 'productName' => product1['name']},
+        {'productId' => product2.id, 'productName' => product2['name']}]
+
+    consumer = @cp.register(
+        random_string('system'), :system, nil, {}, nil, owner_key, [], installed, nil, [],
+        nil, [], nil, nil, nil, nil, nil, 0, nil, 'Layered')
+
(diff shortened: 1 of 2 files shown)


Ceiu (changes requested):
Seems good, and looks like it follows the accepted design.

Just one oddity with the spec test and I'm okay with it.

server/spec/autobind_spec.rb
```diff
@@ -682,4 +682,42 @@
...
+    expect(status['status']).to eq('not specified')
+
+    entitlements = @cp.list_entitlements(:uuid => consumer.uuid)
+
+    # Trying to bind the without SLA exempt.
+    # Then system purpose status should be match.
+
+    entitlements.each do |ent|
+      @cp.unbind_entitlement(ent.id, {:uuid => consumer.uuid})
+    end
```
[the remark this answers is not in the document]
sonalidhome: Good catch! :) 
I updated the spec-test to test more stuff in.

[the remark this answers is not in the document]
sonalidhome: Good catch! :)
I updated the spec-test to test more stuff in.

server/src/main/resources/rules/rules.js
```diff
@@ -506,20 +506,25 @@ function get_pool_priority(pool, consumer) {
         var null_rule_score = 0;
         var mismatch_rule_score = 0;
 
-        log.debug("Evaluating attribute {} with weight {}", attr, attrs[attr]);
-        if (unsatisfiedSet.length === 0 && poolSet.length === 0) {
-            null_rule_score = 0.1;
-            log.debug("NULL rule score for attribute {} and pool {}, which is: {}", [attr, pool.id, null_rule_score]);
+        if(Utils.equalsIgnoreCase(attr, 'support_level') && Utils.equalsIgnoreCase(poolSet[0], 'Layered')) {
+            log.debug("Ignoring the pool prioritization when SLA is Layered", attr, poolSet[0]);
```
nikosmoum, line 510: Similarly to the check for the syspurpose compliance, we shouldn't be checking by hardcoding `Layered` here, but instead checking if pool we are evaluating has the `support_level_exempt` attribute set to true. Something like `pool.retrievePoolAttributeValues("support_level_exempt")` should work here (although I have not tested it, so it needs verification).

nikosmoum (changes requested):
One thing to change. Otherwise we're good. Also, after this is merged, let's also fix this in master

Edit: As discussed, let's remove the bz id from the commit/PR

server/src/main/resources/rules/rules.js
nikosmoum: UPDATE: based on the latest discussion, we should actually remove these rules.js changes all together

sonalidhome: Removed the changes related to rules.js.

PR #2896, opened 2021-01-20
[F] Updated the OrphanCleanupJob to be aware of product to product references (ENT-2011)

Ceiu:
- OrphanCleanupJob will now avoid removing orphaned products which are
  still referenced by other products
- Fixed a null pointer exception that could occur in OwnerProductCurator
  if creating owner-product links with a null owner or product collection


sonalidhome (approved):
Changes look good to me.

nikosmoum (approved):

PR #2899, opened 2021-01-21
[F] Updated tooling to use new tables and APIs for children products (ENT-3429)

Ceiu:
- Test data importer now properly adds provided and derived products
  to the product, rather than the pools
- Updated the org migrator to be aware of the product provided
  products, and no longer manage the provided products tables
  associated with pools

### DIFF
--- a/server/bin/import_test_data.rb
+++ b/server/bin/import_test_data.rb
@@ -182,7 +182,7 @@ def create_product(cp, owner, product)
   product_content = product['content'] || []
   dependent_products = product['dependencies'] || []
   relies_on = product['relies_on'] || []
-  derived_product_id = product['derived_product_id']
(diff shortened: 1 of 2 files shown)


Ceiu:
For those interested in testing the org migrator: 

- import the test data
- arbitrarily choose any of the orgs created during test data import
- export the chosen org using the org migrator
- delete the other orgs using the org deletion endpoint
- dump the SQL which should contain only the data for the chosen org
- clear the database entirely
- import the data for the chosen org using the org migrator
- dump the SQL again
- Use a diff tool to verify the SQL dumps are mostly identical

Note that due to how the migrator works, there may be minor amounts of drift in the timestamps, but otherwise everything should be mostly identical.

nikosmoum (approved):

PR #2900, opened 2021-01-22
[M] ENT-2725: Update system purpose status to NOT consider Layered SLA

sonalidhome:


server/spec/autobind_spec.rb
```diff
@@ -822,4 +822,38 @@
...
+    @cp.consume_product(product2['id'], {:uuid => consumer.uuid})
+
+    status = @cp.get_purpose_compliance(consumer['uuid'])
+    expect(status['status']).to eq('not specified')
+
+    entitlements = @cp.list_entitlements(:uuid => consumer.uuid)
+
+    entitlements.each do |ent|
+      @cp.unbind_entitlement(ent.id, {:uuid => consumer.uuid})
+    end
```
nikosmoum, line 856: The final check here is missing

nikosmoum (changes requested):
1 minor issue; otherwise good

nikosmoum:
Also, please avoid using the flags like [M] in the commit messages. We only add it on the PR summary for convenience

server/spec/autobind_spec.rb
sonalidhome: Updated.

PR #2901, opened 2021-01-22
ENT-2252: Upgrade openapi-generator to latest version

Januson:
- Update gradle to 6.8
 - Update junit5 to 5.7.0
 - Update openapi to 5.0.0
 - Update overriden openapi templates
 - Update AbstractJavaJAXRSServerCodegen to generate default responses with HTTP response code 500


buildSrc/src/main/java/org/openapitools/codegen/languages/AbstractJavaJAXRSServerCodegen.java
```diff
@@ -0,0 +1,316 @@
...
+                    if (isMultipartPost) {
+                        parameter.vendorExtensions.put("x-multipart", "true");
+                    }
+                }
+
+                List<CodegenResponse> responses = operation.responses;
+                if (responses != null) {
+                    for (CodegenResponse resp : responses) {
+                        if ("0".equals(resp.code)) {
+                            resp.code = "500";
```
nikosmoum, line 213: Let's separate the PR into 2 commits to serve as better documentation: first commit adds this file as is, and has a summary `Add default AbstractJavaJAXRSServerCodegen.java from openapi-generator`, and the 2nd commit is all the rest of the changes, with additionally changing this line to return "500" and it's summary will be the same as the current commit, but also adding the explanation for the 500.

nikosmoum (changes requested):
ACK on the whole thing; just 1 minor change, to split the PR into 2 (in order for the custom change that was done in https://github.com/candlepin/candlepin/pull/2809 to be explicitly visible)

buildSrc/src/main/java/org/openapitools/codegen/languages/AbstractJavaJAXRSServerCodegen.java
Januson: Done

nikosmoum:
One more thing I forgot: since some packages were updated (junit, openapi-generator-maven-plugin), we also need to add the corresponding changes in the pom.xml files by running `./gradlew pom`

PR #2905, opened 2021-01-22
[M] Fixed a bug with Content equality checks for fields with empty strings

Ceiu:
- Fixed a bug that could cause equality checks for two Content
  instances to fail if one had empty strings for certain fields
  and the other had nulls

### DIFF
--- a/server/src/main/java/org/candlepin/model/Content.java
+++ b/server/src/main/java/org/candlepin/model/Content.java
@@ -53,6 +53,7 @@ import javax.xml.bind.annotation.XmlRootElement;
 import javax.xml.bind.annotation.XmlTransient;
 
 
+
 /**
  * ProductContent
  */
@@ -338,7 +339,7 @@ public class Content extends AbstractHibernateObject implements SharedEntity, Cl
 
     @Override
     public String getContentUrl() {
-        return contentUrl;
+        return this.contentUrl == null || this.contentUrl.isEmpty() ? null : this.contentUrl;
     }
 
     public void setContentUrl(String contentUrl) {
@@ -351,7 +352,7 @@ public class Content extends AbstractHibernateObject implements SharedEntity, Cl
      */
     @Override
     public String getRequiredTags() {
-        return requiredTags;
+        return this.requiredTags == null || this.requiredTags.isEmpty() ? null : this.requiredTags;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -338,7 +339,7 @@ public void setVendor(String vendor) {
 
     @Override
     public String getContentUrl() {
-        return contentUrl;
+        return this.contentUrl == null || this.contentUrl.isEmpty() ? null : this.contentUrl;
```
nikosmoum, line 342: I don't know in what context the bug occurred, but maybe it is better to move this logic out of the getters and into the equals method? Is it reasonable to think that leaving this in the getters might cause other problems? Theoretical example in https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/controller/ContentManager.java#L753-L755, where if, for whatever reason, the DTOs' contentUrl is an empty string, while the content model's contentUrl always returns null (instead of empty string), resulting in 'updating' the content, which is unneeded churn?

nikosmoum (changes requested):
Couple of questions

server/src/main/java/org/candlepin/model/Content.java
```diff
@@ -509,21 +510,25 @@ public boolean equals(Object other) {
...
+                .append(this.getId(), that.getId())
+                .append(this.getType(), that.getType())
+                .append(this.getLabel(), that.getLabel())
+                .append(this.getName(), that.getName())
+                .append(this.getVendor(), that.getVendor())
+                .append(this.getMetadataExpiration(), that.getMetadataExpiration())
+
+                // These fields require special consideration, as nulls and empty strings are
+                // considered identical for CP's purposes. The accessors should fix this for us,
+                // but if equality checks start failing in the future, this is something to check.
```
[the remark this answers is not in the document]
Ceiu: Done.

server/src/main/java/org/candlepin/model/Content.java
Ceiu: I thought about that, and came to the conclusion that we want our business logic to handle it as soon as it can. Since we transform empty strings to nulls at some point with the goal of treating them the same, it makes sense to me that we should start processing them that way as early and completely as possible.

In the DTO case, the behavior you're describing is already how it happens today: because the translation happens at the DB layer, the DTO will be pulling the DB entity, which will already have converted the previous empty string to null, and the null will be updated to null again. This isn't exactly optimal, but without additional checks to verify that we're not clearing an already cleared field, then the behavior is as-expected and as-designed with respect to the slightly different semantics the API layer has for empty strings vs nulls.

Ceiu: I suppose the alternate approach here would be to instead modify it in the mutator rather than the accessor. I'm down with either.

PR #2907, opened 2021-01-27
[M] 1920582: Added additional validation for cloud registration data (ENT-3440)

Ceiu:
- Added an additional check for data in the /cloud/authorize endpoint
  to trigger a 400 rather than a 500 when no cloud registration
  details are provided
- Fixed a bug with the serialization of exception output when writing
  out a CandlepinException that does not have an exception message
- Candlepin will now list "cloud_registration" as a capability when
  the cloud registration feature is enabled
- Added support for cloud registration to the spec tests
- Added additional testing surrounding the cloud registration feature
- Other minor improvements and optimizations to the spec tests


server/spec/cloud_registration_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+
+
+describe 'Cloud Registration' do
+
+    include CandlepinMethods
+
+    before(:each) do
+        skip("candlepin running in standalone mode") unless is_hosted?
+        skip("cloud registration is not enabled") unless @cp.is_capability_present("cloud_registration")
```
nikosmoum, line 13: Given that, by default, cloud authentication is set to false, and the capability is only present if that property is set to true, that means this test suite will always be skipped when Jenkins will be running spec tests. The alternatives I can think of are either turn cloud auth on by default (probably not a great idea), or, somehow setting it in candlepin.conf before we run spec tests on jenkins.

nikosmoum (changes requested):
Good change as a whole, but the cloud reg spec tests aren't going to be running by default

server/spec/cloud_registration_spec.rb
Ceiu: Yup. There are a number of ways to handle this, but the critical part here is getting the tests available. We can sort out the details in making them autorun in our testing infrastructure later.

On that note, the easiest would probably be having Jenkins write a custom.yaml with the cloud reg setting enabled. That way we don't need to do much on the infrastructure side to get it going.

server/spec/test_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+require 'candlepin_scenarios'
+require 'unpack'
+
+describe 'Certification verification' do
+  include CandlepinMethods
+  include CertificateMethods
+  include Unpack
+
+
+  it 'certificates should contain all content' do
```
nikosmoum, line 11: how does this file relate to the change? was it added by accident?

Ceiu: It doesn't, but it keeps sneaking in there when I do a "git add server/spec" without double checking what got added.

Removed.

PR #2908, opened 2021-01-27
[F] Updated migration for product hierarchy changes (ENT-3445)

Ceiu:


Ceiu:
To test:
- Deploy current master with test data (-gta or -gtma)
- Make note of current pools with derived products and provided products with the following queries:
```
SELECT DISTINCT p.product_uuid, p.derived_product_uuid FROM cp_pool p WHERE p.derived_product_uuid IS NOT NULL;

SELECT DISTINCT pool.product_uuid AS product_uuid, ppp.product_uuid AS provided_product_uuid
 FROM cp2_pool_provided_products ppp
 JOIN cp_pool pool ON pool.id = ppp.pool_id
UNION
SELECT DISTINCT pool.derived_product_uuid AS product_uuid, dpp.product_uuid AS provided_product_uuid
 FROM cp2_pool_derprov_products dpp
 JOIN cp_pool pool ON pool.id = dpp.pool_id;
```
- Check out and deploy this branch (no params, or -m)
- Verify the deployment succeeds
- Verify the database has been properly updated with the following queries:
```
SELECT uuid, derived_product_uuid FROM cp2_products WHERE derived_product_uuid IS NOT NULL;

SELECT * FROM cp2_product_provided_products;
```
The first query will show the new column on the products table (derived_product_uuid), and it should be null in all cases except the ones where a mapping existed in the first query from the initial deployment. The query here should return an identical set of results as the original query.

The second query should return the same mappings that were detected in the original deployment.

Both sets of results can be verified quickly by using a diff tool on the output on the command line from either psql or mysql.

nikosmoum (approved):

PR #2909, opened 2021-01-29
1919458: invalid Content-Type yields 500 (ENT-3439)

sonalidhome:
- Added a CandlepinContentTypeFilter. It will execute before
   each request to validate the content-type.
 - Added the RSpec test.

### DIFF
--- a/server/spec/filter_response_spec.rb
+++ b/server/spec/filter_response_spec.rb
@@ -78,4 +78,17 @@ describe 'Response JSON Filtering' do
       consumer["owner"]["id"].should_not == nil
     end
   end
+
+  it 'should get valid response on invalid content types' do
+    lambda {
+      @cp.get_response("/consumers/#{@consumer1.uuid}", 'application|json')
+    }.should raise_exception(RestClient::BadRequest)
+
+    response = @cp.get_response("/consumers/#{@consumer1.uuid}",'application/json')
+    expect(response.code).to eq(200)
+
+    lambda {
+      @cp.get_response("/consumers/#{@consumer1.uuid}", 'application-json')
+    }.should raise_exception(RestClient::BadRequest)
+  end
 end
--- a/server/src/main/java/org/candlepin/servlet/filter/CandlepinContentTypeFilter.java
+++ b/server/src/main/java/org/candlepin/servlet/filter/CandlepinContentTypeFilter.java
@@ -1,0 +1,79 @@
+/**
+ * Copyright (c) 2009 - 2021 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.servlet.filter;
+
+import com.google.inject.Inject;
+import com.google.inject.Injector;
+import com.google.inject.Singleton;
+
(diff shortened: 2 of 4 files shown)


server/src/main/java/org/candlepin/servlet/filter/EventFilter.java
```diff
@@ -47,29 +49,44 @@
...
+            else {
+                eventSink.rollback();
+                log.debug("Request failed, skipping event sending, status={}", status);
+            }
         }
-        else {
-            eventSink.rollback();
-            log.debug("Request failed, skipping event sending, status={}", status);
+        catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(this.i18n.tr("Invalid Content-Type {0}", contentType));
```
nikosmoum, line 89: Unfortunately this solution will lead to the same stacktrace but with a `IllegalArgumentException` instead of an NPE. Because we're not at the resteasy layer here, we can't rely on the `ExceptionMappers` to send back a proper response, and instead we have to build/send one ourselves. The following piece of code at the start of the method should work, and also avoids the whole Tomcat HTML output:
```
    @Override
    public void doFilter(ServletRequest request, ServletResponse response,
        FilterChain chain) throws IOException, ServletException {

        I18n i18n = injector.getInstance(I18n.class);
        // Validate media type contains a slash, to avoid RESTEasy's inability to handle this
        String contentType = request.getContentType();
        if(contentType != null && !contentType.contains("/")) {
            HttpServletResponse httpResponse = (HttpServletResponse) response;
            httpResponse.setContentType("text/plain");
            httpResponse.setCharacterEncoding("UTF-8");
            httpResponse.setStatus(HttpServletResponse.SC_BAD_REQUEST);
            httpResponse.getWriter().write(i18n.tr("Invalid Content-Type {0}", contentType));
            return;
        }
```

nikosmoum (changes requested):
It is surprising that RESTEasy does not handle content-type values without a slash; I couldn't find any relevant bug for that in their tracker. This approach though has the same problem. I've recommended an alternative solution in the comment

Ceiu (changes requested):
This code only works if the EventFilter happens to be the first filter in the chain to touch the MediaType.

If we're going to work around the issue like this, we should put this in its own filter, and probably adjust the binding priority so that it is invoked as early in the chain as possible.

server/src/main/java/org/candlepin/servlet/filter/EventFilter.java
sonalidhome: Updated the code in the Filter class.

server/src/main/java/org/candlepin/servlet/filter/CandlepinContentTypeFilter.java
```diff
@@ -0,0 +1,70 @@
...
+        FilterChain chain) throws IOException, ServletException {
+
+        // Validate media type contains a slash, to avoid RESTEasy's inability to handle this
+        String contentType = request.getContentType();
+
+        if (contentType != null && !contentType.contains("/")) {
+            HttpServletResponse httpResponse = (HttpServletResponse) response;
+            httpResponse.setContentType("text/plain");
+            httpResponse.setCharacterEncoding("UTF-8");
+            httpResponse.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
nikosmoum, line 54: This isn't sending a message back though. something like this should do: `httpResponse.getWriter().write(i18n.tr("Invalid Content-Type {0}", contentType));`

sonalidhome: Added.

server/spec/filter_response_spec.rb
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 require 'spec_helper'
 require 'candlepin_scenarios'
+require 'candlepin_api.rb'
```
nikosmoum, line 4: `candlepin_scenarios` already requires `candlepin_api` so it's available transitively. no need for this require

nikosmoum (changes requested):
Only a minor redundancy, and then should be good to merge

server/spec/filter_response_spec.rb
sonalidhome: Done.

PR #2914, opened 2021-02-01
[M] Activation keys can no longer be created or updated with pools in SCA mode (ENT-3089)

Ceiu:
- While SCA mode is enabled for a given organization, activation keys can
  no longer be created with pools, or have pools added to them


Ceiu:
This is ready for review, but is pending further discussions with all of the affected stakeholders.

nikosmoum:
Closing this because we still need to get some answers from other stakeholders before applying this change on our side. We'll revisit.

PR #2915, opened 2021-02-02
ENT-2315: Changes to fix Candlepin API docs

wolfdale:
Changes 
- Deprecated ```/candlepin/swagger.json``` endpoint
- Removed unwanted swagger related code.
- Updated Swagger UI  to render ```candlepin-api-spec.yaml```


api/candlepin-api-spec.yaml
```diff
@@ -31,6 +31,34 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /openapi.json:
```
nikosmoum, line 34: I am not sure if we need to serve these dynamically. Once you delete our current static Swagger UI under server/src/main/webapp and instead deploy the latest swagger version as described in the "Plain old HTML/CSS/JS (Standalone)" section here: https://swagger.io/docs/open-source-tools/swagger-ui/usage/installation/ then that should be able to serve both the UI and the json file alone. That should be enough (not sure it's worth it adding an extra endpoint just for the yaml)

nikosmoum (changes requested):

build.gradle
```diff
@@ -439,6 +439,10 @@ project(":api") {
         rename { String fileName ->
             api_spec_path.endsWith(fileName) ? 'openapi.yaml' : fileName  // rename yaml to openapi.yaml
         }
+        copy {
+            from "${projectDir}/candlepin-api-spec.yaml"
+            into "${rootDir}/server/src/main/webapp/docs/"
+        }
```
nikosmoum, line 445: we should also make sure this copying happens on the maven side too (since tito uses maven to build the war)

api/candlepin-api-spec.yaml
wolfdale: I thought our goal was to deprecate ```/candlepin/swagger.json``` end point, which is used to generate the API spec by scanning the code therefore I created these endpoints to serve openAPI spec directly. 
In recent changes 
- I removed the ```/candlepin/openapi.json``` endpoint since the generated spec have issues with (offsetdatetime, default response).
- Kept the ```/candlepin/openapi.yaml``` endpoint since we need fetch raw OpenAPI spec (Just like /swagger.json endpoint).
- Updated to new Swagger UI to render ```candlepin-api-spec.yaml```.

nikosmoum: Yes, when I wrote this comment I wasn't yet aware that swagger.json has issues.

Although I still don't think we need the `/candlepin/openapi.yaml` endpoint, since the yaml is still served statically, and is downloadable from the /docs page. You can either click on it on the Swagger UI page, or just do `curl -k https://localhost:8443/candlepin/docs/candlepin-api-spec.yaml` to get it.

There is still the question though of serving the swagger.json. Personally, I don't think we should serve it since it has the problems you mentioned, but maybe it still has value (because some people might prefer to parse json instead of yaml)? (I think this might be a topic of discussion with the team).

build.gradle
wolfdale: Good point. Made changes to ```api/pom.xml```.

api/candlepin-api-spec.yaml
```diff
@@ -31,6 +30,20 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /openapi.yaml:
```
nikosmoum, line 33: As I described in the other comment, we're already serving this statically, so there is no reason for a new endpoint

wolfdale: Removed.

build.gradle
nikosmoum: You can't add plugins in api/pom.xml, due to the fact we auto-generate the poms with the `gradlew pom` command. Look at the `task pom` in build.gradle to see how to add the plugins during the auto-generation.

api/candlepin-api-spec.yaml
```diff
@@ -1,4 +1,3 @@
-
 openapi: 3.0.2
 info:
   title: Candlepin
```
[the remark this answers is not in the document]
wolfdale: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2546,7 +2524,6 @@ paths:
       description: Retrieves a list of known consumer types
       tags:
           - Consumer
-          - ConsumerType
```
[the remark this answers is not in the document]
wolfdale: Done.

api/candlepin-api-spec.yaml
```diff
@@ -8205,7 +8156,6 @@ paths:
     get:
       description: Retrieve a list of owners the user can register systems to
       tags:
-        - Owners
         - Users
```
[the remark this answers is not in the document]
wolfdale: Sounds good.

build.gradle
wolfdale: I was not aware of this. Changes made to ```build.gradle```

api/candlepin-api-spec.yaml
```diff
@@ -8157,7 +8109,6 @@ paths:
     get:
       description: Retrieves a list of Roles by User
       tags:
-        - Roles
         - Users
       operationId: getUserRoles
       parameters:
```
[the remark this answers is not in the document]
wolfdale: Done.

build.gradle
nikosmoum: In addition to making the plugins auto-generate from the gradle pom command, we should also run the command and commit the pom changes as well. (this is basically what we do now when updating any library in build.gradle, we run `./gradlew pom` and commit both the gradle & pom changes

wolfdale: Done.

PR #2921, opened 2021-02-04
[F] ENT-3503: Remove obsolete swagger deps & annotations

nikosmoum:
- Remove remaining swagger annotations now that we use
  openapi-generator to generate them
- Remove the 'checks' module since the only thing there was
  a custom checkstyle check for missing swagger annotations
- Remove redundant dependencies swagger-jaxrs, swagger-core
  and swagger-models


Ceiu:
ack

PR #2922, opened 2021-02-05
[F] copy API spec file into the war

nikosmoum:


barnabycourt:
Please do not use a new task for this. The file can be placed anywhere in the final war by adding a second "from" section to the war task in the build.gradle. This can be done in the same way that we copy the license into the meta-inf directory.

Ceiu:
ack

image

PR #2925, opened 2021-02-09
ENT-3432: Remove internal registry hostname from github

sonalidhome:
- Updated the code to use `QUAY.IO` open source docker registry
 - Removed the code of candlepin-rhel7 image


docker/docker-compose-build.yml
```diff
@@ -5,9 +5,6 @@ services:
   candlepin-base:
     build: candlepin-base/
     image: ${REGISTRY}/candlepin-base
-  candlepin-rhel7:
-    build: candlepin-rhel7/
-    image: ${REGISTRY}/candlepin-rhel7
```
nikosmoum, line 10: If we're removing this from the compose, we should also remove the whole docker/candlepin-rhel7 directory

jenkins/Jenkinsfile
```diff
@@ -15,7 +15,17 @@ pipeline {
             }
         }
         stage('Test') {
+            environment {
+                CANDLEPIN_QUAY_BOT = credentials("candlepin-quay-bot")
+            }
             parallel {
+                stage('quay-login') {
```
nikosmoum, line 22: I am not sure it will work if we add this stage in the parallel{} block. There might be race conditions where one of the other stages that need to do a docker pull cannot do so due to the login having not run yet. Maybe it would be better if this stage was moved out of parallel{}, between the Trust and Test stages?

nikosmoum (changes requested):
A couple of items to address and then should be good

nikosmoum:
These changes will also have to be ported to the hotfix branches

nikosmoum:
This seems to be failing with this now:
```
+ docker login -u **** -p '' quay.io
Error: Cannot perform an interactive login from a non TTY device
```

PR #2926, opened 2021-02-10
[M] Removed SubscriptionInfo.getDerivedProduct and ImportedEntityCompiler

Ceiu:
- Removed the getDerivedMethod method from SubscriptionInfo and
  Subscription
- Removed the now-obsolete ImportedEntityCompiler class


server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -999,23 +999,11 @@ public Pool convertToMasterPool(SubscriptionInfo sub) {
...
-                if (pdata.getId() == null || pdata.getId().isEmpty()) {
-                    throw new IllegalStateException("Subscription references an incomplete product: " +
-                        pdata);
-                }
-
-                productIds.add(pdata.getId());
-            }
-        }
+        // Gather derived and provided product IDs...
+        this.gatherSubscriptionProductIds(productIds, sub.getProduct());
```
nikosmoum, line 1006: I was looking into if we even need to do a tree traversal here. The reason we were building the `productIds` Set was so that later in the `convertToMasterPoolImpl` method we could link the products to the Pool object. But now Pool only has a single link (the main product), so we can just fetch that single product, put it in the `productMap` and pass it in `convertToMasterPoolImpl`?

Another note here: This method (`convertToMasterPool(SubscriptionInfo sub)`) is not actually being used any more in production in any sense. Looking at the code paths that lead to it (excluding the tests), there's two of them:
1) The Refresher.run() method, which, if you check the comment above, that part of the code is dead at the moment.
2) The POST /owners/{owner_key}/subscriptions endpoint, which is deprecated, and is not called in either Katello or Hosted, and in fact is already removed in the spec first branch.
So in the end, this whole method is useless and can eventually be removed

nikosmoum (changes requested):
1 issue/question, but otherwise good.

Also, it turns out the spec test that fails is failing consistently in master (only in postgres for some reason), that @sonalidhome is looking at

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
Ceiu: > I was looking into if we even need to do a tree traversal here. The reason we were building the productIds Set was so that later in the convertToMasterPoolImpl method we could link the products to the Pool object. But now Pool only has a single link (the main product), so we can just fetch that single product, put it in the productMap and pass it in convertToMasterPoolImpl?

Hrmm... yeah, that's true. Especially with knowledge of how ```convertToMasterPoolImpl``` works. I'll drop the tree traversal and just send a singleton map in with the one product entry.

Regarding the rest of that, we should either mark this method deprecated or remove it entirely when the spec-first bits go through.

nikosmoum: Turns out that we have an investigation task in the backlog to possibly revive that code path, so maybe we shouldn't remove it entirely for now.

PR #2929, opened 2021-02-12
ENT-3546:Address sporadic spec test failure in 

sonalidhome:
derived_product_import_spec.

 - Updated the code to setup the products and pool
   in the correct order.

### DIFF
--- a/server/spec/derived_product_import_spec.rb
+++ b/server/spec/derived_product_import_spec.rb
@@ -35,25 +35,32 @@ describe 'Import', :serial => true do
   end
 
   it 'can retrieve subscripton certificate from derived pool entitlement' do
-    stacked_datacenter_product = create_product(nil, nil, {
-      :attributes => {
-        :virt_limit => "unlimited",
-        :stacking_id => 'mixed-stack',
-        :sockets => "2",
-        'multi-entitlement' => "yes"
-      }
-    })
+
     derived_product = create_product(nil, nil, {
       :attributes => {
           :cores => '6',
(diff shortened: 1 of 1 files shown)


server/spec/derived_product_import_spec.rb
```diff
@@ -54,15 +54,20 @@
       {
         :derived_product_id => derived_product.id
       })
-    pool = @cp.list_owner_pools(@owner['key'], {:product => stacked_datacenter_product.id})[0]
+
+    pools = @cp.list_owner_pools(@owner['key'], {:product => stacked_datacenter_product.id})
+
+    pool  = pools.select do |p|
+      p['type'] == 'NORMAL'
```
nikosmoum, line 61: I believe this might be the wrong way to make the test pass. It's a test about derived products, and, looking at how the test is setting up it's data, its currently wrong. We no longer specify the derived product when creating the pool, but when we create the main (marketing) product. Thus, the order of data creation would be: 1) create the derived product, 2) create the main product and specify the derived product we created in step 1, 3) create the pool by specifying only the main product (and preferably by using the `create_pool()` method instead of the deprecated `create_pool_and_subscription`.

nikosmoum (changes requested):
There's a different way this test should be set up; I've explained inline

server/spec/derived_product_import_spec.rb
sonalidhome: Updated the code to create the products and pool in the correct order.

PR #2930, opened 2021-02-15
[M] Fixed product creation order in test data importer (ENT-3552)

Ceiu:
- The Candlepin test data importer will now resolve product references
  to ensure products are created in the right order during test data
  import
- Changed product certificate generation such that any unexpected
  exception results in an InternalServerError rather than a null
  certificate

### DIFF
--- a/server/bin/import_test_data.rb
+++ b/server/bin/import_test_data.rb
@@ -53,7 +53,7 @@ def create_owner(cp, new_owner)
   displayName = new_owner['displayName']
 
   print "owner: #{owner_name}\n"
-  print "\t displayName: #{displayName}\n"
+  print "    displayName: #{displayName}\n"
 
   owner = cp.create_owner(owner_name, new_owner)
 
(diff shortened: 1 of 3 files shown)


wolfdale (approved):
ack

nikosmoum (changes requested):
Ack, but let's add the bz id in the commit message: 1928249

nikosmoum:
Also, there seems to be a related unit test failure

Ceiu:
Done. I've updated the commit message, but haven't yet addressed the test failure. Looking into that now.

Ceiu:
Test fixed. This also changed a bit of the logic in the product cert stuff. The convention I've landed on is:
- If it's a product that can have a cert, and that cert generation failed, throw an exception
- if the product cannot have a cert (following our numeric-id = engineering product logic), then the generation always returns null.

The affected test has been updated accordingly.

nikosmoum (approved):

PR #2932, opened 2021-02-16
[3.1]1929681: Candlepin will not install in FIPS mode

Januson:
- Switch filedigest algorithm to sha-256


nikosmoum (approved):
ACK, but jenkins is failing because this bz id is referring to the wrong target release. The bz needs to be cloned with Target Release set to 3.1

nikosmoum:
Merging this since the failure is unrelated

PR #2933, opened 2021-02-17
[3.1] ENT-3432: Remove internal registry hostname from github

sonalidhome:
- Updated the code to use `QUAY.IO` open source docker registry
 - Removed the code of candlepin-rhel7 image


nikosmoum (commented):
Ack on the change, but it looks like this is failing with the rate limit errors. Let's also cherry-pick the following 3 commits from master to fix that:
684cb40d3345cac0ae36b137f9593a82cf637e54
ca399517b348f7e1ee02fb7e581bf602178a4a76
f69f6bc4524a6d5175b0cec0f73fbb6ba3f1db98

nikosmoum:
It looks like I missed 1 commit. We also need this one 6ed4cde97f3c700116b5f609821f0bb28cbe6b09

PR #2935, opened 2021-02-17
[M] 1925546: Added additional checks for regenerating SCA certificates (ENT-3521)

Ceiu:
- SCA certificates will now be regenerated on request if the org's
  content, products, or pools change after the last cert update
- Removed now-obsolete fields from the update_pool method in
  candlepin_api

### DIFF
--- a/server/client/ruby/candlepin_api.rb
+++ b/server/client/ruby/candlepin_api.rb
@@ -1137,7 +1137,6 @@ class Candlepin
 
   def create_pool(owner_key, product_id, params={})
     quantity = params[:quantity] || 1
-    provided_products = params[:provided_products] || []
 
     start_date = params[:start_date] || DateTime.now
     end_date = params[:end_date] || start_date + 365
@@ -1146,8 +1145,7 @@ class Candlepin
       'startDate' => start_date,
       'endDate'   => end_date,
       'quantity'  =>  quantity,
-      'productId' => product_id,
-      'providedProducts' => provided_products.collect { |pid| {'productId' => pid} }
+      'productId' => product_id
     }
 
     if params[:branding]
@@ -1166,8 +1164,8 @@ class Candlepin
       pool['orderNumber'] = params[:order_number]
     end
 
-    if params[:derived_product_id]
-      pool['derivedProductId'] = params[:derived_product_id]
+    if params[:upstream_pool_id]
+      pool['upstreamPoolId'] = params[:upstream_pool_id]
     end
 
     if params[:source_subscription]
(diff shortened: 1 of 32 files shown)


Ceiu:
This is a bit of an odd change, and one I'm not entirely sold on myself. The jist of it is to add a date of the last known change to a given org's/owner's content view. If that content view is newer than the cert, then we need to regenerate it, even if it has not expired. The goal is to address a race condition that can occur when the SCA cert is generated while the content view is changing for the org. In such a case, it's possible for the cert to be generated without the content change, but since the content change completes before the cert is stored, it does not get invalidated. Using the dates will allow us to catch when this happens and regenerate the certificate as necessary.

For those reading this without the context of the BZ or previous discussions, the "correct" way to deal with this is to lock on something to prevent this case from ever happening at all, but Candlepin's current design does not allow for this with its current locking model without locking on the org as a whole and kneecapping performance & scalability. This approach was decided as a sufficient enough backup plan to recover from this case when it occurs.

I'm still very open to other solutions here. Additionally, one improvement we could consider here is to check the actual content paths we would include in the cert, take a hash (or other snapshot like thing) and store the hash. Then, rather than just always updating the date on any pool/product/content change, we only update the content view date when the hash/snapshot changes. This somewhat falls apart with respect to consumer-specific environments, which is why I opted not to include it in this PR; however, it is something to consider for future improvements/optimizations.
Outside of that, there is a lot of overlapping bits of code affected by this change which impacted how the implementation came together. There are some improvements that could be made here, but they may have require other far-reaching changes.

Lastly, to verify that this addresses the race condition, the testing steps specified in the BZ can be used:
1. Create the necessary objects within Candlepin for a baseline SCA cert: owner, content, products, pool, consumer
2. Put the org into SCA mode by setting the content access mode to "org_environment" (adding that mode to its list as necessary)
3a. Create a new content and attach it to the engineering product of the pool while doing the next step
3b. Start as many concurrent background tasks as possible which fetch the certificates for the org. Optionally, rather than fetching the whole cert, you can use the endpoint which only fetches the payload (```/consumers/{consumer_uuid}/accessible_content```). 

If 3a and 3b are done at the same time, the first cert payload fetched by step 3b should reflect the initial state of the environment, and further calls after 3a complete should reflect the new environment. If the cert never changes, the race condition has not been solved.

Ceiu:
Currently looking into the failing spec tests

Ceiu:
I've changed the resolution on the last content update check to remove millisecond resolution, but I never saw the failures locally.

The fact that these new tests look to be failing sporadically gives me a lot of concern as to the value of this "fix". If it can't even resolve with a basic one-off "update + fetch" test with reliability, I'm not sure it's going to at all help the more usual case of "update + many fetch" situation that will be common in a real environment.

sonalidhome (approved):

sonalidhome (approved):
Ack. Changes look good to me.

PR #2943, opened 2021-02-22
On SCA mode, stop throwing an error during dry-run auto-attach

sonalidhome:
- Added more spec tests around SCA mode.

### DIFF
--- a/server/spec/autobind_disabled_for_owner_spec.rb
+++ b/server/spec/autobind_disabled_for_owner_spec.rb
@@ -154,4 +154,25 @@ describe 'Autobind Disabled On Owner' do
     job['state'].should == "FAILED"
     job['resultData'].should == "org.candlepin.async.JobExecutionException: Auto-attach is disabled for owner #{owner['key']} because of the content access mode setting."
   end
+
+  it "dry-run-auto-attach returns 200 when on SCA mode" do
+    skip("candlepin running in standalone mode") if not is_hosted?
+
+    owner = create_owner(random_string("test_owner"), nil, {
+        'contentAccessModeList' => 'org_environment,entitlement',
+        'contentAccessMode' => "org_environment"
+    })
+
+    owner = @cp.get_owner(owner['key'])
+    owner['autobindDisabled'] = true
+    @cp.update_owner(owner['key'], owner)
+
+    cp_user = user_client(owner, random_string("testing-user"))
+    consumer = cp_user.register("foofy_test", :system, nil,
+                                {'cpu.cpu_socket(s)' => '8'}, nil, owner['key'], [], [])
+    consumer_cp = Candlepin.new(nil, nil, consumer.idCert.cert, consumer.idCert['key'])
+    pools = consumer_cp.autobind_dryrun(consumer.uuid)
+    expect(pools.length).to eq(0)
+
+  end
 end
--- a/server/spec/consumer_resource_spec.rb
+++ b/server/spec/consumer_resource_spec.rb
@@ -1438,6 +1438,12 @@ describe 'Consumer Resource' do
     expected_exceptions.each { |e| e.should be_an(RestClient::Gone) | be_an(RestClient::ResourceNotFound) }
   end
 
+  it "should not allow to content access certificate body when not SCA mode" do
+    expect {
+      @consumer1.get_content_access_body()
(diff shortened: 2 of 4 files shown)


Ceiu (changes requested):
This doesn't really move the bar substantially as far as testing in hosted vs standalone modes. What I would expect here is to make the "before(:each)" block be hosted-aware and change how/where it creates the necessary data accordingly.

For individual tests which simply do not work because they require an endpoint that is disallowed in hosted mode, those tests should then be individually skipped based on the mode. However, given the current changes toward SCA mode everywhere, I would expect every operation to be testable in some capacity regardless of mode.

sonalidhome:
Updated the PR. Added a few more tests around SCA mode in hosted.

server/spec/content_access_spec.rb
```diff
@@ -883,4 +883,25 @@ def regenerate_cert_test(&updater)
...
+    # System purpose status
+    status = @cp.get_purpose_compliance(consumer['uuid'])
+    status['status'].should == 'disabled'
+
+    # compliance status
+    status = @cp.get_compliance(consumer['uuid'])
+    status['status'].should == 'disabled'
+  end
+
+  it "should not allow auto-attach for consumer with activation keys" do
```
nikosmoum, line 899: This test is wrong, so let's remove it. The scenario it describes is opposite of what we want, meaning registration should _never_ fail in SCA mode, regardless of if activation keys are used, or if they have auto-attach set to true or not.
The reason the test passes is unrelated: it seems to be that you accidentally set the value 'true' as the owner key on the register() method.

server/spec/content_access_spec.rb
```diff
@@ -883,4 +883,25 @@ def regenerate_cert_test(&updater)
     end
   end
 
+  it 'should be disabled for owner in SCA mode' do
```
nikosmoum, line 886: This test was missing, so let's keep it :+1:

server/spec/consumer_resource_spec.rb
```diff
@@ -1438,6 +1438,12 @@ def verify_installed_pids(consumer, product_ids)
     expected_exceptions.each { |e| e.should be_an(RestClient::Gone) | be_an(RestClient::ResourceNotFound) }
   end
 
+  it "should not allow to content access certificate body when not SCA mode" do
```
nikosmoum, line 1441: :+1:

server/spec/autobind_disabled_for_owner_spec.rb
```diff
@@ -154,4 +154,26 @@
     job['state'].should == "FAILED"
     job['resultData'].should == "org.candlepin.async.JobExecutionException: Auto-attach is disabled for owner #{owner['key']} because of the content access mode setting."
   end
+
+  it "should not dryrun when owner is in SCA mode" do
```
nikosmoum, line 158: The premise of this test is also wrong, and it actually exposes a gap we missed.
When we did the changes to no-op  auto-attach and return status 200 when in SCA mode (https://github.com/candlepin/candlepin/pull/2947), we should have _also_ applied that change to the dry-run auto-attach. Let's apply that change now, and change this test to assert the opposite: that dry-run-auto-attach returns 200 when on SCA mode

nikosmoum (changes requested):
1) Let's not make SCA mode testable in standalone mode in this commit, since we're going to change that anyway. We'll do the necessary changes as part of that change.
2) A couple of changes are needed.
3) Let's change the commit message to reflect the behavioural change on dry-run auto-attach instead since we're no longer doing what the original task was. Something like this should do:
```
On SCA mode, stop throwing an error during dry-run auto-attach

- Added more spec tests around SCA mode
```

server/spec/content_access_spec.rb
sonalidhome: Done.

server/spec/autobind_disabled_for_owner_spec.rb
sonalidhome: Updated.

server/spec/content_access_spec.rb
```diff
@@ -883,4 +883,25 @@ def regenerate_cert_test(&updater)
     end
   end
```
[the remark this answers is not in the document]
sonalidhome: ack.

PR #2945, opened 2021-02-24
ENT-3446: Make Jenkins run cloud auth spec

sonalidhome:
- Added a `generate_custom_yaml.py` to generate a 
  custom YAML file to override the `candlepin.auth.cloud.enable` property
  in `candlepin.conf`.
   This is executed only on Jenkins for hosted mode.

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -195,12 +195,13 @@ OPTIONS:
   -j <version>      use a specific Java version instead of the auto-detected default
   -v                enable verbose/debug output
   -i                internationalization - validate translation files
+  -k                run cloud registration rspec test suite (only for "hosted" mode)
   -a <arguments>    extra arguments to pass to the Candlepin deploy script (implies -d)
 HELP
 }
 
 ARGV=("$@")
-while getopts ":dtRqrHulsb:c:a:vj:i" opt; do
+while getopts ":dtRqrHulsb:c:a:vj:i:k" opt; do
     case $opt in
         d  ) DEPLOY="1";;
         t  )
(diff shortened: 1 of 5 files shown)


jenkins/Jenkinsfile
```diff
@@ -78,6 +78,7 @@ pipeline {
                         sh 'sudo chown -R jenkins:jenkins $WORKSPACE'
                         checkout scm
                         sh 'docker login -u "$CANDLEPIN_QUAY_BOT_USR" -p "$CANDLEPIN_QUAY_BOT_PSW" quay.io'
+                        sh 'python jenkins/generate_custom_yaml.py'
```
nikosmoum, line 81: We should try to avoid adding any testing logic in the pipeline script. It should only contain infrastructure-related commands (docker login/logout, git checkout, etc.). There are multiple levels where this script could be run:
1) jenkins/rspec-tests.sh
2) docker/test
3) cp-test
I would strongly prefer to add this script in the candlepin-base image and run it from cp-test, since that is the closest to the actual tests running, and anyone that would download/build the candlepin-base image to run tests locally would have the cloud tests running there too (while if we only left this in the jenkins pipeline script, they wouldn't run)

nikosmoum (changes requested):
It works well, however the pipeline script isn't the ideal place for running this

docker/candlepin-base/cp-test
```diff
@@ -429,6 +431,14 @@ if [ "$RSPEC" == "1" ]; then
     echo "Running rspec tests..."
     CLEAN_CP=1
 
+    # To run the cloud auth rspec test suite.
+    # We generate custom.yaml to enable the
+    # cloud auth property.
+    # This is valid only for hosted mode.
+    if [ "$CLOUD_AUTH_RSPEC" == "1" ] && [ "$HOSTED" == "1" ]; then
+      python jenkins/generate_custom_yaml.py
```
nikosmoum, line 439: The script should probably change location; The `jenkins` folder isn't appropriate since this is not a jenkins-specific script. `candlepin-base` or `server/bin`  is better

PR #2946, opened 2021-02-25
ENT-2360: Remove Qpid from Candlepin

wolfdale:
Changes -
- Removed ```Qpid``` dependency from POM, gradle & gemfiles
- Updated ```broker.xml``` & ```logback.xml```
- Removed qpid spec tests.
- Removed Qpid related code & unit tests.
- Kept the Qpid configuration scripts as it is, since we are using QPID for older candlepin branches.


Januson:
Changes look good. The rest of the pieces can be removed when the branches requiring qpid are no longer required. But this can be quite some time.

I think we could remove it now but it would require keeping a branch of Ansible role with qpid setup and refer to it in the requirements.yml of the older Candlepin branches.

@Ceiu @nikosmoum @sonalidhome What do you think?

wolfdale:
@Januson One option could be to have multiple vagrant files for each major Candlepin branch (something like they have in satellite-deploy). In that case we could use ```git tag``` on candlepin-ansible-roles & checkout particular tag depending on candlepin branch. This could also resolve Java version switching in between candlepin branches that @Ceiu was mentioning.
This will make maintaining older candlepin branches easy.

nikosmoum:
@wolfdale @Januson I think it's a good idea and we should do it, but perhaps not as part of this same task

server/bin/qpid/README.md
```diff
@@ -1,3 +1,5 @@
+[Deprecated since Candlepin-4.0.0]
+
```
nikosmoum, line 2: No point in adding this here; we can just delete the whole file (the whole `server/bin/qpid/` directory actually). It's still going to be available in older branches.

nikosmoum: The `server/bin/qpid_watch.py` script also needs removing

vagrant/README.md
```diff
@@ -66,7 +66,7 @@ tasks that are dependant on the existance of the variable 'my_cool_var'.
    The default port is 35675.
 1. Run `vagrant up`
 
-## Testing QPID
+## Testing QPID [Deprecated since Candlepin v4.0.0]
```
nikosmoum, line 69: Same here: no point keeping this. Just remove the section entirely

server/bin/deploy
```diff
@@ -336,34 +334,6 @@ else
     HOSTEDTEST=""
 fi
 
-# Delete the allmsg Qpid queue. The error messages produced during
-# deletion are silenced
-# This deletion is attempted regardless of whether Qpid is enabled. We do that
-# to prevent the queue to grow indefinitely
-QPID_KEYS_DIR=$PROJECT_DIR/../server/bin/qpid/keys
```
nikosmoum, line 343: We should also remove the creation of the `/etc/candlepin/certs/amqp` directory from the spect.tmpl file.
Specifically these lines seem to be irrelevant now:
```
%{__install} -d -m 755 %{buildroot}/%{_sysconfdir}/%{name}/certs/amqp/

and

## amqp dir, keystore & truststore used by qpid JMS client
%ghost %attr(644, tomcat, tomcat) %{_sysconfdir}/%{name}/certs/amqp/keystore
%ghost %attr(644, tomcat, tomcat) %{_sysconfdir}/%{name}/certs/amqp/truststore
```

nikosmoum (changes requested):
A few more things to remove; otherwise good

server/pom.xml
```diff
@@ -336,16 +336,6 @@
...
-      <artifactId>qpid-common</artifactId>
-      <scope>compile</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.qpid</groupId>
-      <artifactId>qpid-client</artifactId>
-      <scope>compile</scope>
-    </dependency>
     <dependency>
       <groupId>geronimo-spec</groupId>
```
nikosmoum, line 340: This dependency should have also been removed from here. Running `./gradlew pom` is removing it when I test your branch locally

vagrant/README.md
wolfdale: Removed.

server/bin/deploy
wolfdale: Removed.

server/pom.xml
wolfdale: Removed ```geronimo-spec``` dependency.

server/bin/qpid/README.md
wolfdale: The ```configure-qpid.sh``` script inside ```server/bin/qpid/``` is being used by ```docker/candlepin-base/cp-test```. Shouldn't  this affect our docker image builds ? 

I removed ```server/bin/qpid_watch.py``` script.

nikosmoum: You're right. in that case let's keep it!

PR #2947, opened 2021-03-03
1928072: When registering a system to RHSM (ENT-3549)

sonalidhome:
via Anaconda, the user is presented with
a spurious error message if their Organization
has simple content access (SCA) enabled.

 - Updated the code to ignore the exception thrown
   when owner has content mode settings enabled.

### DIFF
--- a/server/spec/autobind_disabled_for_owner_spec.rb
+++ b/server/spec/autobind_disabled_for_owner_spec.rb
@@ -31,7 +31,7 @@ describe 'Autobind Disabled On Owner' do
     exception_thrown.should be true
   end
 
-  it 'autobind should fail when owner is in SCA mode' do
+  it 'autobind should not attach entitlements when owner is in SCA mode' do
     skip("candlepin running in standalone mode") if not is_hosted?
 
     exception_thrown = false
@@ -48,17 +48,10 @@ describe 'Autobind Disabled On Owner' do
       {'cpu.cpu_socket(s)' => '8'}, nil, owner['key'], [], [])
     consumer_cp = Candlepin.new(nil, nil, consumer.idCert.cert, consumer.idCert['key'])
 
-    begin
-      consumer_cp.consume_product()
-    rescue RestClient::BadRequest => e
-      exception_thrown = true
-      ex_message = "Ignoring request to auto-attach. " +
-        "It is disabled for org \"#{owner['key']}\" because of the content access mode setting."
-      data = JSON.parse(e.response)
-      expect(data['displayMessage']).to eq(ex_message)
-    end
+    consumer_cp.consume_product()
+    entitlements = consumer_cp.list_entitlements()
+    expect(entitlements.length).to eq(0)
 
-    expect(exception_thrown).to eq(true)
   end
 
   it 'autobind fails when hypervisor autobind disabled on owner' do
--- a/server/spec/healing_spec.rb
+++ b/server/spec/healing_spec.rb
@@ -173,34 +173,4 @@ describe 'Healing' do
     exception_thrown.should be true
   end
 
-  it 'healing should fail when owner is in SCA mode' do
-    skip("candlepin running in standalone mode") if not is_hosted?
-
-    owner = create_owner(random_string("test_owner"), nil, {
-      'contentAccessModeList' => 'org_environment,entitlement',
-      'contentAccessMode' => "org_environment"
-    })
-    owner = @cp.get_owner(owner['key'])
-
-    expect(owner).to_not be_nil
-
-    cp_user = user_client(owner, random_string("testing-user"))
-    consumer = cp_user.register("foofy_test", :system, nil,
-      {'cpu.cpu_socket(s)' => '8'}, nil, owner['key'], [], [])
-    consumer_cp = Candlepin.new(nil, nil, consumer.idCert.cert, consumer.idCert['key'])
-    exception_thrown = false
-
-    begin
-      consumer_cp.consume_product()
-    rescue RestClient::BadRequest => e
-      exception_thrown = true
-      ex_message = "Ignoring request to auto-attach. " +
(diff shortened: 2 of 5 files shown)


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2177,9 +2177,10 @@ public Response bind(
             }
             catch (AutobindDisabledForOwnerException e) {
                 if (owner.isUsingSimpleContentAccess()) {
-                    throw new BadRequestException(i18n.tr("Ignoring request to auto-attach. " +
+                    log.debug("Ignoring request to auto-attach. " +
                         "It is disabled for org \"{0}\" because of the content access mode setting."
-                        , owner.getKey()));
+                        , owner.getKey());
+                    return null;
```
Januson, line 2183: Instead of just returning null we should build a Response so that it is clear what we are trying to accomplish.

Januson (changes requested):
Changes look good. There is just one change I would like to see to make the intent of that code clearer.

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -235,10 +235,15 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
                     "Auto-attach is disabled for owner \"{0}\"{1}.",
                     owner.getKey(), hypMessage));
             }
-            else {
+            else if (owner.isAutobindDisabled()) {
                 throw new AutobindDisabledForOwnerException(i18n.tr(
                     "Auto-attach is disabled for owner \"{0}\"{1}.",
-                    owner.getKey(), caMessage));
+                    owner.getKey()));
```
nikosmoum, line 241: The `caMessage` variable was accidentally removed here

server/src/main/java/org/candlepin/resource/ConsumerResource.java
nikosmoum: Agreed. A response with status 200 set on it

server/src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -235,10 +235,15 @@ public void adjustEntitlementQuantity(Consumer consumer, Entitlement ent,
...
+            }
+            else {
+                log.debug("Auto-attach is disabled for owner \"{0}\"{1}.",
+                    owner.getKey(), caMessage);
+                return null;
```
nikosmoum, line 246: Looking at the code paths where this method is used in, returning null might be dangerous.
One example i can see is the `EntitleByProductsJob` job that could throw an NPE when it will try to call `.size()` on this list. An empty arraylist is safer.

nikosmoum (changes requested):
A couple of issues to look at, but otherwise good

nikosmoum:
Another note: Lets make the commit message more to-the-point, without mentioning Anaconda. This is a behaviour change in Candlepin that is going to affect all types of clients (not just Anaconda). Something like should do:
```
On SCA mode, stop throwing an error during auto-attach

- No-op and return HTTP status 200 when auto-attaching on SCA mode
```

server/src/main/java/org/candlepin/resource/ConsumerResource.java
sonalidhome: Done.

server/src/main/java/org/candlepin/controller/Entitler.java
sonalidhome: This was removed because in this `if` condition the `caMessage` is empty.

server/src/main/java/org/candlepin/controller/Entitler.java
sonalidhome: Done.

server/spec/autobind_disabled_for_owner_spec.rb
```diff
@@ -48,17 +48,10 @@
...
-    begin
-      consumer_cp.consume_product()
-    rescue RestClient::BadRequest => e
-      exception_thrown = true
-      ex_message = "Ignoring request to auto-attach. " +
-        "It is disabled for org \"#{owner['key']}\" because of the content access mode setting."
-      data = JSON.parse(e.response)
-      expect(data['displayMessage']).to eq(ex_message)
-    end
+    consumer_cp.consume_product()
```
nikosmoum, line 51: Let's change the name of this spec test to better reflect the new behaviour. It should say something like `autobind should not attach entitlements when owner is in SCA mode`

server/spec/healing_spec.rb
```diff
@@ -188,19 +188,11 @@
...
-    begin
-      consumer_cp.consume_product()
-    rescue RestClient::BadRequest => e
-      exception_thrown = true
-      ex_message = "Ignoring request to auto-attach. " +
-        "It is disabled for org \"#{owner['key']}\" because of the content access mode setting."
-      data = JSON.parse(e.response)
-      expect(data['displayMessage']).to eq(ex_message)
-    end
+    consumer_cp.consume_product()
```
nikosmoum, line 192: This spec test is identical with the other one in `autobind_disabled_for_owner_spec.rb `. Let's delete it.

nikosmoum (changes requested):
A couple minor issues I missed on the spec tests

server/spec/autobind_disabled_for_owner_spec.rb
sonalidhome: Updated the spec name.

server/spec/healing_spec.rb
sonalidhome: Done.

nikosmoum:
Ack on the changes, but the commit must have the bz id in front of it

nikosmoum:
@sonalidhome Let's port this to 3.1-HOTFIX too

PR #2949, opened 2021-03-03
[M] Rewrote test data importer (ENT-3593)

Ceiu:
- Completely rewrote the test data importer in python, replacing
  the aging ruby version
- Fixed an incorrect endpoint path in HostedTestResource for
  adding a batch of content to a product


Ceiu:
For *actual* python developers: I am sorry. Feel free to call out better "pythonic" ways of doing things. If it's not horribly inefficient, or clown-shoes syntax, I'll likely switch over to it.

Testing instructions:
- Deploy as you would normally with test data, or invoke the script directly.
- Python2/3 didn't seem to make a difference for me, so hopefully your mileage does not vary here.
- There are now CLI options which may be worth exploring. I.e. you can now target a remote CP installation with this.

jirihnidek (commented):
@Ceiu I would like to do lot of changes. It will be easier to update this PR directly. Is it OK?

Ceiu:
@jirihnidek Depends on the scope of the changes. Generally I would prefer comments so I can make the changes myself; especially in a case like this where I need more practice/experience with the language.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+    host = 'localhost'
+    port = 8443
+    username = 'admin'
+    password = 'admin'
+
+    status_cache = None
+    hosted_test = None
+
+
+    def __init__(self, host, port, username, password):
```
jirihnidek, line 81: Please use default values like `host='localhost', port=8443, ...` instead of class attributes.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+        self.host = host
+        self.port = port
+        self.username = username
+        self.password = password
+
+        self.determine_mode()
+
+    def build_url(self, endpoint):
```
jirihnidek, line 89: Class methods do not have doc strings.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+    def post(self, endpoint, data=None, params=None, headers=None, content_type='application/json', accept_type='application/json'):
+        return self.request('POST', endpoint, data, params, headers, content_type, accept_type)
+
+    def put(self, endpoint, data=None, params=None, headers=None, content_type='application/json', accept_type='application/json'):
+        return self.request('PUT', endpoint, data, params, headers, content_type, accept_type)
+
+    def delete(self, endpoint, data=None, params=None, headers=None, content_type='application/json', accept_type='application/json'):
+        return self.request('DELETE', endpoint, data, params, headers, content_type, accept_type)
+
+    def request(self, req_type, endpoint, data=None, params=None, headers=None, content_type='application/json', accept_type='application/json'):
```
jirihnidek, line 108: Too long line.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+        if params is None: params = {}
+        if headers is None: headers = {}
+
+        # Add our content-type and accept type if they weren't explicitly provided in the headers already
+        if not ci_in(headers, 'Content-Type'): headers['Content-Type'] = content_type
```
jirihnidek, line 113: PEP8 error: E701

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+        if not ci_in(headers, 'Accept'): headers['Accept'] = accept_type
```
jirihnidek, line 114: PEP8 error: E701

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+
+        log.log(LOGLVL_TRACE, "Sending request: %s %s (params: %s, headers: %s, data: %s)" % (req_type, endpoint, params, headers, data))
```
jirihnidek, line 116: Too long line.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+
+        # correct provided products and derived product
+        if 'provided_products' in product_data:
+            provided_products = [ { 'id': value } for value in product_data['provided_products'] ]
+
+        if 'derived_product_id' in product_data:
+            derived_product = { 'id': product_data['derived_product_id'] }
+
+        # Add branding to some of the products
+        # TODO: FIXME: couldn't we just do this in the test data instead?
```
jirihnidek, line 216: I agree.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+    username = 'admin'
+    password = 'admin'
+
+    status_cache = None
+    hosted_test = None
+
+
+    def __init__(self, host, port, username, password):
+        self.host = host
+        self.port = port
```
jirihnidek, line 83: What about adding here `self.prefix` (default value `candlepin`)

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+                if 'content' in data:
+                    content.extend([(owner_key, value) for value in data['content']])
+
+                if 'content' in owner:
+                    content.extend([(owner_key, value) for value in owner['content']])
+
+    return content
+
+# Gather and order products in a way that ensures linear creation is safe
+def gather_products(data, hosted_test):
```
jirihnidek, line 365: The method should be refactored. It is too long.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+    def order_products(owner_key, product_map):
```
jirihnidek, line 366: Shadows name `owner_key` from outer scope. Same for `product_map`.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+        ordered = []
+        processed_pids = []
+
+        def traverse_product(product):
```
jirihnidek, line 370: Is it really necessary to have inner method here?

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+        # to Candlepin
+        for content_ref in product_data['content']:
+            content_map[content_ref[0]] = content_ref[1]
+
+    if content_map:
+        cp.add_content_to_product(owner_key, product_data['id'], content_map)
+
+
+# Process a given JSON file, creating the test data it defines
+def process_file(cp, options, filename):
```
jirihnidek, line 446: This function should be refactored to several functions like `create_owners()`, `create_users()`, etc.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+    owners = {}
+    activation_keys = []
+    engineering_products = {}
+
+    log.info("Importing data from file: %s" % (filename))
+
+
+    # Create owners
+    if 'owners' in data:
+        print('')
```
jirihnidek, line 461: Mixing `print()` and `long.<method>` is not recommended.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+                with open(cert_filename, 'w+') as file:
+                    log.info("  Writing certificate: %s" % (cert_filename))
+                    file.write(product_cert['cert'])
+            except requests.exceptions.HTTPError as e:
+                if e.response.status_code == 404:
+                    log.warn("  Skipping certificate for unused product: %s" % (pid))
+                else:
+                    raise e
+
+        print('')
```
jirihnidek, line 643: This `print()` and `log.info()` should be shifted left.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+        log.info("Creating %d activation keys..." % (len(activation_keys)))
+        for owner_key, key_name, pool in activation_keys:
+            log.info("  Activation key: %s [owner: %s, pool: %s]" % (key_name, owner_key, pool['id'] if pool is not None else 'n/a'))
+
+            key = cp.create_activation_key(owner_key, {'name': key_name})
+            if pool is not None:
+                cp.add_pool_to_activation_key(key['id'], pool['id'])
+
+
+    # Fetch entitlements/certs for engineering products
```
jirihnidek, line 618: IMHO these certificates are not entitlement certificates, but these certificates are `productid` certificates.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+import json
+import os
+import random
+import re
+import requests
+import string
+import sys
+import time
+
+from optparse import OptionParser
```
jirihnidek, line 19: The `optparse` module is considered as obsolete. Please use `argparse`.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+    def create_role(self, role_name, permissions):
+        role = {
+            'name': role_name,
+            'permissions': permissions
+        }
+
+        return self.post('roles', role).json()
+
+    def add_role_to_user(self, username, role_name):
+        return self.post("roles/%s/users/%s" % (role_name, username)).json()
```
jirihnidek, line 182: Using `"%s" % role_name` is considered as obsoleted. Please use "{var}".format(var=variable).

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+            'branding': branding,
+            'providedProducts': provided_products,
+            'derivedProduct': derived_product
+        }
+
+        # Determine URL and submit
+        if self.hosted_test:
+            url = 'hostedtest/products'
+        else:
+            url = "owners/%s/products" % (owner_key)
```
jirihnidek, line 240: Redundant parentheses.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+        log.debug("No test data files provided, defaulting to 'test_data.json'")
+
+        filename = 'test_data.json'
+        if sys.path[0]:
+            filename = "%s/%s" % (sys.path[0], filename)
+
+        args = [filename]
+
+    return (options, args)
+
```
jirihnidek, line 688: Too many white space.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+    return False
+
+# Merges two dictionaries, converting the keys in the source dictionary from snake case to camel case
+# where appropriate.
+def safe_merge(source, dest, include=None, exclude=None):
+    def handle_match(match):
+        return "%s%s" % (match.group(1), match.group(2).upper())
+
+    for key, value in source.items():
+        key = re.sub('(\w)_(\w)', handle_match, key)
```
jirihnidek, line 57: PEP8: W608 invalid escape sequence.

jirihnidek (changes requested):
The Python code requires some changes. Please use flake8, pylint, PyCharm or any other tool checking style of the code.

Entitlement certificates contains definition, but the content is still not generated correctly. When you deploy candlepin with option `-r`, then generated repositories expected unique product ID in the path. When I enable `fake-content` repository:

```
[root@thinkpad-t580 ~]# subscription-manager repos --list-enabled 
+----------------------------------------------------------+
    Available Repositories in /etc/yum.repos.d/redhat.repo
+----------------------------------------------------------+
Repo ID:   fake-content
Repo Name: fake-content
Repo URL:  http://svice:8080/path/to/fake-content
Enabled:   1
```

and I want to install some RPM package from this repository, then installation is terminated with following error:

```
[root@thinkpad-t580 ~]# dnf install insights-client
Updating Subscription Management repositories.
fake-content                                                                                                                                   230 kB/s | 1.0 kB     00:00    
Errors during downloading metadata for repository 'fake-content':
  - Status code: 404 for http://svice:8080/path/to/fake-content/repodata/repomd.xml (IP: 192.168.122.50)
Error: Failed to download metadata for repo 'fake-content': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
Ignoring repositories: fake-content
Last metadata expiration check: 0:55:09 ago on Mon Mar  8 15:37:53 2021.
No match for argument: insights-client
Error: Unable to find a match: insights-client
```

The content should have following path `http://svice:8080/path/to/fake-content/38072` and not only `http://svice:8080/path/to/fake-content`.

```
[vagrant@candlepin vagrant]$ ls /var/lib/tomcat/webapps/ROOT/path/to/fake-content/38072/repodata/repomd.xml 
/var/lib/tomcat/webapps/ROOT/path/to/fake-content/38072/repodata/repomd.xml
```

server/bin/test_data_importer.py
Ceiu: I assume that long is supposed to be "log" -- in which case I generally agree. However, I want blank line logical separators for output, and just doing a log.<anything> outputs a whole bunch of stuff. The alternative is an extra line break at the end of a log line, but then I have to know when the last line of a given loop will be.

server/bin/test_data_importer.py
Ceiu: Fixed

server/bin/test_data_importer.py
Ceiu: Correct. The comment is wrong here.

server/bin/test_data_importer.py
Ceiu: Does python want escaped backslashes for regular expressions (like Java)?

server/bin/test_data_importer.py
Ceiu: The method has no meaning outside of the containing method, it closes on some of the containing method's values (processed_pids), and is recursive in nature. Yes, this could be refactored to not require passing all of that around but that seems rather wasteful given closures are available.

server/bin/test_data_importer.py
Ceiu: Which is fine here -- that inner method is not intended to be using those values from the outer scope. owner_key would probably be fine, but the product map gets rebuilt per-org in standalone mode.

server/bin/test_data_importer.py
Ceiu: Done

server/bin/test_data_importer.py
Ceiu: Will get fixed along with the other conversions of printf-style formatting.

server/bin/test_data_importer.py
Ceiu: Fixed

server/bin/test_data_importer.py
Ceiu: Fixed

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+    return dest
+
+def random_id(prefix, suffix_length=8):
+    suffix = ''.join(random.choice(string.digits) for i in range(suffix_length))
+    return "%s-%s" % (prefix, suffix)
+
+
+
+class Candlepin:
+    # Server configuration
```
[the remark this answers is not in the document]
Ceiu: Holdover from initial testing. Fixed.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+logging.getLogger("requests").setLevel(logging.WARNING)
+logging.getLogger("urllib3").setLevel(logging.WARNING)
+
+log = logging.getLogger('test_data_importer')
+
+
+# TODO: There's likely a better python way of doing this that I don't know, but all the ways that I've
+# learned make shallow copies of the entire dictionary to do the check which seems rather wasteful to me,
+# so we go old school here.
+def ci_in(dict, search):
```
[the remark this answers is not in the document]
Ceiu: Partially fixed. Name is unchanged, variable updated to not conflict with intended variable type.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+
+from optparse import OptionParser
+
+
+
+# Disable HTTP warnings that'll be kicked out by urllib (used internally by requests)
+requests.packages.urllib3.disable_warnings()
+
+# Setup logging
+LOGLVL_TRACE = 5
```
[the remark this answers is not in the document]
Ceiu: How so? The intent here is that it's a constant (upper snake-case), for the log level of trace which isn't defined natively by the logger.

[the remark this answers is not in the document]
Ceiu: Also this follows similar naming conventions elsewhere in CP

server/bin/test_data_importer.py
Ceiu: Done.

server/bin/test_data_importer.py
Ceiu: Done

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+
+class Candlepin:
+    # Server configuration
+    host = 'localhost'
+    port = 8443
+    username = 'admin'
+    password = 'admin'
+
+    status_cache = None
+    hosted_test = None
```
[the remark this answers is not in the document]
Ceiu: removed

server/bin/test_data_importer.py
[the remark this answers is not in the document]
jirihnidek: OK

server/bin/test_data_importer.py
jirihnidek: You can use `log.info('')`.

server/bin/test_data_importer.py
jirihnidek: Thanks. It would be better to use f-strings, but I think that we still want to test it on RHEL/CentOS 7, right? BTW: PyCharm can do this conversation for you (shortcut Alt-Enter).

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+            # Clean up job status
+            # self.delete('jobs', {'id': job_status['id']})
+
+        return job_status
+
+
+
+# Gather and order content in a way that ensures linear creation is safe
+def gather_content(data, hosted_test):
+    content = []
```
jirihnidek, line 337: One more question. Why is list of tuples used here? You create map (dictionary) of contents later, where key for content is content ID. What about duplicity in generated list? I would expect more comments and explanation in this function.

server/bin/test_data_importer.py
jirihnidek: Yes, something like this: '(\\w)_(\\w)' removes the warning.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+
+        return self.status_cache
+
+    def determine_mode(self):
+        if self.hosted_test is None:
+            self.hosted_test = not self.status()['standalone']
+
+            if self.hosted_test:
+                try:
+                    self.get('hostedtest/alive', accept_type='text/plain')
```
jirihnidek, line 140: BTW: I don't see `hostedtest` in https://www.candlepinproject.org/swagger/?url=candlepin/swagger-3.1.16.json#/ What is it? Is it used in production?

server/bin/test_data_importer.py
Ceiu: Done

server/bin/test_data_importer.py
Ceiu: ```log.info('')``` does not result in a blank line, and makes it hard to see the logical break in logging output. I.e.

```
2021-03-16 16:23:51,107 INFO    test_data_importer -- test
2021-03-16 16:23:51,107 INFO    test_data_importer -- 
2021-03-16 16:23:51,107 INFO    test_data_importer -- Connecting to Candlepin @ localhost:8443
```

instead of the desired
```
2021-03-16 16:23:51,107 INFO    test_data_importer -- test

2021-03-16 16:23:51,107 INFO    test_data_importer -- Connecting to Candlepin @ localhost:8443
```

Is there another option we have here to use the logger without printing the usual logging prefix?

server/bin/test_data_importer.py
Ceiu: The tuples here and the map you noticed later have different sets of data and serve different purposes:

The output of gather_content is an array of tuples consisting of the content ID and the key of the owning organization (if applicable). In standalone/Sat mode, the content ID may be present multiple times, as each org gets their own copy of the content; whereas in hosted/portal mode, the content is created globally, and will not be assigned to an org. In both cases it's technically incorrect to have a given tuple of (content_id, owner_key) appear more than once, so I suppose we could use a set here to eliminate that potential risk (assuming set is py2 compatible); however, we still don't validate across files, so the test data still needs to be formatted properly to avoid the headache here.

The output of map_content_to_product (which is what I assume you're referencing) generates the mapping of content_id to a boolean indicating whether or not that content is enabled. This data comes from a different location (product['content']), and only overlaps with data['content'] and data['owner']['content'] in the content_id. This map is passed to Candlepin to indicate which content is mapped to a given product, and is built for each product that specifies content.

server/bin/test_data_importer.py
Ceiu: The hostedtest stuff is not deployed in the war and can only be added from the unshipped deploy script by specifying -H. It is a set of APIs that populate the data store behind some adapters we use for testing a simulated upstream to represent how hosted operates -- very similar in behavior and design to the cloud registration testing.

The previous test data injector was kind of doing this as well, but was hiding the hosted/standalone separation behind the deprecated ```create_pool_and_subscription```, which tried to do both at the same time to varying degrees of success.

server/bin/test_data_importer.py
Ceiu: nack

sub-100 lines is acceptable for something that's already abstracted out (and itself encapsulates an abstracted recursive op). The linter is wrong in this case.

server/bin/test_data_importer.py
Ceiu: fixed

server/bin/test_data_importer.py
Ceiu: fixed

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+
+    for key in dict:
+        if key.lower() == search:
+            return True
+
+    return False
+
+# Merges two dictionaries, converting the keys in the source dictionary from snake case to camel case
+# where appropriate.
+def safe_merge(source, dest, include=None, exclude=None):
```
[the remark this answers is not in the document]
Ceiu: fixed

Ceiu:
Converted the PR back to a draft to prevent merging before the squash.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,706 @@
...
+    def handle_match(match):
```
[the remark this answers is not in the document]
Ceiu: Some of these have been fixed

server/bin/test_data_importer.py
Ceiu: done

Ceiu:
@jirihnidek 
I've addressed the majority of the comments, though I believe I missed a few. Some of them are on the subjective side of things with regards to how much I want to respect the linter (I don't), but others I probably overlooked trying to fix everything in one commit. If you see something that hasn't been addressed, feel free to call it out again.

server/bin/test_data_importer.py
Ceiu: Fixed. I've changed it such that the logger now properly handles empty lines to output nothing.

server/bin/test_data_importer.py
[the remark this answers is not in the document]
Ceiu: I've removed this log level as the debug level hadn't been used yet, so it made more sense to convert the only TRACE output to debug.

server/bin/test_data_importer.py
jirihnidek: Good to know. Thanks. :+1:

jirihnidek:
The issue described in this comment: https://github.com/candlepin/candlepin/pull/2949#pullrequestreview-606370765 is not still addressed. In other words functionality introduced in this commit: https://github.com/candlepin/candlepin/commit/de1955fb2642d5ff2019a65042cd284f05f24903 was lost and it wasn't implemented in new importer.

I will try to explain, why functionality implemented in this PR cannot work for testing repositories. In other words why one content cannot be shared between several products. Lets have two products called `prod_a` (represented by product certificate `a.pem`) and `prod_b` (represented by product certificate `b.pem`). Both products tries to share same content e.g. `never_enabled`. Content contains many attributes. One of the attribute is path to the repository. The content of this repository can looks like this:

```
[vagrant@candlepin vagrant]$ ls -lR /var/lib/tomcat/webapps/ROOT/foo/path/never_enabled/100000000000000/
/var/lib/tomcat/webapps/ROOT/foo/path/never_enabled/100000000000000/:
total 4
drwxr-xr-x 2 tomcat tomcat  107 Mar 18 13:39 RPMS
drwxr-xr-x 2 tomcat tomcat 4096 Mar 18 13:39 repodata

/var/lib/tomcat/webapps/ROOT/foo/path/never_enabled/100000000000000/RPMS:
total 12
-rw-r--r-- 1 tomcat tomcat 2724 Mar 18 13:39 slow-eagle-1-07.noarch.rpm
-rw-r--r-- 1 tomcat tomcat 2728 Mar 18 13:39 tricky-frog-1-0.noarch.rpm
-rw-r--r-- 1 tomcat tomcat 2724 Mar 18 13:39 white-lion-1-0.noarch.rpm

/var/lib/tomcat/webapps/ROOT/foo/path/never_enabled/100000000000000/repodata:
total 32
-rw-r--r-- 1 tomcat tomcat  823 Mar 18 13:39 06ab876ce26821bb45a6fa71a97a4184755d9e7fcfbcd02efe3dd2da8c5a6978-filelists.sqlite.bz2
-rw-r--r-- 1 tomcat tomcat  418 Mar 18 13:39 086cc511029ce3f96ab239c8fe2ed270e28db26acd4e4be3177b5b486bfedb44-other.xml.gz
-rw-r--r-- 1 tomcat tomcat  958 Mar 18 13:39 0dda56a2cf9a1fc46a4c7d08c81290fc3a42e191d7ee294a5c319caa8edeb245-other.sqlite.bz2
-rw-r--r-- 1 tomcat tomcat  846 Mar 18 13:39 13ae4a911a3ec6510f5af6ba648897c5542588689c6dd5bfb61271f79e6b94c6-primary.xml.gz
-rw-r--r-- 1 tomcat tomcat 2203 Mar 18 13:39 1a69f0076718b059c491e989567c58cbfcc53e90a719a01af56aacd18db2894c-primary.sqlite.bz2
-rw-r--r-- 1 tomcat tomcat  345 Mar 18 13:39 6b8cb101c2efc19a2da357765e9c50c7dcb779ec7dc7406009926bf94586672f-filelists.xml.gz
-rw-r--r-- 1 tomcat tomcat 1324 Mar 18 13:39 c9be92b2222293407c23a312b99449473cec1670b27a1365f83172d7b15706bb-productid.gz
-rw-r--r-- 1 tomcat tomcat 3522 Mar 18 13:39 repomd.xml
```

As you can see the repository contains directory `repodata` and this directory has to contain product certificate (`<hash>-productid.gz`). It can contain only one product certificate. This certificate could be `a.pem` or `b.pem`. It could not be some Schrodinger certificate. Thus one content in the fact cannot be shared with more products. Directory RPMS can be shared using some symbolic links, but `repodata` has to be unique for combination product-content.

Ceiu:
@jirihnidek I've updated the content generation steps to use the same type of workaround previously implemented in the ruby test data importer. Let me know if that resolves the repo issues for you.

Ceiu:
@jirihnidek Updated again to add the pid-cid pairing to the end of the content URL to ensure it is unique for the product.

Included also with this commit is some commented out lines which do the same for both the content name and label (as the old Ruby script did), but I'm trying to minimize the amount of data manipulation the script is doing. If repo utilization *still* fails, try uncommenting out the name/label updates on line 589 and 590 and see if that gets us going.

jirihnidek:
> @jirihnidek Updated again to add the pid-cid pairing to the end of the content URL to ensure it is unique for the product.
> 
> Included also with this commit is some commented out lines which do the same for both the content name and label (as the old Ruby script did), but I'm trying to minimize the amount of data manipulation the script is doing. If repo utilization _still_ fails, try uncommenting out the name/label updates on line 589 and 590 and see if that gets us going.

When I did following small change of `server/bin/create_test_repos.py`, then repositories are created according your changes and I was able to install RPM packages from testing repositories:

```diff
diff --git a/server/bin/create_test_repos.py b/server/bin/create_test_repos.py
index 3297eff..68c780f 100755
--- a/server/bin/create_test_repos.py
+++ b/server/bin/create_test_repos.py
@@ -167,7 +167,7 @@ def create_repo_definition(product, content):
         'name': product['id'] + '-' + content['name'],
         'product_id': product['id'],
         'type': content['type'],
-        'content_url': os.path.join(content['content_url'], product['id']),
+        'content_url': os.path.join(content['content_url'], str(product['id']) + '-' + str(content['id'])),
         'gpg_url': content.get('gpg_url', ""),
         'packages': content.get('packages', [])
     }
```

I will do proper review of the PR, but this is promising. :+1:

jirihnidek (changes requested):
I see very good progress in this PR :+1: , I have few requests.

Please include diff proposed in this comment https://github.com/candlepin/candlepin/pull/2949#issuecomment-814028808 to this PR.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,865 @@
...
+        for content_id, enabled in product_data['content']:
+            # Create a unique content specifically for this product-content pairing.
+            content = find_content(content_id).copy()
+
+            content_id = '{pid}-{cid}'.format(pid=product_data['id'], cid=content_id)
+
+            # Modify the base content data so that it is unique for this product
+            content['id'] = content_id
+            # content['name'] = '{name} ({uid})'.format(name=content['name'], uid=product_data['id'])
+            # content['label'] = '{label} ({uid})'.format(label=content['label'], uid=product_data['id'])
```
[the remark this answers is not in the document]
Ceiu: I see. In light of that, I'll use the "pid-cid" syntax for all three of them.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,865 @@
...
+        filename = 'test_data.json'
+        if sys.path[0]:
+            filename = '{script_home}/{filename}'.format(script_home=sys.path[0], filename=filename)
+
+        options.json_files = [filename]
+
+    return options
+
+
+# Execute script
```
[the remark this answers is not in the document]
Ceiu: I'll add this, but for clarity: I don't intend for this to be imported elsewhere, so that wasn't a concern when this started.

As far as the Candlepin class goes: it exists because we currently have nothing for that at present, and we have other initiatives that should eventually supersede the need to manually write one later (OpenAPI/spec first). The Ruby script got to piggyback on the existing spec tests having a generic-ish candlepin_api class to use for sending everything, but we don't yet have something equivalent in Python. As I don't want to write one here, I implemented the bare minimum needed for the test data creation, but in a way that if/when it gets replaced with a full thing later, we can delete the Candlepin class from this file and import the fully-featured version with minimal pain.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,865 @@
...
+            while job_status['state'] not in finished_states:
+                time.sleep(1)
+                job_status = self.get(job_status['statusPath']).json()
+
+            # Clean up job status
+            # self.delete('jobs', {'id': job_status['id']})
+
+        return job_status
+
+def create_owners(cp, data):
```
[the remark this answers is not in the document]
Ceiu: create_owners, create_users, etc. only exist because of the previous request to break up the process_file function. They serve no purpose other than reducing the length of that function to make the linter happy.

The goal of the Candlepin class is to provide the API. The methods mentioned above here provide business logic that itself uses the API.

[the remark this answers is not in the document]
jirihnidek: OK, np.

server/bin/test_data_importer.py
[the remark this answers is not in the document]
jirihnidek: Thanks, good to know.

server/bin/test_data_importer.py
```diff
@@ -0,0 +1,868 @@
...
+        # to Candlepin
+        for content_id, enabled in product_data['content']:
+            # Create a unique content specifically for this product-content pairing.
+            content = find_content(content_id).copy()
+            content_id = '{pid}-{cid}'.format(pid=product_data['id'], cid=content_id)
+
+            # Modify the base content data so that it is unique for this product
+            content['id'] = content_id
+            content['name'] = '{name} ({uid})'.format(name=content['name'], uid=product_data['id'])
+            content['label'] = '{label} ({uid})'.format(label=content['label'], uid=product_data['id'])
```
jirihnidek, line 588: Nitpick. Can you please modify this line to this:

```python
content['label'] = '{label}-{uid}'.format(label=content['label'], uid=product_data['id'])
```

Without this change generated `repo_id` looks strange, because white spaces and special characters in content label are transformed to '-' by subscription-manager, when redha.repo is generated.

```
[root@thinkpad-t580 subscription-manager]# subscription-manager repos --list-enabled 
+----------------------------------------------------------+
    Available Repositories in /etc/yum.repos.d/redhat.repo
+----------------------------------------------------------+
Repo ID:   fake-content--38072-   <======== Too many '-'
Repo Name: fake-content (38072)
Repo URL:  http://svice:8080/path/to/fake-content/38072-3902
Enabled:   1

[root@thinkpad-t580 subscription-manager]# rct cc /etc/pki/entitlement/370058437736336375.pem | grep "Content:" -A 10
Content:
	Type: yum
	Name: fake-content (38072)
	Label: fake-content (38072)
	Vendor: Red Hat
	URL: /path/to/fake-content/38072-3902
	GPG: file:///etc/pki/rpm-gpg/RPM-GPG-KEY-candlepin
	Enabled: True
	Expires: 
	Required Tags: 
	Arches: ALL
```

jirihnidek (changes requested):
Hi,
I have only one small request (`repo_id`). I tested it in standalone mode, hosted mode and SCA mode and it works as expected in all cases.

Thanks

server/bin/test_data_importer.py
jirihnidek: Dnf/Yum accepts only alphanumeric characters and '-' in repository ID.

Ceiu: Unfortunate. Alright, I'll remove the parens.

Ceiu:
Okay, label and name have been updated, and the branch has been squashed.

server/bin/test_data_importer.py
Ceiu: After a few minutes of looking at this, I now realize that "Repo ID" is not the content ID, but either the name or the label of the content. Bah.

jirihnidek (approved):
Hi Chris,
Thanks for your work on this PR. :+1: I'm merging it to master.

Jiri

PR #2956, opened 2021-03-05
[3.1] 1928072: On SCA mode, stop throwing an error during auto-attach (ENT-3549)

sonalidhome:
- No-op and return HTTP status 200 when auto-attaching on SCA mode


nikosmoum (changes requested):
Since this is going into 3.1, I've created a separate bz clone with target release set to 3.1: 1935595. Let's use this in the commit message instead

PR #2965, opened 2021-03-10
1886556: Fixed NPE originating from consumer create API when facts are null

wolfdale:
Issue - 
Consumer create API called with null facts, results in NPE for **all** consumer types (other than ```candlepin```). 

Additional Info - 
We don't calculate the compliance hashes for consumer(while creating consumers) with consumer type ```candlepin```. There is data corruption on stage ENV, where ```isManifest``` flag is set to ```false``` for consumer type ```candlepin``` resulting in forced calculation of hash leading to NPE. 

Changes - 
Defensively added a null check for consumer facts while calculating compliance hashes.

### DIFF
--- a/server/src/main/java/org/candlepin/policy/js/compliance/hash/HashableStringGenerators.java
+++ b/server/src/main/java/org/candlepin/policy/js/compliance/hash/HashableStringGenerators.java
@@ -211,7 +211,9 @@ public class HashableStringGenerators {
             }
 
             String generated = target.getUuid();
-            generated += generateFromCollection(target.getFacts().entrySet(), STRING_ENTRY);
+
+            generated += generateFromCollection(target.getFacts() != null ?
+                target.getFacts().entrySet() : null, STRING_ENTRY);
             generated += generateFromCollection(target.getInstalledProducts(), INSTALLED_PRODUCT);
             generated += generateFromCollection(target.getEntitlements(), ENTITLEMENT);
             return generated;
--- a/server/src/test/java/org/candlepin/policy/js/compliance/hash/ComplianceStatusHasherTest.java
+++ b/server/src/test/java/org/candlepin/policy/js/compliance/hash/ComplianceStatusHasherTest.java
@@ -15,10 +15,10 @@
 
 package org.candlepin.policy.js.compliance.hash;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import org.candlepin.model.Consumer;
 import org.candlepin.model.ConsumerInstalledProduct;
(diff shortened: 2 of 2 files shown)


server/src/main/java/org/candlepin/policy/js/compliance/hash/HashableStringGenerators.java
```diff
@@ -211,7 +211,15 @@ public String generate(Consumer target) {
             }
 
             String generated = target.getUuid();
-            generated += generateFromCollection(target.getFacts().entrySet(), STRING_ENTRY);
+
+            if (target.getFacts() != null) {
```
Januson, line 215: The ComplianceReasonGenerator has the same potential issue with generating from attributes. We should probably fix that as well.

wolfdale: I traced back the code. It seems highly unlikely that we might run into NPE as it create a new instance of ```ComplianceStatus``` which has a [constructor](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/policy/js/compliance/ComplianceStatus.java#L59) that initialize ```ComplianceReason```. Do you have any particular scenario where we could run into such situations ?

server/src/test/java/org/candlepin/policy/js/compliance/hash/ComplianceStatusHasherTest.java
```diff
@@ -47,7 +47,7 @@
     private String initialHash;
     private Owner owner;
 
-    @Before
+    @BeforeEach
```
Januson, line 50: +1 Please switch the Assert imports as well.

wolfdale:
@Januson @nikosmoum Do you think we should port this fix to ```3.X-HOTFIX``` branches ? Also as there is an issue of incorrect data on staging ENV, should we perform manual DB update ?

server/src/test/java/org/candlepin/policy/js/compliance/hash/ComplianceStatusHasherTest.java
wolfdale: Done.

wolfdale:
@nikosmoum @Januson Any thoughts ?

nikosmoum:
@wolfdale We don't have to port this back to any other branch, because facts are never null in the real world (this issue is exposed only when you craft data with curl, etc.). As for the data on stage, you should reach out to that team to ask why/how it got changed to that?

server/src/main/java/org/candlepin/policy/js/compliance/hash/HashableStringGenerators.java
```diff
@@ -211,7 +211,15 @@ public String generate(Consumer target) {
...
+                generated += generateFromCollection(target.getFacts().entrySet(), STRING_ENTRY);
+            }
+            else {
+                // Append null if consumer facts are null.
+                generated += "null";
```
nikosmoum, line 220: What is the purpose of adding this? Couldn't we just not append anything (skip the else alltogether)?

wolfdale: In [generateFromCollection()](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/policy/js/compliance/hash/HashableStringGenerators.java#L66-L69) method we append null if any collection is null. I skipped the method call & did it over here.

nikosmoum: Perhaps it would be better to reduce the confusion by doing this:
`generated += generateFromCollection(target.getFacts() != null ? target.getFacts().entrySet() : null, STRING_ENTRY);`
and letting the `generateFromCollection` do it's own handling

wolfdale: Done.

PR #2975, opened 2021-03-17
1935500: Unexpected repos are available after turning on Simple Content Access

Januson:
- Update content filtering to take consumer facts into account

### DIFF
--- a/server/spec/content_access_spec.rb
+++ b/server/spec/content_access_spec.rb
@@ -30,11 +30,11 @@ describe 'Content Access' do
 
     @content = @cp.create_content(
         @owner['key'], "cname", 'test-content', random_string("clabel"), "ctype", "cvendor",
-        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]]}, true)
+        {:content_url=> '/this/is/the/path',  :modified_products => [@modified_product["id"]], :arches => "x86_64"}, true)
 
     @content_id = @content['id']
 
-    @product = create_product('test-product', 'some product')
+    @product = create_product('test-product', 'some product', :attributes => {:arch => 'x86_64'})
     @cp.add_content_to_product(@owner['key'], @product['id'], @content_id)
     @pool = @cp.create_pool(@owner['key'], @product['id'], {:quantity => 10})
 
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -322,7 +321,8 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
         if (oeca == null || regenerate) {
             log.info("Generating new SCA payload for organization \"{}\"...", owner.getKey());
 
-            String contentJson = this.createPayloadAndSignature(owner, env);
+            byte[] payload = this.createContentAccessDataPayload(owner, env, consumer);
```
Ceiu, line 324: This will cause the content payload to be generated for the owner-env-consumer triplet, but the OwnerEnvContentAccess object is stored only per-owner-environment. I suspect this will lead to a similar problem where content is either missing or extraneous depending on which consumer triggers cert generation first.

server/spec/content_access_spec.rb
```diff
@@ -91,6 +91,36 @@
     }.to raise_exception(RestClient::BadRequest)
   end
 
+  it "should filter content with incorrect architecture" do
```
Ceiu, line 94: We either need to extend this test or add a new test to verify that when a consumer generates a cert based on their arch, it doesn't negatively impact consumers in the same org+env with different arches.

server/src/test/java/org/candlepin/model/OwnerContentCuratorTest.java
```diff
@@ -513,9 +513,9 @@ public void testUpdateOwnerContentReferences() {
         Content unmodified = this.createContent("c1", "c1", owner2);
         Content updated = this.createContent("c1", "c1");
 
-        assertTrue(original.getUuid() != updated.getUuid());
-        assertTrue(original.getUuid() != unmodified.getUuid());
-        assertTrue(updated.getUuid() != unmodified.getUuid());
+        assertNotSame(original.getUuid(), updated.getUuid());
```
Ceiu, line 516: These were actually incorrect entirely before, as the strings representing the UUIDs could never possibly be the same instance across those objects -- even before Content was made to be immutable at the Hibernate level.

While the updated assertion is more apt for how the test was written, it's still logically incorrect and should be using proper string comparing rather than instance checking. However, these tests (and this class as a whole) has nothing to do with the PR.

Ceiu (changes requested):
We still need to verify (and likely fix) that consumers with different arches within the same org+env don't get inappropriate content.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -532,44 +528,47 @@ private String createDN(Consumer consumer, Owner owner) {
...
         this.ownerContentCurator.getActiveContentByOwner(owner.getId())
-            .forEach((content, enabled) -> container.addContent(content, enabled));
+            .forEach(container::addContent);
 
+        Consumer emptyConsumer = new Consumer();
+        emptyConsumer.setOwner(owner);
         emptyConsumer.setEnvironment(environment);
-        emptyEnt.setPool(emptyPool);
-        emptyEnt.setConsumer(emptyConsumer);
+        // TODO Do we need any other data for content filtering?
```
nikosmoum, line 547: I am not aware of anything else we filter based on the consumer. @Ceiu do you?

Ceiu: The only thing I know of that would be fetched from the consumer object aside from arch would be: CPU sockets/cores, ram, storage bands, etc.; nothing that applies for filtering in SCA mode, I think.

nikosmoum: That's all included in the facts, so we should be good

nikosmoum:
I'm closing this, since it's getting stale/has conflicts, and we need to do more design work before moving on to a solution.

PR #2976, opened 2021-03-17
[M] Removed foreign key constraints from obsolete pool tables (ENT-3668)

Ceiu:
- Removed the foreign key constraints from cp2_pool_provided_products
  and cp2_pool_derprov_products, as these tables have been obsoleted
  and existing data in them could prevent pools or products from
  being deleted properly.


sonalidhome (approved):
ack.

PR #2979, opened 2021-03-22
ENT-2580: Updated virt-who report generator testing script

wolfdale:
Changes - 
Updated virt-who report generator testing script.

```"virtWhoType":"libvirt"``` & ```"hypervisor.type":"QEMU"``` properties are kept as constants.

Additional Info -
1) Async format
```
./gen_fake_report.py --hypervisors 1 --guests 1 --report 1 --format async
```
```
{
   "hypervisors":[
      {
         "hypervisorId":{
            "hypervisorId":"a7c9cf7c-67cb-4b68-9553-a279262b5e34"
         },
         "name":"thinkpad-LA738B",
         "guests":[
            {
               "guestId":"ec79c6ca-3146-47bb-8cb1-9fc7b9f11bbd",
               "state":2,
               "attributes":{
                  "virtWhoType":"libvirt",
                  "active":1
               }
            }
         ],
         "facts":{
            "cpu.cpu_socket(s)":3,
            "hypervisor.type":"QEMU",
            "dmi.system.uuid":"c7790a3a-43f7-41ca-9389-644915cc075d"
         }
      }
   ]
}
```
2) Sync format (Underlying API (POST /hypervisor) is now deprecated in master, but I believe it is still in use for older candlepin)
```
./gen_fake_report.py --hypervisors 1 --guests 1 --report 1 --format sync
```
```
{
   "e8a0d445-5312-4b75-85e4-a55413ae4c0b":[
      {
         "guestId":"7207952b-56bd-4453-9a64-7f2c5faca26c",
         "state":5,
         "attributes":{
            "virtWhoType":"libvirt",
            "active":0
         }
      }
   ]
}
```

### DIFF
--- a/server/bin/gen_fake_report.py
+++ b/server/bin/gen_fake_report.py
@@ -4,32 +4,57 @@ import json
 import argparse
 import random
 import sys
+import string
 
 
 class AbstractVirtWhoReportGenerator(object):
 
-    def __init__(self, num_hypervisors, num_guests):
+    def __init__(
+            self, num_hypervisors, num_guests, virt_type, hypervisor_type,
+            hypervisor_id_format):
         self.hypervisors = num_hypervisors
         self.guests = num_guests
+        self.virt_type = virt_type
+        self.hypervisor_type = hypervisor_type
+        self.hypervisor_id_format = hypervisor_id_format
 
     def generate(self):
-        # Returns a json representation of the generated report
+        """
+        Returns a json representation of the generated report
+        """
         return json.dumps(self._generate())
 
     def gen_hypervisor(self, num_guests, min_guests=0, max_guests=None):
-        # Subclasses should implement this method to return a hypervisor with
-        # The given number of guests.
-        # Min guests, max_guests should allow for variation in the amount of
-        # guests for the hypervisor.
-        # Should return a dictionary representation of the guest
+        """
+        Subclasses should implement this method to return a hypervisor with
+        The given number of guests. Min guests, max_guests should allow for
+        variation in the amount of guests for the hypervisor. Should return
+        a dictionary representation of the guest
+        """
         raise NotImplemented()
 
     def gen_guest(self):
-        # Generate a guest. Should return a dictionary representation of the
-        # guest
+        """
+        Generate a guest. Should return a dictionary representation of the
+        guest.
+        """
         return {
             "guestId": str(uuid.uuid4()),
-            "state": random.choice(range(1, 6))
+            "state": random.choice(range(1, 6)),
+            "attributes": {
+                "virtWhoType": self.virt_type,
+                "active": random.choice(range(0, 2))
+            }
+        }
+
+    def gen_facts(self, hypervisor_uuid=str(uuid.uuid4())):
+        """
+        Generate guests facts.
+        """
+        return {
+            "cpu.cpu_socket(s)": random.choice(range(1, 6)),
+            "hypervisor.type": self.hypervisor_type,
+            "dmi.system.uuid": hypervisor_uuid
         }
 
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: server/bin/gen_fake_report.py [hunk 1]
```diff
@@ -38,16 +58,19 @@ class VirtWhoReportGenerator(AbstractVirtWhoReportGenerator):
...
         return report
 
 
 class AsyncVirtWhoReportGenerator(AbstractVirtWhoReportGenerator):
 
     def gen_hypervisor(self, num_guests, min_guests=0, max_guests=None):
-        return {"uuid": str(uuid.uuid4()),
-                "guests": [self.gen_guest() for x in range(0, num_guests)]}
+        return {"hypervisorId": {"hypervisorId": str(uuid.uuid4())},
+                "name": "thinkpad-" + gen_random_string(),
```
Ceiu, line 71: This will function, but I'd like this to be more generic so that consumer names are indicative of what they are. Something like "hypervisor-" or "generated_hypervisor-", etc. would be better for sifting through test data and logs once it hits Candlepin.

Ceiu (changes requested):
Looks good for what it is, however the naming of the hypervisors could be more specific.

Additionally, are the virtWhoType and hypervisor.type fields things that we'll need to change from time to time? Should those be configurable via CLI options?

server/bin/gen_fake_report.py
wolfdale: Changed it to use ```generated_hypervisor``` as prefix.

wolfdale:
We could end up with wanting to test with different values of ```virtWhoType``` and ```hypervisor.type```. Made it configurable with CLI options (with defaults as libvirt & QEMU) of ```--v``` & ```--h```.

server/bin/gen_fake_report.py
```diff
@@ -38,16 +61,19 @@ class VirtWhoReportGenerator(AbstractVirtWhoReportGenerator):
...
+            ]
         return report
 
 
 class AsyncVirtWhoReportGenerator(AbstractVirtWhoReportGenerator):
 
     def gen_hypervisor(self, num_guests, min_guests=0, max_guests=None):
-        return {"uuid": str(uuid.uuid4()),
-                "guests": [self.gen_guest() for x in range(0, num_guests)]}
+        return {"hypervisorId": {"hypervisorId": str(uuid.uuid4())},
```
nikosmoum, line 73: The `hypervisorId` can be of 3 different formats (based on the config) of either: uuid, hostname, hwuuid. If the property is not set, it defaults to uuid. So I think by default we should be setting this to the same uuid that the `dmi.system.uuid` fact value is.

Btw, it could be a nice feature to have a hypervisorId format cli flag, to specify which of the 3 formats it should use, although not that important (so feel free to skip).

server/bin/gen_fake_report.py
```diff
@@ -38,16 +61,19 @@ class VirtWhoReportGenerator(AbstractVirtWhoReportGenerator):
...
+                "name": "generated_hypervisor-" + gen_random_string(),
+                "guests": [self.gen_guest() for x in range(0, num_guests)],
```
nikosmoum, line 75: I don't know why this field was originally called `guests`, but it's wrong. When posting the generated report to the server, it ends up reporting the hypervisor but with 0 guests, because the field is called `guestIds`, not `guests`

nikosmoum (changes requested):

server/bin/gen_fake_report.py
wolfdale: You're right, this was not working with ```guests```, changed it to ```guestIds```

server/bin/gen_fake_report.py
wolfdale: I was unaware of this, but I checked the hypervisor_check_in spec, and it has the same working. Made few more changes to accommodate both ```uuid``` & ```hostname``` as hypervisorID formats. 
- For ```uuid``` format ```hypervisorId``` & ```dmi.system.uuid``` fields will be same.
- For ```hostname``` format ```hypervisorId``` & ```name``` fields will be same 

The ```uuid``` format will remain as default.

server/bin/gen_fake_report.py
```diff
@@ -56,22 +96,47 @@ def _generate(self):
...
+    args.hypervisor_id_format = str.lower(args.hypervisor_id_format).strip()
 
     return args
 
 
+def gen_random_string():
+    """
+    Generates random alphanumeric strings
+    """
+    return ''.join(random.choices(
```
nikosmoum, line 136: I'm getting the following:
```
[vagrant@candlepin vagrant]$ ./server/bin/gen_fake_report.py --hypervisors 1 --guests 2 --reports 1 --format async
Traceback (most recent call last):
  File "./server/bin/gen_fake_report.py", line 161, in <module>
    report = generator.generate()
  File "./server/bin/gen_fake_report.py", line 25, in generate
    return json.dumps(self._generate())
  File "./server/bin/gen_fake_report.py", line 94, in _generate
    range(0, self.hypervisors)]}
  File "./server/bin/gen_fake_report.py", line 79, in gen_hypervisor
    "name": "generated_hypervisor-" + gen_random_string(),
  File "./server/bin/gen_fake_report.py", line 136, in gen_random_string
    return ''.join(random.choices(
AttributeError: 'module' object has no attribute 'choices'
```
I think the method name is `choice`, not `choices`

nikosmoum: It looks like it's because I'm using python 2.7.5 (which is the default in Centos 7) and this method was only added in 3.6.
Let's make sure the script is compatible in general for both versions

wolfdale: yes. It's because of Python 2.7, changes made to work on both python versions.

nikosmoum:
There are some infrastructure related jenkins failures. I'm merging this regadless, since it is not exercised by jenkins tests anyway

PR #2981, opened 2021-03-23
[M] Fixed an issue with the org_migrator's export order of derived products (ENT-3675)

Ceiu:
- Fixed an issue/bug in the org_migrator's export ordering of products
  with derived products in the new product model. Products will now
  be exported such that derived products will occur earlier in the
  list than products which reference them.

### DIFF
--- a/server/bin/org_migrator/org_migrator.py
+++ b/server/bin/org_migrator/org_migrator.py
@@ -9,6 +9,7 @@ from optparse import OptionParser
 import os
 import re
 import zipfile
+import getpass
 
 import cp_connectors as cp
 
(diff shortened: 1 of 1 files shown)


server/bin/org_migrator/org_migrator.py
```diff
@@ -338,7 +338,51 @@ def do_export(self):
...
+
+        products = { row[uuid_idx]: row for row in cursor }
+        cursor.close()
+
+        ordered = []
+        insertion_order = {}
+
+        def insert_row(row):
+            if row[uuid_idx] in insertion_order:
+                # Row already inserted; skip it
```
nikosmoum, line 359: It seems to me the only use of the `insertion_order` map is here, to check that we already processed a row, and we never actually check the values in it (the order number). If so, do we still need this map? Could this check simply be changed into if `row in ordered:`, and we ditch `insertion_order` ?

server/bin/org_migrator/org_migrator.py
```diff
@@ -338,7 +338,51 @@ def do_export(self):
...
+                return
+
+            if row[derived_uuid_idx] is not None:
+                parent_row = products[row[derived_uuid_idx]]
+
+                if parent_row is None:
```
nikosmoum, line 365: Would `child_row` not be a more appropriate name for this variable, since the derived product is child of a main product?

nikosmoum (changes requested):

nikosmoum:
Also, I am unsure of how to test this, since reproducing the issue doesn't seem easy

server/bin/org_migrator/org_migrator.py
Ceiu: We could. This was a decision made for speed purposes, as a key lookup is far faster than a linear search through the list for an entire row object. 

Looks like 479 is our current max product count in prod, so if the equality check on tuples works as expected, this may be unnecessary.

server/bin/org_migrator/org_migrator.py
Ceiu: yes, fixed

Ceiu:
@nikosmoum 
Testing this is rather difficult, as it is contingent on the order in which the DB returns columns. Outside of manipulating the data stored on the drive/memory of our DB, we're kinda contingent on it giving us good or bad output.

nikosmoum:
Ack. merging since jenkins tests are not affected by this script

PR #2984, opened 2021-03-24
ENT-3214: Test SCA in Standalone mode

nikosmoum:
- Add utility method that indirectly sets the content access
  mode of an owner in standalone mode through export/import
- Allow SCA-related tests to run in both hosted and
  standalone modes, and make them mode-aware where needed,
  such as when asserting content urls contain a prefix or not
- Added a couple of missing assertions such as testing
  the removal of content access mode of a manifest consumer
  when sending an empty string for that value on an update request

### DIFF
--- a/server/spec/content_access_spec.rb
+++ b/server/spec/content_access_spec.rb
@@ -12,12 +12,21 @@ describe 'Content Access' do
   include Unpack
 
   before(:each) do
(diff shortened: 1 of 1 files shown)


sonalidhome (approved):
Ack. Changes look good now.

PR #2985, opened 2021-03-25
ENT-3677: Fix for Stack Overflow issue in product equality check.

wolfdale:
Fix for Stack Overflow issue in product equality check.

### DIFF
--- a/server/src/main/java/org/candlepin/model/Product.java
+++ b/server/src/main/java/org/candlepin/model/Product.java
@@ -37,6 +37,7 @@ import org.hibernate.annotations.LazyCollectionOption;
 import org.hibernate.annotations.Type;
 
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -1205,8 +1206,15 @@ public class Product extends AbstractHibernateObject implements SharedEntity, Li
                 .append(this.name, that.name)
                 .append(this.multiplier, that.multiplier)
                 .append(this.attributes, that.attributes)
-                .append(this.derivedProduct, that.derivedProduct)
                 .isEquals();
+
+            // When checking children versioned entities, we're only interested in verifying that the
+            // reference is the same, otherwise we could run into some weird linkage problems (also it's
+            // much faster)
+            Comparator<Product> productUuidComparator = Comparator.comparing(product -> product != null ?
+                product.getUuid() : null, Comparator.nullsLast(Comparator.naturalOrder()));
+
+            equals = equals && productUuidComparator.compare(this.derivedProduct, that.derivedProduct) == 0;
 
             // Check our collections.
             // Impl note: We can't use .equals here on the collections, as Hibernate's special
@@ -1224,7 +1232,7 @@ public class Product extends AbstractHibernateObject implements SharedEntity, Li
 
             // Compare provided product
             equals = equals && Util.collectionsAreEqual(this.providedProducts, that.providedProducts,
-                (b1, b2) -> Objects.equals(b1, b2) ? 0 : 1);
+                productUuidComparator);
 
         }
 
--- a/server/src/test/java/org/candlepin/model/ProductTest.java
+++ b/server/src/test/java/org/candlepin/model/ProductTest.java
@@ -34,7 +34,6 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-
 /**
  * ProductTest
  */
@@ -95,12 +94,26 @@ public class ProductTest {
             "arch2", "type2")
         );
 
+        for (Product product : providedP1) {
+            product.setUuid(product.getId() + "_uuid");
+        }
+
         Set<Product> providedP2 = Util.asSet(
             new Product("ak3", "providedProduct3", "varient3", "version3",
             "arch3" , "type3"),
             new Product("ak4", "providedProduct4", "varient4", "version4",
             "arch4" , "type4")
         );
+
+        for (Product product : providedP2) {
+            product.setUuid(product.getId() + "_uuid");
+        }
+
+        Product derivedProd1 = new Product("derivedProdId1", "dp-name-1");
+        derivedProd1.setUuid(derivedProd1.getId() + "_uuid");
+
+        Product derivedProd2 = new Product("derivedProdId2", "dp-name-2");
+        derivedProd2.setUuid(derivedProd2.getId() + "_uuid");
 
         return Stream.of(
             new Object[] { "Id", "test_value", "alt_value" },
@@ -110,6 +123,7 @@ public class ProductTest {
             new Object[] { "ProductContent", productContent1, productContent2 },
             new Object[] { "DependentProductIds", Arrays.asList("1", "2", "3"), Arrays.asList("4", "5") },
             new Object[] { "Branding", brandings1, brandings2 },
+            new Object[] { "DerivedProduct", derivedProd1, derivedProd2 },
             new Object[] { "ProvidedProducts", providedP1, providedP2 }
         );
     }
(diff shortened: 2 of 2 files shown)


Ceiu:
This ended up being the solution for another problem I was working on as well. I'll leave you to discover your own path here, but if you want to see what I came up with...

<details><summary>
  SPOILER
  </summary>

    @Override
    public boolean equals(Object obj) {
        if (this == obj) {
            return true;
        }

        boolean equals = false;

        if (obj instanceof Product) {
            Product that = (Product) obj;

            // TODO:
            // Maybe it would be better to check the UUID field and only check the following if
            // both products have null UUIDs? By not doing this check, we run the risk of two
            // different products being considered equal if they happen to have the same values at
            // the time they're checked, or two products not being considered equal if they
            // represent the same product in different states.

            equals = new EqualsBuilder()
                .append(this.id, that.id)
                .append(this.name, that.name)
                .append(this.multiplier, that.multiplier)
                .append(this.attributes, that.attributes)
                .isEquals();

            // When checking children versioned entities, we're only interested in verifying that the
            // reference is the same, otherwise we could run into some weird linkage problems (also it's
            // much faster)
            Comparator<Product> productUuidComparator = (p1, p2) ->
                Objects.equals(p1 != null ? p1.getUuid() : null, p2 != null ? p2.getUuid() : null) ? 0 : 1;

            equals = equals && productUuidComparator.compare(this.derivedProduct, that.derivedProduct) == 0;

            // Check our collections.
            // Impl note: We can't use .equals here on the collections, as Hibernate's special
            // collections explicitly state that they break the contract on .equals. As such, we
            // have to step through each collection and do a manual comparison. Ugh.
            equals = equals && Util.collectionsAreEqual(this.dependentProductIds, that.dependentProductIds);

            // Compare content UUIDs
            equals = equals && Util.collectionsAreEqual(this.productContent, that.productContent,
                (pc1, pc2) -> Objects.equals(pc1, pc2) ? 0 : 1);

            // Compare branding collections
            equals = equals && Util.collectionsAreEqual(this.branding, that.branding,
                (b1, b2) -> Objects.equals(b1, b2) ? 0 : 1);

            // Compare provided product
            equals = equals && Util.collectionsAreEqual(this.providedProducts, that.providedProducts,
                productUuidComparator);
        }

        return equals;
    }
</details>

wolfdale:
@Ceiu Looks good, but we still want to test the derived product & provided products equality when UUID doesn't match. right?
Adjusted the comparator to test equality when UUIDs don't match.
```
Comparator<Product> productUuidComparator = (p1, p2) ->
                Objects.equals(p1 != null && p1.getUuid() != null ? p1.getUuid() : p1,
                p2 != null && p2.getUuid() != null ? p2.getUuid() : p2) ? 0 : 1;
```

Ceiu:
Maybe. I would argue no, but if there's a use case here, how we handle null UUIDs gets a bit more complex. Moreover, we also need to consider how Content is checked today and make that line up if we change this.

Basically, as it is now, we only want to check that we're pointing to the correct reference -- as is the critical component of the issue I had. For products and content, this avoids needing to traverse the object graph to ensure our children entities are part of the same organization, and ensures equality of two products is only "true" if both point to the same children. Without this, we risk linking across owners in cases where parallel creation of entities causes two otherwise identical products to exist at the same time for different orgs.

If we do want to start checking full equality when the UUIDs don't match or aren't set, then we need to start doing it if *either* aren't set. That changes how the comparator will work to something a bit closer to:

```
return (p1 != null && p2 != null) && (p1.getUuid() != null && p2.getUuid() != null) ? Objects.equals(p1.getUuid(), p2.getUuid()) : Objects.equals(p1, p2);
```

But, again, I'm not sure what the use case is for this logic (probably unit tests), so I'm not sure if we need it. If we do opt to use this, the content check in ProductContent.equals should also be updated in kind.

server/src/test/java/org/candlepin/model/ProductTest.java
```diff
@@ -160,6 +160,17 @@ public void testBaseEquality() {
...
 
+    /**
+     * This test case is disabled.
+     * We updated product equals method to check only UUIDs for derived product & provided products.
+     * This update is done to fix stackoverflow issues when derived or provided products create a cycle
+     * within the hierarchy.
+     * It avoids needing to traverse the object graph to ensure our children entities are part of the
+     * same organization, & ensures equality of two products is only "true" if both point to the same
+     * children. Without this, we risk linking across owners in cases where parallel creation of entities
+     * causes two otherwise identical products to exist at the same time for different orgs.
```
nikosmoum, line 171: Let's add this comment as part of the commit message (except the 'This test is disabled'), after the summary and empty line

server/src/main/java/org/candlepin/model/Product.java
```diff
@@ -1205,9 +1206,16 @@ public boolean equals(Object obj) {
...
                 .append(this.multiplier, that.multiplier)
                 .append(this.attributes, that.attributes)
-                .append(this.derivedProduct, that.derivedProduct)
                 .isEquals();
 
+            // When checking children versioned entities, we're only interested in verifying that the
+            // reference is the same, otherwise we could run into some weird linkage problems (also it's
+            // much faster)
+            Comparator<Product> productUuidComparator = (p1, p2) ->
+                Objects.equals(p1 != null ? p1.getUuid() : null, p2 != null ? p2.getUuid() : null) ? 0 : 1;
```
Januson, line 1215: 3 ternary operators on a single line are a bit scary. I would suggest taking advantage of Comparator.comparing to compose the comparator.

Something like this.

```
Comparator.comparing(p -> p != null ? p.getUuid() : null, Comparator.nullsLast(Comparator.naturalOrder()))
```

server/src/test/java/org/candlepin/model/ProductTest.java
```diff
@@ -245,4 +256,27 @@ public void testClone(String valueName, Object value1, Object value2) throws Exc
...
+        Product derivedProduct = new Product();
+
+        product.setUuid("test-UUID-1");
+        derivedProduct.setUuid("test-UUID-2");
+
+        // Creating endless cyclic product hierarchy
+        product.setDerivedProduct(derivedProduct);
+        derivedProduct.setDerivedProduct(product);
+
+        assertFalse(product.equals(derivedProduct));
```
Januson, line 271: You can use `assertNotEquals` for a bit more readable assertion on equality.

server/src/main/java/org/candlepin/model/Product.java
Ceiu: An alternate approach here could be to update the ```Util.collectionsAreEqual``` to use a ```BiFunction``` or a custom functional interface instead of a Comparator. Those utilities are fairly old now, and were written before functional interfaces were available to us, and without the knowledge of the BiFunction interface.  
At that point we get rid of one of the ternary operations that normalizes the output for what a Comparator expects.

Might be out of scope for this PR, however.

server/src/test/java/org/candlepin/model/ProductTest.java
Ceiu: I would argue that we don't disable this test. We still want to verify that the equality works as expected with each field populated individually. We should add additional fields to the output of the source method to verify that derived and provided products are checked as expected.

Ceiu (changes requested):

server/src/test/java/org/candlepin/model/ProductTest.java
wolfdale: fixed.

server/src/main/java/org/candlepin/model/Product.java
wolfdale: Changed to use ```comparing```.

server/src/test/java/org/candlepin/model/ProductTest.java
wolfdale: Modified the test case. Also added the summary in git commit.

nikosmoum:
ack from me. I'll leave @Ceiu and @Januson to re-review for their feedback

Ceiu (approved):

PR #2995, opened 2021-04-06
ENT-3491: Addressing sporadic spec test failure in import_update_spec.rb

sonalidhome:
- The issue is in MySQL which does not store partial seconds.
   This was already addressed in commit 82a4051. To resolve,
   added `sleep 2` before the next export operation executes. 
   This ensures there is a time difference between the exports.


wolfdale (approved):

nikosmoum:
@sonalidhome Let's port this to 3.1 and 3.2 branches

PR #2998, opened 2021-04-12
[3.2] change default content access mode list

jdhrh:


Ceiu (approved):

nikosmoum (changes requested):
ack on the change, but let's keep the changelog clear by changing the commit message to the same or similar with the master commit (e.g. " change default content access mode list"), instead of mentioning that it's a backport and the which branch its going in

PR #3000, opened 2021-04-14
[3.1] 1949698: Fix SCA cert regeneration for consumers in environments (ENT-3784)

Ceiu:
- Fixed a duplicate key error that could occur when regenerating
  SCA certs for a consumer that already had an expired or outdated
  certificate that was part of an environment

### DIFF
--- a/server/spec/content_access_spec.rb
+++ b/server/spec/content_access_spec.rb
@@ -43,8 +43,18 @@ describe 'Content Access' do
     sleep 1
   end
 
-  it "does allow addition of the content access level" do
+  def promote_content_to_environment(user, environment, content, enabled = true)
+    job = user.promote_content(environment['id'], [
+      {
+        :contentId => content['id'],
+        :enabled => enabled
+      }
+    ])
 
+    wait_for_job(job['id'], 15)
+  end
+
+  it "does allow addition of the content access level" do
     @cp.update_owner(@owner['key'], {'contentAccessMode' => "entitlement"})
     @owner = @cp.get_owner(@owner['key'])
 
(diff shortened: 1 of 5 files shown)


Ceiu:
Once this is approved and merged, we'll need to forward port this to 3.2 and beyond.

server/spec/content_access_spec.rb
```diff
@@ -726,23 +722,42 @@ def regenerate_cert_test(&updater)
...
-    regenerate_cert_test do
+    client = consumer_client(@user, @consumername, type=:system, username=nil, facts= {'system.certificate_version' => '3.3'})
+
+    regenerate_cert_test(client) do
       @pool['start_date'] = (DateTime.now + 1)
       @cp.update_pool(@owner['key'], @pool)
     end
   end
 
+  it 'should regenerate SCA cert when environment content changes' do
```
nikosmoum, line 749: This test doesn't fail when I run it against a candlepin without the fix. I am also having trouble reproducing the issue manually (either doing unregister/register, or subman refresh)

nikosmoum: I have tried reproducing manually with both using an environment, and without an environment, but still cannot reproduce

Ceiu: And it shouldn't. This test was the first one I wrote to attempt to reproduce it after the initial report. It doesn't hit the issue because the root cause here is that a consumer with a non-null environment needs to have an existing cert that expires naturally which would then be regenerated. The spec tests cannot verify this path without priming/modifying the database, so this is done in the unit tests which were also added as part of this PR.

nikosmoum (approved):

PR #3002, opened 2021-04-15
1935500: Update SCA to cache certs by consumer

Januson:
- Drop the old caching table


server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -309,56 +299,27 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
             existing.setKeyAsBytes(pemEncodedKeyPair);
             existing.setConsumer(consumer);
 
+            log.info("Generating new SCA payload for organization \"{}\"...", owner.getKey());
+
+            StringBuilder pem = new StringBuilder();
+            X509Certificate x509Cert = this.createX509Certificate(consumer, owner,
+                BigInteger.valueOf(serial.getId()), keyPair, startDate, endDate);
             pem.append(new String(this.pki.getPemEncoded(x509Cert)));
+            pem.append(this.createPayloadAndSignature(owner, env));
```
nikosmoum, line 313: Now that we're doing per-consumer-caching, let's pass in the actual consumer object here (and down the call stack of this method) so that arch filtering will also work

server/src/main/resources/db/changelog/20210412145432-drop_content_cache.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20210412145432-1" author="ojanus">
+        <sql>UPDATE cp_cert_serial SET expiration = NOW() WHERE id IN (SELECT serial_id FROM cp_cont_access_cert)</sql>
```
nikosmoum, line 10: Could the results returned from that subquery hit a limit with either mariadb/postgres for amount of args in the IN clause?
Perhaps another way of writing this query would be safer

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -546,40 +507,42 @@ private String createDN(Consumer consumer, Owner owner) {
...
-
-        containerSet.add(container);
         container.setId("content_access");
         container.setName(" Content Access");
 
         this.ownerContentCurator.getActiveContentByOwner(owner.getId())
-            .forEach((content, enabled) -> container.addContent(content, enabled));
+            .forEach(container::addContent);
 
+        Consumer emptyConsumer = new Consumer();
```
Ceiu, line 519: Now that we're passing the consumer in, and these certs are generated per-consumer, do we need to generate a fake consumer here at all?

Ceiu (changes requested):
One question on the use of the consumer we're now passing through.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
Januson: Done.

server/src/main/resources/db/changelog/20210412145432-drop_content_cache.xml
Januson: Done.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
Januson: That's true. We don't need a fake one anymore since we have a real one. Fixed.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -282,9 +272,7 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
             Date expiration = existing.getSerial().getExpiration();
             Date contentUpdate = owner.getLastContentUpdate();
 
-            if (expiration.before(now) || !contentUpdate.before(existing.getUpdated()) ||
-                oeca == null || !contentUpdate.before(oeca.getUpdated())) {
-
+            if (expiration.before(now) || !contentUpdate.before(existing.getUpdated())) {
```
nikosmoum, line 275: I thought that with this change, we could get rid of the `Owner.getLastContentUpdate` workaround?

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -763,7 +765,12 @@ public Owner updateOwnerContentAccess(Owner owner, String updatedList, String up
             owner = this.ownerCurator.merge(owner);
             ownerCurator.flush();
 
-            this.refreshOwnerForContentAccess(owner);
+            this.syncOwnerLastContentUpdate(owner);
+            // TODO cleanup
+            // Delete cert when turning off SCA
+            if (currentMode.equals("org_environment") && updatedMode.equals("entitlement")) {
```
Ceiu, line 771: We should be using the enums here rather than the strings.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -265,106 +260,126 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
+        ContentAccessCertificate savedCert = this.contentAccessCertificateCurator.create(newCert);
+        consumer.setContentAccessCert(savedCert);
+        this.consumerCurator.merge(consumer);
+        return savedCert;
+    }
+
+    /**
+     * Represents a duration of certificate validity
+     */
+    private static class Validity {
```
Ceiu, line 312: I feel this class, and the way its used currently, is unnecessarily fancy given the goal here. If we're going to start "fixing" the way we juggle date and times, I'd like to see a proper utility making use of the java.time package which supplants all the date junk we've been doing up to now -- which would be another PR.

server/src/main/java/org/candlepin/model/ContentAccessCertificate.java
```diff
@@ -69,6 +72,18 @@ public void setConsumer(Consumer consumer) {
...
 
+    public String getContent() {
+        return content;
+    }
+
+    public void setContent(String content) {
+        this.content = content;
+    }
+
+    public String getFullCert() {
```
Ceiu, line 83: This feels wrong to me. Rather than have a special method that builds the "full" cert, the default "getCert" behavior should be the full thing, and the specific components should be fetchable individually.

Anything using the cert generically should not have to fumble with specifics of the SCA cert. It should only be when the individual components of the SCA cert are of interest that we start working with explicit classes and custom methods.

server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1896,9 +1896,11 @@ public void deleteConsumer(@Verify(Consumer.class) String uuid) {
 
         // we want to insert the content access cert to this list if appropriate
         try {
-            Certificate cert = this.contentAccessManager.getCertificate(consumer);
+            ContentAccessCertificate cert = this.contentAccessManager.getCertificate(consumer);
```
Ceiu, line 1899: This follows with what I was saying above: this bit of code shouldn't change, as it is not interested in details about the SCA cert itself, and is just serializing it as a generic cert.

server/src/main/java/org/candlepin/sync/Exporter.java
```diff
@@ -435,7 +435,7 @@ private void exportContentAccessCerts(File baseDir, Consumer consumer) throws IO
             log.debug("Exporting content access certificate: " + contentAccessCert.getSerial());
             File file = new File(contentAccessCertDir.getCanonicalPath(),
                 contentAccessCert.getSerial().getId() + ".pem");
-            CertificateExporter crt = new CertificateExporter();
+            ScaCertificateExporter crt = new ScaCertificateExporter();
```
Ceiu, line 438: More regarding the cert handling behavior here. If it behaved like a normal cert by default, we wouldn't need a custom writer for it.

server/src/main/java/org/candlepin/sync/ScaCertificateExporter.java
```diff
@@ -0,0 +1,38 @@
...
+import org.candlepin.model.ContentAccessCertificate;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+/**
+ * Certificate exporter
+ */
+public class ScaCertificateExporter {
```
Ceiu, line 26: nack on this for reasons stated in other comments

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -265,106 +260,126 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
+        if (!contentUpdate.before(existing.getUpdated())) {
+            existing.setContent(this.createPayloadAndSignature(owner, consumer));
+            this.contentAccessCertificateCurator.saveOrUpdate(existing);
         }
+        return existing;
+    }
 
-        if (existing == null || regenerate) {
-            log.info("Generating new SCA certificate for organization \"{}\"...", owner.getKey());
+    private ContentAccessCertificate createNewScaCertificate(Consumer consumer, Owner owner)
```
Ceiu, line 299: Given that we're using merge, saveOrUpdate, and other such state-agnostic persistence methods, is there any reason this method needs to exist individually? It seems separating the create and update operations isn't doing much for us, and the two could easily just be a single create-or-update style method as it was before.

There's a ton of abstraction happening here and lots of jumping around the code for no real gain.

Ceiu (changes requested):

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
Januson: Done.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
Januson: Sure, moving to time API would help in some aspects. We could even start using it in places like this, but it will not bring many benefits until we start using it in the entities so that we don't have to convert between the two API. The reason for this class is that startDate and endDate need to be created and passed together. I guess it could be modeled as two methods instead but that felt like even more trouble so I picked a value object to encapsulate this date operation with a possibility to parameterize it later.

server/src/main/java/org/candlepin/model/ContentAccessCertificate.java
Januson: Rollbacked to the wrapping solution.

server/src/main/java/org/candlepin/resource/ConsumerResource.java

server/src/main/java/org/candlepin/sync/Exporter.java
Januson: Done.

server/src/main/java/org/candlepin/sync/ScaCertificateExporter.java
Januson: Done.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
Januson: I shuffled a bit more with the methods and inlined some of them. I think that splitting it helps a bit with navigation. There is a lot of low level thing happening with cert generation. This way you can get an overview of what the getCertificate method does and move to details of creation/update if you need to.

nikosmoum:
A note here: Let's avoid using jira ids in the commit message, and instead use the bz of the arch-filtering bug: 1935500 (but the commit message should still make clear that that is only part of the change).

server/spec/content_access_spec.rb
```diff
@@ -92,6 +92,43 @@ def set_content_access_mode_for_standalone_owner(owner, access_mode, force_distr
...
+    @cp.add_content_to_product(@owner['key'], product2['id'], content_id2)
+    @cp.create_pool(@owner['key'], product2['id'], {:quantity => 10})
+
+    @consumer = consumer_client(@user, @consumername, type=:system, username=nil, facts={'system.certificate_version' => '3.3', "uname.machine" => "ppc64"}, owner_key=@owner['key'])
+    certs = @consumer.list_certificates
+    certs.length.should == 1
+
+    json_body = extract_payload(certs[0]['cert'])
+
+    content = json_body['products'][0]['content'][0]
```
nikosmoum, line 117: This test implicitly tests that only content2 (which was added as part of the test) will be available in the cert, while content (which was added as part of the before method) will not be available, but doesn't actually assert this. Let's add an assertion that the json_body does not actually contain a second content (to make sure it's filtered)

server/src/main/java/org/candlepin/util/X509Util.java
```diff
@@ -183,41 +172,8 @@ public String createFullContentPath(String contentPrefix, ProductContent pc) {
...
-                            break searchArch;
-                        }
-                        else {
-                            canUse = false;
-                        }
-                    }
-                }
-            }
+            Set<String> contentArches = archesOf(product, pc);
+            boolean canUse = hasCompatibleArchitecture(consumerArches, contentArches);
```
nikosmoum, line 176: Good job on refactoring this method for readability, but it seems there were no unit tests for it. Let's add some unit tests now, since refactoring code without existing unit tests is prime for regressions appearing

server/src/test/java/org/candlepin/controller/ContentAccessManagerDBTest.java
```diff
@@ -0,0 +1,185 @@
...
+    @Test
+    public void testGetCertificateRegeneratesExpiredCerts() throws Exception {
+        Owner owner = this.createSCAOwner();
+        Consumer consumer = this.createV3Consumer(owner, null);
+
+        this.regenerateExpiredSCACertTest(consumer);
+    }
+
+    @Test
+    public void testGetCertificateRegeneratesExpiredCertsWithEnvironment() throws Exception {
```
nikosmoum, line 177: The tests in this class cover a good amount of use cases, e.g. new SCA cert creation with/without environment, expired SCA cert regeneration with/without environment, but there's another area missing: testing that an existing (non-expired) SCA cert has only it's content payload updated (not the x509/serial) with new content (could be a unit test here, or could be a spec test; up to you)

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -265,121 +260,122 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
-            Date expiration = existing.getSerial().getExpiration();
-            Date contentUpdate = owner.getLastContentUpdate();
+        return wrap(result);
+    }
 
-            if (expiration.before(now) || !contentUpdate.before(existing.getUpdated()) ||
-                oeca == null || !contentUpdate.before(oeca.getUpdated())) {
+    private ContentAccessCertificate createNewScaCertificate(Consumer consumer, Owner owner)
+        throws IOException, GeneralSecurityException {
+        log.info("Generating new SCA certificate for organization \"{}\"...", owner.getKey());
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -265,121 +260,122 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
-            }
-        }
+        ContentAccessCertificate existing = new ContentAccessCertificate();
+        existing.setSerial(serial);
+        existing.setKeyAsBytes(pemEncodedKeyPair);
+        existing.setConsumer(consumer);
 
-        if (existing == null || regenerate) {
-            log.info("Generating new SCA certificate for organization \"{}\"...", owner.getKey());
+        log.info("Generating new SCA payload for organization \"{}\"...", owner.getKey());
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -265,121 +260,122 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
-            existing = this.contentAccessCertificateCurator.create(existing);
-            consumer = this.consumerCurator.merge(consumer);
+            existing.setCert(createX509Cert(consumer, owner, serial, keyPair, oneYearValidity));
+            this.contentAccessCertificateCurator.saveOrUpdate(existing);
         }
-        else {
-            pem.append(existing.getCert());
+        // regenerate content
+        Date contentUpdate = owner.getLastContentUpdate();
+        if (!contentUpdate.before(existing.getUpdated())) {
```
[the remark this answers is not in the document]
Januson: Done.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -265,121 +260,122 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
-            existing.setKeyAsBytes(pemEncodedKeyPair);
-            existing.setConsumer(consumer);
-
-            pem.append(new String(this.pki.getPemEncoded(x509Cert)));
-            existing.setCert(pem.toString());
-            consumer.setContentAccessCert(existing);
-
-            existing = this.contentAccessCertificateCurator.create(existing);
-            consumer = this.consumerCurator.merge(consumer);
+            existing.setCert(createX509Cert(consumer, owner, serial, keyPair, oneYearValidity));
```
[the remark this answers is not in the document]
Januson: Done. Added log statement into createX509Cert.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -265,121 +260,122 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
+            this.contentAccessCertificateCurator.saveOrUpdate(existing);
         }
-        else {
-            pem.append(existing.getCert());
+        // regenerate content
+        Date contentUpdate = owner.getLastContentUpdate();
+        if (!contentUpdate.before(existing.getUpdated())) {
+            existing.setContent(this.createPayloadAndSignature(owner, consumer));
```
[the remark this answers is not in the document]
Januson: Done. Added log statement into createPayloadAndSignature.

server/spec/content_access_spec.rb
Januson: Done.

server/src/main/java/org/candlepin/util/X509Util.java
Januson: Done.

server/src/test/java/org/candlepin/controller/ContentAccessManagerDBTest.java
Januson: Done. Added spec tests.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -265,122 +260,124 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
+        return savedCert;
+    }
 
+    private ContentAccessCertificate updateScaCertificate(Consumer consumer, Owner owner,
+        ContentAccessCertificate existing) throws GeneralSecurityException, IOException {
+        Date now = new Date();
+        Date expiration = existing.getSerial().getExpiration();
+        Date contentUpdate = owner.getLastContentUpdate();
+        // Checks need to be done before any updates!
+        boolean shouldUpdateContent = !contentUpdate.before(existing.getUpdated());
```
nikosmoum, line 304: not sure why this needs to be checked here, instead of right before `if (shouldUpdateContent) {`?

nikosmoum:
ACK from me. I'll let @Ceiu re-review

nikosmoum (approved):

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -265,122 +260,124 @@ public ContentAccessCertificate getCertificate(Consumer consumer)
...
-        if (oeca == null || regenerate) {
-            log.info("Generating new SCA payload for organization \"{}\"...", owner.getKey());
-
-            String contentJson = this.createPayloadAndSignature(owner, env);
-            oeca = new OwnerEnvContentAccess(owner, env, contentJson);
-            oeca = this.ownerEnvContentAccessCurator.saveOrUpdate(oeca);
+        Date contentUpdate = owner.getLastContentUpdate();
+        boolean shouldUpdateContent = !contentUpdate.before(existing.getUpdated());
+        if (shouldUpdateContent || isX509CertExpired) {
+            existing.setContent(this.createPayloadAndSignature(owner, consumer));
```
nikosmoum, line 316: @Januson @Ceiu A bit of a crazy idea I wanted to fly by you two here for the race condition:
What if, after we run `createPayloadAndSignature()` (which we assume to be an expensive operation), we do a refresh of the Owner object to see if the lastContentUpdate has changed during the time it was running? And if it did change, we scrap the newly generated payload and return the cert as is? This way we would reduce even further the chance for the race condition. The client might have stale data, but at least they won't miss the content entirely, since they'll be able to pick it up next time they check in?

So something like:
```
 if (shouldUpdateContent || isX509CertExpired) {
  String payload = this.createPayloadAndSignature(owner, consumer);

  this.ownerCurator.refresh(owner);
  // New content update sneaked in; return without changes
  if (contentUpdate.before(owner.getLastContentUpdate())) {
    return existing;
  }

  // no new content sneaked in, persist changes  
  existing.setContent(payload);
  this.contentAccessCertificateCurator.saveOrUpdate(existing);
 }
```
Do you think this will work? Any downsides?

Ceiu: This could work, but it has the downside of being rather awkward and confusing to users who receive stale content for a blip, and then get corrected.

If we're going to check this post-generation (which is fair), we should just make this part a post-check loop that'll immediately do the regeneration again if we detect an update on the org. It's effectively the same behavior, except we save the user the confusion and re-fetching of the cert.

nikosmoum: Based on discussion in irc: this idea is good, but we'll need to think it through a bit more, with some design/diagrams.

nikosmoum: @Ceiu I had not the time to formulate all the possible proposals to enhance this; given that is holding back this PR from getting merged, and that this PR can stand by itself, are you ok with merging this as is, and treating the enhancement(s) as separate tasks?

Ceiu: Sure

Ceiu (approved):

PR #3003, opened 2021-04-15
[4.0] 1949980: Fix SCA cert regeneration for consumers in environments (ENT-3784)

Ceiu:
- Fixed a duplicate key error that could occur when regenerating
  SCA certs for a consumer that already had an expired or outdated
  certificate that was part of an environment

### DIFF
--- a/server/spec/content_access_spec.rb
+++ b/server/spec/content_access_spec.rb
@@ -92,9 +92,19 @@ describe 'Content Access' do
     end
   end
 
+  def promote_content_to_environment(user, environment, content, enabled = true)
+    job = user.promote_content(environment['id'], [
+      {
+        :contentId => content['id'],
+        :enabled => enabled
+      }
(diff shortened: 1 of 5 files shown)


Ceiu:
We may want to hold off on merging this PR until we decide what we're doing with the caching change in master. If that ends up being better to backport, this fix should be omitted from 4.0.

nikosmoum:
@Ceiu Would it be problematic if we merge this now, but also porting the per-consumer-caching on this branch later once ready?

nikosmoum (approved):

Ceiu:
Yeah. We could also just leave this branch as-is and focus the efforts of the caching changes purely in 4.1. I'm fine with either.

nikosmoum:
Leaving per-consumer-caching for 4.1 only sounds fine I think. Merging this then

PR #3006, opened 2021-04-20
[M] Add timeout to Jenkins pipeline

nikosmoum:


nikosmoum:
Why we need this: jenkins pipelines on Weblate PRs need approval to run, and they wait for the approval indefinitely; combining that with the fact that every time the PR branch is pushed/force pushed a new pipeline is triggered, makes it so that these PRs hoard a large amount of Jenkins resources just waiting for approval. Adding a timeout will make sure the resources are released eventually.

nikosmoum:
Note: I have added the timeout at 5 minutes just for testing. If it works, I will update it to be hours.

sonalidhome (approved):

PR #3010, opened 2021-04-21
[3.2] 1949122: Changed how content access  mode is selected for new consumers (ENT-3755)

Ceiu:
- During consumer creation, the content access mode is now ignored and
  the "best" mode will be selected based on modes available to the
  org in which the consumer is being created

### DIFF
--- a/server/spec/candlepin_scenarios.rb
+++ b/server/spec/candlepin_scenarios.rb
@@ -23,7 +23,7 @@ module CandlepinMethods
     # Set the content access mode list for new owners. We do this here instead of doing it explicitly
     # in each test due to some test objects creating owners internally which need this list.
     if not params['contentAccessModeList']
-      params['contentAccessModeList']= 'org_environment,entitlement'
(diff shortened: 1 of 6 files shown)


Ceiu:
Once this is verified and merged, it will be forward ported to master

Ceiu:
I think this might need some more thought/attention once the use cases here are finished. The more I think on this, the more I'm opposed to this change, to the point where I'm willing to nack my own PR.

This change introduces behavior I absolutely detest -- the application implying it knows what the user wants more than the user. There's a fair argument to be made about input validation, but in the case of valid behavior, an application should never override that and choose for the user by default/design. To throw some buzzwords at this: it's an aggressively anti-user antipattern.

What I would like to see here is that we respect valid values passed for the content access mode, and default to the best available only when the value provided is null, empty, or not present.

nikosmoum:
The double negation on 'anti-user antipattern' then makes it a 'pro-user pattern' right? :rofl: 
Joking aside this is a good point.

PR #3015, opened 2021-04-27
[3.2] 1949122: Changed how content access mode is selected for new consumers (ENT-3755)

Ceiu:
- During the creation of a manifest consumer, if no content access mode
  is provided, a default value will be chosen by using the "best" mode
  from the org's available modes: SCA if available, entitlement mode
  otherwise


Ceiu:
This PR has been updated according to the more reasonable semantics which let the API user dictate exactly what behavior they want, following the conventions we use elsewhere:

- null or absent to indicate that a default should be chosen from the "best" mode available to the org
- empty to indicate the consumer should default to the org's mode
- a specific value indicating the mode to use, which must be available to the org

Januson (approved):

PR #3018, opened 2021-04-28
[M] 1949125: Changed how content access mode is validated for new consumers (ENT-3755)

Ceiu:
- During the creation of a manifest consumer, if no content access mode
  is provided, a default value will be chosen by using the "best" mode
  from the org's available modes: SCA if available, entitlement mode
  otherwise


sonalidhome (approved):
Ack.

PR #3024, opened 2021-04-30
Bumped version of Openapi generator maven (& gradle) plugin

wolfdale:
Bumped version of openapi generator maven (& gradle) plugin to v5.1.0


sonalidhome (approved):
ack.

PR #3028, opened 2021-05-06
1957779: Fix SCA cert regeneration for consumers in environments

Januson:


nikosmoum (commented):
Let's use 1957779 as the bz id (clone of the original)

nikosmoum:
Looks like there's some compilation errors

PR #3030, opened 2021-05-12
ENT-3856: Fix to revoke SCA cert upon client unregistration

wolfdale:
Changes
- Fix to revoke SCA cert when client unregisters.

### DIFF
--- a/server/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/server/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -1877,7 +1877,7 @@ public class ConsumerResource implements ConsumersApi {
         }
 
         consumerRules.onConsumerDelete(toDelete);
-
+        contentAccessManager.removeContentAccessCert(toDelete);
         Event event = eventFactory.consumerDeleted(toDelete);
         consumerCurator.delete(toDelete);
         identityCertService.deleteIdentityCert(toDelete);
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1877,6 +1880,11 @@ public void deleteConsumer(@Verify(Consumer.class) String uuid) {
         }
 
         consumerRules.onConsumerDelete(toDelete);
+        ContentAccessCertificate cert = toDelete.getContentAccessCert();
+
+        if (cert != null) {
+            cacCurator.delete(cert);
+        }
```
nikosmoum, line 1887: Let's use `ContentAccessManager.removeContentAccessCert()` here instead, since it already performs the necessary checks, and we don't have to add another dependency to this class (ContentAccessManager is already a dependency)

nikosmoum: Also, this is a good chance to add some unit tests for `ContentAccessManager.removeContentAccessCert()` itself, since I don't see any in `ContentAccessManagerTest.java`. It's a simple method, but we should strive to increase test coverage of all the code we touch

nikosmoum (changes requested):
A couple minor changes needed.
Also, we should avoid adding jira ids to commits. If a change is not related with a bz, we should not add any id. (the jira id can still be added to the PR summary instead)

server/src/main/java/org/candlepin/resource/ConsumerResource.java
wolfdale: Done. Added unit test for same.

nikosmoum (approved):

PR #3031, opened 2021-05-12
[3.2] 1960150: Fixed NPE when listing future pools with entitlements (ENT-3862)

Ceiu:
- Fixed a NullPointerException that could occur while listing future
  pools that had associated entitlements


Ceiu:
The spec test to verify the issue is resolved needs more work. At the time of writing, it does not yet trigger the issue in unpatched versions of Candlepin.

nikosmoum (approved):
ack. just need to update the commit message to use the 3.2 bz id

PR #3035, opened 2021-05-14
ENT-2798: [RFE] Add service_type system purpose attribute

sonalidhome:
- Added new column `sp_service_type` in `cp_consumer` table.
 - Updated the functionality to calculate the `get_pool_priority`
   (rules.js) to consider service type with lower impact than
   other system purpose attributes.
   For calculations, consumers `serviceType`
   is matched with the product's `support_type` attribute.
 - Updated the system purpose compliance calculations to consider
   new attribute.
 - Updated the `ConsumerDTO`, `ConsumerDTOArrayElement` and,
   `SystemPurposeComplianceStatusDTO` which holds the service
   type attribute. All endpoints that use these DTOs will reflect
   new attribute.

### DIFF
--- a/server/src/main/java/org/candlepin/model/Consumer.java
+++ b/server/src/main/java/org/candlepin/model/Consumer.java
@@ -272,6 +272,11 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
     @OneToMany(mappedBy = "consumer", orphanRemoval = true, cascade = { CascadeType.ALL })
     private Set<ConsumerActivationKey> activationKeys;
 
+    @Column(name = "sp_service_type")
+    @Type(type = "org.candlepin.hibernate.EmptyStringUserType")
+    @Size(max = 255)
+    private String serviceType;
+
     public Consumer(String name, String userName, Owner owner, ConsumerType type) {
         this();
 
@@ -1009,5 +1014,14 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
 
     public Set<ConsumerActivationKey> getActivationKeys() {
         return this.activationKeys;
+    }
+
+    public String getServiceType() {
+        return serviceType;
+    }
+
+    public Consumer setServiceType(String serviceType) {
+        this.serviceType = serviceType;
+        return this;
     }
 }
--- a/server/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/server/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -60,6 +60,7 @@ import java.util.stream.Collectors;
 
 import javax.inject.Provider;
 import javax.inject.Singleton;
+import javax.persistence.EntityManager;
 import javax.persistence.LockModeType;
 import javax.persistence.PersistenceException;
 import javax.persistence.TypedQuery;
@@ -69,6 +70,7 @@ import javax.persistence.criteria.Join;
 import javax.persistence.criteria.JoinType;
 import javax.persistence.criteria.MapJoin;
 import javax.persistence.criteria.Order;
+import javax.persistence.criteria.Path;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
 
@@ -467,66 +469,6 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
         }
 
         return Collections.emptyList();
-    }
-
-    /**
-     * Fetches all unique role attribute values set by all the consumers of the specified owner.
-     *
-     * @param owner
-     *  The owner the consumers belong to.
-     * @return
-     *  A list of the all the distinct values of the role attribute that the consumers belonging to the
-     *  specified owner have set.
-     */
-    @SuppressWarnings("unchecked")
-    @Transactional
-    public List<String> getDistinctSyspurposeRolesByOwner(Owner owner) {
-        return this.createSecureCriteria()
-            .add(Restrictions.eq("ownerId", owner.getId()))
-            .add(Restrictions.neOrIsNotNull("role", ""))
-            .setProjection(Projections.property("role"))
-            .setResultTransformer(Criteria.DISTINCT_ROOT_ENTITY)
-            .list();
-    }
-
-    /**
-     * Fetches all unique usage attribute values set by all the consumers of the specified owner.
-     *
-     * @param owner
-     *  The owner the consumers belong to.
-     * @return
-     *  A list of the all the distinct values of the usage attribute that the consumers belonging to the
-     *  specified owner have set.
-     */
-    @SuppressWarnings("unchecked")
-    @Transactional
-    public List<String> getDistinctSyspurposeUsageByOwner(Owner owner) {
-        return this.createSecureCriteria()
-            .add(Restrictions.eq("ownerId", owner.getId()))
-            .add(Restrictions.neOrIsNotNull("usage", ""))
-            .setProjection(Projections.property("usage"))
-            .setResultTransformer(Criteria.DISTINCT_ROOT_ENTITY)
-            .list();
-    }
-
-    /**
-     * Fetches all unique serviceLevel attribute values set by all the consumers of the specified owner.
-     *
-     * @param owner
-     *  The owner the consumers belong to.
-     * @return
-     *  A list of the all the distinct values of the serviceLevel attribute that the consumers belonging
-     *  to the specified owner have set.
-     */
-    @SuppressWarnings("unchecked")
-    @Transactional
-    public List<String> getDistinctSyspurposeServicelevelByOwner(Owner owner) {
-        return this.createSecureCriteria()
-            .add(Restrictions.eq("ownerId", owner.getId()))
-            .add(Restrictions.neOrIsNotNull("serviceLevel", ""))
-            .setProjection(Projections.property("serviceLevel"))
-            .setResultTransformer(Criteria.DISTINCT_ROOT_ENTITY)
-            .list();
     }
 
     /**
@@ -1280,7 +1222,7 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
      * Builds a collection of predicates to be used for querying consumers using the JPA criteria
      * query API.
      *
-     * @param critBuilder
+     * @param criteriaBuilder
      *  the CriteriaBuilder instance to use to create predicates
 
      * @param root
@@ -1559,6 +1501,55 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
         }
 
         return updated;
+    }
+
+    /**
+     * Fetches all unique system purpose attribute values set by all the consumers
+     * of the specified owner.
+     *
+     * @param owner
+     *  The owner the consumers belong to
+     * @param sysPurposeAttribute
+     *  The type of system purpose attribute needs to be fetched
+     * @return
+     *  A list of the all the distinct values of the system purpose attributes that the consumers
+     *  belonging to the specified owner have set
+     */
+    public List<String> getDistinctSyspurposeValuesByOwner(Owner owner,
+        SystemPurposeAttributeType sysPurposeAttribute) throws RuntimeException {
+
+        EntityManager entityManager = this.getEntityManager();
+        CriteriaBuilder builder = entityManager.getCriteriaBuilder();
+        CriteriaQuery<String> query = builder.createQuery(String.class);
+        Root<Consumer> root = query.from(Consumer.class);
+        Path sysPurposePath;
+
+        switch (sysPurposeAttribute) {
+            case USAGE:
+                sysPurposePath = root.get(Consumer_.usage);
+                break;
+            case ROLES:
+                sysPurposePath = root.get(Consumer_.role);
+                break;
+            case SERVICE_LEVEL:
+                sysPurposePath = root.get(Consumer_.serviceLevel);
+                break;
+            case SERVICE_TYPE:
+                sysPurposePath = root.get(Consumer_.serviceType);
+                break;
+            default:
+                throw new RuntimeException("Unrecognized system purpose attribute: " + sysPurposeAttribute);
+        }
+
+        Predicate notNullPredicate = builder.isNotNull(sysPurposePath);
+        Predicate notEmptyPredicate = builder.notEqual(sysPurposePath, "");
+
+        query.select(sysPurposePath)
+            .where(builder.and(builder.equal(root.get(Consumer_.ownerId), owner.getId()),
+                notNullPredicate, notEmptyPredicate))
+            .distinct(true);
+
+        return entityManager.createQuery(query).getResultList();
     }
 
 }
--- a/server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
+++ b/server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
@@ -157,6 +157,7 @@ public class SystemPurposeComplianceRules {
             Set<String> unsatisfiedAddons = new HashSet<>(consumer.getAddOns());
             String preferredSla = consumer.getServiceLevel();
             String preferredUsage = consumer.getUsage();
+            String preferedServiceType = consumer.getServiceType();
             Pool entitlementPool = entitlement.getPool();
 
             Set<Product> entitlementProducts = new HashSet<>();
@@ -210,21 +211,23 @@ public class SystemPurposeComplianceRules {
                 }
                 unsatisfiedAddons.removeAll(addonsFound);
 
+                String productSLA = product.getAttributeValue(Product.Attributes.SUPPORT_LEVEL);
                 if (!slaExempted &&
                     StringUtils.isNotEmpty(preferredSla) &&
-                    product.hasAttribute(Product.Attributes.SUPPORT_LEVEL)) {
-                    String sla = product.getAttributeValue(Product.Attributes.SUPPORT_LEVEL);
-                    if (sla.equalsIgnoreCase(preferredSla)) {
-                        status.addCompliantSLA(sla, entitlement);
-                    }
+                    preferredSla.equalsIgnoreCase(productSLA)) {
+                    status.addCompliantSLA(productSLA, entitlement);
                 }
 
+                String productUsage = product.getAttributeValue(Product.Attributes.USAGE);
                 if (StringUtils.isNotEmpty(preferredUsage) &&
-                    product.hasAttribute(Product.Attributes.USAGE)) {
-                    String usage = product.getAttributeValue(Product.Attributes.USAGE);
-                    if (usage.equalsIgnoreCase(preferredUsage)) {
-                        status.addCompliantUsage(usage, entitlement);
-                    }
+                    preferredUsage.equalsIgnoreCase(productUsage)) {
+                    status.addCompliantUsage(productUsage, entitlement);
+                }
+
+                String productServiceType = product.getAttributeValue(Product.Attributes.SUPPORT_TYPE);
+                if (StringUtils.isNotEmpty(preferedServiceType) &&
+                    preferedServiceType.equalsIgnoreCase(productServiceType)) {
+                    status.addCompliantServiceType(productServiceType, entitlement);
                 }
             }
         }
@@ -244,12 +247,19 @@ public class SystemPurposeComplianceRules {
         }
 
         if (!slaExempted &&
-            StringUtils.isNotEmpty(consumer.getServiceLevel()) && status.getCompliantSLA().isEmpty()) {
+            StringUtils.isNotEmpty(consumer.getServiceLevel()) &&
+            status.getCompliantSLA() != null && status.getCompliantSLA().isEmpty()) {
             status.setNonCompliantSLA(consumer.getServiceLevel());
         }
 
-        if (StringUtils.isNotEmpty(consumer.getUsage()) && status.getCompliantUsage().isEmpty()) {
+        if (StringUtils.isNotEmpty(consumer.getUsage()) &&
+            status.getCompliantUsage() != null && status.getCompliantUsage().isEmpty()) {
             status.setNonCompliantUsage(consumer.getUsage());
+        }
+
+        if (StringUtils.isNotEmpty(consumer.getServiceType()) &&
+            status.getCompliantServiceType() != null && status.getCompliantServiceType().isEmpty()) {
+            status.setNonCompliantServiceType(consumer.getServiceType());
         }
 
         if (currentCompliance) {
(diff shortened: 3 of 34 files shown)


Ceiu (commented):
The logic seems fine enough to me for the most part, though I've not been in the loop on the design and needs of this feature. Mostly just have a few suggestions for the new code.

server/src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -980,4 +985,12 @@ public void setActivationKeys(Set<ConsumerActivationKey> activationKeys) {
     public Set<ConsumerActivationKey> getActivationKeys() {
         return this.activationKeys;
     }
+
+    public String getServiceType() {
+        return serviceType;
+    }
+
+    public void setServiceType(String serviceType) {
```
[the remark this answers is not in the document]
sonalidhome: Done.

server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
```diff
@@ -230,6 +231,14 @@ public SystemPurposeComplianceStatus getStatus(Consumer consumer,
                         status.addCompliantUsage(usage, entitlement);
                     }
                 }
+
+                if (StringUtils.isNotEmpty(prefferedServiceType) &&
+                    product.hasAttribute(Product.Attributes.SUPPORT_TYPE)) {
```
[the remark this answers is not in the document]
sonalidhome: Updated.

server/src/test/java/org/candlepin/resource/OwnerResourceTest.java
```diff
@@ -2101,6 +2108,13 @@ public void testReturnConsumerSysPurposeValuesForOwner() throws Exception {
             .get(SystemPurposeAttributeType.SERVICE_LEVEL.toString());
         Set<String> expectedSLAs = new HashSet<>(Arrays.asList("sla1", "sla2", "sla3"));
         assertEquals(expectedSLAs, serviceLevelAgreements);
+
+        Set<String> serviceType = result.getSystemPurposeAttributes()
+            .get(SystemPurposeAttributeType.SERVICE_TYPE.toString());
+
+        Set<String> expectedServiceType = new HashSet<>(Arrays.asList("serviceType1",
```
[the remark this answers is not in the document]
sonalidhome: Done.

server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
```diff
@@ -256,6 +265,10 @@ public SystemPurposeComplianceStatus getStatus(Consumer consumer,
             status.setNonCompliantUsage(consumer.getUsage());
         }
 
+        if (StringUtils.isNotEmpty(consumer.getServiceType()) && status.getCompliantServiceType().isEmpty()) {
```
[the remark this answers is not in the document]
sonalidhome: Updated the code with null check for map type object.

server/src/main/resources/rules/rules.js
```diff
@@ -517,7 +527,8 @@ function get_pool_priority(pool, consumer) {
         'roles': 2800,
         'addons': 1400,
         'support_level': 700,
-        'usage': 350
+        'usage': 350,
+        'support_type': 340
```
nikosmoum, line 531: Even though 340 is lower than the weight of usage (350), and higher than the next most powerful weight (150 for the requires_host attribute), and it will work for most use cases, it will not work for a number of use cases.

As the comment above explains, the relationship between each attribute's previous/next weight is important. More specifically: each weight should be larger than all the other attributes below it combined.
In this case this rule is broken because usage (350) will now be less than everything below it combined (service_type+requires_host+virt_only+sockets+ram+cores+vcpu: 670).

Here is an example use case that this will break (as an AutobindRulesTest.java unit test):
<details>
<summary>broken use case</summary>


```
    /*
     * This test demonstrates that a pool that satisfies the consumer's Usage should be selected
     * during autoattach even if all other attributes below it (service_type, sockets, cores & ram, etc.) mismatch,
     * instead of a pool that has a matching service_type, but no other mismatches.
     */
    @SuppressWarnings("checkstyle:localvariablename")
    @Test
    public void testShouldSelectPoolWhenUsageMatchesButServiceTypeSocketsRamCoresMismatch() {
        Product product69 = new Product();
        product69.setId("compliant-69");

        // Consumer specified attributes:
        consumer.setUsage("myusage");
        consumer.setServiceType("myservicetype");
        consumer.setFact("cpu.cpu_socket(s)", "1");
        consumer.setFact("cpu.core(s)_per_socket", "1");
        consumer.setFact("memory.memtotal", "9980456");
        ConsumerInstalledProduct consumerInstalledProduct =
            new ConsumerInstalledProduct(product69);
        consumer.addInstalledProduct(consumerInstalledProduct);

        // --- No satisfied syspurpose attributes on the consumer ---

        // Candidate pools:

        // This pool covers the system's usage attribute, but most other attributes below usage (service_type,
        // cores, ram, etc.) are mismatched.
        Product prodMCT1650 = createSysPurposeProduct(null, null, null,
            null, "myusage", "random_service_type");
        // Set large sockets/cores/ram values to this SKU, to make sure it's priority is reduced compared to
        // the other pool that uses the default priorities for these attributes (prioritization points for
        // cores/sockets/ram works this way because we do not want to to attach a pool that wastes extra
        // resource attributes if we can).
        prodMCT1650.setAttribute(Product.Attributes.SOCKETS, "32");
        prodMCT1650.setAttribute(Product.Attributes.CORES, "32");
        prodMCT1650.setAttribute(Product.Attributes.RAM, "19960912");
        prodMCT1650.addProvidedProduct(product69);
        Pool MCT1650 = TestUtil.createPool(owner, prodMCT1650);
        MCT1650.setId("MCT1650");
        MCT1650.setQuantity(1L);

        // This pool only covers the service_type of the system (and not the usage attribute), but has no
        // other mismatches compared to the previous pool.
        Product genericProduct = createSysPurposeProduct(null, null, null, null, null, "myservicetype");
        genericProduct.addProvidedProduct(product69);
        Pool genericPool = TestUtil.createPool(owner, genericProduct);
        genericPool.setId("genericPool");
        genericPool.setQuantity(1L);

        List<Pool> pools = new ArrayList<>();
        pools.add(MCT1650);
        pools.add(genericPool);

        List<PoolQuantity> bestPools = autobindRules.selectBestPools(consumer,
            new String[]{"compliant-69"}, pools, compliance, null, new HashSet<>(), false);

        assertEquals(1, bestPools.size());
        // We expect MCT1650 to be attached, because the fact that it covers the usage attribute, which is
        // more important than the service_type attribute, should be enough to prioritize it over
        // genericPool, even if other mismatches of attributes of lower importance are in play.
        assertTrue(bestPools.contains(new PoolQuantity(MCT1650, 1)));
    }

```
</details>

To solve this, I would suggest using 350 as the weight for the support_type attribute, and simply double the weight of every attribute above it. So like this:
```
        'products': 11200,
        'roles': 5600,
        'addons': 2800,
        'support_level': 1400,
        'usage': 700,
        'support_type': 350
```

nikosmoum: Also, if we are going to apply this change, we would also have to change the baseline priority that every pool starts with above from:
 `var priority = 545;` to `var priority =  1105;` (to make sure, as the comment above that explains, we never get priorities below zero, when a pool accrues too many mismatches that result in negative points).

And perhaps we need to change those 2 comments that explain the algorithm to reflect these changes.

nikosmoum (changes requested):
The code looks good, but the pool prioritization algorithm needs adjustment so that we will not be breaking some auto-attach edge cases (in this case, a usage match no longer being able to overpower mismatches of less important attributes)

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1338,4 +1338,24 @@ public int cullInvalidConsumerContentAccess(Owner owner, String... existingModes
...
+     * @param owner
+     *  The owner the consumers belong to.
+     * @return
+     *  A list of the all the distinct values of the service types that the consumers belonging to the
+     *  specified owner have set.
+     */
+    @SuppressWarnings("unchecked")
+    @Transactional
+    public List<String> getDistinctSyspurposeServiceTypeByOwner(Owner owner) {
+        return this.createSecureCriteria()
```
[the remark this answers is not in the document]
nikosmoum: I don't believe we really need the secure criteria. The API endpoint this is called from already has a `@Verify(Owner.class)` and that should be enough I think, so this could be re-written in non-secure JPA criteria.

Also, this is a good opportunity to merge the other 3 similar methods that fetch other syspurpose attributes (`getDistinctSyspurposeRolesByOwner`, `getDistinctSyspurposeUsageByOwner`, `getDistinctSyspurposeServicelevelByOwner`) with this one, into a single one that takes a parameter, to avoid duplicating code.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -2420,7 +2431,7 @@ public void testShouldSelectAllNonStackedPoolsThatProvideRoleAddonOrProductButNo
 
         // This pool provides the consumer's installed product, but does not provide his specified role.
         Product prodWithInstalledProductOnly = createSysPurposeProduct(null, null, null,
```
nikosmoum, line 2433: This test is a variation of the previous, but with the usage/sla/service_type pools that we expect to not be attached to have an installed product that the user does not specify on them, so we should do what we did in the previous test:
add service_type to the consumer, and also add a pool that provides that service_type and also a random installed product, and assert that it will not be attached

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -2502,7 +2513,7 @@ public void testShouldSelectPoolWhenSLAMatchesButOtherSysPurposeAttributesAndSoc
 
         // Candidate pools:
         Product prodMCT1650 = createSysPurposeProduct(null, "random_role", "random_addon",
-            "mysla", "random_usage");
+            "mysla", "random_usage", null);
```
nikosmoum, line 2516: This should be "random_service_type" instead of null, since the test expects all attributes other than SLA to have some random value

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -2552,7 +2563,7 @@ public void testShouldSelectPoolWhenRoleMatchesButOtherSysPurposeAttributesAndSo
 
         // Candidate pools:
         Product prodMCT1650 = createSysPurposeProduct(null, "myrole", "random_addon",
-            "random_sla", "random_usage");
+            "random_sla", "random_usage", null);
```
nikosmoum, line 2566: Same here: This should be "random_service_type" instead of null, since the test expects all attributes other than role to have some random value

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -2602,7 +2613,7 @@ public void testShouldSelectPoolWhenAddonsMatchButOtherSysPurposeAttributesAndSo
 
         // Candidate pools:
         Product prodMCT1650 = createSysPurposeProduct(null, "random_role", "myaddon",
-            "random_sla", "random_usage");
+            "random_sla", "random_usage", null);
```
nikosmoum, line 2616: Same here: This should be "random_service_type" instead of null, since the test expects all attributes other than addons to have some random value

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -2650,7 +2661,7 @@ public void testShouldSelectPoolWhenSLAMatchesButOtherSysPurposeAttributesAndVcp
 
         // Candidate pools:
         Product prodMCT1650 = createSysPurposeProduct(null, "random_role", "random_addon",
-            "mysla", "random_usage");
+            "mysla", "random_usage", null);
```
nikosmoum, line 2664: Same here: This should be "random_service_type" instead of null, since the test expects all attributes other than SLA to have some random value

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -2937,13 +2950,13 @@ public void testSelectBestPoolsUsageMatchShouldOverrideAllOtherSyspurposeMismatc
         // --- No satisfied syspurpose attributes on the consumer ---
 
         Product prod1 = createSysPurposeProduct(null, "mismatched_role", "mismatched_addon",
-            "mismatched_sla", "Production");
+            "mismatched_sla", "Production", null);
```
nikosmoum, line 2953: The consumer should have a service_type set, and this should be "random_service_type" (that mismatches with the consumer's)

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -3031,13 +3045,13 @@ public void testSelectBestPoolsAddonMatchShouldOverrideAllOtherSyspurposeMismatc
         // --- No satisfied syspurpose attributes on the consumer ---
 
         Product prod1 = createSysPurposeProduct(null, "mismatched_role", "my_addon",
-            "mismatched_sla", "mismatched_usage");
+            "mismatched_sla", "mismatched_usage", null);
```
nikosmoum, line 3048: The consumer should have a service_type set, and this should be "random_service_type" (that mismatches with the consumer's)

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -3078,13 +3092,13 @@ public void testSelectBestPoolsRoleMatchShouldOverrideAllOtherSyspurposeMismatch
         // --- No satisfied syspurpose attributes on the consumer ---
 
         Product prod1 = createSysPurposeProduct(null, "RHEL Workstation", "mismatched_addon",
-            "mismatched_sla", "mismatched_usage");
+            "mismatched_sla", "mismatched_usage", null);
```
nikosmoum, line 3095: The consumer should have a service_type set, and this should be "random_service_type" (that mismatches with the consumer's)

nikosmoum (changes requested):
Added some more comments around unit tests to make sure we fullproof the prioritization algorithm

server/src/main/resources/rules/rules.js
nikosmoum: Adding a comment here for posterity to note the outcome of more discussion between me and @sonalidhome around the adjustments to the algorithm: In addition to the above changes:
- We found out there is a special case around virt_only (see #2388) for the algorithm, whose weight (+100) need to be more powerfull than all the syspurpose attributes' null rule scores combined. The current changes break that, which lead to having to reduce the null rule modifier to half, from `0.01` to `0.005`.
- The current changes also break another goal of the algorithm, which is that: A given syspurpose attribute match should be more powerful than all of the other attribute mismatches combined (even for the mismatches of attributes that are of higher importance). In order to solve that, we decided to also reduce the mismatch rule modifier to half, from `-0.05` to `-0.025`.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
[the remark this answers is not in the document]
sonalidhome: Added a `getDistinctSyspurposeValuesByOwner(Owner owner, String sysPurposeAttribute)` method to fetch syspurpose attributes based on input parameter.

server/src/main/java/org/candlepin/policy/SystemPurposeComplianceRules.java
```diff
@@ -157,6 +157,7 @@ public SystemPurposeComplianceStatus getStatus(Consumer consumer,
             Set<String> unsatisfiedAddons = new HashSet<>(consumer.getAddOns());
             String preferredSla = consumer.getServiceLevel();
             String preferredUsage = consumer.getUsage();
+            String prefferedServiceType = consumer.getServiceType();
```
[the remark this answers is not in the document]
sonalidhome: Done.

server/src/main/resources/rules/rules.js
sonalidhome: Ack. Changes are updated.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -1600,7 +1611,7 @@ public void testSysPurposePoolPriorityShouldNotBeNegative()
 
         // Candidate pools:
         Product prod = createSysPurposeProduct(null, "mismatched_role", "mismatched_addon",
-            "mismatched_sla", "mismatched_usage");
```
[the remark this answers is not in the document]
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -4055,4 +4070,379 @@ public void selectBestPoolsMatchesAddonsValueWithWhiteSpaces() {
...
+        product69.setId("compliant-69");
+
+        // Consumer specified syspurpose attributes:
+        ConsumerInstalledProduct consumerInstalledProduct =
+            new ConsumerInstalledProduct(product69);
+        consumer.addInstalledProduct(consumerInstalledProduct);
+        consumer.setServiceType("Other ServiceType");
+
+        // Candidate pools:
+        Product prodMCT1650 = createSysPurposeProduct(null, "Smart ServiceType,Other ServiceType", null,
```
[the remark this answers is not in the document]
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
```diff
@@ -2339,36 +2350,36 @@ public void testShouldSelectAllNonStackedPoolsThatProvideRoleAddonOrProductButNo
...
         poolWithUsageOnly.setId("poolWithUsageOnly");
         poolWithUsageOnly.setQuantity(1L);
 
         // This pool does not provide the consumer's installed product, but provides his specified usage.
         Product prodWithSLAOnly = createSysPurposeProduct(null, null, null,
-            "Other SLA", null);
+            "Other SLA", null, null);
         Pool poolWithSLAOnly = TestUtil.createPool(owner, prodWithSLAOnly);
         poolWithSLAOnly.setId("poolWithSLAOnly");
         poolWithSLAOnly.setQuantity(1L);
```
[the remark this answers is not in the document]
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
sonalidhome: Done.

server/src/test/java/org/candlepin/policy/AutobindRulesTest.java
sonalidhome: Done.

server/src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1735,23 +1736,34 @@ public SystemPurposeAttributesDTO getSyspurpose(
...
+            dtoMap.get(SystemPurposeAttributeType.USAGE.toString()).addAll(consumerUsages);
+            dtoMap.get(SystemPurposeAttributeType.SERVICE_LEVEL.toString()).addAll(consumerSLAs);
+            dtoMap.get(SystemPurposeAttributeType.ADDONS.toString()).addAll(consumerAddons);
+            dtoMap.get(SystemPurposeAttributeType.SERVICE_TYPE.toString()).addAll(consumerServiceTypes);
+
+            dto.setOwner(translator.translate(owner, NestedOwnerDTO.class));
+            dto.setSystemPurposeAttributes(dtoMap);
+        }
+        catch (NotFoundException e) {
+            throw new NotFoundException("Unrecognized system purpose attribute.");
```
nikosmoum, line 1765: We shouldn't be throwing such an error on the API, because the user never specified any syspurpose attribute; we, the developer, hardcoded the list of attributes we want to gather. We should probably remove this block from here, and have consumerCurator.getDistinctSyspurposeValuesByOwner() throw a RuntimeException, which doesn't have to be checked here.

sonalidhome: Updated.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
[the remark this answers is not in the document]
nikosmoum: ack from me

server/src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1561,4 +1503,53 @@ public int cullInvalidConsumerContentAccess(Owner owner, String... existingModes
...
+     *
+     * @param owner
+     *  The owner the consumers belong to
+     * @param sysPurposeAttribute
+     *  The type of system purpose attribute needs to be fetched
+     * @return
+     *  A list of the all the distinct values of the system purpose attributes that the consumers
+     *  belonging to the specified owner have set
+     */
+    public List<String> getDistinctSyspurposeValuesByOwner(Owner owner, String sysPurposeAttribute)
```
Ceiu, line 1518: Since we have enums available for it, we should just use the SystemPurposeAttributeType enum instead of a string for the input here. We don't want to expose or couple the metamodel details to business logic, as can break things down the line.

Ceiu (changes requested):
Generally speaking, we should avoid using metamodel classes outside of query building. Doing so risks tying our business logic to the database structure, and could break things quite badly should the backend change in the future.

server/src/main/java/org/candlepin/model/ConsumerCurator.java
sonalidhome: Ack. Updated the code to use `SystemPurposeAttributeType`.

Ceiu (approved):

nikosmoum (approved):

PR #3036, opened 2021-05-14
[ENT-3228] Candlepin project restructuring

wolfdale:
Candlepin project restructuring

- [x] Removed commons & api sub-project from builds
- [x] Server package moved to the project root
- [x] Gradle build working
- [x] Maven build working
- [x] Vagrant deployment working
- [x] Minor openAPI clean-ups 
- [x] Cleanup dependencies in both gradle & maven
- [x] Fix all broken scripts
- [x] Verify artifacts are bundled correctly
- [x] Verify/Fix tito builds & packaging
- [x] Verify/Fix Docker builds
- [x] Update candlepin-roles project [candlepin/ansible-role-candlepin/pull/17]
- [x] Update project Readme files 
- [x] Test with Jenkin & custom docker images by pushing it to Quay
- [x] Rebase with master branch
- [x] Final Cleanups
- [x] Update release pipeline

 ## Project structure

> .
| ── api                                                               
|   ├── candlepin-api-config.json
|   └── candlepin-api-spec.yaml
|
|── bin                                                                  
|   ├── deployment
|   └── scripts
|   
|── buildSrc                                                         
|   ├── build.gradle
|   └── src
|
|── client                                                                   
|   └── ruby
|
|── config                                                                   
|   ├── candlepin
|   └── checkstyle
|── src                                                                   
|   ├── main
|   └── test
|
|── docker   
|── jenkins   
|── po          
|── selinux 
|── spec      
|── vagrant  

Tested this branch against custom docker image that support OLD and NEW project structure. ```candlepin/docker/candlepin-base/cp-test``` is updated to support this.


Ceiu (approved):

server/src/test/resources/version.properties
```diff
@@ -1,2 +0,0 @@
-version=${version}
-release=${release}
```
nikosmoum, line 2: The contents of this were deleted, but the file was left? Is the file needed at all? what was it's original purpose anyway?

po/keys.pot
```diff
@@ -17,124 +17,6 @@ msgstr ""
 "Content-Type: text/plain; charset=CHARSET\n"
 "Content-Transfer-Encoding: 8bit\n"
 
-#: common/src/main/java/org/candlepin/common/exceptions/mappers/BadRequestExceptionMapper.java:72
```
nikosmoum, line 20: something to note here: The `master` component's configuration in Weblate will have to be changed to find out where the .po files live now (since common/ no longer exists)

build.gradle
```diff
@@ -371,94 +491,6 @@ project(":api") {
         withXml = false
     }
 
-    task pom {
```
nikosmoum, line 374: I don't see this task being moved elsewhere. Does this mean the pom.xml is no longer auto-generated?
What was the original reason for auto-generating the pom? One of the benefits of having that was that the developers did not have to remember to update both gradle and maven when they added/updated/removed dependencies, since the release pipeline would run the pom task to generate it. With this gone, we would be going back to release pipeline failures due to missing maven dependencies.

nikosmoum (changes requested):
Some questions/issues I see. I'll have another look next week, since I didn't manage to get through everything yet. Nevertheless, awesome job; I can't wait for a simplified structure :+1:

src/test/java/org/candlepin/controller/refresher/mappers/NodeMapperTest.java
```diff
@@ -12,14 +12,13 @@
  * granted to use or replicate Red Hat trademarks that are incorporated
  * in this software or its documentation.
  */
-package org.candlepin.controller.refresher.builders;
```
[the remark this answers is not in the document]
nikosmoum: (not an issue btw, just me being surprised)

build.gradle
wolfdale: Earlier there were multiple POM files. ```pom``` task was there to update these file so we don't have to go & update every file manually. Now I've merged all dependencies into single pom file. If we want to add/update/remove any dependency we just need to update at two places ie. build.gradle & pom.xml. Also sooner or later we are going to deprecate maven all together. So I figured just to remove all maven stuff from build.gradle & keep both build tools isolated from each other.

po/keys.pot
wolfdale: Noted. 
I'll keep this comment open. Until changes are made.

server/src/test/resources/version.properties
wolfdale: Both files (& its content) are still present in new structure. 
We refer these files within source code for checking compatibility.

wolfdale:
Re-based with main branch, PR up-to-date for review.

wolfdale:
Re-based with main branch, PR is up-to-date.

build.gradle
nikosmoum: Deprecating maven is a long way out yet, and unfortunately, remembering to change even a single pom file when a corresponding gradle change was made was hard (which was the reason we introduced a special check in the release pipeline to see if either only the build.gradle or pom.xml was changed; and then we removed this special check specifically because introduced the gradle pom command instead) and will happen multiple times guaranteed if we remove this, as it did in the past.

Running a candlepin release only to find out it errored out in the build step, which is way too late in the pipeline, and then causing us to post a PR with the pom.xml fix, and re-running the release from scratch, which then would produce a build by skipping a version because the previous run already tagged a .z version, was very frustrating. We have 2 options here:
1. Re-introduce the check for build.gradle/pom.xml discrepancies early in the release pipeline.
2. Add back the gradlew pom command.
3. Introduce a new jenkins PR pipeline job that checks for build.gradle/pom.xml discrepancies.

I don't like #1, because we still need to create a PR with the pom.xml fix every time someone forgot to do those changes. I'm fine with either #2 or #3, but for #3 we would have to make sure it's introduced before we run a 4.1.z release (if not before this PR is merged).

nikosmoum: I wonder what others think is the best course of action here: @Januson @wolfdale @sonalidhome @Ceiu

build.gradle
```diff
@@ -23,7 +23,58 @@ plugins {
...
     id "com.github.ben-manes.versions" version "0.28.0"
     id "org.candlepin.gradle.SpecVersion"
-    id "java"
+}
+
+apply plugin: "war"
+apply plugin: "checkstyle"
+apply plugin: Gettext
+apply plugin: Msgfmt
+apply plugin: Gettext
```
nikosmoum, line 32: Gettext plugin applied twice

nikosmoum (changes requested):
A couple additional minor things

build.gradle
nikosmoum: @wolfdale You could bring this up as a post-scrum to discuss

docker/candlepin-base/cp-test
```diff
@@ -288,10 +286,27 @@ else
...
+# Candlpin project restructuring
+# We need to correctly set the project home path.
+# We define CP_NEW_STRUCTURE boolean to check if
+# we are running with new candlepin project structure.
+CP_NEW_STRUCTURE=true
+PROJECT_DIR="."
+echo "Current branch : $CHECKOUT"
+echo "Target branch : $TARGET"
+
+if [ "$TARGET" != "master" ]; then
```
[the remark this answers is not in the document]
nikosmoum: I just run a 4.1.1 release, so we should probably use 4.1.2 here instead

buildSrc/src/main/java/org/candlepin/gradle/SpecVersion.java
```diff
@@ -28,9 +28,9 @@ public void apply(Project project) throws GradleScriptException {
         String versionFromSpec = null;
         String releaseFromSpec = null;
 
-        // Read the spec file from (projectRoot)/server/candlepin.spec.tmpl
```
nikosmoum, line 31: One more question I just realized: The buildSrc/build.gradle file was left untouched. Is there no way to merge that functionality into the main build.gradle file? (it's not a big deal if we do leave it there, but I am just curious if that was accidentally missed, or something is blocking us from removing it)

build.gradle
wolfdale: Removed.

buildSrc/src/main/java/org/candlepin/gradle/SpecVersion.java
wolfdale: Actually ```buildSrc/build.gradle``` is for separate build whose purpose is to build any tasks, plugins, or other classes which are to be used in main build. We use that for extending openapi, gettext, msgfmt plugin etc, that need to be compile before hand.

nikosmoum: ack

build.gradle
wolfdale: As discussed, added pom task again.
 
@nikosmoum Since we removed the whole parent-child hierarchy in POM files, we need to make sure every tag (such as profile, repository, scm, license tags in POM file) is generated by ```pom``` task. I still figuring out how to generate [this](https://github.com/candlepin/candlepin/blob/master/pom.xml#L46) via gradle maven plugin.

nikosmoum: Yes, I think the pom task only used to generate the children poms, not the parent pom.xml

docker/candlepin-base/cp-test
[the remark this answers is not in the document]
wolfdale: Since we fetch the version from ```candlepin.spec.tmpl``` file. It will be better if we check that this file exists on the root or not.
Something like - 
```
if [ ! -f candlepin.spec.tmpl ]; 
   Old Structure
else
    New Structure
```

build.gradle
wolfdale: Added the gradle pom task. 
We can safely remove the ```activeByDefault``` tag from POM file, since maven profiles are active by default, as we have only one profile defined, we don't have to worry about it.

build.gradle
```diff
@@ -540,5 +578,354 @@ wrapper {
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }.withXml {
+            dependencyManagement.pomConfigurer.configurePom(asNode())
+            asNode().dependencies.'*'.findAll() {
```
wolfdale, line 924: This is done because Gradle maven plugin looks for dependency & dependency management tag in ```build.gradle``` file to generate the pom file. In Gradle we have two logical projects (api & candlepin) which we build/compile from single ```build.gradle``` file. In order to auto-generate classes via openapi-generator we have included ```compile project(":api")``` in dependency of main Gradle build. It make sure auto-generated class are available for main build beforehand.

Since :api project is added as a dependency, gradle pom task pick this as a dependency & add that to POM.xml file. We need to remove this dependency from auto-generated pom file.

nikosmoum: ack

pom.xml
```diff
@@ -1,164 +1,1087 @@
...
+                <sourceFolder>src/gen/java/main</sourceFolder>
+              </configOptions>
+              <templateDirectory>${project.basedir}/buildSrc/src/main/resources/templates</templateDirectory>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.bsc.maven</groupId>
+        <artifactId>maven-processor-plugin</artifactId>
```
nikosmoum, line 496: A question on these 2 plugins: maven-processor-plugin and build-helper-maven-plugin. They didn't seem to be needed beforehand; why are they added now? How where the jpamodel generated-sources added to the war during a maven build before?

build.gradle
nikosmoum: Let's explain this in a short comment right above this code

nikosmoum:
I have just a last question, and a minor comment to be added. After that, a rebase and a squash, this should be good to merge

build.gradle
wolfdale: Done.

pom.xml
wolfdale: Since we are using open-api & jpamodel generators, both generate source code to ```/target/generated-sources``` directory, which creates conflicts. We need to explicitly specify where to keep the generated code (by specifying exact output directory). ```maven-processor-plugin``` (& ```openapi-generator-maven-plugin```) plugin makes job easier by specifying output directory.

```build-helper-maven-plugin``` make sure that multiple source directories (auto-generated & main code) are available during compilation.
The old candlepin project structure, every sub-project (server, common, api) have specific POM file, which does not create any conflict while auto-generating code since each project have their own ```target/generated-sources``` directory.

wolfdale:
Before merge, We need to push docker image (with cp-test changes) to registry. Expectation is, this PR will turn green.

nikosmoum (approved):

PR #3037, opened 2021-05-17
[M] Print rspec results to file

nikosmoum:
- The gradle rspec command will now write the output to
  server/build/reports/tests/rspec/index.html by default
- Use '--no-file' to prevent the file from being generated
- When run in the container that Jenkins uses, it will not
  generate the file by default

### DIFF
--- a/buildSrc/src/main/groovy/Rspec.groovy
+++ b/buildSrc/src/main/groovy/Rspec.groovy
@@ -32,6 +32,9 @@ class Rspec extends DefaultTask {
     @Input
     String spec = ""
 
+    @Input
+    boolean fileOutput = true
+
     @Option(option = 'spec', description = 'The name of the spec file to search. Example: "core_and_ram"')
     void setSpec(final String specfile_name) {
         this.spec = specfile_name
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/groovy/Rspec.groovy
```diff
@@ -52,6 +60,14 @@ class Rspec extends DefaultTask {
                 "-I", "${project.getProjectDir()}/client/ruby"
         ]
 
+        if (fileOutput) {
+            rspec_args.add("--format")
+            rspec_args.add("h")
+            rspec_args.add("--force-color")
+            rspec_args.add("--out")
+            rspec_args.add("rspec_output.html")
```
Ceiu, line 68: Should probably drop this in build/reports/tests/rspec/index.html following along with what junit (attempts?) to do.

Ceiu (changes requested):

buildSrc/src/main/groovy/Rspec.groovy
nikosmoum: Done

Ceiu (approved):

wolfdale:
@nikosmoum this is a breaking change for branches other than ```master```. 
Should we port this to all other HOTFIX branches ?
While testing candlepin docker image (with branches other than master) I found -
```
* What went wrong:
Problem configuring task :candlepin:rspec from command line.
> Unknown command-line option '--no-file'.
```

nikosmoum:
@wolfdale ack, I'll port this back to other banches

PR #3043, opened 2021-05-20
ENT-3811: Trim whitespace in addons and roles

Januson:
- Extracted utillity method Util.toList
- Reused toList for other places where we handled whitespace
- Updated rules to handle whitespace when splitting addons and roles

### DIFF
--- a/server/src/main/resources/rules/rules.js
+++ b/server/src/main/resources/rules/rules.js
@@ -1,4 +1,4 @@
-// Version: 5.41
+// Version: 5.42
 
 /*
  * Default Candlepin rule set.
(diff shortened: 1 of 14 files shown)


server/src/main/resources/rules/rules.js
```diff
@@ -308,7 +308,7 @@ function createPool(pool, consumer) {
         }
         else {
             if (attribute === 'addons' || attribute === 'roles') {
-                poolSet = this.getProductAttribute(attribute).split(',');
+                poolSet = this.getProductAttribute(attribute).trim().split(/\s*,\s*/);
```
Ceiu, line 311: We should probably use the same regex here that also parses out empty/whitespaced comma pairs which we're using above.

Ceiu (changes requested):
This is a good standardization/cleanup. I only saw one thing that could use an update.

server/src/main/resources/rules/rules.js
Januson: Done. Updated to handle whitespaced commas.

Ceiu (approved):

PR #3051, opened 2021-05-27
[4.0] ENT-3911: 1964106: derivedProvidedProducts is now null where it was previously [] 

sonalidhome:
- This was changed while moving derived product from pool to product feature.
   Updated code to return the `Collections.emptySet()` instead of null.
 - Added spec test around this to test the required o/p.

### DIFF
--- a/server/spec/subscription_resource_spec.rb
+++ b/server/spec/subscription_resource_spec.rb
@@ -48,4 +48,32 @@ describe 'Subscription Resource' do
           e.to_s.eql? "bad URI(is not URI?): pools/{pool_id}/cert"
       end
   end
+
+  it 'subscriptions derived/provided products should return empty array when null' do
+      # Product does not have derived/provided products
+      pool = @cp.create_pool(@owner['key'], @some_product.id, { :quantity => 2, :subscriptionId => "test-subscription1" })
+      sublist = @cp.list_subscriptions(@owner['key'])
+      expect(sublist[0]['providedProducts']).to eq([])
+      expect(sublist[0]['derivedProvidedProducts']).to eq([])
+      @cp.delete_pool(pool.id)
+
+      # Product provided products only
+      provided_product = create_product(random_string('provided_product'))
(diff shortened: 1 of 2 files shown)


server/spec/subscription_resource_spec.rb
```diff
@@ -48,4 +48,42 @@
           e.to_s.eql? "bad URI(is not URI?): pools/{pool_id}/cert"
       end
   end
+
+
+  it 'subscriptions derived/provided products should return empty array when null' do
```
wolfdale, line 53: Indentation needs to be fixed. It's uneven at multiple places. Same for #3052 PR.

wolfdale (changes requested):
Changes looks good, need some minor fix up with indentation.

server/spec/subscription_resource_spec.rb
sonalidhome: Done.

wolfdale (approved):

Ceiu (approved):

server/spec/subscription_resource_spec.rb
```diff
@@ -48,4 +48,37 @@
...
+
+      # Product does not have derived/provided products
+      pool = create_pool_and_subscription(@owner['key'], @some_product.id, 2,
```
nikosmoum, line 56: Please don't use the deprecated methods `create_pool_and_subscription` and `delete_pool_and_subscription`.
Instead, use either:
a) `@cp.create_pool()` (and add the subscription_id / upstream_pool_id parameters when you need to simulate the fact that the pool was the result of a manifest import), and `@cp.delete_pool` when you want the test to not be specifically tied to hosted or standalone mode (and run on both), or
b) `create_upstream_subscription` and  `delete_upstream_subscription`, both followed by a call to org_refresh.

Also, this will probably solve the current spec test failures in hosted builds.

nikosmoum (changes requested):

server/spec/subscription_resource_spec.rb
sonalidhome: Ack. Updated code to use new methods.

PR #3052, opened 2021-05-27
[M] ENT-3911: 1965208: derivedProvidedProducts is now null where it was previously []

sonalidhome:
- In OpenAPI, autogenerated classes have collections initializes
   to null by default. It is found in the regression that it affects
   the Katello.
   Updated the `PoolToSubscriptionTranslator` to populate the
   `providedProducts` and `derivedProvidedProducts` with empty set.
   And it will get replaced with the appropritate values later.
 - Added spec test around this to test the required o/p.

### DIFF
--- a/server/spec/subscription_resource_spec.rb
+++ b/server/spec/subscription_resource_spec.rb
@@ -37,4 +37,33 @@ describe 'Subscription Resource' do
 
       @cp.activate_subscription(consumer.uuid, "mail", "locale")
   end
+
+  it 'subscriptions derived/provided products should return empty array when null' do
+
+      # Product does not have derived/provided products
+      pool = @cp.create_pool(@owner['key'], @some_product.id, { :quantity => 2, :subscriptionId => "test-subscription1" })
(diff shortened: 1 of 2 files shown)


wolfdale (approved):

server/spec/subscription_resource_spec.rb
```diff
@@ -37,4 +37,36 @@
 
       @cp.activate_subscription(consumer.uuid, "mail", "locale")
   end
+
+  it 'subscriptions derived/provided products should return empty array when null' do
+
+      # Product does not have derived/provided products
+      pool = create_pool_and_subscription(@owner['key'], @some_product.id, 2,
```
nikosmoum, line 44: Same as the 4.0 PR: Please use the non-deprecated methods for creating and deleting pools

nikosmoum (changes requested):

server/spec/subscription_resource_spec.rb
sonalidhome: ack. It worked now.

PR #3053, opened 2021-06-01
[M] Updated API documentation for the listAll param of /owners/{key}/pools (ENT-3792)

Ceiu:


api/candlepin-api-spec.yaml
```diff
@@ -5596,7 +5596,9 @@ paths:
             type: string
         - name: listall
           in: query
-          description: Include pools that are not suited to the unit's facts
+          description: |
+            Includes pools which are available to the consumer, but not applicable based on the consumer's
+            current system facts.
```
nikosmoum, line 5601: The meaning of the description stayed the same. Perhaps the jira card description was not clear, but we should convey to the user that `listAll` is a misnomer, since setting it to `true` might stop this endpoint from filtering pools based on _some_ facts (such as: architecture, sockets, cores, ram, storage_band, instance_multiplier), but it will _still_ do a LOT of pool filtering based on other facts such as: 
- Filter derived pools if the consumer is a manifest.
- Filter unmapped guest pools if the consumer is not a guest.
- Filter unmapped guest pools if the consumer is not newly-created (and therefore ineligible).
- Filter pools that requires_host if the consumer is a manifest.
- Filter pools that requires_host if the consumer is not a guest.
- Filter pools that requires_host if the consumer is a guest but it's host does not match with the pool's requires_host value.
- Filter pools that requires_consumer if the consumer is a manifest.
- Filter pools that requires_consumer if the consumer is a guest but it's uuid does not match with the pool's requires_consumer value.
- Filter pools that requires_consumer_type if the consumer's type does not match the pool's requires_consumer_type value.
- Filter pools that are 'restrictedToUsername', and the consumer's username does not match the pool's restrictedToUsername value.
- Filter pools that the consumer is already attached to, and they are not multi-entitlement pools.
- Filter pools that are expired.

Thus, `listSomeMore` would be a more applicable name, but since we cannot change the API, we should come up with a description that clarifies what setting it to true will do. Something like this might be better?:
`Includes pools which are not applicable based on some of the consumer's current system facts (such as architecture, sockets, cores, ram, storage_band, instance_multiplier), but will still filter pools based on other factors (such as pools restricted to a consumer with a certain type, username, uuid; pools restricted to guests with a certain host; non multi-entitlement pools that the consumer has already attached; unmapped guest pools that the consumer is ineligible for; and expired pools.`

nikosmoum (changes requested):

api/candlepin-api-spec.yaml
Ceiu: None of what you listed are filtered on the system facts, though.

Regardless, I've updated the block with the paragraph provided and made a few grammatical changes.

nikosmoum: Some of those do have to do with the consumer being a guest (which is reported as the virt.is_guest fact), but yes, most are not fact related. Thanks

PR #3058, opened 2021-06-04
[M] Added recalculation logic for vCPU counts on x86 guest systems (ENT-3926)

Ceiu:
- Add detection for changes to certain CPU facts to determine whether
  or not a guest's core count needs to be recalculated with the
  assumption SMT is actually enabled when it appears to be disabled
- Fixed a bug with the cloud profile fact change detection that would
  fail to detect the removal of a cloud profile fact


Ceiu:
This PR is awaiting more feedback from PMs on the direction we want to go with regard to hardcoding a division by 2 under the assumption that the target system is on a hypervisor using SMT2 but reporting no SMT.

Additionally, something about the way this ended up being implemented with the fact checking and FactEnum stuff just isn't sitting right with me, but I can't put my finger on exactly why. I'm very open to suggestions on how this was done or how it could be done, as well as how we should organize facts in cases like this.

Ceiu:
After giving it some thought, I've figured out what I dislike about the current fact checking code: it splits the facts (and logic) across a number of enums/interfaces, which will eventually lead to badness.

Next week I'll change this as follows:
- Remove the FactEnum interface
- Remove the CloudProfileFacts enum
- Remove the CPUFacts enum
- Add a new enum to the Consumer class named Facts which encapsulates all of the current facts Candlepin references in code (similar to Product.Attributes and Pool.Attributes).
- Add static functions to Consumer.Facts for fetching classes of facts, such as getCloudProfileFacts and getCPUFacts. These will serve as replacements for the currently disparate enums.

This will allow the fact checking logic to stay as it is currently, but centralizes the facts into a single enum and lifts the jank where all of the fact enums have to implement the FactEnum interface so we can treat them the same.

The final Consumer.Facts enum will look something like the following:
```
public static enum Facts {
    CPU_COUNT("cpu.cpu(s)"),
    ...
    
    public Facts(String fact) {
        ...
    }

    public String getFact() { // maybe rename this to name() or key() to be less weird-looking in actual usage
        return this.fact;
    }

    public static Facts[] getCloudProfileFacts() {
        return new Facts[] { /* facts currently specified in the CloudProfileFacts enum */ };
    }

    public static Facts[] getCPUFacts() {
        return new Facts[] { /* facts currently specified in the CPUFacts enum */ };
    }
}
```

nikosmoum (approved):
ack for now

nikosmoum:
Closing this since a new approach/design is needed for implementing this, and also there are conflicts that are easier solved by creating a fresh PR.

PR #3059, opened 2021-06-07
[4.0] Updated refresher to avoid fetching unnecessary versioned entities (ENT-3845)

Ceiu:
- Updated the refresher infrastructure to add a 3rd processing stage
  which performs versioned entity resolution and persistence
  operations to avoid needing to prefetch unnecessary versioned
  entities and massively reducing the memory footprint of refresh
  and manifest import
- Optimized versioned entity queries
- Updated the way nodes store links to other nodes to avoid needing
  to pass through entity mappers to visitors
- Rewrote several test suites to address the above changes and to
  improve test stability


sonalidhome (approved):
Ack.

PR #3063, opened 2021-06-14
[M] ENT-3228; Moved classes of common subpackages to appropriate places

nikosmoum:
- The initial restructuring moved the common module as a package
  under org.candlepin, with it's contents intact. This change
  is removing the common package entirely and moving all the classes
  under it to the appropriate top-level packages. Some examples:
  common/exceptions/ -> exceptions/
  common/auth/ -> auth/
  common/guice/ -> guice/
  etc


Ceiu:
The copyright block got kinda busted during what I assume was automated replacement. Looks like the initial Javadoc style double-asterisk got removed, and everything is indented one space further.

Ceiu (commented):
For the most part, the only thing I would like to see fixed here is the copyright thing -- either standardizing on the new format globally or fixing the new ones which are different.

I did have one question on the packaging choice for some of the RE classes: AuthUtil could probably go in resteasy.filter, since we have lots of auth stuff there already; or maybe the auth specific filters (might be all of them?) could move to the auth package. I see you did this for at least some of the Jackson filters, but I didn't see the same done here (though github literally isn't displaying the full file list for this PR, so I could have missed it).

Nothing major enough to hold up the PR, since a minor repackaging is easy to do later, but just something I was curious on.


Looking back through this, th

nikosmoum:
@Ceiu Reverted the copyright formatting changes, moved AuthUtil and RestEasyOAuthMessage to the resteasy.filter package, fixed checkstyle import order issues, and fixed test class compilation error.

Ceiu (approved):

PR #3064, opened 2021-06-15
Fix for Gettext plugin extension

wolfdale:
Changes -
- Removed the ```-j``` aka ```--join-existing``` flag. This was relevant for older project structure where we have to join the results of ```gettext``` command from ```server``` & ```common``` sub-projects.    
- Removed obsolete reference to /common folder.


nikosmoum (approved):
ACK

PR #3065, opened 2021-06-15
[4.0] 1949980: Update SCA to cache certs by consumer (ENT-3966)

Ceiu:
- Drop the old caching table


Ceiu:
Cherry-pick of #3002 to the 4.0 branch. There were some shenanigans involved in the merge, so at least a cursory glance over most areas would be prudent here.

Ceiu:
Added the missing bits from the SCA cert removal on unregistration fix. I opted not to explicitly cherry-pick that PR, as something about the merge conflict resolution brought in a whole bunch of seemingly random changes, and I didn't feel like fixing a bunch of misc. things or risking missing breaking something else. Instead, I manually reproduced the work from the fix and added it as a new commit on top of the per-consumer cert caching commit.

server/src/main/resources/db/changelog/20210412145432-drop_content_cache.xml
```diff
@@ -0,0 +1,26 @@
...
+    </changeSet>
+
+    <changeSet id="20210412145432-2" author="ojanus" dbms="mysql,mariadb">
+        <sql>UPDATE cp_cert_serial s RIGHT JOIN cp_cont_access_cert c ON c.serial_id = s.id SET expiration = NOW()</sql>
+    </changeSet>
+
+    <changeSet id="20210412145432-3" author="ojanus">
+        <addColumn tableName="cp_cont_access_cert">
+            <column name="content" type="LONGTEXT">
+                <constraints nullable="false"/>
```
nikosmoum, line 20: While the port looks good, I thought I'd give the db upgrade a test, and found out there's a problem. I've:
1. Set up candlepin with test data
2. Switched one of the orgs to use SCA mode
3. Registered a consumer, so as to generate an SCA cert+serial
4. Checked out this branch and tried to deploy 
Got this error:
```
Updating Database
Liquibase update Failed: Migration failed for change set db/changelog/20210412145432-drop_content_cache.xml::20210412145432-3::ojanus:
     Reason: liquibase.exception.DatabaseException: Error executing SQL ALTER TABLE public.cp_cont_access_cert ADD content TEXT NOT NULL: ERROR: column "content" contains null values
```
It looks like when we're adding non-nullable columns, we need to define a default value for the existing rows in the table. Adding an empty string should do the job, since all sca certs are set to be expired, and next check-in will force the regeneration of their content payloads.

nikosmoum: Needless to say, we need to fix this also in 3.1 and master

Ceiu: That sounds like something that needs to be fixed across all the branches, not just in the backport.

PR #3066, opened 2021-06-15
[3.1] Backport SCA per-consumer caching (ENT-3966)

Ceiu:
1949698: Update SCA to cache certs by consumer
- Drop the old caching table
- Only add content from active pools on the SCA certificate


Ceiu:
This PR is the backported work from #3002 and #2799. Most of the code from these two PRs is as-is, with very few exceptions. However, given the nature of the complexity of backporting and merging two commits, at least a cursory glance over the core functionality of both commits should be performed to ensure the logic is as intended.

nikosmoum (approved):

PR #3067, opened 2021-06-16
[M] Removed non-nullable restriction from the SCA cert cache table

Ceiu:
- The cp_cont_access_cert table no longer requires the content column
  to be non-null to address a potential problem with the migration
  of an already-populated table


Ceiu:
The solution implemented here is definitely up for debate -- an equally valid option would be to retain the non-null constraint but have a default value for the column definition. Both options work and both break in the same case: the empty/null value is present and the cert is passed to a subman client. In such a case, the client will break with a ```'NoneType' object has no attribute 'name'
``` error.

PR #3070, opened 2021-06-24
[M] Stop docker logout in Jenkins pipeline stages

nikosmoum:


nikosmoum:
The reason for this change is to overcome the very frequent 'ERROR: for candlepin  unauthorized: access to the requested resource is not authorized' we get in the various pipeline stages. It seems that one stage logging out is preventing other stages to pull container images from quay.io. Logging in multiple times in sequence seems to work just fine, so not logging out should work.

sonalidhome (approved):

PR #3001, opened 2021-04-15
ENT-3218: Setting up SonarQube for Candlepin

sonalidhome:
- Updated the dependencies for SonarQube
 - Added a Jenkins job to generate a report on a sonarqube server

PR #3076, opened 2021-06-30
Update mysql.cnf file owner to be readable in mariadb container

sonalidhome:
This is to investigate/fix the misbehavior of the spec test
in the environment created for #3001
    
- Added below options to get the isolation level
  -N : Don't write column names in results.
  -s : This option results in nontabular output
  format and escaping of special characters.
- Updated the mysql.cnf file permission and
  the file owner to MySQL user and group.

### DIFF
--- a/bin/deployment/bash_functions
+++ b/bin/deployment/bash_functions
@@ -166,10 +166,9 @@ init_mysql_jdbc() {
     # JDBC_DRIVER="org.mariadb.jdbc.Driver"
     # JDBC_JAR="/usr/lib/java/mariadb-java-client.jar"
     # JDBC_URL="jdbc:mariadb://$db_host/$db_name"
-
-    local isolation_level=$(mysql -h $db_host $db_name  --user=$db_user --password="$password"  -e 'select @@global.tx_isolation')
-    local READ_COMMITTED="READ-COMMITTED"
-    if [[ $isolation_level != *$READ_COMMITTED* ]]; then
+    local isolation_level=$(mysql -h $db_host $db_name  --user=$db_user --password="$password" -s -N -e 'select @@global.tx_isolation')
+    echo "isolation_level : $isolation_level"
+    if [[ $isolation_level != "READ-COMMITTED" ]]; then
         err_msg "WARNING: candlepin requires READ-COMMITTED isolation level"
     fi
 }
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -4,7 +4,7 @@ library identifier: 'fh-pipeline-library@master', retriever: modernSCM(
    credentialsId: 'github-api-token-as-username-password'])
 
 pipeline {
-    agent { label 'docker' }
+    agent { label 'candlepin' }
     options {
         skipDefaultCheckout true
         timeout(time: 16, unit: 'HOURS')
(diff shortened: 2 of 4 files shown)


spec/consumer_resource_spec.rb
```diff
@@ -1519,20 +1519,23 @@ def verify_installed_pids(consumer, product_ids)
...
-          @user1.unregister(consumer['uuid'])
-        rescue RestClient::ResourceNotFound => rnf
-          # Expected and OK.
-          expected_exceptions << rnf
-        rescue RestClient::Gone => gone
-          expected_exceptions << gone
-        rescue Exception => e
-          # Unexpected - report it.
-          unexpected_exceptions << e
+        lock.synchronize do
```
nikosmoum, line 1525: This test (and the other ones you have edited in the PR), seems to want to run the threads in parallel by design, in order to test what happens on the server side when concurrent requests happen. If you use a lock these blocks of code, you essentially only let one thread execute this at a time, which defeats the purpose of the test.

I would focus on finding out what changed in the infrastructure first (some examples: perhaps Ruby version/interpreter changed that now uses true multithreading (see regarding the Global Interpreter Lock [here](https://www.toptal.com/ruby/ruby-concurrency-and-parallelism-a-practical-primer), while before it wasn't? perhaps there's difference in how containers work now that our infrastructure moved to fedora 30 (maybe the docker/podman version used has different multithreading support)? 

If we find out that indeed we were not actually running these tests in parallel, we might actually need to fix the way the server responds when _truly concurrent_ requests like these come in.

bin/deployment/bash_functions
```diff
@@ -166,10 +166,10 @@ init_mysql_jdbc() {
     # JDBC_DRIVER="org.mariadb.jdbc.Driver"
     # JDBC_JAR="/usr/lib/java/mariadb-java-client.jar"
     # JDBC_URL="jdbc:mariadb://$db_host/$db_name"
-
-    local isolation_level=$(mysql -h $db_host $db_name  --user=$db_user --password="$password"  -e 'select @@global.tx_isolation')
+    local isolation_level=$(mysql -h $db_host $db_name  --user=$db_user --password="$password" -s -N -e 'select @@global.tx_isolation')
+    echo "isolation_level : $isolation_level"
     local READ_COMMITTED="READ-COMMITTED"
```
nikosmoum, line 171: since we're not using this variable anymore, let's remove it

jenkins/Jenkinsfile
```diff
@@ -4,7 +4,7 @@ library identifier: 'fh-pipeline-library@master', retriever: modernSCM(
    credentialsId: 'github-api-token-as-username-password'])
 
 pipeline {
-    agent { label 'docker' }
+    agent { label 'jmolet-vm0' }
```
nikosmoum, line 7: Let's revert the changes in this file. We can have the infrastructure changes merged first, and then re-run jenkins on this PR to make sure it's passing before merging it.

nikosmoum (changes requested):
Let's squash with a more approriate commit message, and update the PR title too.
And once @Lorquas merges his infrastructure changes we'll re-run the pipeline for this PR to make sure it passes before we merge this

jenkins/Jenkinsfile
Lorquas: The infra has been merged and deployed. The new nodes will be called "candlepin-%". This file can be completely reverted.  You may continue to use the 'docker' agent label or the 'candlepin' label.

PR #3077, opened 2021-07-01
1800811: Pool quantity should never go below -1 (ENT-3974)

wolfdale:
Changes :
- All unlimited pools are restricted to -1 quantity.
- Unmapped guest & Bonus pools specifically to have unlimited quantity in both cases: a) the master pool quantity is unlimited, b) virt_limit is unlimited

### DIFF
--- a/server/spec/pool_unlimited_master_spec.rb
+++ b/server/spec/pool_unlimited_master_spec.rb
@@ -51,10 +51,28 @@ describe 'Unlimited Master Pools' do
       }
     })
 
+    @product_virt_instance_multiplier = create_product(nil, nil, {
+      :attributes => {
+        :virt_limit => "8",
+        'multi-entitlement' => "yes",
+        "instance_multiplier" => 6
+      }
+    })
+
+    @product_virt_product_multiplier = create_product(nil, nil, {
+      :attributes => {
+        'multi-entitlement' => "yes",
+        :virt_limit => "8",
+      },
+      :multiplier => 100
+    })
+
     @pool_no_virt = create_pool_and_subscription(@owner['key'], @product_no_virt.id, -1, [], '', '', '', nil, nil, false)
     @pool_unlimited_virt = create_pool_and_subscription(@owner['key'], @product_unlimited_virt.id, -1, [], '', '', '', nil, nil, false)
     @pool_virt = create_pool_and_subscription(@owner['key'], @product_virt.id, -1, [], '', '', '', nil, nil, false)
     @pool_virt_host_dep = create_pool_and_subscription(@owner['key'], @product_virt_host_dep.id, -1, [], '', '', '', nil, nil, false)
+    @pool_virt_product_multiplier = create_pool_and_subscription(@owner['key'], @product_virt_product_multiplier.id, -1, [], '', '', '', nil, nil, false)
+    @pool_virt_instance_multiplier = create_pool_and_subscription(@owner['key'], @product_virt_instance_multiplier.id, -1, [], '', '', '', nil, nil, false)
 
     @pools = @cp.list_pools :owner => @owner.id, :product => @product_unlimited_virt.id
     @pools.size.should == 2
@@ -63,6 +81,10 @@ describe 'Unlimited Master Pools' do
     @pools = @cp.list_pools :owner => @owner.id, :product => @product_virt.id
     @pools.size.should == 2
     @pools = @cp.list_pools :owner => @owner.id, :product => @product_virt_host_dep.id
+    @pools.size.should == 2
+    @pools = @cp.list_pools :owner => @owner.id, :product => @product_virt_product_multiplier.id
+    @pools.size.should == 2
+    @pools = @cp.list_pools :owner => @owner.id, :product => @product_virt_instance_multiplier.id
     @pools.size.should == 2
   end
 
@@ -73,20 +95,20 @@ describe 'Unlimited Master Pools' do
     ents[0].quantity.should == 300
   end
 
-  it 'allows system to consume limited virt quantity pool' do
+ it 'allows system to consume limited virt quantity pool' do
     skip("candlepin running in hosted mode") if is_hosted?
     @pools = @cp.list_pools :owner => @owner.id, :product => @product_virt.id
     guest_pool = nil
     @pools.each do |pool|
         if get_attribute_value(pool['attributes'], "pool_derived") == "true" and pool.type == 'UNMAPPED_GUEST'
-           guest_pool = pool
+          guest_pool = pool
         end
     end
     @guest_client_unmapped.consume_pool(guest_pool['id'], {:uuid => @guest_unmapped.uuid, :quantity => 4000})
     ents = @guest_client_unmapped.list_entitlements
     ents.size.should == 1
     ents[0].quantity.should == 4000
-  end
+ end
 
  it 'allows mapped guest to consume unlimited quantity pool' do
     skip("candlepin running in hosted mode") if is_hosted?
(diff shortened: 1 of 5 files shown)


nikosmoum (changes requested):
A quick pass over this. I'll take another look when we have clear expectations on the virt bonus pool use case

server/spec/pool_unlimited_master_spec.rb
```diff
@@ -82,10 +82,20 @@
            guest_pool = pool
         end
     end
-    @guest_client_unmapped.consume_pool(guest_pool['id'], {:uuid => @guest_unmapped.uuid, :quantity => 4000})
+
+    # Guest pools are available in limited quantity which is equal to virt_limit (5)
+    expect {
+      @guest_client_unmapped.consume_pool(guest_pool['id'], {:uuid => @guest_unmapped.uuid, :quantity => 4000})
+    }.to raise_error(RestClient::Forbidden)
```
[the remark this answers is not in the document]
wolfdale: I think we need to keep this spec test as it is. I have added two new spec tests relevant to above conditions.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -74,6 +75,12 @@ public PoolRules(PoolManager poolManager, Configuration config, EntitlementCurat
     }
 
     private long calculateQuantity(long quantity, Product product, String upstreamPoolId) {
+        // Negative pool quantities are treated as unlimited quantity
+        // Limiting all negative quantities to -1
+        if (quantity < 0) {
+            return UNLIMITED_QUANTITY;
```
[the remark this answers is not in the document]
wolfdale: Done.

server/spec/post_bind_bonus_pool_spec.rb
```diff
@@ -100,6 +107,19 @@
   @hostlimited_bonus_stacked_pool = sub_pools[0]
   @hostlimited_bonus_stacked_pool['quantity'].should == 90
 
+  @unlimited_product_master_pool = create_pool_and_subscription(owner['key'], unlimited_prod['id'], -100)
```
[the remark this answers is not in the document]
wolfdale: Added more spec tests.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -244,15 +251,28 @@ private boolean hasBonusPool(List<Pool> pools) {
...
      */
-    private String getVirtQuantity(String virtLimit, long quantity) {
+    private String getVirtQuantity(String virtLimit, long masterPoolQuantity, boolean hostLimited) {
+        if (hostLimited && ("unlimited".equals(virtLimit) || masterPoolQuantity == UNLIMITED_QUANTITY)) {
+            return String.valueOf(UNLIMITED_QUANTITY);
+        }
+
         if ("unlimited".equals(virtLimit)) {
             return virtLimit;
         }
```
nikosmoum, line 261: We know that we should treat both Unmapped guest pools and Legacy virt bonus pools the same, so don't need the host_limited value at all in this method. We could also merge the if statement above with this one to be:
```
if ("unlimited".equals(virtLimit) || masterPoolQuantity == UNLIMITED_QUANTITY) {
    return "unlimited";
}
```
since all we care about is to return 'unlimited' if either virt_limit or the master pool is unlimited.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -244,15 +251,28 @@ private boolean hasBonusPool(List<Pool> pools) {
...
             int virtLimitInt = Integer.parseInt(virtLimit);
+
             if (virtLimitInt > 0) {
-                return String.valueOf(virtLimitInt * quantity);
+                long quantity = virtLimitInt * masterPoolQuantity;
+
+                // Master pool quantity is negative (unlimited quantity)
+                // Set bonus pool quantity equal to virt_limit
+                if (quantity < 0) {
+                    return String.valueOf(virtLimitInt);
```
nikosmoum, line 272: This is now incorrect. If `quantity` is negative, that would mean that master pool is negative, so that means we want to return `"unlimited"`. Let's update this and the comment above it

server/spec/pool_unlimited_master_spec.rb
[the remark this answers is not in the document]
nikosmoum: The assertion here is incorrect. The master pool (`@pool_virt`) for the main product (`@product_virt`) has quantity set to -1, which means the unmapped guest pool should have unlimited quantity, which means the guest can attach a quantity as large as they want (e.g. 4000) without error. Essentially, the original test here was correct

nikosmoum (changes requested):

server/spec/pool_unlimited_master_spec.rb
```diff
@@ -130,11 +161,70 @@
...
+   pools = @cp.list_owner_pools(@owner['key'])
+   sub_pools = pools.select do |pool|
+     pool['subscriptionId'] == @pool_virt_product_multiplier['subscriptionId'] &&
+       pool['type'] != 'NORMAL'
+   end
+   sub_pool = sub_pools[0]
+
+   # sub pool quantity is expected to be equal to virt_limit
+   sub_pool = @cp.get_pool(sub_pool['id'])
+   expect(sub_pool['quantity']).to eq(8)
```
nikosmoum, line 191: assertion is wrong here: master pool is unlimited (-1), so the bonus pool should also be unlimited

server/spec/pool_unlimited_master_spec.rb
```diff
@@ -130,11 +161,70 @@
...
+   pools = @cp.list_owner_pools(@owner['key'])
+   sub_pools = pools.select do |pool|
+     pool['subscriptionId'] == @pool_virt_instance_multiplier['subscriptionId'] &&
+       pool['type'] != 'NORMAL'
+   end
+   sub_pool = sub_pools[0]
+
+   # sub pool quantity is expected to be equal to virt_limit
+   sub_pool = @cp.get_pool(sub_pool['id'])
+   expect(sub_pool['quantity']).to eq(8)
```
nikosmoum, line 213: same here: master pool is unlimited (-1), so the bonus pool should also be unlimited

server/spec/post_bind_bonus_pool_spec.rb
```diff
@@ -100,6 +123,41 @@
...
+  @cp.get_pool(@unlimited_product_master_pool['id'])['quantity'].should == -1
+  pools = @cp.list_owner_pools(@owner_key)
+  pools.length.should == 12
+  sub_pools = pools.select do |pool|
+    pool['subscriptionId'] == @unlimited_product_master_pool['subscriptionId'] &&
+      pool['type'] != 'NORMAL'
+  end
+  sub_pools.length.should == 1
+  @unlimited_product_bonus_pool = sub_pools[0]
+  @cp.get_pool(@unlimited_product_bonus_pool['id'])['quantity'].should == 4
```
nikosmoum, line 137: same here: if master pool is unlimited (-1), the bonus pool should be unlimited (-1)

server/spec/post_bind_bonus_pool_spec.rb
```diff
@@ -409,4 +467,127 @@ def ent_derived_test(pool, expected_quantity)
     @cp.get_entitlement(ent.id)['quantity'].should == 10
   end
 
+  it 'should allow limited consumption of Bonus pools for unlimited master pool created with non-zero virt_limit' do
```
nikosmoum, line 470: The premise of this test is wrong. If the master pool is unlimited, the bonus pool should always be unlimited too

server/spec/post_bind_bonus_pool_spec.rb
```diff
@@ -409,4 +467,127 @@ def ent_derived_test(pool, expected_quantity)
...
+    expect(master_pool['quantity']).to eq(-1)
+  end
+
+  it 'should allow unlimited consumption of master pool regardless of virt_limit' do
+    # unlimited_prod is created with -100 quantity & virt_limit of 4
+    master_pool = @cp.get_pool(@unlimited_product_master_pool['id'])
+    expect(master_pool['quantity']).to eq(-1)
+
+    bonus_pool = @cp.get_pool(@unlimited_product_bonus_pool['id'])
+    expect(bonus_pool['quantity']).to eq(4)
```
nikosmoum, line 516: The premise of this test is correct, but this assertion (and the following ones for the same pool) are wrong, since all bonus pools should always be unlimited when the master pool is unlimited

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -837,20 +856,35 @@ private long processVirtLimitPools(List<Pool> existingPools,
...
+                                        isMasterUnlimited = true;
+                                    }
                                 }
                             }
-                            expectedQuantity = (expectedQuantity - adjust) * virtLimit;
+
+                            // Set quantity equal to virt_limit, if master pool quantity is unlimited &
+                            // virt_limit is non-zero.
+                            if (isMasterUnlimited) {
+                                expectedQuantity = virtLimit;
```
nikosmoum, line 877: expectedQuantity = -1

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -837,20 +856,35 @@ private long processVirtLimitPools(List<Pool> existingPools,
...
+
                             for (Pool derivedPool : existingPools) {
                                 String isDerived =
                                     derivedPool.getAttributeValue(Pool.Attributes.DERIVED_POOL);
 
                                 if (isDerived == null) {
                                     adjust = derivedPool.getExported();
+
+                                    if (derivedPool.getQuantity() == UNLIMITED_QUANTITY) {
+                                        isMasterUnlimited = true;
```
nikosmoum, line 869: we also need to make sure the `existingPool` whose quantity we're trying to calculate is either a BONUS or UNMAPPED_GUEST type, and only then set expectedQuantity = -1

wolfdale: Changes made. This check is also added in EntitlementRules [postUnbindVirtLimit](https://github.com/candlepin/candlepin/blob/aade17f8fa0e6db5a64c0f2a439a7f9ec1c22c70/server/src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java#L546) & [adjustHostedBonusPoolQuantity](https://github.com/candlepin/candlepin/blob/aade17f8fa0e6db5a64c0f2a439a7f9ec1c22c70/server/src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java#L733)

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
wolfdale: Fixed.

server/spec/post_bind_bonus_pool_spec.rb
wolfdale: Fixed.

server/spec/post_bind_bonus_pool_spec.rb
wolfdale: Fixed.

server/spec/post_bind_bonus_pool_spec.rb
wolfdale: Fixed.

server/spec/pool_unlimited_master_spec.rb
wolfdale: Fixed.

server/spec/pool_unlimited_master_spec.rb
wolfdale: Fixed.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
wolfdale: Since we treat unmapped_guest & Bonus pool same while calculating the quantities, this check is redundant.

server/src/main/java/org/candlepin/policy/js/pool/PoolRules.java
wolfdale: Fixed.

server/spec/pool_unlimited_master_spec.rb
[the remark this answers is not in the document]
wolfdale: Reverted to original test case.

nikosmoum (changes requested):
A couple things to be changed, but otherwise good

server/spec/pool_unlimited_master_spec.rb
```diff
@@ -130,11 +152,79 @@
...
     guest_pool.should_not be nil
     @guest_client_unmapped.consume_pool(guest_pool['id'], {:uuid => @guest_unmapped.uuid, :quantity => 600})
     ents = @guest_client_unmapped.list_entitlements
     ents.size.should == 1
     ents[0].quantity.should == 600
-  end
+ end
 
+ it 'product multiplier should have no effect on unlimited master pool quantity' do
+    skip("candlepin running in hosted mode") if is_hosted?
```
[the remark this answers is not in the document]
wolfdale: Right, Made changes to support both Candlepin modes for this spec test.

nikosmoum (changes requested):
Ack on the change, but I just realized the branch is wrong. This should be fixed against 4.0 and master

PR #3078, opened 2021-07-01
[M] 1928161: Large CRL file operation causes OOM error in Candlepin

Januson:
- Deleted old crl endpoints
- Deleted crl file handling
- Deleted CrlUpdateJob
- Added a new expoint for retrieving revoked certificate serials
- Added a new CertificateCleanupJob to handle cleanup of expired certificates(identity and content access) and revoked expired certificate serials


nikosmoum (commented):
Based on the failing spec tests, It seems like virt-who check-in is depending on the no-id cert use case. We should figure out if it would be ok to generate a cert anyway, or not.

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
```diff
@@ -0,0 +1,96 @@
...
+        log.debug("Certificate cleanup successfully completed.");
+    }
+
+    private void cleanupExpiredIdentityCerts() {
+        List<Consumer> consumers = this.consumerCurator.findAllWithExpiredIdCerts();
+        log.info("Cleaning up {} expired identity certificates.", consumers.size());
+        for (Consumer consumer : consumers) {
+            IdentityCertificate idCert = consumer.getIdCert();
+            consumer.setIdCert(null);
+            this.identCertCurator.delete(idCert);
```
nikosmoum, line 83: There would be problems here the first time this job would run in Hosted, due to millions of expired id certs lying around, so we should re-think this approach.

Instead of fetching potentially millions of Consumers into memory for the sole purpose of deleting their id certs, we should be doing all these operations through a few queries instead, e.g.:
1. Fetch the serials of id certs that are expired
2. Update cp_consumer.consumer_idcert_id to NULL for those certs who have those serials
3. Delete the cp_id_cert rows for those serials
4. Update the cp_cert_serial to revoked=true for those serials
Also, since I suspect these queries would make use of the IN clause, we should be partitioning the cert serials into batches like we do in other Curators, to guard against dbms limitations there.

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    private void cleanupExpiredContentAccessCerts() {
+        List<Consumer> consumers = this.consumerCurator.findAllWithExpiredContentAccessCerts();
+        log.info("Cleaning up {} expired content access certificates.", consumers.size());
+        for (Consumer consumer : consumers) {
+            ContentAccessCertificate cert = consumer.getContentAccessCert();
+            consumer.setContentAccessCert(null);
+            this.caCertCurator.delete(cert);
```
nikosmoum, line 93: Same thing as for identity certs here: Don't instantiate consumers into memory, use queries, partition queries in batches

nikosmoum (changes requested):
A few things to look into here. Also, let's not use the bugzilla description in the commit and PR summary, but something that more accurately describes this change (e.g. `Introduce new CRL API & Certificate cleanup job`)

api/candlepin-api-spec.yaml
```diff
@@ -2824,48 +2824,19 @@ paths:
...
-              type: string
-      responses:
-        204:
-          description: Certificates successfully unrevoked
+                type: array
+                items:
+                  type: integer
+                  format: int64
         500:
           description: CRL file path not defined in config file
```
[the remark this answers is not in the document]

src/main/java/org/candlepin/model/CertificateSerialCurator.java
```diff
@@ -216,4 +84,25 @@ public int deleteRevokedExpiredAndNotCollectedSerials() {
...
             .getResultList();
     }
+
+    /**
+     * Returns all serial ids that are both expired and revoked.
+     *
+     * @return a list of serial ids
+     */
+    @SuppressWarnings("unchecked")
+    public List<Long> listExpiredRevokedSerialIds() {
```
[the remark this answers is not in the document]
Januson: Done.

api/candlepin-api-spec.yaml
```diff
@@ -2824,48 +2824,19 @@ paths:
     get:
       tags:
         - Certificate Revocation List
-      description: Retrieves the certificate revocation list
+      description: Retrieves the list of all revoked certificate serial ids
```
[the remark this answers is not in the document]

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    @Transactional
+    public void execute(JobExecutionContext context) throws JobExecutionException {
+        log.debug("Starting certificate cleanup.");
+        // Skipping entitlement certificates as those are handled by ExpiredPoolsCleanupJob
+        cleanupExpiredIdentityCerts();
+        cleanupExpiredContentAccessCerts();
+        this.serialCurator.deleteRevokedExpiredSerials();
```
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
Januson: Good point. I did not considered a scenario where is accumulated so much already. Done

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
Januson: Done.

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private void cleanupExpiredIdentityCerts() {
+        List<ExpiredCertificate> allExpiredIdCertificates = this.identCertCurator.listAllExpired();
+        log.info("Cleaning up {} expired identity certificates.", allExpiredIdCertificates.size());
+        for (List<ExpiredCertificate> expiredCertificates : partition(allExpiredIdCertificates)) {
+            List<String> expiredCertIds = certIdsOf(expiredCertificates);
+            List<Long> expiredSerials = serialsOf(expiredCertificates);
+            List<String> consumersToUnlink = this.consumerCurator.findUuidsByIdentCerts(expiredCertIds);
+            int unlinkedConsumers = this.consumerCurator.unlinkIdCertificatesOf(consumersToUnlink);
```
nikosmoum, line 89: Doing these 2 steps as two separate queries runs the risk of a race condition, where if a consumer with an expired cert checks-in in between those 2 queries and gets a new valid certificate, then the `unlinkIdCertificatesOf` method would be unlinking that new valid certificate. Doing it in a single query would make sure the database does it as a single atomic operation

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private void cleanupExpiredContentAccessCerts() {
+        List<ExpiredCertificate> allExpiredCaCertificates = this.caCertCurator.listAllExpired();
+        log.info("Cleaning up {} expired content access certificates.", allExpiredCaCertificates.size());
+        for (List<ExpiredCertificate> expiredCertificates : partition(allExpiredCaCertificates)) {
+            List<String> expiredCertIds = certIdsOf(expiredCertificates);
+            List<Long> expiredSerials = serialsOf(expiredCertificates);
+            List<String> consumersToUnlink = this.consumerCurator
+                .findUuidsByContentAccessCerts(expiredCertIds);
+            int unlinkedConsumers = this.consumerCurator.unlinkCaCertificatesOf(consumersToUnlink);
```
nikosmoum, line 104: Same here: a single query should be used.

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
nikosmoum: In case this is hard to do using HQL, we can always use pure SQL (and make sure to test it against both mariadb + postgres).

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    private void cleanupCertificateSerials() {
+        int deleted = this.serialCurator.deleteRevokedExpiredSerials();
+        log.debug("Cleaning up {} expired and revoked certificate serials.", deleted);
+    }
+
+    private Iterable<List<ExpiredCertificate>> partition(List<ExpiredCertificate> list) {
+        return Iterables.partition(list, this.config.getInt(ConfigProperties.CERT_CLEANUP_BULK_SIZE));
```
nikosmoum, line 117: No need to do this with a custom config entry, especially with a default value of 1000 which is very low and might lead to too many queries. We already have an abstraction for the IN clause use case: `AbstractHibernateCurator.getInBlockSize()` which will use whatever the allowed limit for IN blocks is for the given underlying rdbms

nikosmoum (changes requested):
A couple more changes needed, but we're getting there

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
Januson: Done.

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
Januson: Done.

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
Januson: Done.

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -320,6 +322,10 @@ private ContentAccessCertificate updateScaCertificate(Consumer consumer, Owner o
         return existing;
     }
 
+    private void revoke(ContentAccessCertificate existing) {
```
Ceiu, line 325: We probably want a separate PR for this, as this seems important enough for its own call out.

src/main/java/org/candlepin/model/CertificateSerialCurator.java
```diff
@@ -216,4 +85,46 @@ public int deleteRevokedExpiredAndNotCollectedSerials() {
...
+     * @return a number of revoked serials
+     */
+    @Transactional
+    public int revokeByIds(Collection<Long> serialsToRevoke) {
+        if (serialsToRevoke == null || serialsToRevoke.isEmpty()) {
+            return 0;
+        }
+
+        String hql = "UPDATE CertificateSerial s SET s.revoked = true" +
+            " WHERE s.revoked = false AND s.id IN (:serials)";
```
Ceiu, line 123: Need protection for the IN operator here.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1552,4 +1545,127 @@ public int cullInvalidConsumerContentAccess(Owner owner, String... existingModes
...
+        else if (db.contains("postgresql")) {
+            return this.postgreUnlinkCaCertificates(certIds);
+        }
+        else {
+            throw new PersistenceException(
+                "Cannot execute query: The database dialect is not recognized.");
+        }
+    }
+
+    private String getDbDialect() {
```
Ceiu, line 1666: If we're going to abstract this out, move it up a tier to the AbstractHibernateCurator.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1552,4 +1545,127 @@ public int cullInvalidConsumerContentAccess(Owner owner, String... existingModes
...
+
+    /**
+     * Takes a list of identity certificate id and unlinks them from consumers.
+     *
+     * @param certIds certificate ids to be unlinked
+     * @return a number of unlinked consumers
+     * @throws PersistenceException when unsupported db dialect is detected
+     */
+    @Transactional
+    public int unlinkIdCertificates(Collection<String> certIds) throws PersistenceException {
```
Ceiu, line 1630: Seeing direct SQL for each backend gives me pause. What is the goal here and why can we not let Hibernate handle it nor use a different query?

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
```diff
@@ -129,4 +131,41 @@ private int deleteContentAccessCerts(Set<String> certIdsToDelete) {
...
+     *
+     * @param idsToDelete ids to be deleted
+     * @return a number of deleted certificates
+     */
+    @Transactional
+    public int deleteByIds(Collection<String> idsToDelete) {
+        if (idsToDelete == null || idsToDelete.isEmpty()) {
+            return 0;
+        }
+        String hql = "DELETE FROM ContentAccessCertificate c WHERE c.id IN (:idsToDelete)";
```
Ceiu, line 165: Need protection for the IN operator.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1552,4 +1545,127 @@ public int cullInvalidConsumerContentAccess(Owner owner, String... existingModes
         return entityManager.createQuery(query).getResultList();
     }
 
+    @Transactional
+    private int postgreUnlinkIdCertificates(Collection<String> certIds) {
```
Ceiu, line 1549: In all of these backend-specific queries, we're doing an UPDATE-JOIN on the primary key (cert ID), and then examining the ID of the cert to select which rows to update.

Unless I'm missing something, the join is unnecessary -- especially if we're stepping into raw SQL.

src/main/java/org/candlepin/model/IdentityCertificateCurator.java
```diff
@@ -31,9 +35,40 @@ public IdentityCertificateCurator() {
...
+     *
+     * @param idsToDelete ids to be deleted
+     * @return a number of deleted certificates
+     */
+    @Transactional
+    public int deleteByIds(Collection<String> idsToDelete) {
+        if (idsToDelete == null || idsToDelete.isEmpty()) {
+            return 0;
+        }
+        String hql = "DELETE FROM IdentityCertificate c WHERE c.id IN (:idsToDelete)";
```
Ceiu, line 68: Need protection for IN operator

src/main/java/org/candlepin/resource/HypervisorTypeChecker.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.candlepin.model.ConsumerType;
+import org.candlepin.model.ConsumerTypeCurator;
+
+import com.google.inject.Inject;
+
+import java.util.Objects;
+
+
+public class HypervisorTypeChecker {
```
Ceiu, line 26: This seems like a really complicated way of adding a `.matches`/`.isOfType`/etc. method on ConsumerType, and enum support in ConsumerTypeCurator.

Ceiu (changes requested):

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
```diff
@@ -129,4 +131,41 @@ private int deleteContentAccessCerts(Set<String> certIdsToDelete) {
...
     }
 
+
+    /**
+     * Lists all expired content access certificates that are not revoked.
+     *
+     * @return a list of expired certificates
+     */
+    @SuppressWarnings("unchecked")
+    public List<ExpiredCertificate> listAllExpired() {
```
Ceiu, line 141: Something about using a custom projection object here isn't sitting well with me. I'm not sure we actually need it, and it seems like this (and a few other methods) are optimizations around a join or lookup later.

Ceiu (changes requested):

spec/certificate_cleanup_spec.rb
```diff
@@ -0,0 +1,154 @@
+require 'spec_helper'
+require 'candlepin_scenarios'
+require 'pg'
+require 'mysql'
+
+describe 'Scheduled Jobs' do
```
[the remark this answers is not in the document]
Januson: True. Writing them as integration tests would be better and I originally had them there but that's no longer doable since db specific queries won't work with hsqldb.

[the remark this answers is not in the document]
Januson: Done. Rewrote the query so that the spec test is not needed.

spec/certificate_cleanup_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+    serial.should_not be_nil
+  end
+
+  def expect_expired_serial(serial)
+    Date.parse(serial['expiration']).should < Date.today
+  end
+
+  def expire_serial(id)
+    sql = "UPDATE cp_cert_serial SET expiration = '2020-05-27 03:59:59+00' WHERE id = #{id}"
+    if is_hosted?
```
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/async/tasks/CertificateCleanupJob.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    private void cleanupCertificateSerials() {
+        int deleted = this.serialCurator.deleteRevokedExpiredSerials();
+        log.debug("Cleaning up {} expired and revoked certificate serials.", deleted);
+    }
+
+    private Iterable<List<ExpiredCertificate>> partition(List<ExpiredCertificate> list) {
+        return Iterables.partition(list, this.consumerCurator.getInBlockSize());
```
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/auth/SSLAuth.java
```diff
@@ -37,16 +44,26 @@
  * Remember, certs are easy.
  */
 public class SSLAuth extends ConsumerAuth {
+    private static final Logger log = LoggerFactory.getLogger(SSLAuth.class);
```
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/controller/ContentAccessManager.java
Januson: Done.

src/main/java/org/candlepin/model/CertificateSerialCurator.java
Januson: Done.

src/main/java/org/candlepin/model/ConsumerCurator.java
Januson: Done.

src/main/java/org/candlepin/model/ConsumerCurator.java
Januson: Done.

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
Januson: Done.

src/main/java/org/candlepin/model/ConsumerCurator.java
Januson: Done.

src/main/java/org/candlepin/model/IdentityCertificateCurator.java
Januson: Done.

jturel:
Looking at this from the Katello side... We currently disable the CRLUpdateJob because we don't use a CRL and it's expensive. It looks like Candlepin will not generate a CRL under any circumstance with this change and that's OK with me.

Regarding the CertificateUpdateJob, I see the default schedule is every 12 hours. Would doing this every 24 hours make more sense in a Katello/Sat environment since we don't really deal in revoked certificates?

nikosmoum:
@jturel Hi Jonathon, you are right that the CRL file will no longer be generated with this change.
The CertificateCleanupJob is important for clearing out 2 things though:
- Deletes all expired (but not necessarily revoked) Identity and SCA certificates (these might pile up when hosts register and then never unregister themselves, and time passes so those certs are never revoked, but are expired and therefore invalid).
- Cleans up all the certificate serials that are both expired and revoked (this includes serials of all 3 types of certs: identity, SCA and entitlement).
So it's function is not really linked to the new CRL API/behaviour, but simply for making sure the db tables that hold these don't increase needlessly. If you think Sat/Katello's has less cert/serial load to cleanup compared to Hosted, feel free to set it to 24hours (but I would recommend it always is set to run periodically regardless of the period).

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
Januson: Even after refactoring the query I don't see a good way to replace the projection without splitting it into two queries. Do you have any suggestion on how to do get rid of it?

src/main/java/org/candlepin/resource/HypervisorTypeChecker.java
Januson: As per IRC discussion this will be done in a follow-up PR.

jturel:
Thank you Nikos. We'll stick with the 24 hour interval like we used for the CRLUpdateJob. We've made the period configurable on our end so it's easy to change

nikosmoum:
Let's validate the use cases around the expired identity certs with stakeholders before moving on with this (specifically for the relevant change in `ConsumerResource.getConsumer()` in this PR)

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -663,6 +666,10 @@ public ConsumerDTO getConsumer(@Verify(Consumer.class) String uuid) {
                     consumer = this.regenerateIdentityCertificate(consumer);
                 }
             }
+            else if (!this.hypervisorType.isHypervisor(consumer.getTypeId())) {
+                log.info("No identity certificate found for consumer: {}. Generating a new one", uuid);
+                consumer = this.regenerateIdentityCertificate(consumer);
+            }
```
nikosmoum, line 672: @Januson Based on our previous discussions, we should revert this. We don't want to regenerate the identity cert if there isn't one, because that would mean they either don't need one (it's a hypervisor), or it was expired and the CertCleanupJob deleted it (at which point, they should just be re-registering manually with username/password)

Januson: Done.

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
nikosmoum: Just a note that we discussed this on IRC a few days ago, and the serials being exposed both to the REST API and Adapters API complicates this, and would be better done as a separate task

nikosmoum: Perhaps a naive question, but why do we need the custom object for this query? Does HQL not allow for selecting 2 columns from 2 joined tables? e.g. `SELECT c.id, s.id ...`?

Januson: We don't. Using scalar projection would be fine. I just find DTO projections more descriptive compared to array you get from scalar projection.

nikosmoum: I suppose I can see both the pros/cons of this approach, but as someone reading this query for the first time, I was immediately trying to understand why is this needed, which is distracting and takes away from readability/descriptiveness (at least for me).
The fact we have a separate model class just for this is also not great. If this became a trend, then we'd have model classes mixed in with other model classes that aren't really entity models, but DTO projections. If we are to keep this DTO projection, I'd say we should put it under a sub-package like `model.projections` or something.

I'm fine with either way (keeping DTO projection, or using scalar projection), so we could keep it unless @Ceiu feels strongly about this

nikosmoum:
Ack from me; I'll let @Ceiu re-review

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
Januson: It's not really a separate model. Just a view of model. It's useful when you only need read-only data and it's actually a recommended practice to use projections as they have performance benefits:
- You are not querying  as much data
- DTO and scalar projections are ~20% faster than entities (for the same amount of queried data) as Hibernate managed entities have huge overhead.

Ceiu: Generally speaking, I don't like creating objects for the sake of creating objects. I initially called this area out because the logic throughout the chain is coupled down to the new expiry job, and the projection DTO seemed to be a result of that. However, in this case, I've since come to the conclusion that the logic isn't the problem here, but rather a result of our goofy model surrounding certs and serials.  In cases like this, where we have a clear need for a multi-column projection *and* we need the result of said projection outside of the curator, it's acceptable.

I'd still prefer we didn't have it overall, but the only performant way to remove it that I see involves fixing the cert model, which we've decided is a separate task.

PR #3085, opened 2021-07-26
[3.1] 1985360: Don't filter out content with no arch specified [ENT-4128]

nikosmoum:
- The fix applies to content for both SCA and Entilement certs

### DIFF
--- a/server/src/main/java/org/candlepin/util/X509Util.java
+++ b/server/src/main/java/org/candlepin/util/X509Util.java
@@ -230,7 +230,20 @@ public abstract class X509Util {
         return contentArches;
     }
 
+    /**
+     * Returns true/false depending on whether or not there is a match/mismatch between the consumer's
+     * supported architectures and the content's supported architectures.
+     *
+     * @param consumerArches The architectures that the consumer supports
+     * @param contentArches The architectures that the content supports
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/candlepin/util/X509Util.java
```diff
@@ -230,7 +230,20 @@ public String createFullContentPath(String contentPrefix, ProductContent pc) {
         return contentArches;
     }
 
+    /**
+     * Returns true/false depending on whether or not there is a match/mismatch between the consumer's supported
+     * architectures and the content's supported architectures.
+     *
+     * @param consumerArches The architectures that the consumer supports
+     * @param contentArches The architectures that the content supports
+     * @return False if none of the content's architectures match to any of the consumer archituctures
```
Ceiu, line 239: Typo with the second "architectures" in this line, and the comma should move up (or if it's too long, arch moves down)

nikosmoum: Done. also added a unit test

Ceiu (approved):

Ceiu:
I've manually tested the following and found no issues: checkstyleMain, checkstyleTest, test, rspec (postgresql, postgresql+hosted, mysql, mysql+hosted)

Overriding Jenkins issue and merging.

PR #3087, opened 2021-07-28
ENT-4018: Cleanup docker containers & networks on Jenkins pipeline abort

Januson:
- Add post-step cleanup to pipeline stages

### DIFF
--- a/jenkins/cleanup.sh
+++ b/jenkins/cleanup.sh
@@ -1,0 +1,15 @@
+#!/bin/bash -x
+
+echo "Cleaning up workspace: ${WORKSPACE}"
+
+DIR="$(git rev-parse --show-toplevel)/docker/"
+cd $DIR
+
+# Run the cleanup
+PROJ_NAME="${STAGE_NAME}-${BUILD_TAG}"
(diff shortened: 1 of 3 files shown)


jenkins/cleanup.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Run the cleanup
+PROJ_NAME="${STAGE_NAME}-${BUILD_TAG}"
+docker-compose -p $PROJ_NAME down
+RETVAL=$?
+
+echo "return value: $RETVAL"
+cd -
+
+sudo chown -R jenkins:jenkins ${WORKSPACE}/artifacts
```
Ceiu, line 26: Is all of the stuff surrounding the artifacts directory necessary in the cleanup script? A lot of it looks like carry over from other scripts which might not be important to the "down" operation.

Januson: Done. I deleted redundant lines.

nikosmoum (approved):
I aborted the last jenkins run for this PR and checked all jenkins nodes that had jobs from this PR running, and I can confirm all docker containers and networks were removed after the abort.

PR #3088, opened 2021-07-28
[M] 1984984: Don't filter out content with no arch specified [ENT-4128]

nikosmoum:
- The fix applies to content for both SCA and Entilement certs


Ceiu:
ack. Running test suite manually to verify.

PR #3090, opened 2021-07-29
[M] Log changes to content access mode for owner (ENT-4121)

wolfdale:
Changes
- Logging changes to content access mode for owner.

### DIFF
--- a/src/main/java/org/candlepin/resource/OwnerResource.java
+++ b/src/main/java/org/candlepin/resource/OwnerResource.java
@@ -977,9 +977,12 @@ public class OwnerResource implements OwnersApi {
             owner = this.contentAccessManager
                 .updateOwnerContentAccess(owner, contentAccessModeList, contentAccessMode);
         }
-
-        owner = ownerCurator.merge(owner);
-        ownerCurator.flush();
+        else {
+            // ContentAccessManager does its own merge on owner,
+            // if updates are NOT related to content access we need to take care of it.
+            owner = ownerCurator.merge(owner);
+            ownerCurator.flush();
+        }
 
         // Build and queue the owner modified event
         Event event = this.eventFactory.getEventBuilder(Target.OWNER, Type.MODIFIED)
--- a/src/main/java/org/candlepin/controller/ContentAccessManager.java
+++ b/src/main/java/org/candlepin/controller/ContentAccessManager.java
@@ -767,6 +767,9 @@ public class ContentAccessManager {
                 this.contentAccessCertCurator.deleteForOwner(owner);
             }
             this.eventSink.emitOwnerContentAccessModeChanged(owner);
+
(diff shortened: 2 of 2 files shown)


Ceiu (changes requested):
A better place for the logging would be the ContentAccessManager, which handles this logic for most(/every) means of changing the SCA mode or list.

Check ContentAccessManager:~760

nikosmoum (commented):
@Ceiu I think that is called in both the create and update endpoints, and we only want to log this during the update endpoint. Also, we don't want to log it in case something goes wrong during the rest of the call, which is why @wolfdale added the logging after the flush is done

Ceiu:
@nikosmoum Import as well, but prod likely doesn't care about that. It just seems odd to me to add logging in one specific area, given that we have a centralized place for this logic, that also kicks out a specific event when a change occurs. Not wanting it outputting on creation is weird to me -- especially given that the initial mode is configurable on creation (thus, functionally identical to the state flipping they're interested in checking).

FWIW, ContentAccessManager does its own merge on the owner. In the case where the SCA stuff changes during updateOwner, that second merge+flush is redundant.

nikosmoum:
@Ceiu Ack, let's move this in the `ContentAccessManager` then

nikosmoum:
@wolfdale Also, I think we need to port this to 4.0 too

wolfdale:
@nikosmoum @Ceiu Ack.
I'll cherrypick this to 4.0 once this gets merged.

nikosmoum (approved):
ack from me. I'll let @Ceiu re-review & merge

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -978,8 +978,12 @@ public OwnerDTO updateOwner(@Verify(Owner.class) String key,
                 .updateOwnerContentAccess(owner, contentAccessModeList, contentAccessMode);
         }
 
-        owner = ownerCurator.merge(owner);
-        ownerCurator.flush();
+        // ContentAccessManager does its own merge on owner, if updates are related to content access
+        // this is redundant.
+        if (!updateContentAccess) {
```
Ceiu, line 983: I'm hesitant to okay this, since we intentionally left it off prior to avoid logical coupling to the CAM for merging the product data. With the added comment it's probably fine, as future maintainers would have a hint as to why the merge isn't always done; but in such a case we should reuse the ```updateContentAccess``` check we've already done directly above this block and drop it in as an "else" block.

Ceiu (changes requested):

src/main/java/org/candlepin/resource/OwnerResource.java
wolfdale: I was initially thinking of putting everything inside of else block.
Done.

Ceiu (approved):

PR #3091, opened 2021-07-29
[M] 1980370: Replace DELETE with TRUNCATE where possible; [ENT-4065]

sonalidhome:


Ceiu (approved):

src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
```diff
@@ -20,8 +20,9 @@
             <column name="cont_acc_cert_id"/> <!-- this sets the column value to null -->
         </update>
 
-        <delete tableName="cp_cont_access_cert" />
-
-        <delete tableName="cp_owner_env_content_access" />
+        <sql>
+            TRUNCATE TABLE cp_cont_access_cert cascade;
```
Ceiu, line 24: Actually, the cascade here is dangerous. This will truncate any table with a foreign key relationship, which is not what we want to do here -- we could end up clearing half the DB.

Ceiu (changes requested):

src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
Ceiu: Also, it doesn't look like mysql or mariadb support cascade on TRUNCATE:

- https://dev.mysql.com/doc/refman/8.0/en/truncate-table.html
- https://mariadb.com/kb/en/truncate-table/

src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
```diff
@@ -20,8 +20,29 @@
...
+        <dropForeignKeyConstraint baseTableName="cp_consumer" constraintName="fk_cont_acc_cert" />
+
+        <sql dbms="mysql,mariadb">
+            DELETE FROM cp_cont_access_cert;
+            DELETE FROM cp_owner_env_content_access;
+        </sql>
+
+        <sql dbms="postgresql,oracle,hsqldb">
+            TRUNCATE TABLE cp_cont_access_cert;
+            TRUNCATE TABLE cp_owner_env_content_access cascade;
```
Ceiu, line 32: While CASCADE is safe here (as nothing has a reference into this table), we still don't want to use it because (a) it's dangerous in general (see below) and (b) it causes conflicts with the other DBMSs.

By removing the cascade, we can also remove the DBMS restriction on this block, remove the SQL block above, and allow every DB to use the TRUNCATE operation.

For clarity, when you cascade a truncate operation, it doesn't just delete the rows referencing a row in the table being truncated, *it truncates every table referencing the target table, and any tables referencing them, recursively*. You can see this in action by importing test data into a postgresql deployment, and then truncating the `cp_entitlement` table. By default, there will be zero rows in the entitlements table, and some 700+ pools in cp_pool, but by executing a TRUNCATE TABLE ... CASCADE, several other tables are truncated as well -- even though no rows are actually removed from the cp_entitlement table:

```
candlepin=> SELECT COUNT(*) FROM cp_pool;
 count 
-------
   729
(1 row)

candlepin=> SELECT COUNT(*) FROM cp_entitlement;
 count 
-------
     0
(1 row)

candlepin=> TRUNCATE TABLE cp_entitlement CASCADE;
NOTICE:  truncate cascades to table "cp_ent_certificate"
NOTICE:  truncate cascades to table "cp_pool"
NOTICE:  truncate cascades to table "cp_activationkey_pool"
NOTICE:  truncate cascades to table "cp_pool_attribute"
NOTICE:  truncate cascades to table "cp_product_pool_attribute"
NOTICE:  truncate cascades to table "cp_pool_source_stack"
NOTICE:  truncate cascades to table "cp2_pool_source_sub"
TRUNCATE TABLE

candlepin=> SELECT COUNT(*) FROM cp_pool;
 count 
-------
     0
(1 row)
```

src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
```diff
@@ -20,8 +20,29 @@
             <column name="cont_acc_cert_id"/> <!-- this sets the column value to null -->
         </update>
 
-        <delete tableName="cp_cont_access_cert" />
+        <dropForeignKeyConstraint baseTableName="cp_consumer" constraintName="fk_cont_acc_cert" />
+
+        <sql dbms="mysql,mariadb">
```
Ceiu, line 25: For operations that liquibase actually supports properly, we should use its tags (i.e. the delete tag that this update is replacing).

However, this block is unnecessary with changes to the SQL block below.

Ceiu (changes requested):
Don't forget to add your name as an author on the changeset, as well.

src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
sonalidhome: Agreed. Removed the cascade.

src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
sonalidhome: This was intented to keep MySQL/MariaDB away from Truncate with Cascade. But now I have removed the cascade from both tables. Hence removed this SQL tag as well.

src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
```diff
@@ -6,7 +6,7 @@
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
-    <changeSet id="20201103112757-1" author="crog">
+    <changeSet id="20201103112757-1" author="sdhome">
```
Ceiu, line 9: Convention on this is to have both authors listed, delimited with a comma: `"crog, sdhome"`

Ceiu (changes requested):
One last minor change here and this should be good

src/main/resources/db/changelog/20201103112757-purge-current-sca-certs.xml
sonalidhome: ack.

Ceiu (approved):

PR #3092, opened 2021-07-30
[M] Add back 'docker-compose down' to docker/test script

nikosmoum:
- Even though we do this cleanup in a Jenkins post script, it
  is a good idea to also do it within the docker/test script
  to perform the cleanup when one runs the script manually


sonalidhome (approved):
ack.

PR #3095, opened 2021-08-02
[4.0] Log changes to content access mode for owner (ENT-4121)

wolfdale:
Changes

- Logging changes to content access mode for owner.


sonalidhome (approved):
ack.

PR #3101, opened 2021-08-10
Collapse the api project into the main gradle project.

barnabycourt:
1) Remove the API sub-project in favor of a single project build.
2) Remove the allprojects section as there is now a single project.
3) eliminate the manual setting of the msgfmt plugin generated source directory in favor of having the plugin set it directly on the applied project.
4) Add com.candlepin.gradle as a prefix for the plugin id of all our home grown buildSrc gradle plugins.


wolfdale (approved):
ACK, Looks great.

PR #3103, opened 2021-08-10
[M] Various minor improvements to the cloud registration flow (ENT-4188)

Ceiu:
- Added an additional check to the cloud registration resource
  to prevent token generation when cloud registration is not enabled
- Added additional logging to the AuthenticationFilter to list the
  auth providers used during a given request
- Added logic to consumer registration to avoid proceeding in cases
  where no owner or principal information is available


sonalidhome (approved):
ack.

PR #3106, opened 2021-08-12
Remove spring dependency management in favor of gradle bom support

barnabycourt:
This change removes our use of the spring dependency management plugin.
Gradle now has full support of maven BOM files and can provide equivalent
capabilities by using a BOM to specify the version for groups of related
packages. The gradle BOM support also gives better support for specifying
constraints to update specific libraries to newer versions than those
proscribed by a BOM.

Where I could find a BOM it was included at the beginning of the dependencies block. For the libraries that did not, I specified the versions in-line. 

There were small number of jars that have been updated, added, and removed from the generated jar based on the changes. 

Updated: These are library updates that were specified by the BOM support which the spring-dependency plugin was not finding/updating for us previously. 
commons-codec-1.11.jar -> commons-codec-1.13.jar
commons-lang3-3.2.1.jar -> commons-lang3-3.9.jar
httpcore-4.4.7.jar -> httpcore-4.4.12.jar

Removed: 
guice-multibindings - This is now provided by guice-core as of guice 4.2

Added: All of these were added because of dependencies specified by the BOMs which we were not previously pulling in. Everything appeared to be working without these libs however resteasy really wants them to be there. 
cdi-api-2.0.SP1.jar
javax.el-api-3.0.0.jar
javax.interceptor-api-1.2.jar

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ import org.yaml.snakeyaml.Yaml
 
 buildscript {
     //openapi version has to be set manually in the plugins block as well.
-    ext.openapi_version = '5.2.0'
+    ext.openapi_version = '5.2.1'
 
     repositories {
         mavenCentral()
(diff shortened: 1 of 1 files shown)


barnabycourt:
retest this please

barnabycourt:
retest this please

Ceiu (approved):

PR #3107, opened 2021-08-17
[M] 1986929: Added content_overrides & guestids resources to root resource list

wolfdale:
Changes 
- Due to OpenAPI changes, endpoints to access content override & guestIDs resources were moved to ConsumerResource class. They are now added back to the root resource list. URL links for content override & guestIDs are accessed via methods specific to content override & guestIDs (present in ConsumerResource class).


src/main/java/org/candlepin/resource/RootResource.java
```diff
@@ -75,6 +77,33 @@
...
         addResource(DeletedConsumerResource.class);
+
+        // Due to OpenAPI changes, endpoints to access content override & guestIDs
+        // were moved to ConsumerResource class. Here URL links for content override &
+        // guestIDs are accessed via methods specific to content override & guestIDs
+        Class<?>[] implementedInterfaces = ConsumerResource.class.getInterfaces();
+        Class<?> consumersAPI = null;
+
+        for (Class<?> apiClass: implementedInterfaces) {
+            if (apiClass.getSimpleName().equals("ConsumersApi")) {
```
Ceiu, line 88: While this gets the job done, I'm not at all thrilled about having the names of both the interface and the methods hardcoded into this obscure class. If we need to start manually flagging resources for entry into the list of links, we should be doing something like annotating the methods in question (or flagging them in the API spec to get annotated by the code generator) and then picking that up automatically.

This is just priming us to run into the same bug during a simple refactor and hide it in a one-off error that only shows up the first time the endpoint is hit, and then never again.

I've created a PoC of this which you may use however you like: https://gist.github.com/Ceiu/b6694cba3b5c619987aa4a324a068914

Ceiu (changes requested):

src/main/java/org/candlepin/resource/RootResource.java
wolfdale: Really liked the idea to use annotations here. Thanks.

src/main/java/org/candlepin/resource/RootResource.java
```diff
@@ -85,15 +115,21 @@ public RootResource(Configuration config) {
...
+            if (hideResources.contains(link.getRel())) {
+                log.debug("Hiding supported resource: {}", link.getRel());
+                continue;
+            }
+
+            links.put(link.getRel(), link);
         }
-        return newLinks;
+
+        return new LinkedList<>(links.values()); // :/
```
Ceiu, line 132: This probably should be fixed. I didn't intend for this to go out entirely as-is, since it's rather inefficient. It'd be better if we could just return a collection rather than a list so we don't need to build a new list just to output the correct type.

wolfdale: Ack, Done.

Ceiu (approved):

nikosmoum (approved):
ACK

PR #3109, opened 2021-08-19
Remove unused ldapjdk library

barnabycourt:
I can't find any references to what is provided by this library in our code. Trying it through the test pipeline to see if anything breaks.


sonalidhome (approved):
ack.

PR #3110, opened 2021-08-20
Update keycloak library

barnabycourt:


wolfdale:
We probably need to sync pom.xml file, since we were couple of changes to dependencies. We also doesn't require [this](https://github.com/candlepin/candlepin/blob/master/build.gradle#L697-L706) as we have removed the API projects altogether.

PR #3111, opened 2021-08-20
Update artemis library

barnabycourt:


wolfdale:
pom.xml file need to be synced with these changes.

PR #3115, opened 2021-08-23
[M] 1928161: Add validation to ssl auth

Januson:
- Validate whether the incomming certificate was issued by us


nikosmoum:
Let's validate the use cases around this with stakeholders before moving on with this

nikosmoum (commented):
Both of our main use cases (Hosted & Satellite) already do their own checks for the signature validity of the identity cert, so we should avoid doing double the work. Thus closing this PR.

PR #3118, opened 2021-08-24
[M] Gradle/Maven dependency fix

wolfdale:
**Changes** :
- Instead to reverting back to use Spring dependency management plugin, we now manually handle versions via ext properties. Because of Gradle ```platform``` component, pom task was not able to infer the sub-dependency versions from BOM, as we need to explicitly mention the dependency version for pom auto generation task to work. 
- Removed the ```api``` project clause inside pom task in build.gradle file.
- Upgraded Artemis (#3111) & Keyclock (#3110) dependencies

**Note for reviewer** :
There is one change related to XML header for pom.xml, although was able to build war & deploy Candlepin. 

**Before** 
`<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">` 
**After**
`<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,53 +87,57 @@ ext {
     generatedMetamodels = "${buildDir}/generated/api/src/gen/java"
 }
 
-dependencies {
-    //Specify BOMs that are used for specifying versions of libraries
-    implementation platform('org.jboss.resteasy:resteasy-bom:4.4.3.Final')
-    implementation platform('com.fasterxml.jackson:jackson-bom:2.12.3')
-    implementation platform('com.google.inject:guice-bom:4.2.3')
-    implementation platform('org.junit:junit-bom:5.7.0')
+ext {
+    // Denotes BOM versions
+    resteasy_version = "4.4.3.Final"
+    jackson_version = "2.12.3"
+    guice_version = "4.2.3"
+    junit_version = "5.7.0"
+}
 
+dependencies {
     // Commons
     implementation "commons-codec:commons-codec:1.11"
     implementation "commons-collections:commons-collections:3.2.2"
     implementation "commons-io:commons-io:2.7"
     implementation "commons-lang:commons-lang:2.5"
 
-    //collections
+    // Collections
     implementation "com.google.guava:guava:30.1-jre"
 
     // Gettext libraries used for internationalization & translation
     implementation "com.googlecode.gettext-commons:gettext-commons:0.9.8"
 
-    //Guice Libraries - Version from BOM
-    implementation "com.google.inject.extensions:guice-assistedinject"
-    implementation "com.google.inject.extensions:guice-servlet"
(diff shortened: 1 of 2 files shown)


barnabycourt:
hmm... I'm seeing a number of differences between the libraries included in the gradle generated war & the maven generated war. 

Specifically, it looks like the annotationProcessor libs are not being added to the gradle generated war but are being added to the mvn generated war. 

Also, it appears that a number of other libraries are getting fixed to earlier versions in the mvn build than they are in the gradle build. 

I am testing by using gradle & mvn to generate the war files & then dumping the jar lists out to txt files so that I can diff them more easily. 
For the mvn generated jar: `jar -t -f ./target/candlepin-4.1.5.war | grep jar | sort > ../new_mvn_libs.txt`
For the gradle generaated jar: ` jar -t -f ./build/libs/candlepin-4.1.5.war | grep jar | sort > ../old_libs.txt`

barnabycourt:
@wolfdale Hmm... been digging some more, I'm not sure that the annotation package is a new issue. I see FastInfoset-1.2.15.jar  listed in the current 4.1.5-1 brew build

barnabycourt:
In more detail, comparing the base against brew the only difference I see is the new version of the mvn build appears to not be applying the constraint to javax.validation:validation-api:2.0.1.Final. If all we want to focus on in this PR is making the new build results look like the old build results, this would be the only thing to fix.

build.gradle
```diff
@@ -87,99 +87,103 @@ ext {
...
-    implementation "com.fasterxml.jackson.datatype:jackson-datatype-jdk8"
-    implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
-    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml"
-    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-xml"
+    // Jackson
+    implementation "com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider:$jackson_version"
+    implementation "com.fasterxml.jackson.module:jackson-module-jsonSchema:$jackson_version"
 
-    //Javax
+    // Bean validation API is explicitly added to this version,
```
barnabycourt, line 121: Please call out that this is a transitive dependency of com.fasterxml.jackson.module:jackson-module-jsonSchema in the comment.

barnabycourt (changes requested):

barnabycourt (approved):
lgtm, merge after jenkins

PR #3124, opened 2021-09-03
[4.0] Rewrote the migration script for the product model changes (ENT-4184)

Ceiu:
- Updated the migration scripts for the product model changes
  to correctly move provided and derived products from pools to
  newly created products for the migration
- Product.equals now uses a transient collection of provided
  product UUIDs when loaded from the database to avoid circular
  references when a given set of products are loaded


Ceiu:
This PR will be updated once #3123 is merged

nikosmoum (approved):

PR #3126, opened 2021-09-08
1989981: Fix to avoid exceeding the overall query parameter limit

wolfdale:
Final query was exceeding the overall query parameter limit for /products/owners endpoint

Changes -
- Broke the query to avoid exceeding the limit for IN clause & overall query parameter limit.

### DIFF
--- a/src/main/java/org/candlepin/model/OwnerCurator.java
+++ b/src/main/java/org/candlepin/model/OwnerCurator.java
@@ -16,32 +16,41 @@ package org.candlepin.model;
 
 import org.candlepin.controller.OwnerContentAccess;
 
+import com.google.common.collect.Iterables;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 
 import org.apache.commons.lang3.StringUtils;
-import org.hibernate.Criteria;
 import org.hibernate.LockMode;
 import org.hibernate.LockOptions;
 import org.hibernate.ReplicationMode;
-import org.hibernate.Session;
 import org.hibernate.criterion.DetachedCriteria;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Property;
 import org.hibernate.criterion.Restrictions;
 import org.hibernate.criterion.Subqueries;
-import org.hibernate.sql.JoinType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+
 
 import javax.inject.Singleton;
+import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
 import javax.persistence.TypedQuery;
-
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Expression;
+import javax.persistence.criteria.Join;
+import javax.persistence.criteria.JoinType;
+import javax.persistence.criteria.Predicate;
+import javax.persistence.criteria.Root;
 
 /**
  * OwnerCurator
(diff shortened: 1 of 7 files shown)


Ceiu (changes requested):

src/main/java/org/candlepin/model/OwnerCurator.java
```diff
@@ -199,45 +207,59 @@ public Owner getByUpstreamUuid(String upstreamUuid) {
...
+        Join<Product, Product> derivedProducts = product.join(Product_.derivedProduct, JoinType.LEFT);
+        Join<Product, Product> derivedProvidedProducts =
+            derivedProducts.join(Product_.providedProducts, JoinType.LEFT);
+
+        Expression<String> productExpr = product.get(Product_.id);
+        Expression<String> providedProductsExpr = providedProducts.get(Product_.id);
+        Expression<String> derivedProductsExpr = derivedProducts.get(Product_.id);
+        Expression<String> derivedPPExpr = derivedProvidedProducts.get(Product_.id);
+
+        int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit());
```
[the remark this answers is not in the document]
wolfdale: Currently we have ```IN_OPERATOR_BLOCK_SIZE, "15000"``` & ```QUERY_PARAMETER_LIMIT, "32000"``` in our code, here we are selecting the minimum of these two, I think we are on the safer side. Right ?

src/main/java/org/candlepin/model/OwnerCurator.java
```diff
@@ -199,45 +207,59 @@ public Owner getByUpstreamUuid(String upstreamUuid) {
...
-                return this.cpQueryFactory.buildQuery(session, criteria);
+    public List<Owner> getOwnersWithProducts(Collection<String> productIds) {
+        if (productIds == null || productIds.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        CriteriaBuilder criteriaBuilder = this.entityManager.get().getCriteriaBuilder();
+        CriteriaQuery<String> query = criteriaBuilder.createQuery(String.class);
+        Root<Owner> owner = query.from(Owner.class);
+        query.distinct(true);
```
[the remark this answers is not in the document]
wolfdale: With these updated queries, distinct restriction is applied on the SQL query directly (verified from the debug logs). The first query fetches the distinct owner IDs & second query is for fetching the owner entities. Should we merge these two into a single query?

Since we are not doing any pagination, we are safe from the hibernate returning incorrect data when we deal with fetching distinct IDs with joins & pagination.  
I have added the comment/note back, explaining the issue with pagination for future reference.

src/main/java/org/candlepin/model/OwnerCurator.java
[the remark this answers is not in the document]
nikosmoum: I think what Chris means is that the parameter limit right now is used to calculate the size of the block (let's say 32000 for the first block), but then the block is being added in 4 different IN clauses (productExpr, providedProductsExpr, derivedProductsExpr, derivedPPExpr), which means there will be in total 4 x 32000 parameters, while what we want is the total number of parameters in the whole query to be below 32000. So perhaps we should be targeting `this.getQueryParameterLimit() / 4` as the limit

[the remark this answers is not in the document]
wolfdale: @nikosmoum Got it. Changes made.

src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -120,13 +120,14 @@ public ProductCertificateDTO getProductCertificate(String productUuid) {
...
+    public Iterable<OwnerDTO> getProductOwners(List<String> productUuids) {
         if (productUuids.isEmpty()) {
             throw new BadRequestException(i18n.tr("No product IDs specified"));
         }
 
-        return this.translator.translateQuery(
-            this.ownerCurator.getOwnersWithProducts(productUuids), OwnerDTO.class);
+        List<Owner> owners = this.ownerCurator.getOwnersWithProducts(productUuids);
+        return owners.stream()
+            .map(this.translator.getStreamMapper(Owner.class, OwnerDTO.class)).collect(Collectors.toList());
```
Ceiu, line 130: We don't need to collect this into a list -- we can output a stream of DTOs.

src/main/java/org/candlepin/model/OwnerCurator.java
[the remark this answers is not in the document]
Ceiu: If DISTINCT is being applied properly, then the original issue is resolved and we can combine the two queries back into one -- especially since we're partitioning the queries and collecting the inputs already.

Another thing to consider here is that pagination is busted as soon as we break the query up at all (as we're doing with the partitioning); we need to either sort the product ID collection before we partition, or sort the output list after we run through all of the blocks.

src/main/java/org/candlepin/resource/ProductResource.java
wolfdale: Changed it to Iterable, along with new Pagination logic.

src/main/java/org/candlepin/model/OwnerCurator.java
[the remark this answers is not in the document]
wolfdale: @Ceiu @nikosmoum 
The original return type for query was ```CandlepinQuery``` which means pagination need to be supported with updated queries as well if someone tries to access endpoint with PageRequest. 
To support pagination we need fetch the total number of records, which is to be set as ```maxRecords``` in PageRequest. Here instead of getting the count, we fetch distinct OwnerIds (with JPA the distinct keyword is the part of SQL query) & use this in second query to fetch the owner entities.

With updated queries there could be one potential issue [here](https://github.com/candlepin/candlepin/blob/6e99912c80b1b549dbdceb5802bd92ff3a21bad9/src/main/java/org/candlepin/model/OwnerCurator.java#L231) as in order to support the ```sort_by``` functionality of pagination, we need to pass all the ownerIDs to IN clause by using partition which could still exceed the parameter limits. One solution here could be to break the second query again to fetch sub-set of a page, and later merge the results into single page. What are you suggestions ?

src/main/java/org/candlepin/model/OwnerCurator.java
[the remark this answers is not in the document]
Ceiu: `Math.min(this.getInBlockSize(), this.getQueryParameterLimit() / 4);`

src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -142,15 +156,21 @@ public ProductCertificateDTO getProductCertificate(String productUuid) {
             return null;
         }
 
-        ResultIterator<Owner> iterator = this.ownerCurator.getOwnersWithProducts(productUuids).iterate();
+        // PageRequest set to null to get all the records
+        PageRequest pageRequest = null;
```
Ceiu, line 160: This is a sign that our abstraction is leaky, and why we should avoid the legacy paging stuff wherever possible.

Ceiu (changes requested):

src/main/java/org/candlepin/model/OwnerCurator.java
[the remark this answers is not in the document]
wolfdale: Done

src/main/java/org/candlepin/model/OwnerCurator.java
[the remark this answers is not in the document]
wolfdale: Thanks for detailed explanation, since endpoint was never intended to explicitly support paging, we are okay to not support it. I can check with Katello side to verify that they are not calling API with such request. For now I have removed the paging bits, & used simple partition technique to get the work done.

src/main/java/org/candlepin/resource/ProductResource.java
wolfdale: Removed.

src/main/java/org/candlepin/model/OwnerCurator.java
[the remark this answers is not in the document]
nikosmoum: We need to verify with both Katello and Teamnado that they're not using pagination. If anyone was using pagination, I'd expect that to be Teamndo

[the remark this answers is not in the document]
wolfdale: @nikosmoum I found Katello is not using this API in their codebase as of now. Teamnado have used this endpoint (without pagination) in the past & actively wants to use this for errata notification service, they don't require this API to be paginated as they didn't face any issues in the past.

src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -142,15 +143,14 @@ public ProductCertificateDTO getProductCertificate(String productUuid) {
             return null;
         }
 
-        ResultIterator<Owner> iterator = this.ownerCurator.getOwnersWithProducts(productUuids).iterate();
+        Iterator<Owner> iterator = this.ownerCurator.getOwnersWithProducts(productUuids).iterator();
```
Ceiu, line 146: Sorry for noticing this late, but we can turn all of this into a stream with a couple mappers and/or functions:

```
        Function<Owner, JobConfig> jobConfigMapper = (owner) -> {
            return RefreshPoolsJob.createJobConfig()
                .setOwner(owner)
                .setLazyRegeneration(lazyRegen);
        };

        Function<JobConfig, AsyncJobStatus> jobQueueMapper = (config) -> {
            try {
                return this.jobManager.queueJob(config);
            }
            catch (Exception e) {
                log.debug("Exception occurred while queueing job: {}", config, e);

                return new AsyncJobStatus()
                    .setName(RefreshPoolsJob.JOB_NAME)
                    .setState(AsyncJobStatus.JobState.FAILED)
                    .setJobResult(e.toString());
            }
        }

        return this.ownerCurator.getOwnersWithProducts(productUuids)
            .stream()
            .map(jobConfigMapper)
            .map(jobQueueMapper)
            .map(this.translator.getStreamMapper(AsyncJobStatus.class, AsyncJobStatusDTO.class));
```

src/main/java/org/candlepin/model/OwnerCurator.java
```diff
@@ -189,55 +199,55 @@ public Owner getByUpstreamUuid(String upstreamUuid) {
...
+        Expression<String> derivedProductsExpr = derivedProducts.get(Product_.id);
+        Expression<String> derivedPPExpr = derivedProvidedProducts.get(Product_.id);
+
+        // Block size reduced by 4 times to avoid exceeding the parameters limit as there
+        // are multiple IN clauses in the query below.
+        int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit() / 4);
+        Set<Owner> owners = new HashSet<>();
+
+        for (List<String> block : Iterables.partition(productIds, blockSize)) {
+            List<Predicate> predicates = new LinkedList<>();
```
Ceiu, line 241: Since we have a fixed number of predicates to add to the query, we can avoid the list->array juggling by just building it all at once:

```
Predicate predicate = criteriaBuilder.or(productExpr.in(block), providedProductsExpr.in(block), ..., ...);
query.where(predicate);
```

src/main/java/org/candlepin/model/OwnerCurator.java
```diff
@@ -189,55 +199,55 @@ public Owner getByUpstreamUuid(String upstreamUuid) {
...
+            predicates.add(productExpr.in(block));
+            predicates.add(providedProductsExpr.in(block));
+            predicates.add(derivedProductsExpr.in(block));
+            predicates.add(derivedPPExpr.in(block));
+            query.where(criteriaBuilder.or(predicates.toArray(new Predicate[0])));
+            owners.addAll(this.entityManager.get().createQuery(query).getResultList());
```
Ceiu, line 247: Fetching the entity manager isn't free, and shouldn't be retrieved directly like this. Instead grab it out of the loop using the `getEntityManager` method and store it to be used here.

src/main/java/org/candlepin/model/OwnerCurator.java
```diff
@@ -189,55 +199,55 @@ public Owner getByUpstreamUuid(String upstreamUuid) {
...
         }
 
-        return this.cpQueryFactory.<Owner>buildQuery();
+        return new ArrayList<>(owners);
```
Ceiu, line 250: What is the goal of changing the output type to a list when we're already collecting the data in a set? We're already not maintaining ordering at all if a set is involved, so there's little value to be had by converting it to a list for output.

Ceiu (changes requested):

src/main/java/org/candlepin/resource/ProductResource.java
wolfdale: Done.

src/main/java/org/candlepin/model/OwnerCurator.java
wolfdale: Sure, done.

src/main/java/org/candlepin/model/OwnerCurator.java
wolfdale: Sure, changes made.

src/main/java/org/candlepin/model/OwnerCurator.java
wolfdale: Changes made to directly return set.

Ceiu (approved):

Ceiu:
Changes are good with me. I'll leave the merging to @nikosmoum in case he had any final thoughts.

PR #3127, opened 2021-09-08
Deserialization of LinkedHashSet in OpenAPI to preserve order

sonalidhome:
Currently, We are facing an issue with the insertion order of arrays
with uniqueItems=true, as they are not preserving the order. 
The issue mentioned is [issue](https://github.com/OpenAPITools/openapi-generator/issues/10167).
To solve this issue, added the `@JsonDeserialize(as = LinkedHashSet.class)`
on setter method.

Note: I am not able to create such a scenario in our code for spec test/unit test.
           But this works for the Upcoming PR of having multiple environments on consumer.

### DIFF
--- a/buildSrc/src/main/java/org/openapitools/codegen/languages/AbstractJavaJAXRSServerCodegen.java
+++ b/buildSrc/src/main/java/org/openapitools/codegen/languages/AbstractJavaJAXRSServerCodegen.java
@@ -105,6 +105,8 @@ public abstract class AbstractJavaJAXRSServerCodegen extends AbstractJavaCodegen
             setUseTags(convertPropertyToBoolean(USE_TAGS));
         }
 
+        importMapping.put("JsonDeserialize", "com.fasterxml.jackson.databind.annotation.JsonDeserialize");
+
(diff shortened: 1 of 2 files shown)


nikosmoum:
@sonalidhome I thought we had to add `AbstractJavaJAXRSServerCodegen.java ` in our codebase, which was why I mentioned we need 2 commits (one to add it, and one to make our changes). But since it's already there in our codebase we don't need 2 commits

Ceiu:
How are we hitting this issue? I feel like we may be focusing too much on a specific problem instead of how/which collections we're using.

nikosmoum (commented):
Posting some conclusions from IRC discussions:

Even though this workaround will work for preserving order, using any hash-based data structure (HashSet, LinkedHashSet, HashMap) would cause another problem with a use case we had not considered before: if the user sent duplicate items (e.g. 'a, b, c, a'), one of the duplicates would be silently dropped.

What we should probably do is use an ArrayList, which will both a) preserve insertion order, and b) allow duplicates, so that we can explicitly iterate it and if we find duplicates throw an error.

Given that we don't want to use a LinkedHashSet, then I'd say applying this workaround is unneeded, so I'll close this PR

PR #3128, opened 2021-09-10
1980404: No consumer deletion cleanup when deleting Environment

Januson:
- Added a cleanup to bulk revoke identity and content access
  certificates of the consumers deleted with the environment

### DIFF
--- a/src/main/java/org/candlepin/resource/EnvironmentResource.java
+++ b/src/main/java/org/candlepin/resource/EnvironmentResource.java
@@ -31,16 +31,19 @@ import org.candlepin.exceptions.BadRequestException;
 import org.candlepin.exceptions.ConflictException;
 import org.candlepin.exceptions.IseException;
 import org.candlepin.exceptions.NotFoundException;
-import org.candlepin.guice.PrincipalProvider;
 import org.candlepin.model.AsyncJobStatus;
-import org.candlepin.model.CandlepinQuery;
+import org.candlepin.model.CertificateSerialCurator;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.ConsumerCurator;
 import org.candlepin.model.Content;
+import org.candlepin.model.ContentAccessCertificate;
+import org.candlepin.model.ContentAccessCertificateCurator;
 import org.candlepin.model.Environment;
 import org.candlepin.model.EnvironmentContent;
 import org.candlepin.model.EnvironmentContentCurator;
 import org.candlepin.model.EnvironmentCurator;
+import org.candlepin.model.IdentityCertificate;
+import org.candlepin.model.IdentityCertificateCurator;
 import org.candlepin.model.OwnerContentCurator;
 import org.candlepin.resource.validation.DTOValidator;
 import org.candlepin.util.RdbmsExceptionTranslator;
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -145,21 +147,52 @@ public void deleteEnvironment(@Verify(Environment.class) String envId) {
...
+        for (Consumer c : consumers) {
             log.info("Deleting consumer: {}", c);
 
             // We're about to delete these consumers; no need to regen/dirty their dependent
             // entitlements or recalculate status.
             this.poolManager.revokeAllEntitlements(c, false);
             this.consumerCurator.delete(c);
         }
+        revokeIdentityCertificates(consumers);
+        revokeContentAccessCertificates(consumers);
```
nikosmoum, line 173: Both of these are only revoking the serials, but what we want is to entirely delete the certificates _and_ revoke the serials

src/test/java/org/candlepin/resource/EnvironmentResourceTest.java
```diff
@@ -0,0 +1,217 @@
...
+        this.environmentResource.deleteEnvironment(BAD_ID);
+
+        verifyZeroInteractions(this.poolManager);
+        verifyZeroInteractions(this.consumerCurator);
+        verifyZeroInteractions(this.certificateSerialCurator);
+        verify(this.envCurator).delete(eq(this.environment1));
+    }
+
+    @Test
+    void shouldCleanUpAfterDeletingEnvironment() {
```
nikosmoum, line 170: Great job adding unit tests, especially since there wasn't a unit test class for this resource, but let's also add a spec test too

nikosmoum (changes requested):

src/main/java/org/candlepin/resource/EnvironmentResource.java

src/test/java/org/candlepin/resource/EnvironmentResourceTest.java
Januson: Done.

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -145,21 +155,66 @@ public void deleteEnvironment(@Verify(Environment.class) String envId) {
...
 
+    private void deleteConsumers(Environment environment, List<Consumer> consumers) {
+        if (consumers.isEmpty()) {
+            log.info("No consumers found in environment: {}", environment);
+            return;
+        }
         // Cleanup all consumers and their entitlements:
         log.info("Deleting consumers in environment {}", environment);
-        for (Consumer c : consumers.list()) {
+        for (Consumer c : consumers) {
```
Ceiu, line 172: If we're already iterating over each consumer, it seems like it'd be best to pull the necessary cert IDs here, rather than doing multiple individual iterations for each type and serial.

Ceiu (changes requested):

src/main/java/org/candlepin/resource/EnvironmentResource.java
Januson: Done.

nikosmoum (approved):
ack form me

Ceiu (approved):

PR #3129, opened 2021-09-14
[F] Support multiple environments on consumers [ENT-3896]

sonalidhome:
- Added environments in ConsumerDTO/ConsumerDTOArrayElement and
   environmentIds in Consumer Entity.
 - Updated the `PUT /consumers/{consumer_uuid}`
   and the `POST /consumers` endpoint to support
   multiple environment additions from the DTO.
 - Added and updated the spec tests/ unit tests.
 - Adding/removing/re-prioritizing enviornments
   is done on consumer update.

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -8493,6 +8493,10 @@ components:
                   $ref: '#/components/schemas/ConsumerActivationKeyDTO'
               serviceType:
                 type: string
+              environments:
+                type: array
+                items:
+                  $ref: '#/components/schemas/EnvironmentDTO'
 
     ConsumerActivationKeyDTO:
       description: Represent activation keys used by consumer in registration
@@ -8535,8 +8539,6 @@ components:
               $ref: '#/components/schemas/ReleaseVerDTO'
             owner:
               $ref: '#/components/schemas/NestedOwnerDTO'
-            environment:
-              $ref: '#/components/schemas/EnvironmentDTO'
             entitlementCount:
               type: integer
               format: int64
--- a/spec/consumer_resource_spec.rb
+++ b/spec/consumer_resource_spec.rb
@@ -31,7 +31,7 @@ describe 'Consumer Resource' do
         @owner2['key'],
         [],
         [],
-        nil,
+        [],
         [],
         nil,
         [],
@@ -57,7 +57,7 @@ describe 'Consumer Resource' do
         @owner2['key'],
         [],
         [],
-        nil,
+        [],
         [],
         nil,
         [],
@@ -174,9 +174,8 @@ describe 'Consumer Resource' do
     user_cp = user_client(@owner1, random_string('billy'))
     consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [])
     consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
-
     expect{
-      consumer_client.update_consumer({:environment => {:name => "abc"}})
+      consumer_client.update_consumer({:environments => [{ 'name' => 'abc'}]})
     }.to raise_error(RestClient::ResourceNotFound)
   end
 
@@ -187,12 +186,12 @@ describe 'Consumer Resource' do
       consumer['idCert']['key'])
     env = @cp.create_environment(@owner1['key'], random_string("env_id"), random_string("env_name"))
 
-    expect(consumer.environment).to be_nil
+    expect(consumer.environments).to be_nil
 
-    consumer_client.update_consumer({:environment => {:name => env.name}})
+    consumer_client.update_consumer({:environments => [{:name => env.name}]})
     consumer = @cp.get_consumer(consumer['uuid'])
 
-    expect(consumer.environment).to_not be_nil
+    expect(consumer.environments).to_not be_nil
   end
 
   it 'should let consumer update environment with valid env id only' do
@@ -201,12 +200,12 @@ describe 'Consumer Resource' do
     consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
     env = @cp.create_environment(@owner1['key'], random_string("env_id"), random_string("env_name"))
 
-    expect(consumer.environment).to be_nil
+    expect(consumer.environments).to be_nil
 
-    consumer_client.update_consumer({:environment => {:id => env.id}})
+    consumer_client.update_consumer({:environments => [{"id" => env.id}]})
     consumer = @cp.get_consumer(consumer['uuid'])
 
-    expect(consumer.environment).to_not be_nil
+    expect(consumer.environments).to_not be_nil
   end
 
   it 'should let not consumer update environment with incorrect env id' do
@@ -214,10 +213,9 @@ describe 'Consumer Resource' do
     consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [])
     consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
 
-    expect(consumer.environment).to be_nil
-
+    expect(consumer.environments).to be_nil
     expect{
-      consumer_client.update_consumer({:environment => {:id => 'incorrect_id'}})
+      consumer_client.update_consumer({:environments => [{ 'id' => 'incorrect_id'}]})
     }.to raise_error(RestClient::ResourceNotFound)
 
   end
@@ -433,7 +431,7 @@ describe 'Consumer Resource' do
   it "should let a consumer register with a hypervisorId" do
     some_owner = create_owner(random_string('someowner'))
     client = user_client(some_owner, random_string('bob'))
-    consumer = client.register(random_string('system1'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], nil, [], "aBcD")
+    consumer = client.register(random_string('system1'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], [], [], "aBcD")
     # hypervisorId should always be set to lower case for the database constraint
     consumer['hypervisorId']['hypervisorId'].should == "abcd"
   end
@@ -442,7 +440,7 @@ describe 'Consumer Resource' do
     some_owner = create_owner(random_string('someowner'))
     client = user_client(some_owner, random_string('bob'))
     tags = [ "awesomeos", "awesomeos-workstation", "otherproduct" ]
-    consumer = client.register(random_string('system1'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], nil, [], nil, tags)
+    consumer = client.register(random_string('system1'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], [], [], nil, tags)
     consumer['contentTags'].should =~ tags
   end
 
@@ -450,7 +448,7 @@ describe 'Consumer Resource' do
     some_owner = create_owner(random_string('someowner'))
     client = user_client(some_owner, random_string('bob'))
     annotations = "here is a piece of information, here is another piece."
-    consumer = client.register(random_string('system1'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], nil, [], nil, nil, nil, nil, annotations)
+    consumer = client.register(random_string('system1'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], [], [], nil, nil, nil, nil, annotations)
     consumer['annotations'].should == annotations
   end
 
@@ -461,7 +459,7 @@ describe 'Consumer Resource' do
     created_date = '2015-05-09T13:23:55+0000'
     checkin_date = '2015-05-19T13:23:55+0000'
     consumer = client.register(random_string('system'), type=:system, nil, {}, user_name,
-              owner['key'], [], [], nil, [], nil, [], created_date, checkin_date)
+              owner['key'], [], [], [], [], nil, [], created_date, checkin_date)
     consumer['lastCheckin'].should == checkin_date
     consumer['created'].should == created_date
     #reload to be sure it was persisted
@@ -484,7 +482,7 @@ describe 'Consumer Resource' do
       client = user_client(owner, user_name)
 
       consumer = client.register(random_string('system'), type=:candlepin, nil, {}, user_name, owner['key'],
-        [], [], nil, [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, nil, nil, nil,
+        [], [], [], [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, nil, nil, nil,
         consumer_input_mode)
 
       expect(consumer).to_not be_nil
@@ -507,7 +505,7 @@ describe 'Consumer Resource' do
 
       expect {
         client.register(random_string('system'), type=:candlepin, nil, {}, user_name, owner['key'],
-          [], [], nil, [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, nil, nil, nil,
+          [], [], [], [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, nil, nil, nil,
           consumer_input_mode)
       }.to raise_exception(RestClient::BadRequest)
     end
@@ -520,7 +518,7 @@ describe 'Consumer Resource' do
       client = user_client(owner, user_name)
 
       consumer = client.register(random_string('system'), type=:system, nil, {}, user_name, owner['key'],
-        [], [], nil, [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, nil, nil, nil,
+        [], [], [], [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, nil, nil, nil,
         consumer_input_mode)
 
       expect(consumer).to_not be_nil
@@ -536,7 +534,7 @@ describe 'Consumer Resource' do
 
       expect {
         consumer = client.register(random_string('system'), type=:system, nil, {}, user_name, owner['key'],
-          [], [], nil, [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, nil, nil, nil,
+          [], [], [], [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, nil, nil, nil,
           consumer_input_mode)
       }.to raise_exception(RestClient::BadRequest)
     end
@@ -587,7 +585,7 @@ describe 'Consumer Resource' do
     service_level = 'test_service_level'
 
     consumer = client.register(random_string('system'), type=:system, nil, {}, user_name,
-              owner['key'], [], [], nil, [], nil, [], nil, nil, nil, nil, nil, 0, nil, service_level)
+              owner['key'], [], [], [], [], nil, [], nil, nil, nil, nil, nil, 0, nil, service_level)
 
     expect(consumer['serviceLevel']).to eq(service_level)
 
@@ -602,7 +600,7 @@ describe 'Consumer Resource' do
     client = user_client(owner, user_name)
 
     consumer = client.register(random_string('system'), type=:system, nil, {}, user_name,
-                               owner['key'], [], [], nil, [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil,
+                               owner['key'], [], [], [], [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil,
                                nil, nil, nil, nil, false)
 
     expect(consumer['autoheal']).to eq(false)
@@ -620,7 +618,7 @@ describe 'Consumer Resource' do
     role = 'test_role'
 
     consumer = client.register(random_string('system'), type=:system, nil, {}, user_name,
-              owner['key'], [], [], nil, [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, role)
+              owner['key'], [], [], [], [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, role)
 
     expect(consumer['role']).to eq(role)
 
@@ -637,7 +635,7 @@ describe 'Consumer Resource' do
     usage = 'test_usage'
 
     consumer = client.register(random_string('system'), type=:system, nil, {}, user_name,
-              owner['key'], [], [], nil, [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, usage)
+              owner['key'], [], [], [], [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, usage)
 
     expect(consumer['usage']).to eq(usage)
 
@@ -654,7 +652,7 @@ describe 'Consumer Resource' do
     addons = ['test_addon-1', 'test_addon-2', 'test_addon-3']
 
     consumer = client.register(random_string('system'), type=:system, nil, {}, user_name,
-              owner['key'], [], [], nil, [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, addons)
+              owner['key'], [], [], [], [], nil, [], nil, nil, nil, nil, nil, 0, nil, nil, nil, nil, addons)
 
     expect(consumer['addOns']).to_not be_nil
     expect(consumer['addOns'].size).to eq(addons.size)
@@ -675,7 +673,7 @@ describe 'Consumer Resource' do
     created_date = '2015-05-09T13:23:55.968+0000'
     checkin_date = '2015-05-19T13:25:55.222+0000'
     consumer = client.register(random_string('system'), type=:system, nil, {}, user_name,
-              owner['key'], [], [], nil, [], nil, [], created_date, checkin_date)
+              owner['key'], [], [], [], [], nil, [], created_date, checkin_date)
     consumer = client.get_consumer(consumer['uuid'])
     consumer['lastCheckin'].should == '2015-05-19T13:25:55+0000'
     consumer['created'].should == '2015-05-09T13:23:55+0000'
@@ -684,12 +682,12 @@ describe 'Consumer Resource' do
   it "should not let a consumer register with a used hypervisorId in same the org" do
     some_owner = create_owner(random_string('someowner'))
     client = user_client(some_owner, random_string('bob'))
-    consumer = client.register(random_string('system1'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], nil, [], "aBcD")
+    consumer = client.register(random_string('system1'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], [], [], "aBcD")
     # hypervisorId should always be set to lower case for the database constraint
     consumer['hypervisorId']['hypervisorId'].should == "abcd"
 
     lambda do
-      client.register(random_string('system2'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], nil, [], "abCd")
+      client.register(random_string('system2'), :system, random_string("someuuid"), {}, random_string("uname"), some_owner['key'], [], [], [], [], "abCd")
     end.should raise_exception(RestClient::BadRequest)
   end
 
@@ -870,7 +868,7 @@ describe 'Consumer Resource' do
     user = user_client(@owner1, random_string('billy'))
     lambda do
       consumer = user.register(random_string('machine1'), :candlepin, nil,
-                               {}, nil, nil, [], [], nil, [], nil, [], nil, nil,
+                               {}, nil, nil, [], [], [], [], nil, [], nil, nil,
                                nil, nil, "RHSM/1.0 (cmd=subscription-manager)")
     end.should raise_exception(RestClient::BadRequest)
 
@@ -956,7 +954,7 @@ describe 'Consumer Resource' do
     user_cp = user_client(@owner1, random_string('billy'))
     consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [])
     consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
-    consumer1 = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [], nil, [], "hYpervisor")
+    consumer1 = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [], [], [], "hYpervisor")
     consumer_client1 = Candlepin.new(nil, nil, consumer1['idCert']['cert'], consumer1['idCert']['key'])
 
     consumer = consumer_client.get_consumer()
@@ -1553,6 +1551,91 @@ describe 'Consumer Resource' do
     }.to raise_exception(RestClient::BadRequest)
   end
 
+  it 'should create consumer with multiple environments' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    env1 = @cp.create_environment(@owner1['key'], random_string("env_id-1"), random_string("env_name-1"))
+    env2 = @cp.create_environment(@owner1['key'], random_string("env_id-2"), random_string("env_name-2"))
+
+    consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [], [{'id' => env1.id}, {'id' => env2.id}])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+    expect(consumer.environments.size).to eq(2)
+    expect(consumer.environments[0].name).to eq(env1.name)
+    expect(consumer.environments[1].name).to eq(env2.name)
+  end
+
+  it 'should not create consumer with nil environment in environments list' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    env1 = @cp.create_environment(@owner1['key'], random_string("env_id-1"), random_string("env_name-1"))
+
+    lambda do
+      consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [], [{'id' => env1.id}, {'id' => ''}])
+    end.should raise_exception(RestClient::ResourceNotFound)
+  end
+
+  it 'should update consumer with re-prioritized environments' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    env1 = @cp.create_environment(@owner1['key'], random_string("env_id-1"), random_string("env_name-1"))
+    env2 = @cp.create_environment(@owner1['key'], random_string("env_id-2"), random_string("env_name-2"))
+
+    consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+
+    consumer_client.update_consumer({:environments => [{:name => env1.name}, {:name => env2.name}]})
+    consumer = @cp.get_consumer(consumer['uuid'])
+    expect(consumer.environments[0].name).to eq(env1.name)
+    expect(consumer.environments[1].name).to eq(env2.name)
+
+    # re-prioritized the environments
+    consumer_client.update_consumer({:environments => [{:name => env2.name}, {:name => env1.name}]})
+    consumer = @cp.get_consumer(consumer['uuid'])
+    expect(consumer.environments[0].name).to eq(env2.name)
+    expect(consumer.environments[1].name).to eq(env1.name)
+  end
+
+  it 'should preserve other environment priorities on a consumer when an environment is removed' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    env1 = @cp.create_environment(@owner1['key'], random_string("env_id-1"), random_string("env_name-1"))
+    env2 = @cp.create_environment(@owner1['key'], random_string("env_id-2"), random_string("env_name-2"))
+    env3 = @cp.create_environment(@owner1['key'], random_string("env_id-3"), random_string("env_name-3"))
+
+
+    consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [],
+      [{'id' => env1.id}, {'id' => env2.id}, {'id' => env3.id}])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+
+    expect(consumer.environments[0].name).to eq(env1.name)
+    expect(consumer.environments[1].name).to eq(env2.name)
+    expect(consumer.environments[2].name).to eq(env3.name)
+
+    # Removed the environment 2 (env_id-2).
+    consumer_client.update_consumer({:environments => [{:name => env1.name}, {:name => env3.name}]})
+    consumer = @cp.get_consumer(consumer['uuid'])
+
+    #It should maintain the lowering priority for the rest of other environments.
+    expect(consumer.environments[0].name).to eq(env1.name)
+    expect(consumer.environments[1].name).to eq(env3.name)
+  end
+
+  it 'should return exception when environment is specified more than once' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    env1 = @cp.create_environment(@owner1['key'], random_string("env_id-1"), random_string("env_name-1"))
+    env2 = @cp.create_environment(@owner1['key'], random_string("env_id-2"), random_string("env_name-2"))
+
+    # Exception raised while creating consumer
+    expect {
+      user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [],
+        [{:id => env1.id}, {:id => env2.id}, {:id => env1.id}])
+    }.to raise_exception(RestClient::Conflict)
+
+    consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+
+    # Exception raised while updating consumer
+    expect {
+      consumer_client.update_consumer({:environments => [{:name => env1.name}, {:name => env2.name}, {:name => env1.name}]})
+    }.to raise_exception(RestClient::Conflict)
+  end
+
 end
 
 describe 'Consumer Resource Consumer Fact Filter Tests' do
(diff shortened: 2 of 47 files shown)


src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
```diff
@@ -137,8 +138,16 @@ public ConsumerDTO populate(ModelTranslator translator, Consumer source, Consume
                 dest.setOwner(owner != null ? translator.translate(owner, NestedOwnerDTO.class) : null);
             }
 
-            Environment environment = this.environmentCurator.getConsumerEnvironment(source);
-            dest.setEnvironment(translator.translate(environment, EnvironmentDTO.class));
+            if (source.getEnvironmentIds() != null) {
+                Collection<String> envIDs = source.getEnvironmentIds().values();
+                for (String id : envIDs) {
+                    dest.addEnvironments(translator
+                        .translate(this.environmentCurator.listById(id), EnvironmentDTO.class));
```
nikosmoum, line 145: Instead of iterating over the environment ids here, we should instead create a new method `environmentCurator.getConsumerEnvironments`  (similar to `environmentCurator.getConsumerEnvironment`), which accepts a consumer object and returns the list of enviornments they are in, and then use that new curator method here

src/main/java/org/candlepin/dto/api/v1/ConsumerArrayElementTranslator.java
```diff
@@ -138,7 +138,7 @@ public ConsumerDTOArrayElement populate(ModelTranslator translator,
             }
 
             Environment environment = this.environmentCurator.getConsumerEnvironment(source);
-            dest.setEnvironment(translator.translate(environment, EnvironmentDTO.class));
+            dest.addEnvironments(translator.translate(environment, EnvironmentDTO.class));
```
nikosmoum, line 141: Same here: we should be using a new environmentCurator method that returns all the environments of a consumer

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -778,58 +786,14 @@ public void setSystemPurposeStatusHash(String systemPurposeStatusHash) {
...
 
     /**
-     * Fetches the ID of the environment with which this consumer is associated. If the consumer is not
+     * Fetches the ID of the highest prioritized environment with which
+     * this consumer is associated. If the consumer is not
      * associated with an environment, this method returns null.
      *
      * @return the ID of the environment for this consumer
      */
     public String getEnvironmentId() {
```
nikosmoum, line 795: Let's add a //TODO here that explains that this method is temporary and should be removed later on, not to forget

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -1024,4 +988,32 @@ public Consumer setServiceType(String serviceType) {
         this.serviceType = serviceType;
         return this;
     }
+
+    public Map<String, String> getEnvironmentIds() {
+        return this.environmentIds;
```
nikosmoum, line 993: Let's use a MapView to wrap this map before returning it (see examples of how it's used in other entities like Product).
This will prevent people abusing this by doing `consumer.getEnvironmentIds().put("x", "y")`, while they should be using the `addEnvironment()` method instead

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -1024,4 +988,32 @@ public Consumer setServiceType(String serviceType) {
...
+    }
+
+    public Consumer setEnvironmentIds(Map<String, String> environmentIds) {
+        this.environmentIds = environmentIds;
```
nikosmoum, line 997: Let's use the same logic of adding map elements as other places (e.g. Product.setAttributes), by clearing the original map, and then re-populating it with the elements of the incoming map. Assigning the incoming map directly can be dangerous because someone that holds it's reference would be able to update it later on, which would be reflected on the map in the Consumer object

nikosmoum (changes requested):
A quick review for now. Will take another look tomorrow

api/candlepin-api-spec.yaml
```diff
@@ -8439,8 +8439,6 @@ components:
                 $ref: '#/components/schemas/ReleaseVerDTO'
               owner:
                 $ref: '#/components/schemas/NestedOwnerDTO'
-              environment:
```
Ceiu, line 8442: Are we able to remove this at this point? I think we'll need to support both `environment` and `environments` for some time due to older clients that won't be aware of the new multi-environment functionality.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1608,43 +1614,48 @@ else if (toUpdate.getContentAccessMode() != null) {
...
-        else if (environmentName != null) {
-            String envId = this.environmentCurator.
-                getEnvironmentIdByName(toUpdate.getOwnerId(), environmentName);
+                    if (environmentId == null) {
+                        throw new NotFoundException(i18n.tr(
+                            "Environment with name \"{0}\" could not be found.", environmentName));
+                    }
+                }
+                else {
+                    throw new NotFoundException(i18n.tr("Environment could not be found."));
```
Ceiu, line 1646: This should be a bit more specific to indicate that it's not that we couldn't find the environment, but that there was no environment to find. Something along the lines of the environment being malformed.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1608,43 +1614,48 @@ else if (toUpdate.getContentAccessMode() != null) {
...
+        if (environments != null) {
+            for (EnvironmentDTO environment : environments) {
+                String environmentId = environment == null ? null : environment.getId();
+                String environmentName = environment == null ? null : environment.getName();
 
-                if (!environmentCurator.exists(environmentId)) {
-                    throw new NotFoundException(i18n.tr(
-                        "Environment with ID \"{0}\" could not be found.", environmentId));
+                if (environmentId != null) {
+                    if (!environmentCurator.exists(environmentId)) {
```
Ceiu, line 1631: We may want to consider bulking these calls up into a single query. Do we have an upper limit on the number of environments a consumer is allowed to join? Do we have an expectation of the average number of environments per consumer?

These may be @jturel questions, but should be considered when doing one-off queries in loops.

src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
Ceiu, line 145: This needs to be done with a single query and translation, as we're already leaning into N+1 efficiency issues.

src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
```diff
@@ -137,8 +138,16 @@ public ConsumerDTO populate(ModelTranslator translator, Consumer source, Consume
...
+                }
+            }
+            else {
+                dest.setEnvironments(Collections.emptyList());
```
Ceiu, line 149: Is the specification such that we always output a list, or is null a valid value for the environments property?

src/main/java/org/candlepin/model/Consumer.java
Ceiu, line 993: The internal collection should not be returned directly. Hibernate uses proxy collections that can cause unintended modifications to an object if the collection is reused or modified externally.

We need to encapsulate this properly, both protecting the collection and hiding the implementation details of how we're storing this in the DB.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1608,43 +1614,48 @@ else if (toUpdate.getContentAccessMode() != null) {
...
-                throw new NotFoundException(i18n.tr(
-                    "Environment with name \"{0}\" could not be found.", environmentName));
-            }
+                if (validatedEnvIds.containsKey(environmentId)) {
+                    throw new ConflictException("Environment \"{0}\" already exists: " + environmentId);
+                }
 
-            if (toUpdate.getEnvironmentId() == null || !envId.equals(toUpdate.getEnvironmentId())) {
-                validatedEnvId = envId;
+                validatedEnvIds.put("" + validatedEnvIds.size(), environmentId);
```
Ceiu, line 1653: Don't use string concatenation for this type of logic; things like `String.valueOf` or `.toString` would be better options.

src/test/java/org/candlepin/dto/api/v1/ConsumerTranslatorTest.java
```diff
@@ -123,7 +123,7 @@ protected Consumer initSourceObject() {
         consumer.setServiceType("consumer_service_type");
         consumer.setReleaseVer(new Release("releaseVer"));
         consumer.setOwner(owner);
-        consumer.setEnvironment(environment);
+        consumer.addEnvironment(environment);
```
Ceiu, line 126: We should have multiple environments added here to ensure it translates more than just one.

api/candlepin-api-spec.yaml
Ceiu: Just checked the design doc: this needs to be retained, and the output needs to be adjusted to be a comma-delimited list of environment names.

Ceiu (changes requested):
We've deviated quite a bit from the design doc in this PR, and there are a number of areas in which we're going to be hitting the DB *a lot*, which will be very bad for performance.

Some coordination should be done with our primary stakeholders on the Sat side of things (@jturel would be a good contact here) to gather expectations of averages and worst-case, as well as performance requirements to determine whether or not we can get away with one-off queries, or if we need to be batching them. This is especially true for the consumer translation, as the nature of nested object translation is already going to hit the N+1 issue even with environments being bulk fetched, and it only gets worse if we are doing that such a lookup multiple times per consumer.

api/candlepin-api-spec.yaml
nikosmoum: yep. we should retain the field here, but the output adjustment will be in one of the other task in the epic

src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
nikosmoum: I don't think this was thought of during the specification, and looking at our various translators we're very inconsistent with what we do about collections (sometimes we set to null, sometimes empty collection). It looks like most of the time we set to null, so we should try to standardize that way?

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -305,10 +305,15 @@ public ConsumerDTO createConsumerInEnvironment(String envId, ConsumerDTO consume
         this.validator.validateCollectionElementsNotNull(consumer::getInstalledProducts,
             consumer::getGuestIds, consumer::getCapabilities);
 
-        Environment e = lookupEnvironment(envId);
-        consumer.setEnvironment(translator.translate(e, EnvironmentDTO.class));
+        List<EnvironmentDTO> toUpdateEnvSet = new ArrayList<>();
```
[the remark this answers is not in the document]
nikosmoum: All the changes in this method are unneeded, because this is the legacy API endpoint (subman/katello will start using the main POST /consumers endpoint when registering with environments). There will be a different set of changes needed as part of another task for backwards compatibility (supporting old submans that  will be sending the list of environments as a CSV list in the envId string).

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1653,6 +1664,8 @@ else if (environmentName != null) {
                 this.contentAccessManager.removeContentAccessCert(toUpdate);
             }
 
+            contentAccessManager.syncOwnerLastContentUpdate(owner);
```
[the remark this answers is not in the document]
nikosmoum: I think this was my fault mentioning this in the card; you're right it's not needed, since we're already deleting the consumer's cert it will be regenerated anyway

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1608,43 +1614,48 @@ else if (toUpdate.getContentAccessMode() != null) {
...
+                else {
+                    throw new NotFoundException(i18n.tr("Environment could not be found."));
+                }
 
-            if (envId == null) {
-                throw new NotFoundException(i18n.tr(
-                    "Environment with name \"{0}\" could not be found.", environmentName));
-            }
+                if (validatedEnvIds.containsKey(environmentId)) {
+                    throw new ConflictException("Environment \"{0}\" already exists: " + environmentId);
```
nikosmoum, line 1650: Let's clarify this a bit more by saying `Environment \"{0}\" specified more than once`.
Also, I think we should add a spec test for this use case

src/main/java/org/candlepin/resource/ConsumerResource.java
nikosmoum: My assumption would be the average would be in the single digits, but I shouldn't be assuming; we should make sure to find out about these first :+1:

src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
```diff
@@ -206,7 +215,7 @@ public ConsumerDTO populate(ModelTranslator translator, Consumer source, Consume
         else {
             dest.setReleaseVer(null);
             dest.setOwner(null);
-            dest.setEnvironment(null);
+            dest.setEnvironments(Collections.emptyList());
```
nikosmoum, line 218: _If we're going to standardize around null values_ for no environments, we should also make it null here

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -768,13 +770,17 @@ protected void populateEntity(Consumer entity, ConsumerDTO dto) {
             entity.setReleaseVer(new Release(dto.getReleaseVer().getReleaseVer()));
         }
 
-        if (dto.getEnvironment() != null) {
-            Environment env = environmentCurator.get(dto.getEnvironment().getId());
-            if (env == null) {
-                throw new NotFoundException(i18n.tr("Environment \"{0}\" could not be found.",
-                    dto.getEnvironment().getId()));
+        if (dto.getEnvironments() != null && dto.getEnvironments().size() > 0) {
```
nikosmoum, line 773: `.isEmpty` is more concise than `size() > 0`

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1608,43 +1614,48 @@ else if (toUpdate.getContentAccessMode() != null) {
      * @return boolean status whether any updates are made or not.
      */
     private boolean updateEnvironment(ConsumerDTO updated, Consumer toUpdate) {
```
nikosmoum, line 1616: Let's update the name of the method and javadoc to reflect that this method is now updating multiple environments

nikosmoum (changes requested):
A few more changes suggested; other than that the only things left to find out are the expectations around the average/upper numbers of environments per consumer, and whether or not we can avoid outputting environments for bulk endpoints (such as the ones that return `ConsumerDTOArrayElement`)

api/candlepin-api-spec.yaml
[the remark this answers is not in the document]
sonalidhome: Understood.  Retained the `environment`.

src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
sonalidhome: Done.

src/main/java/org/candlepin/model/Consumer.java
sonalidhome: Done.

src/main/java/org/candlepin/model/Consumer.java
sonalidhome: Done.

src/main/java/org/candlepin/model/Consumer.java
sonalidhome: Agreed. Done.

api/candlepin-api-spec.yaml
sonalidhome: Retained.

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -198,4 +199,10 @@ public String getEnvironmentIdByName(String ownerId, String envName) {
 
         return envId;
     }
+
+    public Environment listById(String id) {
+        DetachedCriteria criteria = this.createSecureDetachedCriteria()
```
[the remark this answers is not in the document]
sonalidhome: Not using this method anymore. Hence removed.

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -100,7 +100,8 @@ public Environment getConsumerEnvironment(Consumer consumer) {
      */
     public CandlepinQuery<Consumer> getEnvironmentConsumers(String environmentId) {
         DetachedCriteria criteria = DetachedCriteria.forClass(Consumer.class)
-            .add(Restrictions.eq("environmentId", environmentId));
+            .createAlias("environmentIds", "env")
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -557,7 +557,8 @@ private int countMatchesByFilters(AbstractHibernateObject<?> object, String obje
     public CandlepinQuery<Entitlement> listByEnvironment(String environmentId) {
         DetachedCriteria criteria = DetachedCriteria.forClass(Entitlement.class)
             .createCriteria("consumer")
-            .add(Restrictions.eq("environmentId", environmentId));
+            .createAlias("environmentIds", "env")
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/main/java/org/candlepin/resource/ConsumerResource.java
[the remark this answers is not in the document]
sonalidhome: Removed.

src/main/java/org/candlepin/resource/EnvironmentResource.java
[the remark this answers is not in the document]
sonalidhome: Reverted these changes and added a TODO for the next task.

src/main/java/org/candlepin/resource/ConsumerResource.java
sonalidhome: Done.

src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
sonalidhome: Yeah, gathered all environments and later translated.

src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
sonalidhome: Done.

src/main/java/org/candlepin/model/Consumer.java
sonalidhome: Now wrapped in MapView.

src/main/java/org/candlepin/resource/ConsumerResource.java
sonalidhome: Done.

src/test/java/org/candlepin/dto/api/v1/ConsumerTranslatorTest.java
sonalidhome: Done.

src/main/java/org/candlepin/resource/ConsumerResource.java
sonalidhome: Done.

src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
sonalidhome: Done.

src/main/java/org/candlepin/resource/ConsumerResource.java
sonalidhome: Done.

src/main/java/org/candlepin/resource/ConsumerResource.java
sonalidhome: Done.

src/main/resources/db/changelog/20210817145050-add-consumer-environments.xml
```diff
@@ -0,0 +1,23 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20210817145050-1" author="sdhome">
+        <createTable tableName="cp_consumer_environments">
+            <column name="cp_consumer_id" type="VARCHAR(32)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="environment_id" type="VARCHAR(255)">
```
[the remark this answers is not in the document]
sonalidhome: ack.

src/main/resources/db/changelog/20210817145050-add-consumer-environments.xml
```diff
@@ -0,0 +1,23 @@
...
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20210817145050-1" author="sdhome">
+        <createTable tableName="cp_consumer_environments">
+            <column name="cp_consumer_id" type="VARCHAR(32)">
+                <constraints nullable="false"/>
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/test/java/org/candlepin/dto/api/v1/ConsumerArrayElementTranslatorTest.java
```diff
@@ -116,7 +116,7 @@ protected Consumer initSourceObject() {
         consumer.setServiceType("consumer_service_type");
         consumer.setReleaseVer(new Release("releaseVer"));
         consumer.setOwner(owner);
-        consumer.setEnvironment(environment);
+        consumer.addEnvironment(environment);
```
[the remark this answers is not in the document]
sonalidhome: Done.

spec/consumer_resource_spec.rb
```diff
@@ -1553,6 +1551,46 @@ def verify_installed_pids(consumer, product_ids)
...
+    consumer = @cp.get_consumer(consumer['uuid'])
+
+    expect(consumer.environments[0].name).to eq(env1.name)
+    expect(consumer.environments[1].name).to eq(env2.name)
+
+    # re-prioritized the environments
+    consumer_client.update_consumer({:environments => [{:name => env2.name}, {:name => env1.name}]})
+    consumer = @cp.get_consumer(consumer['uuid'])
+    expect(consumer.environments[0].name).to eq(env2.name)
+    expect(consumer.environments[1].name).to eq(env1.name)
```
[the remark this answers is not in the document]
sonalidhome: Added a new spec test for this.

src/main/resources/db/changelog/20210817145050-add-consumer-environments.xml
```diff
@@ -0,0 +1,23 @@
...
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20210817145050-1" author="sdhome">
+        <createTable tableName="cp_consumer_environments">
+            <column name="cp_consumer_id" type="VARCHAR(32)">
```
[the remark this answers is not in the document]
sonalidhome: Added an Index.

If we create composite primary key only on cp_consumer_id and environment_id, It gives an exception:
`Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "cp_consumer_environments_pkey" already exists.` 

It works well when we have a uniqueness with the composite primary key on cp_consumer_id, environment_id, and priority.

src/test/java/org/candlepin/dto/api/v1/ConsumerTranslatorTest.java
```diff
@@ -226,8 +232,18 @@ protected void verifyOutput(Consumer source, ConsumerDTO dest, boolean childrenG
                 ConsumerType ctype = this.mockConsumerTypeCurator.getConsumerType(source);
                 this.consumerTypeTranslatorTest.verifyOutput(ctype, dest.getType(), true);
 
-                Environment environment = this.mockEnvironmentCurator.getConsumerEnvironment(source);
-                this.environmentTranslatorTest.verifyOutput(environment, dest.getEnvironment(), true);
+                if (source.getEnvironmentIds() != null) {
+                    for (Map.Entry<String, String> sourceEnvId : source.getEnvironmentIds().entrySet()) {
```
nikosmoum, line 236: This test should be verifying two extra things: first that the destination consumer has translated exactly as many environments as the source consumer had (see `ProductTranslatorTest.java` verifying Branding as an example),  and second that it should also check that the order in which they were translated was preserved (this is a unique use case for this feature which we should be testing)

spec/consumer_resource_spec.rb
```diff
@@ -1553,6 +1551,82 @@ def verify_installed_pids(consumer, product_ids)
...
+  it 'should return exception when environment is specified more than once' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    env1 = @cp.create_environment(@owner1['key'], random_string("env_id-1"), random_string("env_name-1"))
+    env2 = @cp.create_environment(@owner1['key'], random_string("env_id-2"), random_string("env_name-2"))
+
+    consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+
+    expect {
+      consumer_client.update_consumer({:environments => [{:name => env1.name}, {:name => env2.name}, {:name => env1.name}]})
```
nikosmoum, line 1626: Something I noticed: this restriction (not allowing duplicate environments), should not only be applied when updating a consumer, but also when registering a consumer. we should add this restriction in ConsumerResource and also enhance/add this test case here either in this same test or a separate test

spec/consumer_resource_spec.rb
```diff
@@ -1553,6 +1551,82 @@ def verify_installed_pids(consumer, product_ids)
...
   end
 
+  it 'should create consumer with multiple environments' do
+    user_cp = user_client(@owner1, random_string('billy'))
+    env1 = @cp.create_environment(@owner1['key'], random_string("env_id-1"), random_string("env_name-1"))
+    env2 = @cp.create_environment(@owner1['key'], random_string("env_id-2"), random_string("env_name-2"))
+
+    consumer = user_cp.register(random_string('system'), :system, nil, {}, nil, nil, [], [], [{'id' => env1.id}, {'id' => env2.id}])
+    consumer_client = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+    expect(consumer.environments.size).to eq(2)
```
nikosmoum, line 1561: The priority of environments should be preserved not only when updating a consumer, but also when registering. Let's add this assertion to the end of this test

spec/consumer_resource_spec.rb
```diff
@@ -1553,6 +1551,82 @@ def verify_installed_pids(consumer, product_ids)
...
+    expect(consumer.environments[1].name).to eq(env2.name)
+    # re-prioritized the environments
+    consumer_client.update_consumer({:environments => [{:name => env2.name}, {:name => env1.name}]})
+    consumer = @cp.get_consumer(consumer['uuid'])
+    expect(consumer.environments[0].name).to eq(env2.name)
+    expect(consumer.environments[1].name).to eq(env1.name)
+
+  end
+
+  it 'should update consumer with updated environments' do
```
nikosmoum, line 1593: Let's make the name of this test more descriptive of what it tests. Something like 'should preserve other environment priorities on a consumer when an environment is removed'

src/main/resources/db/changelog/20210817145050-add-consumer-environments.xml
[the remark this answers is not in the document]
nikosmoum: If you get this error, then it means we have an error somewhere in our logic that we should definitely investigate. This is exactly the error we should be getting if someone tries to insert more than one priority for the same consumer+environment pair, and we shouldn't be allowing that. We need to find out when and why this happens, and fix it, and add this composite key

nikosmoum (changes requested):
Some more improvements needed

Ceiu (changes requested):

api/candlepin-api-spec.yaml
```diff
@@ -8578,6 +8582,10 @@ components:
               type: string
             serviceType:
               type: string
+            environments:
+              type: array
+              items:
+                $ref: '#/components/schemas/EnvironmentDTO'
```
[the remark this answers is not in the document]
sonalidhome: Removed.

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -198,4 +204,43 @@ public String getEnvironmentIdByName(String ownerId, String envName) {
 
         return envId;
     }
+
+    /**
+     * Fetches the multiple environments for the specified consumer. If the consumer does not have a defined
+     * environments, this method returns null. If the consumer has an invalid environments,
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/test/java/org/candlepin/dto/api/v1/ConsumerTranslatorTest.java
sonalidhome: Done.

spec/consumer_resource_spec.rb
sonalidhome: Done. Added validations on creation as well.

spec/consumer_resource_spec.rb
sonalidhome: Done.

spec/consumer_resource_spec.rb
sonalidhome: Done.

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -1024,4 +994,39 @@ public Consumer setServiceType(String serviceType) {
         this.serviceType = serviceType;
         return this;
     }
+
+    public Map<String, String> getEnvironmentIds() {
```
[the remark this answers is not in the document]
sonalidhome: Done. Changed the accessor and mutator to accept and receive a list.

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -541,25 +541,31 @@ private int countMatchesByFilters(AbstractHibernateObject<?> object, String obje
...
      * @return
-     *  A CandlepinQuery to iterate over the entitlements in the specified environment
+     *  A List to iterate over the entitlements in the specified environment
      */
-    public CandlepinQuery<Entitlement> listByEnvironment(String environmentId) {
-        DetachedCriteria criteria = DetachedCriteria.forClass(Entitlement.class)
-            .createCriteria("consumer")
-            .add(Restrictions.eq("environmentId", environmentId));
+    public List<Entitlement> listByEnvironment(String environmentId) {
+        CriteriaBuilder cb = this.entityManager.get().getCriteriaBuilder();
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -98,11 +100,15 @@ public Environment getConsumerEnvironment(Consumer consumer) {
...
-    public CandlepinQuery<Consumer> getEnvironmentConsumers(String environmentId) {
-        DetachedCriteria criteria = DetachedCriteria.forClass(Consumer.class)
-            .add(Restrictions.eq("environmentId", environmentId));
+    public List<Consumer> getEnvironmentConsumers(String environmentId) {
 
-        return this.cpQueryFactory.<Consumer>buildQuery(this.currentSession(), criteria);
+        String jpql = "SELECT c FROM Consumer c " +
+            "JOIN c.environmentIds e " +
+            "WHERE e = :environmentId ";
+        Query query = this.getEntityManager().createQuery(jpql);
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/main/java/org/candlepin/dto/api/v1/ConsumerTranslator.java
```diff
@@ -137,8 +138,15 @@ public ConsumerDTO populate(ModelTranslator translator, Consumer source, Consume
                 dest.setOwner(owner != null ? translator.translate(owner, NestedOwnerDTO.class) : null);
             }
 
-            Environment environment = this.environmentCurator.getConsumerEnvironment(source);
-            dest.setEnvironment(translator.translate(environment, EnvironmentDTO.class));
+            if (source.getEnvironmentIds() != null) {
+                List<Environment> environments = environmentCurator.getConsumerEnvironments(source);
+                for (Environment environment : environments) {
+                    dest.addEnvironments(translator.translate(environment, EnvironmentDTO.class));
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1604,47 +1610,55 @@ else if (toUpdate.getContentAccessMode() != null) {
...
-                    throw new NotFoundException(i18n.tr(
-                        "Environment with ID \"{0}\" could not be found.", environmentId));
+                if (environmentId != null) {
+                    if (!environmentCurator.exists(environmentId)) {
+                        throw new NotFoundException(i18n.tr(
+                            "Environment with ID \"{0}\" could not be found.", environmentId));
+                    }
                 }
+                else if (environmentName != null) {
+                    environmentId = this.environmentCurator.
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -198,4 +204,43 @@ public String getEnvironmentIdByName(String ownerId, String envName) {
...
+     *  if consumer is null
+     *
+     * @throws IllegalStateException
+     *  if the consumer's defined environment ID is invalid
+     *
+     * @return
+     *  A list of Environments instance for the specified consumer, or null if the consumer does not have a
+     *  defined environments
+     */
+    public List<Environment> getConsumerEnvironments(Consumer consumer) {
```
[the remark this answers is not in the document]
sonalidhome: Updated to use JPQL.

src/main/java/org/candlepin/dto/api/v1/ConsumerArrayElementTranslator.java
```diff
@@ -137,8 +138,15 @@ public ConsumerDTOArrayElement populate(ModelTranslator translator,
                 dest.setOwner(owner != null ? translator.translate(owner, NestedOwnerDTO.class) : null);
             }
 
-            Environment environment = this.environmentCurator.getConsumerEnvironment(source);
-            dest.setEnvironment(translator.translate(environment, EnvironmentDTO.class));
+            if (source.getEnvironmentIds() != null) {
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/main/resources/db/changelog/20210817145050-add-consumer-environments.xml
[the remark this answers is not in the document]
sonalidhome: This is now resolved after making changes in `environmentIds` mutator i.e. `setEnvironmentIds`

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1002,6 +1008,8 @@ public Consumer createConsumerFromDTO(ConsumerDTO consumer, ConsumerType type, P
             Date createdDate = consumerToCreate.getCreated();
             Date lastCheckIn = consumerToCreate.getLastCheckin();
 
+            this.updateEnvironments(consumer, consumerToCreate, false);
```
nikosmoum, line 1011: Using this method for handling environments during consumer registration is fine, but you should remove the environment handling done in the populateEntity() method then, because now we're doing duplicate work.

nikosmoum (changes requested):
One thing to change, but otherwise looks good to me

src/main/java/org/candlepin/resource/ConsumerResource.java
sonalidhome: Done.

nikosmoum (approved):
ack from me; I'll leave @Ceiu to re-review/merge

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -305,8 +305,10 @@ public ConsumerDTO createConsumerInEnvironment(String envId, ConsumerDTO consume
         this.validator.validateCollectionElementsNotNull(consumer::getInstalledProducts,
             consumer::getGuestIds, consumer::getCapabilities);
 
+        //TODO: This needs to support backward compatibility.
+        // Kept as it is to be handled in other task of multi-environment feature.
         Environment e = lookupEnvironment(envId);
-        consumer.setEnvironment(translator.translate(e, EnvironmentDTO.class));
+        consumer.addEnvironments(translator.translate(e, EnvironmentDTO.class));
```
Ceiu, line 311: If the consumer DTO provided as part of this request already includes environments, we would add the environment requested explicitly as part of the resource to the end of that list and then do a normal creation, rather than creating the consumer in the single requested environment.

Is that expected behavior here?

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1604,57 +1598,83 @@ else if (toUpdate.getContentAccessMode() != null) {
...
+                    toUpdate.setContentAccessCert(null);
+                    this.contentAccessManager.removeContentAccessCert(toUpdate);
+                }
+
+                // lazily regenerate certs, so the client can still work
+                poolManager.regenerateCertificatesOf(toUpdate, true);
             }
 
-            // lazily regenerate certs, so the client can still work
-            poolManager.regenerateCertificatesOf(toUpdate, true);
```
Ceiu, line 1657: We still need to do this every time the environment(s) change, right? The certs will have environment-specific details in them, so I can't imagine we can get away from regenerating the certs when this changes.

@nikosmoum

Ceiu (changes requested):
Almost there

src/main/java/org/candlepin/resource/ConsumerResource.java
nikosmoum: Yes we need to, but only in case of a PUT request (which with the current state of the PR will still happen). But it doesn't have to be done during a POST (registration) call, since there's no certificates yet (this is the only path in which this is not run)

src/main/java/org/candlepin/resource/EnvironmentResource.java
nikosmoum: The clients calling this endpoint (legacy subman clients) have no  (and will not ever have) knowledge that you can set environments in the request body; but just to be sure, perhaps we should do a `.clear()` before adding the env from the URL param

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -1024,4 +995,48 @@ public Consumer setServiceType(String serviceType) {
         this.serviceType = serviceType;
         return this;
     }
+
+    public List<String> getEnvironmentIds() {
+        return this.environmentIds == null ? null :
+            this.environmentIds.values().stream().collect(Collectors.toList());
```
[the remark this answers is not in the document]
sonalidhome: Done.

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -1024,4 +995,48 @@ public Consumer setServiceType(String serviceType) {
...
+
+        // Creating a new map instance and assigning it to the environmentIds.
+        // This ensures that, we are deleting the existing entries and inserting the new entries.
+        // This is done to avoid constraint (cp_consumer_environments_pkey) violation exception.
+        // Whenever we try to update the existing data (changing priority), it actually updates
+        // the enviroment Id and not the priority.
+        Map<String, String> envMap = new LinkedHashMap<>();
+        if (environmentIds != null) {
+            for (String environmentId : environmentIds) {
+                envMap.put(String.valueOf(envMap.size()), environmentId);
```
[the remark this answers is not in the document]
sonalidhome: Added exception.

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -1024,4 +997,60 @@ public Consumer setServiceType(String serviceType) {
         this.serviceType = serviceType;
         return this;
     }
+
+    public List<String> getEnvironmentIds() {
+        return this.environmentIds.entrySet().stream()
+            .sorted(Comparator.comparingInt(lhs -> Integer.parseInt(lhs.getKey())))
```
Ceiu, line 1003: Pedantic request: "lhs" is usually used to indicate the left-hand side of a comparison. Since we're isolated from that due to the handy comparator function, let's just call it "entry" or something to indicate that it's working with the elements from the entry set.

Ceiu (changes requested):
One more thing. I promise I'm not trying to be difficult here. :)

src/main/java/org/candlepin/resource/EnvironmentResource.java
Ceiu: Oh, we still need to handle this. Probably by way of either .clear as nikos suggested, or by way of using .setEnvironments (which is my preferred method).

sonalidhome: Updated to use `.setEnvironments`. I hope this will be done in a more detailed manner in the next tasks.

src/main/java/org/candlepin/model/Consumer.java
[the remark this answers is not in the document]
sonalidhome: Agreed, good point. Updated code to use 2 loops, one for deduplication and another for adding entries to the map.

Ceiu (approved):

PR #3136, opened 2021-10-05
2010251: Update date check for content access certificate

wottop:


src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -648,13 +648,16 @@ public boolean hasCertChangedSince(Consumer consumer, Date date) {
         // has any real value.
 
         if (owner.isUsingSimpleContentAccess()) {
-            // Check if the owner's content view has changed since the date
-            if (!date.after(owner.getLastContentUpdate())) {
-                return true;
+            ContentAccessCertificate cert;
+            try {
+                cert = getCertificate(consumer);
```
nikosmoum, line 653: The only place where this method (hasCertChangedSince) is called from is ConsumerResource.getContentAccessBody, which itself is already calling the ContentAccessManager.getCertificate method. With this change, we will be calling the getCertificate method twice. This is a problem not only because we'll be calling it twice, but getCertificate could actually re-generate the cert (either the x509 part, or the payload, or both), which is at odds with what the current method (hasCertChangedSince) tries to do (check whether or not it has changed).

Instead, could we not entirely remove the hasCertChangedSince method, and add the logic that it's supposed to be doing around here?: https://github.com/candlepin/candlepin/blob/candlepin-4.0-HOTFIX/server/src/main/java/org/candlepin/resource/ConsumerResource.java#L1903 (before moving on to build the "ContentAccessListing result" object, check if the cert has changed and if not return 304 status?)

PR #3138, opened 2021-10-13
ENT-3898: Updated the old registration API for backwards compatibility

wolfdale:
Changes :
Updated the old registration API for backwards compatibility to support multi-env

### DIFF
--- a/client/ruby/candlepin_api.rb
+++ b/client/ruby/candlepin_api.rb
@@ -797,6 +797,18 @@ class Candlepin
     return delete(url, params)
   end
 
+  # This is a legacy registration endpoint
+  # (newer subscription-manager clients will not be using it)
+  def create_consumer_in_environments(env_ids, name, type=:system)
+    consumer = {
+        :type => {:label => type},
+        :name => name
+      }
+
+      params = {}
+      return post("/environments/#{env_ids}/consumers", params, consumer)
+  end
+
   def get_product_owners(product_ids)
     url = "/products/owners"
 
--- a/src/main/java/org/candlepin/resource/EnvironmentResource.java
+++ b/src/main/java/org/candlepin/resource/EnvironmentResource.java
@@ -63,6 +63,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.BinaryOperator;
+import java.util.stream.Collectors;
 
 import javax.persistence.PersistenceException;
 import javax.persistence.RollbackException;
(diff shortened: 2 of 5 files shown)


src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -306,12 +330,13 @@ public ConsumerDTO createConsumerInEnvironment(String envId, ConsumerDTO consume
         this.validator.validateCollectionElementsNotNull(consumer::getInstalledProducts,
             consumer::getGuestIds, consumer::getCapabilities);
 
-        //TODO: This needs to support backward compatibility.
-        // Kept as it is to be handled in other task of multi-environment feature.
-        Environment e = lookupEnvironment(envId);
-        consumer.setEnvironments(Arrays.asList(translator.translate(e, EnvironmentDTO.class)));
+        List<Environment> envs = processEnvironmentIds(envId);
+        consumer.setEnvironments(envs.stream()
```
Ceiu, line 334: Given that we're calling the process method in one place, and we're using streams to convert everything, we could combine it into one chain:

```
List<EnvironmentDTO> dtos = Arrays.stream(envId.split("\\s*,\\s*"))
    .map(EnvironmentResource::lookupEnvironment)
    .map(this.translator.getStreamMapper(Environment.class, EnvironmentDTO.class))

consumer.setEnvironments(dtos);
```

As an aside, I wonder if we shouldn't also verify that all of the environments fetched are in the requested owner. Now that we're just passing the requested owner key through, there's some risk of the owner and environment(s) not lining up if it's never validated.

Ceiu (changes requested):

nikosmoum (changes requested):

client/ruby/candlepin_api.rb
```diff
@@ -797,6 +797,17 @@ def demote_content(env_id, content_ids)
     return delete(url, params)
   end
 
+  def create_consumer_in_environments(env_ids, owner_key, name, type=:system)
```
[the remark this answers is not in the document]
wolfdale: Sure, done

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -300,18 +301,29 @@ public AsyncJobStatusDTO demoteContent(@Verify(Environment.class) String envId,
...
-        Environment e = lookupEnvironment(envId);
-        consumer.setEnvironments(Arrays.asList(translator.translate(e, EnvironmentDTO.class)));
+        List<EnvironmentDTO> environmentDTOs = Arrays.stream(envId.split("\\s*,\\s*"))
+            .map(this::lookupEnvironment)
+            .map(this.translator.getStreamMapper(Environment.class, EnvironmentDTO.class))
+            .collect(Collectors.toList());
+
+        // Check if all envs belongs to same org
+        boolean sameOwner = environmentDTOs.stream().map(env -> env.getOwner().getKey())
+            .distinct().count() <= 1;
```
nikosmoum, line 317: There shouldn't be a case where count is <0 here. All environments should belong to an owner, so perhaps this should be `== 1`

nikosmoum (changes requested):
One minor thing, but otherwise ack from me

src/main/java/org/candlepin/resource/EnvironmentResource.java
Ceiu: We're also doing a bunch of extra work by using distinct/count.

`.distinct` works internally by building a LinkedHashSet of the parent stream and then streaming its results (with more nuance than that, but that's the gist of it)

`.count` iterates until the stream is exhausted, very similar to a `for (Object obj : objects) ++count;` in cases where the spliterator backing it doesn't provide the exact size.

Given that we only care about the unique value and whether or not we encounter something that isn't equal to the first org we find, we could use a custom reducer to do both at once:

```
BinaryOperator<String> unify = (prev, next) -> {
    if (prev != null && !prev.equals(next)) {
        throw new BadRequestException(i18n.tr("Two or more environments belong to different organizations"));
    }

    return next;
};

String ownerKey = environmentDTOs.stream()
    .map(env -> env.getOwner().getKey())
    .reduce(null, unify);
```

PR #3142, opened 2021-10-26
[3.1] 1973257: Update method to mark entitlements dirty

wottop:
Should solve issue with performance and the concurrent
database errors as well


nikosmoum:
@wottop Let's port this to 4.0 and master

PR #3144, opened 2021-10-27
ENT-3900: Add support for multi-env entitlement cert regeneration

Januson:
- Generate entitlement cert using multiple environments
- Deduplicate promoted content
- Handle environment-specific content paths

Note: Due to the high amount of shared code I had to handle all content prefixes not just entitlement ones.

### DIFF
--- a/spec/environment_spec.rb
+++ b/spec/environment_spec.rb
@@ -1,8 +1,14 @@
 require 'spec_helper'
 require 'candlepin_scenarios'
 
+def promote_content(content1, env1)
+  job = @cp.promote_content(env1['id'], [{ :contentId => content1['id'] }])
+  wait_for_job(job['id'], 15)
+end
+
 describe 'Environments' do
   include CandlepinMethods
+  include CertificateMethods
 
   before(:each) do
     @expected_env_id = random_string('testenv1')
@@ -215,7 +221,6 @@ describe 'Environments' do
   it 'filters content not promoted to environment' do
     consumer = @org_admin.register(random_string('testsystem'), :system, nil, {},
         nil, nil, [], [], [{ 'id' => @env['id']}])
-    expect(consumer['environments']).not_to be_nil
     consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'],
       consumer['idCert']['key'])
 
@@ -233,7 +238,7 @@ describe 'Environments' do
         }])
     wait_for_job(job['id'], 15)
 
-    pool = create_pool_and_subscription(@owner['key'], product['id'], 10)
+    pool = @cp.create_pool(@owner['key'], product['id'])
     ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
 
     x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
@@ -248,7 +253,6 @@ describe 'Environments' do
 
   it 'regenerates certificates after promoting/demoting content' do
     consumer = @org_admin.register(random_string('testsystem'), :system, nil, {}, nil, nil, [], [], [{ 'id' => @env['id']}])
-    expect(consumer['environments']).not_to be_nil
     consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
 
     product = create_product
@@ -264,7 +268,7 @@ describe 'Environments' do
         }])
     wait_for_job(job['id'], 15)
 
-    pool = create_pool_and_subscription(@owner['key'], product['id'], 10)
+    pool = @cp.create_pool(@owner['key'], product['id'])
     ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
 
     x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
@@ -307,16 +311,12 @@ describe 'Environments' do
     # We'll use @env as env1 with nothing on it
 
     env2 = @cp.create_environment(@owner['key'], random_string("env2"), random_string("env2"))
-    job = @cp.promote_content(env2['id'], [{:contentId => content1['id']}])
-    wait_for_job(job['id'], 15)
-    job = @cp.promote_content(env2['id'], [{:contentId => content2['id']}])
-    wait_for_job(job['id'], 15)
+    promote_content(content1, env2)
+    promote_content(content2, env2)
 
     env3 = @cp.create_environment(@owner['key'], random_string("env3"), random_string("env3"))
-    job = @cp.promote_content(env3['id'], [{:contentId => content1['id']}])
-    wait_for_job(job['id'], 15)
-    job = @cp.promote_content(env3['id'], [{:contentId => content2['id']}])
-    wait_for_job(job['id'], 15)
+    promote_content(content1, env3)
+    promote_content(content2, env3)
     @cp.register(random_string("consumer1"), :system, nil, {}, random_string("consumer1"), @owner['key'], [], [], [{ 'id' => env3['id']}])
     @cp.register(random_string("consumer2"), :system, nil, {}, random_string("consumer2"), @owner['key'], [], [], [{ 'id' => env3['id']}])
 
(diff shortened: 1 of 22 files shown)


nikosmoum (changes requested):
A few things to change; will do another pass tomorrow

src/main/java/org/candlepin/service/impl/PromotedContent.java
```diff
@@ -0,0 +1,121 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * A collection of content promoted to environments.
+ */
+public class PromotedContent {
```
Ceiu, line 33: This class is a glorified wrapper around a hash map, which is fine, but we then use a factory method to utilize it rather than anything that would better fit its property of a wrapper, such as a fluent, builder-style API. This would keep the construction in context, which is implied to be intended by the first log message in the builder function.

```
PromotedContent promotedContent = new PromotedContent()
    .from(environment)
    .from(environment);
```

While I understand this design likely came about because of how we use it today (that is, we get the list of environments from the curator and throw it at the `.from` method), I don't think we should retain design choices from the past if we're going to spend the effort making a new abstraction layer to replace them.

This also eliminates the need for multiple constructors, and fully encapsulates the underlying map.

As a side note, the example above uses a single environment mutator since the order in which environments are added is critical to how this operates. That should be front-and-center with the design here, since it's so central to how this object is going to be used. A `.fromAll` could also work, but it'd need to be limited to order-retaining collections for that very reason.

src/main/java/org/candlepin/service/impl/PromotedContent.java
```diff
@@ -0,0 +1,121 @@
...
+        this.contents = contents;
+    }
+
+    /**
+     * Returns an id of environment to which is the given {@link ProductContent} promoted.
+     *
+     * @param pc a product content for which to find environment
+     * @return an environment id
+     */
+    public String environmentOf(ProductContent pc) {
```
Ceiu, line 82: Should be `environmentIdOf` since we don't actually get the whole environment.

Ceiu (changes requested):
Nikos started to cover most of the other stuff I was going to comment on, so I'll stop piling on and just add some general thoughts here:

Overall, given the complexity surrounding the prefixes and the way we pass content around here, I think we can get some good mileage out of the abstraction concept being implemented. However, I would like to see it fleshed out a bit more rather than having it adapted to the code as it exists today and not quite addressing the real design problems present in the original code.

As it is now, it took quite a bit of back-and-forth through the code to really understand why the objects were being added and what they were actually doing. But, with a bit of refinement, I think they could do a lot of good.

src/main/java/org/candlepin/util/Util.java
```diff
@@ -603,4 +605,8 @@ else if (temporalAccessor instanceof LocalDateTime) {
             return LocalDate.from(temporalAccessor).atStartOfDay().atOffset(ZoneOffset.UTC);
         }
     }
+
+    public static String encode(String text) {
```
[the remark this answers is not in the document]
Januson: Done.

spec/environment_spec.rb
```diff
@@ -378,4 +378,162 @@
...
+    env3 = @cp.create_environment(@owner['key'], random_string("env3"), random_string("env3"))
+    promote_content(content1, env1)
+    promote_content(content2, env2)
+    promote_content(content3, env3)
+    consumer = @cp.register(random_string("consumer1"), :system, nil,
+       facts={'system.certificate_version' => '3.3', "uname.machine" => "x86_64"}, random_string("consumer1"),
+       @owner['key'], [], [], [{ 'id' => env1['id']}, { 'id' => env2['id']}, { 'id' => env3['id']}])
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+
+    pool = create_pool_and_subscription(@owner['key'], product['id'], 10)
```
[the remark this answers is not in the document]
Januson: Done.

spec/environment_spec.rb
```diff
@@ -378,4 +378,162 @@
...
```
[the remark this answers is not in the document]
Januson: Done.

spec/environment_spec.rb
```diff
@@ -378,4 +378,162 @@
...
+    env3 = @cp.create_environment(@owner['key'], random_string("env3"), random_string("test_env_3"))
+    promote_content(content1, env1)
+    promote_content(content2, env2)
+    promote_content(content3, env3)
+    consumer = @cp.register(random_string("consumer1"), :system, nil,
+                            facts={'system.certificate_version' => '3.3', "uname.machine" => "x86_64"}, random_string("consumer1"),
+                            @owner['key'], [], [], [{ 'id' => env1['id']}, { 'id' => env2['id']}, { 'id' => env3['id']}])
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+
+    pool = create_pool_and_subscription(@owner['key'], product['id'], 10)
```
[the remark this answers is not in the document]
Januson: Done.

spec/environment_spec.rb
```diff
@@ -378,4 +378,162 @@
...
+    promote_content(content2, env1)
+    promote_content(content2, env2)
+    promote_content(content2, env3)
+    promote_content(content3, env3)
+    consumer = @cp.register(random_string("consumer1"), :system, nil,
+                            facts={'system.certificate_version' => '3.3', "uname.machine" => "x86_64"}, random_string("consumer1"),
+                            @owner['key'], [], [], [{ 'id' => env1['id']}, { 'id' => env2['id']}, { 'id' => env3['id']}])
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+
+    pool = create_pool_and_subscription(@owner['key'], product['id'], 10)
```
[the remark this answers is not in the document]
Januson: Done.

spec/environment_spec.rb
```diff
@@ -378,4 +378,162 @@
...
+    certs.length.should == 1
+
+    json_body = extract_payload(certs[0]['cert'])
+    json_body['products'].length.should == 1
+    json_body['products'][0]['content'].size.should == 3
+
+    env_names = [env1.name, env2.name, env3.name]
+    json_body['products'][0]['content'].each { |content|
+      env_names.any? { |env_name| content['path'].include? env_name }
+    }
```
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/service/impl/ContentPaths.java
```diff
@@ -0,0 +1,97 @@
...
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.service.impl;
```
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/service/impl/EntitlementContentPrefix.java
```diff
@@ -0,0 +1,117 @@
...
+
+    private final Map<String, String> prefixes = new HashMap<>();
+    private final String ownerPrefix;
+
+    /**
+     * A constructor
+     *
+     * @param owner an owner to be used in construction of content prefixes
+     */
+    public EntitlementContentPrefix(Owner owner) {
```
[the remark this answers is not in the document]
Januson: Done. I did not add overloaded version as there is no real use-case for prefix without environments.

src/main/java/org/candlepin/service/impl/ScaContentPrefix.java
```diff
@@ -0,0 +1,108 @@
...
+    private final String ownerKey;
+    private final boolean standalone;
+
+    /**
+     * A constructor
+     *
+     * @param owner an owner to be used in construction of content prefixes
+     * @param standalone flag denoting whether we are in hosted or standalone deployment
+     */
+    public ScaContentPrefix(Owner owner, boolean standalone) {
```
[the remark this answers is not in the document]
Januson: Done. I did not add overloaded version as there is no real use-case for prefix without environments.

src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
```diff
@@ -172,23 +170,20 @@ public EntitlementCertificate generateEntitlementCert(Entitlement entitlement, P
     public X509Certificate createX509Certificate(Consumer consumer, Owner owner, Pool pool,
         Entitlement ent, Product product, Set<Product> products,
         List<org.candlepin.model.dto.Product> productModels, BigInteger serialNumber,
-        KeyPair keyPair, boolean useContentPrefix)
+        KeyPair keyPair, ContentPaths contentPaths, PromotedContent promotedContent)
```
nikosmoum, line 173: @Januson @Ceiu Given that ContentPaths and PromotedContent are almost universally passed into methods together, and conceptually, paths are one field of the promoted content, would it not be a good idea to merge the ContentPaths into PromotedContent, and only pass a PromotedContent object around? Something like this:
```
class PromotedContent {
    private final ContentPrefix prefix;
    private final Map<String, EnvironmentContent> contents;
    public String getPath(ProductContent pc) {...};
     // + add all the methods that exist in the current PromotedContent class?
}
```

src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
```diff
@@ -399,29 +344,10 @@ private String getContentPrefix(Consumer consumer, Owner owner, boolean useConte
...
-
-        StringBuffer output = new StringBuffer("/");
-        for (String part : contentPrefix.split("/")) {
-            if (!part.equals("")) {
-                output.append(URLEncoder.encode(part, "UTF-8"));
-                output.append("/");
-            }
-        }
-        return output.toString().replace("%24", "$");
+        return v3extensionUtil.getByteExtensions(productModels);
```
nikosmoum, line 350: At this point, this method does nothing other than passing the single argument and returning, so we could remove it and just call the underlying method instead?

nikosmoum (changes requested):
I have only 2 minor requests & a suggestion, but otherwise this looks great :tada: :tada:

src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
Ceiu: Maybe? As pointless as it is, it's bridging the adapter interface for us. We may need to keep the implementation as it is, but we could probably change the code calling it to, instead, call the extension util method directly.

Otherwise, if we really want to remove it, I'd say we should look at all the cert service adapters and check with teamnado to make sure they don't have anything custom in this area. Katello probably as well, but I think we do all of the adapter management for them.

src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
Ceiu: I think I would prefer this overall; but I'm open to other options as well.

spec/environment_spec.rb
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/service/impl/PromotedContent.java
Januson: Done.

src/main/java/org/candlepin/service/impl/PromotedContent.java
Januson: Done.

src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
Januson: Done.

src/main/java/org/candlepin/controller/util/PromotedContent.java
```diff
@@ -0,0 +1,190 @@
...
+     * @param environments environments from which to collect promoted content
+     * @return an instance of {@link PromotedContent}
+     */
+    public PromotedContent withAll(List<Environment> environments) {
+        if (environments == null || environments.isEmpty()) {
+            log.debug("No environments to check for promoted content.");
+            return this;
+        }
+
+        log.debug("Consumer has an environment; checking for promoted content.");
```
Ceiu, line 93: The log line here is out of context and probably unnecessary. Should probably move this back to the cert generation area, or remove it entirely.

src/main/java/org/candlepin/controller/util/EmptyContentPrefix.java
```diff
@@ -0,0 +1,28 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.controller.util;
+
+/**
+ * An empty prefix used for generation of Ueber certificates
+ */
+public class EmptyContentPrefix implements ContentPrefix {
```
Ceiu, line 21: If ContextPrefix is changed to be a functional interface, we can omit this entirely with a simple lambda, which is actually shorter than this class name: `envId -> ""`

Ceiu (changes requested):
The majority of this looks good, outside of some incredibly minor things like whitespace and a log line that doesn't really fit where it landed after the refactor.

However, something about the content prefix stuff irks me a bit. In general I'm not a huge fan of static factory methods and for this kind of thing heavily lean toward the fluent, builder style pattern. But I think there's a bit more to my hesitation than just my preference on the pattern.

I commented in the code that the EmptyContentPrefix could be omitted entirely if ContentPrefix becomes a functional interface, but I think there's a bit of functional overlap between the remaining two impl classes. Not sure if it's close enough to justify mucking with the hierarchy and inheritance, just something that stood out.

I also have some questions surrounding the output of the `.get` method in the case where the environment doesn't exist. A null environment is probably a necessary thing for the cases of generation when the user isn't part of *any* environment; but the environmentId != null but isn't found should probably be a failure state. At the very least, I would like to see the javadoc outline the output and expected behavior in all of the possible cases.

src/main/java/org/candlepin/controller/util/PromotedContent.java
Januson: Done.

src/main/java/org/candlepin/controller/util/EmptyContentPrefix.java
Januson: Done.

src/main/java/org/candlepin/util/X509V3ExtensionUtil.java
```diff
@@ -409,7 +406,7 @@ private Branding getBranding(Pool pool, String productId) {
             // Check if we should override the enabled flag due to setting on promoted content
             if (enableEnvironmentFiltering && consumer.getEnvironmentId() != null) {
```
nikosmoum, line 407: The `getEnvironmentId()` method is to be removed in the next task (SCA cert generation), so we should change this check to instead be `!consumer.getEnvironmentIds().isEmpty()` instead

src/main/java/org/candlepin/util/X509Util.java
```diff
@@ -88,14 +85,14 @@ public boolean apply(Product product) {
...
      */
     public Set<ProductContent> filterProductContent(Product prod, Consumer consumer,
-        Map<String, EnvironmentContent> promotedContent, boolean filterEnvironment,
+        PromotedContent promotedContent, boolean filterEnvironment,
         Set<String> entitledProductIds, boolean contentAccessMode) {
         Set<ProductContent> filtered = new HashSet<>();
 
         for (ProductContent pc : prod.getProductContent()) {
             // Filter any content not promoted to environment.
             if (filterEnvironment && consumer.getEnvironmentId() != null &&
```
nikosmoum, line 94: `!consumer.getEnvironmentIds().isEmpty()`

src/main/java/org/candlepin/util/X509ExtensionUtil.java
```diff
@@ -259,7 +256,7 @@ public X509ExtensionUtil(Configuration config) {
             // content:
             if (enableEnvironmentFiltering && consumer.getEnvironmentId() != null) {
```
nikosmoum, line 257: `!consumer.getEnvironmentIds().isEmpty()`

src/main/java/org/candlepin/controller/util/PromotedContent.java
```diff
@@ -0,0 +1,188 @@
...
+        return this.contents.get(pc.getContent().getId()).getEnabled();
+    }
+
+    /**
+     * Constructs a content path from environment specific prefix and given content.
+     *
+     * @param pc content for which to create a content path
+     * @return full content path
+     */
+    public String getPath(ProductContent pc) {
```
nikosmoum, line 151: I think one of the things @Ceiu was getting at, is that this method (in contrast to other public methods on this class), will still be used when a consumer is not in any environment, and it should still work, so we should update the javadoc of this method to explain that the path might be created with a prefix from either a) owner+environment, b) owner only or c) no prefix (just the content path)

nikosmoum: Also, we should unit test scenarios b) and c), since only scenario a) is unit tested in PromotedContentTest.java

nikosmoum (changes requested):

nikosmoum (approved):

Ceiu (approved):

PR #3145, opened 2021-10-29
[M] Addressed issues with Product.equals and immutable entity versioning (ENT-4464)

Ceiu:
- Fixed a bug in Product.equals and ProductContent.equals which
  would fail to properly compare products when either had children
  that did not yet have a UUID
- Updated how entity versioning is calculated and cached for both
  products and content: the entity version will be retained until
  any mutator which would affect a field that is used in the
  version is called
- Product.equals and Content.equals now attempt to use the UUID
  and entity versions to avoid checking individual fields wherever
  possible
- Product.equals and ProductContent.equals will now perform
  recursive calls to its children where necessary
- Product.setDerivedProduct, Product.addProvidedProduct, and
  Product.setProvidedProducts now attempt to check for a cycle
  in the product definitions and will throw an exception if
  one is detected
- Added an additional test to verify that the refresh operation
  is resulting in deduplicated products where possible

### DIFF
--- a/spec/refresh_pools_spec.rb
+++ b/spec/refresh_pools_spec.rb
@@ -1449,4 +1449,56 @@ describe 'Refresh Pools' do
     product_json = json_body['products'].find {|p| p['id'] == main_product['id']}
     product_json['content'].should == []
   end
+
+  it 'deduplicates products and content' do
+    # Create some orgs
+    owners = []
+    3.times do |i|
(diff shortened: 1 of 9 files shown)


Januson (approved):

Januson (changes requested):
Changes look good but it look like they caused some recursion issue in ProductImporterTest.

Januson (approved):

PR #3146, opened 2021-10-29
[4.0] Optimized refresh to avoid slow, repetitive operations (ENT-4489)

Ceiu:
- Substantially reduced the number of times flush is called
  during a refresh operation by removing the flush from the
  iterative sections to the end of each block where possible
- Changed how subscription pools are fetched during refresh
  to a single instance at the beginning of the operation,
  rather than once for every pool during refresh

### DIFF
--- a/server/spec/refresh_pools_spec.rb
+++ b/server/spec/refresh_pools_spec.rb
@@ -1385,6 +1385,193 @@ describe 'Refresh Pools' do
     product_json['content'].should == []
   end
 
+  it 'regenerates entitlements when pool start date changes' do
+    owner_key = random_string('test_owner')
+    owner = create_owner(owner_key)
+
+    product = create_upstream_product(random_string('test_prod'))
+    sub = create_upstream_subscription(random_string('test_sub'), owner_key, {
+      :product => product,
+      :start_date => Date.today - 20,
+      :end_date => Date.today + 20
+    })
+
+    @cp.refresh_pools(owner_key)
+    pools = @cp.list_pools({:owner => owner.id})
+    expect(pools.length).to eq(1)
+
+    user = user_client(owner, random_string('test_user'))
+    consumer = consumer_client(user, random_string('test_consumer'))
+
+    entitlements = consumer.consume_pool(pools.first['id'])
+    expect(entitlements.length).to eq(1)
+
+    consumer_entitlements = @cp.list_entitlements({:uuid => consumer.uuid})
+    expect(consumer_entitlements.length).to eq(1)
+    entitlement = consumer_entitlements.first
+
+    # Update subscription, then refresh. The entitlements should be regenerated
+    update_upstream_subscription(sub.id, { :start_date => Date.today - 10 })
+    @cp.refresh_pools(owner_key)
+
+    consumer_entitlements = @cp.list_entitlements({:uuid => consumer.uuid})
+    expect(consumer_entitlements.length).to eq(1)
+    new_entitlement = consumer_entitlements.first
+
+    expect(new_entitlement['pool']['id']).to eq(entitlement['pool']['id'])
+    expect(new_entitlement['certificates'].first['id']).to_not eq(entitlement['certificates'].first['id'])
+  end
+
+  it 'regenerates entitlements when pool end date changes' do
+    owner_key = random_string('test_owner')
+    owner = create_owner(owner_key)
+
+    product = create_upstream_product(random_string('test_prod'))
+    sub = create_upstream_subscription(random_string('test_sub'), owner_key, {
+      :product => product,
+      :start_date => Date.today - 20,
+      :end_date => Date.today + 20
+    })
+
+    @cp.refresh_pools(owner_key)
+    pools = @cp.list_pools({:owner => owner.id})
+    expect(pools.length).to eq(1)
+
+    user = user_client(owner, random_string('test_user'))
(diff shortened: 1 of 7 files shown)


Ceiu:
This PR needs a bit more testing to verify that certain scenarios are both covered by our test suite and comments surrounding some of the code removed by this optimization are no longer accurate.

Testing aside, the code is ready for review, and reviewers are encouraged to treat it as final for the purposes of a review.

server/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -745,7 +744,11 @@ void updateFloatingPools(List<Pool> floatingPools, boolean lazy, Map<String, Pro
 
         // Hand off to rules to determine which pools need updating
         List<PoolUpdate> updatedPools = poolRules.updatePools(floatingPools, changedProducts);
-        regenerateCertificatesByEntIds(processPoolUpdates(poolEvents, updatedPools), lazy);
+
+        Set<String> entIds = processPoolUpdates(poolEvents, updatedPools);
+        this.poolCurator.flush();
```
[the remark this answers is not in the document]
Ceiu: The other usage at the end of `updatePoolsForMasterPool` is the call being explicitly targeted by this change, so it's intended to not have a flush following it.

We should probably address *this* flush at some point as well, but is outside the immediate scope of this PR and requires investigation I simply don't have time to do at the moment.

Ceiu:
Note for reviewers: the current unit test failures in ProductImporterTest are addressed by PR #3150

Ceiu:
Update: I've added additional test surrounding the removed flush associated with BZ 1567922. From what I can tell via static analysis and the new tests, we're safe from the old locking issue mentioned in the comment.

server/spec/refresh_pools_spec.rb
```diff
@@ -214,6 +214,154 @@
...
+
+    consumer_entitlements = @cp.list_entitlements({:uuid => consumer1.uuid})
+    expect(consumer_entitlements.length).to eq(1)
+
+    consumer_entitlements = @cp.list_entitlements({:uuid => consumer2.uuid})
+    expect(consumer_entitlements.length).to eq(0)
+  end
+
+  it 'revokes entitlements when pool start date changes' do
+    owner_key = random_string('test_owner')
```
nikosmoum, line 261: These first 3 tests do not really test something significantly different from each other (quantity, end date, start date), so I'd say we only need 1 of them?

server/spec/refresh_pools_spec.rb
```diff
@@ -214,6 +214,154 @@
...
+    @cp.refresh_pools(owner_key)
+
+    consumer_entitlements = @cp.list_entitlements({:uuid => consumer.uuid})
+    new_entitlement = consumer_entitlements.first
+
+    expect(new_entitlement['pool']['id']).to eq(entitlement['pool']['id'])
+    expect(new_entitlement['certificates'].first['id']).to_not eq(entitlement['certificates'].first['id'])
+  end
+
+  it 'revokes entitlements when pool product changes' do
```
nikosmoum, line 333: The comment on the flush for 1567922 is not as detailed as commit summary of the fix. Specifically, it mentions a scenario where entitlement-derived pools being deleted (not just the entitlements being regenerated) after pool/product changes.

So to me it looks like the scenario that could potentially trigger a problem is: 
- We have a main pool with a finite quantity, and also a virt_limit attribute on it
- A host consumes entitlement(s) from it
- A derived pool is created as a result of that entitlement attachment
- The main pool is changed in one of two ways: 
    a) the quantity of it is reduced, resulting in the host's entitlement being revoked (because now it would be overconsuming), and that would mean we need to delete the derived pool that was created because of it, or
    b) virt_limit is removed from the main pool's product (meaning it no longer should provide derived pools),
which would mean we need to delete the derived pool
- That derived pool that was marked for deletion is attempted to be locked+updated later on, which results on some kind of error

nikosmoum (changes requested):
I might be a bit too cautious here, but let's make sure we're testing the right scenario

server/spec/refresh_pools_spec.rb
Ceiu: Ah ha. I'll add a test for this scenario so we don't have to worry about a regression here.

server/spec/refresh_pools_spec.rb
Ceiu: Given the lack of details in the BZ, I was making sure that every change that resulted in entitlement revocation was covered. Yes, there's overlap, but that's the nature of testing each path.

server/spec/refresh_pools_spec.rb
Ceiu: Done.

Ceiu:
A few more changes are necessary here for the new tests.

Ceiu:
I've removed one test which already had coverage, renamed the existing new tests and moved them to the end of the "regenerates entitlements when..." block.

nikosmoum:
@Ceiu I see the same 4 tests (no removed), and one of them `regenerates entitlements when pool start date changes` is at the start of the file, away from the others at the end?

Ceiu:
I removed the quantity test, since we already have one; but apparently copy/paste failed on moving all of the test. I've moved the last one down to the end with the others.

nikosmoum (approved):

PR #3147, opened 2021-11-01
[4.0] Added support for very large JSON files to the upstream sub injector

Ceiu:
- Changed how the upstream subscription injector loads subscriptions
  from JSON files; subscription data will be read individually as a
  pseudo-stream rather than attempting to load the entire JSON file
  at once. This allows for very large JSON files to be read without
  killing the Ruby process.


Ceiu:
This will need to get ported to master once this is accepted.

wolfdale (approved):

PR #3148, opened 2021-11-02
updating ca-certs so the ansible provisioner can run

Lorquas:
This allows the ansible provisioner to use things like urllib and requests so it can install things.  There are other options like a full yum update to the image or turning off SSL checking at this stage (possible more info here https://access.redhat.com/articles/2039753#red-hat-enterprise-linux-7-resolution-4 ), but this seems the fastest and cleanest.

To test this: vagrant destroy; vagrant up  should now work


barnabycourt:
LGTM, it got me past the certificates issue

PR #3151, opened 2021-11-02
[4.0] Update JUnit5 version to 5.7

Januson:
- Bump JUnit version to prevent JUnit vintage from failing to parse version of junit: 4.13.1


nikosmoum (approved):
Ack. Merging despite jenkins failure, since the PR succeded in making sure all unit tests run. The specific failures will be fixed in subsequent PR

PR #3152, opened 2021-11-02
2016418: Update method to mark entitlements dirty [M]

wottop:
Should solve the issue with performance and the concurrent
database errors as well

### DIFF
--- a/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -563,6 +563,53 @@ public class EntitlementCurator extends AbstractHibernateCurator<Entitlement> {
     }
 
     /**
+     * Fetches a the entitlement ids in the specified environment containing the
+     *  specified content ids.
+     *
+     * @param environmentId
+     *  The ID of the environment for which to fetch entitlements
+     *
+     *  @param contentIds
+     *  A collection of content Ids for which to fetch entitlements
+     *
+     * @return List of entitlement ids
+     */
+    public List<String> listEntitlementIdByEnvironmentAndContent(String environmentId,
+        Collection<String> contentIds) {
+        String sql =
+            "select a.id from cp_entitlement a " +
+            "    join cp_consumer b on a.consumer_id=b.id " +
+            "    join cp_environment c on b.environment_id=c.id " +
+            "    join cp_pool d on a.pool_id=d.id " +
+            "    join cp2_product_content f on d.product_uuid=f.product_uuid " +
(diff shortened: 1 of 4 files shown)


src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -562,6 +562,51 @@ private int countMatchesByFilters(AbstractHibernateObject<?> object, String obje
...
+            .createNativeQuery(sql)
+            .setParameter("environmentId", environmentId)
+            .setParameter("contentIds", contentIds)
+            .getResultList();
+
+        sql =
+            "select a.id from cp_entitlement a " +
+            "    join cp_consumer b on a.consumer_id=b.id " +
+            "    join cp_environment c on b.environment_id=c.id " +
+            "    join cp2_pool_provided_products f on a.pool_id=f.pool_id " +
```
nikosmoum, line 597: This table `cp2_pool_provided_products` is deprecated. The only reason it still exists is for backup, until Hosted migrates to 4.0 and we validate 4.0 works, at which point we will delete it. 

You have to use the new table `cp2_product_provided_products` (since now the provided products have marketing products be their parents, not pools). You likely need 1 more join here, but the way you get from Entitlement to Content now should be: cp_entitlement -> cp_pool -> cp2_product_provided_products -> cp2_product_content -> cp2_content

nikosmoum: Also, I was wondering why spec tests don't fail for this (specificaly `regenerates certificates after promoting/demoting content` in `environment_cert_v3_spec.rb`), and it turns out it's because it is currently assigning content directly to the marketing product (which means it's the first query in this method that will return the results), instead of creating an engineering product to add to the marketing product, and then adding the content to the engineering product (which is the realistic scenario); so let's fix that test too

nikosmoum (changes requested):
The relevant spec test needs fixing, along with the second query which needs to use the new 4.0+ table structures

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -562,6 +562,51 @@ private int countMatchesByFilters(AbstractHibernateObject<?> object, String obje
         return this.cpQueryFactory.buildQuery(this.currentSession(), criteria);
     }
 
+    /**
+     * Fetches a the entitlements used by consumers in the specified environment.
```
nikosmoum, line 566: Also, let's fix this comment to read better, and also mention that it's fetching based on content ids (in addition to environment)

nikosmoum (approved):
the changes look good, but the spec test is failing

PR #3153, opened 2021-11-03
[M] Include contentAccessMode in nested Owner [ENT-4349]

jirihnidek:
* ENT-4349
* It is usefull to get information about content access mode,
  when system is registered and JSON document with consumer
  is returned
* TODO:
  - Spec tests
  - Unit tests
  - Anything else?

### DIFF
--- a/src/main/java/org/candlepin/dto/api/v1/NestedOwnerDTOTranslator.java
+++ b/src/main/java/org/candlepin/dto/api/v1/NestedOwnerDTOTranslator.java
@@ -53,6 +53,7 @@ public class NestedOwnerDTOTranslator implements ObjectTranslator<NestedOwnerDTO
         dest.setKey(source.getKey());
         dest.setDisplayName(source.getDisplayName());
         dest.setHref(source.getHref());
+        dest.setContentAccessMode(source.getContentAccessMode());
 
         return dest;
     }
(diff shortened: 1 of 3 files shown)


jirihnidek:
This is WIP PR to be able to test this PR properly: https://github.com/candlepin/subscription-manager/pull/2860

wolfdale:
@jirihnidek Some of the spec test are failing. Also it would be nice to add unit & spec test for this change.

src/main/java/org/candlepin/dto/api/v1/NestedOwnerDTOTranslator.java
```diff
@@ -53,6 +53,7 @@ public Owner populate(ModelTranslator translator, NestedOwnerDTO source, Owner d
         dest.setKey(source.getKey());
         dest.setDisplayName(source.getDisplayName());
         dest.setHref(source.getHref());
+        dest.setContentAccessMode(source.getContentAccessMode());
```
nikosmoum, line 56: Minor note: you don't need to add this here. This is a 'reverse' translator that translates NestedOwnerDTO coming from the API into Owner. Since your change is only about adding this field to the output (and not accepting it in the input), then let's skip this change

nikosmoum:
This PR has not been progressing since opened, so I'm closing for now. The branch (jhnidek/ENT-4349) will still be able to be used for testing against the client, and whenever this needs to be worked again this PR can be reopened

PR #3155, opened 2021-11-04
[M] Upload code coverage report with sonarqube

nikosmoum:
- Sonarqube gradle task now depends on the test task.
- When sonarqube is run from within our test container,
  fail-fast the unit tests to save cycles on Jenkins nodes.
- Run sonarqube stage in parallel with the rest of the
  stages in Jenkins to avoid extending the overall
  duration of the pipeline.
- Print location of JaCoCo code coverage report.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,7 +96,9 @@ between 1.0 and 10.0.  Any CVEs above the maximum allowed CVSS score will cause 
 The reports will be generated automatically under build/reports folder.
 
 ### Checkstyle
-* `./gradlew checkstyleMain`
+* `./gradlew checkstyle` Runs checkstyle for both production and test code.
+* `./gradlew checkstyleMain` Runs checkstyle only for production code.
+* `./gradlew checkstyleTest` Runs checkstyle only for test code.
 
 Buildr provides a Checkstyle task, but we have our own that reads from the  
 Eclipse Checkstyle Plugin configuration.  The Eclipse configuration defines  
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -375,8 +375,13 @@ jacocoTestReport {
...
         xml.enabled true
     }
+
+    doLast {
+        println "Generated JaCoCo test coverage report at $buildDir/reports/jacoco/test/html/index.html"
+    }
 }
 
+project.tasks["sonarqube"].dependsOn "test"
 project.tasks["sonarqube"].dependsOn "jacocoTestReport"
```
Ceiu, line 385: Could we rename this task to something less specific to the backend, such as "coverage"? Also, if it's an independent task, does it work with rspec as well? If not, and if it only works with unit tests, would it make more sense to be an option for unit testing ala `gradlew test --coverage`?

Ceiu (approved):
The core changes of this PR are fine, though I've a question on the name of the underlying task(s) here.

CONTRIBUTING.md
```diff
@@ -118,7 +120,15 @@ path to `checks.xml` and drop the result into `.checkstyle` in your Eclipse
 project directory.
 
 ### Unit Tests
-* `./gradlew tasks` runs Unit tests
+* `./gradlew tests` runs all of the unit tests.
+* `./gradlew tests --tests org.candlepin.controller.Cdn*` runs only the unit
```
Januson, line 124: Typo. Task name should be test. It's also in coverage section.

build.gradle
Januson: This task comes from sonarqube plugin so we cannot rename it, but we could create a coverage task to alias it.

nikosmoum: This plugin only works with java, so it can't work with rspec tests (but it might work once we migrate to java integration tests in the future). As @Januson said, I've created a `coverage` alias for it. I couldn't find a way to run it using a flag on the test task, so you'll have to run `gradlew test coverage`

Ceiu: Sounds good

PR #3159, opened 2021-11-10
[4.0] Duplicate entity versions are now disallowed per-ID (ENT-4509)

Ceiu:
- Versioned entities (products and content) must now be unique per
  (id, version) tuple
- Removed some obsoleted version lookup code
- RefreshPoolsJob now defaults to 3 retries, as the new unique
  restriction could cause parallel refreshes containing the same
  net-new product or content to fail on the first attempt

### DIFF
--- a/server/src/main/java/org/candlepin/controller/ContentAccessManager.java
+++ b/server/src/main/java/org/candlepin/controller/ContentAccessManager.java
@@ -243,48 +243,37 @@ public class ContentAccessManager {
      * some content. The certificate will be regenerated if it is missing, or new content has been made
      * available. Otherwise, it will be simply fetched.
      *
-     * @return Client entitlement certificates.
-     * @throws IOException thrown if there's a problem reading the cert.
-     * @throws GeneralSecurityException thrown security problem
+     * @return Client entitlement certificates
      */
-    public ContentAccessCertificate getCertificate(Consumer consumer)
-        throws GeneralSecurityException, IOException {
-
+    public ContentAccessCertificate getCertificate(Consumer consumer) {
         // TODO: FIXME: Redesign all of this.
-
-        Owner owner = consumer.getOwner();
 
         // Ensure the org is in SCA mode and the consumer is able to process the cert we'll be
         // generating for them.
+        Owner owner = consumer.getOwner();
         if (owner == null || !owner.isUsingSimpleContentAccess() || !this.consumerIsCertV3Capable(consumer)) {
             return null;
         }
 
-        org.candlepin.util.Transactional<ContentAccessCertificate> transaction =
-            this.consumerCurator.transactional(args -> {
-                ContentAccessCertificate existing = consumer.getContentAccessCert();
-                return existing == null ?
-                    createNewScaCertificate(consumer, owner) :
-                    updateScaCertificate(consumer, owner, existing);
-            });
-
-        ContentAccessCertificate result;
         try {
-            result = transaction.allowExistingTransactions()
+            ContentAccessCertificate result = this.consumerCurator.<ContentAccessCertificate>transactional()
+                .allowExistingTransactions()
                 .onRollback(status -> log.error("Rolling back SCA cert (re)generation transaction"))
-                .execute();
-        }
-        catch (IOException | GeneralSecurityException e) {
-            throw e;
+                .execute(args -> {
+                    ContentAccessCertificate existing = consumer.getContentAccessCert();
+                    return existing == null ?
+                        createNewScaCertificate(consumer, owner) :
+                        updateScaCertificate(consumer, owner, existing);
+                });
+
+            return this.wrap(result);
         }
         catch (Exception e) {
-            // Something went horribly wrong...
-            log.error("Could not fetch or (re)generate SCA certificate for consumer {}",
-                consumer.getUuid(), e);
-            return null;
+            log.error("Unexpected exception occurred while fetching SCA certificate for consumer: {}",
+                consumer, e);
         }
 
-        return wrap(result);
+        return null;
     }
 
     private ContentAccessCertificate createNewScaCertificate(Consumer consumer, Owner owner)
--- a/server/src/main/java/org/candlepin/model/OwnerContentCurator.java
+++ b/server/src/main/java/org/candlepin/model/OwnerContentCurator.java
@@ -334,67 +334,7 @@ public class OwnerContentCurator extends AbstractHibernateCurator<OwnerContent> 
     }
 
     /**
-     * Retrieves a map containing all known versions of the content specified by IDs, for all orgs
-     * <em>except</em> the org specified. If no content is found for the specified IDs in other
-     * orgs, this method returns an empty map.
-     *
-     * @param owner
-     *  The owner to exclude from the content lookup. If this value is null, no owner-filtering
-     *  will be performed.
-     *
-     * @param contentIds
-     *  A collection of content IDs for which to fetch all known versions
-     *
-     * @return
-     *  A map containing all known versions of the given content, mapped by Red Hat ID
-     */
-    public Map<String, Set<Content>> getVersionedContentById(Owner owner, Collection<String> contentIds) {
-        Map<String, Set<Content>> result = new HashMap<>();
-
-        if (contentIds != null && !contentIds.isEmpty()) {
-            String jpql;
-
-            if (owner != null) {
-                jpql = "SELECT c FROM OwnerContent oc JOIN oc.content c " +
-                    "WHERE oc.owner.id != :owner_id AND c.id IN (:cids)";
-            }
-            else {
-                jpql = "SELECT c FROM Content c WHERE c.id IN (:cids)";
-            }
-
-            TypedQuery<Content> query = this.getEntityManager().createQuery(jpql, Content.class);
-
-            if (owner != null) {
-                query.setParameter("owner_id", owner.getId());
-            }
-
-            for (Collection<String> block : this.partition(contentIds)) {
-                List<Content> fetched = query.setParameter("cids", block)
-                    .getResultList();
-
-                for (Content entity : fetched) {
-                    Set<Content> idSet = result.get(entity.getId());
-
-                    if (idSet == null) {
-                        idSet = new HashSet<>();
-                        result.put(entity.getId(), idSet);
-                    }
-
-                    idSet.add(entity);
-                }
-            }
-        }
-
-        return result;
-    }
-
-    /**
-     * Fetches all content belonging to organizations other than the one provided having an entity
-     * version equal to one of the versions provided.
-     *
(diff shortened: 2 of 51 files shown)


Ceiu:
Not sure if I'm entirely happy with this solution. It definitely cleans up some things, but basically breaks any parallel operation which creates versioned entities. Debating which set of problems is going to be worse to deal with.

Ceiu:
After *lots* of manual testing, this logic actually seems to be somewhat promising. The code here still needs a lot of refactoring and proper design to not be cringe-worthy, but it seems to perform exceptionally well even in parallel with retries.

Some core changes that will come out of this prototype:
- Split refresh into two transactions: entity mapping and persistence (RefreshWorker)
- Add retry logic to the entity mapper
- Updates to the Transactional object, and shifting usage of the \@Transactional annotation to transactional blocks where appropriate
- Some kind of decorator or extension to the two versioned entity managers to provide a single-time retry if we hit the version constraint violation on entity persistence
- Updating the RefreshPoolsJob's config to have a default retry limit of 3 rather than 1

The implementation of some of the code here is terrible, but reviewers should feel free to call out what they like and don't like anyway.

server/src/main/java/org/candlepin/util/EntityVersioningRetryWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.util;
```
nikosmoum, line 15: Suggestion: this class could go in `org.candlepin.controller.util`, since it's not general utility, but relevant to the classes in the controller package (Product/ContentManagers, RefreshWorker). (Also, I know it's also being used by the *Resource classes, but I feel like it shouldn't; see my other comment on the resource class)

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
```diff
@@ -276,10 +289,13 @@ public ContentDTO updateContent(@PathParam("owner_key") String ownerKey,
             throw new ForbiddenException(i18n.tr("content \"{0}\" is locked", existing.getId()));
         }
 
-        existing = this.contentManager.updateContent(owner, content, true);
+        Content updated = new EntityVersioningRetryWrapper()
+            .attempts(VERSIONING_CONSTRAINT_VIOLATION_ATTEMPTS)
+            .execute(() -> this.contentManager.updateContent(owner, content, true));
```
nikosmoum, line 294: If the retry logic is going to need to be wrapped around all of our Product/Content CRUD operations (or at least most of them), then I wouldn't let that be the burden of the caller. There must be a way to enforce this as the default behavior of the ContentManager/ProductManager methods, and then give the caller a way to opt-out of it for the few times we don't need to retry?

nikosmoum (changes requested):
A couple of suggestions, but looks good

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
Ceiu: Yeah... this is a tough one. I generally agree that this kind of thing should be completely transparent to the callers, but where things started going off the rails here is in that the ContentManager calls into the ProductManager to update affected products when content changes. We wouldn't want the retry to be multi-layered, and I wasn't sure if I wanted to pollute the namespace with a bunch of methods that would only be called by the ContentManager.

The original prototype had this concept that extended the two managers and had methods that did nothing but wrap the call to the superclass's version in the retry logic. It worked, but it was a bit jank looking and feeling, and hit the issue of nesting those transactional objects during content updates, so I dropped it and went with what we have here. The managers being singletons might get in the way of this, but maybe something along the lines of having the standard binding to the retry-wrapped subclass, and the ContentManager uses a binding to get the vanilla ProductManager without retry logic?

Definitely open to suggestions and thoughts here.

server/src/main/java/org/candlepin/util/EntityVersioningRetryWrapper.java
Ceiu: Sure. Once we decide on how to address the other issue, we can figure out where this should go.

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
nikosmoum: I thought Guice by default uses new instances if you don't specify scope, and I don't see a Singleton annotation or binding call in CandlepinModule.java for the manager classes, so I don't think they're singletons. Even if they were, Guice singletons aren't true singletons, but more like limited-instantiation objects, so you could just have 2 instances only of them if you use the Named annotation to differentiate between them.

Also, you could entirely avoid having 2 different manager classes for retry/non-retry, by embbedding the retry logic in the single manager class, and have it be configurable (so that you can set the number of retries to 0), and then defining 2 guice provider methods with Provides and Named annotations, where one returns the manager class configured with retries, and the other without retries (we already do this kind of thing for ObjectMapper classes if you check CandlepinModule.java)

server/src/main/java/org/candlepin/model/OwnerContentCurator.java
```diff
@@ -334,96 +334,22 @@ public int clearContentForOwner(Owner owner) {
...
      * @return
      *  a map containing the contents found, keyed by content ID
      */
-    public Map<String, List<Content>> getContentByVersions(Owner exclude, Collection<Integer> versions) {
+    public Map<String, List<Content>> getContentByVersions(Collection<Integer> versions) {
         Map<String, List<Content>> result = new HashMap<>();
 
         if (versions != null && !versions.isEmpty()) {
-            TypedQuery<Content> query;
+            String jpql = "SELECT c FROM Content c WHERE c.entityVersion IN (:vblock)";
```
nikosmoum, line 349: One more thing just came to mind: Now that we can reuse orphaned entities, what happens if the OrphanCleanupJob runs at the same time? I see the job is using pessimistic write lock, so if it manages to lock the entity first this side might fail to read with a locktimeoutexception, but what if the job acquires the lock after this query is done, but right before we create the new owner-content mapping? Should we be locking entities in this side too?

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
Ceiu: Did we remove the `@Singleton` annotation from the managers? I could have sworn they were flagged as such at one point (and, honestly, they probably should be).

I'll poke around a bit with some implementations and post a gist or two or a commit here and we'll see how they feel.

server/src/main/java/org/candlepin/model/OwnerContentCurator.java
Ceiu: Transaction isolation and referential integrity will save us here, so long as we continue to only create owner-entity mappings in transactions and don't lift that foreign key ref. 

In the case of parallel ops here, we have the following cases:

1. Orphan cleanup gets the lock first. When this happens, the product mapping portion of refresh has either already passed, or is waiting to run. A potential lockwait failure can occur, but by the time the orphan cleanup job is complete and the lock is released, the product will no longer exist and refresh will not see it.
2. Refresh gets the lock first. When this occurs, refresh is either creating a new reference to the product, or removing it from the owner. In the former case, once the lock is acquired in orphan cleanup, the query shouldn't return it. For the later case, it won't matter.

This is largely off the top of my head, so some particulars of how locks propagate with foreign keys, and whether or not the lock acquired by the orphan cleanup job query will acquire it before or after determining the results of the query are a bit fuzzy. Even in the worst-case, I expect that a foreign key violation will save us from doing anything bad.

server/src/main/java/org/candlepin/util/EntityVersioningRetryWrapper.java
Ceiu: This has been moved to the suggested package for now

server/src/main/java/org/candlepin/resource/OwnerContentResource.java
Ceiu: These have all been dropped for the moment, but I'm still conflicted. Given some of the other transaction problems, this might end up getting resolved once I decide on what to do there.

Ceiu: Side note on the singleton thing: there are some context issues that arise when flagging either manager as a singleton. Probably worth investigating eventually, but not part of this task.

server/src/main/java/org/candlepin/util/EntityVersioningRetryWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+     * If the action fails with a versioning-related constraint violation and is at the retry limit,
+     * this method throws a RuntimeException.
+     *
+     * @param action
+     *  the action to retry until successful completion
+     *
+     * @return
+     *  the result of the provided action
+     */
+    public <O> O execute(Supplier<O> action) {
```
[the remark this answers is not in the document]
Ceiu: Done

Ceiu:
This PR has been updated with a couple new commits that add missing tests, make corrections to how retry versioning logic is done (largely dropping it in most places), and cleans up some other implementation details.

At this point it's basically ready for review, but it still needs a rebase/squash before it will be flipped for merge. This will be done after the new code is approved.

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -247,44 +247,35 @@ public ContentAccessManager(
      * @throws IOException thrown if there's a problem reading the cert.
      * @throws GeneralSecurityException thrown security problem
      */
-    public ContentAccessCertificate getCertificate(Consumer consumer)
-        throws GeneralSecurityException, IOException {
```
nikosmoum, line 251: If these exceptions are not on the method contract let's also remove them from the javadoc above

nikosmoum (changes requested):
Only 1 minor issue and then good to merge!

server/src/main/java/org/candlepin/controller/ContentAccessManager.java
Ceiu: done

nikosmoum (approved):

Ceiu:
This PR is blocked, waiting on changes from #3188 to be merged before this can be rebased and merged itself.

Ceiu:
Waiting on backport of #3200 to address the locking issue once and for all.

PR #3162, opened 2021-11-16
ENT-4180: Update the ConsumerDTO for backwards compatibility

wolfdale:
Updated the ConsumerDTO’s ‘environment’ field to not populate the whole EnvironmentDTO, but only EnvironmentDTO’s ‘name’ field with the comma separated list of names of all the consumer’s environments (in priority order).


api/candlepin-api-spec.yaml
```diff
@@ -8435,7 +8435,7 @@ components:
               owner:
                 $ref: '#/components/schemas/NestedOwnerDTO'
               environment:
-                $ref: '#/components/schemas/EnvironmentDTO'
+                type: string
```
nikosmoum, line 8438: This is not supposed to be a String, but an EnvironmentDTO, whose only field being populated will be the 'name' field, which will be a comma-separated list of all the environment names

wolfdale: ah I thought It's only EnvironmentDTO’s ‘name’ field.
Fixed it.

src/test/java/org/candlepin/resource/EnvironmentResourceTest.java
```diff
@@ -241,10 +240,6 @@ void shouldCreateConsumerInMultipleEnvironment() {
         assertEquals(dto.getEnvironments().get(0).getId(), "env1");
         assertEquals(dto.getEnvironments().get(1).getId(), "env3");
         assertEquals(dto.getEnvironments().get(2).getId(), "env2");
-
-        // TODO this need to be changes whenever we make changes to
-        // consumerDTO to handle 'environment'
-        assertNull(dto.getEnvironment());
```
nikosmoum, line 247: I think instead of deleting this we could simply verify that the `dto.getEnvironment().getName()` is populated as expected?

wolfdale: Since this `environment` gets populated during DTO translation, this might not be right place to test this. I have moved this testing bit in [ConsumerTranslatorTest.java](https://github.com/candlepin/candlepin/pull/3162/files#diff-bea22202f35772a874ecbaa6bc0ecb1d84e9401e2e267eff80ead6c32cb5f522R258)

PR #3164, opened 2021-11-16
[M] 2010251: SCA 'lastUpdated' & If-Modified-Since date fixes [ENT-4428]

nikosmoum:
- Now we only wrap the SCA cert into a fresh ContentAccessCertificate
  copy after the db changes are flushed (to make sure the 'updated'
  date that is marked with @PreUpdate annotation has been changed) to
  avoid sending a stale 'lastUpdated' date to the client.
- The date sent in the If-Modified-Since header does not have
  millisecond accuracy, so make sure to round down the server-side
  certificate dates we compare it with to the second.

### DIFF
--- a/spec/content_access_spec.rb
+++ b/spec/content_access_spec.rb
@@ -465,16 +465,22 @@ describe 'Content Access' do
     old_cert = content_body.contentListing.values[0]
     old_x509 = old_cert[0]
     old_content = old_cert[1]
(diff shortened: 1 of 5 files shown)


Ceiu (changes requested):
Changes requested on the 4.0 branch of this PR; this review is being posted purely for bookkeeping reasons.

Ceiu (approved):

PR #3170, opened 2021-11-24
[4.0] Enforce locale when parsing Date from String [ENT-4536]

nikosmoum:
- Starting with Java 9, the CLDR is used as the default locale
provider, which has an effect on how java.text and java.time utils
format/parse certain date formats. Specifically, the format used
by the If-Modified-Since header value is affected when the server
is run with system locale set to some locales such as en_AU or
en_CA, which results in a parsing error to be thrown. Enforcing
the locale used when parsing dates to en_US solves this.

### DIFF
--- a/server/src/test/java/org/candlepin/resteasy/DateFormatterTest.java
+++ b/server/src/test/java/org/candlepin/resteasy/DateFormatterTest.java
@@ -1,0 +1,130 @@
+/**
+ * Copyright (c) 2009 - 2021 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.resteasy;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+
+import java.lang.annotation.Annotation;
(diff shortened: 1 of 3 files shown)


nikosmoum:
Note: this is ready to be reviewed, but is a draft because I am waiting for a proper BZ to be raised, so that I can add the BZ ID on the commit.

server/src/test/java/org/candlepin/resteasy/DateFormatterTest.java
```diff
@@ -0,0 +1,52 @@
...
+        Locale.setDefault(Locale.US);
+    }
+
+    @Test
+    public void testFromStringParseWithAustralianEnglishAsSystemLocale() {
+        Locale.setDefault(new Locale("en", "AU"));
+        DateFormatter formatter = new DateFormatter();
+        Annotation[] annotations = new Annotation[1];
+        DateFormat dateFormatAnnotation = mock(DateFormat.class);
+        try (MockedStatic<FindAnnotation> findAnnotationMock = mockStatic(FindAnnotation.class)) {
```
Januson, line 43: We could use a test class here. That would allow us to drop the mocking and just collect annotation from the class. I think it would help readability and it would be easier to add more tests.

Something like:
```java
    @Test
    public void testFromStringParseWithAustralianEnglishAsSystemLocale() {
        Locale.setDefault(new Locale("en", "AU"));
        DateFormatter formatter = new DateFormatter();
        formatter.setAnnotations(Annotated.class.getAnnotations());

        assertDoesNotThrow(() -> formatter.fromString("Thu, 01 Jan 1970 00:00:00 GMT"));
    }

    @DateFormat("EEE, dd MMM yyyy HH:mm:ss z")
    private static class Annotated {}
```

server/src/test/java/org/candlepin/resteasy/DateFormatterTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class DateFormatterTest {
+
+    @AfterEach
+    public void cleanup() {
+        Locale.setDefault(Locale.US);
+    }
+
+    @Test
+    public void testFromStringParseWithAustralianEnglishAsSystemLocale() {
```
Januson, line 38: Could you add a few more tests? There's a few uncovered cases and their tests should be pretty simple. Like the case with no formatters available.

Januson (changes requested):
Changes look good. I have just a few comments regarding tests.

server/src/test/java/org/candlepin/resteasy/DateFormatterTest.java
nikosmoum: done

server/src/test/java/org/candlepin/resteasy/DateFormatterTest.java
nikosmoum: done

Januson (approved):

PR #3171, opened 2021-11-25
ENT-4183: DB Upgrade for environment data

Januson:
- Migrate cp_consumer.environment_id to cp_consumer_environment table
- Delete cp_conusmer.environment_id column
- Remove redundant code from Consumer.setEnvironmentIds

### DIFF
--- a/src/main/java/org/candlepin/model/Consumer.java
+++ b/src/main/java/org/candlepin/model/Consumer.java
@@ -1010,26 +1010,26 @@ public class Consumer extends AbstractHibernateObject implements Linkable, Owned
             this.environmentIds.clear();
         }
 
-        // Creating a new map instance and assigning it to the environmentIds.
-        // This ensures that, we are deleting the existing entries and inserting the new entries.
-        // This is done to avoid constraint (cp_consumer_environments_pkey) violation exception.
-        // Whenever we try to update the existing data (changing priority), it actually updates
-        // the environment Id and not the priority.
-        LinkedHashMap<String, String> envMap = new LinkedHashMap<>();
         if (environmentIds != null) {
-
-            for (String environmentId : environmentIds) {
-                if (envMap.containsValue(environmentId)) {
-                    throw new DuplicateEntryException("Environment " + environmentId +
-                        " specified more than once.");
-                }
+            HashSet<String> deDuplicatedIds = new HashSet<>(environmentIds);
+            if (environmentIds.size() != deDuplicatedIds.size()) {
+                throw new DuplicateEntryException("One or more environments were specified more than once.");
             }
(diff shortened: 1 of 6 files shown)


src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -1010,26 +1010,21 @@ public Consumer setEnvironmentIds(List<String> environmentIds) {
...
-        // the environment Id and not the priority.
-        LinkedHashMap<String, String> envMap = new LinkedHashMap<>();
         if (environmentIds != null) {
-
-            for (String environmentId : environmentIds) {
-                if (envMap.containsValue(environmentId)) {
-                    throw new DuplicateEntryException("Environment " + environmentId +
-                        " specified more than once.");
-                }
-            }
```
nikosmoum, line 1026: We need to iterate twice here, so I don't think we should be removing this. See here for explanation: https://github.com/candlepin/candlepin/pull/3129#discussion_r722542881 If anything, we might need to add an extra comment here to explain why 2 loops are needed

nikosmoum: There was also a related discussion about adding a composite unique key on the consumerId+environmentId columns, so that we don't have to resort to manually checking for duplicates, but that didn't work out because hibernate likes to add new environmentIds as duplicates before removing the old ones during the re-prioritization use case.

src/main/resources/db/changelog/20211123093750-migrate-environment-column.xml
```diff
@@ -0,0 +1,20 @@
...
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20211123093750-1" author="ojanus">
+        <sql>
+            INSERT INTO cp_consumer_environments(cp_consumer_id, environment_id, priority)
+            SELECT id, environment_id, 0 FROM cp_consumer
```
nikosmoum, line 12: Running this in Hosted (or any setup that does not enforce the use of environments), would fail with non-null constraint violation because it would try to insert NULL values in the environment_id column (because the original cp_consumer.environment_id values are null for consumers). To make sure this doesn't happen, let's add a `WHERE environment_id IS NOT NULL` at the end of that `SELECT`

nikosmoum (changes requested):

nikosmoum (approved):

PR #3172, opened 2021-11-25
[M] Enforce locale when parsing Date from String [ENT-4536]

nikosmoum:
- Starting with Java 9, the CLDR is used as the default locale
  provider, which has an effect on how java.text and java.time utils
  format/parse certain date formats. Specifically, the format used
  by the If-Modified-Since header value is affected when the server
  is run with system locale set to some locales such as en_AU or
  en_CA, which results in a parsing error to be thrown. Enforcing
  the locale used when parsing dates to en_US solves this.

### DIFF
--- a/client/ruby/candlepin_api.rb
+++ b/client/ruby/candlepin_api.rb
@@ -1319,10 +1319,10 @@ class Candlepin
       else
         uuid = @uuid
       end
-      since = params[:since] if params[:since]
+      headers = {:accept => :json}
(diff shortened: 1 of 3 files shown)


nikosmoum:
Note: this is ready to be reviewed, but is a draft because I am waiting for a proper BZ to be raised, so that I can add the BZ ID on the commit.

Januson (approved):

PR #3173, opened 2021-11-30
ENT-4514 : Regen entitlements certs when environment list is modified

wolfdale:
Regenerate entitlements certs when environment list is modified aka environments being added/removed/re-prioritized.

### DIFF
--- a/spec/environment_spec.rb
+++ b/spec/environment_spec.rb
@@ -551,4 +551,504 @@ describe 'Environments' do
       {:content_url=> "/this/is/the/path/#{id}", :arches => "x86_64"}, true)
   end
 
+  it 'should re-gen consumer entitlement cert when higher priority environment is added where provided
+    content is NOT same as higher priority environment' do
+    consumer = @org_admin.register(random_string('testsystem'), :system, nil, {}, nil, nil, [], [],
+      [{ 'id' => @env['id']}])
+
+    expect(consumer['environments']).not_to be_nil
+
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+    product = create_product
+    contentA = create_content
+    contentB = create_content
+    contentC = create_content
+    @cp.add_content_to_product(@owner['key'], product['id'], contentA['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentB['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentC['id'])
+    job = @org_admin.promote_content(@env['id'], [{:contentId => contentB['id']}])
+    wait_for_job(job['id'], 15)
+    pool = @cp.create_pool(@owner['key'], product['id'])
+    ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    serial = ent['certificates'][0]['serial']['serial']
+
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be false
+
+    envB = @org_admin.create_environment(@owner['key'], 'envB', "My Test Env 2", "For test systems only.")
+    job = @org_admin.promote_content(envB['id'],[{:contentId => contentC['id'],}])
+    wait_for_job(job['id'], 15)
+    consumer_cp.update_consumer({:environments => [{:name => envB.name}, {:name => @env.name}]})
+    ent = consumer_cp.list_entitlements()[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    new_serial = ent['certificates'][0]['serial']['serial']
+
+    expect(new_serial).not_to eq(serial)
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be true
+  end
+
+  it 'should re-gen consumer entitlement cert when higher priority environment is added where provided
+    content is NOT same as higher priority environment (production)' do
+    # This spec test is same as above only difference is that,
+    # the content is present on provided products instead of on main product
+    # just like we have on production environment
+    consumer = @org_admin.register(random_string('testsystem'), :system, nil, {}, nil, nil, [], [],
+      [{ 'id' => @env['id']}])
+
+    expect(consumer['environments']).not_to be_nil
+
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+    provided_prod = create_product("123", "123", { :owner => @owner['key']})
+    contentA = create_content
+    contentB = create_content
+    contentC = create_content
+    @cp.add_content_to_product(@owner['key'], provided_prod['id'], contentA['id'])
+    @cp.add_content_to_product(@owner['key'], provided_prod['id'], contentB['id'])
+    @cp.add_content_to_product(@owner['key'], provided_prod['id'], contentC['id'])
+    product = create_product("12345", "12345", {:owner => @owner['key'], :providedProducts => [provided_prod.id]})
+
+    job = @org_admin.promote_content(@env['id'], [{:contentId => contentB['id']}])
+    wait_for_job(job['id'], 15)
+    pool = @cp.create_pool(@owner['key'], product['id'])
+    ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    serial = ent['certificates'][0]['serial']['serial']
+
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be false
+
+    envB = @org_admin.create_environment(@owner['key'], 'envB', "My Test Env 2", "For test systems only.")
+    job = @org_admin.promote_content(envB['id'],[{:contentId => contentC['id'],}])
+    wait_for_job(job['id'], 15)
+    consumer_cp.update_consumer({:environments => [{:name => envB.name}, {:name => @env.name}]})
+    ent = consumer_cp.list_entitlements()[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    new_serial = ent['certificates'][0]['serial']['serial']
+
+    expect(new_serial).not_to eq(serial)
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be true
+  end
+
+  it 'should NOT re-gen consumer entitlement cert when lower priority environment is added where provided
+    content is same as higher priority environment' do
+    consumer = @org_admin.register(random_string('testsystem'), :system, nil, {}, nil, nil, [], [],
+      [{ 'id' => @env['id']}])
+
+    expect(consumer['environments']).not_to be_nil
+
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+    product = create_product
+    contentA = create_content
+    contentB = create_content
+    contentC = create_content
+    @cp.add_content_to_product(@owner['key'], product['id'], contentA['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentB['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentC['id'])
+
+    job = @org_admin.promote_content(@env['id'],[{:contentId => contentB['id']}])
+    wait_for_job(job['id'], 15)
+    pool = @cp.create_pool(@owner['key'], product['id'])
+    ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    serial = ent['certificates'][0]['serial']['serial']
+
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be false
+
+    envB = @org_admin.create_environment(@owner['key'], 'envB', "My Test Env 2", "For test systems only.")
+    job = @org_admin.promote_content(envB['id'],[{:contentId => contentB['id']}])
+    wait_for_job(job['id'], 15)
+    consumer_cp.update_consumer({:environments => [{:name => @env.name}, {:name => envB.name}]})
+    ent = consumer_cp.list_entitlements()[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    new_serial = ent['certificates'][0]['serial']['serial']
+
+    expect(new_serial).to eq(serial)
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be false
+  end
+
+  it 'should re-gen consumer entitlement cert when environment is added having unique content' do
+    consumer = @org_admin.register(random_string('testsystem'), :system, nil, {}, nil, nil, [], [],
+      [{ 'id' => @env['id']}])
+
+    expect(consumer['environments']).not_to be_nil
+
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+    product = create_product
+    contentA = create_content
+    contentB = create_content
+    contentC = create_content
+    @cp.add_content_to_product(@owner['key'], product['id'], contentA['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentB['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentC['id'])
+    job = @org_admin.promote_content(@env['id'],[{:contentId => contentB['id']}])
+    wait_for_job(job['id'], 15)
+    pool = @cp.create_pool(@owner['key'], product['id'])
+    ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    serial = ent['certificates'][0]['serial']['serial']
+
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be false
+
+    envB = @org_admin.create_environment(@owner['key'], 'envB', "envB", "For test systems only.")
+    job = @org_admin.promote_content(envB['id'],[{:contentId => contentC['id']}])
+    wait_for_job(job['id'], 15)
+    consumer_cp.update_consumer({:environments => [{:name => @env.name}, {:name => envB.name}]})
+    ent = consumer_cp.list_entitlements()[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    new_serial = ent['certificates'][0]['serial']['serial']
+
+    expect(new_serial).not_to eq(serial)
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be true
+  end
+
+  it 'should NOT re-gen consumer entitlement cert when environment is added where entitlement does
+    not provide content' do
+    consumer = @org_admin.register(random_string('testsystem'), :system, nil, {}, nil, nil, [], [],
+      [{ 'id' => @env['id']}])
+
+    expect(consumer['environments']).not_to be_nil
+
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+    product = create_product
+    contentA = create_content
+    contentB = create_content
+    contentC = create_content
+    contentD = create_content
+    @cp.add_content_to_product(@owner['key'], product['id'], contentA['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentB['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentC['id'])
+    job = @org_admin.promote_content(@env['id'],[{:contentId => contentB['id'],}])
+    wait_for_job(job['id'], 15)
+    pool = @cp.create_pool(@owner['key'], product['id'])
+    ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    serial = ent['certificates'][0]['serial']['serial']
+
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be false
+
+    envB = @org_admin.create_environment(@owner['key'], 'envB', "envB", "For test systems only.")
+    job = @org_admin.promote_content(@env['id'],[{:contentId => contentD['id'],}])
+    wait_for_job(job['id'], 15)
+    consumer_cp.update_consumer({:environments => [{:name => @env.name}, {:name => envB.name}]})
+    ent = consumer_cp.list_entitlements()[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    new_serial = ent['certificates'][0]['serial']['serial']
+
+    expect(new_serial).to eq(serial)
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be false
+  end
+
+  it 'should re-gen consumer entitlement cert when environment priority are reversed and content
+    originates from different environment' do
+    envB = @org_admin.create_environment(@owner['key'], 'envB', "envB", "For test systems only.")
+    consumer = @org_admin.register(random_string('testsystem'), :system, nil, {}, nil, nil, [], [],
+      [{ 'id' => @env['id']}, { 'id' => envB['id']}])
+
+    expect(consumer['environments']).not_to be_nil
+
+    consumer_cp = Candlepin.new(nil, nil, consumer['idCert']['cert'], consumer['idCert']['key'])
+    product = create_product
+    contentA = create_content
+    @cp.add_content_to_product(@owner['key'], product['id'], contentA['id'])
+
+    job = @org_admin.promote_content(@env['id'],[{:contentId => contentA['id'],}])
+    wait_for_job(job['id'], 15)
+    job = @org_admin.promote_content(envB['id'],[{:contentId => contentA['id'],}])
+    wait_for_job(job['id'], 15)
+
+    pool = @cp.create_pool(@owner['key'], product['id'])
+    ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    serial = ent['certificates'][0]['serial']['serial']
+
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be true
+
+    consumer_cp.update_consumer({:environments => [{:name => envB.name}, {:name => @env.name}]})
+    ent = consumer_cp.list_entitlements()[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+    new_serial = ent['certificates'][0]['serial']['serial']
+
+    expect(new_serial).not_to eq(serial)
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be true
+  end
+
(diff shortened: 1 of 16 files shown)


src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1675,8 +1682,14 @@ else if (environmentName != null) {
...
 
-                // lazily regenerate certs, so the client can still work
-                poolManager.regenerateCertificatesOf(toUpdate, true);
+                if (firstPass) {
+                    // lazily regenerate certs, so the client can still work
+                    poolManager.regenerateCertificatesOf(toUpdate, true);
+                }
+                else {
+                    List<String> updatedEnvsList = selectEnvironment(preExistingEnvIds, validatedEnvIds);
+                    poolManager.regenerateCertificatesOf(toUpdate, toUpdate.getEnvironmentIds(), updatedEnvsList, true);
```
nikosmoum, line 1691: Without having looked at the PR in detail yet, one thing that I can see is that this whole process is going to be expensive, which is why I suggested to do it using a query. If hard to do with a single query, then two or three of them would be preferable than doing all this in java

wolfdale: Yes I understand, I am going to break the query into multiple JPQL queries. I wanted to write the spec test to test out couple of scenarios.

src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
```diff
@@ -334,6 +340,95 @@ public void regenerateCertificatesOf(Consumer consumer, boolean lazy) {
...
+            }
+        }
+
+        return ents;
+    }
+
+    private List<String> getContentUUIDsFromEntitlement(Entitlement entitlement) {
+        List<String> listOfPc = new ArrayList<>();
+        // Entitlement content UUIDs
+        Collection<ProductContent> pc = entitlement.getPool().getProduct().getProductContent();
```
Ceiu, line 403: Given the frequency at which we're calling and using the pool and product, they should be stored locally.

Also, add a line of whitespace between the blocks below.

src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
```diff
@@ -328,6 +333,122 @@ public void regenerateCertificatesOf(Consumer consumer, boolean lazy) {
         this.regenerateCertificatesOf(consumer.getEntitlements(), lazy);
     }
 
+    @Transactional
+    public void regenerateCertificatesOf(Consumer consumer, List<String> updatedEnvironments,
```
nikosmoum, line 337: Instead of adding yet another overloaded `regenerateCertificatesOf` method in this class (there are 7 already), can we break all this functionality down a bit? I don't think `EntitlementCertificateGenerator.java` needs to know anything about filtering entitlements; its sole responsibility should be generating them.
I would suggest moving the methods `filterEntitlements` and `getContentUUIDsFromEntitlement` into their own class called something like `EntitlementEnvironmentFilter`, which will be used at the ConsumerResource.java level; and once the list of filtered is returned, we pass them into one of the existing `EntitlementCertificateGenerator` methods

nikosmoum: This has the added benefit of making the filtering functionality much more unit-testable

src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1827,6 +1827,13 @@ public void regenerateCertificatesOf(Consumer consumer, boolean lazy) {
         this.ecGenerator.regenerateCertificatesOf(consumer, lazy);
     }
 
+    @Override
+    public void regenerateCertificatesOf(Consumer consumer, List<String> updatedEnvironments,
+        List<String> environmentChangeList, boolean lazy) {
+        this.ecGenerator.regenerateCertificatesOf(consumer, updatedEnvironments,
+            environmentChangeList, lazy);
+    }
```
nikosmoum, line 1835: follow up from my other comment: if we break out the entitlement filtering functionality to a separate class, this method is also not needed

src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
```diff
@@ -328,6 +333,122 @@ public void regenerateCertificatesOf(Consumer consumer, boolean lazy) {
...
+     * regenerated.
+     *
+     * @param consumer
+     *  The consumer for which to regenerate entitlement certificates
+     *
+     * @param updatedEnvironments
+     *  List of updated environment
+     *
+     * @param environmentChangeList
+     *  List of environment that gets changed
```
nikosmoum, line 368: Can we explain a bit more what this field is supposed to be? The explanation doesn't look much different from the one above. If it's about environments that got "added, removed, or reprioritized", then we need to explain it here. If we could find a better name for the variable it would be great, but I can't think of something better, so this is fine (as long as the javadoc explains in detail)

src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
```diff
@@ -328,6 +333,122 @@ public void regenerateCertificatesOf(Consumer consumer, boolean lazy) {
...
+        List<Entitlement> entitlements = filterEntitlements(consumer,
+            updatedEnvironments, environmentChangeList);
+
+        // TODO - Assumes only 1 entitlement certificate exists per entitlement
+        this.regenerateCertificatesOf(entitlements, lazy);
+    }
+
+    /**
+     * Select consumers entitlement whose entitlement certs need to be
+     * regenerated.
```
nikosmoum, line 359: Let's be more descriptive here and mention the `why` we only want to regenerate some entitlements and not others

nikosmoum (changes requested):
Suggested some refactoring and better documentation; will do another pass later

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1685,6 +1695,36 @@ else if (environmentName != null) {
...
+     *
+     * @param preExistingEnvs
+     *  Pre existing environment list
+     * @param updatedEnvs
+     *  Updated environment list
+     *
+     * @return
+     *  Returns the unique environmentIds that are added, removed or reordered
+     */
+    private List<String> getEnvironmentChangeList(List<String> preExistingEnvs, List<String> updatedEnvs) {
```
nikosmoum, line 1709: This method is conceptually also playing a role in the filtering of entitlements, so it could be rolled into the new `EntitlementEnvironmentFilter` class. The whole interface could be used with a fluent API as such:
```
new EntitlementEnvironmentFIlter()
    .setPreExistingEnvironments(preExistingEnvs)
    .setUpdatedEnvironments(updatedEnvs)
    .setConsumer(toUpdate)
    .filterEntitlements();
```

nikosmoum (approved):

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,147 @@
...
+
+        // Envs Added
+        envChangeList.addAll(CollectionUtils.subtract(this.updatedEnvs, this.preExistingEnvs));
+
+        // Envs removed
+        envChangeList.addAll(CollectionUtils.subtract(this.preExistingEnvs, this.updatedEnvs));
+
+        // Envs reordered (Index within list changes)
+        for (String env : this.updatedEnvs) {
+            if (this.preExistingEnvs.indexOf(env) != this.updatedEnvs.indexOf(env)) {
```
nikosmoum, line 81: This is not exactly correct. For example:
- if `preExistingEnvs: A B C` 
- and `updatedEnvs: B C`
- Then this code will consider both B and C to be re-ordered, while they haven't been re-ordered compared to each other.
Another example
- if `preExistingEnvs: A B` 
- and `updatedEnvs: A C B`
- Same here: A and B are not reordered (compared to each other).

The scenario we want to catch here is something like this:
- if `preExistingEnvs: A B` 
- and `updatedEnvs: B C A`
- Now A and B are reordered (compared to each other).

spec/environment_spec.rb
```diff
@@ -551,4 +551,268 @@ def new_content(id)
...
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentD['id']}.1")).to be false
+
+    consumer_cp.update_consumer({:environments => [{:name => envB.name}, {:name => @env.name}]})
+    ent = consumer_cp.list_entitlements()[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentD['id']}.1")).to be false
+  end
```
nikosmoum, line 817: Let's also add spec tests for the removal scenarios

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -1301,4 +1301,17 @@ private Predicate ilike(CriteriaBuilder cb, Expression<String> expression, Strin
         Predicate[] array = new Predicate[predicates.size()];
         return predicates.toArray(array);
     }
+
+    public List<String> getProductContentUUIDsByEntitlementId(String entitlementId) {
+        String sql = "SELECT c.uuid FROM cp_entitlement e " +
+            "JOIN cp_pool p ON p.id = e.pool_id " +
+            "JOIN cp2_product_content pc ON pc.product_uuid = p.product_uuid " +
```
nikosmoum, line 1308: in a production scenario, this would never return any data, because the main product that is referenced on the pool (p.product_uuid) would never actually have any content itself. it is only engineering products that have contents.

I would suggest keeping this query as is, but also add a second query to this method, which instead of just this join, has 2 joins, to make sure it links content that are hanging on provided products:
```
JOIN cp2_product_provided_products pp ON p.product_uuid=pp.product_uuid
JOIN cp2_product_content pc ON pp.provided_product_uuid=pc.product_uuid
```

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,147 @@
...
+                    .getEnvironmentContentUUIDs(environmentId);
+
+                if (this.updatedEnvs.contains(environmentId)) {
+                    currentHigher.addAll(this.updatedEnvs
+                        .subList(0, this.updatedEnvs.indexOf(environmentId)));
+                }
+                else {
+                    // Case where this environment is being deleted
+                    currentHigher.addAll(this.updatedEnvs);
+                }
```
nikosmoum, line 120: This will work in specific sub-cases of when the environment is being deleted, but I think I found an issue with the original algorithm implementation. I will have another go at it to find what exactly needs done for the rest of deletion sub-cases

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -236,4 +234,15 @@ public String getEnvironmentIdByName(String ownerId, String envName) {
             .setParameter("consumerId", consumer.getId())
             .getResultList();
     }
+
+    public List<String> getEnvironmentContentUUIDs(String environmentId) {
```
nikosmoum, line 238: This method seems more appropriate for `EnvironmentContentCurator.java`

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,147 @@
...
+
+                for (String higherEnv : currentHigher) {
+                    envListByCurrentHigher.addAll(this.environmentCurator
+                        .getEnvironmentContentUUIDs(higherEnv));
+                }
```
nikosmoum, line 125: This is a query being run within 3 nested loops. In the case the deleteEnvironment API endpoint, where we'll be using this filtering code, we'll also be looping over multiple consumers, so that would be 4 nested loops, which means thousands of queries, which is very prohibitive. Let's make a first step here, and turn this loop into a single query that fetches the content uuids of all environment ids it is being passed; and then let's look for more optimizations.

I am really re-considering whether this task is really worth it; this might be less performant than actually regenerating all of the entitlements blindly (This is a post-scrum/longer discussion we should have after the holidays though).

nikosmoum (changes requested):

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,147 @@
...
+     *  List of entitlement selected for certificate regeneration
+     */
+    public List<Entitlement> filterEntitlements() {
+        List<Entitlement> entitlements = new ArrayList<>();
+        List<String> environmentChangeList = getEnvironmentChangeList();
+
+        for (Entitlement e : consumer.getEntitlements()) {
+            boolean entToRegen = false;
+            List<String> contentUUIDs = this.entitlementCurator
+                .getProductContentUUIDsByEntitlementId(e.getId());
```
nikosmoum, line 105: Looking at this loop:
- we really don't really need the entitlement objects fetched in memory; we only need their entitlement ids (in order to do further queries)
- Calling `consumer.getEntitlements()` is performing 1 query (because that collection is lazily loaded), and then within the loop we perform N more queries, which is 1 for each entitlement. This is a classic case of N+1, so I would propose to cut this whole thing down to a single query by changing the curator method to something like this:
`Map<String, List<String>> entitlementIdToUUIDList = entitlementCurator.getEntitledContentUUIDsByConsumer(consumer)`
where the maps key is the entitlement id and value is the list of UUIDs for each,
and then iterate through the keys:
```
for (String entitlementID : entitlementIdToUUIDList.keys()) {
    ...
}
```

nikosmoum (changes requested):

spec/environment_spec.rb
```diff
@@ -551,4 +551,268 @@ def new_content(id)
...
+    contentA = create_content
+    contentB = create_content
+    contentC = create_content
+    contentD = create_content
+    @cp.add_content_to_product(@owner['key'], product['id'], contentA['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentB['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentC['id'])
+    job = @org_admin.promote_content(@env['id'],[{:contentId => contentB['id'],}])
+    wait_for_job(job['id'], 15)
+    pool = create_pool_and_subscription(@owner['key'], product['id'], 10)
```
nikosmoum, line 686: use `create_pool` instead of this deprecated method (same applies for all the other tests in this PR).

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,147 @@
...
+     * being added, removed or re-prioritized for a consumer.
+     * Idea is to avoid regenerating all of the consumer's entitlements certs & filter the
+     * necessary entitlements who's certificates are actually needed to be re-generated.
+     *
+     * @return
+     *  List of entitlement selected for certificate regeneration
+     */
+    public List<Entitlement> filterEntitlements() {
+        List<Entitlement> entitlements = new ArrayList<>();
+        List<String> environmentChangeList = getEnvironmentChangeList();
```
nikosmoum, line 100: Also, I know we discussed this offline, but just adding a note for myself when I re-review after the break: We want to abstract this filtering functionality (and the two queries that will be made in total) a step further, by accepting multiple consumers (as opposed to performing 2 queries per consumer, which would be another N+1 problem case).

I'd say you can change the APIs on this class as you see the best fit for this

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1618,6 +1618,8 @@ else if (toUpdate.getContentAccessMode() != null) {
      */
     private boolean updateEnvironments(ConsumerDTO updated, Consumer toUpdate, boolean existing) {
         List<EnvironmentDTO> environments = updated.getEnvironments();
+        List<String> preExistingEnvIds =
+            toUpdate.getEnvironmentIds().isEmpty() ? null : toUpdate.getEnvironmentIds();
```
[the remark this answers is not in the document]
wolfdale: done.

src/main/java/org/candlepin/resource/ConsumerResource.java
wolfdale: Done

src/main/java/org/candlepin/model/EnvironmentCurator.java
wolfdale: Moved to `EnvironmentContentCurator` class.

src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
```diff
@@ -71,23 +72,25 @@
...
+    private EnvironmentCurator environmentCurator;
 
 
     @Inject
     public EntitlementCertificateGenerator(EntitlementCertificateCurator entitlementCertificateCurator,
         EntitlementCertServiceAdapter entCertServiceAdapter, EntitlementCurator entitlementCurator,
         PoolCurator poolCurator, EventSink eventSink, EventFactory eventFactory,
-        ContentAccessManager contentAccessManager, OwnerCurator ownerCurator) {
+        ContentAccessManager contentAccessManager, OwnerCurator ownerCurator,
+        EnvironmentCurator environmentCurator) {
```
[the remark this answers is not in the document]
wolfdale: Fixed

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,147 @@
...
+
+        for (Entitlement e : consumer.getEntitlements()) {
+            boolean entToRegen = false;
+            List<String> contentUUIDs = this.entitlementCurator
+                .getProductContentUUIDsByEntitlementId(e.getId());
+
+            for (String environmentId : environmentChangeList) {
+                List<String> currentHigher = new ArrayList();
+                Set<String> envListByCurrentHigher = new HashSet<>();
```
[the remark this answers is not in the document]
wolfdale: Done.

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,147 @@
...
+     * @return
+     *  List of entitlement selected for certificate regeneration
+     */
+    public List<Entitlement> filterEntitlements() {
+        List<Entitlement> entitlements = new ArrayList<>();
+        List<String> environmentChangeList = getEnvironmentChangeList();
+
+        for (Entitlement e : consumer.getEntitlements()) {
+            boolean entToRegen = false;
+            List<String> contentUUIDs = this.entitlementCurator
```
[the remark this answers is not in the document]
wolfdale: Done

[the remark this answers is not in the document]
wolfdale: Done

src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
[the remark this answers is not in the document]
wolfdale: Fixed

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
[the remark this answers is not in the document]
wolfdale: Done.

src/main/java/org/candlepin/controller/EntitlementCertificateGenerator.java
wolfdale: this method is now removed.

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
wolfdale: Fixed

src/main/java/org/candlepin/model/EntitlementCurator.java
wolfdale: Sure, Added the both type of queries within same method.
Shouldn't we write spec test closer to production scenario ? Lot of the spec test have the content attached to main product.

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
wolfdale: Reduced the number of queries to 2. Since we are only interested in the IDs, we now fetch everything before the processing begins.

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
wolfdale: Done.

spec/environment_spec.rb
wolfdale: Done

spec/environment_spec.rb
wolfdale: Done

src/main/java/org/candlepin/model/EntitlementCurator.java
wolfdale: Currently for this specific query, there are no guarding against IN operator. I'll add that once I have more clarity on how do we want to split these queries.

Januson (changes requested):
Changes look good. There's just a few missing tests and a few suggestions.

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,206 @@
...
+            .getEnvironmentContentUUIDs(allEnvIds);
+        Map<String, List<String>> mapOfEntitlementContentUUID = this.entitlementCurator
+            .getEntitlementContentUUIDs(consumerIds);
+
+        for (String entitlementId : mapOfEntitlementContentUUID.keySet()) {
+            boolean entToRegen = false;
+            List<String> entitlementContentUUIDs = mapOfEntitlementContentUUID.get(entitlementId);
+
+            for (String environmentId : environmentChangeList) {
+                List<String> currentHigher = new ArrayList();
```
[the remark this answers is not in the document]
wolfdale: Done

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,206 @@
...
+                if (this.updatedEnvs.contains(environmentId)) {
+                    currentHigher.addAll(this.updatedEnvs
+                        .subList(0, this.updatedEnvs.indexOf(environmentId)));
+                }
+                else {
+                    currentHigher.addAll(this.updatedEnvs);
+                }
+
+                for (String higherEnv : currentHigher) {
+                    Optional.ofNullable(mapOfEnvironmentContentUUID.get(higherEnv))
```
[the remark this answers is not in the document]
wolfdale: sure, fixed

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -1301,4 +1302,80 @@ private Predicate ilike(CriteriaBuilder cb, Expression<String> expression, Strin
         Predicate[] array = new Predicate[predicates.size()];
         return predicates.toArray(array);
     }
+
+    /*
```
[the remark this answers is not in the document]
wolfdale: Added description.

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -1301,4 +1302,80 @@ private Predicate ilike(CriteriaBuilder cb, Expression<String> expression, Strin
...
+
+        return contentMap;
+    }
+
+    private void getMapOfEntIDAndContentUUID(List<Object[]> rawData, Map<String, List<String>> map) {
+        for (Object[] obj : rawData) {
+            String entitlementId = obj[0].toString();
+            List<String> listOfContentUUIDs;
+
+            if (map.containsKey(entitlementId)) {
```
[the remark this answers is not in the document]
wolfdale: done.

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -1301,4 +1302,80 @@ private Predicate ilike(CriteriaBuilder cb, Expression<String> expression, Strin
...
+            else {
+                listOfContentUUIDs = new LinkedList<>();
+            }
+
+            listOfContentUUIDs.add(obj[1].toString());
+            map.put(entitlementId, listOfContentUUIDs);
+        }
+    }
+
+    /*
```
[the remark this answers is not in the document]
wolfdale: done.

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -1301,4 +1302,80 @@ private Predicate ilike(CriteriaBuilder cb, Expression<String> expression, Strin
...
+     * Get entitlements by ID (in bulk)
+     *
+     * @params
+     *  entitlementIds - list of entitlement IDs
+     *
+     * @returns
+     *  List of entitlements
+     */
+    public List<Entitlement> getEntitlements(Iterable<String> entitlementIds) {
```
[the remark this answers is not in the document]
wolfdale: sure, added the unit test.

src/main/java/org/candlepin/model/EnvironmentContentCurator.java
```diff
@@ -56,4 +60,52 @@ public EnvironmentContent getByEnvironmentAndContent(Environment e, Content cont
...
+     * Process the result list (row) which contains entitlement ID & content UUID to
+     * build a map.
+     */
+    private void processData(Map<String, List<String>> contentUUIDMap, List<Object[]> resultList) {
+        for (Object[] result : resultList) {
+            String envId = result[0].toString();
+
+            List<String> ids;
+
+            if (contentUUIDMap.containsKey(envId)) {
```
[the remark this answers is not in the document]
wolfdale: done

src/main/java/org/candlepin/model/EnvironmentContentCurator.java
```diff
@@ -56,4 +60,52 @@ public EnvironmentContent getByEnvironmentAndContent(Environment e, Content cont
             .list();
     }
 
+    /*
```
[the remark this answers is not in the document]
wolfdale: done

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1664,6 +1673,8 @@ else if (environmentName != null) {
         }
 
         if (validatedEnvIds.size() > 0) {
+            List<String> preExistingEnvIds =
```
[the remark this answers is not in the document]
wolfdale: fixed.

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+/*
```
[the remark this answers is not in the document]
wolfdale: done

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -1301,4 +1302,80 @@ private Predicate ilike(CriteriaBuilder cb, Expression<String> expression, Strin
         Predicate[] array = new Predicate[predicates.size()];
         return predicates.toArray(array);
     }
+
+    /*
+     *
+     */
+    public Map<String, List<String>> getEntitlementContentUUIDs(List<String> consumerIds) {
```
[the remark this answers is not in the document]
wolfdale: Waiting on this comment - https://github.com/candlepin/candlepin/pull/3173#discussion_r773143209
After which I'll add the Unit test

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,206 @@
...
+ * (addition, removal or re-prioritization) in consumer existing environment.
+ * This filtering avoids re-generating all the consumer entitlement certificates.
+ */
+public class EntitlementEnvironmentFilter {
+    private EntitlementCurator entitlementCurator;
+    private EnvironmentContentCurator environmentContentCurator;
+    private List<String> preExistingEnvs;
+    private List<String> updatedEnvs;
+    private List<String> consumerIds;
+    private List<String> envRemoved;
```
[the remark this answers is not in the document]
wolfdale: Initially I did the same thing, but it was pointed out in the review comments.
I implement this approach which I liked -https://github.com/candlepin/candlepin/pull/3173#discussion_r770536813

src/main/java/org/candlepin/model/EntitlementCurator.java
wolfdale: As discussed, we want to keep both of these queries. Added the Unit test for this method.

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
wolfdale: Fixed.

src/main/java/org/candlepin/model/EntitlementCurator.java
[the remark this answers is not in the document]
wolfdale: Unit test added.

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -0,0 +1,207 @@
...
+
+        for (int counter = 0; counter < this.updatedEnvs.size(); counter++) {
+            String env = this.updatedEnvs.get(counter);
+
+            if (order.containsKey(env)) {
+                if (order.get(env) != counter) {
+                    environmentChangeList.add(env);
+                }
+            }
+        }
```
nikosmoum, line 98: This implementation has the same problem as the previous one; it still will report some items as re-ordered, even though they were not re-ordered (compared to each other): https://gist.github.com/nikosmoum/cdec89f3607ffdf1774a6a671c0d92b3 (the 2 example unit tests here should pass, but they don't)

This is a place that would benefit from some unit testing, but being a private method we can't. I would suggest extracting out this as a separate public static method in a new ListUtil class, that will be unit testable

nikosmoum (changes requested):

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
wolfdale:
```
List preexisting = List.of("A", "B");
List updated = List.of("A", "C", "B");
```

In above scenario, the Priority of environment B is reduced. Which means that environment B got re-prioritized. Shouldn't we include in the environment change list ?

wolfdale: Added the new reordering logic in CP Utils class.

src/test/java/org/candlepin/util/UtilTest.java
```diff
@@ -394,4 +394,26 @@ void encodesSpecialCharacters() {
...
+        assertTrue(Util.getReorderedItems(preexisting, updated).isEmpty());
+    }
+
+    @Test
+    public void testReorderItemListDifferentOrder() {
+        List<String> preexisting = List.of("A", "B", "C");
+        List<String> updated = List.of("A", "C", "B");
+        List<String> reordered = Util.getReorderedItems(preexisting, updated);
+        assertFalse(reordered.isEmpty());
+        assertFalse(reordered.contains("A"));
```
nikosmoum, line 410: I think being more specific about asserting what got re-ordered is better (in this case asserting that C and B got re-ordered), since now this test could pass even if only one of C or B was in the reordered list

src/main/java/org/candlepin/util/Util.java
```diff
@@ -620,4 +621,37 @@ else if (temporalAccessor instanceof LocalDateTime) {
     public static String encodeUrl(String text) {
         return URLEncoder.encode(text, StandardCharsets.UTF_8);
     }
+
+    /**
+     * Returns the list items which got re-ordered in modified list.
```
nikosmoum, line 626: Let's explain in more detail that this method will only find the items that were re-ordered compared to each other (not as a side-effect of adding/removing items)

spec/environment_spec.rb
```diff
@@ -551,4 +551,469 @@ def new_content(id)
...
+    pool = @cp.create_pool(@owner['key'], product['id'])
+    ent = consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be true
+
+    consumer_cp.update_consumer({:environments => [{:name => envB.name}, {:name => @env.name}]})
+    ent = consumer_cp.list_entitlements()[0]
```
nikosmoum, line 783: All the spec test names are given names such as 'should re-gen / should NOT regen', but they aren't actually asserting whether the entitlement serial changed or not. For the ones that happen to have obvious changes in content (like content added/removed) that is not a problem, but for tests like this one, it is, so let's add this check to all of the tests.

spec/environment_spec.rb
```diff
@@ -551,4 +551,469 @@ def new_content(id)
...
+    x509 = OpenSSL::X509::Certificate.new(ent['certificates'][0]['cert'])
+    extensions_hash = Hash[x509.extensions.collect { |ext| [ext.oid, ext.value] }]
+
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentB['id']}.1")).to be true
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentA['id']}.1")).to be false
+    expect(extensions_hash.has_key?("1.3.6.1.4.1.2312.9.2.#{contentC['id']}.1")).to be false
+  end
+
+  it 'should re-gen consumer entitlement cert when environment priority are reversed' do
```
nikosmoum, line 755: The name of this test is not entirely accurate. We should only regen the entitlement cert when the priority is reversed, AND at least one content got added/removed or originated from a different environment than before. Given that this test has 2 environments that each provide 1 unique content, the reversion of order should not result in the entitlement being regenerated (hence my other comment on the need to actually check the entitlement serial changed or not).

In addition to adding the entitlement serial check, we should change this test to something more useful like, there is only one content (ContentA), that is provided by both environments, and when the environment order is reversed, we assert that the entitlement did get regenerated (and also change it's name to something like `should re-gen consumer entitlement cert when environment priority are reversed and content originates from different environment`)

nikosmoum (changes requested):

nikosmoum (changes requested):

src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -1301,4 +1302,62 @@ private Predicate ilike(CriteriaBuilder cb, Expression<String> expression, Strin
...
+                .createNativeQuery(getContentFromProduct)
+                .setParameter("consumerIds", block)
+                .getResultList();
+            getMapOfEntIDAndContentUUID(contentFromProduct, contentMap);
+        }
+
+        for (List<String> block : Iterables.partition(consumerIds, blockSize)) {
+            List<Object[]> contentFromProvidedProducts = this.getEntityManager()
+                .createNativeQuery(getContentFromProvidedProducts)
+                .setParameter("consumerIds", consumerIds)
```
nikosmoum, line 1347: this is still setting the consumerIds instead of the block we partitioned

wolfdale: good catch. fixed it.

spec/environment_spec.rb
```diff
@@ -551,4 +551,504 @@ def new_content(id)
...
+    @cp.add_content_to_product(@owner['key'], product['id'], contentA['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentB['id'])
+    @cp.add_content_to_product(@owner['key'], product['id'], contentC['id'])
+
+    job = @org_admin.promote_content(@env['id'],[{:contentId => contentB['id'],}])
+    wait_for_job(job['id'], 15)
+    job = @org_admin.promote_content(envB['id'],[{:contentId => contentD['id'],}])
+    wait_for_job(job['id'], 15)
+    job = @org_admin.promote_content(envB['id'],[{:contentId => contentB['id'],}])
+    wait_for_job(job['id'], 15)
```
nikosmoum, line 871: Unsure why you added this, but its good because it has uncovered a bug. If consumer has specified environments @env, envB, which both provide contentB, and their order is reversed (envB, @env), then the _entitlement_ should get regenerated (while right not for this test it isn't). This test is basically the same as the `should re-gen consumer entitlement cert when environment priority are reversed and content originates from different environment` one, but that one is running corrently (while this one isn't). Could it be because of the new changes in `isCertRegenerationRequired`? Anyway, it needs investigation

wolfdale: Fixed.
It was due to recent changes to the ```isCertRegenerationRequired``` method.

nikosmoum (approved):

PR #3174, opened 2021-12-03
[4.0] Upgrade mysql & mariadb driver versions

nikosmoum:


nikosmoum:
It seems upgrading mysql-connector-java breaks the liquibase version we use, and upgrading liquibase is not straightforward at the moment, so I'm closing this.

PR #3176, opened 2021-12-06
[M] ENT-3930: Compile against jss from the file system

nikosmoum:
- Candlepin is already loading jss (both the .jar and the .so) as a
  library from the file system when deployed in Tomcat.
  For compilation purposes though a fixed jss jar was used as a
  maven/gradle dependency in a fedorapeople repo. Also, for unit
  testing, the .so was loaded on startup using a hack in Java code.
  Now in gradle, load them both (.jar and .so) directly from
  the file system for compilation/unit testing purposes, and in maven
  use the jar (only) from the file system for compilation.
- Starting with Fedora 35 and RHEL9, the jss library has upgraded
  to major version 5, and the file names have now changed from
  'jss4.jar' to 'jss.jar', and from 'libjss4.so' to 'libjss.so'.
  Now they are loaded regardless of their name.


nikosmoum (commented):
marking as draft because there's more to this

nikosmoum:
Note: In case someone wanted to test these changes by building the rpm on fedora 29+, you'd also have to have the fix from https://github.com/candlepin/candlepin/pull/3183

Januson (approved):

PR #3177, opened 2021-12-06
2028765: Change to query for correct updating and performance [M]

wottop:
As written it was taking more time than needed while updating all consumers in the org.


Ceiu (changes requested):
Since we're updating the query, can we also use this opportunity to update the aliases to something a bit clearer? `consumer`, `hypervisor`, and `owner` would probably be sufficient here.

Ceiu (approved):

nikosmoum (approved):

PR #3181, opened 2021-12-10
[M] Removed PKCS1 conversion when writing PEM keys (ENT-4452)

Ceiu:
- Private keys will no longer be converted to PKCS1 when written
  to PEM files


Ceiu:
While this is part of the FIPS epic, it's small and simple enough that I've opted to merge it in directly rather than waiting for a feature branch.

nikosmoum (approved):

PR #3190, opened 2021-12-17
[M] Restored write lock on orphan product UUID query

Ceiu:
- Restored the write lock which was erroneously removed when
  updating the orphan product lookup
- Updated the conflict spec test to ensure the orphan cleanup
  job actually completes successfully

### DIFF
--- a/src/main/java/org/candlepin/model/OwnerContentCurator.java
+++ b/src/main/java/org/candlepin/model/OwnerContentCurator.java
@@ -36,6 +36,7 @@ import java.util.Set;
 
 import javax.inject.Singleton;
 import javax.persistence.EntityManager;
+import javax.persistence.LockModeType;
 import javax.persistence.Query;
 import javax.persistence.TypedQuery;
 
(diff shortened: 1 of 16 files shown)


Ceiu:
Note that how well this works depends on the transaction isolation and locking configuration used by the underlying DB.

Januson (changes requested):
Looks like there is still some issue. Spec test "Product Versioning should cleanup orphans without interfering with normal actions" keeps failing.

Ceiu:
After some probing, it doesn't look to be an issue with the transaction isolation itself, but a race condition with (product?) creation and update. Going to dig into this further.

Ceiu:
Added additional read locks on the version resolution, which addresses the issues we were seeing. However, Hibernate has some real hangups about locking on immutable entities which are already in the session. I don't think this is a problem for actual live runtime, but it destroys our test suite.

I'm very open to feedback on the bit where we check for hsqldb as a means of determining whether or not we're in our test environment. Primarily because I *haaaate* when the test environment has an impact on production code, and secondly, because checking the schema to work around an issue in Hibernate is janky as well.

In the interim, I'm going to reach out to some Hibernate folks and see if I can convince them to shore up the inconsistency with locking immutable entities, and hopefully just support it, as I feel we have a valid use case here for where such a thing could be needed.

src/main/java/org/candlepin/model/OwnerContentCurator.java
```diff
@@ -351,6 +371,10 @@ public int clearContentForOwner(Owner owner) {
             TypedQuery<Content> query = this.getEntityManager()
                 .createQuery(jpql, Content.class);
 
+            if (lockMode != null && !this.getDatabaseDialect().contains("hsql")) {
```
nikosmoum, line 374: It's unfortunate that we have to check for hsql here, but since we are, this warrants a good comment that explains why it's done

client/ruby/candlepin_api.rb
```diff
@@ -531,6 +531,33 @@ def trigger_job(job, async=false)
     end
   end
 
+  def wait_for_job(job_id, max_wait_time=30, cleanup_job=true)
```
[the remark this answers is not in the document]
Ceiu: I thought I remembered writing this before. I'll update it to include the cleanup job option and use that version.

src/main/java/org/candlepin/model/OwnerContentCurator.java
Ceiu: Done.

Ceiu:
Something still seems off. I cannot get it to fail on my end against mysql, but the spec tests fail in CI 100% of the time.

nikosmoum:
@Ceiu Have you tried running it locally using the docker container?
E.g. 
./docker/test -c "cp-test -r product_versioning"

Ceiu:
Closing this PR as the focus of the fix has changed entirely. Will reopen with new details.

PR #3195, opened 2022-01-04
ENT-3901: SCA cert regeneration for multi-environment consumers

Januson:
- Handle content from multiple environments in content access certs
- Remove OU from SCA DN

### DIFF
--- a/src/main/java/org/candlepin/controller/ContentAccessManager.java
+++ b/src/main/java/org/candlepin/controller/ContentAccessManager.java
@@ -19,6 +19,7 @@ import org.candlepin.config.ConfigProperties;
 import org.candlepin.config.Configuration;
 import org.candlepin.controller.util.ContentPrefix;
 import org.candlepin.controller.util.PromotedContent;
+import org.candlepin.controller.util.ScaContainerContentPrefix;
 import org.candlepin.controller.util.ScaContentPrefix;
 import org.candlepin.model.CertificateSerial;
 import org.candlepin.model.CertificateSerialCurator;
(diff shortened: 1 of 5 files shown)


wolfdale (changes requested):
Minor suggestions, apart from that changes looks good to me.

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -465,45 +477,8 @@ else if (type.isType(ConsumerTypeEnum.HYPERVISOR)) {
...
-            .append(", O=")
-            .append(Rdn.escapeValue(owner.getKey()));
-
-        if (consumer.getEnvironmentId() != null) {
-            Environment environment = this.environmentCurator.getConsumerEnvironment(consumer);
-            sb.append(", OU=").append(Rdn.escapeValue(environment.getName()));
-        }
-
-        return sb.toString();
+        return "CN=" + consumer.getUuid() + ", O=" + Rdn.escapeValue(owner.getKey());
```
wolfdale, line 481: Using StringBuilder instead of raw Strings would be more appropriate here.

src/test/java/org/candlepin/controller/util/ScaContainerContentPrefixTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import java.util.List;
+import java.util.stream.Stream;
+
```
wolfdale, line 31: Please add javadoc here

wolfdale (changes requested):

src/main/java/org/candlepin/controller/ContentAccessManager.java
wolfdale: With recent Java (8 or higher) versions JVM internally make use of StringBuilder when we try to concatenation simple string just like above. So we can skip this comment.

wolfdale (approved):

PR #3196, opened 2022-01-07
2036746: Fix Illegal embedded sign character

Januson:
- Update testdata to generate content id without dash
- Strip non-numeric characters from content id before we use it for oid

### DIFF
--- a/bin/deployment/test_data_importer.py
+++ b/bin/deployment/test_data_importer.py
@@ -549,7 +549,7 @@ def map_content_to_product(cp, owner_key, product_data, content_data):
         for content_id, enabled in product_data['content']:
             # Create a unique content specifically for this product-content pairing.
             content = find_content(content_id).copy()
-            content_id = '{pid}-{cid}'.format(pid=product_data['id'], cid=content_id)
+            content_id = '{pid}{cid}'.format(pid=product_data['id'], cid=content_id)
 
             # Modify the base content data so that it is unique for this product
             content['id'] = content_id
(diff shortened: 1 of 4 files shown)


barnabycourt:
If the test was embedding an illegal sign character, but that character was being accepted by the API, shouldn't the validation be updated for the create ProductContent to disallow creation with the invalid data?

Januson:
@barnabycourt Sure. If we are ok with changing the API we can start rejecting them at creation. I updated only the OID creation as it keeps the API behavior.

Ceiu:
Given that this is a limitation of v1 certs, and we've never imposed this limitation before, it probably doesn't make sense to do so now. Are there other factors here that might cause problems if there are non-numeric characters in a content ID?

nikosmoum:
I agree with Chris. If non-numeric characters in ids already work with v3 certs, and there are users out there that use them, we'd break them, so I don't think we want to validate at the API. The current change seems better since it only happens at the v1 cert generation path; either that, or throwing a human readable exception at this point that says 'ids with non-numeric characters are not supported for clients with v1 certificates'

Ceiu:
Do we have a means of limiting a CP deployment to only v1 certs? If not, it probably doesn't make sense to add the limit on the API. 

My only concern with this, I guess, is the question of whether or not the content ID being slightly different in the cert OID has any impact with a content lookup or fetch op from a v1-restricted client. Are we at risk of any ID desync that we'd need to address?

nikosmoum:
@Ceiu I think in order to avoid such issues, it might be cleaner to just throw a descriptive error during the v1 cert generation (instead of the current error). Especially since no customer has actually complained about this (it was find by us).

nikosmoum (approved):
ACK from me

PR #3200, opened 2022-01-17
[M] Updated locking model surrounding orphan entity cleanup

Ceiu:
- Added the cp_system_locks table to act as a basis for
  a system-wide read/write locking system
- Added AbstractHibernateCurator.getSystemLock for obtaining
  named locks backed by the database
- Removed the write locks on queries used by orphan entity
  cleanup and various product and content versioning operations
- Added use of system locks during refresh, product or content
  API endpoints, and the orphan entity cleanup job
- Updated the conflict spec test to ensure the orphan cleanup
  job actually completes successfully

### DIFF
--- a/spec/content_versioning_spec.rb
+++ b/spec/content_versioning_spec.rb
@@ -222,7 +222,7 @@ describe 'Content Versioning' do
     length = 100
 
     # Repeat this test a few(ish) times to hopefully catch any synchronization error
-    (1..5).each do
+    (1..10).each do
       o1_uuids = []
       o2_uuids = []
 
@@ -253,10 +253,15 @@ describe 'Content Versioning' do
       end
 
       sleep 1
-      @cp.trigger_job("OrphanCleanupJob");
+      job = @cp.trigger_job("OrphanCleanupJob")
 
       updater.join
       generator.join
+      job_status = wait_for_job(job['id'])
+
+      # Verify the orphan cleanup job completed successfully
+      expect(job_status).to_not be_nil
+      expect(job_status['state'].upcase).to eq('FINISHED')
 
       # Verify the contents created/updated still exist
       o1_uuids.each do |uuid|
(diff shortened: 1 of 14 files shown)


nikosmoum (commented):
This looks great, but playing the devil's advocate a bit, could this deadlock not happen in theory?:

Thread1 OrphanCleanup locks: product
Thread2 ContentManager locks: content
Thread1 OrphanCleanup tries to lock: content (WAITING on Thread2)
Thread2 ContentManager calls into ProductManager, tries to lock: product (WAITING on Thread1)

Knowing that ContentManager might need to call into ProductManager, should we maybe have ContentManager acquire both product+content locks (in this order) preemptively?

wolfdale:
@nikosmoum I think ContentManager (& ProductManager) are trying to acquire `PESSIMISTIC_READ` lock. It allows other transactions reading data but deny any transaction trying to acquire `PESSIMISTIC_WRITE` lock.

I think will look something like this -

Thread1 OrphanCleanup locks: product - PESSIMISTIC_WRITE - Lock Granted
Thread2 ContentManager locks: content - PESSIMISTIC_READ - Lock Granted
Thread1 OrphanCleanup tries to lock: content - PESSIMISTIC_WRITE - Failed
Thread2 ContentManager calls into ProductManager, tries to lock: product PESSIMISTIC_READ - Lock Granted

nikosmoum:
Ah right, that looks correct :+1: so probably we're safe

Ceiu:
Hrmm... you might actually have a point here. Once a write lock is obtained, the read lock will get blocked. So the example given by @wolfdale would actually deadlock instead of granting on the last step. If we flip the lock order around in orphan cleanup, we're probably safe. Alternatively, we could make it all one lock if there's still a deadlock path that's high enough risk that we need to care about it. We're at an edge case of an edge case of an edge case here.

nikosmoum:
@Ceiu If we're flipping the lock order in OrphanCleanup, we should probably also flip it to be the same in Refresh, or otherwise that would definitely deadlock, no?

nikosmoum (approved):

PR #3202, opened 2022-01-19
[3.2] Use new maven mead target

nikosmoum:
- This avoids a bug in gettext, which has been fixed in RHEL9,
  which resulted in no translation classes being compiled.


Ceiu:
Does this work for builds not targeting RHEL9?

nikosmoum:
@Ceiu We don't build against RHEL9 yet. This is only about running the maven build in RHEL9 (not the RPM builds), as a way to get around a bug that gettext has in RHEL8 and earlier (and otherwise it doesn't matter which OS the maven build runs on)

nikosmoum:
Also, this is a cherry-pick from master/4.0, which already have this change (and it works)

PR #3206, opened 2022-01-21
2044944: System purpose compliance updates to correct db column; ENT-4600

wottop:


nikosmoum (changes requested):
ACK on the change, but let's also add a unit test for this use case

nikosmoum (approved):

nikosmoum:
There's a checkstyle failure to be fixed, but after that it's good to merge

nikosmoum:
@wottop Let's port this to 4.0 too, using https://bugzilla.redhat.com/show_bug.cgi?id=2044946

PR #3209, opened 2022-02-01
[M] Updated Vagrantfile to support el8-ready Ansible role (ENT-4468)

Ceiu:
- Updated the Vagrant files to support the forthcoming Candlepin
  Ansible role for individual el7 and el8 support
- Updated bash_functions to support either the mysql or mariadb
  connector jars (el7/el8 support)
- Improved Tomcat version detection logic in the deploy script

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,8 +1,14 @@
 VAGRANTFILE_API_VERSION = "2"
 
+ANSIBLE_TAGS_VAR = "ansible_tags"
+ANSIBLE_SKIP_TAGS_VAR = "ansible_skip_tags"
+ANSIBLE_VAR_PREFIX = "cp_"
+
+
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  config.vm.box = "centos/7"
   config.vm.synced_folder ".", "/vagrant", type: "sshfs", sshfs_opts_append: "-o nonempty"
+  config.vm.host_name = "candlepin.example.com"
+  config.ssh.forward_agent = true
 
   # Set up the hostmanager plugin to automatically configure host & guest hostnames
   if Vagrant.has_plugin?("vagrant-hostmanager")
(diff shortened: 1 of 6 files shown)


Ceiu:
This PR has a co-dependency on the Ansible role changes from https://github.com/candlepin/ansible-role-candlepin/pull/19

bin/deployment/bash_functions
```diff
@@ -158,13 +173,17 @@ init_mysql_jdbc() {
...
+    select_jdbc_jar "/usr/lib/java/mariadb-java-client.jar" "/usr/share/java/mysql-connector-java.jar"
+    evalrc $? "No JDBC jar file available for MariaDB/MySQL"
+
+    JDBC_DRIVER="com.mysql.cj.jdbc.Driver"
     JDBC_URL="jdbc:mysql://$db_host/$db_name"
 
-    # Change once the mariadb-java-client RPM is available in EPEL
+    # FIXME:
+    # Use the mariadb driver and URL when the mariadb-java-client RPM is available in EL7 and the
+    # Liquibase errors are fully resolved with the newer drivers
```
nikosmoum, line 185: Did you hit these liquibase errors after all?

vagrant/candlepin.yml
```diff
@@ -1,16 +1,25 @@
...
-    candlepin_checkout: /vagrant
+    candlepin_home: /vagrant
+
+    cp_configure_postgresql: true
+    cp_configure_mariadb: true
+    cp_configure_user_env: true
+    cp_configure_debugging: true
+
+    cp_git_checkout: false
+    cp_deploy: false
```
nikosmoum, line 21: why is this set to false?

Vagrantfile
```diff
@@ -1,8 +1,13 @@
...
 
+ANSIBLE_TAGS_VAR = "ansible_tags"
+ANSIBLE_SKIP_TAGS_VAR = "ansible_skip_tags"
+ANSIBLE_VAR_PREFIX = "cp_"
+
+
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  config.vm.box = "centos/7"
   config.vm.synced_folder ".", "/vagrant", type: "sshfs", sshfs_opts_append: "-o nonempty"
+  config.vm.host_name = "candlepin.example.com"
```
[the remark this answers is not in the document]
Ceiu: We don't *need* it, but it's a fallback for if we ever add one and forget to explicitly set it.

bin/deployment/bash_functions
Ceiu: Specifically when changing the JDBC_DRIVER to MariaDB's. It doesn't like something with a few places we use NULL. I forget the details and I didn't look too deeply into it.

vagrant/candlepin.yml
Ceiu: For spinning up a Vagrant VM where we're mapping in the CP directory, we don't need to checkout a specific branch during the provisioning step, nor do we need to deploy. However, QE uses those flags to spin up a machine that checks out a specific branch and deploys it without having a local copy (and not running in a Vagrant-managed VM).

nikosmoum: I understand why `cp_git_checkout` is set to false here, since we're mapping from the host, but why not deploy immediately? Doing a vagrant up and then manually deploying is a bit less convenient, when ansible can deploy it for us

Ceiu: Given how we use our VMs (we're not constantly recreating them on a daily basis), I didn't think that'd be particularly useful. If it is on a one-off, you can always run the following to get that behavior:

```
cp_deploy=true vagrant up [el7/el8]
```

nikosmoum: I just tried that but it didn't work. Are you sure that variable is being passed in like that?:
```
TASK [candlepin : Ensure the Candlepin directory exists] ***********************
ok: [el7]

TASK [candlepin : Install Candlepin gem bundle] ********************************
changed: [el7]

TASK [candlepin : Ensure the Candlepin directory exists] ***********************
skipping: [el7]

TASK [candlepin : Check for legacy Candlepin deploy script] ********************
skipping: [el7]

TASK [candlepin : Update Candlepin deploy script path to use legacy path] ******
skipping: [el7]

TASK [candlepin : Deploy Candlepin] ********************************************
skipping: [el7]
```

Ceiu: There's a block in the vagrantfile which is supposed to be copying in environment variables starting with "cp_" to the ansible vars. Either that's not working, or my understanding of the variable precedence is incorrect. I'll take a look.

Ceiu: Fixed the issue that was preventing the passthrough from working as expected.

Ceiu:
@nikosmoum 
Having done it now, I don't know that a script is even necessary to do the custom role bits. You literally just create a directory and then a single text file and put your stuff in there. I've outlined the process in the README.md. Let me know if you feel that's enough.

Ceiu:
Post-merge tasks:
- [ ] Update docs on candlepinproject.org

PR #3214, opened 2022-02-08
Allow SCA update on owner in standalone without need for import

wottop:


spec/content_access_spec.rb
```diff
@@ -587,166 +551,6 @@
     }.to raise_exception(RestClient::BadRequest)
   end
 
-  it "will show up in the owner at distributor on export/import" do
```
Ceiu, line 590: Should we still verify the manifest includes this information for old Candlepins? What is the plan for SCA distribution?

spec/candlepin_scenarios.rb
```diff
@@ -831,20 +831,10 @@ def set_content_access_mode_for_standalone_owner(owner, access_mode, force_distr
...
-      })
-    else
-      # Create a new owner that is SCA-capable, but not SCA-enabled
-      owner = create_owner(random_string("test_owner"), nil, {
-          'contentAccessModeList' => 'org_environment,entitlement',
-          'contentAccessMode' => "entitlement"
-      })
-      # and then indirectly make the owner SCA-enabled
-      owner = set_content_access_mode_for_standalone_owner(owner, 'org_environment')
-    end
```
nikosmoum, line 847: Now that we don't have this restriction, we no longer need the `set_content_access_mode_for_standalone_owner` method either, so we can delete it

spec/content_access_spec.rb
nikosmoum: Yes I think we should verify this. The export should still be exporting the content access mode value

wottop: I believe there is another test that finds the value in the export file. I will check.
This test is no longer valid as it checks the org on the receiving end after import.

wottop: I did add a test to ensure that the contentAccessMode is still in the export

nikosmoum (approved):

PR #3215, opened 2022-02-09
[4.0] 1964376: Fix to include products from new entitlements selected via auto-attach or bulk-pool attach operations

wolfdale:
Fix -
While generating entitlement certificates, products from existing entitlements on a consumer are included as entitled products, and not from new entitlements selected as a result of an auto-attach or bulk-pool attach.
This fix is to include products from new entitlements selected via auto-attach or bulk-pool attach operations as well.

(cherry picked from commit 58fe40e2febaccfaf8b627d95df01ad666e74f6d)

### DIFF
--- a/server/spec/conditional_content_spec.rb
+++ b/server/spec/conditional_content_spec.rb
@@ -1,5 +1,6 @@
 require 'spec_helper'
 require 'candlepin_scenarios'
+include CertificateMethods
 
 describe 'Conditional Content and Dependent Entitlements' do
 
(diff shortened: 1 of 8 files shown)


nikosmoum (approved):

nikosmoum (changes requested):
The commit message starts with the wrong BZ id (1964379 instead of 1964376). It needs updating before merge

wolfdale:
Fixed the bug ID.

PR #3219, opened 2022-02-11
ENT-3899: Support environment deletion in multi-env

Januson:


spec/environment_spec.rb
```diff
@@ -545,6 +550,131 @@ def promote_content(content1, env1)
...
+                            facts={'system.certificate_version' => '3.3', "uname.machine" => "x86_64"}, random_string("consumer2"),
+                            @owner['key'], [], [], [{ 'id' => env1['id']}, { 'id' => env2['id']}])
+
+    @cp.delete_environment(env1['id'])
+
+    lambda {
+      @cp.get_consumer(consumer1['uuid'])
+    }.should raise_exception(RestClient::Gone)
+
+    cons = @cp.get_consumer(consumer2['uuid'])#.should_not be nil
```
nikosmoum, line 579: leftover commented out code

spec/environment_spec.rb
```diff
@@ -545,6 +550,131 @@ def promote_content(content1, env1)
...
+    cons['environments'].length.should == 1
+  end
+
+  it 'should only enable content promoted by consumers environment' do
```
nikosmoum, line 583: This test is a duplicate of `'filters content not promoted to environment'` from `environment_cert_v3.rb`

spec/environment_spec.rb
```diff
@@ -545,6 +550,131 @@ def promote_content(content1, env1)
     }
   end
 
+  it 'should delete consumer together with his last environment' do
```
nikosmoum, line 553: This test already covers what `should cleanup consumers of deleted environment` is asserting, so I think we should remove that test in favor of this one

spec/environment_spec.rb
```diff
@@ -545,6 +550,131 @@ def promote_content(content1, env1)
...
+    @cp.delete_environment(env1['id'])
+
+    certs = consumer_cp.list_certificates
+    certs.length.should == 2
+
+    contents = contents_of(certs[0]) + contents_of(certs[1])
+    contents.size.should == 1
+  end
+
+  it 'should revoke entitlements granted by the demoted content' do
```
nikosmoum, line 646: This is a duplicate of `'regenerates certificates after promoting/demoting content'` in `environment_cert_v3.rb`

spec/environment_spec.rb
```diff
@@ -545,6 +550,131 @@ def promote_content(content1, env1)
...
+    consumer_cp.consume_pool(pool['id'], {:quantity => 1})[0]
+    certs = consumer_cp.list_certificates
+    certs.length.should == 1
+
+    json_body = extract_payload(certs[0]['cert'])
+    json_body['products'].length.should == 1
+    json_body['products'][0]['content'].length.should == 1
+  end
+
+  it 'should revoke entitlements granted by the deleted environment' do
```
nikosmoum, line 612: An improvement on this test would be to rename it `should regenerate only entitlements affected by a deleted environment`, and the assertions at the end should be that a) the entitlement from pool2 did not get regenerated, by comparing the old/new serials (and contains 1 content), and b) the entitlement from pool1 _did_ get regenerated, by comparing the old/new serials (and contains 0 content)

nikosmoum (changes requested):

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
```diff
@@ -92,94 +53,128 @@ public EntitlementEnvironmentFilter setConsumerToBeUpdated(List<String> consumer
...
+    private Set<String> getEnvironmentChangeList(List<String> envRemoved,
+        List<String> preExistingEnvs, List<String> updatedEnvs) {
+
+        Set<String> environmentChangeList = new HashSet<>();
+
+        // Envs Added
+        environmentChangeList.addAll(CollectionUtils.subtract(updatedEnvs, preExistingEnvs));
+
+        // Env removed are kept in separate list for special use cases
+        envRemoved.addAll(CollectionUtils.subtract(preExistingEnvs, updatedEnvs));
```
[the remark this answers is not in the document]
nikosmoum: looking at the calling code, this might actually be a problem, because the envRemoved variable is shared between all consumers, so this `addAll` keeps adding environments of each consumer we loop over into it, and we don't want that (we want envRemoved to be per-consumer only)

nikosmoum (changes requested):
A few changes, but otherwise this does the job

spec/environment_spec.rb
Januson: Done.

spec/environment_spec.rb
Januson: Done.

spec/environment_spec.rb
Januson: Done.

spec/environment_spec.rb
Januson: Done.

spec/environment_spec.rb
Januson: Done.

src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -1863,6 +1863,11 @@ public void regenerateCertificatesOf(Owner owner, String productId, boolean lazy
         this.ecGenerator.regenerateCertificatesOf(owner, productId, lazy);
     }
 
+    @Override
+    public void regenerateCertificatesByEntitlementIds(Collection<String> entitlements, boolean lazy) {
```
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1692,18 +1689,14 @@ else if (environmentName != null) {
...
-                        .regenerateCertificatesByEntitlementIds(entitlementsToBeRegenerated, true);
+                    EnvironmentUpdates environmentUpdates = new EnvironmentUpdates();
+                    environmentUpdates.put(toUpdate.getId(), preExistingEnvIds, validatedEnvIds);
+                    Set<String> entitlementsToBeRegenerated = entitlementEnvironmentFilter
+                        .filterEntitlements(environmentUpdates);
+                    poolManager.regenerateCertificatesByEntitlementIds(entitlementsToBeRegenerated, true);
                 }
                 else {
-                    entitlementCertificateGenerator.regenerateCertificatesOf(toUpdate, true);
+                    poolManager.regenerateCertificatesOf(toUpdate, true);
```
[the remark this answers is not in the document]
Januson: I saw that pool manager already has related methods so I put it there, but you are right. Using generator directly is better idea.

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -157,42 +163,81 @@ public void deleteEnvironment(@Verify(Environment.class) String envId) {
...
             return;
         }
+        // No need to delete env from consumers manually. It will be handled by cascading delete.
+        List<String> consumerIds = consumers.stream()
+            .map(Consumer::getId)
+            .collect(Collectors.toList());
+
+        Set<String> entitlementsToBeRegenerated = entitlementEnvironmentFilter
+            .filterEntitlements(prepareEnvironmentUpdates(environment, consumerIds));
+        poolManager.regenerateCertificatesByEntitlementIds(entitlementsToBeRegenerated, true);
```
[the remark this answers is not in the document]

src/main/java/org/candlepin/resource/util/EntitlementEnvironmentFilter.java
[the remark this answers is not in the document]
Januson: Good catch. Done.

nikosmoum (approved):

PR #3221, opened 2022-02-14
Setup generated client for java spec tests

wottop:
This is step 1 of the migration to java spec tests

1. is the package structure acceptable?
2. are there any other naming issues?
3. should this be squashed before merging?

I ran this last on the new el8 vagrant instance.

The execution is "gradlew clean spec"


spec-tests/src/test/java/org/candlepin/spec/OwnerResourceTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+@SpecTest
+public class OwnerResourceTest {
+
+    private ApiClientFactory apiClient = ClientBuilders.instance();
+    private OwnerClient api;
+
+    @BeforeEach
+    void setUp() {
+        this.api = new OwnerClient(apiClient.createInstance());
```
Ceiu, line 42: This kind of setup is going to be everywhere, and several spec tests use several APIs. We may end up in a situation where something like the DatabaseTestFixture ends up being a common parent class.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/PropertySource.java
```diff
@@ -0,0 +1,22 @@
...
+ *  Red Hat trademarks are not licensed under GPLv2. No permission is
+ *  granted to use or replicate Red Hat trademarks that are incorporated
+ *  in this software or its documentation.
+ */
+
+package org.candlepin.spec.bootstrap.client;
+
+import java.util.Properties;
+
+public interface PropertySource {
```
Ceiu, line 20: This is a type-specific Provider interface.

Ceiu (changes requested):
Gave this a first pass; tried to focus on the most obvious stuff:

- PR needs a squash
- There is a bit of scaffolding which kind of exists in a place that something else already does
- How this will expand to some of the larger, more complex tests remains to be seen, but some of the boilerplate bits I've seen imply it's going to be a bit painful as-is (the setup/teardown stuff, needing to individually initiate a bunch of API wrapper classes, and the idea of needing wrapper classes at all).

Also, the PR itself doesn't build for me on a fresh build. Attempting to run it using the command provided in the PR comment results in the following:

```
OwnerResourceTest > shouldCreateOwner() FAILED
    org.candlepin.ApiException: java.net.ConnectException: Failed to connect to localhost/0:0:0:0:0:0:0:1:8443
        at org.candlepin.ApiClient.execute(ApiClient.java:916)
        at org.candlepin.resource.OwnerApi.createOwnerWithHttpInfo(OwnerApi.java:587)
        at org.candlepin.resource.OwnerApi.createOwner(OwnerApi.java:566)
        at org.candlepin.spec.OwnerResourceTest.shouldCreateOwner(OwnerResourceTest.java:55)
```

wottop:
Working forward with feature branch

PR #3224, opened 2022-02-15
[M] Fix Sonarqube jenkins task

nikosmoum:
- Use the new sonarqube candlepin project key:
  'chainsaw:candlepin-server'.
- Use the correct branch name variable for uploading a PR analysis
  instead of the default master branch analysis.


wottop:
ack

PR #3234, opened 2022-02-24
Migrate to JUnit5

Januson:
- Migrate remaining JUnit4 tests to JUnit5
- Remove JUnit4 dependency
- Update testing(JUnit, Mockito) dependencies to latest versions

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,7 +92,7 @@ ext {
     resteasy_version = "4.4.3.Final"
     jackson_version = "2.12.3"
     guice_version = "4.2.3"
-    junit_version = "5.7.0"
+    junit_version = "5.8.2"
 }
 
 dependencies {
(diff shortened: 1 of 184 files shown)


Ceiu:
Looks good.

We should also update all the uses of `@ParameterizedTest` to use the test name syntax so the output clearly indicates which test is being run; the default output without it is super unclear. If that's done in this PR or a follow up PR, I'm alright with either.

The format I've been using across the board is `@ParameterizedTest(name = "{displayName} [{index}]: {1}, {4}")` where the {1} and {4} in this case are the arguments that matter for the test case. The part following `[{index}]:` will change on a per-test basis.

Januson:
@Ceiu Good point. Gradle's test logging output leaves much to be desired. But I think we should address the cause rather than symptoms in individual tests. For example we could try [test-logger plugin](https://github.com/radarsh/gradle-test-logger-plugin) to make our test logging nicer.

nikosmoum (approved):
ack; I think any improvement in the output can go in a separate PR

PR #3235, opened 2022-02-24
[M] Addressed inconsistent test failures in ConsumerResourceTest

Ceiu:
- Split testSearchConsumersRequiresPagingForLargeResultSets test
  into two tests to avoid an issue where the mocked object would
  sometimes return the wrong value


Ceiu:
For reviewers: feel free to mash the "restart tests" button a number of times to ensure this isn't still a problem in Jenkins.

nikosmoum (approved):

PR #3237, opened 2022-02-25
[FIPS] Updated certificate generation and Tomcat configuration scripts (ENT-4447)

Ceiu:
- Updated gen_certs.sh to use FIPS-compatible certificate generation
  tools and options
- Updated update-server-xml.py in accordance with the changes made
  to gen_certs and current configuration standards
- Updated the deploy script in accordance with the changes made to
  gen_certs and update-server-xml.py


Ceiu:
Notes for reviewers: 

- No one aside from us uses this cert generation process or Tomcat configuration, so we're (thankfully) isolated from needing to coordinate with others as far as the changes to these scripts are concerned. Anyone using our deploy script to stand up Candlepin will still be in good shape here (once the TODO is sorted later).
- This will only operate in the F35 when FIPS is enabled. EL7 and EL8 should work fine if FIPS mode is not enabled. This is a known limitation due to a bug in Tomcat that was fixed in 9.0.38 which is not yet present in EL8. EL7 is not targeted for FIPS changes/concerns, so, at least for the moment, any failures there in this regard under FIPS are not a concern.
- Candlepin will still not fully start due to other issues that need to be sorted out as part of the FIPS effort. The expected result here is that under FIPS mode, the fatal error will show up in Tomcat's localhost log. If it fails to start due to issues in catalina.log, something is not working correctly. Under non-FIPS mode, Candlepin should be fully functional in all development environments.

bin/deployment/update-server-xml.py
```diff
@@ -12,7 +12,16 @@
 logging.basicConfig(level=logging.INFO, format="%(levelname)-7s %(message)s")
 logger = logging.getLogger('update_server_xml')
 
-
+SSL_CIPHERS = [
```
nikosmoum, line 15: We should also update cpsetup with the new config. I think the ideal would be to avoid duplication by extracting the cert/key generation and tomcat configuration in a separate python module, that can be used by both cpsetup and the deploy script, but that could be done as a separate task.

bin/deployment/update-server-xml.py
```diff
@@ -217,105 +215,104 @@ def _build_node(self):
...
+            ("ciphers", ",".join(SSL_CIPHERS))
         ])
 
         certificate = libxml2.newNode("Certificate")
         self._add_attributes(certificate, [
-            ("certificateKeystoreFile", "conf/keystore"),
-            ("certificateKeystorePassword", "password"),
-            ("certificateKeystoreType", "JKS"),
+            ('certificateFile', '/etc/candlepin/certs/candlepin-ca.crt'),
+            ('certificateKeyFile', '/etc/candlepin/certs/candlepin-ca.key'),
```
nikosmoum, line 294: Is there an advantage to using the direct PEM file? I was thinking using the keystore/truststore format would be more future-proof if we eventually wanted to have multiple CA certs (for splitting responsibilities on signing Identity and SCA certs with different keys). Although if that's hard to do it can always be left for later, since it is not an immediate need.

bin/deployment/update-server-xml.py
Ceiu: No one uses cpsetup, and the delta that has grown between it and deploy is one that's been the subject of several discussions and at least one epic with several tasks attempting to fix it. Both cpsetup and cpdb are in dire need of love, but both, I feel, are outside the scope of the FIPS work.

bin/deployment/update-server-xml.py
Ceiu: We need the cert/key file anyway for CP (as it is expecting it to not be in a keystore). We could certainly change it, but it would require a couple things here:

- CP is updated to support both reading files directly for backward compatibility purposes, and reading them from a keystore (and all of the fallout that comes of that)
- Even more branching on the configuration surrounding FIPS mode, as *which* keystore format we use will depend on what is available via keytool and the security providers invoked by Tomcat's protocol/ssl handler.

It's important to note that this configuration for TC is not used outside of our team. Sat has their own setup process with cert generation and TC configuration, as does IT. If we're going to move forward on not having the PEM files on the filesystem, I would argue it's a separate task entirely as it has major impacts on our stakeholders and how CP itself functions.

bin/deployment/update-server-xml.py
nikosmoum: cpsetup is the only way of getting candlepin deployed after installing it as an rpm (which is something that we occasionally have to test), aside from actually doing all that work by hand. But I agree it is outside of the scope for FIPS (just that eventually we should update it)

Ceiu: I wasn't aware of the RPM thing (since we do it ourselves so seldomly. That being the case, I'll do a very ugly update of cpsetup (since it looks pretty hack itself) sometime on Monday to basically just do the current changes without the legacy keystore bits.

In the future, we should probably set it up so that the RPM install script itself calls cpsetup so it's not necessary to do it after the fact -- especially if we plan to move to Spring Boot. (Also merge cpdb into cpsetup and convert the whole mess into a proper python package/module of sorts).

bin/deployment/update-server-xml.py
nikosmoum: > We need the cert/key file anyway for CP (as it is expecting it to not be in a keystore). We could certainly change it, but it would require a couple things here:
> 
>     * CP is updated to support both reading files directly for backward compatibility purposes, and reading them from a keystore (and all of the fallout that comes of that)

Why would CP itself need to be updated for reading from keystore? that's Tomcat's job to handle. That was what it was always doing (the server.xm configured to use `keystoreFile`, etc.). In fact, using the cert/key PEM files directly (instead of using a keystore) was only recently introduced in [Tomcat 9](https://bz.apache.org/bugzilla/show_bug.cgi?id=59344).

> It's important to note that this configuration for TC is not used outside of our team. Sat has their own setup process with cert generation and TC configuration, as does IT. If we're going to move forward on not having the PEM files on the filesystem, I would argue it's a separate task entirely as it has major impacts on our stakeholders and how CP itself functions.

I wouldn't say not having PEM files on the filesystem at all. We'd still need to generate them and add them to the keystore, and tomcat would be reading the keystore only (as it always did before this PR). (and from what I understand, that's also what Satellite had been [doing](https://github.com/theforeman/puppet-candlepin/blob/49fec4a9624e2d65a87f6db919bb59f1fb17aa0c/templates/tomcat/server.xml.erb#L83-L87)).


>     * Even more branching on the configuration surrounding FIPS mode, as _which_ keystore format we use will depend on what is available via keytool and the security providers invoked by Tomcat's protocol/ssl handler.

That is definitely a pain, and probably why we should leave this as is, and use a keystore when we have a need for it in the future.

bin/scripts/code/setup/cpsetup
```diff
@@ -317,7 +328,7 @@ def main(argv):
 
     print(command)
 
-    run_command(command)
+    # run_command(command)
```
nikosmoum, line 331: Was this left over from testing?

bin/scripts/code/setup/cpsetup
```diff
@@ -173,43 +177,50 @@ class TomcatSetup(object):
...
+            run_command_with_sudo('mkdir -p %s' % self.cert_home)
 
-        # Tomcat 8.5+ requires the cert be separate from the key when used
-        # as both a keystore and a trust store
-        run_command_with_sudo('keytool -keystore %s -noprompt -importcert -storepass password '
-                              '-alias candlepin_ca -file %s' % (self.keystore, self.ca_cert))
-        run_command_with_sudo('chmod a+r %s/*' % self.cert_home)
+        # if os.path.exists(self.ca_key) and os.path.exists(self.ca_cert):
+        #     print("Certificates already exist, skipping...")
+        #     return
```
nikosmoum, line 198: If this is not needed, let's remove it

nikosmoum (changes requested):

bin/deployment/deploy
```diff
@@ -82,18 +82,30 @@ isrpm() {
...
+
+    if [ ! -f $CP_KEYSTORE ]; then
+        info_msg "Legacy keystore does not exist; skipping symlink creation/update"
+        return 0 # nothing to do -- legacy keystore hasn't been generated
+    fi
+
+    if [ ! -f $CONTAINER_KEYSTORE ]; then
         info_msg "Symlinking keystore for tomcat"
-        $SUDO ln -s /etc/candlepin/certs/keystore $CONTAINER_CONF_DIR/keystore
+        $SUDO ln -s $CP_KEYSTORE $CONTAINER_KEYSTORE
```
nikosmoum, line 99: Running the deploy script on the Centos Stream 8 vm fails with:
```
...
Generating certificates...
Done. New certificate generated: /etc/candlepin/certs/candlepin-ca.crt
Generating legacy JKS keystore...
Certificate was added to keystore
Done. Keystore generated: /etc/candlepin/certs/tc_keystore.jks
Symlinking keystore for tomcat
ln: failed to create symbolic link '/etc/tomcat/keystore': File exists
```
And every time after the initial it still fails with:
```
...
Certificates already generated; using existing certs
Symlinking keystore for tomcat
ln: failed to create symbolic link '/etc/tomcat/keystore': File exists
```

bin/scripts/code/setup/cpsetup
Ceiu: Oops, yup. Restored.

bin/deployment/update-server-xml.py
Ceiu: > Why would CP itself need to be updated for reading from keystore? that's Tomcat's job to handle. That was what it was always doing (the server.xm configured to use keystoreFile, etc.). In fact, using the cert/key PEM files directly (instead of using a keystore) was only recently introduced in [Tomcat 9](https://bz.apache.org/bugzilla/show_bug.cgi?id=59344).

I'm aware, but CP itself also uses the certs/keys for setting the issuer and key identifier when creating certs. These are loaded directly from the PEM files (ConfigProperties.CA_CERT and ConfigProperties.CA_KEY), so no matter what we do with TC, CP still needs them available unless we're making a decision to move CP to load them cert/key from a keystore -- which invokes the need for backward compatibility I mentioned above, as well as determining which keystore formats we can use depending on system limitations.

bin/deployment/deploy
Ceiu: \*grumble\*
I fixed this once already. Hopefully I just accidentally reverted something. Looking into it.

bin/deployment/update-server-xml.py
nikosmoum: Looked into the [docs](https://tomcat.apache.org/tomcat-9.0-doc/config/http.html#Nested_Components) a bit more and apparently there is support  for multiple certificates to be associated with a single SSLHostConfig, so we don't need to use a keystore to have more than one cert.

bin/deployment/deploy
nikosmoum: nevermind, this was just me having deleted the /etc/candlepin/certs (to test with a 'fresh' checkout) directory without actually deleting the /etc/tomcat/keystore symlink. When I deleted both the dirt and symlink then it worked

nikosmoum (approved):

PR #3239, opened 2022-02-28
ENT-3756: Implementation of SCA-by-default for newly created Owners

wolfdale:
Changed the default content access mode to ORG_ENVIRONMENT for newly created orgs.

### DIFF
--- a/spec/refresh_pools_spec.rb
+++ b/spec/refresh_pools_spec.rb
@@ -43,6 +43,20 @@ describe 'Refresh Pools' do
     end
   end
 
+  it 'contains the proper return value when owner is auto generated' do
+    result = @cp.refresh_pools("auto-generated-owner", false, true)
+
+    if !is_hosted?
+      expect(result).to be_nil
+    else
+      expect(result).to eq("Pools refreshed for owner: auto-generated-owner")
+
+      owner = @cp.get_owner("auto-generated-owner")
+      expect(owner['displayName']).to eq("auto-generated-owner")
+      expect(owner['contentAccessMode']).to eq("org_environment")
+    end
+  end
+
   it 'creates the correct number of pools' do
     owner = create_owner random_string('some-owner')
 
@@ -1688,5 +1702,4 @@ describe 'Refresh Pools' do
     expect(product_uuids.size).to eq(owners.size)
     expect(product_uuids).to all(eq(product_uuids.first))
   end
-
 end
--- a/src/main/java/org/candlepin/resource/OwnerResource.java
+++ b/src/main/java/org/candlepin/resource/OwnerResource.java
@@ -854,75 +854,7 @@ public class OwnerResource implements OwnersApi {
 
     @Override
     public OwnerDTO createOwner(OwnerDTO dto) {
-
-        // Verify that we have an owner key (as required)
-        if (StringUtils.isBlank(dto.getKey())) {
-            throw new BadRequestException(i18n.tr("Owners must be created with a valid key"));
-        }
-
-        Owner owner = new Owner();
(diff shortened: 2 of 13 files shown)


Ceiu (changes requested):

src/main/java/org/candlepin/controller/OwnerManager.java
```diff
@@ -203,4 +213,217 @@ private void removeAllContentForOwner(Owner owner) {
...
+     * @param entity
+     *  The entity instance to populate
+     *
+     * @param dto
+     *  The DTO containing the data with which to populate the entity
+     *
+     * @throws IllegalArgumentException
+     *  if either entity or dto are null
+     */
+    public void populateEntity(Owner entity, OwnerDTO dto) {
```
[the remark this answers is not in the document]
wolfdale: Moved back the logic to resource layer.

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -2317,4 +2248,73 @@ public void remove(String ownerKey, String contentId) {
         this.contentAccessManager.syncOwnerLastContentUpdate(owner);
     }
 
+    private Owner createOwnerFromDTO(OwnerDTO ownerDTO) {
```
nikosmoum, line 2251: As discussed yesterday and based on the proposed design, this method should be in the service layer (OwnerManager) as you had it before, with the difference that it will be accepting an Owner instead of an OwnerDTO

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -2317,4 +2248,73 @@ public void remove(String ownerKey, String contentId) {
...
+        // Verify that we have an owner key (as required)
+        if (StringUtils.isBlank(ownerDTO.getKey())) {
+            throw new BadRequestException(i18n.tr("Owners must be created with a valid key"));
+        }
```
nikosmoum, line 2255: This should stay in the resource layer (so in `OwnerResource.createOwner()`), along with the reverse-translations call(`populateEntity`), which creates the Owner object, which is then passed into the service layer (`OwnerManager.createOwner()`)

nikosmoum (changes requested):

spec/refresh_pools_spec.rb
```diff
@@ -43,6 +43,15 @@
     end
   end
 
+  it 'contains the proper return value when owner is auto generated' do
+    result = @cp.refresh_pools("auto-generated-owner", false, true)
+    if !is_hosted?
+      expect(result).to be_nil
+    else
+      expect(result).to eq("Pools refreshed for owner: auto-generated-owner")
```
nikosmoum, line 51: let's also fetch the owner object itself to verify it's there, and assert that it's contentAccessMode is set to org_environment

src/main/java/org/candlepin/resource/OwnerResource.java
nikosmoum: It turns out doing this is has consequences that are hard to fix without a wider refactoring. That is the kind of task that needs to be done as a PoC. The current changes are good enough for sharing the owner creating code between the 2 endpoints.

src/main/java/org/candlepin/resource/OwnerResource.java
nikosmoum: Same here: see my other comment about doing a PoC about refactoring this

spec/refresh_pools_spec.rb
wolfdale: Done

nikosmoum (approved):

PR #3240, opened 2022-03-02
[M] 2059131: Reduce the facts accepted by the compliance status hasher

Januson:


src/main/java/org/candlepin/policy/js/compliance/hash/HashableStringGenerators.java
```diff
@@ -204,6 +205,15 @@ public String generate(Entry<String, String> target) {
      */
     private static class ConsumerGenerator implements HashableStringGenerator<Consumer> {
 
+        private static final Set<String> RELEVANT_FACTS = Set.of(
```
nikosmoum, line 208: Even though for now it's only the Hasher that cares about these, conceptually they are part of the business logic (they belong to the compliance policy). So instead of hiding them here, let's move move them into a proper Enum called something like `ComplianceStatusFacts` (similar to `CloudProfileFacts`), placed in org.candlepin.policy.js.compliance package.

src/main/java/org/candlepin/policy/js/compliance/hash/HashableStringGenerators.java
```diff
@@ -212,13 +222,21 @@ public String generate(Consumer target) {
...
 
-            generated += generateFromCollection(target.getFacts() != null ?
-                target.getFacts().entrySet() : null, STRING_ENTRY);
+            generated += generateFromCollection(factsOf(target), STRING_ENTRY);
             generated += generateFromCollection(target.getInstalledProducts(), INSTALLED_PRODUCT);
             generated += generateFromCollection(target.getEntitlements(), ENTITLEMENT);
             return generated;
         }
 
+        private Collection<Entry<String, String>> factsOf(Consumer target) {
```
nikosmoum, line 231: This method could be added directly onto the Enum

src/test/java/org/candlepin/policy/js/compliance/hash/SystemPurposeStatusHasherTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.xnap.commons.i18n.I18n;
+
+import java.util.HashSet;
+import java.util.Set;
+
+class SystemPurposeStatusHasherTest {
```
nikosmoum, line 37: :+1: for code coverage increase

nikosmoum (changes requested):

src/main/java/org/candlepin/policy/js/compliance/hash/HashableStringGenerators.java
Januson: Done.

src/main/java/org/candlepin/policy/js/compliance/hash/HashableStringGenerators.java
Januson: Done.

nikosmoum (approved):

PR #3244, opened 2022-03-02
Update NestedOwnerDTO to include Content Access Mode; ENT-4349

wottop:


src/main/java/org/candlepin/dto/api/v1/NestedOwnerDTOTranslator.java
```diff
@@ -53,6 +53,7 @@ public Owner populate(ModelTranslator translator, NestedOwnerDTO source, Owner d
         dest.setKey(source.getKey());
         dest.setDisplayName(source.getDisplayName());
         dest.setHref(source.getHref());
+        dest.setContentAccessMode(source.getContentAccessMode());
```
nikosmoum, line 56: No need for this here. This is a 'reverse' translator that translates NestedOwnerDTO coming as input from the API into an Owner. We only want to add this field to the output (and not accepting it in the input)

nikosmoum (changes requested):
one minor change, but otherwise all good

PR #3245, opened 2022-03-02
[M] Updated certificate generation and Tomcat configuration scripts (ENT-4447)

Ceiu:
- Updated gen_certs.sh to use FIPS-compatible certificate generation
  tools and options
- Updated update-server-xml.py in accordance with the changes made
  to gen_certs and current configuration standards
- Updated the deploy script in accordance with the changes made to
  gen_certs and update-server-xml.py
- Updated cpsetup to use same commands and configuration used in the
  updated gen_certs.sh and update-server-xml.py


Ceiu:
This version should address the issues with the self-signed cert.

Ceiu:
Getting the cert into the trust chain is proving to be a pain. Will finish fixing this tomorrow.

Ceiu:
This should be ready to go now. cpsetup won't work on el7/TC7 as we don't have a good way to detect various things (TC version or FIPS) without rewriting much of it.

nikosmoum (approved):

PR #3246, opened 2022-03-08
[M] Separate docker-compose responsibilities

nikosmoum:
- Add docker-compose option for only running the candlepin container
  when there is no need for a db (e.g. when running unit tests,
  checkstyle, translation validations), to avoid waisting resources.
- Move the requirement for a sonarqube certificate as docker secret
  to it's own docker-compose file, so that the default/postgres
  one can be run locally without it.

### DIFF
--- a/docker/README.mkd
+++ b/docker/README.mkd
@@ -31,7 +31,10 @@ and bring up a database at hostname `db` on the running container
 
 By default the `test` script will run candlepin against a postgres database.  You can change this behavior with the
 `-m` or `-p` flags which will use mysql or postgres respectively. To change the default permanently,
-run the `switch-defaults` script.
+run the `switch-defaults` script. Additionally, you can use the `-x` flag which will use no database at all (useful
+for running tasks that do not require a running server and database, such as unit tests, checkstyle, translation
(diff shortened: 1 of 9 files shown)


Ceiu:
Rather than have multiple compose files, is the Sonarqube stuff something that could be added during the docker-compose command itself? Not sure what level of support we need for actual docker, but the podman-backed docker-compose has options that can be passed in at each stage which might make the SQ bits optional without context-aware composition files.

nikosmoum:
I could not find anything on passing secrets to docker-compose on the command line.
As for podman it seems `podman run` has a --secret flag, but podman-compose doesn't. In any case, I would consider moving to podman or podman-compose as a separate, more involved task

Januson:
Using compose-override might work. We wouldn't need to duplicate the whole compose file, just extend it with sonar setup.

nikosmoum:
@Januson We could use it, but it wouldn't give us much since we'd still need to make the docker/test script apply or not apply the override file, depending on what we want to run. I think what @Ceiu is getting at is not using a separate file at all. If you both think using an override file works, then I can add it

PR #3248, opened 2022-03-08
Remove unnecessary SpecTest annotation

nikosmoum:


Ceiu:
The changes present look fine, however, the comment in the PR implies `status_resource_spec.rb` was removed as well, but this isn't showing up in the diff. Did this change not get committed yet?

nikosmoum:
@Ceiu That removal was in the initial commit of the PR, but then I reverted it because we don't actually have the java spec tests running in Jenkins (we should start deleting rspec tests after our java specs are running in jenkins). I'll edit out that PR comment

Ceiu:
Failure in Jenkins is a periodic test failure unrelated to the changes here. Merging.

PR #3256, opened 2022-03-11
[M] Fixed a potential collision with product entity versioning (ENT-4805)

Ceiu:
- Fixed a bug that could cause a versioning collision in some cases
- Added an additional exception when a version collision is detected
  during product and content management
- Added method chaining to mutators in ProductContent and Branding

### DIFF
--- a/src/main/java/org/candlepin/model/Product.java
+++ b/src/main/java/org/candlepin/model/Product.java
@@ -1300,6 +1300,7 @@ public class Product extends AbstractHibernateObject implements SharedEntity, Li
                 .append(this.attributes);
 
             if (this.derivedProduct != null) {
+                builder.append("derived_product");
                 builder.append(this.derivedProduct.getEntityVersion());
             }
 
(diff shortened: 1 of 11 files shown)


src/main/resources/db/changelog/20220314130000-clear_product_entity_versions.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
```
nikosmoum, line 7: We might want to make it a habit to start using the `4.8.xsd` version since if we don't then liquibase will complain if we try to use newer XML tag features than are available in 3.1 (see [here](https://docs.liquibase.com/concepts/changelogs/xml-format.html)). But this is just a note for future reference (not change needed here).

src/main/java/org/candlepin/model/Product.java
```diff
@@ -1308,42 +1309,46 @@ public int getEntityVersion() {
...
+            // sure to add our children entities into it in a consistent manner.
 
             if (!this.providedProducts.isEmpty()) {
-                accumulator = 0;
-                for (Product product : this.providedProducts) {
-                    accumulator += (product != null ? product.getEntityVersion() : 0);
-                }
+                builder.append("provided_products");
 
-                builder.append(accumulator);
```
nikosmoum, line 1319: I don't understand how this had an order dependency problem. If we had let's say 2 provided products, one with entity version 20 and another with entity version 30, the accumulator would always end up being 50, regardless of the order of those 2 providedProducts being processed?

nikosmoum (commented):
The code looks fine, just asking a question to understand the problem

src/main/java/org/candlepin/model/Product.java
Ceiu: The ordering thing was an issue the accumulator was intended to solve, and did, but also introduced two issues: if two different products had collections of children whose accumulated entity version or hashcode just happened to be the same, it would cause them to collide -- even if they occurred in different sets of children (i.e. product 1 has branding with an accumulated hashcode of A, and product 2 with content with an accumulated version of A, they will resolve to the same version).

New tests have been added to cover both cases, and may be a better way of understanding what I'm struggling to explain here.

nikosmoum: I understand now, thanks

PR #3257, opened 2022-03-14
[M] Update MySQL connector to use org.mariadb connector (ENT-4387)

Ceiu:
- Update deploy script config generation and build.gradle to use
  MariaDB Java connector rather than the MySQL java connector

### DIFF
--- a/bin/deployment/bash_functions
+++ b/bin/deployment/bash_functions
@@ -177,14 +177,8 @@ init_mysql_jdbc() {
     select_jdbc_jar "/usr/lib/java/mariadb-java-client.jar" "/usr/share/java/mysql-connector-java.jar"
     evalrc $? "No JDBC jar file available for MariaDB/MySQL"
 
-    JDBC_DRIVER="com.mysql.cj.jdbc.Driver"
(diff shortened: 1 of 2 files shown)


Ceiu:
This updates the scripts to use the mariadb connector, but does not touch the actual versions of the inclusions. A second update/PR should be posted with those.

nikosmoum:
Apparently it's an issue in mariadb driver itself, which was attempted to be fixed, resulting in a different issue, waiting to be fixed: https://github.com/liquibase/liquibase/issues/2431

nikosmoum:
Using mariadb java client 2.7.5 works (the bug seems to be introduced only in 3.x)

nikosmoum:
Also a different question is, why do we rely on the version available in EL7/8? We're already shipping it in the war, so we should be able to re-use it in our scripts, in the same way we do for liquibase. That way we'll always be using mariadb, and we can get rid of mysql-connector-java from our build.gradle (this should be included in the task we already have in the backlog probably)

nikosmoum (approved):

PR #3262, opened 2022-03-18
[M] Upgrade artemis to 2.20.0

nikosmoum:
- Update the external artemis deployment script to work on python3,
  and to use the latest artemis version.


Ceiu (approved):
Seems fine. I suspect that hard-coded "localhost" is going to be a problem in the FIPS world, but we'll see if it still matters by then.

PR #3265, opened 2022-03-18
[M] Added support for the Java spec tests to cp-test (ENT-4620)

Ceiu:
- The cp-test script has been updated to support running the
  Java spec tests along with rspec when the spec tests are
  invoked

### DIFF
--- a/docker/candlepin-base/cp-test
+++ b/docker/candlepin-base/cp-test
@@ -110,6 +110,18 @@ build_rspec() {
     fi
 }
 
+build_spec() {
+    # TODO: Add support for filtering spec tests once it's supported
+
+    # Check that both gradlew exists and the task exists
+    if ! test -f ./gradlew || ! ./gradlew tasks | grep -q "^spec\b"; then
+        echo "Warning: Java spec tests not found"
+        return;
+    fi
+
+    ./gradlew spec
+}
+
 trapex() {
     target="$1"
     shift
@@ -139,6 +151,7 @@ collect_artifacts() {
         lint.log
         buildr.log
         unit_tests.log
+        spec.log
         rspec.log
         translation.log
         sonarqube_upload.log
@@ -220,7 +233,7 @@ OPTIONS:
   -d                deploy a live candlepin
   -t                populate Candlepin database with test data (implies -d)
   -R                populate Candlepin environment with test repositories (implies -t)
-  -r [filter]       run rspec test suite (implies -d); may be filtered by test
+  -r [filter]       run spec test suite (implies -d); may be filtered by test
                     suite and name
   -H                run rspec tests in "hosted" mode (implies -r and -d)
   -u                run unit test suite N number of times, where N is number
@@ -257,7 +270,7 @@ while getopts ":dtRqrHulskib:c:a:vj:n" opt; do
             REPODATA="1"
             ;;
         r  )
-            RSPEC="1"
+            SPEC_TESTS="1"
             DEPLOY="1"
 
             ARG="${ARGV[$OPTIND - 1]}"
@@ -268,7 +281,7 @@ while getopts ":dtRqrHulskib:c:a:vj:n" opt; do
             ;;
         H  )
             HOSTED="1"
-            RSPEC="1"
+            SPEC_TESTS="1"
             DEPLOY="1"
             ;;
         u  ) UNITTEST=$((UNITTEST + 1))
@@ -290,7 +303,7 @@ while getopts ":dtRqrHulskib:c:a:vj:n" opt; do
         v  ) VERBOSE="1";;
         j  ) JAVA_VERSION="${OPTARG}";;
         i  ) TRANSLATE="1";;
-        k  ) CLOUD_AUTH_RSPEC="1";;
+        k  ) CLOUD_AUTH_SPEC="1";;
         n  ) SONARQUBE="1";;
         ?  ) usage; exit;;
     esac
@@ -466,10 +479,10 @@ if [ "$DEPLOY" == "1" ]; then
 
     DEPLOY_FLAGS="-g"
 
-    if [ "$RSPEC" == "1" ] && [ "$HOSTED" == "1" ] \
+    if [ "$SPEC_TESTS" == "1" ] && [ "$HOSTED" == "1" ] \
         && [ "$CP_NEW_STRUCTURE" == false ] && ($PROJECT_DIR/bin/deploy '-?' | grep -q -- '-H'); then
         DEPLOY_FLAGS="$DEPLOY_FLAGS -H -a"
-    elif [ "$RSPEC" == "1" ] && [ "$HOSTED" == "1" ] \
+    elif [ "$SPEC_TESTS" == "1" ] && [ "$HOSTED" == "1" ] \
         && [ "$CP_NEW_STRUCTURE" == true ] && ($PROJECT_DIR/bin/deployment/deploy '-?' | grep -q -- '-H'); then
         DEPLOY_FLAGS="$DEPLOY_FLAGS -H -a"
     else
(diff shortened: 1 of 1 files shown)


Ceiu:
Due to our design and intent to replace rspec tests as we port them over to Java rather than treating them as a separate, net-new set of tests, I've opted to fire both spec tests whenever running PRs rather than having a separate set of tasks for each DB combo (again).

docker/candlepin-base/cp-test
```diff
@@ -110,6 +110,17 @@ build_rspec() {
     fi
 }
 
+build_spec() {
+    # TODO: Add support for filtering spec tests once it's supported
+
+    # Check that both gradlew exists and the task exists
+    if ! test -f ./gradlew || ! ./gradlew tasks | grep -q "^spec\b"; then
```
nikosmoum, line 117: There's no need to check that gradlew exists any more; any branches that don't have gradle are no longer supported

docker/candlepin-base/cp-test
```diff
@@ -520,12 +531,15 @@ if [ "$DEPLOY" == "1" ]; then
...
 fi
 
-if [ "$RSPEC" == "1" ]; then
-    echo "Running rspec tests..."
+if [ "$SPEC_TESTS" == "1" ]; then
+    echo "Running spec tests..."
     CLEAN_CP=1
 
+    tee /var/log/candlepin/spec.log < /tmp/teepipe &
+    build_spec $SPEC_FILTER
```
nikosmoum, line 539: don't we need to redirect to /tmp/teepipe here?

nikosmoum (changes requested):
Just a couple minor comments, but the major issue is the tests aren't actually running in jenkins yet. Quay says the last time the candlepin-base image was updated was 14th March, so you need to push the image there for Jenkins to use it in the PR

docker/candlepin-base/cp-test
Ceiu: The gradlew check is mostly defensive programming and ensuring that we're in the right directory for the reference (since older builds which do support it may have us in the server directory).

Unless you see it hurting something, I'd feel better leaving it there.

Ceiu:
@nikosmoum Correct, I have not yet pushed an image to Quay since I wasn't sure if my reinterpretation of the task would be okay. If we're alright with the concept, I'll push an image up and see if it runs properly.

nikosmoum:
It looks good to me; it's better to avoid the overhead of spinning new jenkins runners and containers, so when you have the chance, push the image & re-run the build

Ceiu:
The task runs both spec tests; just waiting now for confirmation on artifact retention:
```
Task :spec-tests:test

DefaultPropertiesTest > canReadDefault() PASSED

OwnerApiTest > shouldCreateScaOwner() PASSED

OwnerApiTest > shouldCreateOwner() PASSED

OwnerApiTest > failsToCreateOwnerWithInvalidParent() PASSED

StatusApiTest > retrievesServerStatus() PASSED

--------------------------------------------------------------------
|  Results: SUCCESS (5 tests, 5 successes, 0 failures, 0 skipped)  |
--------------------------------------------------------------------

> Task :autoLintGradle
> Task :buildSrc:compileJava UP-TO-DATE

...

> Task :buildSrc:check UP-TO-DATE
> Task :buildSrc:build UP-TO-DATE
> Task :rspec

Activation Keys
  should allow activation key field filtering
...
```

docker/candlepin-base/cp-test
nikosmoum: You're right. This will run in older branches too, so if the java spec isn't there, we should return but not fail

nikosmoum:
I don't see the collection of the spec.log file. It's not in the main console output like the rest, e.g.:
```
Collecting artifacts...
‘/var/log/candlepin/access.log’ -> ‘/candlepin-dev/artifacts/access.log’
‘/var/log/candlepin/audit.log’ -> ‘/candlepin-dev/artifacts/audit.log’
‘/var/log/candlepin/candlepin.log’ -> ‘/candlepin-dev/artifacts/candlepin.log’
‘/var/log/candlepin/error.log’ -> ‘/candlepin-dev/artifacts/error.log’
‘/var/log/candlepin/rspec.log’ -> ‘/candlepin-dev/artifacts/rspec.log’
‘/var/log/tomcat/catalina.out’ -> ‘/candlepin-dev/artifacts/catalina.out’
‘/var/log/tomcat/catalina.2022-03-23.log’ -> ‘/candlepin-dev/artifacts/catalina.2022-03-23.log’
‘/var/log/tomcat/localhost.2022-03-23.log’ -> ‘/candlepin-dev/artifacts/localhost.2022-03-23.log’
‘/var/log/tomcat/manager.2022-03-23.log’ -> ‘/candlepin-dev/artifacts/manager.2022-03-23.log’
‘/var/log/tomcat/host-manager.2022-03-23.log’ -> ‘/candlepin-dev/artifacts/host-manager.2022-03-23.log’
```
but I also checked the node itself and it's not there

nikosmoum:
I think you'll need to re-build and push the container image for the new changes to take effect

docker/candlepin-base/cp-test
Ceiu: Fixed

Ceiu:
Comments addressed and new images pushed to quay. The comment about redirecting to the teepipe was why the new spec file wasn't getting written, so that should be good now as well.

docker/candlepin-base/cp-test
nikosmoum: In the latest run, java spec tests aren't running, and the `Warning: Java spec tests not found` is shown.

Ceiu: ARRRRRRRGH, my test code to verify what happens when the task doesn't exist got left in. Fixed.

docker/candlepin-base/cp-test
nikosmoum: The spec.log still isn't collected. Looking at the `collect_artifacts()` method, you need to add spec.log in the list of `CANDLEPIN_LOGS` to collect. Also, another thing which I don't know if it makes a difference, but I see it used elsewhere here, is to remove and setup the teepipe again when you want to use it more than once (between spec and rspec):
``` 
rm -f /tmp/teepipe
mkfifo /tmp/teepipe
```

Ceiu: Added spec.log to CANDLEPIN_LOGS.

Regarding the pipe, we're safe to reuse it. Just to be certain, I tried it manually to verify:
```
(08:26:51) crog@cl-vm-1:~/devel/candlepin-alt (crog/4.0/64bit_entity_versions) $ mkfifo /tmp/teepipe
(08:28:49) crog@cl-vm-1:~/devel/candlepin-alt (crog/4.0/64bit_entity_versions) $ tee a.txt < /tmp/teepipe &
[1] 187052
(08:29:07) crog@cl-vm-1:~/devel/candlepin-alt (crog/4.0/64bit_entity_versions) $ echo "hello" > /tmp/teepipe
hello
[1]+  Done                    tee a.txt < /tmp/teepipe
(08:29:15) crog@cl-vm-1:~/devel/candlepin-alt (crog/4.0/64bit_entity_versions) $ tee b.txt < /tmp/teepipe &
[1] 187064
(08:29:23) crog@cl-vm-1:~/devel/candlepin-alt (crog/4.0/64bit_entity_versions) $ echo "hello again" > /tmp/teepipe
hello again
[1]+  Done                    tee b.txt < /tmp/teepipe
(08:29:30) crog@cl-vm-1:~/devel/candlepin-alt (crog/4.0/64bit_entity_versions) $ cat a.txt
hello
(08:29:34) crog@cl-vm-1:~/devel/candlepin-alt (crog/4.0/64bit_entity_versions) $ cat b.txt
hello again
```

PR #3266, opened 2022-03-18
[M] Changed entity versions to a 64bit value; ENT-4831

Ceiu:
- Entity versions on Product, ProductContent, and Content are now
  64 bits instead of 32bits
- Rewrote the version calculation *again*, ensuring that even
  null and empty collections are represented in the final version
  hash
- The initial value and multipliers have been made unique between
  all versioned entities

### DIFF
--- a/src/main/java/org/candlepin/model/Product.java
+++ b/src/main/java/org/candlepin/model/Product.java
@@ -17,6 +17,7 @@ package org.candlepin.model;
 import org.candlepin.model.dto.ProductData;
 import org.candlepin.service.model.ProductInfo;
 import org.candlepin.util.ListView;
+import org.candlepin.util.LongHashCodeBuilder;
 import org.candlepin.util.MapView;
 import org.candlepin.util.SetView;
 import org.candlepin.util.Util;
@@ -51,6 +52,7 @@ import java.util.Set;
 import java.util.Stack;
 import java.util.StringTokenizer;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.persistence.CollectionTable;
 import javax.persistence.Column;
@@ -237,7 +239,7 @@ public class Product extends AbstractHibernateObject implements SharedEntity, Li
     private Set<String> dependentProductIds;
 
     @Column(name = "entity_version")
-    private Integer entityVersion;
+    private Long entityVersion;
 
     @Column
     @Type(type = "org.hibernate.type.NumericBooleanType")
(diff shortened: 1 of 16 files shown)


Ceiu:
I'm planning to add another test here if it doesn't end up taking too long to run, and I should probably add tests for the LongHashCodeBuilder.

src/main/resources/db/changelog/changelog-create.xml
```diff
@@ -1259,5 +1259,5 @@
     <include file="db/changelog/20211110083000-disallow_duplicate_versioned_entities.xml"/>
     <include file="db/changelog/20211123093750-migrate-environment-column.xml"/>
     <include file="db/changelog/20220117152105-add_system_lock_table.xml"/>
-    <include file="db/changelog/20220314130000-clear_product_entity_versions.xml"/>
+    <include file="db/changelog/20220314130000-64bit_entity_versions"/>
```
wottop, line 1262: .xml needed

src/main/resources/db/changelog/changelog-testing.xml
```diff
@@ -2351,5 +2351,5 @@
     <include file="db/changelog/20211110083000-disallow_duplicate_versioned_entities.xml"/>
     <include file="db/changelog/20211123093750-migrate-environment-column.xml"/>
     <include file="db/changelog/20220117152105-add_system_lock_table.xml"/>
-    <include file="db/changelog/20220314130000-clear_product_entity_versions.xml"/>
+    <include file="db/changelog/20220314130000-64bit_entity_versions"/>
```
wottop, line 2354: .xml

wottop (changes requested):

wottop (approved):

src/test/java/org/candlepin/model/ProductTest.java
```diff
@@ -406,9 +406,9 @@ public void testSetProvidedProductsChecksForCyclesOnProvidedProducts(int depth)
         Product p3 = spy(new Product())
             .setId("p3");
 
-        doReturn(1).when(p1).getEntityVersion();
```
Januson, line 409: There's no need to mock these. In fact we don't even need to spy on them anymore.

src/main/java/org/candlepin/util/LongHashCodeBuilder.java
```diff
@@ -0,0 +1,449 @@
...
+/**
+ * A version/hash builder based on the design and principles of the Apache HashCodeBuilder (which
+ * itself is based on principles outlined in Joshua Bloch's Effective Java). However, unlike the
+ * HashCodeBuilder, this class provides no reflection or object-introspection functionality.
+ *
+ * The functionality is very similar, if not identical to the aforementioned HashCodeBuilder, except
+ * the output is a 64-bit long instead of a 32-bit integer. This should, in theory, cut down on the
+ * number of collisions.
+ */
+public class LongHashCodeBuilder {
```
Januson, line 28: Please add some tests.

src/main/java/org/candlepin/model/Product.java
```diff
@@ -1340,8 +1350,12 @@ public int getEntityVersion() {
                     .sorted()
                     .forEach(builder::append);
             }
+            else {
+                builder.append((Object) null);
+            }
 
-            if (!this.branding.isEmpty()) {
+            Collection<Branding> branding = this.getBranding();
+            if (branding != null && !branding.isEmpty()) {
```
Januson, line 1358: It would be nice if the builder could handle these cases for us. There is a lot of repetition that could be just one of two methods on the builder.

Januson (changes requested):
The changes are looking good. I have just a few suggestions and builder could use some tests.

src/test/java/org/candlepin/model/ProductTest.java
Ceiu: I don't understand. We're ensuring the output of getEntityVersion is fixed so that the scenario where we have sets of children that have the same sum don't cause a collision.

I opted for `spy` here instead of `mock` so I didn't have to worry about mocking or breaking anything else that is used in the mapping when it's added as a child of another product.

src/main/java/org/candlepin/model/Product.java
Ceiu: I had considered it, and still might, but with the ordering bits in the middle throwing a wrench in it, I'm not sure how I want the interface to look for it. And given it's only used in one place today, it wasn't a priority of mine. 

If I can come up with something that won't look wonky or be highly specific to what's already in place, I'll add something for this.

src/test/java/org/candlepin/model/ProductTest.java
Ceiu: Added some comments around this to clarify the purpose of the mocking of `.getEntityVersion()`

src/main/java/org/candlepin/util/LongHashCodeBuilder.java
Ceiu: ack, still working on this one

Ceiu: Done. Tests added

src/main/java/org/candlepin/model/Product.java
Ceiu: I tried adding something here, but nothing worked that removed enough from the `Product.getEntityVersion` method. In a generic method, we can't assume the caller wants it sorted or filtered, and we don't know the mapping. So all we end up doing is abstracting out the lines that would add the prefix gate and the `.forEach` line. The rest of the logic ends up remaining in product. But I have reworked it to be more flow-through and consistent.

Januson (approved):

PR #3271, opened 2022-03-24
[M] Removed the matches parameter on entitlement APIs (ENT-4325)

JoshAlbrecht23:
- Removed on consumers and owners APIs

### DIFF
--- a/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -2469,7 +2469,7 @@ public class ConsumerResource implements ConsumersApi {
 
     @Override
     public List<EntitlementDTO> listEntitlements(@Verify(Consumer.class) String consumerUuid,
-        String productId, Boolean regen, String matches, List<KeyValueParamDTO> attrFilters) {
+        String productId, Boolean regen, List<KeyValueParamDTO> attrFilters) {
 
         Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
         PageRequest pageRequest = ResteasyContext.getContextData(PageRequest.class);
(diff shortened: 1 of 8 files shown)


src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2469,15 +2469,15 @@ private Entitlement verifyAndLookupEntitlement(String entitlementId) {
...
+        String productId, Boolean regen, List<KeyValueParamDTO> attrFilters) {
 
         Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
         PageRequest pageRequest = ResteasyContext.getContextData(PageRequest.class);
         if (regen) {
             revokeOnGuestMigration(consumer);
         }
 
-        EntitlementFilterBuilder filters = EntitlementFinderUtil.createFilter(matches, attrFilters);
+        EntitlementFilterBuilder filters = EntitlementFinderUtil.createFilter(null, attrFilters);
```
Ceiu, line 2480: Changing these to nulls is fine for this specific task, but somewhat highlights a shortcoming in our APIs here. Create a new task on our backlog to clean up the FilterBuilder hierarchy to have chainable mutators so we can eliminate the need to have a static factory function doing our construction of them, and can just build them directly as we need them.

At the absolute least, there should an overload of the `createFilter` function that doesn't require both parameters to be present. If we don't call this from any other place, we could just update the existing function signature to not have a matches parameter. In either case, work for a separate task.

Ceiu (approved):

Ceiu (changes requested):
Looks like the spec tests need to be updated or removed as appropriate.

src/main/java/org/candlepin/resource/ConsumerResource.java
JoshAlbrecht23: That sounds good to me. I had initially created an overloaded method for the createFilter function, but I wasn't sure if it was in scope of the Jira, so I removed it.

Logged Jira:
https://issues.redhat.com/browse/ENT-4848

nikosmoum (approved):

Ceiu (approved):

PR #3272, opened 2022-03-24
2068274: Product locking for content add/remove

wottop:


wottop:
Needs some mariadb investigation

src/main/resources/db/changelog/20220329154201-owner-product-key-change.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20220329154201-1" author="wpoteat">
```
Ceiu, line 9: Each logical step of this should be done as a separate change set to make error handling and reverts much easier. It also clarifies which pieces are DB-specific with specific changesets there. PostgreSQL, for instance, has no need to do the shenanigans with the primary/foreign keys, and shouldn't be subject to any goofy rules MySQL has.

src/main/resources/db/changelog/20220329154201-owner-product-key-change.xml
```diff
@@ -0,0 +1,53 @@
...
+
+        <!-- Now that we've added the primary key back, we can re-add the foreign keys -->
+        <addForeignKeyConstraint constraintName="cp2_owner_products_fk1"
+                                 baseTableName="cp2_owner_products"
+                                 baseColumnNames="owner_id"
+                                 referencedTableName="cp_owner"
+                                 referencedColumnNames="id"
+        />
+
+        <addForeignKeyConstraint constraintName="cp2_owner_products_fk2"
```
Ceiu, line 44: Use normal indentation rules for the continuation here.

src/main/resources/db/changelog/20220329154201-owner-product-key-change.xml
```diff
@@ -0,0 +1,53 @@
...
+            WHERE op.product_uuid = prod.uuid)
+        </sql>
+
+        <!-- On MySQL/MariaDB, we need to drop *all* foreign keys before we can drop our primary key -->
+        <dropAllForeignKeyConstraints baseTableName="cp2_owner_products"/>
+
+        <dropPrimaryKey tableName="cp2_owner_products" constraintName="cp2_owner_products_pk"/>
+
+        <addPrimaryKey tableName="cp2_owner_products" columnNames="owner_id, product_id"
+                   constraintName="cp2_owner_products_pk"/>
```
Ceiu, line 34: Extra indentation here

src/main/java/org/candlepin/model/OwnerProduct.java
```diff
@@ -137,6 +140,7 @@ protected void applyObjectIds() {
         }
 
         this.ownerId = owner.getId();
+        this.productId = product.getId();
```
Ceiu, line 143: We should add a check above to ensure the product ID is populated before we persist.

src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -282,16 +318,20 @@ public long getOwnerCount(Product product) {
...
+            this.getEntityManager()
+                .createQuery(jpql)
+                .setParameter("owner_id", owner.getId())
+                .setParameter("product_id", productId)
+                .getSingleResult();
+        }
+        catch (NoResultException nre) {
+            return false;
+        }
+        return true;
```
Ceiu, line 334: Let's rearrange the return statements so that our default output is false, and the confirmation only happens from within the try block. Largely meaningless, but helps readability for any refactors in the future and keeps it consistent with other, similar code.

Ceiu (changes requested):

JoshAlbrecht23 (approved):

src/main/resources/db/changelog/20220329154201-owner-product-key-change.xml
```diff
@@ -0,0 +1,52 @@
...
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20220329154201-1" author="wpoteat">
+        <addColumn tableName="cp2_owner_products">
+            <column name="product_id" type="VARCHAR(255)"/>
```
Ceiu, line 11: Product ID is a 32 character column, we should make this line up

src/main/resources/db/changelog/20220329154201-owner-product-key-change.xml
```diff
@@ -0,0 +1,52 @@
...
+            SET product_id = prod.product_id
+                FROM cp2_products prod
+                WHERE op.product_uuid = prod.uuid
+        </sql>
+    </changeSet>
+
+    <changeSet id="20220329154201-3" author="wpoteat"  dbms="mysql, mariadb">
+        <sql>
+            UPDATE cp2_owner_products op
+            SET product_id=(SELECT prod.product_id
```
Ceiu, line 27: We should use the UPDATE-JOIN syntax here rather than a subquery:
```
UPDATE cp2_owner_products op JOIN cp2_products p ON p.uuid = op.product_uuid SET op.product_id = p.product_id
```

Ceiu (changes requested):

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1995,13 +1997,14 @@ public ProductDTO addContent(String ownerKey, String productId, String contentId
...
-        // Get the matching owner_product & lock it while we are doing the update for this org
+        // Lock the owner_product while we are doing the update for this org
         // This is done in order to prevent collisions in updates on different parts of the product
-        OwnerProduct ownerProduct = ownerProductCurator.getOwnerProductByProductId(owner, productId);
-        ownerProductCurator.lock(ownerProduct);
-        ownerProductCurator.refresh(ownerProduct);
-
-        Product product = ownerProduct.getProduct();
+        if (!this.ownerProductCurator.lockOwnerProduct(owner, productId, LockModeType.PESSIMISTIC_WRITE)) {
+            throw new NotFoundException(
```
Ceiu, line 2003: This statement can fit on a single line

Ceiu (changes requested):

src/main/java/org/candlepin/resource/OwnerResource.java
Ceiu: I didn't repeatedly tag each instance, but all of them could be addressed the same way.

Normally I wouldn't hold up a PR for this, but since we need to re-run tests anyway to see if the SonarQube one will play nice, this minor correction could serve to do that.

Ceiu (approved):

PR #3273, opened 2022-03-24
[M] Added support for skipping pool creation while processing test data (ENT-4849)

Ceiu:
- Marketing products tagged with the "skip_pools" attribute will
  now be skipped during subscription/pool creation


Ceiu:
For testers:
Test data should no longer create pools for awesome-os-basic-vdc or the two dev-sku-products.

jsefler:
QE DevSKU and DataCenter tests using the TESTDATA have been failing against candlepin-4, but have been passing against candlepin-3.x.  This pull request fixes the expected behavior for both "dev-sku" testing and data center skus that employ host-limited derived pool creation.

wottop:
ack

PR #3276, opened 2022-03-28
[M] Updated content override accept a long string of overrides (ENT-4139)

JoshAlbrecht23:
- Updated the values column in cp_content_override table to be text.

### DIFF
--- a/src/main/java/org/candlepin/model/ContentOverride.java
+++ b/src/main/java/org/candlepin/model/ContentOverride.java
@@ -48,6 +48,8 @@ public abstract class ContentOverride<T extends ContentOverride, P extends Abstr
 
     /** Name of the table backing this object in the database */
     public static final String DB_TABLE = "cp_content_override";
+    public static final int MAX_NAME_AND_LABEL_LENGTH = 255;
+    public static final int MAX_VALUE_LENGTH = 2048;
 
     @Id
     @GeneratedValue(generator = "system-uuid")
@@ -57,7 +59,7 @@ public abstract class ContentOverride<T extends ContentOverride, P extends Abstr
     private String id;
 
     @Column(name = "content_label")
-    @Size(max = 255)
+    @Size(max = MAX_NAME_AND_LABEL_LENGTH)
     @NotNull
     private String contentLabel;
 
@@ -65,7 +67,7 @@ public abstract class ContentOverride<T extends ContentOverride, P extends Abstr
     @NotNull
     private String name;
 
-    @Size(max = 255)
+    @Size(max = MAX_VALUE_LENGTH)
     @NotNull
     private String value;
 
--- a/src/main/java/org/candlepin/util/ContentOverrideValidator.java
+++ b/src/main/java/org/candlepin/util/ContentOverrideValidator.java
@@ -18,6 +18,7 @@ import org.candlepin.config.ConfigProperties;
 import org.candlepin.config.Configuration;
 import org.candlepin.dto.api.v1.ContentOverrideDTO;
 import org.candlepin.exceptions.BadRequestException;
+import org.candlepin.model.ContentOverride;
 
 import com.google.inject.Inject;
 
@@ -39,7 +40,6 @@ import java.util.Set;
  */
 public class ContentOverrideValidator {
 
-    public static final int MAX_VALUE_LENGTH = 255;
     public static final Set<String> DEFAULT_BLACKLIST = Collections.<String>unmodifiableSet(
         new HashSet<String>(Arrays.asList("", "name", "label")));
     public static final Set<String> HOSTED_BLACKLIST = Collections.<String>unmodifiableSet(
@@ -81,6 +81,7 @@ public class ContentOverrideValidator {
             Set<String> invalidLabels = new HashSet<>();
             Set<String> invalidProps = new HashSet<>();
             Set<String> invalidValues = new HashSet<>();
+            Set<String> invalidLengthValues = new HashSet<>();
 
             for (ContentOverrideDTO override : overrides) {
                 if  (override != null) {
@@ -88,23 +89,29 @@ public class ContentOverrideValidator {
                     String name = override.getName();
                     String value = override.getValue();
 
-                    if (label == null || label.length() == 0 || label.length() > MAX_VALUE_LENGTH) {
+                    if (label == null || label.length() == 0 ||
+                        label.length() > ContentOverride.MAX_NAME_AND_LABEL_LENGTH) {
                         invalidLabels.add(label != null ? label : "null");
                     }
 
                     if (name == null || this.blacklist.contains(name.toLowerCase()) ||
-                        name.length() > MAX_VALUE_LENGTH) {
+                        name.length() > ContentOverride.MAX_NAME_AND_LABEL_LENGTH) {
 
                         invalidProps.add(name != null ? name : "null");
                     }
 
-                    if (value == null || value.length() == 0 || value.length() > MAX_VALUE_LENGTH) {
+                    if (value == null || value.length() == 0) {
                         invalidValues.add(value != null ? value : "null");
+                    }
+
+                    if (value != null && value.length() > ContentOverride.MAX_VALUE_LENGTH) {
+                        invalidLengthValues.add(name != null ? name : "null");
                     }
                 }
             }
 
-            if (!invalidLabels.isEmpty() || !invalidProps.isEmpty() || !invalidValues.isEmpty()) {
+            if (!invalidLabels.isEmpty() || !invalidProps.isEmpty() || !invalidValues.isEmpty() ||
+                !invalidLengthValues.isEmpty()) {
                 StringBuilder builder = new StringBuilder();
 
                 if (!invalidLabels.isEmpty()) {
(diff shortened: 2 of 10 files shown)


src/main/java/org/candlepin/util/ContentOverrideValidator.java
```diff
@@ -98,7 +98,7 @@ public void validate(Collection<? extends ContentOverrideDTO> overrides) {
                         invalidProps.add(name != null ? name : "null");
                     }
 
-                    if (value == null || value.length() == 0 || value.length() > MAX_VALUE_LENGTH) {
+                    if (value == null || value.length() == 0) {
```
Ceiu, line 101: Even if we have a database column that has no explicit limit, we should not be accepting unlimited length strings. We should decide on a long, but still reasonable limit here regardless what we do with the column type.

My suggestion here would be 1024 characters. Maaaaaybe 2048 if we've got some customers out there that want to push that field to its extremes. This also has the knock on effect of guiding where the database type goes.

src/main/resources/db/changelog/20220328090000-accommodate_large_overrides.xml
```diff
@@ -0,0 +1,21 @@
...
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="202203288090000" author="JoshAlbrecht23">
+        <comment>
+            Accommodate large content overrides for consumers.
+        </comment>
+
+        <modifyDataType
+            columnName="value"
+            newDataType="clob"
```
Ceiu, line 16: For stuff like this, we have a set of data types in datatypes.xml for Liquibase. In the case of unlimited-length strings, we have `text.type` which would be apt here.

src/test/java/org/candlepin/util/ContentOverrideValidatorTest.java
```diff
@@ -317,25 +318,21 @@ public void testValidateWithEmptyOverrideValue() {
...
-        List<ContentOverrideDTO> overrides = this.buildOverridesList(3);
-
-        // We expect this invocation to pass
-        this.validator.validate(overrides);
-
-        // Add our invalid override...
-        StringBuilder builder = new StringBuilder();
-        while (builder.length() < ContentOverrideValidator.MAX_VALUE_LENGTH) {
-            builder.append("longstring");
+        StringBuilder stringBuilder = new StringBuilder();
```
Ceiu, line 321: It's not strictly necessary to rename this variable to be so verbose. In this context we don't have any other builders to confuse it with, the declaration and lifecycle are limited to the immediate scope, and we're staying below our method length requirement, so finding the declaration from any of its uses should be trivial.

If this is purely a stylistic choice, then feel free to ignore this comment.

Ceiu (changes requested):

src/main/java/org/candlepin/util/ContentOverrideValidator.java
JoshAlbrecht23: I think that 1024 characters is a reasonable amount, and we can always easily increase it if there are some customers that need it increased.

src/test/java/org/candlepin/util/ContentOverrideValidatorTest.java
```diff
@@ -317,25 +318,21 @@ public void testValidateWithEmptyOverrideValue() {
...
+        while (stringBuilder.length() < 1000) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good call. Added a public static MAX_VALUE_LENGTH to the ContentOverrideValidator class where we can define and access the max length of the value field.

src/main/resources/db/changelog/20220328090000-accommodate_large_overrides.xml
JoshAlbrecht23: I see. Due to other changes, I changed the newDataType to varchar(1024)

src/main/java/org/candlepin/util/ContentOverrideValidator.java
```diff
@@ -130,6 +137,15 @@ public void validate(Collection<? extends ContentOverrideDTO> overrides) {
...
                 }
 
+                if (!invalidLengthValues.isEmpty()) {
+                    if (builder.length() > 0) {
+                        builder.append('\n');
+                    }
+
+                    builder.append(i18n.tr("The following override values exceed the max {0} " +
+                        "character length: {1}", MAX_VALUE_LENGTH, String.join(", ", invalidLengthValues)));
+                }
```
JoshAlbrecht23, line 147: I figured we could add some better output for exceedingly long values.

src/main/java/org/candlepin/util/ContentOverrideValidator.java
ptoscano: FYI: the case in point [1] has a length of 950 characters, so 1024 would be still too close (but higher, yes) to that limit; hence, I'd suggest to use at least 2048 or even 5000 characters (not sure whether there is a need for sticking to powers of 2).

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1969341

JoshAlbrecht23: I updated the character limit to 2048

Ceiu: > not sure whether there is a need for sticking to powers of 2

Nope, just faster math in some cases and nicer bitwise ops -- both things which don't matter here. 

It's important to note that as this column gets larger and larger, we're up against some real painful facts about it. Namely, as of today, we have some 226 million rows to worry about, and querying it is already incredibly painful. While it would be nice to use a text type here and not need to worry about updating it in the future, query performance would likely suffer. But, we can't just slap a huge number on a varchar type for the same reason. If we exceed the maximum row size, we're going to be up against the same perf problem TEXT has, as, at least in the case of MySQL/MariaDB, variable length columns will get chopped up and stored off-table.

That said, given what I'm seeing in the table right now, I would argue that we should not exceed ~4096 characters. It's well above the use case, and the sheer storage requirements for such absurd potential table sizes with such a limit is likely going to get us a talking to from our friends in IT.

src/main/java/org/candlepin/util/ContentOverrideValidator.java
Ceiu: This is going to output a list of *very* long values, to the point where the error message itself might not be valuable as written. It's probably better to output the key here rather than the value, or at least, truncate the value to something more consumable.

src/main/resources/db/changelog/20220328090000-accommodate_large_overrides.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="202203288090000" author="JoshAlbrecht23">
```
Ceiu, line 9: It's not super important in this case, but the `-#` suffix of the change set is used as an informal-ish ID of each change set in a given file. For the sake of consistency, we should add the missing `-1` on the change set ID here.

Ceiu (changes requested):

src/main/java/org/candlepin/model/ContentOverride.java
```diff
@@ -65,7 +65,7 @@
     @NotNull
     private String name;
 
-    @Size(max = 255)
+    @Size(max = 2048)
```
Ceiu, line 68: Since the table behind this class is shared with `ActivationKeyContentOverrides`, the annotation should also be updated in that class as well.

Ceiu (changes requested):

src/main/resources/db/changelog/20220328090000-accommodate_large_overrides.xml
JoshAlbrecht23: I see! I added the -1 to the id and updated the filename and updated the changelogs

Ceiu: Sorry, the file name was fine. It's just the changeset ID that needed updating.

JoshAlbrecht23: No worries! I wasn't sure if the changeSet id should match what is in the filename or not. I changed the filename back and updated the changelogs back

src/main/java/org/candlepin/util/ContentOverrideValidator.java
```diff
@@ -130,6 +137,16 @@ public void validate(Collection<? extends ContentOverrideDTO> overrides) {
...
                 }
 
+                if (!invalidLengthValues.isEmpty()) {
+                    if (builder.length() > 0) {
+                        builder.append('\n');
+                    }
+
+                    builder.append(i18n.tr("The following override value sizes exceed the max {0} " +
+                        "character length: {1}", MAX_VALUE_LENGTH,
+                        String.join(", ", Integer.toString(invalidLengthValues.size()))));
```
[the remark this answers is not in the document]
JoshAlbrecht23: No worries! I added the change to use the name

Ceiu (approved):

Ceiu:
@ptoscano Did you want a chance to review this?

ptoscano:
> @ptoscano Did you want a chance to review this?

Ah, seems OK to a non-candlepin eye. I leave this to your best judgement :)

The only minor thing I'd do is make the value in the `@Size` decorator not hardcoded but rather use the `MAX_VALUE_LENGTH` constant, so they don't get out of sync accidentally.

JoshAlbrecht23:
That's a good point! I agree, it makes more sense to have the size constraint definition live in the model class instead of having it defined in the validator class.

Ceiu (approved):

PR #3278, opened 2022-03-29
[M] Removed default-rules.js (ENT-1050)

JoshAlbrecht23:
Removing the deprecated default-rules.js and removing the method in the Exporter so that future manifests will not include the file.


nikosmoum (approved):
Looks good to me, but I'll have to wait with the merge because jenkins is still unavailable from the outage.

Januson (changes requested):
export_spec needs to be updated. There are some tests for the legacy rules.

JoshAlbrecht23:
@Januson good catch! I updated the export_spec

PR #3280, opened 2022-03-31
[M] Added consumer name to cp_deleted_consumer table (ENT-4570)

JoshAlbrecht23:
We are wanting to add the consumer name to the cp_deleted_consumer table so that you get back more information than the uuid of the consumer when you use the deleted_consumers endpoint.

I included a screen shot and the steps of the workflow I used to test:
1. I registered using subscription-manager, and checked cp_consumer table to get uuid and name of the consumer.
2. I unregistered and checked the cp_deleted_consumers table and verified that the correct consumer name was inserted.
3. I then verified the correct deleted consumers  information was returned from the deleted_consumers endpoint.

workflow

### DIFF
--- a/spec/deleted_consumer_spec.rb
+++ b/spec/deleted_consumer_spec.rb
@@ -18,7 +18,10 @@ describe 'Deleted Consumer Resource' do
     consumer1.unregister(consumer1.uuid)
 
     deleted_consumers = @cp.get_deleted_consumers(date=date)
-    deleted_consumers.map { |i| i['consumerUuid'] }.should include(uuid)
+    deleted_consumer = deleted_consumers.find { |item| item['consumerUuid'].include?(uuid) }
+    expect(deleted_consumer['consumerUuid']).to eq(uuid)
+    expect(deleted_consumer['consumerName']).to eq(consumername1)
+
   end
 
 end
--- a/src/main/java/org/candlepin/model/DeletedConsumer.java
+++ b/src/main/java/org/candlepin/model/DeletedConsumer.java
@@ -55,6 +55,11 @@ public class DeletedConsumer extends AbstractHibernateObject {
     @NotNull
     private String consumerUuid;
 
+    @Column(name = "consumer_name", length = 255)
+    @Size(max = 255)
+    @NotNull
+    private String consumerName;
+
     /**
      * using the id instead of actual Owner because the owner could be deleted
      * and we still want to keep this record around.
@@ -108,6 +113,15 @@ public class DeletedConsumer extends AbstractHibernateObject {
     public DeletedConsumer setOwnerId(String oid) {
         ownerId = oid;
         return this;
+    }
+
+    public DeletedConsumer setConsumerName(String consumerName) {
+        this.consumerName = consumerName;
+        return this;
+    }
+
+    public String getConsumerName() {
+        return consumerName;
     }
 
     public String getOwnerId() {
--- a/src/test/java/org/candlepin/model/DeletedConsumerTest.java
+++ b/src/test/java/org/candlepin/model/DeletedConsumerTest.java
@@ -15,42 +15,138 @@
 package org.candlepin.model;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
 
 import org.junit.jupiter.api.Test;
 
+import java.util.UUID;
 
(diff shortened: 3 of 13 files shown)


spec/deleted_consumer_spec.rb
```diff
@@ -19,6 +19,8 @@
 
     deleted_consumers = @cp.get_deleted_consumers(date=date)
     deleted_consumers.map { |i| i['consumerUuid'] }.should include(uuid)
+    deleted_consumers.map { |i| i['consumerName'] }.should include(consumername1)
```
Ceiu, line 22: Update this to use the `expect()` style of test rather than `.should`

src/test/java/org/candlepin/model/DeletedConsumerTest.java
```diff
@@ -16,41 +16,87 @@
...
+    private String consumerUuid = UUID.randomUUID().toString();
+    private String consumerName = "consumer-name";
+    private String ownerId = "owner-id";
+    private String ownerKey = "owner-key";
+    private String ownerDisplayName = "owner-display-name";
+    private String principalName = "principal-name";
+
+    @BeforeEach
+    private void beforeEach() {
+        deletedConsumer = new DeletedConsumer(consumerUuid, ownerId, ownerKey, ownerDisplayName);
```
Ceiu, line 38: Since we're updating this area, let's take the opportunity to detach from the injection-style constructor and use chained accessors instead so we can eventually remove those old constructors.

As a stylistic note: 
I'm not a huge fan of priming common test entities or objects in the initializers like this, as it makes the test suites fairly brittle and obscure as they get older and longer (see `ConsumerResourceTest` or `OwnerResourceTest` for an example of what this looks like extrapolated over a decade or more). It's fine for now, but will eventually become a maintenance headache.
What I like to see, instead, is a full test of the accessors and mutators from a base state. For example:

```
public void testGetSetConsumerUuid() {
  DeletedConsumer deletedConsumer = new DeletedConsumer();
  String testValue = "test_consumer_uuid";

  // Verify initial state is as we expect
  assertNull(deletedConsumer.getConsumerUuid());

  // Update the field, and ensure we're returning a self-ref
  DeletedConsumer out = deletedConsumer.setConsumerUuid(testValue);
  assertSame(out, deletedConsumer);

  // Verify final state is as expected/provided
  assertEquals(testValue, deletedConsumer.getConsumerUuid());
}
```

This can be broken up into separate tests if you'd prefer to be more strict about following unit test guidelines, but can get a little clunky as it's rather annoying to test the accessor without using the mutator or a purpose-built constructor (involves reflection and knowledge of the UUT's implementation, which is itself a no-no for good unit tests).

There's also an argument to be made that we should be testing mutators have no side effects on the other fields, but that requires much more infrastructure around both knowing and invoking the other accessors after every mutator. We've done this in the old DTO tests (check any of the 3.x branches for the API DTO tests for an example of this -- or maybe the manifest DTO tests in current master), but it's somewhat involved and clunky in its own right, and comes with its own baggage and headaches.

Ceiu (changes requested):

src/test/java/org/candlepin/model/DeletedConsumerTest.java
JoshAlbrecht23: This is great information! I like the example you gave of starting from a base state and then testing the accessors and mutators for that particular class member variable. I went ahead and implemented the rest of the tests in the same fashion and added a test for the constructors.

I can see how you could break up the accessors into their own tests without the mutators, but like you said, that would involve reflection to populate the private class fields and probably hard coding the field names and I think avoiding that is best.

src/main/java/org/candlepin/model/DeletedConsumer.java
```diff
@@ -75,8 +80,9 @@
     @Column(name = "principal_name")
     private String principalName;
 
-    public DeletedConsumer(String cuuid, String oid, String okey, String oname) {
+    public DeletedConsumer(String cuuid, String cname, String oid, String okey, String oname) {
```
Ceiu, line 83: Don't worry about updating these old constructors. This should either be left as-is, or removed entirely.

Ceiu (changes requested):
Just one last thing with the legacy entity constructor.

src/main/java/org/candlepin/model/DeletedConsumer.java
JoshAlbrecht23: Sounds good to me! Updated

Ceiu (approved):

PR #3281, opened 2022-03-31
[M] Updated several dependencies to their latest available versions (ENT-4808)

Ceiu:


JoshAlbrecht23 (approved):

nikosmoum (approved):
Change looks fine, but checkstyle has a weird error

nikosmoum (changes requested):
Actually just remembered we have the other build.gradle files to update too.
Shouldn't be too much of a problem (and if any is, we can create separate tasks)

nikosmoum:
ack

PR #3282, opened 2022-04-04
[M] Add openapi-generator 5.4.0 upgrade accompanying changes

nikosmoum:
- Merge changes from upstream v5.4.0 to our local copies of:
    * AbstractJavaJAXRSServerCodegen.java
    * api.mustache
    * pojo.mustache
    * Keep our custom workarounds introduced in d6a9b9e7f, bfa2bc88f
      and 733fc0b8a
- Remove our custom workaround from 25c2c0d48 because the fix is in
  upstream and use the new method convention 'put/add*Item()'
- Fix issue with ambiguous operationId name generation
- Update the client openapi config to use the JDK8 date/time API

### DIFF
--- a/buildSrc/src/main/java/org/openapitools/codegen/languages/AbstractJavaJAXRSServerCodegen.java
+++ b/buildSrc/src/main/java/org/openapitools/codegen/languages/AbstractJavaJAXRSServerCodegen.java
@@ -49,7 +49,7 @@ public abstract class AbstractJavaJAXRSServerCodegen extends AbstractJavaCodegen
     protected boolean useBeanValidation = true;
     protected boolean useTags = false;
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractJavaJAXRSServerCodegen.class);
+    private final Logger LOGGER = LoggerFactory.getLogger(AbstractJavaJAXRSServerCodegen.class);
 
     public AbstractJavaJAXRSServerCodegen() {
         super();
@@ -61,16 +61,16 @@ public abstract class AbstractJavaJAXRSServerCodegen extends AbstractJavaCodegen
         apiPackage = "org.openapitools.api";
         modelPackage = "org.openapitools.model";
 
-        // clioOptions default redifinition need to be updated
+        // clioOptions default redefinition need to be updated
         updateOption(CodegenConstants.INVOKER_PACKAGE, this.getInvokerPackage());
         updateOption(CodegenConstants.ARTIFACT_ID, this.getArtifactId());
         updateOption(CodegenConstants.API_PACKAGE, apiPackage);
         updateOption(CodegenConstants.MODEL_PACKAGE, modelPackage);
-        updateOption(this.DATE_LIBRARY, this.getDateLibrary());
+        updateOption(DATE_LIBRARY, this.getDateLibrary());
 
         additionalProperties.put("title", title);
         // java inflector uses the jackson lib
-        additionalProperties.put("jackson", "true");
+        additionalProperties.put(JACKSON, "true");
 
         cliOptions.add(new CliOption(CodegenConstants.IMPL_FOLDER, CodegenConstants.IMPL_FOLDER_DESC).defaultValue(implFolder));
         cliOptions.add(new CliOption("title", "a title describing the application").defaultValue(title));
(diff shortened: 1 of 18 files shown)


src/main/java/org/candlepin/resource/PoolResource.java
```diff
@@ -90,7 +90,7 @@ public PoolResource(ConsumerCurator consumerCurator, OwnerCurator ownerCurator,
     @Override
     @Deprecated
     @SecurityHole
-    public List<PoolDTO> list(String ownerId, String consumerUuid, String productId,
+    public List<PoolDTO> listPools(String ownerId, String consumerUuid, String productId,
```
Ceiu, line 93: :+1:

buildSrc/src/main/java/org/openapitools/codegen/languages/AbstractJavaJAXRSServerCodegen.java
```diff
@@ -49,7 +49,7 @@
     protected boolean useBeanValidation = true;
     protected boolean useTags = false;
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractJavaJAXRSServerCodegen.class);
+    private final Logger LOGGER = LoggerFactory.getLogger(AbstractJavaJAXRSServerCodegen.class);
```
Ceiu, line 52: We don't want this to be a const?

buildSrc/src/main/resources/templates/pojo.mustache
```diff
@@ -46,53 +48,54 @@ import com.fasterxml.jackson.annotation.JsonValue;
...
-  public void {{setter}}({{{datatypeWithEnum}}} {{name}}) {
+  @JsonProperty("{{baseName}}")
+  {{#vendorExtensions.x-setter-extra-annotation}}{{{vendorExtensions.x-setter-extra-annotation}}}
+  {{/vendorExtensions.x-setter-extra-annotation}}public void {{setter}}({{{datatypeWithEnum}}} {{name}}) {
     this.{{name}} = {{name}};
   }
 
   {{#isArray}}
-  public boolean add{{nameInCamelCase}}({{{items.datatypeWithEnum}}} {{name}}) {
+  public {{classname}} add{{nameInCamelCase}}Item({{{items.datatypeWithEnum}}} {{name}}Item) {
```
Ceiu, line 64: This change brings up an interesting question/debate: Generally with mutators that can do nothing (as is the case when adding elements to a set or map, or removing from a collection), we return a boolean so the caller can know if the operation had an effect on the object. 

I guess there's an argument to be made that if the operation is to be considered idempotent, then the boolean output has no real value, since the add will always ensure the element is there, or fail with an exception; with remove being the inverse -- ensuring the element doesn't exist in the set. This example kind of fails in the array/list implementation of remove, since it'd be possible for "remove" to succeed on repeated calls, without fully removing the target element. For instance:

```
list = ["a", "b", "a", "c", "a"];
list.remove("a").remove("a"); // an instance of "a" still exists in the list.
```

Whether or not this matters is one of usability and expectation, I suppose. I can assume a stance from this change, but what are your thoughts on this?

Ceiu (approved):
Looks good; some good naming cleanup here. I do have a few non-blocking questions below, though.

buildSrc/src/main/java/org/openapitools/codegen/languages/AbstractJavaJAXRSServerCodegen.java
nikosmoum: This isn't our code, it's just a local copy of the upstream version, so I think we should try to keep it as close to the original. Here I just merged the changes from the v5.4.0 version of the [file](https://github.com/OpenAPITools/openapi-generator/blob/5.4.x/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaJAXRSServerCodegen.java), while keeping our only [custom change.](https://github.com/candlepin/candlepin/commit/d6a9b9e7fa2b3fcc2bb5ca0cd99423d6f6e6e292#diff-222916468ed7d67decffd7c8eb373abbd37dbd7b6c338821cf1031f29468d927)

Ceiu: Ahh, alright

buildSrc/src/main/resources/templates/pojo.mustache
Ceiu: Gotcha

PR #3284, opened 2022-04-05
adding milestones to abort previous builds

Lorquas:
This PR uses the built in milestones plugin https://www.jenkins.io/doc/pipeline/steps/pipeline-milestone-step/  to stop previous builds.  Instead of manually tracking milestone steps as in the docs of this plugin we're setting them to the build number so that the latest milestone is always the latest job.  We are also setting the previous milestone in case it wasn't set.  (see https://issues.jenkins.io/browse/JENKINS-43353 ).   Some implementations of this do this in a for loop to ensure that all previous builds have a milestone set  (for example: https://github.com/RedHatInsights/insights-tag-service/blob/44abadea01427601e8962f35f099f870d4b2e754/Jenkinsfile#L27 ), however this does take time and it slows down the system for PRs with a large number of builds (hundreds?).  If we see any slowing down I can only do the last N of the loop or something.

Any reviewer can test this by going to the job for this PR in jenkins and clicking "build now" many times.  As they queue they will all be halted except the last one.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -4,13 +4,26 @@ library identifier: 'fh-pipeline-library@master', retriever: modernSCM(
    credentialsId: 'github-api-token-as-username-password'])
 
 pipeline {
-    agent { label 'candlepin' }
+    agent none
     options {
         skipDefaultCheckout true
         timeout(time: 16, unit: 'HOURS')
     }
     stages {
+        stage('Stop Old Builds') {
+            // purely a jenkins step; no env needed
+            agent none
+            steps {
+                script {
+                    // stops all previous running builds
+                    for (int i = 1; i <= (env.BUILD_NUMBER as int); i++) {
+                       milestone ordinal: i
+                    }
+                }
+            }
+        }
         stage('Trust') {
+            agent none
             steps {
                 enforceTrustedApproval("candlepin","rhsm-jenkins-github-app")
             }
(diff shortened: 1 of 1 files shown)


JoshAlbrecht23:
There was some discussion in the Candlepin chat on if we should have functionality like this in a centralized repo so that it can be consumed by other jenkins jobs as well, but that may be work for a separate task and for now we can put the functionality in Candlepin Jenkins file

JoshAlbrecht23:
In the last discussion regarding this PR, there was concern about having to update all of the Jenkins files to get this functionality. We wanted to investigate and see if it is possible to set a Jenkins configuration or something and have all of the jenkins jobs inherit that functionality without having to modify all the Jenkins files. This may require logging an investigation jira!

Lorquas:
OK so This PR now implements the disabledConcurrentBuild feature added to the pipelines (which has now been installed on jenkins), and skips the milestone hack.

I tried setting this in candlepin-jobs however this specifically doesn't work for multibranch pipelines.  These property classes aren't implemented for this job type, and the whole thing stack traces if I try to force write these configs.

This PR also now correctly sets the agents for each step which should make things run faster (It doesn't have to wait for a candlepin node for the trust step because it doesn't need one)

Because of the two notes above, I recommend - unfortunately - to push these changes to all of the branches we care about.

JoshAlbrecht23 (approved):

nikosmoum (approved):

PR #3293, opened 2022-04-12
[M] Added a check for null service level for consumer (ENT-3977)

JoshAlbrecht23:
Stack trace for the issue:
https://gist.github.com/JoshAlbrecht23/cef375638668ea41ac74c06e094d8884

### DIFF
--- a/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -1031,16 +1031,16 @@ public class ConsumerResource implements ConsumersApi {
             // activation keys have higher priority than those on consumer
             this.updateSystemPurposeData(consumer, consumerToCreate);
 
-            if (keys.size() > 0) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1043,6 +1039,10 @@ public Consumer createConsumerFromDTO(ConsumerDTO consumer, ConsumerType type, P
                     "");
             }
 
+            if (keys.size() > 0) {
+                consumerBindUtil.handleActivationKeys(consumerToCreate, keys, owner.isAutobindDisabled());
+            }
+
```
JoshAlbrecht23, line 1045: I figured it would be a good idea to handle the service level logic first and then go into the handleActivationKeys to help prevent issues with a null service level value on the Consumer object.

Ceiu: :+1:

Ceiu (approved):

PR #3296, opened 2022-04-12
[M] Added principal name for system generated jobs (ENT-4855)

JoshAlbrecht23:
Setting the principal name for system generated job instead of it just being null. I included some output from the cp_async_jobs table from testing the code.

```
candlepin=> Select id, name, job_key, principal from cp_async_jobs;
                id                |          name          |        job_key         |           principal           
----------------------------------+------------------------+------------------------+-------------------------------
 ff808181801e769a01801e80d0380b21 | ActiveEntitlementJob   | ActiveEntitlementJob   | System_ActiveEntitlementJob
 ff808181801e769a01801e80d04d0b22 | ExpiredPoolsCleanupJob | ExpiredPoolsCleanupJob | System_ExpiredPoolsCleanupJob
 ff808181801e769a01801e817aaa0b23 | ManifestCleanerJob     | ManifestCleanerJob     | admin
(3 rows)
```

### DIFF
--- a/src/main/java/org/candlepin/async/JobManager.java
+++ b/src/main/java/org/candlepin/async/JobManager.java
@@ -89,7 +89,7 @@ import javax.transaction.Synchronization;
 
 // FIXME: The transaction management here is getting out of hand. The annotations should probably
 // be removed entirely and replaced with explicit transaction management, and the protected methods
-// using them should be reverted to their propery visibility (private).
+// using them should be reverted to their proper visibility (private).
 
 
 /**
@@ -164,9 +164,11 @@ public class JobManager implements ModeChangeListener {
         @Override
         public void execute(org.quartz.JobExecutionContext context) throws org.quartz.JobExecutionException {
             String jobKey = context.getJobDetail().getKey().getName();
+            Principal principal = new JobPrincipal(SystemPrincipal.NAME + "." + jobKey);
+            ResteasyContext.pushContext(Principal.class, principal);
 
             try {
-                log.trace("Queuing job: {}", jobKey);
+                log.trace("Queuing scheduled job: {}", jobKey);
                 this.manager.queueJob(JobConfig.forJob(jobKey));
             }
             catch (JobException e) {
@@ -209,7 +211,7 @@ public class JobManager implements ModeChangeListener {
 
     /**
      * The JobMessageSynchronizer commits or rolls back a messenger transaction upon the completion
-     * of a database tranaction.
+     * of a database transaction.
      */
     private static class JobMessageSynchronizer implements Synchronization {
         /** An array of states that are valid to synchronize against */
--- a/src/main/java/org/candlepin/auth/Principal.java
+++ b/src/main/java/org/candlepin/auth/Principal.java
@@ -130,5 +130,4 @@ public abstract class Principal implements Serializable, java.security.Principal
             return "";
         }
     }
-
 }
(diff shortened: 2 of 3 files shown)


Ceiu (changes requested):
nack

Few things to note here:
- We don't want to couple the auth system to the job system by adding a name builder as a static function in the principal class. While it's technically generic enough that it doesn't have to be only applied to the job system, it being written for the job system in particular means work in one area will affect the other.
- While this accomplishes the goal, it's hitting everything that comes from something that doesn't have a principal -- not necessarily system generated jobs. The distinction won't matter much in practice, but it does affect what the "correct" implementation is in this case. If we're focused on purely scheduled jobs, then the scheduling bridge should probably be responsible for sorting out the principal and injecting it in the context as it schedules the job. If we want this to be the default for when we don't have a principal at all, there's already a place where we have handling for such a fallback. Injecting the handler during the job status conversion creates redundancy later where the principal is created while standing up the job environment. The global default could be moved there as a one-liner that gets rid of use of the SystemPrincipal almost entirely [1].
- Dot notation is a bit safer here, as underscores have been used in the job names in the past, and is a more common delimiter for denoting hierarchy traversal.

[1]
```
Principal principal = new JobPrincipal(name != null ? name : "System." + status.getJobKey());
```

Note that the magic string could be replaced by a constant declared somewhere, depending on level of effort desired, and what class should hold responsibility over its declaration. There are a lot of answers here, ranging from JobManager to SystemPrincipal, or even "no one."

JoshAlbrecht23:
@Ceiu Thank you for the detailed response! I had to do some digging into the JobManager functionality to understand functionality a little more to compare the different approaches you listed. To me, the one that made the most sense was to have the scheduling bridge handle the injection of the principal into the context as it schedules a job. I also think it makes sense to declare the "System" name as constant in the SystemPrincipal class.

src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -89,7 +89,7 @@
 
 // FIXME: The transaction management here is getting out of hand. The annotations should probably
 // be removed entirely and replaced with explicit transaction management, and the protected methods
-// using them should be reverted to their propery visibility (private).
+// using them should be reverted to their property visibility (private).
```
Ceiu, line 92: +1 on the spelling correction, but it was intended to be "proper" not "property"

src/main/java/org/candlepin/async/JobManager.java
```diff
@@ -164,6 +164,8 @@ public Job newJob(TriggerFiredBundle bundle, Scheduler scheduler) throws Schedul
         @Override
         public void execute(org.quartz.JobExecutionContext context) throws org.quartz.JobExecutionException {
             String jobKey = context.getJobDetail().getKey().getName();
+            Principal principal = new JobPrincipal(SystemPrincipal.NAME + "." + jobKey);
+            ResteasyContext.pushContext(Principal.class, principal);
 
             try {
                 log.trace("Queuing job: {}", jobKey);
```
Ceiu, line 171: I know this is unrelated to your work, and is incredibly minor, but can we use this opportunity to change the trace line to read "Queuing scheduled job: {}"?

Ceiu (changes requested):
Much better this time around. Just a minor correction to a correction, and one additional ask since we're touching the scheduling bridge.

src/main/java/org/candlepin/async/JobManager.java
JoshAlbrecht23: Oh haha yeah, that makes more sense!

src/main/java/org/candlepin/async/JobManager.java
JoshAlbrecht23: Sounds good to me! "Queuing scheduled job" makes more sense for this tracer

Ceiu (approved):

PR #3303, opened 2022-04-14
Update cache jars [ENT-4860]

wottop:
The issue we had with empty collections was because this was fixed in hibernate 5.5.3:
https://hibernate.atlassian.net/browse/HHH-4808

We were relying on the connection remaining open when the CandlepinQuery is passed to the interceptor.
The setting "hibernate.connection.release_mode" allows it to behave the way it did before the bug was fixed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -167,11 +167,11 @@ dependencies {
     implementation "org.hibernate.validator:hibernate-validator-annotation-processor:6.2.3.Final"
 
     // Hibernate
-    implementation 'org.hibernate:hibernate-c3p0:5.4.6.Final'
+    implementation 'org.hibernate:hibernate-c3p0:5.6.7.Final'
     // Ehcache (for use with hibernate primarily)
-    implementation "org.hibernate:hibernate-jcache:5.4.6.Final"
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -167,11 +167,11 @@ dependencies {
     implementation "org.hibernate.validator:hibernate-validator-annotation-processor:6.2.3.Final"
 
     // Hibernate
-    implementation 'org.hibernate:hibernate-c3p0:5.4.6.Final'
+    implementation 'org.hibernate:hibernate-c3p0:5.6.7.Final'
```
nikosmoum, line 170: Is the hibernate criteria removal what's stopping us from upgrading to 6.0.0? Is that happening in a separate card?

wottop: It seemed more prudent to go to the highest version 5 as there was an issue that was fixed in the PR.
The criteria removal will definitely be a larger change as well as my hesitancy to jump on an x.0.0 release.
My opinion is to make a separate card for a v6 jump and give it some time to settle in on the hibernate side.

nikosmoum: ack

PR #3307, opened 2022-04-18
[M] Removed mysql-connector-java from gradle dependencies. (ENT-4783)

JoshAlbrecht23:
We are also making a change to use the mariadb-java-client jar from the classpath so that we always use the same version that is defined by gradle. I was able to verify the changes be deploying candle pin in a Vagrant vm for el7, el8, and f35.

### DIFF
--- a/bin/deployment/bash_functions
+++ b/bin/deployment/bash_functions
@@ -174,7 +174,8 @@ init_mysql_jdbc() {
     local password="$DBPASSWORD"
     [ -z "$password" ] && password=''
 
-    select_jdbc_jar "/usr/lib/java/mariadb-java-client.jar" "/usr/share/java/mysql-connector-java.jar"
+    local maria_client_jar=$(find $CP_EXTRACT_CLASSPATH -type f -name "mariadb-java-client*.jar")
+    select_jdbc_jar $maria_client_jar
     evalrc $? "No JDBC jar file available for MariaDB/MySQL"
 
     JDBC_DRIVER="org.mariadb.jdbc.Driver"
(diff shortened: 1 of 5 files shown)


bin/deployment/bash_functions
```diff
@@ -174,7 +174,9 @@ init_mysql_jdbc() {
     local password="$DBPASSWORD"
     [ -z "$password" ] && password=''
 
-    select_jdbc_jar "/usr/lib/java/mariadb-java-client.jar" "/usr/share/java/mysql-connector-java.jar"
+    local cp_classpath="/var/lib/tomcat/webapps/candlepin/WEB-INF/lib"
+    local maria_client_jar=$(JARS=("$cp_classpath"/mariadb-java-client*.jar); IFS=:; read -r JAR <<< "$JARS"; echo "$JAR")
+    select_jdbc_jar $maria_client_jar "/usr/share/java/mysql-connector-java.jar"
```
nikosmoum, line 179: Let's remove the `/usr/share/..` from here. We no longer want to depend on mysql-connector-java, nor on any mariadb/mysql jars from any place in the filesystem other than our own webapp lib folder

nikosmoum (changes requested):

bin/deployment/bash_functions
JoshAlbrecht23: Removed the `/usr/share/java/mysql-connector-java.jar`

JoshAlbrecht23: Well, it looks like the initialization of the database is happening before the app is deployed to `/var/lib/tomcat/webapps/candlepin/WEB-INF/lib` and that is causing issues on the first that CP is deployed because the jars do not yet exist in that directory

JoshAlbrecht23: That makes sense! I added those changes, but while I was testing the error messaging for when the jar cannot be found, we saw that the error messages are being swallowed because 'set -e' is set in the deploy script and so functions returning a non 0 exit code ends the termination of the whole script. 

I'll log a jira to improve the script error handling so that we can have a better understanding of why the script is terminating.

PR #3308, opened 2022-04-18
[M] 2075587: Fixed attribute storage issue causing version collisions (ENT-4907)

Ceiu:
- Fixed a bug with null-versioned product attributes causing
  version collisions due to being used in the version calculation
  before being silently discarded during entity persistence
- Null-valued attributes are now silently discarded at the API
  layer to retain the current behavior
- Product entities will no longer accept null values for product
  attributes, as Hibernate throws them out anyway

### DIFF
--- a/src/main/java/org/candlepin/controller/ProductManager.java
+++ b/src/main/java/org/candlepin/controller/ProductManager.java
@@ -247,7 +247,7 @@ public class ProductManager {
         Product entity = new Product()
             .setId(productData.getId());
 
-        applyProductChanges(entity, productData, productMap, contentMap);
+        entity = applyProductChanges(entity, productData, productMap, contentMap);
 
         // Check if we have an alternate version we can use instead.
         List<Product> alternateVersions = this.ownerProductCurator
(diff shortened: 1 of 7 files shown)


Ceiu:
I've gone back and forth a couple times on this, and ultimately landed on having the API layer act as it does currently, as some of our users may be expecting this behavior out of null attributes. Changing it to be a no-op, as I had originally proposed, creates a weird set of interactions where a null-valued attribute is discarded during creation, discarded during update if the attribute isn't already there, and a no-op when it is. It's kind of the same as discarding it in all cases, with the null-as-no-change behavior only applies in a very specific scenario. And when that's considered with how the attribute mapping is only ever updated when the map itself is non-null, it's very unlikely we'll have users giving us an attribute map that has all of the attributes currently present with null values.

However, all that said, I'm willing to be convinced otherwise if someone feels strongly another way. I'm also slightly worried that this issue can occur in other such collections or mappings, like Content.modifiedProductIds, Pool.attributes, and Product.dependentProductIds.

src/main/java/org/candlepin/controller/ProductManager.java
```diff
@@ -790,6 +787,32 @@ private static Product applyProductChanges(Product entity, ProductInfo update,
...
+
+            incoming.forEach((k, v) -> {
+                if (v != null) {
+                    filtered.put(k, v);
+                }
+            });
+
+            entity.setAttributes(filtered);
+        }
+    }
```
JoshAlbrecht23, line 814: Would it make more sense to put this in the setAttributes method on the Product class? That would help prevent someone from accidentally using the setAttributes without using the applyProductAttributeChanges and creating entries with null values.

JoshAlbrecht23: I spoke too soon. You already have the change in the Product setAttributes method!

JoshAlbrecht23 (approved):

Ceiu:
Unit tests fixed.

Also, I had previously forgot to mention the test case which lead to this discovery and change:
1. Create an org -- no special attributes required. 
```$ ./client/ruby/cpc create_owner admin```
1. Create a product with a null-valued attribute
```curl -k -u admin:admin -X POST --data '{"id": "test_prod", "name": "test_prod", "attributes":[{"name":"dead_attribute"}]}' --header 'Content-Type:application/json' --header 'Accept:application/json' "https://localhost:8443/candlepin/owners/admin/products"```
1. Delete the product
```curl -k -u admin:admin -X DELETE "https://localhost:8443/candlepin/owners/admin/products/test_prod"
1. Attempt to create the product again, using the exact same command as above.

If the bug persists, the second creation will fail with an entity version collision. If the bug is fixed, the creation will succeed. In all cases, the dead attribute should not be present in the output of the product.

PR #3311, opened 2022-04-19
1962572: Filter list of available system purpose values

wottop:
Filter out the values for products which do not have live pools


Ceiu (changes requested):

src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -289,6 +290,52 @@ public CandlepinQuery<Product> getProductsByIdsUsingOwnerKey(String ownerKey,
...
+            .getResultList();
+
+        sql =
+            "SELECT DISTINCT a.name, a.value " +
+            "FROM cp2_product_attributes a JOIN cp_pool b ON a.product_uuid=b.product_uuid " +
+            "WHERE name IN ('support_level') " +
+            "AND b.owner_id=:owner_id " +
+            "AND b.endDate > CURRENT_TIMESTAMP " +
+            "AND a.product_uuid NOT IN " +
+            "(SELECT a.product_uuid FROM cp2_product_attributes a " +
```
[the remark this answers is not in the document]
wottop: Against stage it was not much faster and the highest number of products that any owner has is about 700.
I still made it a join instead of a subselect because I was already there.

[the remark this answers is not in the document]
Ceiu: What do the query plans look like between the two?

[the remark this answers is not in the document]
wottop: http://pastebin.test.redhat.com/1046481

Ceiu (changes requested):
Need some unit tests and documentation for the new curator methods, and a few other improvements.

src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -289,6 +290,68 @@ public CandlepinQuery<Product> getProductsByIdsUsingOwnerKey(String ownerKey,
...
+            "WHERE c.value IS NULL or c.value != 'true'";
+
+        result.addAll(this.getEntityManager()
+            .createNativeQuery(sql)
+            .setParameter("owner_id", ownerId)
+            .getResultList());
+
+        Map<String, Set<String>> attributeMap = new HashMap<>();
+        for (Object[] attMap : result) {
+            attributeMap.computeIfAbsent((String) attMap[0], key -> new LinkedHashSet<>())
```
Ceiu, line 350: I've been thinking about this for a bit, and I'm curious what the intent behind the linked hash set is here. Since we're not doing any sorting in the queries above, retaining the insertion order of the set that gets built here doesn't seem to be very useful -- the output order of the queries aren't guaranteed to be consistent between executions.

If this is something we need to retain, we'll need to add some kind of sorting or some other mechanism to ensure the output of the queries is consistent. Otherwise, we can just use a normal set here and not worry about the order.

src/test/java/org/candlepin/model/OwnerProductCuratorTest.java
```diff
@@ -696,4 +699,80 @@ public void testGetProductsByVersionsDoesntFailWithLargeDataSets() {
 
         this.ownerProductCurator.getProductsByVersions(versions);
     }
+
+    @Test
+    public void testGetSyspurposeAttributesByOwner() {
```
Ceiu, line 704: Let's also add tests for our base cases and invalid inputs

Ceiu (changes requested):

src/test/java/org/candlepin/resource/OwnerResourceTest.java
```diff
@@ -2005,24 +2006,21 @@ public void testReturnProductSysPurposeValuesForOwner() throws Exception {
...
-        p3.setAttribute(SystemPurposeAttributeType.USAGE.toString(), "development");
+        Set<String> mockAddons = new LinkedHashSet<>();
+        mockAddons.add("hello");
+        mockAddons.add("world");
+        mockAddons.add("earth");
+        mockMap.put(SystemPurposeAttributeType.ADDONS.toString(), mockAddons);
 
-        List<Product> dummyProducts = new ArrayList<>(Arrays.asList(p1, p2, p3));
-        when(mockQuery.list()).thenReturn(dummyProducts);
+        Set<String> mockUsage = new LinkedHashSet<>();
```
Ceiu, line 2019: Don't use a LinkedHashSet anymore here, or the equality check will fail if the LHS is the object doing the comparison and the order happens to be "wrong."

src/test/java/org/candlepin/model/OwnerProductCuratorTest.java
```diff
@@ -696,4 +699,87 @@ public void testGetProductsByVersionsDoesntFailWithLargeDataSets() {
...
+        expected.put(SystemPurposeAttributeType.ADDONS.toString(), addons);
+        expected.put(SystemPurposeAttributeType.ROLES.toString(), roles);
+        expected.put(SystemPurposeAttributeType.SERVICE_LEVEL.toString(), support);
+
+        Map<String, Set<String>> result = this.ownerProductCurator.getSyspurposeAttributesByOwner(owner);
+        assertEquals(result, expected);
+    }
+
+    @Test
+    public void testGetNoSyspurposeAttributesByOwner() {
```
Ceiu, line 779: Still need tests for the null owner and ownerId cases

Ceiu (changes requested):

Ceiu (approved):

PR #3316, opened 2022-04-22
2070159: Update resource locator annotation warnings

wottop:
Checks the interfaces for the Resource classes
Downgraded log warning to info for PUT/POST with no @Consumes as this
 is the expected state

### DIFF
--- a/src/main/java/org/candlepin/resteasy/ResourceLocatorMap.java
+++ b/src/main/java/org/candlepin/resteasy/ResourceLocatorMap.java
@@ -215,14 +215,20 @@ public class ResourceLocatorMap implements Map<Method, ResourceLocator> {
             String name = m.getDeclaringClass() + "." + m.getName();
             registered.append("\t" + name + "\n");
 
-            if (!m.isAnnotationPresent(Produces.class)) {
+            if (!m.isAnnotationPresent(Produces.class) &&
+                !isAnnotationPresentInInterface(m, Produces.class)) {
                 missingProduces.append("\t" + name + "\n");
             }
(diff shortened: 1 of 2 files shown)


JoshAlbrecht23:
These changes look good to me, but I'd like to get a second opinion!

src/main/java/org/candlepin/resteasy/ResourceLocatorMap.java
```diff
@@ -234,7 +240,8 @@ protected void logLocators() {
                 continue;
             }
 
-            if (!m.isAnnotationPresent(Consumes.class)) {
+            if (!m.isAnnotationPresent(Consumes.class) &&
```
Januson, line 243: Tests are hitting NPE at line 255. I guess we have to seach for the interface annotation as well.

src/main/java/org/candlepin/resteasy/ResourceLocatorMap.java
```diff
@@ -215,14 +215,20 @@ protected void logLocators() {
             String name = m.getDeclaringClass() + "." + m.getName();
             registered.append("\t" + name + "\n");
 
-            if (!m.isAnnotationPresent(Produces.class)) {
+            if (!m.isAnnotationPresent(Produces.class) &&
```
Januson, line 218: Could you add some tests? We don't have any for this class.

Januson (changes requested):
Changes look good, but it looks like they introduced a NPE.

Januson (approved):

PR #3322, opened 2022-04-27
2078780: Added filter for fetching SCA certs with cert serial Id

JoshAlbrecht23:
I also verified the changes with the following steps:

1. Register with Subscription Manager to an organization that is org_environment content access mode.
2. Use a Get request on the candlepin/consumers/<Consumer ID>/certificates endpoint and use and serial number as a query parameter. (curl -k -u admin:admin --request GET --header 'accept: application/json' "https://localhost:8443/candlepin/consumers/<Consumer ID>/certificates?serials=12345678")
3. Verify that we do not get any certificates back.

### DIFF
--- a/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -95,7 +95,6 @@ import org.candlepin.model.DistributorVersion;
 import org.candlepin.model.DistributorVersionCapability;
 import org.candlepin.model.DistributorVersionCurator;
 import org.candlepin.model.Entitlement;
-import org.candlepin.model.EntitlementCertificate;
 import org.candlepin.model.EntitlementCurator;
 import org.candlepin.model.EntitlementFilterBuilder;
 import org.candlepin.model.EnvironmentContentCurator;
@@ -163,6 +162,7 @@ import java.io.IOException;
 import java.security.GeneralSecurityException;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2068,13 +2068,11 @@ public List<CertificateDTO> getEntitlementCertificates(@Verify(Consumer.class) S
...
         Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
-        ConsumerType ctype = this.consumerTypeCurator.getConsumerType(consumer);
 
         revokeOnGuestMigration(consumer);
         poolManager.regenerateDirtyEntitlements(consumer);
 
         Set<Long> serialSet = this.extractSerials(serials);
-
         List<CertificateDTO> returnCerts = new LinkedList<>();
         List<EntitlementCertificate> allCerts = entCertService.listForConsumer(consumer);
```
Ceiu, line 2077: We can use some of the streaming API to reduce the code footprint and only have the filtering logic expressed once:

```
Stream<EntitlementCertificate> certStream = this.entCertService.listForConsumer(consumer)
    .stream();

Certificate caCert = this.contentAccessManager.getCertificate(consumer);
if (caCert != null) {
    certStream = Stream.concat(certStream, Stream.of(caCert));
}

Set<Long> serialSet = this.extractSerials(serials);
if (!serialSet.isEmpty()) {
    certStream.filter(cert -> serialSet.contains(cert.getSerial().getId()));
}

return certStream.map(translator.getStreamMapper(EntitlementCertificate.class, CertificateDTO.class))
```

I dunno if that's any cleaner though. Some improvements could be to do the stream building in a utility function or to always add the ca cert output and the filter out nulls (safer in the event `listForConsumer` ever misbehaves).

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2068,13 +2068,11 @@ public List<CertificateDTO> getEntitlementCertificates(@Verify(Consumer.class) S
         }
 
         Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
-        ConsumerType ctype = this.consumerTypeCurator.getConsumerType(consumer);
```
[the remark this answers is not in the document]
Ceiu: :+1:

src/main/java/org/candlepin/resource/ConsumerResource.java
JoshAlbrecht23: I like combining both the certificates into a single collection and then filtering them all in a utility method! I wasn't sure if you were thinking of having the filtering method as a static method in some utility class or something, but I couldn't find a class that seemed to make sense. So, I put the logic in a private method in the ConsumerResource class.

Ceiu: The one thing we want to avoid is extraneous copies. The streaming methods provide a view of the base data, and `Stream.concat` gives us a lazy concatenated view of two streams. 

The location of the utility method is fine, since it's a one-off here, and is what I had in mind. However, I was meaning to combine the individual cert and the list of certs in the utility method, since that's the messiest part of this code-wise, rather than moving the filtering out.

I'll take a closer look at this tomorrow morning and see if we can take this one step further.

JoshAlbrecht23: Yeah, I figured we could just add the individual cert that is returned from contentAccessManager.getCertificate to the list of certificates for the consumer, and then create a stream from that total collection of certificates that way we only have one stream we need to filter.

I also added filtering for null values in filterCertificatesBySerialIds to protect against a null value being returned from contentAccessManager.getCertificate.

Ceiu (changes requested):
Something I noticed while writing up the code for the comment chain we have going on in the getEntitlementCertificates method is that this code is used for half the output of the actual endpoint `exportCertificates`. What we've been working on will address half of the issue, but the export logic looks to also be subject to this bug.

We should check to make sure both variations of the output are fixed before merging.

src/main/java/org/candlepin/resource/ConsumerResource.java
JoshAlbrecht23: Okay, this makes sense. I added a `collect(Collectors.toList())` to the certStream.map so that we can still return a List and remain passive.

JoshAlbrecht23:
> Something I noticed while writing up the code for the comment chain we have going on in the getEntitlementCertificates method is that this code is used for half the output of the actual endpoint `exportCertificates`. What we've been working on will address half of the issue, but the export logic looks to also be subject to this bug.
> 
> We should check to make sure both variations of the output are fixed before merging.

I verified that a GET request on the candlepin/consumers/{Consumer Id}/certificates?serials=12345678 endpoint does not return an incorrect certificate. I also verified that candlepin/consumers/{Consumer Id}/certificates does return the correct certificates.

JoshAlbrecht23:
> Something I noticed while writing up the code for the comment chain we have going on in the getEntitlementCertificates method is that this code is used for half the output of the actual endpoint `exportCertificates`. What we've been working on will address half of the issue, but the export logic looks to also be subject to this bug.
> 
> We should check to make sure both variations of the output are fixed before merging.

Good call about the other half of the `exportCertificates`. After testing, we definitely need to address that functionality as well. Unfortunately the the Exporter exports the content access certs and the entitlement certs into two different directories, and so I don't think that we will be able to filter all the certs at once. I added the filtering logic in `Exporter.exportContentAccessCerts` and added unit tests and functionally verified the changes are reflected in the zip file that is exported!

src/main/java/org/candlepin/resource/ConsumerResource.java
Ceiu: Okay, that's fine here, especially since the way this end point works is a bit wonky; but in the future we should look for opportunities to convert those to streams when we're operating on potentially medium to large lists. Also makes for a good chance to deal with the spec file a bit more.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2068,29 +2068,52 @@ public List<CertificateDTO> getEntitlementCertificates(@Verify(Consumer.class) S
...
+     * additional certificates that might not be included in the collection (such as the content
+     * access cert). Null elements will be filtered from the resulting stream.
+     *
+     * @param certificates
+     *  a collection of certificates to include in the certificate stream
+     *
+     * @param extraCerts
+     *  any additional certificates to include in the certificate stream
+     *
+     * @Return
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good catch, updated!

Ceiu (approved):

PR #3323, opened 2022-04-28
Update postgresql container to use version 12

wottop:
Updated the candlepin repo at quay.io to include the postgresql 12.10 docker image from docker.io.


Ceiu:
What was the reason we opted to keep our own version of the postgresql container instead of using an official container image?

PR #3324, opened 2022-04-29
ENT-4613: Migrate Identity and Authorization spec

Januson:


spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ActivationKeys.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+
+        private NestedOwnerDTO owner;
+
+
```
JoshAlbrecht23, line 42: Remove extra new line

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Consumers.java
```diff
@@ -0,0 +1,76 @@
...
+            ConsumerDTO consumer = new ConsumerDTO();
+            consumer.setName(requireNonNullElseGet(this.name, () -> StringUtil.random("test_consumer")));
+            if (this.owner != null) {
+                consumer.setOwner(this.owner);
+            }
+            ConsumerTypeDTO type = new ConsumerTypeDTO();
+            type.label("system");
+            consumer.type(type);
+            consumer.putFactsItem("system.certificate_version", "3.3");
+            // todo fill in rest of the data
```
JoshAlbrecht23, line 71: Update to the following so the todo is marked by IDEs:
`// TODO: fill in rest of the data`

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClient.java
```diff
@@ -0,0 +1,100 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+
+import java.util.Objects;
+
+/**
+ * An entrypoint of Candlepin client. Serves as a crossroads between specific APIs.
+ */
+public class ApiClient {
```
JoshAlbrecht23, line 40: I'm a little confused on why we need this wrapper class, and the naming is confusing as well considering it is org.candlepin.spec.bootstrap.client.ApiClient that wraps the org.candlepin.ApiClient.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClients.java
```diff
@@ -0,0 +1,63 @@
...
+ * in this software or its documentation.
+ */
+package org.candlepin.spec.bootstrap.client;
+
+import org.candlepin.dto.api.v1.CertificateDTO;
+
+/**
+ * Static factory created for easy selection of appropriate authentication method.
+ */
+public final class ApiClients {
```
JoshAlbrecht23, line 22: The name ApiClients is confusing to me, and do we need this static class? Can we expect our spec tests to extend the SpecTestFixture class which already has the ApiClientFactory? If we do, we could move these methods to the abstract class and let the spec tests inherit them for free.

JoshAlbrecht23 (changes requested):

spec-tests/src/main/java/org/candlepin/spec/bootstrap/assertions/StatusCodeAssertions.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public static ApiException catchApiException(ThrowableAssert.ThrowingCallable code) {
+        ApiException exception = AssertionsForClassTypes.catchThrowableOfType(code, ApiException.class);
+        assertThat(exception).as("Expected ApiException but nothing was thrown").isNotNull();
+        return exception;
+    }
+
+    private static AbstractThrowableAssert<?, ? extends Throwable> assertReturnCode(
+        ApiException exception, int statusCode) {
```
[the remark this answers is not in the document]
Januson: Done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClientFactory.java
```diff
@@ -109,4 +174,123 @@ private static SSLContext getSslContext(TrustManager[] trustAllCerts) {
...
+    }
+
+    private static class ActivationKeyInterceptor implements Interceptor {
+
+        private final String ownerKey;
+        private final String activationKeys;
+
+        public ActivationKeyInterceptor(String ownerKey, String activationKeys) {
+            this.ownerKey = ownerKey;
+            this.activationKeys = activationKeys;
```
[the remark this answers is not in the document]
Januson: I don't think it's needed. This interceptor is used to enable activation key auth for endpoints that do not accept activation key as a parameter. So there's only a few uses for it and auth will fail if they are not present. I'll add a comment.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/UserUtil.java
```diff
@@ -0,0 +1,48 @@
...
+import org.candlepin.spec.bootstrap.data.builder.Roles;
+
+public final class UserUtil {
+
+    private UserUtil() {
+        throw new UnsupportedOperationException();
+    }
+
+    public static UserDTO createUser(ApiClient client, OwnerDTO owner) throws ApiException {
+        //        user = user_client(owner, random_string("user"))
```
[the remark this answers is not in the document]
Januson: Done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/UserUtil.java
```diff
@@ -0,0 +1,48 @@
...
+        RolesApi roles = client.roles();
+        UsersApi usersClient = client.users();
+        UserDTO user = new UserDTO()
+            .username(StringUtil.random("test_user"))
+            .password("password")
+            .superAdmin(false);
```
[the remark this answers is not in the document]
Januson: I was thinking more along the line of having createUser/createAdminUser. But I didn't wanted to add a pile of code that would be left unused untill different task gets to deal with different kinds of user permissions.

src/test/java/org/candlepin/auth/TrustedUserAuthTest.java
```diff
@@ -87,10 +87,11 @@ public void missingUsernameHeader() {
 
     @Test
     public void normalTrustedAuth() throws Exception {
+        User u = new User(USERNAME, "pass");
+        when(userService.findByLogin(eq(USERNAME))).thenReturn(u);
```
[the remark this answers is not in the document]
Januson: Why? `doReturn` is useful for stubbing void methods but `when` is better for normal cases as it's type-safe.

src/main/java/org/candlepin/auth/TrustedUserAuth.java
```diff
@@ -59,6 +61,11 @@ public Principal getPrincipal(HttpRequest httpRequest) {
             return createPrincipal(username);
         }
 
+        UserInfo user = this.userServiceAdapter.findByLogin(username);
+        if (user == null) {
+            throw new NotAuthorizedException(i18nProvider.get().tr("User does not exist!"));
```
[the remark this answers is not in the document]
Januson: Done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/cert/X509Cert.java
```diff
@@ -0,0 +1,92 @@
...
+    public X509Cert(X509Certificate certificate) {
+        this.certificate = Objects.requireNonNull(certificate);
+    }
+
+    public String subject() {
+        Principal subjectDN = certificate.getSubjectDN();
+        return subjectDN.getName();
+    }
+
+    public  LocalDateTime notBefore() {
```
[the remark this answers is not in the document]
Januson: Done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ActivationKeys.java
Januson: Done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Consumers.java
Januson: Done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClient.java
Januson: The purpose is to handle creation of different APIs and sharing an instance of okhttp client between them. Naming could be improved. I'll try to figure something out but I am open to suggestions.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClients.java
Januson: The point here is to have a well defined entry point for the client separate from the test code. Treating the client as a library will help us keeping it maintainable. Abstract parent of all spec test is likely to become a dumping ground for everything related like DatabaseTestFixture.

wottop:
I am seeing the following failure:

http://pastebin.test.redhat.com/1049850

Januson:
@wottop Have you redeployed Candlepin? There is a small change that adds missing check for existence of trusted user. We accepted auth request with any user even if it does not exist.

spec-tests/src/test/java/org/candlepin/spec/auth/ActivationKeyAuthSpecTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+        StatusDTO status = client.status();
+
+        assertThat(status).isNotNull();
+    }
+
+    @Test
+    @DisplayName("activation key auth requires owner")
+    void shouldRequireOwner() {
+        var client = ApiClients.noAuth().consumers();
```
Ceiu, line 90: Nack on everything "var." Java is a strongly-typed language and I would like to keep it that way.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClients.java
Ceiu: While I agree that DBTestFixture is somewhat bloated with stuff that could go elsewhere, I feel all the scaffolding we're building here is the same class of problem just spread out more. Like we have ApiClients (plural), ApiClient (singular), a pile of interceptors, and a bunch of other stuff needed for each individual test. In a lot of ways, we're moving the pain from bloat on a class that we just use for a certain type of test, to needing a ton of boilerplate in the tests themselves. Eventually that boilerplate is going to be gathered up in a single class again, and we'll be right back to the DBTestFixture problem.

Ceiu (changes requested):
I'm not really a fan of the split of logic between the various clients making requests, and the output validation of said request. To implement a new test, it feels like we're building two classes that live in separate places, and the logic has to be kept in sync for a general test. On top of that, our client classes have a downgrade in that there's no general-purpose request APIs for just sending a request of a type with certain headers and/or parameters. Really, that should be the basis of any client object, and the other extensions should be little more than wrappers around it -- if they even need to exist. Worse, because they're all separated, the more complex tests are going to end up using a bunch of them.

Also, use of the "var" keyword outside of very specific circumstances is a no-go for me. Static analysis gets much more painful when the typing isn't clear, and I don't really feel like dealing with the pain of unexpected type swapping when the callee is refactored in the future.

spec-tests/src/test/java/org/candlepin/spec/auth/ActivationKeyAuthSpecTest.java
JoshAlbrecht23: +1

Januson:
@Ceiu Could you elaborate? I am not sure what you mean by the split of logic. There is a logic that allows you get an authenticated instance of the client but it is built on the code we get from OpenAPI generator.

Regarding the downgrade, that was kind of the point right? To have a client of our API generated according to the specification, not to have a general purpose request builder. We could create a wrapper that enables use of all endpoints at once but we would have to manage it manually and keep it in sync with generated classes since it would be using them in the background. With the current solution you have to select which resource you want to talk with but it's easily navigable since it corresponds 1:1 to our API.

spec-tests/src/test/java/org/candlepin/spec/auth/ActivationKeyAuthSpecTest.java
Januson: Done. I just wonder what issue are you hitting with var? Java is a strongly-typed language, type inference does not change that and even with extensive refactoring it's still compile-time checked.

Ceiu: Static analysis. I understand how var works, it's just an utter pain to read code that makes heavy use of it, and there are a handful of edge cases where it can cause problems during refactors due to its semi-dynamic nature that I'd really rather not deal with.

Ceiu:
The biggest thing I'm really not looking forward to here is when we start getting into the more in-depth spec tests that require creating owners, products, content, pools, etc.. We'll have to get the OwnerClient, and then the ProductClient, and then the ContentClient, etc. just to do our data setup. There's a ton of that boilerplate/scaffolding that looks like it's primed to get out of hand real quick as soon as the test is no longer operating within the context of a single type or resource.

Januson:
It's a tradeoff. Having all endpoint calls in one place would be nice from a caller's perspective, but it could be hard to navigate. The overhead of keeping it in sync would be on us and maintaining a class with thousands of lines is no fun.

I added ApiClients/ApiClient to simplify getting an authenticated client and passing it around. You don't have to keep those instances around, instead you can use them inlined like:
```java
client.owners().create(...)
client.products().create(...)
...
client.consumers().update(...)
```

It's a little bit of typing but you know exactly to which resource you are talking to. *Api classes do no initialization so creating them is cheap and even if we end up in a situation where it becomes a problem we can cache creation of those *Api wrappers in ApiClient.

One complication is when you need to used multiple clients with different auth methods. Like starting with basic auth and then creating a couple consumer clients. Unfortunately I don't see a way around this.

Ceiu:
I would argue that when it comes to designing the framework or API for something, the caller *is* the focus. With the spec tests, it's not really important to know which specific resource we're hitting, only which endpoint(s) we're attempting to test. That being the case, there's actually an argument there that the endpoint should be explicitly stated in the test itself rather than being placed behind a wrapper; even if that wrapper is just a facade for a generated API interface.

I dunno, it's an interesting set of problems and approaches to said problems, and I think our team as a whole never really being clear about what we want and need out of the spec tests is why we're in this boat to begin with.

For the time being this is fine, I suppose. However my long-term concern is as we keep going down this road, even if we decide we don't like it later, the level of effort required to pull back goes up, decreasing the probability we'll ever have the chance to address such things even if we exceed some threshold of allowed pain.

Ceiu:
Oh, it was just pointed out to me that this doesn't remove the old rspec versions of these tests. We should remove those test suites if they've been obsoleted.

Januson:
@Ceiu Done. Removed obsolete spec test.

PR #3327, opened 2022-05-03
Optimize hibernate use in AsyncJobStatusCuratorTest

barnabycourt:
General updates to the AsyncJobStatusCuratorTest unit tests. Locally, this reduced runtime from over 20 minutes to ~30 seconds.  
* Remove extraneous flushes
* Re-organize the createJobsForQueryTests() in order to reduce database load / time to execute.

### DIFF
--- a/src/test/java/org/candlepin/model/AsyncJobStatusCuratorTest.java
+++ b/src/test/java/org/candlepin/model/AsyncJobStatusCuratorTest.java
@@ -38,6 +38,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Random;
 import java.util.Set;
+import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


Ceiu:
I took the idea and did a bit of optimization and cleanup on it [1]. I'd be curious to hear your take on it and see if it passes with the same level of efficiency.

[1] https://gist.github.com/Ceiu/781b551e8c46dd59fd18dfdb8900dc73

barnabycourt:
@Ceiu I updated based on some of the suggestions in the gist. I made some changes to the suggestions for the createJob() method in order to drop an unused parameter & make sure that the block on flush was passed through to the generic create method. 

I removed the block size from the in clause as HSQLDB does not have an in clause limit & this call is only used in testing. I tried setting a 1k limit for the in clause in HSQLDB but that seemed to cause general slowdowns in other parts of the suite.

Ceiu (approved):

PR #3328, opened 2022-05-04
2078029: Unlimited bonus pool quantity adjustment after manifest change

wottop:


src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java
```diff
@@ -749,16 +749,19 @@ private PoolOperationCallback adjustHostedBonusPoolQuantity(PoolManager poolMana
...
-                        List<Pool> pools = subscriptionPoolMap.get(pool.getSubscriptionId());
-                        if (pools != null) {
-                            for (int idex = 0; idex < pools.size(); idex++) {
-                                Pool derivedPool = pools.get(idex);
-                                if (derivedPool.getAttributeValue(Pool.Attributes.DERIVED_POOL) != null) {
+                    // getting all pools matching the sub id. Filtering out
+                    // the 'parent'.
+                    List<Pool> pools = subscriptionPoolMap.get(pool.getSubscriptionId());
+                    if (pools != null) {
+                        for (int idex = 0; idex < pools.size(); idex++) {
```
Ceiu, line 756: Let's modernize this loop and address the piles of nesting. Either streams or a for-each loop would clean it up substantially. For example:
```
pools.stream()
    .filter(pool -> pool.getAttributeValue(Pool.Attributes.DERIVED_POOL) != null)
    .forEach(pool -> poolOperationCallback.setQuantityToPool(pool, 
        pool.getQuantity().equals(notConsumerLocally) ? 0 : -1));
```

Ceiu (changes requested):

src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java
```diff
@@ -749,19 +749,14 @@ private PoolOperationCallback adjustHostedBonusPoolQuantity(PoolManager poolMana
...
-                                Pool derivedPool = pools.get(idex);
-                                if (derivedPool.getAttributeValue(Pool.Attributes.DERIVED_POOL) != null) {
-                                    poolOperationCallback.setQuantityToPool(derivedPool, 0);
-                                }
-                            }
-                        }
-                    }
+                    // getting all pools matching the sub id. Filtering out
+                    // the 'parent'.
+                    List<Pool> pools = subscriptionPoolMap.get(pool.getSubscriptionId());
```
Ceiu, line 754: We still need the null check here, or make use of `getOrDefault`

Ceiu (changes requested):

src/main/java/org/candlepin/policy/js/entitlement/EntitlementRules.java
```diff
@@ -749,19 +749,15 @@ private PoolOperationCallback adjustHostedBonusPoolQuantity(PoolManager poolMana
...
-                    }
+                    // getting all pools matching the sub id. Filtering out
+                    // the 'parent'.
+                    List<Pool> pools = subscriptionPoolMap.getOrDefault(pool.getSubscriptionId(),
+                        new ArrayList<>());
+                    final Long exportCount = notConsumedLocally;
+                    pools.stream()
+                        .filter(thisPool -> thisPool.getAttributeValue(Pool.Attributes.DERIVED_POOL) != null)
+                        .forEach(thisPool -> poolOperationCallback.setQuantityToPool(thisPool,
+                        pool.getQuantity().equals(exportCount) ? 0 : -1));
```
JoshAlbrecht23, line 760: Do we also need to check that the export count is not 0? I'm a little confused on the expected outcome for the case where the exportCount is 0 and the pool.quantity() is also 0.

wottop: Quantity is total in the pool.
Quantity consumed are those used by consumers reporting to the server.
Quantity exported are those put into a manifest to be used in a Satellite.

If pool.getQuantity() is 0, then the others must also be 0.

If there is a bonus pool for this master pool, it would be logical [and the way it has always been] that it has a 0 quantity as well. You would have a master/bonus pool combo where no entitlements are available.

nikosmoum (approved):
ack from me; I'll let others re-review/merge

Ceiu (approved):

PR #3332, opened 2022-05-11
2083612: Add ssl_verify capability based on config [m] ENT-4976

wottop:


src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -149,6 +149,9 @@ private ConfigProperties() {
     // Messaging
     public static final String CPM_PROVIDER = "candlepin.messaging.provider";
 
+    // OCSP stapling
+    public static final String SSL_VERIFY = "candlepin.sslverifystatus";
```
jirihnidek, line 153: Can you add it to `config/candlepin/candlepin.conf.template` somehow? E.g. It could be mentioned in comment.

nikosmoum, line 153: Let's also set the default value (false) lower down in the same class as we do with other properties

nikosmoum (commented):
Let's also use BZ id (for this and backports)

src/main/java/org/candlepin/config/ConfigProperties.java
wottop: So added

wottop: Done

Ceiu: Why are we defaulting it to false if this is just a flag that has no barring on CP itself? From how this has been described to me multiple times now, it sounds like it's going to be available more than it's not, and it's up to each individual host to decide to shut it off if they don't/can't support it. If we default to false, every existing Sat will need to explicitly add the entry to turn it on.

Not gonna lie, this whole thing seems real hacky and weird; I'm not a fan. It feels like CP should be at least checking something locally with the stack to check if the feature is something we can support even if the config is set to true.

nikosmoum: @Ceiu It's the opposite: Satellite as of now does not use OCSP, so has no need for this config/capability. It is only Hosted that will roll out OCSP, and they can easily set this to true explicitly at that moment.

Unfortunately candlepin can't actually check anything, because OCSP is not provided by the candlepin/tomcat stack. It's provided by the CDN. To the extent I was involved in the design for this, I've not heard of a better solution to this.

Ceiu: ack; disregard my comment

nikosmoum (approved):

src/main/java/org/candlepin/config/ConfigProperties.java
nikosmoum: btw, adding this to the template means this property would only be added to candlepin.conf if the sslverifystatus_enabled variable is set to true, which is the kind of thing that the deploy script usually does, but I don't see the corresponding change to deploy script. So if left as is, this property won't be added to candlepin.conf

nikosmoum: in addition to that though: there is no way for anyone to actually test this feature except in Hosted Prod, so I don't see the point of setting it to true?

nikosmoum: If @jirihnidek is ok with not including this in the .conf file, I can merge the PRs as they are

JoshAlbrecht23 (approved):

src/main/java/org/candlepin/config/ConfigProperties.java
jirihnidek: @nikosmoum I think it is OK and the PR can be merged.

PR #3338, opened 2022-05-13
[M] Changed Maven plugin to maven-publish for Gradle 7+ (ENT-4797)

JoshAlbrecht23:
Changed build.gradle to use the maven-publish plugin instead of the maven plugin for compatibility using Gradle 7+.  Unfortunately we are somewhat limited with the properties and methods with maven-publish and I had to generate most of the pom.xml with the withXml method.

https://docs.gradle.org/current/dsl/org.gradle.api.publish.maven.MavenPom.html

**Pom is generated running the following:**
`gradlew generatePomFileForLocalPublication`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ plugins {
     // Openapi version is set in the buildscript block for use in the pom generation as well
     id "org.openapi.generator" version '5.4.0'
     id "java"
-    id "maven"
+    id "maven-publish"
     id "war"
     id "checkstyle"
     id 'org.owasp.dependencycheck' version '7.0.0'
@@ -37,6 +37,7 @@ plugins {
     id 'org.candlepin.gradle.gettext'
     id 'org.candlepin.gradle.msgfmt'
     id "org.candlepin.gradle.SpecVersion"
+    id "org.candlepin.gradle.PomToolkit"
 }
 
 group = "org.candlepin"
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -488,285 +488,286 @@ tasks.withType(Checkstyle) {
     exclude "**/i18n/Messages*"
 }
 
-task pom {
```
Januson, line 491: Can you add "pom" alias to the generatePomFileForLocalPublication task?

build.gradle
```diff
@@ -488,285 +488,286 @@ tasks.withType(Checkstyle) {
...
+                createProfileNode(profilesNode, 'build-with-jss4', '/usr/lib64/jss/jss4.jar', '4.9.1')
+                createProfileNode(profilesNode, 'build-with-jss5-plus', '/usr/lib64/jss/jss.jar', '5.0.0')
             }
-        }.writeTo("pom.xml")
+        }
     }
+  }
+}
+
+static void createDependencyNode(Node parentNode, String groupId, String artifactId, String version, String scope, String systemPath) {
```
Januson, line 594: This task grew quite a lot. It should probably move to buildSrc but it depends on how long we will keep it around before migrating away from pom files.

Januson (changes requested):
Changes look good. I have just a small request and a note whether it's worth to move the task out of the buil.gradle file.

build.gradle
JoshAlbrecht23: Good call! I added the "pom" alias

build.gradle
JoshAlbrecht23: I'm not sure if there are plans to move away from the pom files and if it makes sense to move to the buildSrc directory or not. I'm curious on what @Ceiu thinks.

Ceiu: As far as I know, the pom files are necessary due to the how our build system works and the various pieces within it. We've recently discussed moving to different build systems which wouldn't need the pom files, but I don't think we're very far along on that yet.

As far as compartmentalizing the task, if there's a clean way to do it such that it ends up in a logical place and is self-contained (i.e. not split across several files/modules), that might be better than throwing piles of stuff in the build.gradle file.

JoshAlbrecht23: I created a new class, PomToolkit, and moved over the pom xml creation logic to static utility methods that can be used!

Januson (approved):

PR #3341, opened 2022-05-18
[M] Added grace period for product deletion during refresh/import (ENT-4881)

Ceiu:
- Products orphaned during refresh or manifest import can now
  have a grace period before they will be deleted
- Added the config "candlepin.refresh.orphaned_entity_grace_period"
  to control the duration of the grace period in days, or disable
  it entirely. By default this option is set to 30 days.

### DIFF
--- a/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -285,6 +285,10 @@ public class ConfigProperties {
     // How long (in seconds) to wait for job threads to finish during a graceful Tomcat shutdown
     public static final String ASYNC_JOBS_THREAD_SHUTDOWN_TIMEOUT = "candlepin.async.thread.shutdown.timeout";
 
+    // How many days to allow a product to be orphaned before removing it on the next refresh or
+    // manifest import. Default: 30 days
+    public static final String ORPHANED_ENTITY_GRACE_PERIOD = "candlepin.refresh.orphan_entity_grace_period";
+
     /**
      * Fetches a string representing the prefix for all per-job configuration for the specified job.
      * The job key or class name may be used, but the usage must be consistent.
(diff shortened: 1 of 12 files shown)


Ceiu:
For this implementation, I opted to keep the configuration aspect as the CandlepinPoolManager level, but I could see an argument to moving the config parsing lower. There's no real concrete reasoning behind the decision to how it landed, so if this doesn't jive with others or someone feels strongly about doing it a different way, I'm open to suggestions.

src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -285,6 +285,10 @@ private ConfigProperties() {
     // How long (in seconds) to wait for job threads to finish during a graceful Tomcat shutdown
     public static final String ASYNC_JOBS_THREAD_SHUTDOWN_TIMEOUT = "candlepin.async.thread.shutdown.timeout";
 
+    // How many days to allow a product to be orphaned before removing it on the next refresh or
+    // manifest import. Default: 30 days
+    public static final String ORPHANED_ENTITY_GRACE_PERIOD = "candlepin.refresh.orphan_entity_grace_period";
```
nikosmoum, line 290: I am unsure if we want this to be configurable. It would give customers a way to keep content that isn't covered by a subscription for too long or even indefinitely, bloating the database

Ceiu: True enough, but the worst-case here is they turn it off and we're back to how 3.x and prior operate.

Ceiu:
I'm realizing there's an optimization/cleanup that could be done with the product date fetching that could also be upgraded to use a per-product lookup as a fallback in the current state mismatch case that I might do next week.

nikosmoum (changes requested):
The only question I have is if it is possible to write a spec test for this? Otherwise it looks good

nikosmoum (approved):

PR #3343, opened 2022-05-19
ENT-4812: Add dependabot configuration

Januson:


nikosmoum (commented):
Have you tested this locally with dependabot-script? Does it catch gradle dependencies fine, without the need for moving them into a dependencies.gradle file?

PR #3345, opened 2022-05-24
Use msgfmt to validate the .po files [m] [ENT-3971]

wottop:


ptoscano:
Sorry, I realized I made a mistake when proposing this proposal: the right command for this job is not `msgmerge`, but `msgfmt`!
So you need something like: `msgfmt -o /dev/null -c file.po`

wottop:
This PR will fail CI until the issues in the PO files a corrected in Weblate. we should wait to merge until then.

nikosmoum (approved):

PR #3346, opened 2022-05-24
[M] Added new inactive consumer cleanup job (ENT-4984)

JoshAlbrecht23:
Implemented a new inactive consumer cleanup job based on the following design document.

Design Document: https://docs.google.com/document/d/1QnttE45Ch8AuW7WuBEdAlbTi2VLBF267043zzzTbrtE/edit#


src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1001,6 +1008,107 @@ else if (systemUuid != null) {
...
+                    serialsToRevoke.add(idCert.getSerial().getId());
+                }
+                ContentAccessCertificate contentAccessCert = consumer.getContentAccessCert();
+                if (contentAccessCert != null) {
+                    caCertsToDelete.add(contentAccessCert.getId());
+                    serialsToRevoke.add(contentAccessCert.getSerial().getId());
+                }
+
+                this.delete(consumer);
+            });
```
JoshAlbrecht23, line 1052: In the design document it mentions inserting the rows into the cp_deleted_consumers table with a single insert statement with a subquery for the entire batch.  This is ideal, but I wasn't able to find a way to populate the non-null key column "id" for the cp_deleted_consumers table this way. I tried using the random_uuid function, but got an error saying I did not have permissions to use the function.

This isn't ideal as we are deleting the consumers individually and maybe doing some performance testing on this would be a good idea to get an idea on how it performs.

**Insert statement I was using:**
```
INSERT INTO cp_deleted_consumers (created, updated, consumer_uuid, owner_id, owner_displayname, owner_key, principal_name)
SELECT NOW(), NOW(), consumer.uuid, consumer.owner_id, owner.displayname, owner.account, :principalName
FROM cp_consumer consumer
JOIN cp_owner owner on owner.id=consumer.owner_id
WHERE consumer.uuid IN (:consumerIds)
```

Ceiu: Couple things to note here:
- Likely due to an old database limitation or some other such decision, our UUID columns aren't *real* UUID columns; we pull the hyphens out, bringing the character count down to 32 (from 36 IIRC). So even if you had a properly functioning UUID function, it probably wouldn't fit and/or would be truncated.
- PostgreSQL and MySQL/MariaDB have different syntax for doing UUID generation, so you wouldn't be able to generalize it anyway
- HQL/JPQL don't have any nice ways of doing it directly. You could try doing the general function call (`function(...)`), but I'd have to look up how the INSERT ... INTO works with HQL/JPQL. I remember it being fairly limited, so it might not even be an option. I'm not clear on the details there atm, however.

All that said, it's not critical that we have well-formed UUIDs for our objects. It's definitely a nice-to-have, but it's not strictly required for our purposes. We've hit similar problems in other migrations and tasks, and have used the following to get around it: `MD5(CONCAT(<some safe set of unique columns and/or NOW()>))`

MD5 isn't cryptographically secure anymore, but it's still functional for non-security purposes, and it happens to generate 32-character strings. So long as the inputs are selected carefully and we do our best to minimize the chance of collision, that should be sufficient.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1001,6 +1008,107 @@ else if (systemUuid != null) {
...
 
+    /**
+     * Deletes inactive consumers and creates {@link DeletedConsumer} records.
+     * The identity certificate and content access certificate for the consumers
+     * are removed and their serials are revoked.
+     *
+     * @param inactiveConsumerIds - ids of inactive consumers to delete.
+     * @return the number of consumers that have been deleted.
+     */
+    @Transactional
```
Ceiu, line 1019: We should avoid adding transaction management to/in curator methods.

Ceiu (changes requested):
There are some other things I would change, but I'm expecting larger changes based on the conversation surrounding the SELECT INTO statement.

src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -485,11 +487,11 @@ public static String jobConfig(String jobKey, String cfgName) {
                 OrphanCleanupJob.DEFAULT_SCHEDULE);
             this.put(jobConfig(UnmappedGuestEntitlementCleanerJob.JOB_KEY, ASYNC_JOBS_JOB_SCHEDULE),
                 UnmappedGuestEntitlementCleanerJob.DEFAULT_SCHEDULE);
+            this.put(jobConfig(InactiveConsumerCleanerJob.JOB_KEY, ASYNC_JOBS_JOB_SCHEDULE),
+                ConfigProperties.ASYNC_JOBS_MANUAL_SCHEDULE);
```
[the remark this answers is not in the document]
nikosmoum: This is by design. The job is not supposed to be run in Satellite at all, only in Hosted, so it's to be disabled by default and Hosted can override it in candlepin.conf

src/test/java/org/candlepin/async/tasks/InactiveConsumerCleanerJobTest.java
```diff
@@ -124,4 +136,46 @@ private void setRetentionDaysConfiguration(String configurationName, int retenti
...
+     * because the hsqldb no longer has the MD5 class.
+     *
+     * @param input - used to generate the MD5 hash.
+     * @return a generated MD5 hash.
+     * @throws RuntimeException
+     */
+    public static final String md5Encode(String input) throws RuntimeException {
+        byte[] data;
+        try {
+            data = input.getBytes("ISO-8859-1");
```
Ceiu, line 171: It's a moot point if we don't actually need to hash, but we'd generally want to use UTF-8 here rather than 8859. Our actual DB text fields are all UTF8, so our logic should match/reflect that.

src/test/java/org/candlepin/async/tasks/InactiveConsumerCleanerJobTest.java
```diff
@@ -68,14 +75,20 @@ public void testExecutionWithInactiveCheckedInTime() throws JobExecutionExceptio
         inactiveConsumerCleanerJob.execute(context);
 
         consumerCurator.flush();
+        consumerCurator.clear();
+        deletedConsumerCurator.flush();
```
Ceiu, line 79: Flush and clear are global operations. We only need to do them once, regardless of which curator handles the op.

Ceiu (changes requested):
While native queries are okay when necessary, we still want to avoid them if the JPA layer will handle what we need to do properly and efficiently.

nikosmoum (changes requested):

src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Objects;
+
+
+
+/**
+ * If enabled, the InactiveConsumerCleanerJob periodically runs and removes inactive consumers
+ * based on their last checked in date and last updated dated. The identity certificate and
+ * content access certificate are removed and their serials are revoked.
+ */
+public class InactiveConsumerCleanerJob implements AsyncJob {
```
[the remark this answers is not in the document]
nikosmoum: And actually, you could even throw the entitlement variable there: an old lastcheckin consumer with an entitlement, and the other without entitlement

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1001,6 +1010,191 @@ else if (systemUuid != null) {
...
+        if (lastUpdatedRetention == null) {
+            throw new IllegalArgumentException("Last updated retention date cannot be null.");
+        }
+
+        String sql =
+            "SELECT consumer.id " +
+            "FROM cp_consumer consumer " +
+            "JOIN cp_consumer_type type ON type.id=consumer.type_id " +
+            "LEFT JOIN " +
+            "    (SELECT DISTINCT consumer_id FROM cp_entitlement) ent ON ent.consumer_id=consumer.id " +
```
nikosmoum, line 1185: Also, I don't think the subquery here will work for the purpose of fetching only consumers without entitlements.
it should probably just be `...LEFT JOIN cp_entitlement ent ON ent.consumer_id=consumer.id...`

src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
```diff
@@ -0,0 +1,117 @@
...
+            DEFAULT_LAST_CHECKED_IN_RETENTION_IN_DAYS);
+        Date nonCheckedInRetention = getRetentionDate(CFG_LAST_UPDATED_IN_RETENTION_IN_DAYS,
+            DEFAULT_LAST_UPDATED_IN_RETENTION_IN_DAYS);
+
+        int deletedCount = 0;
+        List<String> inactiveConsumerIds = consumerCurator.getInactiveConsumerIds(lastCheckedInRetention,
+            nonCheckedInRetention);
+        for (List<String> batch : Iterables.partition(inactiveConsumerIds,
+            ConsumerCurator.INACTIVE_CONSUMER_BLOCK_SIZE)) {
+            if (log.isDebugEnabled()) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Removed the check and left just the `log.debug`

src/main/java/org/candlepin/config/ConfigProperties.java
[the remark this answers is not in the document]
JoshAlbrecht23: Yeah, I think that we can remove the DEFAULT_SCHEDULE since it won't be used

src/main/java/org/candlepin/model/ConsumerCurator.java
JoshAlbrecht23: Removed, and added the transaction management in the InactiveConsumerCleanerJob

src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
```diff
@@ -0,0 +1,117 @@
...
+                log.debug("Cleaning inactive consumers with a batch of ids: {}", batch);
+            }
+
+            deletedCount += (int) consumerCurator.deleteInactiveConsumers(batch);
+        }
+
+        if (deletedCount > 0) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Adjusted the logger.info message and the job result to include the number of consumers that were cleaned up.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1028,42 +1030,127 @@ public int deleteInactiveConsumers(Collection<String> inactiveConsumerIds) {
...
+            serialIds.addAll(this.currentSession()
+                .createSQLQuery(caCertSql)
+                .setParameter("certIds", caCertIds)
+                .getResultList());
+        }
 
-        // Need to flush to avoid foreign key constraint violations with the removed consumers.
-        flush();
+        if (idCertIds != null && !idCertIds.isEmpty()) {
+            String idCertSql =
```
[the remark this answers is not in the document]
JoshAlbrecht23: Changed to use HQL.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1028,42 +1030,127 @@ public int deleteInactiveConsumers(Collection<String> inactiveConsumerIds) {
...
+     * @param consumerIds - ids of the consumers to delete.
+     * @return the number of consumer that were deleted.
+     */
+    public int deleteConsumers(Collection<String> consumerIds) {
+        if (consumerIds == null || consumerIds.isEmpty()) {
+            return 0;
+        }
 
-        return inactiveConsumerIds.size();
+        String deleteConsumersStatement = "DELETE FROM cp_consumer WHERE id IN (:consumerIds)";
```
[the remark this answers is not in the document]
JoshAlbrecht23: Changed to use HQL.

src/test/java/org/candlepin/async/tasks/InactiveConsumerCleanerJobTest.java
JoshAlbrecht23: Awesome! I didn't realize they were global operations.

src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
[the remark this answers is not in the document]
JoshAlbrecht23: I'm currently working on adding the spec tests

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -88,6 +90,9 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
     /** Regular expression used to convert fact expressions to SQL-safe LIKE expressions */
     private static final Pattern FACT_TRANSLATION_REGEX = Pattern.compile("(\\\\?+)([*?]|(?<=\\\\)\\\\)");
 
+    /** Recommended batch size for removing inactive consumers */
+    public static final int INACTIVE_CONSUMER_BLOCK_SIZE = 1000;
```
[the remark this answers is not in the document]
JoshAlbrecht23: Moved this config to the InactiveConsumerCleanerJob class and made it configurable so that it can be adjusted!

src/main/java/org/candlepin/model/ConsumerCurator.java
JoshAlbrecht23: Good catch! Updated

src/main/java/org/candlepin/model/ConsumerCurator.java
JoshAlbrecht23: So, after writing a spec test to test the job, I was running into an error about the update needing to be in a transaction when the method in InactiveConsumerCleanerJob that called the curator had the transactional annotation. Moving the Transactional annotation back to the curator method resolved that error.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1028,42 +1030,127 @@ public int deleteInactiveConsumers(Collection<String> inactiveConsumerIds) {
...
+     * and {@link IdentityCertificate} ids.
+     *
+     * @param caCertIds - ids of content access certificates.
+     * @param idCertIds - ids of identity certificates.
+     * @return serial ids.
+     */
+    public List<Long> getSerialIdsForCerts(Collection<String> caCertIds, Collection<String> idCertIds) {
+        List<BigInteger> serialIds = new ArrayList<>();
+        if (caCertIds != null && !caCertIds.isEmpty()) {
+            String caCertSql =
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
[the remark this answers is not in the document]
JoshAlbrecht23: I added a spec test however I wasn't able to test the null last checked in date with an outdated last updated time because the update time is automatically set at the current date like you said.

api/candlepin-api-spec.yaml
```diff
@@ -4100,6 +4100,12 @@ paths:
         - Jobs
       operationId: scheduleJob
       security: []
+      requestBody:
+        required: false
+        content:
+          application/json:
+            schema:
+              type: object
```
JoshAlbrecht23, line 4108: While implementing the spec tests for the new inactive consumer job, I ran into the following issue with OpenApi when using a new jobs client that extends the generated JobsApi class.
The issue is that the POST on the /jobs/schedule/{jobKey} endpoint does not have a request body defined and so an NPE occurs in the ApiClient.buildRequest method. Adding an object as the request body was the only way I could find to work around the issue.

**OpenAPI bug**
https://github.com/OpenAPITools/openapi-generator/issues/11833

```
InactiveConsumerCleanerJobTest > shouldDeleteInactiveConsumers() FAILED
    java.lang.NullPointerException: Parameter specified as non-null is null: method okhttp3.MediaType$Companion.parse, parameter $this$toMediaTypeOrNull
        at okhttp3.MediaType$Companion.parse(MediaType.kt)
        at okhttp3.MediaType.parse(MediaType.kt)
        at org.candlepin.ApiClient.buildRequest(ApiClient.java:1131)
        at org.candlepin.ApiClient.buildCall(ApiClient.java:1087)
        at org.candlepin.resource.JobsApi.scheduleJobCall(JobsApi.java:953)
        at org.candlepin.resource.JobsApi.scheduleJobValidateBeforeCall(JobsApi.java:965)
        at org.candlepin.resource.JobsApi.scheduleJobWithHttpInfo(JobsApi.java:1008)
        at org.candlepin.resource.JobsApi.scheduleJob(JobsApi.java:987)
        at org.candlepin.spec.bootstrap.client.JobsClient.scheduleJob(JobsClient.java:27)
        at org.candlepin.spec.InactiveConsumerCleanerJobTest.shouldDeleteInactiveConsumers(InactiveConsumerCleanerJobTest.java:93)
```

nikosmoum: Is adding the request body on the API spec the only way to work around this? It would be preferable if we can work around it on the client side somehow.

nikosmoum: Also, your branch needs rebasing since it has conflicts after https://github.com/candlepin/candlepin/pull/3324 was merged in master (and that one has a lot of changes on the java spec tests, so you'll have to adjust your spec test code)

src/main/java/org/candlepin/model/ConsumerCurator.java
nikosmoum: Let's identify the root cause behind that error since there must be something else going wrong; we should be able to set the transaction level at the level we want

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @Test
+    public void shouldDeleteInactiveConsumers() throws Exception {
+        Instant inactiveTime = Instant.now().minus(DEFAULT_LAST_CHECKED_IN_RETENTION_IN_DAYS + 10,
+            ChronoUnit.DAYS);
+        Instant activeTime = Instant.now().minus(DEFAULT_LAST_CHECKED_IN_RETENTION_IN_DAYS - 10,
+            ChronoUnit.DAYS);
+
+        ConsumerDTO inactiveConsumer = buildConsumer(inactiveTime.atOffset(ZoneOffset.UTC));
+        ConsumerDTO activeConsumer = buildConsumer(activeTime.atOffset(ZoneOffset.UTC));
+        ConsumerDTO activeConsumer2 = buildConsumer(activeTime.atOffset(ZoneOffset.UTC));
```
nikosmoum, line 88: Let's make one of the consumers have a null lastcheckin time. Their updated date will not be more than 30 days in the past, so we can assert that they will not get deleted.
Adding a manifest consumer that is inactive should also be easy (and assert they're not getting deleted)

nikosmoum: Also, we really need to verify that inactive consumers that have entitlements are not getting deleted, but I understand that would mean porting over the pools, products and entitlements client objects to java, so it is probably too much for a single task. 

So this begs the question: Maybe we should keep writing spec tests in Ruby for now (at least the complicated ones like this one), until all those are migrated to Java?
What does everyone think? @Januson @Ceiu @wottop @JoshAlbrecht23

nikosmoum (changes requested):

src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: What error? In general, if a given function only works because of the transactional annotation, that means our transaction management is busted, or a utility method has a hard-coded flush that needs to be gated or removed. We should seek to fix that issue rather than patching it up in a way that obscures/bandaids it as we've done in the past.

Our transaction management on the whole is pretty bad as a result of misuse of these annotations, and this has a direct impact on our locking model. We should avoid creating new situations where the boundaries of our transactions are unclear if at all possible.

src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: Resolving this, as the new queries don't hit this issue.

src/main/java/org/candlepin/model/ConsumerCurator.java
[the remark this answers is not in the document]
Ceiu: Looks like the other ones got updated, but the query for this comment in particular still looks to be native.

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
```diff
@@ -0,0 +1,124 @@
...
+            ConsumerTypeDTO newType = new ConsumerTypeDTO();
+            newType.setManifest(false);
+            newType.setLabel(Util.randomString("Test Type"));
+            consumerType = consumerTypeClient.createConsumerType(newType);
+        }
+    }
+
+    @Test
+    public void shouldDeleteInactiveConsumers() throws Exception {
+        Instant inactiveTime = Instant.now().minus(DEFAULT_LAST_CHECKED_IN_RETENTION_IN_DAYS + 10,
```
Ceiu, line 81: Style request: if possible, try to add line breaks on a method invocation rather than splitting up the parameters for a given call. For example:

```
Instant inactiveTime = Instant.now()
    .minus(DEFAULT_LAST_CHECKED_IN_RETENTION_IN_DAYS + 10, ChronoUnit.DAYS);
```

Makes it easier to see the individual steps at a glance, and what they are.

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
Ceiu: I'm torn. While it may be out of scope for this particular PR, doing it in Ruby at all is just doing work that we know is going to be ripped out in the immediate future. If it's just a matter of porting over client objects, that should be pretty simple, right? Maybe that could be a separate PR that blocks this one and then this gets rebased on it.

If it's a complicated task to get the requirements for writing a Java spec test in, then I could see defaulting to Ruby for the time being. So I guess, for me, it comes down to how much work it'd be to spin up the necessary components to write the desired tests.

JoshAlbrecht23: I like opening up a separate PR for porting over the client objects and then rebasing this PR with those changes! I'm working on that now.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ConsumerClient.java
```diff
@@ -0,0 +1,73 @@
...
+import org.candlepin.ApiException;
+import org.candlepin.dto.api.v1.ConsumerDTO;
+import org.candlepin.resource.ConsumerApi;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class ConsumerClient extends ConsumerApi implements SpecTestClient {
+
+    private final Set<String> createdConsumerUuids = new HashSet<>();
```
Januson, line 28: We should drop cleanup for now. It will be added later if needed.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ConsumerTypeClient.java
```diff
@@ -0,0 +1,64 @@
...
+import org.candlepin.ApiException;
+import org.candlepin.dto.api.v1.ConsumerTypeDTO;
+import org.candlepin.resource.ConsumerTypeApi;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class ConsumerTypeClient extends ConsumerTypeApi implements SpecTestClient {
+
+    Set<String> createdConsumerTypeIds = new HashSet<>();
```
Januson, line 28: Same as ConsumerClient. We should drop cleanup for now.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/JobsClient.java
```diff
@@ -0,0 +1,128 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class JobsClient extends JobsApi implements SpecTestClient {
+
+    /** The default amount of time we should wait for a job to finish processing **/
+    private static final long DEFAULT_TIMEOUT_MILLISECONDS = 60000;
+
+    private Set<String> jobIds = new HashSet<>();
```
Januson, line 31: Same as ConsumerClient. We should drop cleanup for now.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/SpecTestFixture.java
```diff
@@ -50,7 +50,13 @@ protected <O> O getClient(Class<O> outputClass) {
     private <O> O createClient(Class<O> clientClass)  {
         switch (clientClass.getName()) {
             case "org.candlepin.spec.bootstrap.client.OwnerClient":
-                return (O) new OwnerClient(apiClientFactory.createAdminClient());
+                return (O) new OwnerClient(apiClientFactory.createClient());
```
Januson, line 53: Please add these to the ApiClients.

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
Januson, line 46: Please use @SpecTest annotation.

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @BeforeEach
+    private void beforeEach() throws ApiException {
+        consumerClient = getClient(ConsumerClient.class);
+        ownerClient = getClient(OwnerClient.class);
+        jobsClient = getClient(JobsClient.class);
+        consumerTypeClient = getClient(ConsumerTypeClient.class);
+
+        if (owner == null) {
+            owner = ownerClient.createOwner(new OwnerDTOBuilder().build());
```
Januson, line 68: Take a look at Owners class. It provides factory methods for randomized owner instances.

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
```diff
@@ -0,0 +1,124 @@
...
+        AsyncJobStatusDTO status = jobsClient.scheduleJob(JOB_KEY, null);
+        boolean isComplete = jobsClient.waitForJobToComplete(status.getId());
+        assertTrue(isComplete);
+
+        // Verify that the inactive consumer has been deleted.
+        final String inactiveConsumer1Uuid = inactiveConsumer.getUuid();
+        ApiException thrown1 = Assertions.assertThrows(ApiException.class, () -> {
+            consumerClient.getConsumer(inactiveConsumer1Uuid);
+        });
+        assertEquals(410, thrown1.getCode());
```
Januson, line 103: Take a look at StatusCodeAssertions a collection of status code related assetions.

Januson (changes requested):

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
JoshAlbrecht23: Created the following PR for adding the client objects: https://github.com/candlepin/candlepin/pull/3407/

src/main/java/org/candlepin/model/ConsumerCurator.java
JoshAlbrecht23: So, instead of using the Transactional annotation on a method in InactiveConsumerCleanerJob that calls the delete method in the curator, I used the org.candlepin.util.Transactional wrapper and that worked!

```
      Transactional<Integer> transaction = this.consumerCurator.transactional((args) -> {
          if (args == null || args.length != 1) {
              throw new IllegalArgumentException("Unexpected value received for arguments: " + args);
          }
      
          List<String> consumerIds = (List<String>) args[0];
          return consumerCurator.deleteInactiveConsumers(consumerIds);
      });
```

Ceiu: Hrmm... looks like any bulk update or delete op at the Hibernate/JPA layer requires a transaction. That's a bit unfortunate/annoying, as it kinda violates our general design goals here. In the case where a curator method is doing a flush and the failure is a result of that, that's definitely something we should be dealing with via better transaction management or moving flush control out. But this case, I'm uncertain. I would like our curator methods to be callable from any context when they're not doing weird things.

Thoughts?

JoshAlbrecht23: @Ceiu  I see what you are saying! I almost think that we need to break up the deleteInactiveConsumers method in the ConsumerCurator class into individual methods that do one thing, and for the bulk update/delete operations we just build the Query object and return it. Then have a deleteInactiveConsumers method in a Manager (org.candlepin.controller package) class that handles the transaction and calling each individual curator method and invoking the updates/deletes.

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
JoshAlbrecht23: Closed the PR as we can use the *API.java classes and only create *Client.java classes for rare cases.

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
JoshAlbrecht23: Updated!

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
JoshAlbrecht23: Awesome! Used Owners.random()

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
JoshAlbrecht23: Updated and added annotation

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ConsumerTypeClient.java
JoshAlbrecht23: Removed *Client.java classes and using the *API.java classes.

api/candlepin-api-spec.yaml
JoshAlbrecht23: Awesome! I removed the request body and the endpoint works as expected with that version rolled back

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1001,6 +1006,180 @@ else if (systemUuid != null) {
...
+     * @return the number of consumer that were deleted.
+     */
+    public int deleteConsumers(Collection<String> consumerIds) {
+        if (consumerIds == null || consumerIds.isEmpty()) {
+            return 0;
+        }
+
+        Collection<Consumer> consumersToDelete = getConsumers(consumerIds);
+        super.bulkDelete(consumersToDelete);
+        return consumersToDelete.size();
```
JoshAlbrecht23, line 1140: With the spec test I realized that deleting all of the consumers with a single HQL statement based on the consumer ids causes FK constraint violations for the following FKs for Consumer.

This isn't ideal as we have to load all the consumers in memory and individually delete each Consumer with org.hibernate.Session.delete() so that the cascading cleans up all of the FK constraints. I wasn't able to find another way to handle this with Hibernate, but I could be missing something. The alternative is to have methods for each constraint that deletes them before deleting the actual consumer rows. 

**FK Constraints for Consumer**
1. facts
2. environmentIds
3. activationKeys
4. hypervisorId
5. capabilities
6. guestIds
7. installedProducts
8. keyPairData
9. Entitlements

spec-tests/src/test/java/org/candlepin/spec/InactiveConsumerCleanerJobTest.java
JoshAlbrecht23: That makes sense! Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/SpecTestFixture.java
JoshAlbrecht23: Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/ConsumerDTOBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+        this.serviceType = serviceType;
+        return this;
+    }
+
+    public ConsumerDTOBuilder setEnvironments(List<EnvironmentDTO> environments) {
+        this.environments = environments;
+        return this;
+    }
+
+    public ConsumerDTO build() {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Removed this ConsumerDTOBuilder class and added to the builder in the Consumers class

src/main/java/org/candlepin/model/ConsumerCurator.java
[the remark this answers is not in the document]
JoshAlbrecht23: Oops, yeah I missed this one. I removed the delete statement based on the inactive consumer ids because of the constraint violations. I added a comment in regards to that!

src/main/java/org/candlepin/model/ConsumerCurator.java
nikosmoum: I'm not sure I understand the problem. Wasn't the delete in a transaction before anyway, since the @Transactional annotation was on it? What is the difference with using the annotation and the transactional block in this case?

Also, I would agree that conceptually the deleteInactiveConsumers method shouldn't be in a curator, but in the job class itself, which then calls into the individual curator methods (the deletedConsumersStatement query would have to be extracted into it's own curator method in that case)

spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobTest.java
```diff
@@ -0,0 +1,157 @@
...
+        consumerWithEntitlement = consumerApi
+            .createConsumer(consumerWithEntitlement, null, owner.getKey(), null, false);
+        activeConsumer = consumerApi.createConsumer(activeConsumer, null, owner.getKey(), null, false);
+        activeConsumer2 = consumerApi.createConsumer(activeConsumer2, null, owner.getKey(), null, false);
+
+        createProductAndPoolForConsumer(consumerWithEntitlement);
+
+        AsyncJobStatusDTO status = jobsApi.scheduleJob(JOB_KEY);
+        boolean isComplete = JobUtil.waitForJobToComplete(status.getId(), jobsApi);
+        assertTrue(isComplete);
```
nikosmoum, line 121: Can we also assert that the job's status is `FINISHED` (because it could complete with a different state, like FAILED)?

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/JobUtil.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param jobId - Id of the job to wait for.
+     * @param timeout - max duration in milliseconds to wait for the job to finish.
+     * @param api - api used to determine if job has completed.
+     * @return true if the job has finished processing successfully or false if the job
+     *         has failed or exceeded the provided timeout duration.
+     * @throws ApiException
+     * @throws InterruptedException
+     */
+    public static boolean waitForJobToComplete(String jobId, long timeout, JobsApi api)
```
nikosmoum, line 57: Returning a boolean is a bit cumbersome here. When using this method we want to be able to assert on AsyncJobStatusDTO in general (the status field mostly, but also the rest), so it would be better if this method returns the AsyncJobStatusDTO itself. And the while loop below stops when either we've entered one of the [4 terminal states](https://github.com/candlepin/candlepin/blob/bb8d37d50c11d1eaa92751e8dabd741bc420d38e/src/main/java/org/candlepin/model/AsyncJobStatus.java#L77-L83), or the timeout has elapsed

nikosmoum (changes requested):
A few more things to iron out, but we're close

src/main/java/org/candlepin/model/ConsumerCurator.java
nikosmoum: Ugh, I should have seen this during design :/
The third, ideal alternative would be to add `ON DELETE CASCADE` on the foreign keys in the db themselves, which we have a card for in the backlog. This feature is where it would be ideal to have that done, but I am conflicted on doing this now because it would slow this task down.

Januson: We should investigate if CASCADE even is a viable solution. From what I've seen it suffers from the same performance issues as deleting with hibernate. If it really is just as slow we would have to implement bulk delete of all these separate items ourselves of live with hibernate solution.

src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: Agreed. I'm not too fond of spinning up a controller/manager class for one-offs, but the method could definitely be paired down as @nikosmoum described. The curators are intended to be minimalist classes which perform database operations, with any surrounding logic being limited to data shaping. We don't always get it right, but that's the goal.

As far as the exact problem, I'm not *entirely* sure -- looks like it could be a new restriction as of Hibernate 5.3? I took the exception Josh hit and did some googling. Many of the results don't really go over the why, and are just people telling others to slap the annotation on the method in question and move on; which pains me more than I can express in words. I didn't see anything clearly stated in the Hibernate user guide regarding bulk operations requiring a transaction, so there might be something else at play here worth investigating.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Consumers.java
```diff
@@ -58,12 +61,20 @@ public Builder withOwner(OwnerDTO owner) {
             return this;
         }
 
+        public Builder withLastCheckin(OffsetDateTime lastCheckin) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Added the withType method to the builder

spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobTest.java
```diff
@@ -0,0 +1,157 @@
...
+        if (owner == null) {
+            owner = ownerApi.createOwner(Owners.random());
+        }
+
+        if (consumerType == null) {
+            ConsumerTypeDTO newType = new ConsumerTypeDTO();
+            newType.setManifest(false);
+            newType.setLabel(StringUtil.random("Test Type"));
+            consumerType = consumerTypeApi.createConsumerType(newType);
+        }
```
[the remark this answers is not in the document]
JoshAlbrecht23: Awesome! I wasn't aware that we were populating pre-built types in the db. Updated!

spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+        createProductAndPoolForConsumer(consumerWithEntitlement);
+
+        AsyncJobStatusDTO status = jobsApi.scheduleJob(JOB_KEY);
+        boolean isComplete = JobUtil.waitForJobToComplete(status.getId(), jobsApi);
+        assertTrue(isComplete);
+
+        // Verify that the inactive consumer has been deleted.
+        final String inactiveConsumer1Uuid = inactiveConsumer.getUuid();
+        assertGone(() -> consumerApi.getConsumer(inactiveConsumer1Uuid));
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good idea! I made a call to the deleted_consumers api to verify that the correct inactive consumer is returned.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/JobUtil.java
JoshAlbrecht23: Yeah, definitely makes more sense to return the AsyncJobStatusDTO.

src/main/java/org/candlepin/model/ConsumerCurator.java
JoshAlbrecht23: Adding the `ON DELETE CASCADE` to the foreign keys makes a lot of sense. I added the change in a new change set 20220613000000-add_delete_cascade_to_consumer_fks.

spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobTest.java
```diff
@@ -0,0 +1,171 @@
...
+        String jobId = jobsApi.scheduleJob(JOB_KEY).getId();
+        AsyncJobStatusDTO status = JobUtil.waitForJobToComplete(jobId, jobsApi);
+        assertEquals("FINISHED", status.getState());
+
+        // Verify that the inactive consumer has been deleted.
+        final String inactiveConsumer1Uuid = inactiveConsumer.getUuid();
+        assertGone(() -> consumerApi.getConsumer(inactiveConsumer1Uuid));
+        // Verify that the inactive consumer has been moved to the deleted_consumers table.
+        List<DeletedConsumerDTO> deletedConsumers = consumerApi
+            .listByDate(inactiveConsumer.getCreated().toString());
```
nikosmoum, line 131: I was very confused when I saw `consumerApi.listByDate` returning a list of `DeletedConsumerDTO`. It looks like the /deleted_consumers API gets bundled within the ConsumerApi class on the client side, which is very unintuitive/misleading. After a bit of digging, I found out the decision to roll APIs into the same client class happens based on the tags field we set on the spec yaml, and we happened to [set it to 'Consumer'](https://github.com/candlepin/candlepin/blob/5fa273be2319bf61008162428cef9e661c480d62/api/candlepin-api-spec.yaml#L2841-L2845) (along with all the /consumer endpoints). I just tested it and if we change the tag to `DeletedConsumer`, then we'll get this API rolled into it's own `DeletedConsumerApi` client class

src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: I'd also like to have more data on the performance implications, but my general stance on them are to use them within reason.

I also agree that object linkage and DB triggers or triggered behavior needs to be limited to logical objects rather than just done on every link. In this case, that means if we delete a consumer, it also makes sense to delete the things that make up the consumer's children (facts, caps, tags, etc.), but NOT where other objects point to it. In those cases we need to make a decision on a case-by-case basis depending on how critical it is to both track when such a thing goes missing (i.e. pool -> entitlement). Setting a field to null rather than a full cascade might also make more sense in some instances.

nikosmoum: Since I see at least 2 people have concerns about performance, and also about how exactly we should go about it (cascade delete vs setting to null), then I think we should leave the cascade as a separate investigation, and go with bulk deletes of all these separate tables on this task.

JoshAlbrecht23: It makes sense that we should only adding cascades on the fields that are owned by the consumers and I updated the change set to only include the FKs from the cp_consumer_content_tags, cp_consumer_facts, cp_consumer_capability, and  cp_consumer_guests tables.

I also did some performance testing by creating 100k inactive consumers and 100k active consumers, and then performing the InactiveConsumerCleanerJob with the FKs having the cascade deletes. The average time to delete a batch of 1000 inactive consumers was 72.71649 ms and the total time to complete the job was 13266 ms.

I was also curious to see what the performance for using the Hibernate deletion without having the cascade on delete on the FKs and it is pretty bad and was taking over 10,000 ms to delete a batch of 1000 inactive consumers on the same data set. I didn't finish the job as it was taking a long time!

nikosmoum: While not fully a representative perf test (there would have to be ~millions of rows in all related tables to say it is), given that I see we already have cascades for some children tables of consumer (cp_consumer_environments, cp_consumer_activation_key, cp_consumer_hypervisor), I'd say let's go for it.

The only thing is that we need to add 2 more cascades for the cp_consumer_guests_attributes (child table of  cp_consumer_guests) and cp_installed_products tables (child table of cp_consumer).

nikosmoum:
There seems to be a unit test failing also

src/main/java/org/candlepin/model/ConsumerCurator.java
JoshAlbrecht23: Added the cascades for the FK on cp_consumer_guests_attributes table and the cp_installed_products table.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/JobUtil.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        if (api == null) {
+            throw new IllegalArgumentException("Api cannot be null.");
+        }
+
+        long startTime = System.currentTimeMillis();
+        long elapsedMilliseconds = 0L;
+        AsyncJobStatusDTO status = null;
+        while (Long.compare(elapsedMilliseconds, timeout) <= 0) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/JobUtil.java
```diff
@@ -0,0 +1,92 @@
...
+ * in this software or its documentation.
+ */
+
+package org.candlepin.spec.bootstrap.data.util;
+
+import org.candlepin.ApiException;
+import org.candlepin.dto.api.v1.AsyncJobStatusDTO;
+import org.candlepin.resource.JobsApi;
+
+public final class JobUtil {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good point, moved it to the JobClient class which extends the JobsApi class.

nikosmoum (approved):
Looks good to me. I'll let others re-review and merge

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/JobsClient.java
```diff
@@ -0,0 +1,92 @@
...
+        long startTime = System.currentTimeMillis();
+        long elapsedMilliseconds = 0L;
+        AsyncJobStatusDTO status = null;
+        while (elapsedMilliseconds <= timeout) {
+            status = this.getJobStatus(jobId);
+            if (status == null) {
+                return status;
+            }
+
+            if (status.getState().equals("FINISHED") ||
```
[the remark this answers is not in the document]
JoshAlbrecht23: I like this a lot more than what I had! Updated to use the set of terminal states.

Januson (approved):

PR #3347, opened 2022-05-25
ENT-4812: Add dependabot configuration

Januson:
- Moved dependency versions to dependencies.gradle
- Moved plugin versions to dependencies.gradle
- Added plugins in buildSrc to allow applying them without version


Ceiu:
Looks good to me. I have some concerns about having references to our deps in a different file than our actual dependency definitions, but it's probably just me overthinking things.

Going to leave this up for a second set of eyes due to my inexperience with all things Gradle.

Ceiu (approved):

PR #3349, opened 2022-05-30
2076914: Translate few strings.

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2076914
  - The BZ was opened against sub-man, but we will be able to
    verify the fix only using sub-man or cockpit plugin for
    sub-man
* Card ID: ENT-4943


src/main/java/org/candlepin/resteasy/filter/AuthenticationFilter.java
```diff
@@ -164,10 +168,12 @@ public void filter(ContainerRequestContext requestContext)
                 principal = new NoAuthPrincipal();
             }
             else if (!config.getBoolean(ConfigProperties.AUTH_OVER_HTTP) && !request.isSecure()) {
-                throw new BadRequestException("Please use SSL when accessing protected resources");
+                throw new BadRequestException(
+                    i18n.tr("Please use SSL when accessing protected resources"));
```
nikosmoum, line 172: This doesn't have to be in a new line; there's enough space on the previous one I believe

src/main/java/org/candlepin/resteasy/filter/AuthenticationFilter.java
```diff
@@ -164,10 +168,12 @@ public void filter(ContainerRequestContext requestContext)
...
             }
             else {
-                throw new NotAuthorizedException("Invalid credentials.");
+                throw new NotAuthorizedException(
+                    i18n.tr("Invalid credentials."));
```
nikosmoum, line 176: same here; could be on the previous one

nikosmoum (changes requested):
There's some compilation errors in unit tests and some checkstyle issue based on jenkins, and I have one minor request with formatting in the inline comments, but the gist of the change itself looks good

src/main/java/org/candlepin/resteasy/filter/AuthenticationFilter.java
jirihnidek: Fixed

src/main/java/org/candlepin/resteasy/filter/AuthenticationFilter.java
jirihnidek: Fixed

nikosmoum (approved):

PR #3417, opened 2022-06-13
Add primary owner support [ENT-3647]

wottop:


src/main/java/org/candlepin/resource/UserResource.java
```diff
@@ -196,6 +196,17 @@ public Stream<OwnerDTO> listUserOwners(@Verify(User.class) String username) {
         return null;
     }
 
+
+    /**
+     * Retrieve the primary owner the user can register systems to.
+     */
+    @Override
+    public OwnerDTO getPrimaryOwner(@Verify(User.class) String username) {
```
nikosmoum, line 204: Instead of adding a new API on our side, I was thinking we just roll this change into the existing `listUserOwners` endpoint. It will still be returning a Stream<OwnerDTO>, but we'll always be returning a single owner in that list (the primary owner from `user.getPrimaryOwner()`. 

Satellite supports multiple owners, but they don't actually use this API (they have their own Katello API that handles the fetching of the list of owners), so we wouldn't be breaking them. On the other hand, this API is being used in Hosted, and from previous discussions I remember they already have a way to discern a 'primary' owner (they just didn't have a way to communicate this to us, which is what this PR addresses).

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1341,12 +1341,14 @@ private Owner setupOwner(Principal principal, String ownerKey) {
         if (ownerKey == null && (principal instanceof UserPrincipal)) {
             // check for this cast?
             List<String> ownerKeys = ((UserPrincipal) principal).getOwnerKeys();
-
-            if (ownerKeys.size() != 1) {
+            if (ownerKeys.size() == 0 ||
+                (ownerKeys.size() > 1 &&
+                ((UserPrincipal) principal).getPrimaryOwner() == null)) {
```
nikosmoum, line 1346: This can probably be merged/simplified. We don't need to do a getOwnerKeys. We only need to check if `getPrimaryOwner() == null` and throw the exception, since we consider having a single/primary owner the only valid state

nikosmoum (changes requested):

src/main/java/org/candlepin/resource/ConsumerResource.java
nikosmoum: Actually, nevermind, it's probably better to do what Chris mentioned in his comment

nikosmoum (changes requested):
UserInfo.getPrimaryOwner is needed, and needs to be called in UserAuth and passed into the UserPrincipal() constructor

nikosmoum (approved):

PR #3418, opened 2022-06-13
[4.0] Fixed JDBC packet size overrun in updateOwnerProductReferences

Ceiu:
- Fixed a bug that could occur while updating owner-product
  references during refresh or manifest import in organizations
  with many activation keys
- Removed use of the bulkSQLUpdate native query builder

### DIFF
--- a/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
+++ b/server/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
@@ -1200,6 +1200,13 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
      * number of actual changes made is significant to the caller, the criteria should also include
      * the keySet from the given values.
      *
+     * @deprecated
+     *  The performance benefits this method provides have been surpassed by general improvements to
+     *  both Candlepin itself and Hibernate. While this made sense at the time it was written, it is
+     *  no longer more performant than even a simple approach of firing off several single-row
+     *  updates, and comes with a host of restrictions and gotchas. No new code should be using this
+     *  method, and existing code should be updated to drop use it as it is updated.
+     *
      * @param table
      *  The name of the table to update
      *
@@ -1216,6 +1223,7 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
      * @return
      *  the number of rows updated as a result of this query
      */
+    @Deprecated
     protected int bulkSQLUpdate(String table, String column, Map<Object, Object> values,
         Map<String, Object> criteria) {
 
--- a/server/src/main/java/org/candlepin/model/OwnerProductCurator.java
+++ b/server/src/main/java/org/candlepin/model/OwnerProductCurator.java
@@ -802,49 +802,122 @@ public class OwnerProductCurator extends AbstractHibernateCurator<OwnerProduct> 
             return;
         }
 
+        // TODO:
         // Should we step through the UUID map and verify that it doesn't try to map anything weird,
         // (like a UUID to itself), or define multiple remappings?
 
-        Session session = this.currentSession();
+        this.updateOwnerProductJoinTable(owner, productUuidMap);
+        this.updateOwnerProductPools(owner, productUuidMap);
+        this.updateOwnerProductActivationKeys(owner, productUuidMap);
 
-        Map<String, Object> criteria = new HashMap<>();
-        Map<Object, Object> uuidMap = Map.class.cast(productUuidMap);
-        criteria.put("product_uuid", productUuidMap.keySet());
-        criteria.put("owner_id", owner.getId());
+        // Looks like we don't need to do anything with product certificates, since we generate
+        // them on request. By leaving them alone, they'll be generated as needed and we save some
+        // overhead here.
+    }
 
-        // Owner products
-        int count = this.bulkSQLUpdate(OwnerProduct.DB_TABLE, "product_uuid", uuidMap, criteria);
+    /**
+     * Part of the updateOwnerProductReferences operation; updates the owner to product join table.
+     * See updateOwnerProductReferences for additional details.
+     *
+     * @param owner
+     * @param productUuidMap
+     */
+    private void updateOwnerProductJoinTable(Owner owner, Map<String, String> productUuidMap) {
+        String sql = "UPDATE " + OwnerProduct.DB_TABLE + " SET product_uuid = :updated " +
+            "WHERE product_uuid = :current AND owner_id = :owner_id";
+
+        Query query = this.getEntityManager()
+            .createNativeQuery(sql)
+            .setParameter("owner_id", owner.getId());
+
+        int count = 0;
+        for (Map.Entry<String, String> entry : productUuidMap.entrySet()) {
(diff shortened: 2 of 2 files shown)


Ceiu:
This should be forward ported to newer branches once accepted

JoshAlbrecht23 (approved):

Ceiu:
Initial perf tests are not looking good, but we'll have to see how they compare with a data set as large as I'm using.

Ceiu:
Converted this to a draft so it won't get merged while I'm testing this.

Ceiu:
I'm glad I thought to do some perf testing here. I setup a pile of test data consisting of the following:
- 3 orgs
- 200,000 products per org; 600k total products
- 25,000 pools per org; 75k total pools
- 20,000 activation keys per org; 60k total activation keys
- 3 products per activation key; 180k total activation key products

It should be noted that this data set is intentionally absurd, and is intended to really push the limits of actual database packet limits as well as get a clear timing difference between the algorithms.

With this data, I setup a test to remap 20k products within the org and time the results. The `updateOwnerProductReferences` operation has three distinct steps, targeting each primary table: update cp2_owner_products, update cp_pool, and update cp2_activation_key_products.

The implementation in the PR at the time of writing performed passably up until the activation key bit. Each step took roughly the following, observed over several runs:
- 2-5s to update cp2_owner_products
- 90-120s to update cp_pools
- 45+ minutes updating cp2_activation_key_products (test was terminated around 45m).

This was obviously unacceptable, even for intentionally silly data, but I at least had a baseline I expected the original implementation to beat. Running the same test against it, however, went poorly:
- 4-5 minutes to update cp2_owner_products
- 30+ minutes updating cp_pools (test was terminated around 35 minutes in)

...so before it even got to the part I expected it to beat, it was struggling; and this was struggling after having issues with the original test setup (more on that later). In any event, at this point I abandoned the original implementation and focused on how to improve the new implementation, as the data gathered thus far has shown the simple approach is more than sufficiently performant.

Given that evidence, I focused on removing the IN operator from the activation key update, and trying to pair down the number of rows that need to be updated. By changing the activation key query to pull only keys that actually used the products being updated and mapping the result, this allowed the new implementation to not run fewer update statements, but also allowed for the updates to leverage the table's primary key for very fast lookups.

The resulting implementation brought the update time for the activation key step down from north of 45 minutes to under 4 seconds in this test, for a total time of around 110-130s total execution time:

```
2022-06-17 20:03:54,334 [...] DEBUG org.candlepin.async.tasks.TestJob - RUNNING TEST 1
2022-06-17 20:03:54,335 [...] DEBUG org.candlepin.async.tasks.TestJob - PREPARING TEST DATA
2022-06-17 20:03:54,344 [...] DEBUG org.candlepin.async.tasks.TestJob - STARTING TEST
2022-06-17 20:03:59,778 [...] DEBUG org.candlepin.model.OwnerProductCurator - 20000 owner-product relations updated
2022-06-17 20:05:52,608 [...] DEBUG org.candlepin.model.OwnerProductCurator - 5000 pools updated
2022-06-17 20:05:56,391 [...] DEBUG org.candlepin.model.OwnerProductCurator - 10001 activation keys updated
2022-06-17 20:05:56,586 [...] DEBUG org.candlepin.async.tasks.TestJob - TEST COMPLETE. TOTAL ELLAPSED TIME: 122242ms
2022-06-17 20:05:56,586 [...] DEBUG org.candlepin.async.tasks.TestJob - RUNNING TEST 2
2022-06-17 20:05:56,586 [...] DEBUG org.candlepin.async.tasks.TestJob - PREPARING TEST DATA
2022-06-17 20:05:56,599 [...] DEBUG org.candlepin.async.tasks.TestJob - STARTING TEST
2022-06-17 20:06:01,800 [...] DEBUG org.candlepin.model.OwnerProductCurator - 20000 owner-product relations updated
2022-06-17 20:07:58,544 [...] DEBUG org.candlepin.model.OwnerProductCurator - 5000 pools updated
2022-06-17 20:08:00,800 [...] DEBUG org.candlepin.model.OwnerProductCurator - 10001 activation keys updated
2022-06-17 20:08:00,887 [...] DEBUG org.candlepin.async.tasks.TestJob - TEST COMPLETE. TOTAL ELLAPSED TIME: 124287ms
2022-06-17 20:08:00,888 [...] DEBUG org.candlepin.async.tasks.TestJob - RUNNING TEST 3
2022-06-17 20:08:00,888 [...] DEBUG org.candlepin.async.tasks.TestJob - PREPARING TEST DATA
2022-06-17 20:08:00,896 [...] DEBUG org.candlepin.async.tasks.TestJob - STARTING TEST
2022-06-17 20:08:04,646 [...] DEBUG org.candlepin.model.OwnerProductCurator - 20000 owner-product relations updated
2022-06-17 20:09:53,624 [...] DEBUG org.candlepin.model.OwnerProductCurator - 5000 pools updated
2022-06-17 20:09:55,886 [...] DEBUG org.candlepin.model.OwnerProductCurator - 10001 activation keys updated
2022-06-17 20:09:55,914 [...] DEBUG org.candlepin.async.tasks.TestJob - TEST COMPLETE. TOTAL ELLAPSED TIME: 115018ms
```

While I was mostly happy with this result, I felt we could go a bit further. The owner-product and activation key-product join tables were updating quickly, but cp_pool was still a bit slow. A quick examination of the table showed that we didn't have an index on either column we were using to select rows. After slapping down an index on the product_uuid column and re-running the test, our total execution time for the entire update operation was brought down from around 2 minutes to 10-12 seconds -- easy perf gains. Given 4.0 is mostly frozen, we won't be adding the index in this PR, but the forward ports will include it.

As a final note, the original test setup attempted to perform an update on all 200k products within the org. This was dismissed after noting that the original implementation couldn't even handle it, as it has an upper limit of around 25k products, and the pain of watching seconds tick by waiting for results got to be too much when both versions were taking on the order of hours or days to complete. Though, my curiosity did get the better of me, and I ran a single iteration of the full 200k product test against the newest implementation to see the numbers, and I was mostly pleased at the results:
- ~45s for cp_owner_products
- ~23 minutes for cp_pool (no index on product_uuid)
- ~15s for cp2_activation_key_products
- Total runtime: 1445s (24 minutes and change)

I do want to note that this is a single run, so it is does not provide us enough raw data to make sweeping statements about its performance at this scale. But, we can see from the recorded times, it's following mostly linear time (10x data size, ~10-12x runtime).

One major takeaway from all of this is that with the improvements within Hibernate over the years, and better transaction management in Candlepin itself, the hacky `bulkSQLUpdate` method is no longer needed. Even the naive approach of spamming individual updates is now outperforming it when clustered behind a transaction, and is far safer as our data grows. Going forward, use of the method should be avoided. As such, this update declares the method as deprecated and any new code attempting to use it in the future should be rejected outright.

nikosmoum:
Can you explain a bit what the actual bug/error was? I assume this is different from the other PR for the missing mappings

Ceiu:
@nikosmoum 
In orgs with lots of activation keys, the original logic would generate packets which were larger than MySQL/MariaDB would accept and refresh would be permanently blocked in affected orgs.

nikosmoum (approved):

PR #3423, opened 2022-06-16
Update checkstyle jars and correct issues [ENT-4863]

wottop:
An attempt to use a patch previously submitted against the checkstyle repo solved some of these violations but caused others. In light of the headache that maintaining our own spin of the jar would cause, it seems more prudent to make the few corrections needed in the code. The rule definition in checkstyle.xml is not granular enough to account for the scenarios.

Perhaps this is just a temporary bridge to some other code-style tool, like spotless.


JoshAlbrecht23 (approved):

nikosmoum (approved):
This workaround is really ugly, but it can't be helped for now I guess.

Ceiu:
Did we explore adding a custom check that exists outside the jar, or bumping the issue with the checkstyle folks?

nikosmoum:
I left a comment in that github issue, but I wouldn't expect much of it given it's been open for more than a year already

nikosmoum:
They're response is 'PRs are welcome'

PR #3425, opened 2022-06-21
Set gradle to not fail when no unit tests found in spec module

Januson:


Ceiu:
This looks to address the immediate concern of the 'test' task of the spec-test being invoked when attempting to run unit tests, and failing any unit test run that's invoked with a filter.

PR #3427, opened 2022-06-21
[4.1] Fixed JDBC packet size overrun when updating owner-entity references (ENT-5128)

Ceiu:
- Fixed a bug that could occur while updating owner-product
  references during refresh or manifest import in organizations
  with many activation keys
- Fixed a similar bug that could occur while updating environment
  content references during refresh or manifest import in orgs
  with many environments
- Removed the bulkSQLUpdate and bulkSQLDelete operations from
  AbstractHibernateCurator, as they are unsafe and no longer
  outperform easier/better approaches


src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -141,34 +139,14 @@ public List<Environment> listWithContent(Set<String> contentIds) {
      *  the number of environments deleted
      */
     public int deleteEnvironmentsForOwner(Owner owner) {
-        String jpql = "SELECT e.id FROM Environment e WHERE e.owner.id = :owner_id";
+        String jpql = "DELETE FROM Environment env WHERE env.owner.id = :owner_id";
```
nikosmoum, line 142: Hibernate is taking care of deleting the environment-content data here right?

nikosmoum (commented):
There's some issue where spec tests can't clean up the owners that is related here, but other than that looks good

src/main/java/org/candlepin/model/EnvironmentCurator.java
Ceiu: It should be, yes. If it weren't, we'd get angry FK violations. I suspect it might be limited to what we specify should cascade, but in the case of environment-content, we're cascading everything.

src/main/java/org/candlepin/model/OwnerContentCurator.java
```diff
@@ -428,47 +428,93 @@ public void updateOwnerContentReferences(Owner owner, Map<String, String> conten
...
+            "WHERE content_uuid = :current AND owner_id = :owner_id";
 
-        for (Map<String, String> block : this.partitionMap(contentUuidMap)) {
-            Map<String, Object> criteria = new HashMap<>();
-            Map<Object, Object> uuidMap = Map.class.cast(block);
-            criteria.put("content_uuid", block.keySet());
-            criteria.put("owner_id", owner.getId());
+        Query query = this.getEntityManager()
+            .createNativeQuery(sql)
+            .setParameter("owner_id", owner.getId());
```
JoshAlbrecht23, line 448: Possible NPE here, we may want to check that owner isn't null first

src/main/java/org/candlepin/model/OwnerContentCurator.java
```diff
@@ -428,47 +428,93 @@ public void updateOwnerContentReferences(Owner owner, Map<String, String> conten
...
+        EntityManager entityManager = this.getEntityManager();
+
+        String jpql = "SELECT env.id, content.uuid FROM EnvironmentContent ec " +
+            "JOIN ec.content content " +
+            "JOIN ec.environment env " +
+            "WHERE env.owner.id = :owner_id " +
+            "  AND content.uuid IN (:content_uuids)";
+
+        Query query = entityManager.createQuery(jpql)
+            .setParameter("owner_id", owner.getId());
```
JoshAlbrecht23, line 477: Possible NPE here, we may want to check that owner isn't null first. It looks like this may also need to be done for the methods: 
updateOwnerProductJoinTable
updateOwnerProductPools
updateOwnerProductActivationKeys

JoshAlbrecht23 (approved):

nikosmoum (approved):

PR #3435, opened 2022-06-22
[4.0] Updated diagnostic tooling for Candlepin 4.0 model changes

Ceiu:
- Updated the org migrator to support pulling all products referenced
  by a given org, rather than just those explicitly mapped in the
  owner-products table. This allows exporting orgs that may not be in
  a valid state with respect to the owner-products or owner-content
  mappings
- Updated cp_connectors to add convenient support of the iterables
  for use with the IN operator
- Updated the upstream_subscription_injector script to support JSON
  files containing products rather than subscriptions
- Renamed upstream_subscription_injector to upstream_data_injector
- HostedTestResource now optionally creates children entities when
  creating or updating subscriptions or products depending on the
  value of the "create_children" query parameter
- Removed several unused or extraneous functions from the deprecated
  block of code in hostedtest_api.rb


Ceiu:
This needs to be forward ported to 4.1 and master upon acceptance.

Januson (commented):
Changes look good, but there was a failed spec test in `import_update_spec` that might be related to the changes.

Ceiu:
Looks like it was a weird one-off. Rerunning the tests works fine and I am unable to reproduce it locally

Januson (approved):

PR #3436, opened 2022-06-23
[M] Update diagnostic tooling for Candlepin 4.x model changes

Ceiu:
- Updated the org migrator to support pulling all products referenced
  by a given org, rather than just those explicitly mapped in the
  owner-products table. This allows exporting orgs that may not be in
  a valid state with respect to the owner-products or owner-content
  mappings
- Updated cp_connectors to add convenient support of the iterables
  for use with the IN operator
- Updated the upstream_subscription_injector script to support JSON
  files containing products rather than subscriptions
- Renamed upstream_subscription_injector to upstream_data_injector
- HostedTestResource now optionally creates children entities when
  creating or updating subscriptions or products depending on the
  value of the "create_children" query parameter
- Removed several unused or extraneous functions from the deprecated
  block of code in hostedtest_api.rb


Ceiu:
This should be backported to 4.1 upon acceptance. It includes changes not present in the 4.0 version of this branch pertaining to multi-environment changes and the OpenAPI changes.

client/ruby/hostedtest_api.rb
```diff
@@ -330,7 +306,10 @@ def create_upstream_subscription(subscription_id, owner_key, params = {})
       raise "Attempting to create a subscription without a product"
     end
 
-    return @cp.post('hostedtest/subscriptions', {}, subscription)
+    # TODO: We're defaulting to creating children on creation to maintain existing behavior
+    # form ancient tests. When the hosted tests are ported to Java, this behavior should not
```
JoshAlbrecht23, line 310: Typo: "form" should be "from" and the same needs to be changed on line 341

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
```diff
@@ -192,18 +238,18 @@ public OwnerInfo createOwner(OwnerDTO owner) {
...
 
-
-
     /**
      * Creates a new subscription from the subscription JSON provided. Any UUID
      * provided in the JSON will be ignored when creating the new subscription.
      *
-     * @param subscriptionDTO
+     * @param subscription
      *  A Subscription object built from the JSON provided in the request
```
JoshAlbrecht23, line 253: Missing the createChildren parameter

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
```diff
@@ -266,7 +326,7 @@ public SubscriptionInfo getSubscription(@PathParam("subscription_id") String sub
      * Updates the specified subscription with the provided subscription data.
      *
      * @param subscriptionId the ID of the subscription to update
-     * @param subscriptionDTO
+     * @param subscription
```
JoshAlbrecht23, line 329: Missing createChildren parameter

JoshAlbrecht23 (changes requested):

client/ruby/hostedtest_api.rb
Ceiu: fixed

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
Ceiu: Fixed both; though my desire to just remove all of it from this class is growing.

Ceiu:
Something's busted here. Investigating...

Ceiu:
Okay, sorted. A couple of the new queries were still using cursors without 'with'.

PR #3447, opened 2022-07-01
Remove unneeded rpm dependencies from container image

nikosmoum:
- Remove postgres and mariadb dependencies, since they are run as
  separate containers anyway.
- Use postgresql jdbc from the java deps instead of the rpm.
- Remove a number of unneeded rpm dependencies.
- Use pg_isready from postgres 12 instead of 9.6.


nikosmoum:
Note: I already pushed the container image with these changes to quay, in order to test them with jenkins on this PR

JoshAlbrecht23 (approved):

PR #3448, opened 2022-07-01
Remove status rspec test since it is ported to Java

nikosmoum:


nikosmoum:
The failures here (specifically the `No JDBC jar file available for PostgreSQL`) is because I pushed a new container image to quay with the changes for #3447 for that PR to get tested properly. So this PR will have to get merged after it

Januson (approved):

PR #3454, opened 2022-07-11
[4.1] Added support for skipping pool creation while processing test data (ENT-4849)

Ceiu:
- Marketing products tagged with the "skip_pools" attribute will
  now be skipped during subscription/pool creation


Ceiu:
For testers:
Test data should no longer create pools for awesome-os-basic-vdc or the two dev-sku-products.

JoshAlbrecht23 (approved):

PR #3459, opened 2022-07-13
[M] Migrate product_resource_spec to java (ENT-5195)

JoshAlbrecht23:
- Moved the product_resource_spec.rb to ProductResourceSpecTest.java
- Removed product_resource_spec.rb
- Updated GET /products/subscriptions return type to array in api spec yaml


api/candlepin-api-spec.yaml
```diff
@@ -7166,7 +7166,9 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/AsyncJobStatusDTO'
+                type: array
```
JoshAlbrecht23, line 7169: I was getting an issue with the generated API class failing to deserialize because Candlepin is returning a list and the generated API is expecting a single item json.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/SpecTestFixture.java
```diff
@@ -41,6 +45,19 @@ protected <O> O getClient(Class<O> outputClass) {
...
+     *
+     * @return true if Candlepin is in standalone mode.
+     * @throws ApiException when unable to retrieve the status.
+     */
+    protected boolean isStandalone() throws ApiException {
+        ApiClient client = ApiClients.admin();
+        StatusApi statusApi = client.status();
+        StatusDTO status = statusApi.status();
+        return status.getStandalone();
+    }
```
JoshAlbrecht23, line 59: I'm not sure if this is the best place for this method. I'm assuming that there are other spec tests that will be dictated on whether or not Candlepin is running in standalone mode or not.

nikosmoum: Looks like @Januson also wrote this as two annotations we can set on methods/classes: @OnlyInHosted and @OnlyInStandalone. Perhaps we keep that?

JoshAlbrecht23: Yeah, I like the use of the annotations a lot more!

Januson: This PR will be probably merged sooner that the one for owner spec. Once it's merged I'll rebase and add the annotation.

spec-tests/src/test/java/org/candlepin/spec/products/ProductResourceSpecTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+    @Test
+    @DisplayName("throws exception on refresh with no products")
+    public void refreshWithNoProductsException() throws Exception {
+        assertBadRequest(() -> productsApi.refreshPoolsForProducts(new ArrayList<>(), false));
+    }
+
+    @Test
+    @DisplayName("censors owner information on owner-agnostic retrieval")
+    public void ownerAgnosticeRetrievalCensorship() throws Exception {
```
nikosmoum, line 156: It seems to me that the purpose of this test has been lost in time. Looking at the git blame over time, this test used to make sure that fetching a product, the ProductDTO would only show the owner that the user/client that makes the request has access to. But ProductDTO hasn't had the owner as part of the DTO for a long time, so there is no point to this test now. Let's just remove it

JoshAlbrecht23: Removed

spec-tests/src/test/java/org/candlepin/spec/products/ProductResourceSpecTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+    @BeforeAll
+    public void beforeAll() throws Exception {
+        ApiClient client = ApiClients.admin();
+        productsApi = client.products();
+        ownerApi = client.owners();
+        ownerProductApi = client.ownerProducts();
+        jobsClient = getClient(JobsClient.class);
+    }
+
```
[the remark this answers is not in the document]
JoshAlbrecht23: Added

spec-tests/src/test/java/org/candlepin/spec/products/ProductResourceSpecTest.java
```diff
@@ -0,0 +1,241 @@
...
+    private Map<OwnerDTO, Collection<ProductDTO>>  setupOrgProductsAndPools(Collection<OwnerDTO> owners)
+        throws ApiException {
+        Map<OwnerDTO, Collection<ProductDTO>> ownerToProducts = new HashMap<>();
+        for (OwnerDTO owner : owners) {
+            ProductDTO product1 = createProduct(owner.getKey(), null, null);
+            ProductDTO product2 = createProduct(owner.getKey(), null, null);
+            ProductDTO product3 = createProduct(owner.getKey(), null, null);
+            ProductDTO providedProduct = createProduct(owner.getKey(), null, Arrays.asList(product1));
+            ProductDTO derivedProvidedproduct
+                = createProduct(owner.getKey(), providedProduct, Arrays.asList(product2, product3));
```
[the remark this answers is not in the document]
JoshAlbrecht23: I ended up reworking the setupOrgProductsAndPools method to match what is done in the original ruby method, and included the original names.

spec-tests/src/test/java/org/candlepin/spec/products/ProductResourceSpecTest.java
```diff
@@ -0,0 +1,241 @@
...
+        ApiClient client = ApiClients.admin();
+        productsApi = client.products();
+        ownerApi = client.owners();
+        ownerProductApi = client.ownerProducts();
+        jobsClient = getClient(JobsClient.class);
+    }
+
+    @Test
+    @DisplayName("retrieves owners by product")
+    public void retrieveOwnersByProduct() throws Exception {
```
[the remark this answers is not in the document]
JoshAlbrecht23: I see. I definitely missed that. To make things easier I reworked the setupOrgProductsAndPools method and used all the same variable names and logic as the original ruby file and then used the same logic for the tests.

spec-tests/src/test/java/org/candlepin/spec/products/ProductResourceSpecTest.java
```diff
@@ -0,0 +1,241 @@
...
+            AsyncJobStatusDTO scheduledJobStatus = scheduledJobStatuses.get(0);
+            verifyRefreshPoolsForProducts(scheduledJobStatus);
+
+            productIdFromAllOwners.add(productIds.get(0));
+        }
+
+        // Verify we refresh pools for each owner
+        List<AsyncJobStatusDTO> scheduledJobStatus
+            = productsApi.refreshPoolsForProducts(productIdFromAllOwners, true);
+        assertEquals(3, scheduledJobStatus.size());
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/SpecTestFixture.java
JoshAlbrecht23: @Januson  Awesome. I left a TODO comment on the test and commented out the test annotation.

PR #3460, opened 2022-07-14
ENT-4612: Migrate owner_resource_spec to Java (Part 1)

Januson:
- Migrate the first half of owner_resource tests to Java
- Add annotations for skipping tests based on candlepin mode


Ceiu (approved):
Seems fine. Addresses a number of concerns I've mentioned in my spec test migration notes, and the only things I really saw here that I didn't like are design choices that aren't outliers for the Java spec tests on the whole.

PR #3461, opened 2022-07-14
ENT-4809: Enable DTO validation with resteasy-validator

bakajstep:


api/candlepin-api-spec.yaml
```diff
@@ -8687,16 +8687,11 @@ components:
       allOf:
         - $ref: '#/components/schemas/TimestampedEntity'
         - type: object
-          required:
-            - name
-            - contentLabel
```
nikosmoum, line 8692: Why were the 'required' and 'minLength' attributes removed?

spec/activation_key_spec.rb
```diff
@@ -323,6 +323,7 @@ def hash_diff(h1, h2)
       @cp.update_activation_key(@activation_key)
     }.should raise_exception(RestClient::BadRequest)
 
+    #TODO Here should be raised exception, but validator evaluate it like it was OK
```
nikosmoum, line 326: My guess would be because you removed the `minLength: 1` from the ContentOverrideDTO.name field in the yaml file

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -25,91 +30,18 @@
...
-import org.candlepin.auth.Access;
-import org.candlepin.auth.ActivationKeyPrincipal;
-import org.candlepin.auth.ConsumerPrincipal;
-import org.candlepin.auth.Principal;
-import org.candlepin.auth.SecurityHole;
-import org.candlepin.auth.SubResource;
-import org.candlepin.auth.UpdateConsumerCheckIn;
-import org.candlepin.auth.UserPrincipal;
-import org.candlepin.auth.Verify;
+import org.candlepin.auth.*;
```
nikosmoum, line 33: We don't use star imports (except for static imports in unit test classes), so the checkstyle stage in Jenkins will fail here (see https://github.com/candlepin/candlepin/blob/358b5d0f790265a36591811b1c6f26fada64d122/config/checkstyle/checkstyle.xml#L185-L187)
Same goes for all of the files in the PR

src/main/java/org/candlepin/resource/validation/DTOValidator.java
```diff
@@ -14,21 +14,12 @@
  */
 package org.candlepin.resource.validation;
 
-import com.google.inject.Inject;
-
+import javax.validation.Validator;
```
nikosmoum, line 17: This will probably fail checkstyle too. The import order config needs javax imports after java imports: https://github.com/candlepin/candlepin/blob/358b5d0f790265a36591811b1c6f26fada64d122/config/checkstyle/checkstyle.xml#L202-L209

nikosmoum (changes requested):
A few things to fix; but the direction is good.
Also, I know this is In Progress, but just to note: the final PR and commit summary should be something descriptive (but short), such as: `Enable DTO validation with resteasy-validator`

src/main/java/org/candlepin/resource/ConsumerResource.java
bakajstep: yes, Ondrej has already pointed this out to me so I fixed it in the last commit

api/candlepin-api-spec.yaml
nikosmoum: I see. I guess this is the problem with applying validation throughout the whole API. You can't control when to validate a DTO on only some endpoints and not others, unless you do manual validation.

If this is a one-off instance, then I guess it's ok to remove the validation annotations from it, but if it needs to be done on multiple DTOs, then I would start thinking whether or not we should leave the DTOValidator.validateConstraints method so that we can do validations manually where we need to, or if we even blanket-applying validation to all endpoints even works for us?

spec/activation_key_spec.rb
nikosmoum: If we're going to remove the required and minLength attributes from the DTO, then we should update this test to stop expecting an exception

api/candlepin-api-spec.yaml
Januson: That is the issue with applying global validations retroactively. It shows us where we did a poor job with manual validation. If one endpoint requires validation while others don't they should not use the same DTO so that the validation is visible in swagger.

Januson: Looking at the jenkins report, there is quite a few similar issues. Most of them seem to be the spec test sending an incomplete request. Like creating a [content without uuid](https://github.com/candlepin/candlepin/blob/master/api/candlepin-api-spec.yaml#L8626).

Januson (changes requested):

spec/activation_key_spec.rb
```diff
@@ -326,16 +326,15 @@ def hash_diff(h1, h2)
...
-    @activation_key['contentOverrides'] = [{:contentLabel => nil, :name => 'value', :value => '123' }]
+    output = @cp.update_activation_key(@activation_key)
+    output['products'].should eq(@activation_key['products'])
+    output['name'].should eq(@activation_key['name'])
 
+    @activation_key['contentOverrides'] = [nil]
     lambda {
       @cp.update_activation_key(@activation_key)
-    }.should raise_exception(RestClient::BadRequest)
+    }.should raise_exception(RestClient::InternalServerError)
```
nikosmoum, line 336: This doesn't look right. We should never expect a 500 error. If this is an error we expect we should be throwing a specific relevant exception (such as 400 Bad Request)

dependencies.gradle
```diff
@@ -101,3 +101,6 @@ libraries["rhino"] = "org.mozilla:rhino:1.7.14"
 libraries["slf4j"] = "org.slf4j:slf4j-api:1.7.36"
 libraries["spotbugs"] = "com.github.spotbugs:spotbugs-annotations:4.7.1"
 libraries["swagger"] = "io.swagger:swagger-annotations:1.6.6"
+libraries["resteasyValidator"] = "org.jboss.resteasy:resteasy-validator-provider:4.4.3.Final"
+libraries["javaxElApi"] = "javax.el:javax.el-api:3.0.0"
```
nikosmoum, line 105: We don't actually need to add the javax.el-api library at all. It looks like hibernate-validator is using the library bundled within Tomcat itself, so let's remove it from both gradle and pom.xml

nikosmoum (changes requested):

dependencies.gradle
bakajstep: done

nikosmoum (commented):
There's some unit tests that need to get fixed, and we need to get all the commits squashed into 1 with a descriptive summary and if needed any more details after leaving an empty line (see [here](https://www.candlepinproject.org/docs/candlepin/developers.html#commit-messages)).

nikosmoum (approved):
We generally try to avoid adding the jira id (ENT-XXXX) to the commit message; we only add the bugzilla id (when the change is related to a bugzilla). If there isn't a bugzilla involved, we only add the jira id on the PR title. But that's just a note for the future; the change itself is good to merge after jenkins passes

PR #3462, opened 2022-07-14
Convert job_schedule_spec and job_status_spec to java spec

wottop:
Note that a scheduler test for import was marked TODO and not created. Waiting for the proper conversion of that functionality which will make this test trivial.

I have included a change to the user client creation because it will become a "master key" if this is not done. We should look at this a bit more.

The paging parameters were added to the job list API yaml so that the values could be passed from the tests. If there is a better way, I'll be happy to change it.


spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
```diff
@@ -0,0 +1,498 @@
...
+    private OwnerApi ownerApi;
+    private OwnerProductApi ownerProductApi;
+    private ConsumerApi consumerApi;
+    private JobsClient jobsClient;
+    private ProductDTO product;
+    private PoolDTO pool;
+
+
+    @BeforeEach
+    public void beforeEach() throws ApiException {
```
Januson, line 76: We should setup clients per test class. There's no need to recreated them for each test.

spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
```diff
@@ -0,0 +1,498 @@
...
+        pool.setEndDate(Instant.now().plus(365, ChronoUnit.DAYS).atOffset(ZoneOffset.UTC));
+        pool = ownerApi.createPool(owner.getKey(), pool);
+    }
+
+    @Test
+    @DisplayName("should find an empty list if the owner key is wrong")
+    public void shouldFindEmptyListOwnerNotExist() throws Exception {
+        List<AsyncJobStatusDTO> jobs = jobsClient.listMatchingJobStatusForOrg("totally-made-up",
+            null, null);
+        assertThat(jobs == null || jobs.size() == 0);
```
Januson, line 108: AssertJ's `assertThat` works a bit differenty. You put what you want to test into `assertThat` and follow with a chain of contextual assertions.

Some examples:
```java
assertThat(jobs).isEmpty();
assertThat(jobs)
    .isNotNull()
    .isNotEmpty();
assertThat(jobs).hasSize(10);
assertThat(id).isEqualTo(anotherId);
```

Januson (changes requested):

spec-tests/src/test/java/org/candlepin/spec/jobs/JobScheduleSpecTest.java
```diff
@@ -0,0 +1,148 @@
...
+    @BeforeEach
+    public void beforeEach() throws ApiException {
+        client = ApiClients.admin();
+    }
+
+    @Test
+    @DisplayName("should not schedule arbitrary tasks")
+    public void shouldNotScheduleArbitraryTasks() throws Exception {
+        try {
+            client.jobs().scheduleJob("totally-made-up");
```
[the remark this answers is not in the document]
wottop: Thanks. I missed that.

wottop:
This will need some changes after https://github.com/candlepin/candlepin/pull/3460 is merged.

spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
```diff
@@ -0,0 +1,480 @@
...
+            null, List.of(product.getId()), null, null, null, true, null, null));
+        assertEquals(consumer.getUuid(), bindStatus.getPrincipal());
+        jobsClient.waitForJobToComplete(bindStatus.getId(), 15);
+    }
+
+    @Test
+    @DisplayName("should allow admin to view any job status")
+    public void shouldAllowAdminToViewJobStatus() throws Exception {
+        AsyncJobStatusDTO jobStatus = ownerApi.healEntire(owner.getKey());
+        jobsClient.waitForJobToComplete(jobStatus.getId(), 15);
```
JoshAlbrecht23, line 161: Should this be 15 seconds or 15 milliseconds?

It looks like this also a question for some of the other tests in this file using the waitForJobToComplete method.

spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
```diff
@@ -0,0 +1,480 @@
...
+        List<OwnerDTO> owners = new ArrayList<>();
+        for (int i = 0; i < totalThreads; i++) {
+            owners.add(ownerApi.createOwner(Owners.random()));
+        }
+        // For each owner create a Thread which refreshes that owner, and saves the job status
+        for (OwnerDTO owner : owners) {
+            Thread t = new Thread(() -> {
+                try {
+                    jobs.add(ownerApi.healEntire(owner.getKey()));
+
```
JoshAlbrecht23, line 184: Extra line

spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
```diff
@@ -0,0 +1,480 @@
...
+    @Test
+    @DisplayName("should allow sorting of jobs")
+    public void shouldAllowSortingOfJobs() throws Exception {
+        AsyncJobStatusDTO job1 = ownerApi.healEntire(owner.getKey());
+        AsyncJobStatusDTO job2 = ownerApi.healEntire(owner.getKey());
+        AsyncJobStatusDTO job3 = ownerApi.healEntire(owner.getKey());
+        List<AsyncJobStatusDTO> jobs = jobsClient.listJobStatuses(null, null, null,
+            Set.of(owner.getKey()), null, null, null, null, null, null, null, "asc", "id");
+        // Since we're not setting the order, we can't guarantee which job we'll get, just that we'll
+        // get exactly one of them.
```
JoshAlbrecht23, line 248: I think this was copied from the shouldAllowPagingOfJobs test

spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    @Test
+    @DisplayName("should allow user to view job status of consumer in managed org")
+    public void shouldAllowUserToViewConsumerJobStatus() throws Exception {
+        ConsumerDTO consumer = new Consumers.Builder()
+            .withOwner(owner)
+            .build();
+        consumer = consumerApi.createConsumer(consumer, user.getUsername(), owner.getKey(), null, false);
+        // need a consumer api client
```
JoshAlbrecht23, line 279: Should this be removed? The only reason why I don't think that we need the comment is that the variable name describes that it is a consumer api client. This also applies to the other methods that have the same comment.

spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
```diff
@@ -0,0 +1,480 @@
...
+        AsyncJobStatusDTO bindStatus = AsyncJobStatusDTO.fromJson(consumerApi.bind(consumer.getUuid(),
+            null, List.of(product.getId()), null, null, null, true, null, null));
+        jobsClient.waitForJobToComplete(bindStatus.getId(), 15);
+
+        try {
+            userClient.jobs().getJobStatus(bindStatus.getId());
+            fail("User should not view job status outside of managed org");
+        }
+        catch (ApiException ae) {
+            System.out.println(ae);
```
JoshAlbrecht23, line 393: Was this accidentally left from troubleshooting?

JoshAlbrecht23 (changes requested):
Some formatting changes and some questions on a couple things

spec-tests/src/test/java/org/candlepin/spec/jobs/JobScheduleSpecTest.java
```diff
@@ -0,0 +1,137 @@
...
+        pool.setEndDate(Instant.now().atOffset(ZoneOffset.UTC).minus(10, ChronoUnit.DAYS));
+        pool = ownerApi.createPool(owner.getKey(), pool);
+
+        // verify pool exists before
+        assertNotNull(poolsApi.getPool(pool.getId(), null, null));
+
+        AsyncJobStatusDTO jobStatus = client.jobs().scheduleJob("ExpiredPoolsCleanupJob");
+        getClient(JobsClient.class).waitForJobToComplete(jobStatus.getId(), 15);
+
+        PoolDTO finalPool = pool;
```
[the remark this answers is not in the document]
wottop: Itellij flags it with "Variable used in lambda expression should be final or effectively final" otherwise.

spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
Januson: Method is renamed, but the annotations is still @BeforeEach

spec-tests/src/test/java/org/candlepin/spec/jobs/JobScheduleSpecTest.java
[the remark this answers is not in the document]
JoshAlbrecht23: Oh I see!! That's my bad.

spec-tests/src/test/resources/settings.properties
```diff
@@ -1,4 +1,4 @@
-spec.test.client.host=localhost
+spec.test.client.host=candlepin-el8.example.com
```
JoshAlbrecht23, line 1: I think this change is okay, but is there a reason why the host name needed to be updated?

JoshAlbrecht23 (changes requested):
Small changes

spec-tests/src/test/resources/settings.properties
wottop: no, I have to change it to run it locally, then remember to flip it back so it will run the tests on Jenkins.

JoshAlbrecht23 (approved):

api/candlepin-api-spec.yaml
```diff
@@ -3809,6 +3809,22 @@ paths:
             type: string
             format: date-time
           description: Filter jobs to those on or before this date
+        - in: query
+          name: page
```
nikosmoum, line 3813: Can we add the `description` and `example` attributes for each paging argument here (same as [here](https://github.com/candlepin/candlepin/blob/e003c4dab79a3dc5d3623f3b72ff8af43a688a8d/api/candlepin-api-spec.yaml#L5615-L5638) I suppose)? It helps with the swagger documentation for users.

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/jobs/JobScheduleSpecTest.java
```diff
@@ -0,0 +1,137 @@
...
+        OwnerDTO owner = ownerApi.createOwner(Owners.random());
+        ProductDTO product = new ProductDTO();
+        product.setId(StringUtil.random("ID"));
+        product.setName(StringUtil.random("Test Product"));
+        product = client.ownerProducts().createProductByOwner(owner.getKey(), product);
+
+        PoolDTO pool = new PoolDTO();
+        pool.setProductId(product.getId());
+        // TODO: Setting these on the pool causes issues in Jenkins.
+        // Will put back after pool tests have been implemented
```
nikosmoum, line 91: pool_resource_spec has been merged, so you can add these now; although I'm not sure how that affects the tests here

spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
```diff
@@ -0,0 +1,446 @@
...
+
+        product = new ProductDTO();
+        product.setId(StringUtil.random("ID"));
+        product.setName(StringUtil.random("Test Product"));
+        product = ownerProductApi.createProductByOwner(owner.getKey(), product);
+
+        pool = new PoolDTO();
+        pool.setProductId(product.getId());
+        // TODO: Setting these on the pool causes issues in Jenkins.
+        // Will put back after pool tests have been implemented
```
nikosmoum, line 92: same here: pool_resource_spec port is merged now

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/jobs/JobScheduleSpecTest.java
nikosmoum: Thinking a bit more on this: why do we even need to set these? It doesn't make a difference for the job spec tests (they already pass). Just setting the quantity should be enough

wottop: I was reproducing the test as written in Ruby. I will rebase and see if the new pool test branch changes anything.

spec-tests/src/test/resources/settings.properties
nikosmoum, line 1: You forgot this in the PR, so tests are failing on jenkins

nikosmoum (approved):

PR #3463, opened 2022-07-21
[M] Migrated content_resource_spec to java (ENT-5197)

JoshAlbrecht23:
Migrated the tests in content_resource_spec.rb to ContentResourceSpecTest.java


spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
```diff
@@ -0,0 +1,92 @@
...
+        for (String certificate : certificates) {
+            // Retrieve the compressed data body
+            certificate = certificate.split("-----BEGIN ENTITLEMENT DATA-----\n")[1];
+            certificate = certificate.split("-----END ENTITLEMENT DATA-----")[0];
+            byte[] compressedBody = fromBase64(certificate.getBytes());
+
+            // Decompress the data
+            Inflater decompressor = new Inflater();
+            decompressor.setInput(compressedBody);
+            byte[] decompressedBody = new byte[1024];
```
JoshAlbrecht23, line 79: I'm not sure how we can know how much to allocate for the decompressed body here

spec-tests/src/test/java/org/candlepin/spec/content/ContentResourceSpecTest.java
```diff
@@ -0,0 +1,286 @@
...
+    @DisplayName("should filter content with mismatched architecture from the product")
+    public void filterContentWithMismatchedArchitectureFromProduct() throws Exception {
+        OwnerDTO owner = ownerApi.createOwner(Owners.random());
+        // Even though this product has no arches specified, and should normally not be filtered out,
+        // the product it belongs to has an architecture that mismatches with the system's,
+        // so we do expect it to get filtered out.
+        ContentDTO content1 = createContent(owner.getKey(), StringUtil.random("cname1"),
+            StringUtil.random("test-content1"), StringUtil.random("clabel1"), StringUtil.random("ctype1"),
+            StringUtil.random("cvendor1"), "/this/is/the/path", null);
+        AttributeDTO attribute1 = new AttributeDTO();
```
Januson, line 175: Check ProductAttributes/Products. I added enum for easier handling of products and attributes.

spec-tests/src/test/java/org/candlepin/spec/content/ContentResourceSpecTest.java
```diff
@@ -0,0 +1,286 @@
...
+        newProduct.setMultiplier(multiplier);
+        if (attribute != null) {
+            newProduct.addAttributesItem(attribute);
+        }
+
+        return ownerProductApi.createProductByOwner(ownerKey, newProduct);
+    }
+
+    private PoolDTO createPool(String ownerKey, ProductDTO product) throws ApiException {
+        PoolDTO pool = new PoolDTO();
```
Januson, line 278: Check Pools.random()

spec-tests/src/test/java/org/candlepin/spec/content/ContentResourceSpecTest.java
```diff
@@ -0,0 +1,286 @@
...
+            StringUtil.random("cvendor1"), "/this/is/the/path", null);
+        AttributeDTO attribute1 = new AttributeDTO();
+        attribute1.setName("arch");
+        attribute1.setValue("x86_64");
+        ProductDTO product1 = createProduct(StringUtil.random("product"), owner.getKey(), null, attribute1);
+        product1 = ownerProductApi.addContent(owner.getKey(), product1.getId(), content1.getId(), true);
+
+        // This content has no arches specified, but the product it belongs to has an arch that
+        // matches with that of the system's, so we do NOT expect it to get filtered out.
+        ContentDTO content2 = createContent(owner.getKey(), StringUtil.random("cname2"),
```
Januson, line 183: Could you add this to Contents.random()? It would be great having a factory method for randomized contents.

Januson (changes requested):
Changes look good. There are just a few places where we could make the client API or the data generation a bit more convenient.

spec-tests/src/test/java/org/candlepin/spec/content/ContentResourceSpecTest.java
```diff
@@ -0,0 +1,286 @@
...
+            .withName(StringUtil.random("consumer-name"))
+            .withType(consumerType)
+            .withOwner(owner)
+            .withFacts(facts)
+            .build();
+
+        consumer = consumerApi.createConsumer(consumer, null, owner.getKey(), null, false);
+        consumerApi.bind(consumer.getUuid(), pool1.getId(), new ArrayList<>(), null, "", "",
+            false, "", new ArrayList<>());
+        consumerApi.bind(consumer.getUuid(), pool2.getId(), new ArrayList<>(), null, "", "",
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good call! Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
```diff
@@ -0,0 +1,92 @@
...
+     * Extracts the decoded and uncompressed bodies of the entitlement certificates.
+     *
+     * @param jsonPayload - the json that contains the encoded and compressed entitlement certificates
+     * @return the entitlement certificate bodies
+     * @throws DataFormatException if unable to decompress the body of the certificate
+     * @throws JsonMappingException if unable to parse the json
+     * @throws JsonProcessingException if unable to parse the json
+     */
+    public static List<JsonNode> extractEntitlementCertificatesFromPayload(Object jsonPayload)
+        throws DataFormatException, JsonMappingException, JsonProcessingException {
```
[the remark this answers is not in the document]
JoshAlbrecht23: I changed to throw an Exception as one of the exceptions did not extend RunTimeException

spec-tests/src/test/java/org/candlepin/spec/content/ContentResourceSpecTest.java
```diff
@@ -0,0 +1,286 @@
...
+        PoolDTO pool1 = createPool(owner.getKey(), product1);
+        PoolDTO pool2 = createPool(owner.getKey(), product2);
+
+        ConsumerTypeDTO consumerType = new ConsumerTypeDTO();
+        consumerType.setManifest(false);
+        consumerType.setLabel("system");
+        Map<String, String> facts = new HashMap<>();
+        facts.put("system.certificate_version", "3.3");
+        facts.put("uname.machine", "ppc64");
+        ConsumerDTO consumer = new Consumers.Builder()
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/content/ContentResourceSpecTest.java
JoshAlbrecht23: Awesome! Updated to use Products and ProductAttributes

spec-tests/src/test/java/org/candlepin/spec/content/ContentResourceSpecTest.java
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/content/ContentResourceSpecTest.java
JoshAlbrecht23: Added Contents class and added random method for generating random contents

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
```diff
@@ -33,7 +35,7 @@ private Products() {
 
     public static ProductDTO random() {
         return new ProductDTO()
-            .id(StringUtil.random("test_id"))
+            .id(String.valueOf(new Random().nextInt(1000)))
```
JoshAlbrecht23, line 38: The product id has to be an number or it causes issues with the certificate encoding.

https://github.com/candlepin/candlepin/blob/master/spec/candlepin_scenarios.rb#L66

spec-tests/src/test/java/org/candlepin/spec/content/ContentResourceSpecTest.java
```diff
@@ -0,0 +1,286 @@
...
+            .withFacts(facts)
+            .build();
+
+        consumer = consumerApi.createConsumer(consumer, null, owner.getKey(), null, false);
+        consumerApi.bind(consumer.getUuid(), pool1.getId(), new ArrayList<>(), null, "", "",
+            false, "", new ArrayList<>());
+        consumerApi.bind(consumer.getUuid(), pool2.getId(), new ArrayList<>(), null, "", "",
+            false, "", new ArrayList<>());
+
+        Object jsonPayload = consumerApi.exportCertificates(consumer.getUuid(), null);
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Extracts the decoded and uncompressed bodies of the entitlement certificates.
+     *
+     * @param jsonPayload - the json that contains the encoded and compressed entitlement certificates
+     * @return the entitlement certificate bodies
+     * @throws DataFormatException if unable to decompress the body of the certificate
+     * @throws JsonMappingException if unable to parse the json
+     * @throws JsonProcessingException if unable to parse the json
+     */
+    public static List<JsonNode> extractEntitlementCertificatesFromPayload(Object jsonPayload)
```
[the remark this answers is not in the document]
Januson: We could theoretically add those DTOs to the spec file. It is part of our API anyway, so we could generate them and link it in a comment as the compressed output.

If not then yes. client.model seems like a good fit.

@nikosmoum @Ceiu @wottop What do you think?

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
```diff
@@ -20,20 +20,24 @@
...
+    private static final int MAX_ID = 10000;
+
     private Products() {
         throw new UnsupportedOperationException();
     }
 
     public static ProductDTO random() {
         return new ProductDTO()
-            .id(StringUtil.random("test_id"))
+            .id(String.valueOf(new Random().nextInt(MAX_ID)))
```
nikosmoum, line 40: This is one area where @Ceiu has also changed, so whichever PR gets merged first, the other one will have to adjust (and I think @Ceiu's change is more complete since it takes care of both engineering and market product id formats)

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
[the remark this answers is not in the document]
nikosmoum: I think using DTOs (either from spec file, or manually crafted), would add an abstraction layer that would make our ability to assert on the cert payload properly hard, or even impossible in some cases (this is the problem that @Ceiu has pointed out before with the general DTOs in responses). I think we should keep using JsonNodes for parsing certificate payloads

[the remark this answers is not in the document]
Januson: @nikosmoum Using DTOs wouldn't prevent us from using json. We could use DTO version for the general case and still be able to fallback on raw json in those rare cases where we need to. The inverse of this is probably what we will end up doing for getting json instead of DTO.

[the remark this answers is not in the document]
Ceiu: Y'all have my core thoughts in mind already, so I'll skip that. So, that aside, especially in light of the cert unification and stuff I'd like to see, I would really like to avoid making the nested objects a critical or important aspect of the certs if we can.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
Ceiu: If my PR happens to get merged first, there's a new utility method for this in `Products.randomSKU()` that may be of value here.

JoshAlbrecht23: Yeah, I think using your new utility method is a better solution!

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
JoshAlbrecht23: Yeah, I agree! I think using the new Products.randomSKU() would make more sense.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
```diff
@@ -0,0 +1,87 @@
...
+        String json = String.valueOf(jsonPayload);
+        if (json == null || json.isEmpty() || json.isBlank()) {
+            return new ArrayList<>();
+        }
+
+        // Retrieves the certificate substrings from the json.
+        List<String> certificates = new ArrayList<>();
+        int certEndIndex = 0;
+        int certStartIndex = json.indexOf("cert=", certEndIndex);
+        while (certStartIndex != -1 && certEndIndex != -1) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: I don't think we will be able to do this because the json does not have quotes on the key and value. When I try to parse as a list of maps I get the following error.

```
com.fasterxml.jackson.databind.JsonMappingException: Unexpected character ('c' (code 99)): was expecting double-quote to start field name
         at [Source: (String)"[{created=2022-07-27T20:43:35+0000, ...
```

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
nikosmoum: the pool_resource_spec PR is merged, so you can rebase and use that now

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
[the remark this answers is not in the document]
nikosmoum: Maybe the best approach is partial DTOs? meaning, instead of adding the whole nested hierarchy, just add the fields we need to assert on (products, content) and not the rest (order, subscription, etc.)?

Also, we'll need a naming convention here. Maybe CertProductDTO/CertContentDTO/etc to differentiate between the other DTOs.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
JoshAlbrecht23: Awesome. I updated to use the new method.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
```diff
@@ -43,7 +43,7 @@ public static ProductDTO random() {
      *  a product DTO with a randomly generated SKU ID and name
      */
     public static ProductDTO randomSKU() {
-        String id = String.valueOf((int) Math.random() * 100000);
+        String id = String.valueOf((int) (Math.random() * 100000));
```
JoshAlbrecht23, line 46: I was getting an id value of 0 and I think the issue was that the double returned from Math.random() was being cast to an integer before multiplying by 100,000 and so it was rounding up to 0 and causing the 0 * 100,000 operation to result in 0.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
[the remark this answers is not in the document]
JoshAlbrecht23: I think that can work, but would these classes only be used for spec-tests and would that make it a DTO? Like you said before we do lose that ability to assert on the payload properly as well. I think we should leave it as a json and have a broader discussion about the DTOs in general and what we should do in cases where asserting against the json is necessary.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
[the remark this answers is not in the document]
JoshAlbrecht23: Actually, after seeing how the ConsumerClient.fetchCertificates() is able to use the ObjectMapper to create a list of CertificateDTOs, I did the same thing and then populated a list of certificates. This is great because I really disliked parsing the String! Thanks for pointing this out!

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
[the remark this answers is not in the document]
nikosmoum: Additionally It just feels like overkill defining DTOs just for the purpose of using them for asserting in spec tests, when the assertions are very simple/exactly the same across all tests (going 2-3 levels in the tree to check the products and content fields)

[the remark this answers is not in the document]
nikosmoum: For the sake of not blocking this PR, let's merge now, and discuss the need for defining DTOs elsewhere

PR #3464, opened 2022-07-21
[M] Ported some pool resource spec tests to Java (ENT-4795)

Ceiu:


spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
```diff
@@ -0,0 +1,673 @@
...
+    private PoolDTO createPool(ApiClient client, OwnerDTO owner) throws ApiException {
+        ProductDTO product = this.createProduct(client, owner);
+        return this.createPool(client, owner, product);
+    }
+
+    private PoolDTO createPool(ApiClient client, OwnerDTO owner, ProductDTO product) throws ApiException {
+        OffsetDateTime startDate = OffsetDateTime.now();
+        OffsetDateTime endDate = startDate.plusYears(1);
+
+        PoolDTO pool = new PoolDTO()
```
Januson, line 138: Check Pools.random()

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
```diff
@@ -0,0 +1,673 @@
...
+        return client.owners().createOwner(Owners.random());
+    }
+
+    private ConsumerDTO createConsumer(ApiClient client, OwnerDTO owner, String name, String type,
+        Map<String, String> facts) throws ApiException {
+
+        ConsumerTypeDTO ctype = new ConsumerTypeDTO()
+            .label(type);
+
+        ConsumerDTO consumer = new ConsumerDTO()
```
Januson, line 95: Check ConsumerTypes/Consumer.random

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
```diff
@@ -0,0 +1,673 @@
...
+        ProductDTO product = new ProductDTO()
+            .id(productId)
+            .name("test_product-" + productId);
+
+        return client.ownerProducts()
+            .createProductByOwner(owner.getKey(), product);
+    }
+
+    // ugh...
+    private AttributeDTO buildAttribute(String key, String value) {
```
Januson, line 123: Check ProductAttributes

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
```diff
@@ -0,0 +1,673 @@
...
+        consumer.putFactsItem("system.certificate_version", "3.3");
+
+        return client.consumers()
+            .register(consumer);
+    }
+
+    private ProductDTO createProduct(ApiClient client, OwnerDTO owner) throws ApiException {
+        String productId = String.valueOf((int) Math.random() * 10000);
+
+        ProductDTO product = new ProductDTO()
```
Januson, line 114: Check Products.random

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
```diff
@@ -0,0 +1,673 @@
...
+    // }
+
+    @Test
+    public void shouldIncludeProductBranding() throws ApiException {
+        ApiClient adminClient = ApiClients.admin();
+        OwnerDTO owner = this.createOwner(adminClient);
+
+        // Impl note: it is not strictly necessary to have the product ID match anything for the
+        // purposes of this test
+        BrandingDTO brand1 = new BrandingDTO()
```
Januson, line 632: Can you add a factory for random brandings?

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
```diff
@@ -0,0 +1,673 @@
...
+     * Generates a random suffix of 7 characters or fewer. Alternative to StringUtil.random which
+     * generates too many characters for fields which have small length limitations.
+     *
+     * @param prefix
+     *  The base string to concatenate to the generated suffix
+     *
+     * @return
+     *  the provided base string/prefix concatenated with a randomly generated suffix
+     */
+    private String randomSuffix(String prefix) {
```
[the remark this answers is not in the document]
Ceiu: Done

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
Ceiu: Ended up having to add some new functions there, but this has been moved.

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
Ceiu: ProductAttributes currently lives as an enum which seems like it's trying to programmatically standardize the attribute names (as we do in CP code), but the way you have to work with it to build attribute DTOs is still clunky.

This method exists because I wanted a quick one-liner that would take a key-value pair and turn it into the object without in-line method chaining. With the ProductAttributes version, I basically end up with the same thing I already have, but with an additional wrapper to unwrap:

```
private AttributeDTO buildAttribute(String key, String value) {
    return new ProductAttributes(key)
        .withValue(value)
        .value();
}
```

There's probably more we could do here to improve that factory.

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
Ceiu: In both the previous version and current version of the consumers factory, the usage is slightly clunky without changing potentially every other use of it out there. Specifically, that NestedOwnerDTO and OwnerDTO aren't at all related makes it real annoying to make the appropriate factory functions that cover our use cases (again, without clobbering other tests).

I've started on this a couple times now, and in both cases, I felt that leaving the manual construction will be cleaner.

ConsumerTypes looks to suffer the same problem as ProductAttributes in that it's more work to build/unwrap the DTO from the enum than it is to just build the DTO.

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
```diff
@@ -0,0 +1,673 @@
...
+  // end
+
+    // @Test
+    // public void includesNullValuesInJSON() {
+        // This test isn't possible given that we don't know if the fields in question were present in the
+        // JSON by the time we can operate on the data via DTO.
+    // }
+
+    @Test
+    public void doesNotLetConsumersViewPoolEntitlements() throws ApiException {
```
nikosmoum, line 197: Can we use the `@DisplayName("")` annotation to specify human readable names for each test case?

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
Ceiu: Done, though I'm not sure how much value it provides.

spec-tests/src/test/java/org/candlepin/spec/PoolSpecTest.java
Ceiu: I guess? Have we standardized on any verbiage, tense, or such for the names? If we're going to switch to something that makes quickly jumping to a test worse, I hope we're at least making text consistent.

Ceiu: Done. Since we're (to my knowledge) lacking a convention, I'm standardizing on "should" and "should not" without a leading noun, and reasonable descriptions of what the test does from there. Additionally, I updated the test names to match the display name so we can still find tests by quickly looking up the symbol.

nikosmoum: ack

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
```diff
@@ -0,0 +1,625 @@
...
+        assertNotNull(attributes);
+
+        assertThat(attributes)
+            .isNotNull()
+            .hasSizeGreaterThanOrEqualTo(2)
+            .containsEntry("suggested_quantity", "1")
+            .containsEntry("compliance_type", "Standard");
+    }
+
+  // it 'should allow fetching content delivery network by pool id' do
```
nikosmoum, line 496: is the ruby test comment needed here, if the java test is already commented below, and the ruby test still exists in the .rb file?

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
```diff
@@ -0,0 +1,625 @@
...
+        ApiClient consumerClient = this.createConsumerClient(userClient, owner);
+
+        PoolDTO fetched = consumerClient.pools()
+            .getPool(pool.getId(), null, null);
+
+        assertNotNull(fetched);
+        assertEquals(pool.getId(), fetched.getId());
+    }
+
+  // it 'includes null values in json' do
```
nikosmoum, line 153: same here: can we remove the ruby test comment since it's still in the .rb file?

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
```diff
@@ -0,0 +1,625 @@
...
+  //       result_cdn = @cp.get_cdn_from_pool(pool['id'])
+  //       result_cdn.name.should == cdn.name
+  //       result_cdn.url.should == cdn.url
+  //   end
+  //   @cp.delete_owner(@import_owner['key'])
+  //   @cp_export.cleanup
+  // end
+
+    // @Test
+    // public void shouldAllowFetchingCDNByPoolId() throws ApiException {
```
nikosmoum, line 523: Where are we on the CDN usage investigation? Is it not used at all by either Sat/Hosted?

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
```diff
@@ -0,0 +1,625 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+
+
+@SpecTest
+class PoolSpecTest {
+
+    private OwnerDTO createOwner(ApiClient client) throws ApiException {
+        return client.owners().createOwner(Owners.random());
```
nikosmoum, line 60: I know I am opening the can of worms again, but if we feel the need to be using such utility methods, and all (or at least most) of the private methods in this class seem generic enough to be able to be used in other spec tests, shouldn't we extract them in appropriate utility class(es)? Could be grouped into classes such as `OwnerUtils`, `PoolUtils`, etc.?

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
```diff
@@ -0,0 +1,625 @@
...
+            .value(value);
+    }
+
+    private PoolDTO createPool(ApiClient client, OwnerDTO owner) throws ApiException {
+        ProductDTO product = this.createProduct(client, owner);
+        return this.createPool(client, owner, product);
+    }
+
+    private PoolDTO createPool(ApiClient client, OwnerDTO owner, ProductDTO product) throws ApiException {
+        PoolDTO pool = Pools.randomUpstream(product);
```
nikosmoum, line 105: _If_ we decide to extract these methods to utility classes for reusability, then this method (and the one above) are hiding the fact that they set the upstream subscription properties; so we should probably use a different method name for them to differentiate them from custom pools that don't have those properties.

Alternatively, since for the majority of use cases the pool being custom vs imported from upstream does not matter, we shouldn't be calling the randomUpstream method within any such utility method, and instead let the developer explicitly use it only in the few tests that need them

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
nikosmoum: And the answer might be no here. Maybe it's fine for each test class to define it's own methods for readability

Ceiu: I would prefer not in these cases. I begrudgingly accept factory methods for building DTOs that can then be further manipulated or whatever as needed, but I don't know how valuable it is to make a global method that creates such a DTO and does some kind of operation on the CP side of things transparently.

It starts getting into the realm of one thing our unit tests do incorrectly with its global utilities in that we create something, and then we have to modify it, so instead of just doing create -> persist, we do create -> persist -> update -> persist. The unit test the does same amount of work (config + persist), but we've doubled up on the number of remote hits.

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
Ceiu: Doesn't seem to be. Next step would be checking with PMs to sort out removal details or if there's some feature or something critical that's on their roadmap.

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
Ceiu: If my memory is correct, the upstream requirement is more common than not. It ended up being a real pain when I tried to fully remove the ruby `create_subscription_and_pool` stuff some years back.

But more to your point, I don't think it's necessary to turn every local abstraction into a global utility method. Something I hit when attempting to do so with the consumer stuff, which is even more generic than this, is that turning this local method into a global function means more overloads and updating of existing functions to try to make every caller from every class able to use it. It ended up making them ugly and cumbersome, with no fewer than 4 variants of the signature.

That's kind of why the `randomUpstream` function exists -- I was requested to move a local context to a global one, but had to avoid breaking existing tests which I have no interest in updating in this PR. It doesn't add a ton of value IMO, and hides details about the pool creation that may or may not be critical to the individual test. And on the topic of the properties that make a pool appear like an upstream pool, rather than discouraging the use of such a function, I would much rather we sort out why those attributes are critical, and update pool so they're not. That's non-trivial, especially in the import bits where it's absolutely required today (it's how local pools are mapped to stuff from the manifest); but it'd be a much better long-term solution than trying to use function A and then falling back to function B if it doesn't work, without much thought to the details.

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
Ceiu: done

spec-tests/src/test/java/org/candlepin/spec/pools/PoolSpecTest.java
Ceiu: done

nikosmoum (approved):

nikosmoum:
I will let @Januson re-review&merge

Januson (approved):

PR #3465, opened 2022-07-21
[M] Removed length requirement on cloud registration signature (ENT-5283)

Ceiu:
- Cloud registration info no longer has a size requirement on
  the signature field; the field is still required to be present,
  and null values are still not permitted, but empty signature
  strings are now allowed

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -8391,7 +8391,6 @@ components:
           minLength: 1
         signature:
           type: string
-          minLength: 1
 
     ConsumerDTO:
       description: Represents a consumer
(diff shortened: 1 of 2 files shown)


Ceiu:
Upon acceptance, this should be backported to 4.1.

JoshAlbrecht23 (approved):

api/candlepin-api-spec.yaml
```diff
@@ -8391,7 +8391,6 @@ components:
           minLength: 1
         signature:
           type: string
-          minLength: 1
```
Januson, line 8394: We should probably remove signature from required section above.

Ceiu: That permits nulls and not supplying the field at all. I suppose that's a matter of semantics, but I restored the requirement after checking our design docs. They weren't 100% clear, but leaned more toward "must specify a signature" than "optional."

Januson (approved):

PR #3467, opened 2022-07-22
ENT-5256: Create client for HostedTestResource

Januson:
- Added HostedTestApi client
- Added spec tests for the HostedTestApi
- Updated HostedTestResource to use DTOs


src/main/java/org/candlepin/dto/hosted/SubscriptionTranslator.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+
+/**
+ * The SubscriptionTranslator provides translation from Subscription objects to the
+ * SubscriptionDTO (hosted test API)
+ */
+public class SubscriptionTranslator implements ObjectTranslator<Subscription, SubscriptionDTO> {
```
nikosmoum, line 38: Need to add unit tests for the new translator classes

spec/refresh_pools_spec.rb
```diff
@@ -1281,7 +1281,7 @@ def test_entitlement_regeneration
 
     # Yield to encapsulating test, applying any change it may have made
     sub = yield(owner, sub)
-    update_upstream_subscription(sub.id, sub)
+    sub = update_upstream_subscription(sub.id, sub)
```
Ceiu, line 1284: If we're always going to fetch the sub from upstream instead of trusting the block to manipulate it properly, it'd be better to just leave the job up to the container and do it all here.

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
```diff
@@ -315,8 +314,14 @@ void onlySuperAdminCanRefresh() throws ApiException {
         ApiClient readWriteClient = ApiClients.trustedUser(readWriteUser.getUsername());
         ApiClient adminClient = ApiClients.trustedUser(adminUser.getUsername());
 
-        ProductDTO product = createProduct(owner);
-        owners.createPool(owner.getKey(), Pools.random(product));
+        if (CandlepinMode.isHosted()) {
```
Ceiu, line 317: Creation of a pool up or downstream is extraneous for this test as-written. We currently don't care if refresh does anything, just that we can invoke it.

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
```diff
@@ -263,9 +273,9 @@ public OwnerInfo createOwner(OwnerDTO owner) {
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/subscriptions")
-    public SubscriptionInfo createSubscription(
+    public SubscriptionDTO createSubscription(
```
Ceiu, line 276: PoolDTO wasn't a fit here?

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
```diff
@@ -460,28 +486,30 @@ public boolean deleteProduct(@PathParam("product_id") String productId) {
     @GET
     @Path("/content")
     @Produces(MediaType.APPLICATION_JSON)
-    public Collection<? extends ContentInfo> listContent() {
-        return this.datastore.listContent();
+    public Collection<ContentDTO> listContent() {
+        return this.datastore.listContent().stream()
+            .map(content -> this.translator.translate(content, ContentDTO.class))
```
Ceiu, line 491: streammapper + return type

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
```diff
@@ -305,8 +316,10 @@ public SubscriptionInfo createSubscription(
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/subscriptions")
-    public Collection<? extends SubscriptionInfo> listSubscriptions() {
-        return this.datastore.listSubscriptions();
+    public Collection<SubscriptionDTO> listSubscriptions() {
```
Ceiu, line 319: Couple things here:
- when translating a collection like this, we should use the streamMapper, so we don't have to do the class resolution thing per-instance
- We could change the output to a Stream\<SubscriptionDTO\> here and save the collection step

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
```diff
@@ -379,28 +400,30 @@ public boolean deleteSubscription(@PathParam("subscription_id") String subscript
     @GET
     @Path("/products")
     @Produces(MediaType.APPLICATION_JSON)
-    public Collection<? extends ProductInfo> listProducts() {
-        return this.datastore.listProducts();
+    public Collection<ProductDTO> listProducts() {
```
Ceiu, line 403: streammapper + return type

Ceiu (changes requested):

spec/refresh_pools_spec.rb
Januson: Removed. This was a forgotten bit.

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
Januson: Done

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
Januson: Done.

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
Januson: Done.

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
Januson: Done.

src/main/java/org/candlepin/dto/hosted/SubscriptionTranslator.java
Januson: Done.

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
Januson: This might need a separate task. I think it will work but it needs a lot of changes (mostly in children creation and ruby tests) due to the differences in how PoolDTO and SubscriptionDTO handles producs.

spec/refresh_pools_spec.rb
Ceiu: This was in reference to each of the tests below (1310, 1323, and 1336) all fetching the subscription from upstream every time instead of just returning the subscription as-is. The fetching into this update will end up being a slow no-op since any subscription changes between the two will be lost.

The existing code could be optimized a bit to check if there's even a difference to make to justify an update/refresh call, but at the very least we shouldn't be doing a fetch->update.

src/main/java/org/candlepin/dto/api/v1/SubscriptionTranslator.java
```diff
@@ -0,0 +1,116 @@
...
+        return dest;
+    }
+
+    private Set<ProductDTO> translateProducts(ModelTranslator translator, Collection<ProductData> products) {
+        if (products == null) {
+            return new HashSet<>();
+        }
+
+        return products.stream()
+            .map(product -> translator.translate(product, ProductDTO.class))
```
Ceiu, line 112: use stream mapper here

src/main/java/org/candlepin/dto/shim/CertificateSerialInfoTranslator.java
```diff
@@ -0,0 +1,76 @@
...
+        CertificateSerialDTO dest) {
+        if (source == null) {
+            throw new IllegalArgumentException("source is null");
+        }
+
+        if (dest == null) {
+            throw new IllegalArgumentException("dest is null");
+        }
+
+        dest
```
Ceiu, line 68: extraneous linebreak

src/main/java/org/candlepin/dto/shim/ProductDTOTranslator.java
```diff
@@ -85,22 +91,65 @@ public ProductData populate(ModelTranslator modelTranslator, ProductDTO source,
...
+        if (branding == null) {
+            return new HashSet<>();
+        }
+
+        ObjectTranslator<BrandingDTO, Branding> translator = modelTranslator
+            .findTranslatorByClass(BrandingDTO.class, Branding.class);
+
+        return branding.stream()
+            .filter(Objects::nonNull)
+            .map(p -> translator.translate(modelTranslator, p))
```
Ceiu, line 138: Fetching the translator above saves us from the repeated resolution, but we still could use the stream mapper function to streamline the code here.

src/main/java/org/candlepin/dto/shim/ProductDTOTranslator.java
```diff
@@ -85,22 +91,65 @@ public ProductData populate(ModelTranslator modelTranslator, ProductDTO source,
...
+        if (products == null) {
+            return new HashSet<>();
+        }
+
+        ObjectTranslator<ProductDTO, ProductData> translator = modelTranslator
+            .findTranslatorByClass(ProductDTO.class, ProductData.class);
+
+        return products.stream()
+            .filter(Objects::nonNull)
+            .map(p -> translator.translate(modelTranslator, p))
```
Ceiu, line 152: Same here -- getStreamMapper

src/main/java/org/candlepin/dto/shim/ProductInfoTranslator.java
```diff
@@ -0,0 +1,181 @@
...
+        if (source == null) {
+            throw new IllegalArgumentException("source is null");
+        }
+
+        if (dest == null) {
+            throw new IllegalArgumentException("dest is null");
+        }
+
+        dest.id(source.getId())
+            .uuid(source.getId())
```
Ceiu, line 84: This should be null rather than being an incorrect value

src/main/java/org/candlepin/dto/shim/ContentInfoTranslator.java
```diff
@@ -66,7 +67,8 @@ public ContentDTO populate(ModelTranslator translator, ContentData source, Conte
             throw new IllegalArgumentException("dest is null");
         }
 
-        dest.uuid(source.getUuid())
+        dest
+            .uuid(source.getId())
```
Ceiu, line 71: Extraneous line break, plus we're setting the UUID to an incorrect value. We're better off not setting setting it to null rather than putting bad data in the field.

src/main/java/org/candlepin/dto/shim/SubscriptionInfoTranslator.java
```diff
@@ -0,0 +1,130 @@
...
+            .upstreamConsumerId(source.getUpstreamConsumerId())
+            .upstreamEntitlementId(source.getUpstreamEntitlementId())
+            .upstreamPoolId(source.getUpstreamPoolId());
+
+        if (translator != null) {
+            dest.cdn(translator.translate(source.getCdn(), CdnDTO.class))
+                .certificate(translator.translate(source.getCertificate(), CertificateDTO.class))
+                .owner(translator.translate(source.getOwner(), NestedOwnerDTO.class));
+            ProductInfo product = source.getProduct();
+            if (product != null) {
```
Ceiu, line 95: Line break before the if statement

Also need to set product to null on the outbound DTO in the no-translator and no-product cases.

src/main/java/org/candlepin/dto/shim/SubscriptionInfoTranslator.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private Set<ProductDTO> translateProducts(
+        ModelTranslator translator, Collection<? extends ProductInfo> products) {
+        if (products == null) {
+            return new HashSet<>();
+        }
+
+        return products.stream()
+            .map(product -> translator.translate(product, ProductDTO.class))
```
Ceiu, line 126: stream mapper

src/main/java/org/candlepin/hostedtest/HostedTestResource.java
Ceiu: ack

Ceiu (changes requested):

src/main/java/org/candlepin/dto/api/v1/SubscriptionTranslator.java
Januson: Sorry. I forgot to delete this translator. Same with ProductDTO and ContentDTO translators. Done.

src/main/java/org/candlepin/dto/shim/CertificateSerialInfoTranslator.java
Januson: Done.

src/main/java/org/candlepin/dto/shim/ProductDTOTranslator.java
Januson: Done.

src/main/java/org/candlepin/dto/shim/ProductDTOTranslator.java
Januson: Done.

src/main/java/org/candlepin/dto/shim/ProductInfoTranslator.java
Januson: Done.

src/main/java/org/candlepin/dto/shim/ContentInfoTranslator.java
Januson: Done.

src/main/java/org/candlepin/dto/shim/SubscriptionInfoTranslator.java
Januson: Do we need to be explicit for properties other than collections? Objects will just default to null.

src/main/java/org/candlepin/dto/shim/SubscriptionInfoTranslator.java
Januson: Done.

src/main/java/org/candlepin/dto/shim/SubscriptionInfoTranslator.java
Ceiu: The populate method is supposed to be idempotent, and is used for both cases where we're translating and populating. That's why we want to set everything to a known state based on the source object.

spec/refresh_pools_spec.rb
Januson: Unfortunately, those fetches are needed for the same reason I added those assignments of product after adding content to it.

Without them the creation of subscription updates its products. This wasn't a problem before, because hosted resource didn't use DTOs so all attributes defaulted to null. Even collections. So now when we create a product, it has no content. We add a content, but the dto we use in subscription still has no content, so the content gets deleted during subscription creation. This was not a problem before because content was null so it was ignored during subscription creation.

As an alternative to fetching we could set those collections to null, but that would require more upkeep.

Ceiu: Given that we aren't changing the product at all (even functionally), that should no-op. DTOs or not, the way the tests function, sending the subscription back up even with the field changed on the content that we've already changed upstream should end up doing nothing. If this is not the case, there's a regression in the hosted test logic.

However, my larger point here is that what this code is doing as a result of the changes to the individual tests is the following:

```
sub = get_upstream_subscription(subscription.id) # value returned from the block
update_upstream_subscription(sub.id, sub) # effective no-op since we'd be sending the sub we just fetched back up
```

Now, given how the tests actually ended up, the update itself is fairly extraneous. The remainder of the base function doesn't even use it, and if it's critical to the test, the test itself should probably be explicitly doing an update. That said, we could remove the entire subscription returning from the block, have the tests which want to update the subscription do their own update, and then drop the update in the base function entirely. The only silver lining of keeping it is that it seems to have exposed an issue we may not have caught otherwise.

Ceiu: To recap discussion from IRC:
The issue you were seeing was due to the change to the content in two of the three tests being made on the legacy paths to provided and derived provided products. Once the target content path was updated to use the 4.0+ pathing, the update of the subscription no longer clobbered the change made in the test.

With that sorted, we can safely remove that whole semi-transparent update logic, since it's only used in two of the ten tests using the `test_entitlement_regeneration` function, and apparently caused a similar concern in two others. The final diff ends up looking as follows:

```diff
diff --git a/spec/refresh_pools_spec.rb b/spec/refresh_pools_spec.rb
index ac5c8f5bd..f0ae95d5d 100644
--- a/spec/refresh_pools_spec.rb
+++ b/spec/refresh_pools_spec.rb
@@ -1279,9 +1279,8 @@ describe 'Refresh Pools' do
 
     expect(concat_serials(entitlement, bonus_entitlement)).to eq(serial_concat)
 
-    # Yield to encapsulating test, applying any change it may have made
-    sub = yield(owner, sub)
-    update_upstream_subscription(sub.id, sub)
+    # Yield to encapsulating test
+    yield(owner, sub)
 
     # verify serial does not change on content update request that does not regenerate cert
     entitlement = @cp.get_entitlement(entitlement['id'])
@@ -1306,8 +1305,6 @@ describe 'Refresh Pools' do
       content = subscription['product']['productContent'].first['content']
       content.modifiedProductIds = [product_id2]
       update_upstream_content(content.id, content)
-
-      subscription
     }
   end
 
@@ -1316,11 +1313,9 @@ describe 'Refresh Pools' do
       product_id2 = random_string(nil, true)
       product2 = create_upstream_product(product_id2, { :name => 'test_prod_2' })
 
-      content = subscription['providedProducts'][0]['productContent'][0]['content']
+      content = subscription['product']['providedProducts'].first['productContent'].first['content']
       content.modifiedProductIds = [product_id2]
       update_upstream_content(content.id, content)
-
-      get_upstream_subscription(subscription.id)
     }
   end
 
@@ -1329,11 +1324,9 @@ describe 'Refresh Pools' do
       product_id2 = random_string(nil, true)
       product2 = create_upstream_product(product_id2, { :name => 'test_prod_2' })
 
-      content = subscription['derivedProvidedProducts'][0]['productContent'][0]['content']
+      content = subscription['product']['derivedProduct']['providedProducts'].first['productContent'].first['content']
       content.modifiedProductIds = [product_id2]
       update_upstream_content(content.id, content)
-
-      get_upstream_subscription(subscription.id)
     }
   end
 
@@ -1367,8 +1360,7 @@ describe 'Refresh Pools' do
       add_content_to_product_upstream(prov_product.id, content_id)
 
       subscription['product']['providedProducts'].push(prov_product)
-
-      subscription
+      update_upstream_subscription(subscription.id, subscription)
     }
 
     prov_product = json_body['products'].find {|p| p['id'] == prov_product_id}
@@ -1378,7 +1370,7 @@ describe 'Refresh Pools' do
   it 'regenerates entitlements when provided product is removed' do
     json_body, main_product = test_entitlement_regeneration { |owner, subscription|
       subscription['product']['providedProducts'] = []
-      subscription
+      update_upstream_subscription(subscription.id, subscription)
     }
 
     json_body['products'].size.should == 1
@@ -1389,8 +1381,6 @@ describe 'Refresh Pools' do
       content = subscription['product']['productContent'].first['content']
       content.label = 'shakeItOff'
       update_upstream_content(content.id, content)
-
-      subscription
     }
 
     product_json = json_body['products'].find {|p| p['id'] == main_product['id']}
@@ -1402,8 +1392,6 @@ describe 'Refresh Pools' do
       content = subscription['product']['productContent'].first['content']
       content.releaseVer = 'badBlood'
       update_upstream_content(content.id, content)
-
-      subscription
     }
 
     # releasever is not in json
@@ -1414,8 +1402,6 @@ describe 'Refresh Pools' do
       content = subscription['product']['productContent'].first['content']
       content.vendor = 'blankSpace'
       update_upstream_content(content.id, content)
-
-      subscription
     }
 
     product_json = json_body['products'].find {|p| p['id'] == main_product['id']}
@@ -1435,10 +1421,6 @@ describe 'Refresh Pools' do
       })
 
       add_content_to_product_upstream(product.id, content.id)
-
-      # Need to return the updated, upstream subscription here so we don't risk clobbering the addition.
-      # We shouldn't, anyway, but there's no need to risk it unnecessarily.
-      get_upstream_subscription(subscription.id)
     }
 
     product_json = json_body['products'].find {|p| p['id'] == main_product['id']}
@@ -1454,10 +1436,6 @@ describe 'Refresh Pools' do
       content = product['productContent'].first['content']
 
       remove_content_from_product_upstream(product.id, content.id)
-
-      # Need to return the updated, upstream subscription here so we don't risk clobbering the removal.
-      # We shouldn't, anyway, but there's no need to risk it unnecessarily.
-      get_upstream_subscription(subscription.id)
     }
 
     product_json = json_body['products'].find {|p| p['id'] == main_product['id']}
```

src/main/java/org/candlepin/dto/shim/SubscriptionInfoTranslator.java
Januson: Done.

Ceiu:
The only hang up I have left with this is that the translators need to be explicitly setting all of the fields they recognize on every possible branch. We're relying too much on defaults when the translators have no knowledge or control over such a thing, which is not only risky behavior, but also prevents us from ever reliably using the population behavior.

Ceiu (approved):

PR #3470, opened 2022-07-26
[M] 2110028: Addressed failure with liquibase changeset (ENT-5284)

JoshAlbrecht23:
- Updated add_delete_cascade_to_consumer_fks.xml to check if
  fks exist before trying to remove with individual change sets.

Since one of the newly added FK constraint name (fk_consumer_guests_attributes_consumer_guests) is pretty long. I did some investigation to see if this could be a potential problem since the length of the Candlepin constraint names have been limited in the past.

**Postgres:**
- Default max constraint name length is 64 bytes
- Based on default deployment, our postgres db is using UTF-8 encoding (checked using \l)
- Best case the identifier length can be 64 characters long and worst case it can be 16 characters long. However, I think that is safe to assume that we will be only using single byte characters for constraint names.
- https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

**MariaDB:**
- Default max constraint name length is 64 characters which is stored in UTF-8.
- https://mariadb.com/kb/en/identifier-names/

**Oracle Database:**
- Compatibility version greater than 12.2 or higher can have up to 128 byte for the constraint name and for versions less than 12.2, they can have up to 30 bytes for the name.
- https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Database-Object-Names-and-Qualifiers.html#GUID-75337742-67FD-4EC0-985F-741C93D918DA

Based on this, I think that we were limiting the length of the constraint names because of the limitations of supporting a pre 12.2 version of Oracle Database, but if we no longer support Oracle Database then the max length of a constraint name should be 64 characters long.


src/main/resources/db/changelog/20220722000000-add_delete_cascade_to_consumer_fks.xml
```diff
@@ -6,56 +6,93 @@
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
-    <changeSet id="20220613000000-1" author="jalbrech">
+    <changeSet id="20220722000000-1" author="jalbrech">
+        <preConditions onFail="MARK_RAN">
```
Ceiu, line 10: With the removal and recreation of the foreign keys in the same changeset, we'll correctly remove the FK only if it exists and rebuild it as we expect, but if a system somehow doesn't have the FK at all, we won't attempt to build one for them.

src/main/resources/db/changelog/20220722000000-add_delete_cascade_to_consumer_fks.xml
```diff
@@ -6,56 +6,93 @@
...
             onDelete="CASCADE" />
-        
+    </changeSet>
+
+    <changeSet id="20220722000000-3" author="jalbrech">
+        <preConditions onFail="MARK_RAN">
+            <foreignKeyConstraintExists foreignKeyName="fk_cnsmr_capability_cnsmr"/>
+        </preConditions>
+
+        <dropForeignKeyConstraint  baseTableName="cp_consumer_capability" constraintName="fk_cnsmr_capability_cnsmr"/>
```
Ceiu, line 44: All the drops have extra whitespace between the command and the table name. Is this intentional?

Ceiu (changes requested):

src/main/resources/db/changelog/20220722000000-add_delete_cascade_to_consumer_fks.xml
JoshAlbrecht23: Good catch, that wasn't intentional. Removed the extra whitespace

src/main/resources/db/changelog/20220722000000-add_delete_cascade_to_consumer_fks.xml
JoshAlbrecht23: For each FK I added a change set for removing the existing FK if it exists and another change set for adding the new FK with the cascade on delete.

src/main/resources/db/changelog/20220722000000-add_delete_cascade_to_consumer_fks.xml
```diff
@@ -0,0 +1,104 @@
...
+        </preConditions>
+        <dropForeignKeyConstraint baseTableName="cp_consumer_capability" constraintName="fk_cnsmr_capability_cnsmr"/>
+    </changeSet>
+
+    <changeSet id="20220722000000-6" author="jalbrech">
+        <addForeignKeyConstraint baseTableName="cp_consumer_capability"
+            baseColumnNames="consumer_id"
+            referencedTableName="cp_consumer"
+            referencedColumnNames="id"
+            constraintName="fk_cnsmr_capability_cnsmr"
```
Ceiu, line 53: Let's update this constraint name to match the others. The other FK names beyond this point are also in a different general format (fk_... vs table_fk). We've not been consistent on naming thus far on a global basis, so it's probably not exactly critical which one we go with, but we should be consistent within a single changeset file.

Ceiu (changes requested):
Looks good, just a minor change to the naming of the FKs and this should be ready.

src/main/resources/db/changelog/20220722000000-add_delete_cascade_to_consumer_fks.xml
JoshAlbrecht23: Good catch, I missed that they were using different formats! I updated them to fk_<base table name>_<reference table name>

src/main/resources/db/changelog/20220722000000-add_delete_cascade_to_consumer_fks.xml
```diff
@@ -0,0 +1,104 @@
...
+        </preConditions>
+        <dropForeignKeyConstraint baseTableName="cp_consumer_guests_attributes" constraintName="fk_guest_attribute"/>
+    </changeSet>
+
+    <changeSet id="20220722000000-12" author="jalbrech">
+        <addForeignKeyConstraint baseTableName="cp_consumer_guests_attributes"
+            baseColumnNames="cp_consumer_guest_id"
+            referencedTableName="cp_consumer_guests"
+            referencedColumnNames="id"
+            constraintName="fk_consumer_guests_attributes_consumer_guests"
```
JoshAlbrecht23, line 101: @Ceiu I added the results of an investigation on max size of constraint names to the description of the PR.

Ceiu: Looks like 128bytes is the defacto standard across the three at the moment. I don't see any reason we would use non-ASCII characters for our object names, so 64 characters seems like a safe, reasonable limit.

Ceiu (approved):

PR #3471, opened 2022-07-27
Use jdbc jars from java dependencies instead of rpm

nikosmoum:
- The postgresql/mysql/mariadb rpms are no longer part of the
  docker image. Use the jars from our java dependencies instead.

### DIFF
--- a/bin/bash_functions
+++ b/bin/bash_functions
@@ -116,6 +116,17 @@ cert_from_pkcs12() {
     fi
 }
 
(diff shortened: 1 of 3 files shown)


nikosmoum:
Note: porting this since we still need to support this branch, and even plan to backport our release pipeline changes in the future (moving to PNC) here, so it needs to be in working state.

Ceiu (approved):

PR #3480, opened 2022-07-28
[M] Added spec-test naming standards documentation

JoshAlbrecht23:
There was some discussion about the standardization of spec-test test names and display names and I thought that we should document these standards.

### DIFF
--- a/spec-tests/README.md
+++ b/spec-tests/README.md
@@ -1,11 +1,35 @@
 # Candlepin - Java base spec tests
 Java spec tests for candlepin. 
   
-# Getting Started
-To run spec tests use the following commnand.
+# Running Tests
+To run spec tests use the following command.
 ```
 ./gradlew clean spec
 ```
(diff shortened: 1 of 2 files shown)


spec-tests/README.md
```diff
@@ -1,11 +1,35 @@
 # Candlepin - Java base spec tests
 Java spec tests for candlepin.
```
nikosmoum, line 2: Can we add a link from the top-level CONTRIBUTING.md (https://github.com/candlepin/candlepin/blob/master/CONTRIBUTING.md#spec-tests) to here? Specifically from the java spec-test section there. It's nice having extra README files, but if they're hidden/unknown to users they're not useful and tend to even get stale faster (I didn't even realize there was a README in the spec-tests dir until now)

JoshAlbrecht23: That's a really good point! I added a link in the CONTRIBUTING.md file.

nikosmoum (approved):

PR #3482, opened 2022-07-29
[4.0-HOSTED] 2112391: Add back the collected field to certs in manifest

nikosmoum:
- Add back the 'collected' field from certificates exported in
  manifests, to avoid breaking old (2.4 or older) candlepins
  importing manifests that expect it to be there.

### DIFF
--- a/server/src/main/java/org/candlepin/dto/manifest/v1/CertificateSerialDTO.java
+++ b/server/src/main/java/org/candlepin/dto/manifest/v1/CertificateSerialDTO.java
@@ -36,6 +36,9 @@ public class CertificateSerialDTO extends TimestampedCandlepinDTO<CertificateSer
     protected BigInteger serial;
     protected Date expiration;
     protected Boolean revoked;
+    // we export this to false by default now that the CRL functionality is removed to avoid breaking old
+    // candlepin imports.
+    protected Boolean collected = Boolean.FALSE;
 
     /**
      * Initializes a new CertificateSerialDTO instance with null values.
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/candlepin/dto/manifest/v1/CertificateSerialDTO.java
```diff
@@ -36,6 +36,9 @@ public class CertificateSerialDTO extends TimestampedCandlepinDTO<CertificateSer
     protected BigInteger serial;
     protected Date expiration;
     protected Boolean revoked;
+    protected Boolean collected = Boolean.FALSE; // we export this to false by default now that the CRL
```
Ceiu, line 39: Does this comment block pass the linter? Generally we put these as a block above the target line.

Ceiu (approved):
This should get them sorted. If you feel up to it, reformatting that comment would be better to my eyes, but it's not even remotely critical.

server/src/main/java/org/candlepin/dto/manifest/v1/CertificateSerialDTO.java
nikosmoum: it does pass. what target line?

nikosmoum: ah I see what you mean

nikosmoum: updated

Ceiu:
Running the psql spec tests locally worked as expected, I don't see why the mysql ones would fail with these changes, but the changes for the mysql jar stuff also blocked local testing in this branch.

Tentatively approving changes in the event our timeline is tight enough to justify applying logic over raw bureaucracy.

PR #3483, opened 2022-07-29
[M] Updated Vagrant configuration to support upcoming Ansible changes (ENT-5303)

Ceiu:
- Updated the requirements.yml to include non-core Ansible collections
  used by the Candlepin role
- Updated the Vagrantfile to use a newer compatibility mode and support
  fetching both roles and collections via ansible-galaxy

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -20,9 +20,20 @@ ANSIBLE_VARS = {
 
 def configure_ansible_provisioning(vm_config)
   vm_config.vm.provision "ansible" do |ansible|
+    # ansible.verbose = "vvv"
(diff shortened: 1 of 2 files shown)


Ceiu:
This will probably need to be backported to any supported branches to minimize dev pain once the role changes go through.

JoshAlbrecht23 (approved):

JoshAlbrecht23:
The changes look good to me, but I think another review from someone who is more familiar with Ansible would be a good idea!

Lorquas (approved):
LGTM  vagrant up works and the galaxy roles install correctly.

PR #3485, opened 2022-08-01
ENT-5212: Break up resources classes into smaller one.

bakajstep:
- OwnerResource broken down into OwnerResource, OwnerProductResource, OwnerContentResource.
- The tests for it were divided equally.
- ConsumerResource broken down into ConsumerResource, GuestIdResource.


src/test/java/org/candlepin/resource/OwnerContentResourceTest.java
```diff
@@ -0,0 +1,297 @@
...
+        assertEquals(cdto.getName(), output.getName());
+
+        Content entity = this.ownerContentCurator.getContentById(owner, cdto.getId());
+        assertNotNull(entity);
+        assertEquals(cdto.getName(), entity.getName());
+    }
+
+    @Test
+    public void createContentWhenContentAlreadyExistsAndLocked()  {
+        // Note:
```
Ceiu, line 160: This note is no longer in line with the actual behavior of the test, so let's remove it.

Ceiu (changes requested):
Looks pretty good, overall a very good change.

Let's remove the outdated comment, and also remove the trailing period from the commit message.

Ceiu (approved):

PR #3488, opened 2022-08-01
Convert huffman decoding to java [ENT-5230]

wottop:
This also converts one of the tests to ensure that it works

### DIFF
--- a/spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/X509HuffmanDecodeUtil.java
+++ b/spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/X509HuffmanDecodeUtil.java
@@ -1,0 +1,448 @@
+/**
+ * Copyright (c) 2009 - 2022 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.spec.bootstrap.data.util;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.zip.Inflater;
+import java.util.zip.InflaterOutputStream;
+
+
+public class X509HuffmanDecodeUtil {
+
+    private static final Logger log = LoggerFactory.getLogger(X509HuffmanDecodeUtil.class);
+    private static final Object END_NODE = new Object();
+
+    private long pathNodeId = 0;
+    private long huffNodeId = 0;
+
+
+    public X509HuffmanDecodeUtil() {
+    }
+
+    public HuffNode makeTrie(List<HuffNode> nodesList) {
+        // drop the first node if path node value, it is not needed
+        if (nodesList.get(0).getValue() instanceof PathNode) {
+            nodesList.remove(0);
+        }
+        while (nodesList.size() > 1) {
+            int node1 = findSmallest(-1, nodesList);
+            int node2 = findSmallest(node1, nodesList);
+            HuffNode hn1 = nodesList.get(node1);
+            HuffNode hn2 = nodesList.get(node2);
+            HuffNode merged = mergeNodes(hn1, hn2);
+            nodesList.remove(hn1);
+            nodesList.remove(hn2);
+            nodesList.add(merged);
+        }
+        return nodesList.get(0);
+    }
+
+    private int findSmallest(int exclude, List<HuffNode> nodes) {
+        int smallest = -1;
+        for (int index = 0; index < nodes.size(); index++) {
+            if (index == exclude) {
+                continue;
+            }
+            if (smallest == -1 || nodes.get(index).getWeight() <
+                nodes.get(smallest).getWeight()) {
+                smallest = index;
+            }
+        }
+        return smallest;
+    }
+
+    private HuffNode mergeNodes(HuffNode left, HuffNode right) {
+        return new HuffNode(null, left.weight + right.weight, left, right);
+    }
+
+    public List<String> hydrateContentPackage(byte[] payload) throws IOException {
+        List<HuffNode> pathDictionary = new ArrayList<>();
+        List<HuffNode> nodeDictionary = new ArrayList<>();
+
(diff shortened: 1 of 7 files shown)


spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
```diff
@@ -0,0 +1,146 @@
...
+        consumerApi = client.consumers();
+    }
+
+    @Test
+    @DisplayName("should encode many content urls")
+    public void shouldEncodeManyContentUrls() throws Exception {
+        ProductDTO product = new ProductDTO();
+        // needs a numeric id so it will appear in the cert
+        Random rand = new Random();
+        product.setId("" + Math.abs(rand.nextInt()));
```
nikosmoum, line 84: I think we have utilities already for this you an use (Products.randomEng() and Products.randomSKU()). Although looking through those we probably need to switch their names (SKU products don't need numeric ids, while eng products do)

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
```diff
@@ -0,0 +1,146 @@
...
+        ownerApi = client.owners();
+        owner = ownerApi.createOwner(Owners.random());
+        ownerContentApi = client.ownerContent();
+        ownerProductApi = client.ownerProducts();
+        consumerApi = client.consumers();
+    }
+
+    @Test
+    @DisplayName("should encode many content urls")
+    public void shouldEncodeManyContentUrls() throws Exception {
```
nikosmoum, line 80: Now that this is ported it can be removed from the ruby file

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUnpackUtil.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private CertificateUnpackUtil() {
+        throw new UnsupportedOperationException();
+    }
+
+    private static final Random RANDOM = new Random();
+
+    public static String random(String prefix) {
+        return prefix + "-" + System.nanoTime() + "-" + RANDOM.nextInt(100000);
+    }
```
nikosmoum, line 49: This method (and variable) seem unused so we can remove them (also, we already have utility methods in Products.java for generating random ids)

nikosmoum (changes requested):
I can't say I have feedback for the decoding code, and it looks like it works, so that's great :+1: 
Just have a couple minor requests

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
```diff
@@ -0,0 +1,143 @@
...
+        pool = ownerApi.createPool(owner.getKey(), pool);
+
+        AsyncJobStatusDTO refresh = ownerApi.refreshPools(owner.getKey(), false);
+        if (refresh != null) {
+            client.jobs().waitForJobToComplete(refresh.getId(), 15000);
+        }
+        ConsumerDTO consumer = client.consumers().register(Consumers.random(owner));
+        ApiClient consumerClient = ApiClients.trustedConsumer(consumer.getUuid());
+        consumerApi = consumerClient.consumers();
+        String bindResult = consumerApi.bind(consumer.getUuid(),
```
Januson, line 122: This might be a good place to add an overloaded method to avoid passing all those null every time we want to bind.

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+        for (int i = 0; i < 100; i++) {
+            ContentDTO content = Contents.random();
+            content.setContentUrl(
+                String.format("/content/dist/rhel/$releasever-%s/$basearch-%s/debug-%s", i, i, i));
+            content = ownerContentApi.createContent(owner.getKey(), content);
+            product = ownerProductApi.addContent(owner.getKey(), product.getId(), content.getId(), true);
+        }
+
+        PoolDTO pool = new PoolDTO();
```
Januson, line 109: This can be Pools.random()

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
```diff
@@ -0,0 +1,143 @@
...
+        AsyncJobStatusDTO refresh = ownerApi.refreshPools(owner.getKey(), false);
+        if (refresh != null) {
+            client.jobs().waitForJobToComplete(refresh.getId(), 15000);
+        }
+        ConsumerDTO consumer = client.consumers().register(Consumers.random(owner));
+        ApiClient consumerClient = ApiClients.trustedConsumer(consumer.getUuid());
+        consumerApi = consumerClient.consumers();
+        String bindResult = consumerApi.bind(consumer.getUuid(),
+            pool.getId(), null, null, null, null, false, null, null);
+        JsonNode result = new ObjectMapper().readTree(bindResult);
```
Januson, line 124: If we are not going to return DTO, we should at least return JsonNode so that spec tests don't have to handle their own object mappers.

nikosmoum:
The code looks fine, but the new test is failing with an NPE

Januson (changes requested):
Changes look good. I have just a few suggestions.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUnpackUtil.java
```diff
@@ -0,0 +1,124 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.security.cert.X509Certificate;
+import java.util.Base64;
+import java.util.zip.InflaterInputStream;
+
+
+public final class CertificateUnpackUtil {
```
JoshAlbrecht23, line 38: I added a CertificateUtil class which has a extractEntitlementCertificatesFromPayload method that uses some of the logic of these methods like extractPayload() and decompress(). I think that we should combine these two utility classes so that we are not duplicating code and having two different classes doing the same thing,

https://github.com/candlepin/candlepin/blob/master/spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java

JoshAlbrecht23 (commented):
Just a comment about the CertificateUnpackUtil class

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/X509HuffmanDecodeUtil.java
```diff
@@ -0,0 +1,454 @@
...
+            else {
+                name += (char) b;
+            }
+        }
+
+        pathDictionary.add(new HuffNode(END_NODE, weight));
+        List<HuffNode> triePathDictionary = new ArrayList<>(pathDictionary);
+        HuffNode pathTrie = makeTrie(triePathDictionary);
+
+        StringBuffer nodeBits = new StringBuffer();
```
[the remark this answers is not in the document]
wottop: Because the class is instantiated for each use, I think we are safe.

Ceiu:
Can we do this with JSS rather than BouncyCastle? Right now we've got our foot in both libraries, and it's been the cause of a fair few headaches related to FIPS compatibility. If we're going to be married to JSS for the foreseeable future, I'd rather we commit new code to it rather than continuing to straddle the fence and use two libraries which are completely incompatible.

PR #3493, opened 2022-08-02
[M] Ported the owner-content spec tests to Java (ENT-5199)

Ceiu:
- Ported the owner_content_resource rspec tests to Java
- Fixed a bug in OwnerContentResource.updateContent which would cause
  an exception or error if the content DTO contained a different ID
  than that specified in the URL


Ceiu:
Posting this as a draft, as there are still three tests to add, pending discussions or implementation in other areas.

spec/owner_content_resource_spec.rb
```diff
@@ -125,6 +125,7 @@
     original_serial.should_not == new_serial
   end
 
+  # TODO: This should be moved to the owner product resource spec tests
```
nikosmoum, line 128: Can it be moved as part of this PR? Otherwise whenever the owner_product card is picked up this test could be forgotten

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -0,0 +1,685 @@
...
+    }
+
+    @Test
+    @DisplayName("should not allow updating locked content")
+    public void shouldNotAllowUpdatingLockedContent() throws Exception {
+        ApiClient adminClient = ApiClients.admin();
+        OwnerDTO owner = this.createOwner(adminClient);
+
+        // TODO: This test cannot be completed until the upstream bits are available, since we can no
+        // longer query or set the locked flag via API
```
nikosmoum, line 460: In addition to the upstream bits, you also need the ability of sending a custom JSON payload for the content PUT request here, since the locked flag isn't on the DTO. I think we need to extend the scope of the task 'Ability to assert on JSON response' to also include 'Ability to send custom JSON payload to an endpoint'.

Until those are done though it might be better to mark this with the `@Disabled("Depends on X and Y features being ported to java spec tests")` annotation, to skip it entirely during runs

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -0,0 +1,685 @@
...
+  //   expect(c3set.size).to eq(1)
+  //   expect(c3set[0]['id']).to eq(c1_id)
+
+  //   c4set = @cp.list_content(@owner['key'], {:page=>4, :per_page=>1, :sort_by=>"id"})
+  //   expect(c4set.size).to eq(0)
+  // end
+
+    @Test
+    @DisplayName("should list content in sorted pages")
+    public void shouldListContentInSortedPages() throws Exception {
```
nikosmoum, line 193: Same here, can be marked as disabled

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -0,0 +1,685 @@
...
+  //   expect(c3set.size).to eq(1)
+  //   expect(c3set[0]['id']).to eq(c3_id)
+
+  //   c4set = @cp.list_content(@owner['key'], {:page=>4, :per_page=>1})
+  //   expect(c4set.size).to eq(0)
+  // end
+
+    @Test
+    @DisplayName("should list content in pages")
+    public void shouldListContentInPages() throws Exception {
```
nikosmoum, line 123: Same here, can be marked as disabled

nikosmoum (changes requested):
A few minor requests. I think the prerequisites for this PR might take a while to get done, so I don't think we should let the PR sit. Instead, just mark the java spec tests as disabled, leave the ruby versions of the tests untouched, and merge the PR. We can add the porting of the leftover tests separately once the other work items are done.

spec/owner_content_resource_spec.rb
Ceiu: Done

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
Ceiu: done

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
Ceiu: done

nikosmoum:
@Ceiu The global paging params PR was merged, so now this is unblocked

Ceiu:
Whoops, jumped the gun a bit. Need to add the tests for RUD operations on locked entities.

Ceiu:
Temporarily "blocked" on #3535 to make improvements to some of the content creation bits in this change.

Ceiu:
@nikosmoum This should be fully ready for review now

spec/owner_content_resource_spec.rb
```diff
@@ -5,22 +5,6 @@
...
-
-    @content_id = @content['id']
-
-    @product = create_product('test-product', 'some product', {:multiplier => 4})
-    @cp.add_content_to_product(@owner['key'], @product['id'], @content_id)
-  end
-
-
-
   it 'lists content in pages' do
```
nikosmoum, line 8: Now that you've ported these tests this file can be deleted

nikosmoum (changes requested):

spec/owner_content_resource_spec.rb
Ceiu: Done

PR #3494, opened 2022-08-02
2110031: add config for enabling/disabling the DB check on startup

wottop:
The report of the missing sets will still happen, but the runtime error
 will not be thrown. This will allow the startup to continue

### DIFF
--- a/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -129,6 +129,8 @@ public class ConfigProperties {
     // Space separated list of resources to hide in GET /status
     public static final String HIDDEN_CAPABILITIES = "candlepin.hidden_capabilities";
 
+    public static final String HALT_ON_LIQUIBASE_DESYNC = "candlepin.db.halt_on_liquibase_desync";
+
     // Authentication
     public static final String TRUSTED_AUTHENTICATION = "candlepin.auth.trusted.enable";
     public static final String SSL_AUTHENTICATION = "candlepin.auth.ssl.enable";
@@ -409,6 +411,7 @@ public class ConfigProperties {
             // submit one when registering.
             this.put(HIDDEN_RESOURCES, "environments");
             this.put(HIDDEN_CAPABILITIES, "");
+            this.put(HALT_ON_LIQUIBASE_DESYNC, "true");
             this.put(SSL_VERIFY, "false");
 
             this.put(FAIL_ON_UNKNOWN_IMPORT_PROPERTIES, "false");
--- a/src/main/java/org/candlepin/guice/CandlepinContextListener.java
+++ b/src/main/java/org/candlepin/guice/CandlepinContextListener.java
@@ -367,6 +367,9 @@ public class CandlepinContextListener extends GuiceResteasyBootstrapServletConte
      * @throws RuntimeException if there are missing changesets or a LiqubaseException
      */
     protected void checkDbChangelog() {
+        if (!config.getBoolean(HALT_ON_LIQUIBASE_DESYNC)) {
+            return;
+        }
         try {
             Liquibase liquibase = getLiquibase();
             List<ChangeSet> unrunChangesets = liquibase.listUnrunChangeSets(null, null);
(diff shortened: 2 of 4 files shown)


src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -129,6 +129,8 @@ private ConfigProperties() {
     // Space separated list of resources to hide in GET /status
     public static final String HIDDEN_CAPABILITIES = "candlepin.hidden_capabilities";
 
+    public static final String DB_HALT_STARTUP = "candlepin.db.halt_startup";
```
Ceiu, line 132: "halt_startup" to me implies that we're trying to prevent something DB related from running its startup routine(s). If we're going to continue on the path of still running the checks every time, but not failing on error, we should rename this to something more intuitive, such as "halt_on_error", "allow_errors", or "allow_liquibase_desync".

Ceiu (changes requested):
On the surface this works, but feels like it misses the larger point of the RFE. In cases where we'll want to disable the check, the logging output isn't really of consequence, and will end up being extraneous noise.

I think having an option to ignore errors during the check is fine, but we should still have a more universal off switch for the feature as a whole.

src/main/java/org/candlepin/config/ConfigProperties.java
nikosmoum: I think we only want a config option to skip the entire check (instead of not failing if it errors), so that kind of option would need a different name such as `candlepin.check_db_updated`

wottop: I was going for "more information is better", but you are correct as they did not request this.
I reserve the right to bring this back up if it gets asked for :)

src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -371,6 +371,9 @@ private void insertValidationEventListeners(Injector injector) {
      * @throws RuntimeException if there are missing changesets or a LiqubaseException
      */
     protected void checkDbChangelog() {
+        if (!config.getBoolean(HALT_ON_LIQUIBASE_DESYNC)) {
+            return;
+        }
```
JoshAlbrecht23, line 376: I think that we should still log out the missing changesets so that it is clear that there are missing changesets, but then have a log message that states that we will omit these changes due to the configuration. I think if we hide the missing changesets from the logs it will be easy for people to add a false to the halt config and forget about it. I'm open to opinions on this though!

We could put the "Please update the database.", "Aborting Candlepin initialization ...", and the RuntimeException in a conditional for when the HALT_ON_LIQUIBASE_DESYNC is true so that it is displayed only when CP should halt.

JoshAlbrecht23 (commented):
Question regarding if we should still log the missing changesets.

src/main/java/org/candlepin/guice/CandlepinContextListener.java
nikosmoum: The check for missing changesets on startup was added for a specific use case: upstream katello/foreman/candlepin users.
When it comes to Satellite customers, they don't actually need this check because the satellite installer tooling takes care of database upgrades (it's just a nice-to-have insurance policy that they'll probably never need).
And when it comes to Hosted, they always check if/what the changesets that need to be run are before an upgrade, so they don't have a need for it (that's why they asked for this config to disable it entirely).

So tl;dr, I don't think any big issue would be caused if someone turned this check off and forgot about it, and for hosted to be able to turn the check off entirely also gives a startup speed boost (since the liquibase db tables arent checked).

JoshAlbrecht23: I see! Thanks for the context around the check. I'm fine with not logging out the missing changesets since they aren't crucial to any of the consumers in this case!

Ceiu: This is a case where if we wanted the logging (which is a fair ask), it'd be separate configurations: one for whether or not to run the check at all including all the logs, and one to actually allow CP to continue running in such a state.

Given the initial implementation of this PR, that's where I expected it to go, but it's not strictly required for the RFE in question, and I'm not going to hold up the PR for it to be added back in.

JoshAlbrecht23 (approved):

PR #3504, opened 2022-08-05
[M] Fixed cp-test artifact collection and spec test filtering (ENT-5313)

Ceiu:
- Addressed an issue with the usage of trap and set -e in cp-test
  which caused errors in testing to skip cleanup and artifact
  collection
- Addressed an issue with a previous refactoring of the rspec
  filtering that missed the assignement of the filter variable,
  causing spec filters to not be evaluated or used

### DIFF
--- a/docker/cp-test
+++ b/docker/cp-test
@@ -10,7 +10,7 @@
 # See usage below.
 
 # Exit on any error:
-set -e
+set -eE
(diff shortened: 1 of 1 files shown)


Lorquas (approved):
Looks good; merging

PR #3508, opened 2022-08-11
ENT-5308: Add and apply global paging params to OpenApi spec

Januson:


api/candlepin-api-spec.yaml
```diff
@@ -9696,3 +9682,33 @@ components:
...
+      schema:
+        type: string
+        example: name
+    paging_sort_by:
+      name: sort_by
+      description: Property to use for ordering
+      in: query
+      schema:
+        type: string
+        example: asc
```
Ceiu, line 9714: Need a new line at EOF

api/candlepin-api-spec.yaml
```diff
@@ -9696,3 +9682,33 @@ components:
     basicAuth:
       type: http
       scheme: basic
+
+  parameters:
+    paging_page:
+      name: page
+      description: Number of a page to return
```
Ceiu, line 9689: Not your original wording, but we should use this chance to correct the phrasing here to something like "page [number|index|offset] to return"

Ceiu (changes requested):
Two very minor changes in the spec file, and a general question: Do we have a way to add stuff like this without impacting every API method which might use the parameters? We're going to run into a lot of pain when it comes to adding filters where virtually every method will need the include/exclude parameters, on top of the paging, on top of...

api/candlepin-api-spec.yaml

api/candlepin-api-spec.yaml

api/candlepin-api-spec.yaml
```diff
@@ -9696,3 +9682,33 @@ components:
...
+      schema:
+        type: string
+        example: name
+    paging_sort_by:
+      name: sort_by
+      description: Property to use for ordering
+      in: query
+      schema:
+        type: string
+        example: asc
```
nikosmoum, line 9714: The two 'example' values for the order and sort_by are inverted here. An example of the `order` param would be 'asc' or 'desc', and an example of the `sort_by` param would be 'name' or 'id'

PR #3509, opened 2022-08-11
[M] Migrated spec test exporter infrastructure to java (ENT-5299)

JoshAlbrecht23:


spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,329 @@
...
+    private OwnerContentApi ownerContentApi;
+    private CdnApi cdnApi;
+    private RulesApi rulesApi;
+    private JobsClient jobsClient;
+
+    private StandardExportFactory exportFactory;
+    private File export;
+
+    @BeforeAll
+    void beforeAll() throws Exception {
```
Ceiu, line 76: I've posted this on a few other posts, but global state config is bad for test isolation. In particular this targets things like the factory and consumer UUID at the bottom of this method, but the majority of the clients aren't stateless, which leaks into tests. In this particular case it's *probably* fine since literally every test actually uses the export factory, but the export itself is probably a bit extraneous at this level. Long-term, however, I don't want to get into a case like the ConsumerResource unit tests, which are simultaneously clunky individually due to a bunch of setup, but also interdependent because of their reliance on global test state.

As an aside, the name "client" for the extensions of the API wrappers is very misleading. It makes it look like they encapsulate owner permissions, when it's not. We should probably rename those.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,329 @@
...
+        File[] certFiles = productsDir.get(0).listFiles((file, str) -> {
+            return str.toLowerCase().contains(".pem");
+        });
+        assertEquals(expectedNumberOfCerts, certFiles.length);
+    }
+
+    @Test
+    @DisplayName("should allow manifest creation with read only user")
+    void shouldAllowManifestCreationWithReadOnlyUser() throws Exception {
+        exportFactory.createExportWithReadOnlyUser();
```
Ceiu, line 318: We should probably verify the created export exists and has the appropriate properties. There are no assertions at the factory level, so this test only verifies that this particular call completes without an exception-level error.

Really the entirety of that util method should be moved to this test, and then have assertions sprinkled about.

Ceiu (changes requested):

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/AbstractExportFactory.java
```diff
@@ -0,0 +1,196 @@
...
+    protected UsersApi userApi;
+    protected JobsClient jobsClient;
+
+    protected ConsumerDTO consumer;
+    protected OwnerDTO owner;
+    protected List<File> filesToCleanup = new ArrayList<>();
+    protected CdnDTO cdn;
+    protected Map<String, ProductDTO> productIdToProduct = new HashMap<>();
+
+    public AbstractExportFactory(ConsumerClient consumerClient, OwnerClient ownerClient,
```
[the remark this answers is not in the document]
JoshAlbrecht23: Nope! Passing in the admin client is a much better approach. Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/StandardExportFactory.java
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * A factory for building standard manifest exports. Creates all of the needed products, content,
+ * and pools for generating a standard manifest.
+ */
+public class StandardExportFactory extends AbstractExportFactory {
+    private OwnerProductApi ownerProductApi;
+    private OwnerContentApi ownerContentApi;
+    private CdnApi cdnApi;
+
+    public StandardExportFactory(ConsumerClient consumerClient, OwnerClient ownerClient,
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated to use the single admin client

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/StandardExportFactory.java
```diff
@@ -0,0 +1,193 @@
...
+        consumerClient.bind(consumer.getUuid(), product3Pool.getId(), 1);
+        consumerClient.bind(consumer.getUuid(), productUpPool.getId(), 1);
+        consumerClient.bind(consumer.getUuid(), productVdcPool.getId(), 1);
+
+        super.cdn = cdnApi.createCdn(Cdns.random());
+    }
+
+    private PoolDTO createPool(String ownerKey, String productId, long quantity, Set<BrandingDTO> brandings)
+        throws ApiException {
+        PoolDTO pool = Pools.random();
```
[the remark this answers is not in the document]
JoshAlbrecht23: I didn't even notice that the DTOs had those "set-less" mutators. That is interesting. Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/FileUtil.java
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * @param path containing a potential zip file.
+     * @return true if the file is a zip or false if it is not.
+     */
+    public static boolean isZip(Path path) {
+        if (path == null || !Files.exists(path)) {
+            throw new IllegalArgumentException("Path is null or file does not exist.");
+        }
+
+        if (path.toFile().isDirectory()) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: I forgot to add a comment on this test because I was confused by the original test. The original Ruby test only had the factory creating the read only export and so I wasn't sure if the test was just making sure that the export could be created without error or not. I moved over the logic from createExportWithReadOnlyUser() into the test and then added assertions on the export!

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/FileUtil.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        List<File> directories = new ArrayList<>();
+        if (!rootDirectory.isDirectory()) {
+            return directories;
+        }
+
+        // Add all matching directories in the current directory
+        List matchedDirs = Arrays.asList(rootDirectory.listFiles((file, str) -> {
+            return str.toLowerCase().equals(name.toLowerCase());
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/FileUtil.java
```diff
@@ -0,0 +1,152 @@
...
+            throw new IllegalArgumentException("Root directory is null.");
+        }
+
+        List<File> directories = new ArrayList<>();
+        if (!rootDirectory.isDirectory()) {
+            return directories;
+        }
+
+        // Add all matching directories in the current directory
+        List matchedDirs = Arrays.asList(rootDirectory.listFiles((file, str) -> {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Thanks for the information on the use of `Arrays.asList` vs `List.of`! I moved this method into the ExportSpecTest as a private method.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/FileUtil.java
```diff
@@ -0,0 +1,152 @@
...
+    public static File unzip(Path zipPath, Path targetDirectory) throws IOException {
+        if (zipPath == null || !Files.exists(zipPath)) {
+            throw new IllegalArgumentException("Zip file is null or does not exist.");
+        }
+
+        if (targetDirectory == null) {
+            throw new IllegalArgumentException("Target directory is null.");
+        }
+
+        if (!isZip(zipPath)) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/StandardExportFactory.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        this.ownerProductApi = ownerProductApi;
+        this.ownerContentApi = ownerContentApi;
+        this.cdnApi = cdnApi;
+
+        initialize();
+    }
+
+    private void initialize() throws ApiException {
```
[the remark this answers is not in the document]
JoshAlbrecht23: I created methods bindPoolsToConsumer, createPoolsForProducts, and createPool in the AbstractExportFactory class to help with the repetitive code.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: Yeah, I was following the pattern of the original ruby tests for how a lot of this was setup. I did see a note on the Exporter class in the candlepin_scenarios mentioning how creating the exports takes a lot of time, and I'm wondering if that is why the pattern of using a single factory and export for all the tests was chosen for export_spec.rb. 

Also, I removed the "client" from the names of the API wrappers throughout!

candlepin_scenarios Exporter comment:
https://github.com/candlepin/candlepin/blob/f4b25d0b9b765280074c600fd30c1c467fff9a8d/spec/candlepin_scenarios.rb#L319-L327

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/StandardExportFactory.java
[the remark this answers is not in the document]
Ceiu: My apologies here, I was meaning to abstract some of the repeated blocks of code into a separate private method in this class, not making some of the behavior more generic and exposed externally.

For instance, some of it could be abstracted into like a "createVDCProduct" method which creates the content, SKU and engineering entities. However, since it's all one-off interconnected stuff, so sometimes that's hard to break up; and if not broken up properly, it can get even worse to read than a long method. I'll leave the judgement up to you, but the protected methods in the parent class isn't what I had in mind.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/FileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+            return null;
+        }
+
+        Files.createDirectory(targetDirectory);
+        try (ZipInputStream zipInputStream = new ZipInputStream(Files.newInputStream(zipPath))) {
```
Ceiu, line 67: I've been looking at this for a bit now, and I'm beginning to wonder: given that we can query for a ZipEntry specifically off a ZipFile, is it necessary for our tests to even unzip the export archives we receive? We could just treat it as a fully-contained file system and avoid needing most of this class and the cleanup it mandates. Moreover, if we ensure the export file itself is created as a temp file (via File.createTempFile), then it'll automatically be cleaned up by the JVM and we won't need to do any management of it at all.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
Ceiu: Ahh, I misread it as a `@BeforeEach`.

I'm not sure how to feel here, since we still have cases where we create this export, and then it's not used for some of the tests which, instead, create their own export. This feels like a real good place to use the junit 5 nested test class bits, with its own beforeAll to spinup the necessary to create the single export, and then all the individual tests which examine it (and can do so in isolation) can become part of that nested test class. From there, the other tests which either make their own exports can live on in the base/outer test class and operate individually.

This allows us to start isolating actual cases where we have a need for shared resource or state creation, and minimize the frequency at which we toss stuff into the global or class scope.

Ceiu (changes requested):

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/StandardExportFactory.java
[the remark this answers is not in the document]
Ceiu: As a general note, if you prefer this format, there are several places in the code still using the non-fluent style. This is a stylistic thing, however, so it's not blocking if you don't want to deal with it.

Ceiu:
As a general thought, it's really feeling a lot like some of the design of the spec bits are being inherited from the old ruby code, for better or worse. In the explicit case of the exporter, and the general availability of better tooling for working with zip files, the "exporter" itself is becoming little more than just predefined test data spinup. With some of the organizational tools we get out of junit, I'm wondering how much of that logic needs to exist in its own separate utility.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: This makes a lot of sense and I like the use of the nested test class here!

nikosmoum:
@Ceiu How would this be used by other tests if it doesn't exist in its own separate utility?

Ceiu:
I'm not sure that it strictly *needs* to be used by other tests, honestly. Looking at what we're getting out of the export infrastructure, it basically boils down to creating a bunch of entities (as per normal test behavior), invoking the export endpoint, and then the file management stuff like unzipping and destroying files.

In Java, we probably don't actually need the unzipping stuff, as Java provides an API for dealing with that and pulling the actual file data as we'd need, and the invocation is a one to three line bit of code. It's a convenience to get around making a bunch of pools, products, and content -- that's where the bulk of the value is here.

Also, a number of the tests using the exporter are in the wrong suites. Like the job schedule spec has some import tests, import_update and import_warning should be part of import, etc..

The larger concern I'm having/seeing is that with all the work going into these new spec tests, we shouldn't be worried about tying ourselves to the decisions made in Ruby upward of 12 years ago by a group of devs who probably aren't/weren't Ruby experts. Even in the little bit of work I've done in the spec tests, I've seen and fixed a lot of shortcomings in the original design or overlooked tests. If we fall into the trap of "just" porting the existing logic over close to 1:1, we're going to inherit a lot of pain and miss a fantastic opportunity to shore up some of our gaps in testing.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/StandardExportFactory.java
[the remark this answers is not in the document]
JoshAlbrecht23: That makes sense! Like you said some of test data initialization is hard to break up. The createVDCProduct makes sense and I added it. I moved the bindPoolsToConsumer, createPoolsForProducts, and createPool methods to private methods in the ExportSpecTest.

src/main/java/org/candlepin/async/tasks/ImportJob.java
```diff
@@ -230,7 +230,7 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
                 storedFileId, overrides, uploadedFileName);
 
             log.info("Async import complete.");
-            context.setJobResult(importRecord);
+            context.setJobResult(importRecord.toString());
```
JoshAlbrecht23, line 233: This was causing a de-serialization issue with the AsyncJobStatusDTO object because the "resultData" field would be an object instead of a string when you fetched the status of the ImportJob using the /jobs/{job_id} endpoint.

JoshAlbrecht23:
@Ceiu I agree that we really don't need exporter factories because tests that test an export should live in the ExportSpecTest.java file. I moved the logic from the exporter factory classes to the ExportSpecTest file and moved the "Standard Export Tests" to a nested test class so that the test class can handle creating/destroying the test data and can be isolated. This can be a pattern that the other ruby export factories can follow when they are migrated as well.

I also moved over the job schedule spec tests that use the standard exporter.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,579 @@
...
+        }
+
+        @AfterAll
+        void afterAll() throws Exception {
+            export.close();
+        }
+
+        @Test
+        @DisplayName("should export consumer types")
+        void shouldExportConsumerTypes() throws Exception {
```
Ceiu, line 189: This test is backed by the utility method which makes very strong assumptions about files never having conflicting names between directories. We should be pulling very specific files or file data on an individual basis to verify it exists where and how we expect it. For example, I'd expect this test to look something closer to the following:

```
List<ConsumerTypeDTO> types = this.consumerTypeApi.getConsumerTypes();
for (ConsumerTypeDTO ctype : types) {
  ZipEntry entry = this.export.getEntry("path/to/types/" + ctype.getName() + ".json");
  assertNotNull(entry);

  try (InputStream istream = this.export.getInputStream(entry)) {
    ConsumerTypeDTO parsed = ApiClient.MAPPER.readValue(istream, ConsumerTypeDTO.class);
    assertEquals(ctype, parsed);
  }
}
```

Bonus points for also verifying that there aren't additional files left in the directory that contains the JSON files for consumer types.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,579 @@
...
+            for (File jsonFile : entitlementFiles.values()) {
+                JsonNode root = new ObjectMapper().readTree(Files.readString(jsonFile.toPath()));
+                assertNull(root.get("consumer"));
+            }
+        }
+
+        @Test
+        @DisplayName("should export candlepin 0.x backward compatible product definitions")
+        void shouldExportCandlepinBackwardCompatibleProductDefinitions() throws Exception {
+            List<ZipEntry> entitlements = ExportUtil.getFilesInDirectory(ENTITILEMENTS_PATH, export);
```
Ceiu, line 246: This comment applies to this test and several others, but a lot of this code could be streamlined with use of streams directly from the zip file:

```
this.export.stream()
  .filter(entry -> entry.isFile() && entry.getName().beginsWith(ENTITLEMENTS_PATH))
  .forEach(entry -> {
    // JSON Node processing
  })
```

Ceiu (changes requested):
I didn't get too deep into the individual tests since my initial requests on the utilities will have a pretty large impact on them, and it didn't feel right to keep nagging on the same thing over and over.

Ceiu:
Oh, a general thought about the use of ObjectMappers: they can have serialization extensions bolted onto them, so we should try to be consistent about which ones we use. Every time we spin up a new one, it's vanilla and may (de)serialize our objects differently than the general-purpose ObjectMapper held by ApiClient.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Retrieves all {@link ZipEntry}s in a provided path inside of the provided {@link ZipFile}.
+     *
+     * @param path location of the directory to retrieve zip entries from.
+     * @param zipFile the zip file system to retrieve the zip entries from.
+     * @return the zip entries in the provided zip file and target directory.
+     */
+    public static List<ZipEntry> getFilesInDirectory(String path, ZipFile zipFile) {
+        List<ZipEntry> foundFiles = new ArrayList<>();
+        Enumeration<? extends ZipEntry> zipEntries = zipFile.entries();
```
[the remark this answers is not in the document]
Ceiu: I should probably clarify before it's too late: when comparing the name to a path, if we're going to do it correctly, we should also verify that the match we're making is on a proper directory separator so we don't get into this situation:

path: "/my/target/dir"
"/my/target/dir" - exact match
"/my/target/dir/file" - also a match (starts with path)
"/my/target/dirname/file" - shouldn't match but does (starts with path, but not in the way we want)

The inverse is true of filenames:
filename: "myfile.txt"
"/my/target/dir/myfile.txt" - exact match
"/my/target/dir/thisismyfile.txt" - also matches but shouldn't

This is less of a concern with ZipEntrys since they appear to use the FQ name for everything so we're likely to pass in the FQ path as well, but it's something to keep in mind.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
```diff
@@ -0,0 +1,133 @@
...
+     */
+    public static ZipFile getExportArchive(File manifest) throws IOException {
+        if (manifest == null) {
+            throw new IllegalArgumentException("Manifest is null.");
+        }
+
+        try (ZipFile zipFile = new ZipFile(manifest)) {
+            ZipEntry entry = zipFile.getEntry(EXPORT_NAME);
+            File tmp = File.createTempFile("export", ".zip");
+            tmp.deleteOnExit();
```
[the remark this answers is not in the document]
JoshAlbrecht23: I meant to add a comment on this because I was confused on why the `File.createTempFile` would not cleanup the file when the VM terminates. Taking a look at the java docs, it mentions that you have to invoke the deleteOnExit method if you want the file to be deleted automatically.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#createTempFile(java.lang.String,java.lang.String,java.io.File)

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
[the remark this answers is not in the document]
JoshAlbrecht23: Updated to use a steam instead of enumeration, and I added a helper method isEntryInDirectory that verifies that the file starts with the target directory. To address the sub directory issue I check if there is an additional '/' in the path after the beginning target directory substring. I'm not sure if there is a better way of validating the file is not in sub directory of the target path.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /**
+     * Get a single {@link ZipEntry} that matches the provided file name in a given {@link ZipFile}.
+     *
+     * @param fileName the name of the file to retrieve.
+     * @param zipFile the zip file to retrieve the {@link ZipEntry} from.
+     * @return a zip entry that matches the provided file name.
+     */
+    public static ZipEntry getFile(String fileName, ZipFile zipFile) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Copies the provided {@link ZipEntry}s to temporary {@link File}s that can be used to read content from.
+     *
+     * @param zipFile the zip file to read zip entries from.
+     * @param entries the individual files that should be copied to temporary files.
+     * @return a map of the zip entry file name to the corresponding temporary file.
+     * @throws IOException if unable to read from the zip entries or write to the new files.
+     */
+    public static Map<String, File> copyToTempFiles(ZipFile zipFile, Collection<ZipEntry> entries)
```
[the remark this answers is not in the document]
JoshAlbrecht23: I see what you are saying, using the FileInputStream instead of creating all these temp files makes life so much better. I removed this method since it's not needed anymore.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,579 @@
...
+    void shouldAllowManifestCreationWithReadOnlyUser() throws Exception {
+        ApiClient client = ApiClients.admin();
+        OwnerDTO owner = client.owners().createOwner(Owners.random());
+        ConsumerDTO readOnlyConsumer = createReadOnlyConsumer(client, owner);
+
+        File readOnlyManifest = createExport(client, readOnlyConsumer.getUuid(), null);
+        ZipFile readOnlyExport = ExportUtil.getExportArchive(readOnlyManifest);
+        List<ZipEntry> exportDir = ExportUtil.getFilesInDirectory(EXPORT_PATH, readOnlyExport);
+        List<ZipEntry> consumerFiles = exportDir.stream()
+            .filter(e -> e.getName().contains("consumer.json"))
```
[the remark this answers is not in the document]
JoshAlbrecht23: That makes sense! I updated this throughout to not use 'e' as the single character name and replaced the name comparison to use `endsWith()` instead of `contains`. I applied this to other tests as well!

For the verification that the entry is a file, I found out that all the entries in the ZipFile are files, it doesn't have entries for just directories, so all the entries `isDirectory()` will return false.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: Should we use the getFilesInDirectory method to also make sure that the zip entries are not in a sub directory?

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,579 @@
...
+            .collect(Collectors.toList());
+        assertEquals(1, consumerFiles.size());
+
+        Map<String, File> actual = ExportUtil.copyToTempFiles(readOnlyExport, consumerFiles);
+        JsonNode root = new ObjectMapper()
```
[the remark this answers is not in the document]
JoshAlbrecht23: The issue that I run into is that most of the json files in the manifest can't be de-serialized to the object classes because there are unrecognized fields in the json files. I included the error message that is thrown when I try to de-serialize into the json file as a ConsumerDTO class.

```
  com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "created" (class org.candlepin.dto.api.v1.NestedOwnerDTO), not marked as ignorable (5 known properties: "contentAccessMode", "id", "displayName", "key", "href"])
   at [Source: (ZipFileInflaterInputStream); line: 1, column: 100] (through reference chain: org.candlepin.dto.api.v1.ConsumerDTO["owner"]->org.candlepin.dto.api.v1.NestedOwnerDTO["created"])
      at app//org.candlepin.spec.ExportSpecTest.shouldAllowManifestCreationWithReadOnlyUser(ExportSpecTest.java:123)
```

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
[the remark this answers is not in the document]
JoshAlbrecht23: It looks like setting the ObjectMapper DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES configuration to false resolves that issue!

src/main/java/org/candlepin/async/tasks/ImportJob.java
Ceiu: nack on this

The way this is intended to work is that any job result is JSON serialized. There's an argument to be made as to whether or not we should be serializing model objects (even fake ones), but changing this to .toString completely changes the output of this job.

If there's a problem on the spec test side of things, that should be updated instead.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/FileUtil.java
JoshAlbrecht23: Updated to use the ZipFile as a self-contained files system and ensured that it is cleaned up by the JVM! This cleaned up a lot!

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        String remainingPath = fullyQualifiedPath
+            .substring(directoryPath.length(), fullyQualifiedPath.length());
+
+        // Return false if the file is in a sub-directory.
+        return !remainingPath.contains("/");
+    }
+
+}
```
JoshAlbrecht23, line 206: I'm not sure if there is a better way of validating that the ZipEntry is not in a sub directory.

src/main/java/org/candlepin/async/tasks/ImportJob.java
JoshAlbrecht23: That makes sense. I used the Request api to avoid the serialization issues!

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,621 @@
...
+            .providedProducts(new HashSet<>())
+            .accountNumber("12345")
+            .orderNumber("6789")
+            .endDate(OffsetDateTime.now().plusYears(5))
+            .branding(new HashSet<>(brandings));
+
+        return ownerApi.createPool(ownerKey, pool);
+    }
+
+    private void waitForJob(ApiClient client, String jobId) throws IOException {
```
JoshAlbrecht23, line 595: I added this method because the `JobsUtil.waitForJob()` method throws errors for shouldPurgeImportRecords test when trying to retrieve the job status and de-serializing the job status object. So, to avoid this, I'm using the Request api and just reading the state of the job off the json.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
Ceiu: Can you provide some context? At the surface, the fetching of the ZipEntry in itself should be enough to verify its path:

```
ZipFile file = ...
ZipEntry entry = file.getEntry("/path/to/file.txt"); // entry.txt is in /path/to if this returns a ZipEntry
```

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
Ceiu: What exception are you seeing?

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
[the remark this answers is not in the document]
Ceiu: That's unfortunate. Good catch.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: This is that issue where the import job is returning an object in the resultData field for the AsyncJobStatusDTO instead of a string, and so the ApiClient can't deserialize the json into the AsyncJobStatusDTO.

```
ExportSpecTest > Standard Export Tests > should purge import records FAILED
    java.lang.IllegalArgumentException: Expected the field `resultData` to be a primitive type in the JSON string but got `{"created":1662061459887,"updated":1662061459887,"id":"ff80818182fa8d340182fa9539af3072","status":"SUCCESS","statusMessage":"test_owner-97717E27 file imported forcibly.","fileName":"8991662c-124e-41c8-a629-bbcb6e3c87f9-export-11672463340225012102.zip","generatedBy":"admin","generatedDate":1662061459092,"upstreamConsumer":{"created":1662061459887,"updated":1662061459887,"id":"ff80818182fa8d340182fa9539af3073","uuid":"8991662c-124e-41c8-a629-bbcb6e3c87f9","name":"test_consumer-0413CED7","type":{"id":"1003","label":"candlepin","manifest":true},"ownerId":"ff80818182fa8d340182fa9538423045","contentAccessMode":"org_environment","webUrl":"cdn-name80224C92","apiUrl":"https://cdn.test.com"}}`
        at org.candlepin.dto.api.v1.AsyncJobStatusDTO.validateJsonObject(AsyncJobStatusDTO.java:670)
        at org.candlepin.dto.api.v1.AsyncJobStatusDTO$CustomTypeAdapterFactory$1.read(AsyncJobStatusDTO.java:695)
        at org.candlepin.dto.api.v1.AsyncJobStatusDTO$CustomTypeAdapterFactory$1.read(AsyncJobStatusDTO.java:685)
        at com.google.gson.TypeAdapter$1.read(TypeAdapter.java:199)
        at com.google.gson.Gson.fromJson(Gson.java:1058)
        at com.google.gson.Gson.fromJson(Gson.java:1016)
        at com.google.gson.Gson.fromJson(Gson.java:959)
        at org.candlepin.JSON.deserialize(JSON.java:246)
        at org.candlepin.ApiClient.deserialize(ApiClient.java:847)
        at org.candlepin.ApiClient.handleResponse(ApiClient.java:1055)
        at org.candlepin.ApiClient.execute(ApiClient.java:979)
        at org.candlepin.resource.JobsApi.getJobStatusWithHttpInfo(JobsApi.java:541)
        at org.candlepin.resource.JobsApi.getJobStatus(JobsApi.java:519)
        at org.candlepin.spec.bootstrap.client.api.JobsClient.waitForJob(JobsClient.java:66)
        at org.candlepin.spec.bootstrap.client.api.JobsClient.waitForJob(JobsClient.java:107)
        at org.candlepin.spec.ExportSpecTest$StandardExporter.shouldPurgeImportRecords(ExportSpecTest.java:297)

```

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
```diff
@@ -0,0 +1,206 @@
...
+        if (path == null || path.isEmpty()) {
+            throw new IllegalArgumentException("Path is null or empty");
+        }
+
+        if (type == null) {
+            throw new IllegalArgumentException("Type is null or empty");
+        }
+
+        ZipEntry entry = manifest.getEntry(path);
+        try (InputStream istream = manifest.getInputStream(entry)) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Looks like we get a NPE in `manifest.getInputStream()`. We should check that entry is found based on the provided path and not null before creating the InputStream!

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/ExportUtil.java
JoshAlbrecht23: This is for the getFilesInDirectory method and so I was wanting to return all files in the directory path that is provided. Alternatively we could get rid of the getFilesInDirectory method an just verify that specific expected files exists by fetching the ZipEntries like you said, but we can't validate that an unexpected file is not added to the directory. Maybe that isn't much of a concern though, but I figured it was a nice to have.

The goal of the isEntryInDirectory was to prevent returning ZipEntry when the file exists in a sub directory of the expected path:
expected directory path: `/path/to/`
file in sub directory `/path/to/ugh/file.txt`

JoshAlbrecht23: Well, really using the export's stream and filtering on the start of the file name matching the path will give us all the files in that directory and sub directories and so we can just assert on that collection and we can tell if an unexpected file exists or an expected file does not exists! Sorry @Ceiu, I think you said this in another comment and I was just over thinking things.  I removed the isEntryInDirectory method.

```
export.stream()
  .filter(entry -> !entry.isDirectory() && entry.getName().startsWith(path))
```

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
Ceiu: This looks like something is wrong in the spec file or the client generation. That field should be a generic string at the API level. Before we work around it, we should check if there's something we can fix there to make it so we don't need to give it special attention.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,649 @@
...
+import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+@SpecTest
+class ExportSpecTest {
+    private static final String EXPORT_JOB_KEY = "ExportJob";
+    private static final String RECORD_CLEANER_JOB_KEY = "ImportRecordCleanerJob";
+    private static final String UNDO_IMPORTS_JOB_KEY = "UndoImportsJob";
+    /** The following are directory location paths within a manifest */
```
Ceiu, line 101: This should use the one-line comment format so Javadoc and IDE doc parsers don't attach this line to the EXPORT_PATH declaration.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Request.java
```diff
@@ -75,6 +77,7 @@ public Request(org.candlepin.ApiClient client) {
         this.pathParams = new HashMap<>();
         this.headers = new HashMap<>();
         this.queryParams = new HashMap<>();
+        this.formParams = new HashMap<>();
```
Ceiu, line 80: This isn't cleared in the reset method.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,649 @@
...
+
+            initializeData();
+
+            manifest = createExport(client, consumer.getUuid(), cdn);
+            export = ExportUtil.getExportArchive(manifest);
+        }
+
+        @AfterAll
+        void afterAll() throws Exception {
+            export.close();
```
Ceiu, line 177: Make sure to check that this is null in the event an exception is thrown in `createExport`.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,649 @@
...
+                String actualCdnLabel = root.get("cdnLabel").asText();
+                assertEquals(cdn.getLabel(), actualCdnLabel);
+            }
+        }
+
+        @Test
+        @DisplayName("should not include consumer json in entitlements")
+        void shouldNotIncludeConsumerJsonInEntitlements() throws Exception {
+            int target = ENTITILEMENTS_PATH.length() - 1;
+            AtomicInteger count = new AtomicInteger(0);
```
Ceiu, line 220: Is this a workaround for parallel streams, or did the compiler complain about it a non-final variable being referenced from the closure?

If it's the former, we could avoid it with a call to `.sequential()` before the filters (though I'm pretty sure `.stream()` on a collection is sequential already).

If it's the latter, a cleaner way to do this would be to drop `.forEach` in exchange for the non-terminal `.peek`:

```
int count = export.stream()
  .sequential() // probably optional/extraneous
  .filter(...)
  .filter(...)
  .peek(entry -> { ... })
  .count();

assertEquals(6, count);
```

This can be applied to all of the methods using this.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,649 @@
...
+                        .setPathParam("owner_key", importOwner.getKey())
+                        .addQueryParam("force", "SIGNATURE_CONFLICT")
+                        .addQueryParam("force", "MANIFEST_SAME")
+                        .addHeader("Accept", "application/json")
+                        .addHeader("Content-Type", "multipart/form-data")
+                        .setFormParam(manifest.getName(), manifest)
+                        .execute();
+
+                    assertEquals(200, response.getCode());
+                    JsonNode root = ApiClient.MAPPER.readTree(response.getBody());
```
Ceiu, line 316: Hrmm... I didn't expect this to be needed so soon. This could be added as part of the Response object directly as another method for fetching the body:

```
    /**
     * Attempts to process the body as a JSON-formatted UTF-8 string. If the body cannot be
     * deserialized into a JSON object, this method throws an exception.
     *
     * @throws JsonDeserializationException
     *  if the body cannot be deserialized into a JSON object
     *
     * @return
     *  a JsonNode representing the body of the response
     */
    public JsonNode deserialize() {
        try {
            return ApiClient.MAPPER.readTree(this.getBody());
        }
        catch (IOException e) { // Impl note: JsonProcessingException *is* an IOException
            throw new JsonDeserializationException(e);
        }
    }
```

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,649 @@
...
+                .filter(entry -> entry.getName().startsWith(PRODUCTS_PATH))
+                .filter(entry -> entry.getName().lastIndexOf('/') == target)
+                .filter(entry -> entry.getName().endsWith(".json"))
+                .forEach(entry -> {
+                    try {
+                        productCount.incrementAndGet();
+                        String name = new File(entry.getName()).getName().replace(".json", "");
+                        ProductDTO expected = productIdToProduct.get(name);
+                        assertNotNull(expected);
+                        ProductDTO actual = ExportUtil
```
Ceiu, line 393: No check on the output of this call. Should add an assertNotNull following it.

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1549,7 +1549,7 @@ public AsyncJobStatusDTO importManifestAsync(
     }
 
     @Override
-    public CandlepinQuery<ImportRecordDTO> getImports(
+    public Iterable<ImportRecordDTO> getImports(
```
Ceiu, line 1552: :+1:

Ceiu (changes requested):

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,649 @@
...
+        String cdnLabel = cdn == null ? null : cdn.getLabel();
+        String cdnName = cdn == null ? null : cdn.getName();
+        String cdnUrl = cdn == null ? null : cdn.getUrl();
+        File export = apiClient.consumers().exportData(consumerUuid, cdnLabel, cdnName, cdnUrl);
+        export.deleteOnExit();
+
+        return export;
+    }
+
+    private File createExportAsync(ApiClient apiClient, String consumerUuid)
```
nikosmoum, line 548: @Ceiu @JoshAlbrecht23 This and the above method (createExport) look generic enough to me to be added to the ExporterUtil class for use by the import tests that are to be imported later

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,649 @@
...
+            assertEquals(expectedNumberOfCerts.get(), actualNumberOfCerts.get());
+        }
+
+        @Test
+        @DisplayName("should not schedule non cron tasks")
+        public void shouldNotScheduleNonCronTasks() throws Exception {
+            assertForbidden(() -> jobsApi.scheduleJob(UNDO_IMPORTS_JOB_KEY));
+        }
+
+        private void initializeData() throws Exception {
```
nikosmoum, line 448: @Ceiu @JoshAlbrecht23 Even though I understand we don't need an 'exporter' abstraction in the same way we had in the ruby tests, we still have about 6 `import_*.rb` spec tests to be migrated that need this kind of data created in a manifest for importing. Can we make this generic and expose it for reuse? It's just too large for copy and pasting

JoshAlbrecht23: We could have an ExportFactory that has a builder that lets you define what should be included in the export. Then have the ExportFactory be able to return the objects that were created so that you can assert against the export. That way you can use the ExportFactory in both the ExportSpecTests and the ImportSpecTests and can easily create an export based on just what you need in each individual test.

```
ExportFactory factory = new ExportFactory.Builder()
    .withVirtualProductPool()
    .withDerivedProduct()
    .build();

ZipFile export = factory.createExport();

```

nikosmoum: While it could work, this might be a bit too overengineered for what we need. I agree with @Ceiu now that we should probably let this be as is, and the import spec tests will define their own data, because we really don't need everything from this initializeData method. Especially since most of those import_*.rb files have only 1-2 tests each, so they need minimal amounts of data in the manifest, and adding an abstraction layer would just be too much. So, tl;dr, disregard my comment!

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
nikosmoum: Looking at this again, the createExport method is too small to have value porting.
The createExportAsync does not seem to be used anywhere, for now. The ruby tests were all running twice, once using the async and once using the sync endpoints, but I don't think need that for now. Looking at hosted logs and katello code, the async has never been used. Given that, I think we should do what @Ceiu proposed yesterday, which is to leave most tests here to continue using the sync endpoint, but make only one of them use the async one, to make sure it's functional for now. (in the mean time we can start discussions on getting Satellite to using the async one)

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -0,0 +1,649 @@
...
+        assertEquals(readOnlyConsumer.getName(), actualConsumer.getName());
+
+        assertEquals(owner.getId(), actualConsumer.getOwner().getId());
+        assertEquals(owner.getKey(), actualConsumer.getOwner().getKey());
+        assertEquals(owner.getDisplayName(), actualConsumer.getOwner().getDisplayName());
+    }
+
+    @Nested
+    @DisplayName("Standard Export Tests")
+    @TestInstance(Lifecycle.PER_CLASS)
```
Januson, line 134: Why the test needs the PER_CLASS lifecycle? It can break parallelization by making tests share instance properties.

Januson (changes requested):

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: Good point! The PER_CLASS lifecycle should not be necessary and we can remove it.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: I was able to fix and validate the issue by changing the resultData field type to an object in the candlepin-api-spec.yaml file. This would also be nice because we can avoid adding all the multipart/form-data functionality to the Request object. 

https://github.com/candlepin/candlepin/blob/master/api/candlepin-api-spec.yaml#L8239

A side note, I also ran into this issue with the async export job as the result data is an ExportResult object instead of String. We can add a ExportResultDTO object to candlepin-api-spec.yaml and use the ObjectMapper to deserialize the resultData field to the ExportResultDTO.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: I added the test 'shouldExportProductsAsync' to verify that the createExportAsync method is working properly!

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: Added

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: Good catch! Updated

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: Updated

api/candlepin-api-spec.yaml
```diff
@@ -8909,6 +8911,16 @@ components:
...
+    ExportResultDTO:
+      description: Represents the result of an export job
+      properties:
+        exportedConsumer:
+          type: string
+        exportId:
+          type: string
+        href:
+          type: string
+
```
JoshAlbrecht23, line 8923: I added the ExportResultDTO so that we can deserialize the ExportResult that returned in the resultData field on the AsyncJobStatusDTO object when we use the `exportDataAsync` api method.

api/candlepin-api-spec.yaml
```diff
@@ -8236,7 +8238,7 @@ components:
               type: string
               example: "/jobs/ff808081554a3e4101554a3e9033005d"
             resultData:
-              type: string
+              type: object
```
JoshAlbrecht23, line 8241: Updated to object type as some of the async jobs return objects instead of a string in the resultData field and this causes deserialization issues with the OpenApi apis.

Ceiu (approved):
Tentative approval pending further investigation of that spec change on the async job status DTO.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
JoshAlbrecht23: @Ceiu I did some validating in Candlepin 4.0 to verify that the changes to the resultData field are passive. I triggered an async export job using the `/consumers/{consumer_uuid}/export/async` endpoint and then got the job status using the `/jobs/{job_id}` endpoint and validated that the resultData field has an object and not a string.

CP 4.0job status result for async export:
```
{
  "created" : "2022-09-13T20:58:30+0000",
  "updated" : "2022-09-13T20:58:30+0000",
  "id" : "ff80818183389410018338a5720e0034",
  "name" : "Export Manifest",
  "group" : null,
  "origin" : "candlepin-el8.example.com",
  "executor" : "candlepin-el8.example.com",
  "principal" : "admin",
  "state" : "FINISHED",
  "previousState" : "RUNNING",
  "startTime" : "2022-09-13T20:58:30+0000",
  "endTime" : "2022-09-13T20:58:30+0000",
  "attempts" : 1,
  "maxAttempts" : 1,
  "statusPath" : "/jobs/ff80818183389410018338a5720e0034",
  "resultData" : {"exportedConsumer":"35398b97-11f2-428c-a3e5-fa286bdaddc8","exportId":"ff80818183389410018338a572dd0035","href":"/consumers/35398b97-11f2-428c-a3e5-fa286bdaddc8/export/ff80818183389410018338a572dd0035"},
  "key" : "ExportJob"
}
```

JoshAlbrecht23:
> Tentative approval pending further investigation of that spec change on the async job status DTO.

Added the investigation to comment: https://github.com/candlepin/candlepin/pull/3509#discussion_r970083047

Ceiu:
@Januson You good with the changes here?

PR #3512, opened 2022-08-15
Use formatter instead of toString for date field [ENT-5301]

wottop:
toString truncates the seconds if the date hits an exact minute.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobSpecTest.java
@@ -45,6 +45,7 @@ import org.junit.jupiter.api.Test;
 
 import java.time.Instant;
 import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


nikosmoum (approved):

spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobSpecTest.java
```diff
@@ -114,8 +115,9 @@ public void shouldDeleteInactiveConsumers() throws Exception {
         final String inactiveConsumer1Uuid = inactiveConsumer.getUuid();
         assertGone(() -> consumerApi.getConsumer(inactiveConsumer1Uuid));
         // Verify that the inactive consumer has been moved to the deleted_consumers table.
+        DateTimeFormatter fmt = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'");
         List<DeletedConsumerDTO> deletedConsumers = deletedConsumerApi
-            .listByDate(inactiveConsumer.getCreated().toString());
+            .listByDate(fmt.format(inactiveConsumer.getCreated()));
```
JoshAlbrecht23, line 120: I think we should be able to use the the ISO-8601 static formatters that DateTimeFormatter already has instead of creating a new one. I believe that the ISO_DATE_TIME formatter should work since it includes the offset data that is included in the OffsetDateTime object that is returned from `inactiveConsumer.getCreated()`.

`DateTimeFormatter.ISO_DATE_TIME.format(inactiveConsumer.getCreated());`

JoshAlbrecht23 (commented):
Comment on using a static ISO formatter.

JoshAlbrecht23 (approved):

PR #3515, opened 2022-08-17
Update pool creation for Jenkins

wottop:
Not sure why, but the original pool creation in these tests causes Jenkins to fail off and on.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/jobs/JobScheduleSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/jobs/JobScheduleSpecTest.java
@@ -32,7 +32,8 @@ import org.candlepin.spec.bootstrap.client.ApiClients;
 import org.candlepin.spec.bootstrap.client.SpecTest;
 import org.candlepin.spec.bootstrap.client.api.JobsClient;
 import org.candlepin.spec.bootstrap.data.builder.Owners;
-import org.candlepin.spec.bootstrap.data.util.StringUtil;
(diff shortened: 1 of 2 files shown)


spec-tests/src/test/java/org/candlepin/spec/jobs/JobScheduleSpecTest.java
```diff
@@ -80,19 +81,14 @@ public void shouldPurgeExpiredPools() throws Exception {
...
         OwnerDTO owner = ownerApi.createOwner(Owners.random());
-        ProductDTO product = new ProductDTO();
-        product.setId(StringUtil.random("ID"));
-        product.setName(StringUtil.random("Test Product"));
+        ProductDTO product = Products.random();
         product = client.ownerProducts().createProductByOwner(owner.getKey(), product);
 
-        PoolDTO pool = new PoolDTO();
-        pool.setProductId(product.getId());
-        pool.setSubscriptionId(StringUtil.random("source_sub"));
```
nikosmoum, line 90: Not setting the subscriptionId field for both these tests has fixed the issue in this case, but we should preemptively make a change to not allow this kind of flakiness to happen in the future: we can delete the `StringUtil.random` method and rename `StringUtil.randomSuffix` into `StringUtil.random`, and use that throughout the code, since it will produce a prefix with shorter/limited length

wottop: I will make a separate branch for this.

PR #3518, opened 2022-08-18
2118759: Leave manifest consumer identity certificate

wottop:
Plus a bit of checkstyle cleanup

### DIFF
--- a/src/test/java/org/candlepin/dto/api/v1/GuestIdArrayElementTranslatorTest.java
+++ b/src/test/java/org/candlepin/dto/api/v1/GuestIdArrayElementTranslatorTest.java
@@ -75,9 +75,9 @@ public class GuestIdArrayElementTranslatorTest extends AbstractTranslatorTest<Gu
             assertEquals(source.getId(), dto.getId());
             assertEquals(source.getGuestId(), dto.getGuestId());
             assertEquals(source.getCreated(), dto.getCreated() != null ?
-                new Date(dto.getCreated().toInstant().toEpochMilli()) : null);
+                Date.from(dto.getCreated().toInstant()) : null);
             assertEquals(source.getUpdated(), dto.getUpdated() != null ?
-                new Date(dto.getUpdated().toInstant().toEpochMilli()) : null);
+                Date.from(dto.getUpdated().toInstant()) : null);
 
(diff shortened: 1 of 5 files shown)


nikosmoum (approved):

src/test/java/org/candlepin/dto/api/v1/GuestIdArrayElementTranslatorTest.java
```diff
@@ -75,9 +75,9 @@ protected void verifyOutput(GuestId source, GuestIdDTOArrayElement dto, boolean
             assertEquals(source.getId(), dto.getId());
             assertEquals(source.getGuestId(), dto.getGuestId());
             assertEquals(source.getCreated(), dto.getCreated() != null ?
-                new Date(dto.getCreated().toInstant().toEpochMilli()) : null);
+                (new Date(dto.getCreated().toInstant().toEpochMilli())) : null);
```
Ceiu, line 78: We can just use .getTime() to get the milliseconds straight from the date object

Ceiu (changes requested):
We should also patch up the place where the NPE occurs so that it either completes when the cert is null, or throws an exception that is more indicative of the problem.

src/test/java/org/candlepin/dto/api/v1/GuestIdArrayElementTranslatorTest.java
wottop: created is an OffsetDateTime. Not sure I can reduce what is here.

wottop:
Added catch and log message for null cert.

src/test/java/org/candlepin/dto/api/v1/GuestIdArrayElementTranslatorTest.java
Ceiu: Ahh, my mistake; didn't process that this was from the DTO. Also didn't notice that we're getting millis to convert to a date. In that case, we can skip the millis bit entirely:

```
Date.from(dto.getCreated().toInstant())
```

It's too bad there's no shared interface between Instant and OffsetDateTime. The amount of `.toInstant()`s we're going to have around the code is going to be high for the foreseeable future.

Ceiu (approved):

PR #3519, opened 2022-08-18
[M] Added same-ID entity version collision resolution (ENT-5324)

Ceiu:
- The versioned entity update logic will now clear the entity version
  for an existing entity on collision when the colliding entities
  have the same entity ID

### DIFF
--- a/src/main/java/org/candlepin/controller/ContentManager.java
+++ b/src/main/java/org/candlepin/controller/ContentManager.java
@@ -32,8 +32,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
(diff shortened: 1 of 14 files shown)


Ceiu:
A great deal of care should be taken when reviewing and evaluating this code. Getting the versioning wrong only leads to pain, so please pose any questions in logic, no matter how minor.

src/main/java/org/candlepin/controller/ContentManager.java
```diff
@@ -156,20 +156,36 @@ public Content createContent(Owner owner, ContentInfo contentData) {
...
+                    this.ownerContentCurator.mapContentToOwner(candidate, owner);
+                    return candidate;
                 }
 
-                log.debug("Converging content with existing version: {} => {}", entity, alt);
+                // If we have a version collision, and the entity IDs are the same, there's likely
+                // some shenanigans going on. Rather than halting all behavior, let's just clear
+                // the old entity's version so we start mapping to the new one.
+                // If we have a collision where the contents are actually different, we won't
+                // fail at this point (but we *will* fail), and we won't be able to detect it.
```
nikosmoum, line 184: Where (and in what way) would we fail if this happened?

Ceiu: When a flush is performed at some later point, Hibernate will throw an exception. I've been wondering if the restriction should fall back to unique on ID + entity version to avoid this type of thing all together.

PR #3527, opened 2022-08-23
[M] Added an API for performing generic requests

Ceiu:
- Added the Request and Response objects for building generic
  requests without needing a specific API class or method

### DIFF
--- a/spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Request.java
+++ b/spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Request.java
@@ -1,0 +1,332 @@
+/**
+ * Copyright (c) 2009 - 2022 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.spec.bootstrap.client.request;
+
+import org.candlepin.ApiException;
+import org.candlepin.Pair;
+import org.candlepin.spec.bootstrap.client.ApiClient;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import okhttp3.Call;
+
+
+
+/**
+ * The Request class provides a fluent-style builder API for creating generic requests. The Request
+ * instance needs an ApiClient instance to perform the actual work, and can be provided either
+ * directly during construction, or can be pulled automatically from a client-level ApiClient
+ * object.
+ */
+public class Request {
+
+    private static final Pattern PATH_PARAM_REGEX = Pattern.compile("\\{([^}]+)\\}");
+
+    private final org.candlepin.ApiClient client;
+
+    private String basePath;
+    private String endpoint;
+    private String method;
+    private String body;
+
+    private Map<String, String> pathParams;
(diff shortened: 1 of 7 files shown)


spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Request.java
```diff
@@ -0,0 +1,331 @@
...
+     * Builds a new Request instance which will use the specified ApiClient to perform the actual
+     * request upon execution.
+     *
+     * @param client
+     *  the org.candlepin.ApiClient instance to use to execute the request
+     *
+     * @throws IllegalArgumentException
+     *  if the provided client is null
+     */
+    public Request(org.candlepin.ApiClient client) {
```
JoshAlbrecht23, line 67: We can use just `ApiClient client` instead of using the full package

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Request.java
```diff
@@ -0,0 +1,332 @@
...
+     */
+    public static Request from(ApiClient client) {
+        if (client == null) {
+            throw new IllegalArgumentException("client is null");
+        }
+
+        return new Request(client.getApiClient());
+    }
+
+    // TODO: add a .to method which accepts an endpoint enum which define the HTTP method and endpoint
```
JoshAlbrecht23, line 100: Is this TODO intentional? Just want to double check and make sure this wasn't a method that you wanted to add in this PR and just forgot to implement it.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Response.java
```diff
@@ -0,0 +1,175 @@
...
+     *  if the response code, body, or headers cannot be read or processed
+     */
+    public Response(okhttp3.Response response) {
+        if (response == null) {
+            throw new IllegalArgumentException("response is null");
+        }
+
+        try {
+            this.code = response.code();
+            this.body = response.body().bytes();
```
JoshAlbrecht23, line 58: Is it possible for the `response.body()` to be null?

JoshAlbrecht23 (changes requested):
One change request and a couple questions

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Request.java
Ceiu: Different ApiClients (`org.candlepin.ApiClient` and `org.candlepin.spec.bootstrap.client.ApiClient`). Due to the naming conventions used, one of the two must be fully qualified, and I opted to FQ the shorter of the two.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Request.java
Ceiu: Intentional, but won't be done anytime soon. If there's a push for how I feel all of this should actually work [1], then we'll investigate how much effort it'd be to add it.

[1] Basically, the spec file should generate some enums which themselves implement a shared interface (ApiEndpoint?). Each enum would provide the basic info for the endpoint: path, method, and some help info in string form for a nice .toString output. If such a thing comes to exist, the code for the request would look like the following:

```
Request.from(client)
  .to(OwnerContent.LIST_CONTENT) // GET /owners/{owner_key}/content; would set both fields
  .setPathParam("owner_key", owner.getKey())
  ...
```

I'd want to sort out something to codify the path params somehow, but that's potentially non-trivial (enums with enums?).

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Response.java
Ceiu: Hrmm... maybe. That's largely copied from what the generated code does, and the [docs](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response/body/) suggest it shouldn't be null in the cases we use it.

We could throw a check in there just to be safe, though.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Request.java
JoshAlbrecht23: Oh, that's my bad! I didn't realize they were different ApiClients.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Response.java
Ceiu: Done

JoshAlbrecht23 (approved):

PR #3528, opened 2022-08-23
Slow docker image builds

Lorquas:
Currently there is some bug with Docker (most likely our configuration of it) where it CPU limits commands run in the container.  When yum inside the container hits 100% CPU usage it hangs for some undetermined time and makes each package install VERY slow.  I've tried manually opening up Docker to use all CPUs and other available resources but this does not prevent any one CPU from hitting full usage.

Conveniently, podman does not have any of this problem, and since the build stage of candlepin tests don't strictly require docker, I'm doing a simple fix of using that for the builds only.

Note that this doesn't include a fix for the other build script in `docker/gating-test-images/build.sh`  and we will likely hit that problem there also.  This PR makes the job run much faster but we might need to discuss what to do in that script.

This allows for the fixes in: https://github.com/candlepin/candlepin-jobs/pull/85

### DIFF
--- a/docker/build-images
+++ b/docker/build-images
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 unset CDPATH
-SCRIPT_NAME=$( basename "$0" )
-SCRIPT_HOME=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+SCRIPT_NAME=$(basename "$0")
+SCRIPT_HOME=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 CURRENT_DIR=$PWD
 
 # defaults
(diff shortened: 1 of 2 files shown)


docker/build-images
```diff
@@ -29,61 +29,78 @@ OPTIONS:
...
-  CP_VERSION="$( git describe | cut -d- -f 2 )"
-  for IMG in ${TAG_IMAGES}; do
-    docker tag $REGISTRY/$IMG:latest $REGISTRY/$IMG:$CP_VERSION
-  done
+  python -c 'import yaml,sys;y=yaml.safe_load(sys.stdin); print(" ".join(y["services"].keys()))' <$SCRIPT_HOME/docker-compose-build.yml
 }
 
 echo "Building images..."
 cd $SCRIPT_HOME
-docker-compose -f docker-compose-build.yml build $BUILD_ARGS $IMAGE && tag_images
```
nikosmoum, line 79: If we're not longer using docker-compose-build.yml let's delete that file entirely. the `get_image_names` method above that is also using it can just list them hardcoded (it's just 2, and we rarely if ever change them)

Lorquas: Done

PR #3529, opened 2022-08-24
Convert entitlement_certificate_v3 to java spec

wottop:


spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
```diff
@@ -17,121 +17,493 @@
...
+    }
+
+    /**
+     * This test covers the case where the system supports 3.0 certs, but
+     * the server is creating 3.4 certs, and the product contains attributes
+     * supported by 3.0.
+     * @throws Exception
+     */
+    @Test
+    @DisplayName("generated a version 3.4 certificate when requesting a 3.0 certificate")
```
nikosmoum, line 200: Let's follow the convention to start with `should/should not` for all test method/displaynames in the file like we've been doing elsewhere

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
```diff
@@ -17,121 +17,493 @@
...
+
+    @Test
+    @DisplayName("generated a version 3.4 certificate")
+    public void shouldGenerateThreeFourCert() throws Exception {
+        consumerApi.bindProduct(system.getUuid(), product.getId());
+        EntitlementDTO ent = consumerApi.listEntitlements(system.getUuid()).get(0);
+        String value = CertificateUtil.standardExtensionValueFromCert(
+            ent.getCertificates().iterator().next().getCert(), "1.3.6.1.4.1.2312.9.6");
+        assertEquals("3.4", value);
+        consumerApi.unbindAll(system.getUuid());
```
nikosmoum, line 222: If the `system` is created in the `beforeAll`, why do you need to unbind from it at the end of each test? It should be a separate consumer for each one

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
```diff
@@ -17,121 +17,493 @@
...
 
     @BeforeAll
     public void beforeAll() throws ApiException {
         client = ApiClients.admin();
         ownerApi = client.owners();
-        owner = ownerApi.createOwner(Owners.random());
         ownerContentApi = client.ownerContent();
         ownerProductApi = client.ownerProducts();
         consumerApi = client.consumers();
+        hostedTestApi = client.hosted();
```
nikosmoum, line 98: Can we move all these clients in the BeforeEach, and delete the BeforeAll, to minimize the state sharing between tests?

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
```diff
@@ -17,121 +17,493 @@
...
+            ProvidedProductDTO engProvidedProduct = (new ProvidedProductDTO())
+                .productName(engProduct.getName())
+                .productId(engProduct.getId());
+            pool = Pools.random(mktProduct)
+                .contractNumber("12345")
+                .accountNumber("6789")
+                .orderNumber("order1")
+                .providedProducts(Set.of(engProvidedProduct));
+            pool = ownerApi.createPool(owner.getKey(), pool);
+            assertThat(pool.getBranding()).hasSize(1);
```
nikosmoum, line 386: shouldn't this be asserted outside the if/else to make sure it's checked for both hosted/standalone?

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
wottop: Carry over from the ruby

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateV3SpecTest.java
Januson: Do we care about the tests sharing a client? It should be immutable and thread safe.

wottop: Now that they are in the BeforeEach, there should not be a sharing issue

nikosmoum (approved):

PR #3531, opened 2022-08-24
[M] Migrated owner_product_resource_spec to java (ENT-5198)

JoshAlbrecht23:
Migrated the spec tests from owner_product_resource_spec to java and included the test 'should throw exception while inserting product without name' from owner_content_resource_spec.rb.

There are two tests dependent on ENT-5308 (Add and apply global paging params to OpenApi spec) and so I added a TODO comment on them and added the ruby code commented out in the test place holder.


spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/request/Request.java
```diff
@@ -157,7 +157,7 @@ public Request setBasePath(String path) {
      * endpoint, minus the common components of the path that would be captured by the base path.
      * <p></p>
      * The endpoint path may optionally include parameters which will be replaced by the values set
-     * in any path parameters later. For example, in the path "owners/{owner_key}/products", the
+     * in any path parameters later. For example, in the path "/owners/{owner_key}/products", the
```
JoshAlbrecht23, line 160: It looks like we need the forward slash in the path, so I figured it would be a good idea to add that to the example in the javadoc

Ceiu: :+1:

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -0,0 +1,799 @@
...
+        sleep(1000);
+        ProductDTO prod3 = ownerProductApi.createProductByOwner(ownerKey, Products.random()
+            .id("test_product-3"));
+
+        Response response = Request.from(client)
+            .setPath(OWNER_PRODUCTS_PATH)
+            .setPathParam("owner_key", ownerKey)
+            .addQueryParam("page", "1")
+            .addQueryParam("per_page", "1")
+            .addQueryParam("sort_by", "id")
```
Ceiu, line 404: We should specify the sort order here so we're not at the risk of the default changing breaking this test. Otherwise the test logic needs to be updated to be handle either order, or the test name needs to be updated to indicate the default ordering is part of the test.

spec-tests/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
     implementation libraries["jacksonJsr310"]
     implementation libraries["assertj"]
     implementation libraries["commonsCodec"]
+    implementation libraries["guava"]
```
Ceiu, line 27: What is this for?

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -0,0 +1,799 @@
...
+    @Test
+    @DisplayName("should lists all products in bulk fetch")
+    public void shouldListsAllProductsInBulkFetch() throws Exception {
+        OwnerDTO owner = ownerApi.createOwner(Owners.random());
+        String ownerKey = owner.getKey();
+        ProductDTO prod1 = ownerProductApi.createProductByOwner(ownerKey, Products.random());
+        ProductDTO prod2 = ownerProductApi.createProductByOwner(ownerKey, Products.random());
+        ProductDTO prod3 = ownerProductApi.createProductByOwner(ownerKey, Products.random());
+
+        List<ProductDTO> products = ownerProductApi.getProductsByOwner(ownerKey, new ArrayList<>());
```
Ceiu, line 304: Non-blocking comment: you can use `List.of()` if you just need a quick empty list, and don't mind the immutability.

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -0,0 +1,799 @@
...
+        assertEquals(3, products.size());
+        assertTrue(products.contains(prod1));
+        assertTrue(products.contains(prod2));
+        assertTrue(products.contains(prod3));
+
+        // Pick two products to use in a bulk get
+        products = ownerProductApi.getProductsByOwner(ownerKey, List.of(prod1.getId(), prod3.getId()));
```
Ceiu, line 311: These should be separate test cases -- one for specifying products and one that doesn't.

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -0,0 +1,799 @@
...
+        String ownerKey = owner.getKey();
+        ProductDTO prod1 = ownerProductApi.createProductByOwner(ownerKey, Products.random());
+        ProductDTO prod2 = ownerProductApi.createProductByOwner(ownerKey, Products.random());
+        ProductDTO prod3 = ownerProductApi.createProductByOwner(ownerKey, Products.random());
+
+        List<ProductDTO> products = ownerProductApi.getProductsByOwner(ownerKey, new ArrayList<>());
+        assertEquals(3, products.size());
+        assertTrue(products.contains(prod1));
+        assertTrue(products.contains(prod2));
+        assertTrue(products.contains(prod3));
```
Ceiu, line 308: The AssertJ bits that are included in the spec test can make this set of checks a bit nicer looking:

```
assertThat(products)
  .isNotNull()
  .hasSize(3)
  .containsOnly(prod1, prod2, prod3);
```

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -0,0 +1,799 @@
...
+        product.setDerivedProduct(derivedProduct);
+        product = hostedTestApi.createProduct(product);
+
+        return product;
+    }
+
+    private SubscriptionDTO createUpstreamSubscription(OwnerDTO owner, ProductDTO product)
+        throws ApiException {
+        SubscriptionDTO subscription = hostedTestApi.createSubscription(Subscriptions.random(owner, product));
+        AsyncJobStatusDTO job = ownerApi.refreshPools(owner.getKey(), false);
```
Ceiu, line 778: Be *VERY* careful here. Automating/hiding the refresh op lead to very problem we had with the first iteration of the hosted test API, which is a problem that still infects the spec tests to this day, some 7 years later. 
This method looks especially close to the very deprecated `create_pools_and_subscriptions` method from the ruby hostedtest_api.rb, which has side effects which are sometimes unexpected and hides or performs actions that the test author and maintainers need to be aware of if the test is to work as intended.

You'll need the upstream logic bits for testing locked entities (which I don't see at present), but I feel like this could be better represented/abstracted with a set of nested tests and the common setup.

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -0,0 +1,799 @@
...
+        ProductDTO derivedProvidedProduct =
+            Iterables.getOnlyElement(product.getDerivedProduct().getProvidedProducts());
+        assertBadRequest(() -> ownerProductApi
+            .deleteProductByOwner(owner.getKey(), derivedProvidedProduct.getId()));
+    }
+
+    @Test
+    @OnlyInHosted
+    @DisplayName("should return forbidden request when deleting derived provided product attached to sub")
+    public void shouldReturnForbiddenRequestWhenDeletingDerivedProvidedProductAttachedToSub()
```
Ceiu, line 543: This test can be generalized with custom products and pools.

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -0,0 +1,799 @@
...
+        OwnerDTO owner = ownerApi.createOwner(Owners.random());
+        ProductDTO product = createProductWithProvidedAndDerivedProduct(owner.getKey());
+        ProductDTO derivedProduct = product.getDerivedProduct();
+        assertBadRequest(() -> ownerProductApi.deleteProductByOwner(owner.getKey(), derivedProduct.getId()));
+    }
+
+    @Test
+    @OnlyInHosted
+    @DisplayName("should return forbidden request on attempt to delete derived product attached to sub")
+    public void shouldReturnForbiddenRequestOnAttemptToDeleteDerivedProductAttachedToSub()
```
Ceiu, line 518: This test can be generalized with custom products and pools -- no need for the hosted test API restriction.

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -0,0 +1,799 @@
...
+    }
+
+    @Test
+    @DisplayName("should return bad request when inserting product with null id")
+    public void shouldReturnBadRequestWhenInsertingProductWithNullId() throws Exception {
+        OwnerDTO owner = ownerApi.createOwner(Owners.random());
+        ProductDTO product = Products.random()
+            .id(null);
+
+        Request request = Request.from(client)
```
Ceiu, line 724: What is the `createProductByOwner` endpoint lacking that these two tests need?

Ceiu (changes requested):

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
```diff
@@ -38,4 +39,27 @@ public static String random(String base) {
...
+     *
+     * @param length
+     *  The length of the generated random string.
+     *
+     * @return
+     *  a randomly generated string.
+     *
+     * @throws IllegalArgumentException if length is less than or equal to 0.
+     */
+    public static String random(int length) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: I like this and should help with our random string collision issues. The test that is using the new method is OwnerProductResourceSpecTest.shouldReturnCorrectExceptionForConstraintViolations

spec-tests/build.gradle
JoshAlbrecht23: There were a lot of instances where we needed to get a single element out of a set and guava has a nice Iterables.getOnlyElement. Alternatively we could use a stream to get the element after we assert the set size is one, but the length is a little long.
`product.getBranding().stream().findAny().get()`

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
```diff
@@ -38,4 +39,27 @@ public static String random(String base) {
...
+        if (length <= 0) {
+            throw new IllegalArgumentException("Length must be larger than 0");
+        }
+        StringBuilder builder = new StringBuilder(length);
+        for (int i = 0; i < length; i++) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: I like this a lot better than the for loop! I also started a StringBuilder with the prefix if it is not null and appended the random suffix to it.

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
JoshAlbrecht23: I see! Updated to ascending sort order

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
JoshAlbrecht23: Updated to two separate tests: shouldListsAllProductsInBulkFetch and shouldListsAllSpecifiedProductsInBulkFetch

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
JoshAlbrecht23: I need to get use to using AssertJ package because I do really like how nice this looks. Updated

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
JoshAlbrecht23: I took a look and saw what you were doing in the owner-content spec suite and I added a nested test class that handles the creation of the upstream testing/locking.

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
JoshAlbrecht23: Updated

spec-tests/build.gradle
Ceiu: ack

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Branding.java
```diff
@@ -59,6 +60,20 @@ public static BrandingDTO random(String name) {
...
+    /**
+     * Builds a random BrandingDTO instance using the provided product's id.
+     *
+     * @param product
+     *  the product to use for populating BrandingDTO's product Id.
+     *
+     * @return a BrandingDTO instance with a randomly generated name and type using the product's id.
+     */
+    public static BrandingDTO random(ProductDTO product) {
+        BrandingDTO branding = random();
```
Ceiu, line 72: Non-blocking comment: this could be written as a two liner:


```
return random()
  .productId(product.getId());
```

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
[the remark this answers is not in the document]
Ceiu: ack

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
Ceiu: :+1:

Ceiu (changes requested):
Getting there; just a couple oversights and I think this is in a good place.

Is it safe to assume this is going to be rebased after the StringUtil stuff goes through? That may have an impact on this PR.

JoshAlbrecht23:
> Getting there; just a couple oversights and I think this is in a good place.
> 
> Is it safe to assume this is going to be rebased after the StringUtil stuff goes through? That may have an impact on this PR.

Yeah, I was planning on waiting for the StringUtil changes to get through first and rebasing this PR

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
JoshAlbrecht23: Created the JsonManipulationTests nested test class to have tests that serialize nulls, and then a set of tests that do no serialize the null values.

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
JoshAlbrecht23: Removed

src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -150,6 +150,18 @@ public ProductDTO addContent(String ownerKey, String productId, String contentId
     public ProductDTO createProductByOwner(String ownerKey, ProductDTO dto) {
         this.validator.validateCollectionElementsNotNull(
             dto::getBranding, dto::getDependentProductIds, dto::getProductContent);
+        if (dto.getId() == null) {
```
Ceiu, line 153: Few things here:
- Let's move the simple property checks to before the collection validation, also add whitespace between the checks so they don't look "connected" at a glance (current lack of whitespace matches our normal if-else syntax)
- We can combine the null and empty checks, as it'll reduce the number of strings to translate, and shrinks this block a bit.
- On the note of translations, let's try to use a string that's already present if possible, as it'll avoid making the translators have to translate the same message twice, or build up weird deviations if the verbiage we use doesn't lend itself well to the destination language.
- Don't punctuate the end of the error message, as it's not a complete sentence, and it looks funny when encapsulated. As far as the beginning capitalization goes... we're not at all consistent on that. Even looking at just my own error messages I've been inconsistent with that, so eh... follow your heart. 

Something like this should suffice:
```
if (dto.getId() == null || dto.getId().matches("^\\s*$")) {
  throw new BadRequestException(this.i18n.tr("product has a null or invalid ID"));
}

if (dto.getName() == null || ...
```

Ceiu (changes requested):
One last set of changes on the input validation side of things, and this is good with me

src/main/java/org/candlepin/resource/OwnerProductResource.java
JoshAlbrecht23: I checked the /po/keys.pot file for existing keys and it didn't look like there were any existing ones that we can use (I believe that's the correct file to check).  I like the message you provided for the invalid id and I did the same for the name. Those will be very reusable for other locations where product id and name are verified!

Ceiu (approved):

PR #3534, opened 2022-08-25
Migrate entitlement_certificate_spec to java spec

wottop:


Januson (changes requested):
Looks good. I added some suggestions. Apart from that, I am a bit worried how hosted conditions complicate some of the tests.

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateSpecTest.java
```diff
@@ -0,0 +1,474 @@
...
+
+    @Test
+    @DisplayName("multiple entitlement in excess will be deleted when " +
+        "existing subscription quantity is decreased")
+    public void deleteMultiEntWhenSubQuantityDecreased() throws Exception {
+        //  this entitlement makes the counts inconclusive
+        consumerApi.unbindAll(system.getUuid());
+        ProductDTO prod = Products.randomEng().addAttributesItem(
+            new AttributeDTO().name("multi-entitlement").value("yes"));
+        if (CandlepinMode.isHosted()) {
```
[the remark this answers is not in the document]
Ceiu: Yes. A lot of those tests are only merged into one because of the old `create_pool_and_subscription` method which did everything in one and hid the details. From what I recall when I attempted to rip it out, very few tests actually *needed* the hosted test bits, and even fewer had reason to change test operation depending on the mode.

Breaking up the tests will, at the very least, make it obvious which is which, and will probably also highlight areas where the differentiation isn't necessary anymore.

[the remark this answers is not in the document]
JoshAlbrecht23: Yeah, I agree on breaking up the tests based on hosted/standalone. I had to do the same for owner product resource spec tests and it makes the test more clear on what is being tested.

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateSpecTest.java
```diff
@@ -0,0 +1,474 @@
...
+    private OwnerClient ownerApi;
+    private EntitlementsApi entitlementsApi;
+    private OwnerContentApi ownerContentApi;
+    private OwnerProductApi ownerProductApi;
+    private ConsumerClient consumerApi;
+    private HostedTestApi hostedTestApi;
+    private ProductDTO monitoring;
+    private PoolDTO pool;
+    private SubscriptionDTO subscription;
+    ConsumerDTO system;
```
JoshAlbrecht23, line 79: Should this be private?

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateSpecTest.java
```diff
@@ -0,0 +1,474 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+import java.util.List;
+import java.util.stream.Collectors;
+
+
+@SpecTest
+public class EntitlementCertificateSpecTest {
```
JoshAlbrecht23, line 66: We will want to match the naming standard when it comes to the test name and display name:
```
@Test
@DisplayName("should retrieve content")
public void shouldRetrieveContent() {
    // Test logic
}
```

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateSpecTest.java
```diff
@@ -0,0 +1,474 @@
...
+            pool = Pools.random(coolApp);
+            pool = ownerApi.createPool(owner.getKey(), pool);
+        }
+        consumerApi.bindProduct(system.getUuid(), coolApp.getId());
+        List<CertificateDTO> oldCerts = consumerApi.fetchCertificates(system.getUuid());
+        assertThat(oldCerts).hasSize(2);
+        AsyncJobStatusDTO result = entitlementsApi.regenerateEntitlementCertificatesForProduct(
+            coolApp.getId(), false);
+        if (result != null) {
+            client.jobs().waitForJob(result.getId());
```
JoshAlbrecht23, line 177: We may want to validate that the job was successfully completed

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateSpecTest.java
```diff
@@ -0,0 +1,474 @@
...
+        consumerApi.unbindAll(system.getUuid());
+        ProductDTO prod = Products.randomEng().addAttributesItem(
+            new AttributeDTO().name("multi-entitlement").value("yes"));
+        if (CandlepinMode.isHosted()) {
+            prod = hostedTestApi.createProduct(prod);
+            subscription = Subscriptions.random(owner, prod);
+            subscription = hostedTestApi.createSubscription(subscription);
+            AsyncJobStatusDTO refresh = ownerApi.refreshPools(owner.getKey(), false);
+            if (refresh != null) {
+                client.jobs().waitForJob(refresh.getId());
```
JoshAlbrecht23, line 226: We should validate that the job successfully completed. It looks like this will need to be done through out where the `waitForJob` method is being used.

JoshAlbrecht23 (changes requested):
Some changes that can be made throughout:
- Breaking up the tests based on hosted/standalone.
- Updating the test name and display name to match the standard.
- Validating jobs are completing successfully by asserting against the job status.

JoshAlbrecht23 (changes requested):
Just a small change with the visibility of two methods

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateSpecTest.java
```diff
@@ -0,0 +1,634 @@
...
+        assertEquals(oldCerts.size(), newCerts.size());
+        // System has two certs, but we only regenerated for one product, so the
+        // other serial should have remained the same:
+        assertThat(newCerts)
+            .filteredOn(oldCerts::contains)
+            .hasSize(1);
+    }
+
+    @Test
+    @OnlyInStandalone
```
nikosmoum, line 197: I know you've duplicated the tests and using the OnlyInStandalone/OnlyInHosted annotations in each version of the test because the ruby version was using the create_pool_and_subscription method, but we really don't need to do that.

We should only be using the OnlyInStandalone/OnlyInHosted when we are testing functionality that is different between the two modes (e.g. testing refresh, or locked vs custom pools/products, or hosted-only legacy bonus pools).

In the case for this spec test suite, we're just testing entitlement certs behaviour which should be the same in both hosted and standalone, so I'd say just remove the 'hosted' version of the tests,  just use the normal createPool/createProduct APIs, and remove the annotations so that the tests will run in both modes anyway

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateSpecTest.java
```diff
@@ -0,0 +1,634 @@
...
+        EntitlementDTO ent = consumerApi.listEntitlements(system.getUuid()).get(0);
+        assertEquals(ent.getEndDate(), pool.getEndDate());
+    }
+
+    @Test
+    @OnlyInStandalone
+    @DisplayName("should delete single entitlement in excess when subscription quantity is decreased")
+    public void shouldDeleteSingleEntInExcessWhenSubQuantityIsDecreasedStandalone() throws Exception {
+        //  this entitlement makes the counts inconclusive
+        consumerApi.unbindAll(system.getUuid());
```
nikosmoum, line 214: This appears in a few tests (I know it's been ported as is from the ruby version), but it's a code smell that can lead to test flakiness. In such a case, where enough of the tests require a piece of data to be created in the beforeEach (in this case a consumer and it's entitlements), while others don't, then we should split one of the two groups in a nested class.

Let's create a nested class only for the tests like this one, that don't want `system` to be pre-entitlement by the `beforeEach`

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateSpecTest.java
```diff
@@ -0,0 +1,634 @@
...
+    private ConsumerClient consumerApi;
+    private HostedTestApi hostedTestApi;
+    private JobsClient jobApi;
+    private ProductDTO monitoring;
+    private PoolDTO pool;
+    private SubscriptionDTO subscription;
+    private ConsumerDTO system;
+
+    @BeforeAll
+    public void beforeAll() throws ApiException {
```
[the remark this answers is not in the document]
wottop: The class members didn't need to be static, so I made them non-static.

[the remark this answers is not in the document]
nikosmoum: It will have to be static once Ondrej's PR goes in because he's changing the `TestInstance.Lifecycle` (https://github.com/candlepin/candlepin/pull/3557/files#diff-a79a06da6cfad80f38cf42205b904d5f2ee7f37048f03bba288f0114c34da756). So depending on which PR goes in first, the other will have to adjust theirs

nikosmoum (approved):

spec-tests/src/test/java/org/candlepin/spec/EntitlementCertificateSpecTest.java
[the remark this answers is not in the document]
nikosmoum: I've merged Ondrej's PR, so this one needs changing because it's failing with `BeforeAll method 'public void org.candlepin.spec.EntitlementCertificateSpecTest.beforeAll() throws org.candlepin.invoker.client.ApiException' must be static`

PR #3535, opened 2022-08-25
[M] Improved spec test random string generation and job waiting routines (ENT-5354)

Ceiu:
- Updated how random string generation works, allowing for specific
  character sets to be used, removing the automatic hyphenation, and
  generating fixed-length strings rather than dynamic-lengthed ones
- Updated the API around waiting for jobs, permitting both the job ID
  and the job object itself; the latter of which will also verify
  the job itself exists and has a valid job ID
- waitForJob now throws an exception if a job vanishes during the
  wait loop, or if the job does not terminate within the alloted time
- Updated several tests and builders to use the new string generation
  method with appropriate charsets
- Updated several tests to also verify jobs they wait on terminate
  with a successful status

### DIFF
--- a/spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
+++ b/spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
@@ -21,21 +21,114 @@ public final class StringUtil {
 
     private static final Random RANDOM = new Random();
 
+    /** Default length of randomly generated strings */
+    public static final int DEFAULT_RANDOM_STRING_LENGTH = 8;
+
+    /** Character set consisting of upper- and lower-case alphabetical ASCII characters */
+    public static final String CHARSET_ALPHABETICAL = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+
+    /** Character set consisting of numeric ASCII characters */
+    public static final String CHARSET_NUMERIC = "0123456789";
+
+    /** Character set consisting of characters representing base-16 (hexidecimal) values (0-9, A-F) */
+    public static final String CHARSET_NUMERIC_HEX = "0123456789ABCDEF";
+
+    /** Character set consisting of numeric and upper- and lower-case alphabetical ASCII characters */
+    public static final String CHARSET_ALPHANUMERIC = CHARSET_ALPHABETICAL + CHARSET_NUMERIC;
+
     private StringUtil() {
         throw new UnsupportedOperationException();
     }
 
     /**
-     * Generates a random suffix of 7 characters or fewer.
+     * Generates a random string of characters from the specified charset string.
      *
-     * @param base
-     *  The base string to concatenate to the generated suffix
+     * @param length
+     *  the length of the string to generate; must be a positive integer
+     *
+     * @param charset
+     *  a string representing the character set to use for generating the string
+     *
+     * @throws IllegalArgumentException
+     *  if length is a non-positive integer, or charset is null or empty
      *
      * @return
-     *  the provided base string concatenated with a randomly generated suffix
+     *  a randomly generated string using the characters from the specified charset string
      */
-    public static String random(String base) {
-        return base + "-" + RANDOM.nextInt(100000);
+    public static String random(int length, String charset) {
+        if (length < 1) {
+            throw new IllegalArgumentException("length is a non-positive integer");
+        }
+
+        if (charset == null || charset.isEmpty()) {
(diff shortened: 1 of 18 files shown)


spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
```diff
@@ -21,21 +21,114 @@ public final class StringUtil {
...
+     *
+     * @return
+     *  a randomly generated string using the characters from the specified charset string appended
+     *  to the provided prefix
+     */
+    public static String random(String prefix, int length, String charset) {
+        String suffix = random(length, charset);
+
+        return (prefix != null && !prefix.isEmpty()) ?
+            prefix +  suffix :
```
JoshAlbrecht23, line 98: Extra space after '+'

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
```diff
@@ -21,21 +21,114 @@ public final class StringUtil {
...
+     *  a string representing the character set to use for generating the string
+     *
+     * @throws IllegalArgumentException
+     *  if length is a non-positive integer, or charset is null or empty
+     *
+     * @return
+     *  a randomly generated string using the characters from the specified charset string appended
+     *  to the provided prefix
+     */
+    public static String random(String prefix, int length, String charset) {
```
JoshAlbrecht23, line 94: Do we want to add the hyphen between the prefix and suffix automatically so that the caller doesn't always have to always add it to their prefix? I guess that would give the caller less control over creating a random string exactly how they want it, but it would mean our random strings would be uniform.

JoshAlbrecht23 (changes requested):
An extra space and a question regarding the hyphen between the prefix and suffix.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
```diff
@@ -21,21 +21,114 @@ public final class StringUtil {
...
+    /** Character set consisting of upper- and lower-case alphabetical ASCII characters */
+    public static final String CHARSET_ALPHABETICAL = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+
+    /** Character set consisting of numeric ASCII characters */
+    public static final String CHARSET_NUMERIC = "0123456789";
+
+    /** Character set consisting of characters representing base-16 (hexidecimal) values (0-9, A-F) */
+    public static final String CHARSET_NUMERIC_HEX = "0123456789ABCDEF";
+
+    /** Character set consisting of numeric and upper- and lower-case alphabetical ASCII characters */
```
JoshAlbrecht23, line 36: Extra '-' after upper

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
Ceiu: Fixed

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
Ceiu: I went back and forth on this myself, and landed on not including it, since it (a) mucks with the number of characters output by the string in a way that's not good for our random space problem and (b) doesn't make sense for every format of string, and it's probably clearer overall to have the caller specify.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
Ceiu: The hyphen connects to "case" following "lower-", and this is shorthand for the following: "upper-case and lower-case" or "upper[-]/lower-case"

JoshAlbrecht23: ooh that's my bad!

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/StringUtil.java
JoshAlbrecht23: Both points make a lot of sense. I'm good with the way you implemented it!

JoshAlbrecht23 (approved):

nikosmoum (approved):

PR #3538, opened 2022-08-29
Enable parallel execution of spec tests

Januson:
- Extract test logger settings into a convention file
- Enable parallel execution of spec tests

Note: We've written a fair amount of spec tests and questions are starting to arise whether the choices we are making now will affect tests running in parallel. We should start with parallel execution in order to flush out issues early.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,14 +17,13 @@ plugins {
     id "org.owasp.dependencycheck"
     id "jacoco"
     id "org.sonarqube"
-    id "com.adarshr.test-logger"
-
 
     // The following are plugins are plugins we wrote in the buildSrc folder
     id "org.candlepin.gradle.gettext"
     id "org.candlepin.gradle.msgfmt"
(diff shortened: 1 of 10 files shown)


Ceiu:
Looks like the output turns into just a soup of individual test one-liners? Parameterized tests get kinda messy if they have a name format specified, but look awful when not run in parallel.

IIRC we added a separate test runner or log output for that nicer output, but it's being obsoleted by this PR. That being the case, perhaps it's a good time to also remove that.

Januson:
Yeah. Test logging is not ideal. Problem is, there's probably nothing we can do about it. I think it might be worth keeping the plugin. Mostly because without it we would have to configure test logging manually again only to achieve the same mixed pile of output.

I wonder if it could hold on to output until it has output for the whole tests class. That probably depends on whether JUnit outputs enough events to determine when the tests class is finished.

PR #3543, opened 2022-08-31
Update liquibase to 4.15.0

Januson:


Ceiu (approved):
Seems fine, but why did we remove the positioning on the columns?

PR #3544, opened 2022-08-31
[M] Add packages to distinguish classpath for server and client side

bakajstep:
- Add server package to server side
- Add client package to client side


api/candlepin-api-config.json
```diff
@@ -1,8 +1,8 @@
 {
-  "modelPackage": "org.candlepin.dto.api.v1",
-  "apiPackage": "org.candlepin.resource",
-  "invokerPackage": "org.candlepin",
-  "groupId": "org.candlepin",
+  "modelPackage": "org.candlepin.server.dto.api.v1",
+  "apiPackage": "org.candlepin.server.resource",
```
Ceiu, line 3: Let's keep things consistent with the existing naming schemes and move the client/server further in:

```
"modelPackage": "org.candlepin.dto.api.[server|client].v1"
"apiPackage": "org.candlepin.resource.[server|client].v1"
```

I'm unsure what to make of the invokerPackage and groupId, as those don't seem to be impacting much else. I assume it's for the other generated bits like the top-level utility classes and exceptions? If so, I'm unsure where to put those, and without more control over the organization, we won't be able to match everything up anyway.

Any other thoughts on this? @nikosmoum @Januson @JoshAlbrecht23 ?

Ceiu (changes requested):

api/candlepin-api-config.json
nikosmoum: I agree with what you proposed on the modelPackage/apiPackage.
As for the other two I don't think it matters much: for invokerPackage we could set it to `org.candlepin.invoker.[server|client]` (It only affects the client anyway; we don't have anything autogenerated on the server side for this). And the groupId is irrelevant as far as I understand so we could just leave it at `org.candlepin`

bakajstep: so it should look like this:
```
"modelPackage": "org.candlepin.dto.api.[server|client].v1"
"apiPackage": "org.candlepin.resource.[server|client].v1"
"invokerPackage": org.candlepin.invoker.[server|client]
"groupId": "org.candlepin"
```
Do you agree with that?

Ceiu: That's good with me.

JoshAlbrecht23: That looks good to me as well!

bakajstep: done

nikosmoum (approved):

PR #3549, opened 2022-09-01
[M] Added minor state management to the CandlepinContextListener (ENT-5386)

Ceiu:
- Added state tracking to the CandlepinContextListener to avoid
  exceptions on shutdown if initialization doesn't complete
  successfully
- Reformatted string generation in cases where unrun Liquibase
  changes are detected

### DIFF
--- a/src/main/java/org/candlepin/guice/CandlepinContextListener.java
+++ b/src/main/java/org/candlepin/guice/CandlepinContextListener.java
@@ -89,6 +89,25 @@ import javax.servlet.ServletContextEvent;
 public class CandlepinContextListener extends GuiceResteasyBootstrapServletContextListener {
     public static final String CONFIGURATION_NAME = Configuration.class.getName();
 
+    /**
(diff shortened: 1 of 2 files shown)


Ceiu:
For reviewers: an easy way to test this would be to do a normal deployment with a fresh database (-g), open the database of choice and delete some entries from the databasechangelog table, then restart tomcat:

```
$ ./bin/deployment/deploy -g
$ psql -U candlepin candlepin -c "DELETE FROM databasechangelog WHERE filename = 'db/changelog/20211110083000-disallow_duplicate_versioned_entities.xml';"
$ sudo truncate --size=0 /var/log/candlepin/* && sudo rm /var/log/tomcat/* && sudo systemctl restart tomcat
```

From there, the output in candlepin.log should contain the now-missing changelog info, and tomcat/localhost-<date>.log should contain *only* a RuntimeException surrounding the Liquibase state error.

nikosmoum (approved):

PR #3571, opened 2022-09-09
Convert request_scanner.rb to Python

ptoscano:
Create a Python version of request_scanner.rb to perform the same task in a language that will still be used in the future. The implementation mostly follows the Ruby version for the logic, doing things in a Pythonic way:
- `argparse` instead of `OptionParser`
- a custom `argparse.Action` for getting regular expression objects instead of strings
- Python's `logging` instead of homebrew functions; `logging` works fine for our case with only few config tweaks
- handle errors directly on file opening, and item removal from dict, rather than checking beforehand
- pre-compile one more regex that was inline in a condition
- iterate directly on keys of dictionaries in case only the keys are needed
- format the code using the `'black'` [1] tool, so there is a standard codestyle matching the modern Python practices

The resulting output is almost the same; the only difference is due to the logging and applies to warning lines: instead of "WARN:" there is "WARNING:".

Side effect of the conversion: the Python version seems faster than the Ruby version.

[1] https://github.com/psf/black


Ceiu (changes requested):
This works pretty well; I forget what refactor work was needed on the logic, but we can make those later as they come up again.

The only change necessary here would be a typo correction in the commit header: the 'n' is missing from "convert"

Ceiu:
Nope, nothing you missed, just something I know I wanted to do at one point for one of the investigations, but since forgot. If it's important, it'll come up again and we can change it at that time.

PR #3573, opened 2022-09-12
[M] Removed direct output of CandlepinQuery from the resource endpoints (ENT-5400)

Ceiu:
- Updated all endpoints to not return a CandlepinQuery directly, instead processing it immediately
- Removed the ModelTranslator.translateQuery method
- Removed the CandlepinQueryInterceptor


Ceiu:
The automatic paging provided by the CPQ interceptor throws a major wrench into this change. Investigating alternate solutions here.

PR #3575, opened 2022-09-14
[M] Migrate pool_unlimited_master_spec to Java (ENT-5372)

wottop:


spec-tests/src/test/java/org/candlepin/spec/pools/PoolUnlimitedMasterSpecTest.java
```diff
@@ -0,0 +1,341 @@
...
+            productVirtInstanceMuliplier);
+        productVirtProductMuliplier = ownerProductApi.createProductByOwner(owner.getKey(),
+            productVirtProductMuliplier);
+
+        poolNoVirt = createUnlimitedPool(productNoVirt);
+        poolUnlimitedVirt = createUnlimitedPool(productUnlimitedVirt);
+        poolVirt = createUnlimitedPool(productVirt);
+        poolVirtHostDep = createUnlimitedPool(proudctVirtHostDep);
+        poolVirtInstanceMuliplier = createUnlimitedPool(productVirtInstanceMuliplier);
+        poolVirtProductMuliplier = createUnlimitedPool(productVirtProductMuliplier);
```
Januson, line 149: It looks like most of the setup is not actually needed for every test. Most of the tests require only 1-2 pools and don't care about the rest. It could be beneficial to split it up.

The only place that actually needs all of them is `shouldHaveCreatedCorrectPools`, but I think those assertions could be split and integrated into other tests as well.

spec-tests/src/test/java/org/candlepin/spec/pools/PoolUnlimitedMasterSpecTest.java
```diff
@@ -0,0 +1,341 @@
...
+        product = ownerProductApi.createProductByOwner(owner.getKey(), product);
+        // create pool with -100 quantity
+        PoolDTO unlimitedPool = Pools.random(product)
+            .quantity(-100L)
+            .subscriptionId(StringUtil.random("sub"))
+            .subscriptionSubKey(StringUtil.random("key"))
+            .upstreamPoolId(StringUtil.random("pool"));
+        unlimitedPool = ownerApi.createPool(owner.getKey(), unlimitedPool);
+        final String poolId = unlimitedPool.getId();
+        sleep(2000);
```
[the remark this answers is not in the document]
wottop: Without it the listOwnerPools happens before the pool creation is done. I get back 0 pools.

spec-tests/src/test/java/org/candlepin/spec/pools/PoolUnlimitedMasterSpecTest.java
```diff
@@ -0,0 +1,373 @@
...
+            .type(ConsumerTypes.System.value())
+            .facts(Map.of("virt.uuid", guestUuid, "virt.is_guest", "true"));
+        guest = consumerApi.createConsumer(guest);
+        guestClient = ApiClients.trustedConsumer(guest.getUuid());
+        guestUnmapped = Consumers.random(owner)
+            .name(StringUtil.random("host"))
+            .type(ConsumerTypes.System.value())
+            .facts(Map.of("virt.is_guest", "true"));
+        guestUnmapped = consumerApi.createConsumer(guestUnmapped);
+        guestUnmappedClient = ApiClients.trustedConsumer(guestUnmapped.getUuid());
```
nikosmoum, line 110: We shouldn't be using trusted consumers (which are using the cp-consumer header in the background) for all of these clients. The ruby test is simply passing in the identity cert+key. The equivalent in java would be:
`ApiClients.ssl(guestUnmapped)`.

Which makes me wonder if we have other java spec tests doing the same (I'll check)

spec-tests/src/test/java/org/candlepin/spec/pools/PoolUnlimitedMasterSpecTest.java
```diff
@@ -0,0 +1,373 @@
...
+
+    @BeforeEach
+    public void beforeEach() throws ApiException, InterruptedException {
+        owner = ownerApi.createOwner(Owners.random());
+        String guestUuid = StringUtil.random("uuid");
+
+        physicalSystem = Consumers.random(owner)
+            .name(StringUtil.random("host"))
+            .type(ConsumerTypes.System.value())
+            .guestIds(List.of(new GuestIdDTO().guestId(guestUuid)));
```
nikosmoum, line 96: Can we also remove the owner & consumers creation out of the before (maybe with a similar utility method as the one for products/pools? as per the guidelines we should only be creating data here if they are to be read-only, but the tests here are attaching entitlements to them

nikosmoum (changes requested):

nikosmoum (approved):

PR #3591, opened 2022-09-22
[M] Migrated refresh_pools_spec to Java (ENT-5379)

JoshAlbrecht23:
- removed refresh_pools_spec.rb
- split up the certificate extraction method in CertificateUtil
- made owner access mode declarations public


spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,877 @@
...
+        assertNotNull(updatedEnt);
+        assertEquals(1, updatedEnt.getCertificates().size());
+        CertificateDTO updatedCert = updatedEnt.getCertificates().stream()
+            .collect(MoreCollectors.onlyElement());
+        assertNotEquals(entSerial.asText(), updatedCert.getSerial().getSerial());
+    }
+
+    @Test
+    @OnlyInHosted
+    void shouldRegenerateEntitlementsWhenRequiredProductsChanges() throws Exception {
```
JoshAlbrecht23, line 583: This is a pretty large test and I wasn't sure if it could be broken up or not and so I kept it the same so I didn't lose any test coverage on functionality. However, if someone knows a good way of breaking this up, that would be great.

Original test:
https://github.com/candlepin/candlepin/blob/master/spec/refresh_pools_spec.rb#L642

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,877 @@
...
+            for (BrandingDTO actualBrand : actual) {
+                if (expectedBrand.getProductId().equals(actualBrand.getProductId())) {
+                    compareBrandings(expectedBrand, actualBrand);
+                    found = true;
+                }
+            }
+
+            assertTrue(found, "missing branding " + expectedBrand.getName());
+        }
+    }
```
JoshAlbrecht23, line 848: I wasn't sure if there is a better way of comparing the two unordered collections of brandings when we only want to assert against the product id, name, and type. External libraries like AssertJ or apache commons-collections all use the equals method on the BrandingDTO object and that includes the created/updated date time.

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,877 @@
...
+        ProductDTO prod3 = adminClient.hosted().createProduct(Products.random());
+        providerProduct.setProvidedProducts(Set.of(prod3));
+        providerProduct = adminClient.hosted().updateProduct(providerProduct.getId(), providerProduct);
+
+        refreshPools(adminClient, ownerKey);
+        pools = adminClient.pools().listPoolsByOwner(owner.getId());
+        assertEquals(1, pools.size());
+        PoolDTO pool = pools.get(0);
+        assertEquals(1, pool.getProvidedProducts().size());
+        assertEquals(prod3.getId(), pool.getProvidedProducts().stream()
```
Ceiu, line 158: This stream collector seems like a really complicated way of doing:

`<collection>.iterator().next()`

Given that we're pulling the provided products collection twice, it may even make sense to do that above and then use assertJ tests on it:

```
Collection<ProductDTO> providedProducts = pool.getProvidedProducts();
assertThat(providedProducts)
  .isNotNull()
  .hasSize(1)
  .singleElement()
  .returns(prod3.getId(), ProductDTO::getId);
```

Or you could do the whole block as one long chain, as mentioned in another comment somewhere in this review.

Ceiu (changes requested):
For the most part this is good. The only real blocking changes here are the missing modifier keywords and a couple of tests which aren't testing exactly as the test seems to intend.

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
JoshAlbrecht23: I like just clearing the created/updated fields since none of the tests care about those fields. If a future test needed to assert on the created and updated fields then they can use the assertThat to compare the elements and use the equals method on the object.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Owners.java
```diff
@@ -28,9 +28,9 @@
  */
 public final class Owners {
 
-    private static final String ACCESS_MODE_LIST = "org_environment,entitlement";
-    private static final String ENT_ACCESS_MODE = "entitlement";
-    private static final String SCA_ACCESS_MODE = "org_environment";
+    public static final String ACCESS_MODE_LIST = "org_environment,entitlement";
+    public static final String ENT_ACCESS_MODE = "entitlement";
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,877 @@
...
+        assertEquals(expectedSubscriptionCount, pools.size());
+        for (PoolDTO pool : pools) {
+            assertNotNull(pool.getSubscriptionId());
+            assertTrue(expectedSubIds.contains(pool.getSubscriptionId()));
+        }
+    }
+
+    @Test
+    @OnlyInHosted
+    void shouldDetectChangesInProviderProducts() {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,877 @@
...
+        providerProduct = adminClient.hosted().createProduct(providerProduct);
+        adminClient.hosted().createSubscription(Subscriptions.random(owner, providerProduct));
+
+        refreshPools(adminClient, ownerKey);
+        List<PoolDTO> pools = adminClient.pools().listPoolsByOwner(owner.getId());
+        assertEquals(1, pools.size());
+
+        PoolDTO actualPool = pools.get(0);
+        Set<String> expectedProvProdIds = Set.of(prod1.getId(), prod2.getId());
+        assertEquals(expectedProvProdIds.size(), actualPool.getProvidedProducts().size());
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,877 @@
...
+        }
+
+        // Remove the old provided products and add a new one...
+        ProductDTO prod3 = adminClient.hosted().createProduct(Products.random());
+        providerProduct.setProvidedProducts(Set.of(prod3));
+        providerProduct = adminClient.hosted().updateProduct(providerProduct.getId(), providerProduct);
+
+        refreshPools(adminClient, ownerKey);
+        pools = adminClient.pools().listPoolsByOwner(owner.getId());
+        assertEquals(1, pools.size());
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated, but I had to break it into two assertThat blocks as I was getting the following error for the returns check. I think that the extracting method is wrapping the set and not allowing you to check the field with the returns method.

`The method returns(T, Function<Set<ProvidedProductDTO>,T>) in the type AbstractObjectAssert<ObjectAssert<Set<ProvidedProductDTO>>,Set<ProvidedProductDTO>> is not applicable for the arguments (String, Function<ProvidedProductDTO,String>)`

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,877 @@
...
+        dataCenterProduct.setDerivedProduct(null);
+        adminClient.hosted().updateProduct(dataCenterProduct.getId(), dataCenterProduct, false);
+
+        refreshPools(adminClient, ownerKey);
+        pools = adminClient.pools()
+            .listPools(owner.getId(), null, dataCenterProduct.getId(), null, null, null, null, null, null);
+
+        assertThat(pools)
+            .hasSize(2)
+            .extracting(PoolDTO::getProductId)
```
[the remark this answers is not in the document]
JoshAlbrecht23: Yeah, updated!

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,877 @@
...
+            .contains(dataCenterProduct.getId());
```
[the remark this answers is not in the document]
JoshAlbrecht23: Yeah, this should be `.containsOnly` because both pool productIds should be the dataCenterproduct id

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,877 @@
...
+                    compareBrandings(expectedBrand, actualBrand);
+                    found = true;
+                }
+            }
+
+            assertTrue(found, "missing branding " + expectedBrand.getName());
+        }
+    }
+
+    private void verifyCertProductsAndContent(JsonNode certBody,
```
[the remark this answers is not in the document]
JoshAlbrecht23: I went back and forth on this when writing the method, and I like the idea of keeping it a map in case we end up adding tests that want to check content for multiple products.

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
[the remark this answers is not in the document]
Ceiu: Did you include `.singleElement` twice in that chain?

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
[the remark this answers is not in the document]
Ceiu: I think in practice, setting up the map will be more work than calling the method multiple times. In terms of setting this up to be used in other places/tests going forward, we'd get more value flipping the logic here and converting the JsonNode into a map of pids-to-cids, and then processing it in-test:

```
private static Map<String, List<String>> toProductContentIdMap(JsonNode certNode) {
  // TODO: add reasonable node assertions or checks
  Map<String, List<String>> output = new HashMap<>();

  JsonNode products = certNode.get("products");
  
  products.forEach(productNode -> {
    String productId = productNode.get("id").asText();

    // note that the json could very well list the same content multiple times, 
    // so we can't safely use a set here.
    List<String> contentIds = output.computeIfAbsent(productId, () -> new ArrayList<>()); 

    content = productNode.get("content");
    content.forEach(contentNode -> contentIds.add(contentNode.get("id").asText()));
  })

  return output;
}

...

Map<String, List<String>> certProductContentIds = toProductContentIdMap(certNode);
assertThat(certProductContentIds).containsOnly(
  entry("expected pid1", List.of("expected", "content", "ids")),
  entry("expected pid2", List.of("expected cid")),
  entry("expected pid3", List.of("expected cid 2")));

// Or, for the more verbose among us:
assertThat(certProductContentIds)
  .isNotNull()
  .hasSize(3)
  .containsOnlyKeys("expected", "product", "ids");

assertThat(certProductContentIds.get("expected"))
  .isNotNull()
  .hasSize(3)
  .containsOnly("expected", "content", "ids");
```

Not a big deal either way, just something to consider going forward. The more complex and context-specific the input is to a utility method, the less likely it is to be used as intended, or at all.

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,883 @@
...
+        assertEquals(expected.getName(), actual.getName());
+        assertEquals(expected.getType(), actual.getType());
+    }
+
+    private void compareBrandings(Collection<BrandingDTO> expected, Collection<BrandingDTO> actual) {
+        expected.stream().forEach(brand -> {
+            brand.setCreated(null);
+            brand.setUpdated(null);
+        });
+        actual.stream().forEach(brand -> {
```
Ceiu, line 846: `Collection` (or more specifically, `Iterable`) also has a `forEach` method. It's not necessary to build a stream first.

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
Ceiu: That's fine -- just make sure to document that it's happening so it doesn't take anyone by surprise.

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,883 @@
...
+
+import java.time.OffsetDateTime;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+@SpecTest
+public class RefreshPoolsTest {
```
Ceiu, line 66: Just caught this during manual testing: should probably name this RefreshPoolsSpecTest to not confuse it with the unit tests.

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,883 @@
...
+
+        // Verify the entitlement cert has changed as a result
+        EntitlementDTO updatedEnt = adminClient.entitlements().getEntitlement(ent.get("id").asText());
+        assertThat(updatedEnt.getCertificates())
+            .hasSize(1)
+            .map(CertificateDTO::getSerial)
+            .isNotNull()
+            .map(CertificateSerialDTO::getSerial)
+            .isNotNull()
+            .isNotEqualTo(serial.asLong());
```
Ceiu, line 764: This test chain isn't performing as you'd expect from the naming of the `map` function. Generally speaking, `extracting` and `map` off a IterableAssert map the whole collection using the mapping function provided, leaving you with another collection as the target under test.

The end result of this by the time you reach `.isNotEqualTo` is a single-element collection containing the serial. As such, the collection will never be equal to the target serial.

To do what you intend here, we need to use `.singleElement` to get out of the collection context, and then from there we can use `.extracting` to map to the desired type:

```
assertThat(updatedEnt.getCertificates())
    .singleElement()
    .isNotNull()
    .extracting(CertificateDTO::getSerial)
    .isNotNull()
    .extracting(CertificateSerialDTO::getSerial)
    .isNotEqualTo(serial.asLong());
```

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
```diff
@@ -0,0 +1,883 @@
...
+        refreshPools(adminClient, ownerKey);
+        pools = adminClient.pools().listPoolsByOwner(owner.getId());
+        assertThat(pools)
+            .isNotNull()
+            .hasSize(1);
+
+        assertThat(pools.get(0).getProvidedProducts())
+            .isNotNull()
+            .singleElement()
+            .returns(prod3.getId(), from(ProvidedProductDTO::getProductId));
```
Ceiu, line 162: Not a change request, but just making sure you know: the `from` here is purely optional syntactic sugar to make it "read" like an English sentence.

Ceiu (changes requested):

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
Ceiu: Also, while I'm thinking on it; since these are the generated DTOs, you could do it inline with method chaining:

```
expected.forEach(brand -> brand.created(null).updated(null));
actual.forEach(brand -> brand.created(null).updated(null));
```

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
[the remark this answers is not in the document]
JoshAlbrecht23: I like having a method that returns the map of product id to content ids and asserting against the map in the test method a lot more than the method I had! Updated

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
JoshAlbrecht23: Good catch! Updated the file name

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsTest.java
JoshAlbrecht23: Oh, I see what you are saying. Updated

Ceiu (approved):

PR #3594, opened 2022-09-26
[M] Added support for the device_auth capability and status (ENT-3599)

Ceiu:
- Added support for the device_auth capability
- Added new fields to the status DTO for indicating OAuth2 device auth details (realm, url, client_id, and scope)
- Updated keycloak flows to also populate the device auth fields and capabilities

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -9447,11 +9447,22 @@ components:
           uniqueItems: true
           items:
             type: string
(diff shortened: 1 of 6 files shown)


Ceiu:
For some context here: reworking Keycloak and all of its various levels of indirection and adapters is non-trivial compared to the MVP of just adding the new fields. If we still want to go down that road of making Keycloak more generic, it should be done as a separate PR after this one goes through the review process.

nikosmoum (approved):

PR #3595, opened 2022-09-26
[M] 2129177: Retain order of activation keys on db lookup (ENT-5423)

wottop:


src/main/java/org/candlepin/model/activationkeys/ActivationKeyCurator.java
```diff
@@ -95,7 +96,10 @@ public List<ActivationKey> findByKeyNames(String ownerKey, Collection<String> ke
                 .setParameter("keys", keyNameBlock)
                 .getResultList());
         }
-
+        List<String> sortOrder = new ArrayList<>();
+        sortOrder.addAll(keyNames);
+        Collections.sort(foundActivationKeys,
+            Comparator.comparing(item -> sortOrder.indexOf(item.getName())));
```
Ceiu, line 102: While this works, we should not be doing this globally at the curator level. The `ConsumerResource.findActivationKeys` method is a prime spot to do this, as that's the only time we'll care about ensuring the input and output retain the same ordering. It also saves us from having to create a new copy of the list since we'll know the exact type and ordering already.

As a bonus, we also move the testing of the ordering to the resource level where it belongs as it's become part of the registration spec now.

Ceiu (changes requested):

src/main/java/org/candlepin/model/activationkeys/ActivationKeyCurator.java
nikosmoum: To add to what @Ceiu said, can we rename and/or add a javadoc on the findActivationKeys method so that developers know that it finds activation keys while keeping the order? Maybe `findActivationKeysInOrder`?

wottop: Still needs a spec test. I pushed it prematurely.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1063,13 +1064,26 @@ public Consumer createConsumerFromDTO(ConsumerDTO consumer, ConsumerType type, P
...
+    /**
+     * We need to make sure the requested keys are available. In addition, the order in
+     *  which they are specified in the call should be maintained for their later application.
+     *
+     * @param principal
+     * @param owner
+     * @param keyStrings
+     * @return
+     */
+    List<ActivationKey> findActivationKeysInOrder(
```
Ceiu, line 1076: Missing scope modifier

src/test/java/org/candlepin/resource/ConsumerResourceTest.java
```diff
@@ -1322,6 +1325,29 @@ public void testFindConsumersByFact() {
         // to bother checking that we got the exact ones we're expecting.
     }
 
+    @Test
+    void keysFoundInExpectedOrder() {
```
Ceiu, line 1329: scope modifier

nikosmoum:
Looks like there's some activation key registration unit tests failing

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceActivationKeySpecTest.java
```diff
@@ -0,0 +1,100 @@
...
+    static OwnerDTO owner;
+    static ActivationKeyApi activationKeyApi;
+
+
+    @BeforeAll
+    static void beforeAll() {
+        client = ApiClients.admin();
+        ownerClient = client.owners();
+        consumerClient = client.consumers();
+        owner = Owners.random();
```
nikosmoum, line 56: Let's not make the owner globally available as per the guidelines. There's only 1 test for now, but the rest of them will be ported at some point

nikosmoum (approved):
Looks good to me. I'll leave @Ceiu re-review & merge

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1063,15 +1064,30 @@ public Consumer createConsumerFromDTO(ConsumerDTO consumer, ConsumerType type, P
...
+     * We need to make sure the requested keys are available. In addition, the order in
+     *  which they are specified in the call should be maintained for their later application.
+     *
+     * @param principal
+     * @param owner
+     * @param keyStrings
+     * @return
+     */
+    private List<ActivationKey> findActivationKeysInOrder(
         Principal principal, Owner owner, Set<String> keyStrings) {
```
Ceiu, line 1077: Because we're still accepting a set here, we've already lost the requested order. Given that we also do the parsing in this class, we probably should just pass in the original string here rather than a pre-parsed collection. That avoids having to re-wrap the items, and might clean up the comparator a bit.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1063,15 +1064,30 @@ public Consumer createConsumerFromDTO(ConsumerDTO consumer, ConsumerType type, P
...
         if (keyStrings.isEmpty()) {
             return Collections.emptyList();
         }
 
         List<ActivationKey> foundKeys = this.activationKeyCurator.findByKeyNames(owner.getKey(), keyStrings);
-
-        Set<String> found = foundKeys.stream()
+        List<String> sortOrder = new ArrayList<>();
```
Ceiu, line 1083: This block of collection juggling is unnecessary. The entirety of it can be replaced by a single call to `List.sort`:

```
List<String> keyNames = List.of(keyString.split("\s*,\*"));
List<ActivationKey> keys = this.activationKeyCurator.findByKeyNames(...);

// insert validation of resolved keys here

keys.sort(Comparator.comparing(key -> keyNames.indexOf(key.getName()));

return keys;
```

Note that we also need to properly handle the case where a given activation key is specified multiple times, and our principal check and general warning output is incorrect as well. We should always be warning when an AK is specified that doesn't exist (it's not user facing anyway), and if the intent is that non-AKPrincipals can't use AKs, then we shouldn't do *any* processing of fetched AKs in those cases.

Ceiu (changes requested):

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1063,15 +1064,30 @@ public Consumer createConsumerFromDTO(ConsumerDTO consumer, ConsumerType type, P
         }
     }
 
-    private List<ActivationKey> findActivationKeys(
+    /**
+     * We need to make sure the requested keys are available. In addition, the order in
+     *  which they are specified in the call should be maintained for their later application.
+     *
+     * @param principal
```
Ceiu, line 1071: We should also fill these in

src/main/java/org/candlepin/resource/ConsumerResource.java
wottop: The intent appears to be that we only log if it is a non-AK principal. I just dropped the condition.

Ceiu (changes requested):

wottop:
I used the linked hash map because it eliminates an explicit sort of the returned AK objects.

Ceiu (approved):

PR #3603, opened 2022-09-29
Migrate import_spec to Java

Januson:
While working on this spec I hit a couple trade-offs. There are other options to solve these but with drawbacks of their own.
- Generating exports is slow. I chose to generate them once per class in beforeAll
- Tests have to be run twice. Once for sync and once for async import endpoint
- JUnit5 does not provide any class level parameterization. To solve sync/async suite I chose inheritance where all tests are defined in sync spec and async spec inherits them while only redefining the import method.
- To enable inheritance while still have the ability to generate exports per-class I had to change the test life-cycle to PER_CLASS


spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/Importer.java
```diff
@@ -0,0 +1,93 @@
...
+import org.candlepin.invoker.client.ApiException;
+import org.candlepin.spec.bootstrap.client.ApiClient;
+import org.candlepin.spec.bootstrap.data.builder.Export;
+import org.candlepin.spec.bootstrap.data.builder.ExportGenerator;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+
+public final class Importer {
```
JoshAlbrecht23, line 29: Do we want to call this ImporterUtil?

JoshAlbrecht23 (commented):
If we are wanting to go with the ExportGenerator, I wonder if we should create the export generator with a Builder object that dictates what objects get created for the export. This would allow tests to generate an export with exactly what they need and you wouldn't need to have the different setups (full/minimal). Then allow the created object to be accessed from the ExportGenerator and asserted against like you currently have.

I'm curious what @Ceiu thinks because we were having similar conversations with the Export tests.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ExportGenerator.java
```diff
@@ -0,0 +1,302 @@
...
+import java.time.OffsetDateTime;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class ExportGenerator implements AutoCloseable {
```
nikosmoum, line 48: During the exporter spec migration (https://github.com/candlepin/candlepin/pull/3509) after discussion it was decided to not create an 'exporter' abstraction/infrastructure, because it is unnecessarily creating data (products, pools, etc.) all at once, while each test method only needs a minimal amount. And that the export process was primarily slow just because of that creating of a massive amount of data. Instead, each test should create the test data it needed, and then do an export with just that data

spec-tests/src/test/java/org/candlepin/spec/imports/ImportErrorSpecTest.java
```diff
@@ -0,0 +1,205 @@
...
+    private Export exportOld;
+    private Export export;
+    private Importer importer;
+
+    @BeforeAll
+    public void beforeAll() {
+        admin = ApiClients.admin();
+        owner = admin.owners().createOwner(Owners.random());
+        UserDTO user = UserUtil.createUser(admin, owner);
+        userClient = ApiClients.trustedUser(user.getUsername());
```
nikosmoum, line 76: Can we replace all `ApiClients.trustedUser(username)` with `ApiClients.basic(username, password)` through out this PR (and for all other spec tests from now on)? The ApiClients.trustedUser and ApiClients.trustedConsumer should never be used in spec tests (aside from the few tests that explicitly test those authentication methods, e.g. TrustedConsumerAuthSpecTest & TrustedUserAuthSpecTest).

Januson: Done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ExportGenerator.java
Januson: Done. Tests got significantly slower, but the change got rid of lifetime changes so it might be a worthy tradeoff.

JoshAlbrecht23:
> @JoshAlbrecht23 Depends if we have a use-cases for that. We could definitely make export content configurable. The question is whether it would be worth it to introduce this additional complexity when we don't have that many tests that actually care about the shape of the data. It would be worth it if we wanted to test with a lot of different data setups.

That's a good point! There really aren't that many existing tests that would benefit from that functionality and so it probably isn't worth adding all that complexity.

JoshAlbrecht23 (approved):
It looks good to me, but it may be a good idea to get another +1 from someone else

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ExportGenerator.java
Januson: I am keeping it as a separate commit for now so that they can be compared.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/Importer.java
```diff
@@ -0,0 +1,80 @@
...
+import org.candlepin.spec.bootstrap.client.ApiClient;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+
+public final class Importer {
+
+    private final ApiClient client;
+    private final boolean async;
```
nikosmoum, line 30: Apologies for finding this late, but we don't need conditionals for running the test sync or async, or any of the abstractions around this, because we plan to remove the sync endpoint in the next sprint or so (it is deprecated and no longer used), so all importing will be async from now on

spec-tests/src/test/java/org/candlepin/spec/imports/ImportSuccessSpecTest.java
```diff
@@ -0,0 +1,357 @@
...
+    private ConsumerDTO consumer;
+    private Importer importer;
+
+    @BeforeEach
+    public void beforeAll() {
+        admin = ApiClients.admin();
+        owner = admin.owners().createOwner(Owners.random());
+        UserDTO user = UserUtil.createUser(admin, owner);
+        userClient = ApiClients.basic(user);
+        consumer = userClient.consumers().createConsumer(Consumers.random(owner));
```
nikosmoum, line 80: Should this be in the before method? I see this consumer is only used in 1 test method

spec-tests/src/test/java/org/candlepin/spec/imports/ImportErrorSpecTest.java
```diff
@@ -0,0 +1,208 @@
...
+    private ApiClient userClient;
+    private Importer importer;
+
+    @BeforeEach
+    public void beforeAll() {
+        admin = ApiClients.admin();
+        owner = admin.owners().createOwner(Owners.random());
+        UserDTO user = UserUtil.createUser(admin, owner);
+        userClient = ApiClients.basic(user);
+        userClient.consumers().createConsumer(Consumers.random(owner));
```
nikosmoum, line 61: This is creating a consumer and not using it anywhere?

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ExportCdn.java
```diff
@@ -0,0 +1,41 @@
...
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.spec.bootstrap.data.builder;
+
+public class ExportCdn {
```
nikosmoum, line 18: Could the autogenerated CdnDTO not be used instead of this?

nikosmoum (changes requested):
A few minor issues and some questions, but this is looking good

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClients.java
```diff
@@ -40,6 +41,10 @@ public static ApiClient basic(String username, String password) {
         return new ApiClient(CLIENT_FACTORY.createClient(username, password));
     }
 
+    public static ApiClient basic(UserDTO user) {
```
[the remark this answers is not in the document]
Januson: Done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ExportGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+        consumerApi = client.consumers();
+        rolesApi = client.roles();
+        usersApi = client.users();
+    }
+
+    /**
+     * Destroys created scaffolding by deleting its owner.
+     */
+    @Override
+    public void close() {
```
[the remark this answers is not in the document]
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/imports/ImportErrorSpecTest.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    @Test
+    public void shouldReturnCorrectErrorStatusMessageOnADuplicateImport() {
+        Export export = createMinimalExport();
+        this.importer.doImport(owner.getKey(), export.file());
+        assertConflict(() -> this.importer.doImport(owner.getKey(), export.file()))
+            .hasMessageContaining("MANIFEST_SAME");
+
+        List<ImportRecordDTO> pools = userClient.owners().getImports(owner.getKey());
```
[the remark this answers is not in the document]
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/imports/ImportSuccessSpecTest.java
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/imports/ImportErrorSpecTest.java
Januson: Done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ExportCdn.java
Januson: It could but it would have to carry webUrl in one of the unrelated properties such as name which I found confusing.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/Importer.java
Januson: Done.

nikosmoum (approved):

PR #3605, opened 2022-09-29
Make junit report output visible from jenkins

barnabycourt:


barnabycourt:
Closing because we don't want to eliminate the isolation that running in the container provides.

PR #3606, opened 2022-09-30
[M] 2130671: Candlepin ignores the 'include' parameter

Januson:
- Removed adding filtering mixin to individual DTOs. By adding the mixin directly to the Object all DTOs will inherit it.
- Removed legacy filters marked for removal
- Ported filtering spec test for easier verification


spec-tests/src/test/java/org/candlepin/spec/ResponseFilteringSpecTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    private static ApiClient admin;
+    private static OwnerDTO owner;
+    private static ConsumerDTO consumer;
+
+    @BeforeAll
+    static void beforeAll() {
+        admin = ApiClients.admin();
+        owner = admin.owners().createOwner(Owners.random());
+        UserUtil.createUser(admin, owner);
```
nikosmoum, line 52: This is creating a user but isn't using it? was it accidentally leftover?

spec-tests/src/test/java/org/candlepin/spec/ResponseFilteringSpecTest.java
```diff
@@ -0,0 +1,174 @@
...
+            .addQueryParam("exclude", "href")
+            .setPath("/consumers/" + consumer.getUuid())
+            .execute()
+            .deserialize(ConsumerDTO.class);
+
+        assertThat(consumerFull)
+            .usingRecursiveComparison()
+            .ignoringFields("href")
+            .isEqualTo(consumerFiltered);
+        assertThat(consumerFiltered.getHref()).isNull();
```
nikosmoum, line 69: Instead of null, can we check that the field is actually missing from the response?
I know that the href should never be null for a consumer, but in case that happened through a bug, this test would still pass for the wrong reason.

Same issue exists for all the other tests that assert on excluded fields

spec-tests/src/test/java/org/candlepin/spec/ResponseFilteringSpecTest.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    @Test
+    void shouldAllowFiltersOnEncapsulatedLists() {
+        admin.consumers().createConsumer(Consumers.random(owner));
+        admin.consumers().createConsumer(Consumers.random(owner));
+        JsonNode consumers = Request.from(admin)
+            .addQueryParam("type", "system")
+            .addQueryParam("include", "id")
+            .addQueryParam("include", "owner.id")
```
nikosmoum, line 140: This test is using the general GET /consumers API, which, in the serial ruby tests that have cleanup is acceptable, however in the java version we have no cleanup, so if we were unlucky and this test run around the end of the build, this could be fetching the entirety of all created system consumers. Can we filter the query more by owner key?

spec-tests/src/test/java/org/candlepin/spec/ResponseFilteringSpecTest.java
```diff
@@ -0,0 +1,174 @@
...
+            .setPath("/consumers")
+            .execute()
+            .deserialize(JsonNode.class);
+
+        assertThat(consumers)
+            .isNotEmpty()
+            .allSatisfy(consumer -> {
+                assertThat(consumer.get("id")).isNotNull();
```
nikosmoum, line 148: This should also be asserting that each consumer object only has 2 fields

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/ResponseFilteringSpecTest.java
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/ResponseFilteringSpecTest.java
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/ResponseFilteringSpecTest.java
Januson: Done.

nikosmoum (approved):

Ceiu:
Can we add the missing scope modifiers? It's hard to tell between developer intent and templated code with the unfortunate choice Sun made with the syntax for "package private," and it doesn't really buy is anything anyway.

PR #3608, opened 2022-09-30
[M] 2128583: Updated GuestIdDeserializer to use existing ObjectMapper (ENT-5422)

JoshAlbrecht23:
- Updated GuestIdDeserializer to use existing ObjectMapper

This is a fix for the original PR (#3606) for the issue

### DIFF
--- a/src/main/java/org/candlepin/jackson/GuestIdDeserializer.java
+++ b/src/main/java/org/candlepin/jackson/GuestIdDeserializer.java
@@ -16,14 +16,13 @@ package org.candlepin.jackson;
 
 import org.candlepin.dto.api.server.v1.GuestIdDTO;
 import org.candlepin.exceptions.CandlepinJsonProcessingException;
-import org.candlepin.util.ObjectMapperFactory;
 
 import com.fasterxml.jackson.core.JacksonException;
 import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.core.TreeNode;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
-import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectReader;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/resteasy/JsonProvider.java
```diff
@@ -147,14 +147,16 @@ public JsonProvider(boolean indentJson) {
...
         customModule.addSerializer(Date.class, new DateSerializer());
+
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        configureHateoasObjectMapper(mapper, indentJson);
+
         // Ensure we handle releaseVer fields properly
         customModule.addDeserializer(ReleaseVerDTO.class, new ReleaseVersionWrapDeserializer());
         customModule.addDeserializer(ConsumerTypeDTO.class, new ConsumerTypeDeserializer());
-        customModule.addDeserializer(GuestIdDTO.class, new GuestIdDeserializer());
+        customModule.addDeserializer(GuestIdDTO.class, new GuestIdDeserializer(mapper.reader()));
```
JoshAlbrecht23, line 157: It appears that we cannot use `ObjectMapper.readerFor()` because the ObjectMapper has a TypeFactory that caches the class type and this is referenced in `ObjectMapper.registerModule(customModule)` and breaks the custom deserializers.

src/main/java/org/candlepin/jackson/GuestIdDeserializer.java
```diff
@@ -36,36 +35,41 @@
  */
 public class GuestIdDeserializer extends JsonDeserializer<GuestIdDTO> {
     private static Logger log = LoggerFactory.getLogger(GuestIdDeserializer.class);
-    private static ObjectMapper mapper = ObjectMapperFactory.getObjectMapper();
+
+    private ObjectReader reader;
+
+    /**
+     * Creates a new GuestIdDeserializer using the specified reader as the
+     * default/base reader for processing guest IDs received in object form.
```
Ceiu, line 43: Missing doc for the object reader param

src/main/java/org/candlepin/jackson/GuestIdDeserializer.java
```diff
@@ -36,36 +35,41 @@
  */
 public class GuestIdDeserializer extends JsonDeserializer<GuestIdDTO> {
     private static Logger log = LoggerFactory.getLogger(GuestIdDeserializer.class);
```
Ceiu, line 37: We should also make this a constant

src/main/java/org/candlepin/jackson/GuestIdDeserializer.java
```diff
@@ -36,36 +35,41 @@
...
         }
-        else if (rootNode.isObject()) {
-            log.debug("Processing GuestIdDTO as a containing object node.");
-            return mapper.convertValue(rootNode, GuestIdDTO.class);
+        else if (node.isObject()) {
+            log.debug("Processing node as a GuestIdDTO instance: {}", node);
+
+            return reader.readValue(node, GuestIdDTO.class);
         }
         else {
```
Ceiu, line 68: We can actually drop the "else" here and just throw the exception directly from this point. Saves us a bit of indentation.

Ceiu (changes requested):
Three minor changes would be nice, but won't affect functionality. Also we should have a third person review this since the PR contains code both of us have written.

src/main/java/org/candlepin/jackson/GuestIdDeserializer.java
JoshAlbrecht23: Updated

src/main/java/org/candlepin/jackson/GuestIdDeserializer.java
JoshAlbrecht23: Updated

src/main/java/org/candlepin/jackson/GuestIdDeserializer.java
JoshAlbrecht23: Updated

Ceiu (approved):

nikosmoum (approved):

PR #3609, opened 2022-09-30
[M] Migrated instance_spec and multiplier_spec to Java (ENT-5381)

JoshAlbrecht23:
- removed instance_spec.rb
- removed multiplier_spec.rb


spec-tests/src/test/java/org/candlepin/spec/InstanceSpecTest.java
```diff
@@ -0,0 +1,203 @@
...
+            .extracting("type")
+            .containsExactlyInAnyOrder("NORMAL", "UNMAPPED_GUEST");
+
+        assertThat(guestClient.pools().listPoolsByConsumer(guestUser.getUuid()))
+            .isNotNull()
+            .hasSize(1);
+
+        physicalClient.consumers().bindProduct(physicalUser.getUuid(), engProd.getId());
+
+        int expectedQuantity = ((int) Math.ceil((double) cpuSockets / socketCount) * socketMultiplier);
```
JoshAlbrecht23, line 106: The original test asserts against a hard coded quantity value of 8 which made understanding the test difficult, so I wanted to calculate the expected quantity for clarity.

I believe this is correct based on some of the logic I saw in the rules.js quantity calculation methods and some of the testing I did , but maybe @wottop can help verify?

Januson (approved):

PR #3611, opened 2022-10-04
[4.1] 2131997: Added deserializer for GuestIdDTO (ENT-5422)

JoshAlbrecht23:
- deserializer that creates GuestIdDTO from id string
- deserializer that creates GuestIdDTO from containing object
- updated GuestIdDeserializer to use existing ObjectMapper


nikosmoum (changes requested):
The commits should have the 4.1 BZ id: 2131997

nikosmoum (approved):

PR #3613, opened 2022-10-05
[M] Migrated refresh_pools_spec to Java - part 2 (ENT-5380)

JoshAlbrecht23:
- removed refresh_pools_spec.rb


spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsSpecTest.java
```diff
@@ -847,6 +844,892 @@ public void shouldInvalidateEntitlementsWhenPoolQuantityIsReduced() {
...
+
+        refreshPools(adminClient, owner.getKey());
+        pools = adminClient.pools().listPoolsByOwner(owner.getId());
+        assertThat(pools).hasSize(1);
+
+        assertThat(adminClient.entitlements().getEntitlement(ents.get(0).get("id").asText()))
+            .isNotNull()
+            .extracting(EntitlementDTO::getCertificates)
+            .returns(1, Set::size)
+            .extracting(certs -> certs.iterator().next())
```
Ceiu, line 1493: A more syntactically clear way of doing this would be using the assertion factory and `singleElement`:

```
assertThat(adminClient.entitlements().getEntitlement(ents.get(0).get("id").asText()))
    .isNotNull()
    .extracting(EntitlementDTO::getCertificates, as(collection(CertificateDTO.class)))
    .singleElement()
    .isNotNull()
    .extracting(CertificateDTO::getSerial)
    .isNotNull()
    .isNotEqualTo(serialBeforeUpdate.longValue());
```

We get the size check and the extraction in one line rather than two.

You'll need to add the following imports for the above code:
```
import static org.assertj.core.api.Assertions.as;
import static org.assertj.core.api.InstanceOfAssertFactories.collection;
```

This applies to all of these assertion chains throughout the suite.

Ceiu (approved):
Tests look good, and if you're up for it, there's a different way to do some of the cert assertion chains that may read a little nicer for extracting the single element we expect.

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsSpecTest.java
JoshAlbrecht23: Perfect!! Thanks for this. I was really confused on why I could not use `.singleElement()` after extracting the set of CertificateDTOs and I ended up with that hacky approach. I'm still getting use to the assertj library, but I do like it a lot!

Ceiu: :+1:

Ceiu (approved):

PR #3615, opened 2022-10-06
Migrate hypervisor_check_in to Java part 2

wottop:
Includes cleanup to part 1


spec-tests/src/test/java/org/candlepin/spec/HypervisorCheckInSpecTest.java
```diff
@@ -763,14 +771,793 @@ public void shouldAllowASingleGuestToBeMigratedAndRevokeHostLimitedEntsHosted()
...
+            guestRH00271Entitlement.getPool().getProductId().equals(x.getPool().getProductId()))
+            .collect(Collectors.toList()).get(0);
+        assertNotNull(updatedGuestRH00271Entitlement);
+        EntitlementDTO updatedGuestRH00051Entitlement = newEnts.stream()
+            .filter(x -> guestRH00051Entitlement.getPool().getProductId().equals(x.getPool().getProductId()))
+            .collect(Collectors.toList()).get(0);
+        assertNotNull(updatedGuestRH00051Entitlement);
+
+        rh00051Cert = null;
+        rh00271Cert = null;
```
JoshAlbrecht23, line 1407: You shouldn't need to assign to null since the are already being reassigned to the output of the stream filtering in the next couple of lines

JoshAlbrecht23 (changes requested):

spec-tests/src/test/java/org/candlepin/spec/HypervisorCheckInSpecTest.java
```diff
@@ -146,45 +111,42 @@ private void initConsumersAndPools() {
...
-        guestConsumer2 = Consumers.random(owner)
-            .name(StringUtil.random("guest"))
-            .type(ConsumerTypes.System.value())
-            .facts(Map.of("virt.uuid", guest2VirtUuid, "virt.is_guest",
-                "true", "uname.machine", "x86_64"));
-        guestConsumer2 = consumerApi.createConsumer(guestConsumer2);
+    private void setupOwnerUserClient() {
+        owner = ownerApi.createOwner(Owners.random());
+        UserDTO user = UserUtil.createUser(client, owner);
+        userClient = ApiClients.trustedUser(user.getUsername());
```
nikosmoum, line 122: Can we replace all `ApiClients.trustedUser(username)` with `ApiClients.basic(username, password)` through out this PR (and for all other spec tests from now on)? The ApiClients.trustedUser and ApiClients.trustedConsumer should never be used in spec tests (aside from the few tests that explicitly test those authentication methods, e.g. TrustedConsumerAuthSpecTest & TrustedUserAuthSpecTest).

JoshAlbrecht23 (approved):

PR #3616, opened 2022-10-06
[M] Refresh Mapping Resolution (ENT-5384)

Ceiu:
Added logic for detecting and fixing entity mapping issues during refresh

- During refresh and manifest import, certain issues with org-entity
  mapping issues will be detected and, if present, will remap products
  and content to ensure correct entity usage

Reverted product/content link from activation key and environments

- ActivationKey no longer references products directly, instead
  referencing products by Red Hat product ID
- Environment no longer references content directly, instead
  referencing content by Red Hat content ID
- Removed the ActivationKey and Environment entity remapping logic
  from org refresh and product/content update flows
- Internal entity model API improvements, cleanup, and changes
- Removed some long-unused code

Removed orphaned product grace period functionality

- Removed the orphaned product grace period functionality entirely,
  as its purpose is superceded by removing the hard link between
  activation key and environment to product and content

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -8805,8 +8805,8 @@ components:
     EnvironmentContentDTO:
       description: EnvironmentContent represents the promotion of content into a particular environment.
       properties:
-        content:
-          $ref: "#/components/schemas/ContentDTO"
+        contentId:
+          type: string
         enabled:
           type: boolean
 
--- a/src/main/java/org/candlepin/controller/refresher/RefreshWorker.java
+++ b/src/main/java/org/candlepin/controller/refresher/RefreshWorker.java
@@ -16,6 +16,7 @@ package org.candlepin.controller.refresher;
 
 import org.candlepin.controller.ContentManager;
 import org.candlepin.controller.ProductManager;
+import org.candlepin.controller.refresher.RefreshResult.EntityState;
 import org.candlepin.controller.refresher.builders.ContentNodeBuilder;
 import org.candlepin.controller.refresher.builders.NodeFactory;
 import org.candlepin.controller.refresher.builders.PoolNodeBuilder;
@@ -29,12 +30,16 @@ import org.candlepin.controller.refresher.visitors.NodeProcessor;
 import org.candlepin.controller.refresher.visitors.PoolNodeVisitor;
 import org.candlepin.controller.refresher.visitors.ProductNodeVisitor;
 import org.candlepin.controller.util.EntityVersioningRetryWrapper;
+import org.candlepin.model.Content;
 import org.candlepin.model.ContentCurator;
 import org.candlepin.model.Owner;
 import org.candlepin.model.OwnerContentCurator;
 import org.candlepin.model.OwnerProductCurator;
+import org.candlepin.model.Pool;
 import org.candlepin.model.Pool.PoolType;
 import org.candlepin.model.PoolCurator;
+import org.candlepin.model.Product;
+import org.candlepin.model.ProductContent;
 import org.candlepin.model.ProductCurator;
 import org.candlepin.service.model.ContentInfo;
 import org.candlepin.service.model.ProductContentInfo;
@@ -49,8 +54,11 @@ import org.slf4j.LoggerFactory;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.persistence.EntityTransaction;
 import javax.persistence.LockModeType;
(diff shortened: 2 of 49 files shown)


src/main/java/org/candlepin/controller/refresher/RefreshWorker.java
```diff
@@ -508,18 +578,43 @@ public RefreshResult execute(Owner owner) {
             this.contentMapper.clearExistingEntities();
 
             // Add in our existing entities
-            this.poolMapper.addExistingEntities(
-                this.poolCurator.listByOwnerAndTypes(owner.getId(), PoolType.NORMAL, PoolType.DEVELOPMENT));
+            List<Pool> pools = this.poolCurator
```
JoshAlbrecht23, line 581: Do we need obtain a system lock for the poolCurator?

src/main/java/org/candlepin/model/Environment.java
```diff
@@ -114,32 +115,62 @@ public String getOwnerId() {
...
+        return this;
+    }
+
+    public Environment addEnvironmentContent(EnvironmentContent envcontent) {
+        if (envcontent == null) {
+            throw new IllegalArgumentException("envcontent is null");
+        }
+
+        // Ensure it's set to this environment. Really this should be immutable and created internally.
+        envcontent.setEnvironment(this);
```
JoshAlbrecht23, line 138: Do we need to include this to the `setEnvironmentContent(Set<EnvironmentContent> environmentContent)` method for the elements in that collection?

JoshAlbrecht23 (commented):
Overall the changes make sense! During the actual implementation, it would be interesting to get some performance metrics of the RefreshWorker against an SQL dump of non-prod.

src/main/java/org/candlepin/model/Environment.java
Ceiu: Yeah, we probably should. That whole method should probably be written to encapsulate the collection rather than just using whatever external collection it's given as well.

src/main/java/org/candlepin/controller/refresher/RefreshWorker.java
Ceiu: Probably not. At least, not at present. Pools are currently gated behind the owner locks elsewhere in refresh and the various pool manipulating operations. If pool scope changes, or we give our locking model a proper pass/refactor, that'd definitely be something to consider.

nikosmoum (commented):
The first change (entity mapping resolution) seems good, but I am unsure if the removal of hard links between environment/content will make the revert of the orphan grace period possible

src/main/java/org/candlepin/controller/refresher/mappers/AbstractEntityMapper.java
```diff
@@ -149,6 +163,12 @@ public boolean addImportedEntity(String id, I entity) {
             throw new IllegalArgumentException("entity is null");
         }
 
+        I existing = this.importedEntities.get(id);
+        if (existing != null && !existing.equals(entity)) {
+            log.warn("Remapping imported entity with a different entity version; discarding previous..." +
```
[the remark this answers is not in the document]
Ceiu: Imported (upstream) entities aren't ever "dirty" in a way we care about, as we have no control over how they're mapped or delivered to us and we have nothing to clean up there as a result.

With existing (local) entities, having two references to the same RH ID indicates our scope is either multi-org (bad) or a single org is referencing two different versions of a given entity, and we need to manually fix that mapping.

api/candlepin-api-spec.yaml
```diff
@@ -8888,8 +8888,8 @@ components:
     EnvironmentContentDTO:
       description: EnvironmentContent represents the promotion of content into a particular environment.
       properties:
-        content:
-          $ref: "#/components/schemas/ContentDTO"
+        contentId:
+          type: string
```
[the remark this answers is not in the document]
Ceiu: Hrmm... investigating the origins of the environment/key cleanup code, it goes back a very long time -- roughly 10 years [1].

So apparently the way refresh worked in the past was what was saving them -- we would unlink products and content from pools as they were cleaned up and deleted, but the products or content themselves would stick around indefinitely. But if the content was ever manually deleted, it would be removed from the environment all the same.

I *really* hate the orphan grace period code and flow, but it seems this change would also introduce weird behavior relative to past versions. Going to think on this a bit before fully committing to that cleanup removal; maybe there's a better option. The UUID -> ID change is probably still a net good one, but we'd still need to do some cleanup on the deletion of products and content regardless.

[1] https://github.com/candlepin/candlepin/commit/678f1f0b2f91d1241f693aa3d45b71c391b76ae5

Ceiu:
Closing this as the production version of it will be handled in other PRs.

PR #3620, opened 2022-10-10
[M] Fix bug with KeyValueParamDTO

bakajstep:
- Switch back parsing KeyValueParam before openApi generation


Ceiu:
Before I give this a full review, can I get an example of the issue that lead to the need to make these changes? I would like to understand the root issue so the intent and direction of this PR is clearer to me; as well as give me a chance to look at other options which may give us a better solution to the problem.

Ceiu:
One of the major issues we have here is that we're basically trying to straddle the line between manual parsing and keeping the old abstractions in place. From my understanding of the problem after talking with others who've hit it and recreating a reproducer for it, we're moving in the direction of not providing an abstraction at all. This being the case, all of our code processing those key:value pair strings will be done manually or semi-manually going forward. Either by way of directly processing the string as-is, or by invoking whatever utility or abstraction we have in place to do the parsing and validation. You can see this occurring even in this PR with the for loops which build the `KeyValueParameter` objects, only to rip them apart in the next line.

Given where we want to be, the goal of this PR should be to write a utility or abstraction such that we don't have a lot of maintenance in code going forward, and it's not clunky or cumbersome to use. Needing to pass around the translation provider is a hint that our abstraction is too leaky, and there's a lot of legacy code from the old abstractions that just doesn't need to exist anymore.

If we can't get the OpenAPI generator to do the parsing magic itself for us by way of a deserializer we provide, then I feel our next best option is a fairly simple utility class which provides bare-bones deserialization of a single kvpair string or collections thereof. There's an argument to be made about where the responsibility of exception generation should occur, and that can be handled in a number of ways. If we scope the utility class to the resource package, it could very well just generate the appropriate `BadRequestException` itself. Otherwise, if we want to retain some of the old exception logic and mappers for whatever reason, we could just update the CandlepinParameterParseException to do proper translation of its strings. However, that gets messy in a number of cases and probably ends up requiring static injection or some other clunky means of providing the translation bits (not to mention the difficulty and error-prone code in translating the individual pieces that will exist elsewhere). Also, we only ever use it in one place, so it doesn't really have a ton of value as it is.

For example, we could do a lot with a single parser class:
```
public class KeyValueStringParser {
    private static final Pattern PARSE_PATTERN = Pattern.compile(":");

    @Inject
    private static Provider<I18n> i18nProvider;

    public KeyValueStringParser() {
        throw new UnsupportedOperationException();
    }

    public static List<Pair<String, String>> parseKeyValuePairs(Iterable<String> kvpStrings) {
        List<Pair<String, String>> output = new ArrayList<>();

        if (kvpStrings != null) {
            kvpStrings.forEach(kvpstr -> output.add(parseString(kvpstr)));
        }

        return output;
    }

    public static List<Pair<String, String>> parseKeyValuePairs(String... kvpStrings) {
        return parseKeyValuePair(kvpStrings != null ? List.of(kvpStrings) : null);
    }

    public static Pair<String, String> parseKeyValuePair(String kvpString) {
        return kvpString != null ? parseString(kvpString) : null;
    }

    private static Pair<String, String> parseString(String kvpString) {
        if (kvpString == null) {
            throw new BadRequestException(buildExceptionString(null));
        }

        String[] parts = PARSE_PATTERN.split(kvpString,  2);
        if (parts.length < 2) {
            throw new BadRequestException(buildExceptionString(kvpString));
        }

        return Pair.of(parts[0], parts[1]);
    }

    private static String buildExceptionString(String... args) {
        I18n i18n = i18nProvider.get();
        return i18n.tr("Failed to parse parameter \"{0}\"; expected format: key:value", args);
    }

}
```

Using this, the facts loop in ConsumerResource.searchConsumers becomes:
```
KeyValueStringParser.parseKeyValuePairs(facts)
    .forEach(kvpair -> queryArgs.addFact(kvpair.getKey(), kvpair.getValue()));
```

We could also add some additional utility methods for converting to maps or streams as well, depending on need.

Ceiu:
Obsoleted by #3623 due to outages

PR #3622, opened 2022-10-11
ENT-5390: Fix bundle install

Januson:
- Bundler had trouble accessing rubygems when ipv6 was used. All requests were timing-out. This fix changes the precedence so that ipv4 is used for accessing rubygems.
- Centos 7 was missing ruby-devel dependency which caused issues for some of the gems.


Ceiu:
Can we include more details on the change here? What was wrong with bundle install?

Ceiu (approved):

PR #3623, opened 2022-10-11
[M] ENT-5448: Updated how key-value strings are handled at the API level

Ceiu:
- Removed the KeyValueParamDTO and replaced all endpoints using it with a simple collection of strings
- Removed all abstractions, converters, or interceptors built to process and convert KeyValueParamDTO instances, as they no longer have use with the removal of the KeyValueParamDTO
- Removed CandlepinParameterParseException and its associated mappers, as this exception was very oddly built, only used for a very specific conversion, and the error message wasn't translating properly
- Added a new utility class, KeyValueStringParser, to provide utility functions for key-value strings into lists, streams, or maps of key-value pairs
- Updated several endpoints and tests accordingly


Ceiu:
This change addresses a shortcoming in how the OpenAPI generated client bits handled the KeyValueParamDTO stuff. On the server end, it handled the conversion "fine" (though it required converters to do it), but we had no such thing on the client end. This would result in the query parameter being serialized into an object rather than the desired key-value string:

`fact=class%20KeyValueParamDTO%20%7B%0A%20%20%20%20key%3A%20fact1%0A%20%20%20%20value%3A%20value1%0A%7D`

The immediate fix put in by others in a handful of places was simply to remove the usage of the KeyValueParamDTO and use simple strings instead. This works fine enough, but introduced several places where we'd need to do manual parsing and validation of the strings -- which is annoying and prone to error and inconsistencies. 

This PR aims to provide a standardized utility for performing the parsing and validation, in a way that is convenient to use (one or two lines in the general case). Additionally, with the removal of the KeyValueParamDTO, several bits of now-obsolete code were also removed and, where appropriate, replaced with something using the new utility.

As a note, the way it's currently setup, this does make use of some static injection, which is kinda jank, but overall pretty clean from a usage perspective. There's another way to go about this -- by making this class require instantiation where it's used and then pass in the required i18n bits at build time -- but that's clunky in its own right. I'm open to feedback on this front, but I want to reiterate that the focus is for the usability of the utility to be very lightweight and simple.

src/main/java/org/candlepin/resource/util/KeyValueStringParser.java
```diff
@@ -0,0 +1,271 @@
...
+ * the exceptions are intended to be passed through to the user and contain both the appropriate
+ * HTTP error code and translated error message. Use of these functions should not require
+ * additional error handling or input validation in the general case.
+ * <p></p>
+ * Key-value strings are parsed according to the format "{key}:{value}", where only the first colon
+ * is recognized as a delimiter, and all additional colons will be rolled into the value. If the
+ * string is null, empty, or otherwise does not contain any instances of the delimiter, it is
+ * considered malformed, and will cause functions in this class to throw a BadRequestException.
+ */
+public class KeyValueStringParser {
```
Januson, line 51: Could we turn this into normal utility class with constructor injection? Static injections are greatly discouraged even by the guice team. They are inconvenient to test and will make maintenance of our API layer difficult.

src/test/java/org/candlepin/resource/util/KeyValueStringParserTest.java
```diff
@@ -0,0 +1,616 @@
...
+
+    @Test
+    public void testParseArrayDisallowsNullValuesInCollection() {
+        String[] kvstrings = new String[] { "key1:value1", null, "key3:value3" };
+
+        assertThrows(BadRequestException.class, () -> KeyValueStringParser.parseKeyValuePairs(kvstrings));
+    }
+
+    @Test
+    public void testParseArrayDisallowsEmptyValuesInCollection() {
```
JoshAlbrecht23, line 279: You could add another test for a blank string as well. This can also apply to some of the other tests like where you are testing mapKeyValuePairs with an Iterable

Januson (changes requested):
Changes are looking good. It's cool that we can specify string formats like that in the spec. I have only one comment regarding the static injection.

src/test/java/org/candlepin/resource/util/KeyValueStringParserTest.java
Ceiu: I don't understand the request. This test does handle empty strings already (parameter 2). If you're referencing a string containing whitespace (i.e. `"  "`), is there something the undelimited case isn't covering that you'd like to see?

src/main/java/org/candlepin/resource/util/KeyValueStringParser.java
Ceiu: I've got a commit prepped where this change has been made, but I don't see what we gain from it outside of avoiding static injection. The KeyValueStringParser is stateless by design, so the change ends up going from:

```java
KeyValueStringParser.parseKeyValuePairs(attrFilters)
  .forEach(kvpair -> poolFilters.addAttributeFilter(kvpair.getKey(), kvpair.getValue()));
```

to:

```java
new KeyValueStringParser(this.i18n).parseKeyValuePairs(attrFilters)
  .forEach(kvpair -> poolFilters.addAttributeFilter(kvpair.getKey(), kvpair.getValue()));
```

which is basically the same thing, but with manual construction of an object that is immediately discarded.

We could inject a provider to each resource that uses it, which would change it to:

```java

@Inject
public OwnerResource(..., Provider<KeyValueStringParser> kvsParserProvider, ...) {
  ...

  this.kvsParserProvider = Objects.requireNonNull(kvsParserProvider);
  // Alternatively, this could just be the KeyValueStringProvider directly
}

...

this.kvStringParserProvider[.get()].parseKeyValuePairs(attrFilters)
  .forEach(kvpair -> poolFilters.addAttributeFilter(kvpair.getKey(), kvpair.getValue()));
```

This works, but now we need to update the constructor so we accept and store a reference to a utility that, again, isn't storing any state and is, effectively an implementation detail. We're using a composition pattern to inject fixed behavior to avoid injection (which really is already just avoiding making the resources have to handle and translate an exception message).

And then the tests need to be updated to pass in the provider/instance on invocation. But the core idea is to *not* have different implementations of that parsing, so I don't see why we would want control over those details or need explicit testing around knowledge of the calls into this utility outside of explicit testing of it in the targeted suite.

Something I had considered to ease any static injection pain is having a default i18n instance as the default, and let static injection override it:

```java
    private static Provider<I18n> i18nProvider = () -> 
        I18nFactory.getI18n(KeyValueStringParser.class, Locale.US, I18nFactory.FALLBACK);

    @Inject
    public static Provider<I18n> initialize(Provider<I18n> i18nProvider) {
        if (i18nProvider == null) {
            throw new IllegalArgumentException("i18nProvider is null");
        }

        Provider<I18n> prev = KeyValueStringParser.i18nProvider;
        KeyValueStringParser.i18nProvider = i18nProvider;

        return prev;
    }
```

This would get rid of any test injector setup (though it's only needed in the test suite for this class), and allow for one-off changes if necessary.

Is there something I'm overlooking here? Any other thoughts on it?

Ceiu:
Something I should probably note: the "key:value" format in the string doesn't actually do anything; I'm just following the convention already set in place.

OpenAPI generators are supposed to silently ignore formats they don't support and treat them as the default format for a given type, and, unfortunately, the format is not displayed in the Swagger UI, so it looks like it's for dev documentation purposes only.

JoshAlbrecht23 (approved):

src/main/java/org/candlepin/resource/util/KeyValueStringParser.java
Januson: Creating the instance where needed could work. It's just a small utility class so we can justify it not being injected into resources. The main issue is that static injection creates a dependency on Guice which will get in our way when we'll want to migrate to different injection framework.

Ceiu: ack; will convert to non-static utility and locally created instances

Ceiu: done

Januson (approved):

PR #3624, opened 2022-10-12
[M] Eliminate trustedConsumer/trustedUser usage in spec tests

nikosmoum:
- While porting spec tests from ruby to java, there has been a heavy use of ApiClients.trustedConsumer and ApiClients.trustedUser auth methods which are using the 'magic' headers (cp-consumer/cp-user). The ruby tests never used those, and instead use the identity cert + key and username + password for authentication. This change switches all existing java spec tests to use the ApiClients.basic(username, password) and ApiClients.ssl(consumer) which are the more realistic authentication use cases.

### DIFF
--- a/spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClientFactory.java
+++ b/spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClientFactory.java
@@ -80,9 +80,9 @@ public class ApiClientFactory {
         return apiClient;
     }
 
-    public ApiClient createCloudAuthClient(String token) {
+    public ApiClient createBearerTokenAuthClient(String token) {
         ApiClient apiClient = createDefaultClient();
-        apiClient.setHttpClient(createOkHttpClient(new CloudAuthInterceptor(token)));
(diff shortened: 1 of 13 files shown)


nikosmoum:
Note for developers:
When porting tests, here are the equivalent authentication methods to use for each ruby method:
```
# Ruby => Java
Candlepin.new(username, password) or user_client() => ApiClients.basic(username, password)

Candlepin.new('admin', 'admin') => ApiClients.admin()

Candlepin.new(nil, nil, cert, key) or consumer_client() => ApiClients.ssl(consumerDTO)

Candlepin.new => ApiClients.noAuth()
```

Ceiu:
Changes look good to me, but we should add some documentation on each of those methods in `ApiClients` so this doesn't happen again in the future. Especially since once this PR is merged, that comment is going to get buried.

Ceiu (changes requested):
Changes look good to me, but we should add some documentation on each of those methods in `ApiClients` so this doesn't happen again in the future. Especially since once this PR is merged, that comment is going to get buried.

nikosmoum:
@Ceiu Done, also I've renamed the cloud auth client into bearer token client since that is more accurate

PR #3625, opened 2022-10-14
[M] 2134502: Satellite cannot enable or sync satellite-tools repo

wottop:
- The path tree condensing algorithm was dropping nodes in some specific cases
- This code was untouched for 10 years and only after a change was made to the
  hashing of content did it surface.
- Included are sets of data, where it previously did and did not work correctly,
  in unit tests so that we will see any future problems sooner.
- The urls in the files should all be compressed and decompressed correctly
  if the algorithm is operating properly.

### DIFF
--- a/src/main/java/org/candlepin/util/X509V3ExtensionUtil.java
+++ b/src/main/java/org/candlepin/util/X509V3ExtensionUtil.java
@@ -589,7 +589,8 @@ public class X509V3ExtensionUtil extends X509Util {
                     // all grandparents with name now point to merged node
                     for (PathNode pn : toRemove.getParents()) {
                         for (NodePair child : pn.getChildren()) {
-                            if (child.getName().equals(name)) {
+                            if (child.getName().equals(name) &&
+                                child.getConnection().isEquivalentTo(merged)) {
                                 child.setConnection(merged);
                             }
                         }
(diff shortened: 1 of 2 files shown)


Ceiu:
Can we update the title to fit the 70 character limit, and then add some bullets about what the targeted change was?

src/test/resources/urlcompression/content_from_11.txt
```diff
@@ -0,0 +1,843 @@
+/content/rhb/rhel/client/5/5Client/$basearch/devtoolset/debug
```
Ceiu, line 1: Is it necessary to have these lengthy list of repo paths as a separate file for the reproducer? Could we pair this down to only the handful we need to reproduce the issue in code and write specific tests surrounding them?

Trying to read this, I can't tell which paths are or aren't valid with respect to the test at hand.

barnabycourt:
might be simpler with a more paired down test set for the failure mode was 
`/content/dist/rhel/server/6/$releasever/$basearch/satellite/6.0/os
/content/dist/rhel/server/6/$releasever/$basearch/satellite/6.0/source/SRPMS
/content/dist/layered/rhel9/x86_64/sat-client/6/source/SRPMS
/content/beta/layered/rhel8/x86_64/sat-tools/6/source/SRPMS
/content/beta/layered/rhel8/x86_64/sat-tools/6/os
/content/dist/layered/rhel8/x86_64/sat-tools/6.8/os
/content/dist/layered/rhel8/x86_64/sat-client/6/os
/content/dist/layered/rhel9/x86_64/sat-client/6/os
/content/dist/layered/rhel8/x86_64/sat-client/6/source/SRPMS`

nikosmoum (approved):
Looks good to me. I'll let @Ceiu re-review & merge.
Also, we should port this to 4.1 too

Ceiu:
@nikosmoum @wottop 
Just waiting for the test data to get paired down. Looking at it, it's still not at all clear which bits of those are necessary or what overlap was causing problems. Ideally we'd have only two conflicting paths for the test case(s), or whatever pairs of paths that can break output as a parameterized test data provider.

Basically, I just want it to be in a place where we have tests that directly highlight the issue rather than a huge test file.

Ceiu (approved):

PR #3627, opened 2022-10-17
[M] ENT-5370: Migrate oauth_spec to Java

bakajstep:
- Added support for OAuth1 API client
- Migrate oauth_spec.rb
- Delete oauth_spec.rb


bakajstep:
serialization_spec.rb is an old code that we don't need to migrate.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClients.java
```diff
@@ -129,6 +129,43 @@ public static ApiClient bearerToken(String token) {
...
+     */
+    public static ApiClient oauthUser(String oauthConsumer, String oauthSecret, String username) {
+        return new ApiClient(CLIENT_FACTORY.createOauthClient(oauthConsumer, oauthSecret)
+            .addDefaultHeader("cp-user", username));
+    }
+
+    /**
+     * Returns a client to make API calls with, that uses an OAuth consumer authentication.
+     *
+     * @param oauthConsumer an OAuth consumer
```
Ceiu, line 159: The term "consumer" is overloaded in this context (OAuth and Candlepin both have their own definitions of a consumer), so clarification is needed to differentiate them.

spec-tests/src/test/java/org/candlepin/spec/auth/OauthSpecTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Test
+    void shouldReturnsAUnauthorizedIfOauthSecretDoesNotMatch() {
+        ApiClient oauth = ApiClients.oauth(OAUTH_CONSUMER, "badsecret");
+        assertUnauthorized(() -> oauth.users().listUsers());
+    }
+
+    @Test
+    void shouldLetsACallerActAsAUser() {
```
Ceiu, line 55: This method, and the one following it, should read "shouldLetACaller..." -- the "Lets" should not be pluralized.

spec-tests/src/test/java/org/candlepin/spec/auth/OauthSpecTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Test
+    void shouldLetsACallerActAsAConsumer() {
+        ApiClient adminClient = ApiClients.admin();
+        OwnerDTO owner = adminClient.owners().createOwner(Owners.random());
+        ConsumerDTO consumer = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ApiClient oauthConsumer = ApiClients.oauthConsumer(
+            OAUTH_CONSUMER, OAUTH_SECRET, consumer.getUuid());
+        ConsumerDTO consumerFromServer = oauthConsumer.consumers().getConsumer(consumer.getUuid());
+        assertThat(consumerFromServer.getId())
```
Ceiu, line 73: Add a null check and return check here:

```java
assertThat(consumerFromServer)
  .isNotNull()
  .returns(consumer.getId(), ConsumerDTO::getId);
```

Ceiu (changes requested):

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClients.java
bakajstep: is that enough, or should I add something else I'm not sure about the acting like sentence

spec-tests/src/test/java/org/candlepin/spec/auth/OauthSpecTest.java
bakajstep: done

spec-tests/src/test/java/org/candlepin/spec/auth/OauthSpecTest.java
bakajstep: done

spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/ApiClients.java
Ceiu: The "acting like" sentences are extraneous and somewhat confusing, but the addition of "Candlepin" on the param is sufficient here.

Ceiu (approved):

PR #3636, opened 2022-10-22
[M] ENT-5377: Migrate import_update and import_excess spec to Java

wottop:
Merged the ruby into one java file to reduce common code
 and considering that the excess spec is an update test


spec-tests/src/test/java/org/candlepin/spec/imports/ImportUpdateSpecTest.java
```diff
@@ -0,0 +1,173 @@
...
+        //  before creating another export.
+        Thread.sleep(2000);
+        Export updatedExport = exportGenerator.updateWithProductNoEnt().export();
+        doImport(importOwner.getKey(), updatedExport.file());
+
+        assertThat(exportGenerator.getManifestConsumerEntitlements()).hasSize(0);
+        assertThat(admin.owners().getOwnerSubscriptions(importOwner.getKey())).hasSize(0);
+    }
+
+    private void doImport(String ownerKey, File export) {
```
Ceiu, line 165: Why are we abstracting away a one-liner? If this is done commonly enough that the `List.of()` is becoming burdensome, we should be adding an overload of `importManifestAsync` which takes only the ownerKey and export file instead of this.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ExportGenerator.java
```diff
@@ -177,4 +250,17 @@ private PoolDTO createPool(String ownerKey, ProductDTO product) throws ApiExcept
 
         return ownerApi.createPool(ownerKey, pool);
     }
+
+    public ContentDTO getContent() {
```
Ceiu, line 254: This won't work correctly for exports that attach more than one product or content.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/ExportGenerator.java
```diff
@@ -134,7 +137,77 @@ public ExportGenerator withProduct(ProductDTO product, ContentDTO content) {
         return this;
     }
 
-    private File createExport(String consumerUuid, ExportCdn cdn) {
+    public ExportGenerator updateWithProduct() {
```
Ceiu, line 140: This should not be done as an abstracted bit of logic at this level.

Really, the ExportGenerator should not exist at all, but if it's going to continue living on, it should be little more than boilerplate reduction and not inherit any of the old "does everything for everyone, everywhere" ideology that the original rspec code had.

Ceiu (changes requested):

Ceiu (approved):

PR #3640, opened 2022-10-25
[M] Migrate content_versioning_spec to Java (ENT-5454)

JoshAlbrecht23:
- Removed content_versioning_spec.rb


spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
```diff
@@ -0,0 +1,314 @@
...
+
+@SpecTest
+public class ContentVersioningSpecTest {
+    private ApiClient adminClient;
+    private OwnerDTO owner;
+
+    @BeforeEach
+    public void beforeEach() {
+        adminClient = ApiClients.admin();
+        owner = adminClient.owners().createOwner(Owners.random());
```
Ceiu, line 47: Let's move the owner creation to the tests themselves, and if we're going to retain the adminClient in this scope, change this to a beforeAll.

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
```diff
@@ -0,0 +1,314 @@
...
+
+    @Test
+    public void shouldCreateOneContentInstanceWhenSharedByMultipleOrgs() throws Exception {
+        OwnerDTO owner2 = adminClient.owners().createOwner(Owners.random());
+        ContentDTO expectedContent = Content.random();
+
+        ContentDTO owner1Content = adminClient.ownerContent().createContent(owner.getKey(), expectedContent);
+        ContentDTO owner2Content = adminClient.ownerContent().createContent(owner2.getKey(), expectedContent);
+
+        assertThat(owner1Content)
```
Ceiu, line 58: We're missing the most critical aspect of this test: verifying that the UUIDs on both content is the same. I know this is probably covered by the `isEqualTo` check, but given that it's the core component here, I would expect it to be done explicitly.

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
```diff
@@ -0,0 +1,314 @@
...
+    public void shouldCreateANewInstanceWhenMakingChangesToAnExistingInstance() throws Exception {
+        ContentDTO content = adminClient.ownerContent().createContent(owner.getKey(), Content.random());
+
+        content.setName(content.getName() + "-2");
+        ContentDTO updatedContent = adminClient.ownerContent()
+            .updateContent(owner.getKey(), content.getId(), content);
+
+        assertThat(updatedContent)
+            .isNotNull()
+            .extracting(ContentDTO::getUuid)
```
Ceiu, line 131: This could be a `.doesNotReturn` check instead

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
```diff
@@ -0,0 +1,314 @@
...
+
+            Runnable owner2ContentGeneration = () -> {
+                for (int j = batchOffset; j < batchOffset + numberOfContent; j++) {
+                    String id = contentIdPrefix + j;
+                    ContentDTO createdContent = adminClient.ownerContent()
+                        .createContent(owner2.getKey(), Content.random().id(id));
+                    owner2CreatedContentUuids.add(createdContent.getUuid());
+                }
+            };
+            Thread owner2GenerationThread = new Thread(owner2ContentGeneration);
```
Ceiu, line 284: We can condense this down by removing the variable declaration of the runnable above and rearranging things a bit:

```java
Thread owner1UpdatesThread = new Thread(() -> {
  // runnable contents
});

Thread owner2GenerationThread = new Thread(() -> {
  // runnable contents
});

owner1UpdatesThread.start();
owner2GenerationThread.start();

AsyncJobStatusDTO job = adminClient.jobs().scheduleJob("OrphanCleanupJob");

owner1UpdatesThread.join(60000);
owner2GenerationThread.join(60000);
```

Also note that the job invocation needs to occur *before* we wait for the update and  creation loops to complete.

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
```diff
@@ -0,0 +1,314 @@
...
+            owner2GenerationThread.start();
+
+            // Wait for the threads to finish
+            owner1UpdatesThread.join(10000);
```
Ceiu, line 288: 10s is probably very tight for this test. We should bump this to 30 or 60s, possibly even longer. Especially since if it happens to finish faster, it won't wait the full duration.

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
```diff
@@ -0,0 +1,314 @@
...
+            owner2GenerationThread.join(10000);
+
+            AsyncJobStatusDTO job = adminClient.jobs().scheduleJob("OrphanCleanupJob");
```
Ceiu, line 291: I've commented on this above, but to make sure it's individually captured: this needs to occur between the worker threads starting and waiting for them to terminate.

Ceiu (changes requested):

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
JoshAlbrecht23: Moved the owner creation and the adminClient creation to the individual tests

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
JoshAlbrecht23: That's a good point, we should explicitly verify that the UUIDs are the same and not rely on the `isEqualTo` check. Updated

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
JoshAlbrecht23: That makes sense. Updated

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/content/ContentVersioningSpecTest.java
JoshAlbrecht23: I updated to 90s to be on the safer side

Ceiu (approved):

PR #3641, opened 2022-10-26
ENT-5453: Migrate product_versioning_spec to Java

wottop:


spec-tests/src/test/java/org/candlepin/spec/products/ProductVersioningSpecTest.java
```diff
@@ -0,0 +1,664 @@
...
+            Products.random().id(id).name(name));
+        assertEquals(prod1.getUuid(), prod2.getUuid());
+        assertEquals(prod1.getUuid(), prod3.getUuid());
+
+        ProductDTO prod4 = ownerProductApi.updateProductByOwner(owner2.getKey(), id,
+            new ProductDTO().name("new product name"));
+        assertNotEquals(prod4.getUuid(), prod1.getUuid());
+
+        List<ProductDTO> prods = ownerProductApi.getProductsByOwner(owner2.getKey(), new ArrayList<>());
+        assertThat(prods).hasSize(1);
```
Ceiu, line 139: These can be turned into assertion chains:

```java
assertThat(prods)
  .isNotNull()
  .singleElement()
  .returns(prod4.getUuid(), ProductDTO::getUuid)
  .doesNotReturn(prod2.getUuid(), ProductDTO::getUuid);
```

`.singleElement` checks that the object is a collection that contains exactly one element, but then also extracts that one element for further evaluation.

Alternatively, you can use the `extracting` method to get the UUID before doing normal equals comparisons:

```java
assertThat(prods)
  .isNotNull()
  .singleElement()
  .extracting(ProductDTO::getUuid)
  .isEqualTo(prod4.getUuid())
  .isNotEqualTo(prod2.getUuid());
```

spec-tests/src/test/java/org/candlepin/spec/products/ProductVersioningSpecTest.java
```diff
@@ -0,0 +1,664 @@
...
+                    ProductDTO prod = ownerProductApi.createProductByOwner(owner2.getKey(),
+                        new ProductDTO().name(id).id(id));
+                    o2Uuids.add(prod.getUuid());
+                }
+            });
+            generator.start();
+
+            sleep(1000);
+            AsyncJobStatusDTO job = client.jobs().scheduleJob("OrphanCleanupJob");
+            updater.join();
```
Ceiu, line 541: We should add an upper-bound on these joins so we don't wait indefinitely in the event something gets stuck somewhere. In the content version of these tests we went with 90s.

spec-tests/src/test/java/org/candlepin/spec/products/ProductVersioningSpecTest.java
```diff
@@ -0,0 +1,664 @@
...
+            generator.join();
+            AsyncJobStatusDTO jobStatus = client.jobs().waitForJob(job);
+            assertEquals("FINISHED", jobStatus.getState());
+
+            // Verify the products created/updated still exist
+            assertThat(o1Uuids.stream()
+                .filter(x -> client.products().getProduct(x) != null)
```
Ceiu, line 548: If we're going to try to do this in one go like this, the assertion mapper would be a better option:

```java
assertThat(o1Uuids)
  .map(uuid -> client.products().getProduct(uuid))
  // other checks here
```

After the call to .map, the collection will contain product entities which can then be verified however necessary.

spec-tests/src/test/java/org/candlepin/spec/products/ProductVersioningSpecTest.java
```diff
@@ -0,0 +1,664 @@
...
+        OwnerDTO owner2 = ownerApi.createOwner(Owners.random());
+
+        String prefix = "test-product-";
+        int offset = 10000;
+        int length = 100;
+
+        // Repeat this test a few(ish) times to hopefully catch any synchronization error
+        for (int i = 0; i < 10; i++) {
+            final int base = offset;
+            List<String> o1Uuids = new ArrayList<>();
```
Ceiu, line 507: Might want to use synchronized lists here to avoid failures as a result of local race conditions.

Ceiu (changes requested):
Seems to match the base logic overall, but there are a couple issues in the thread-based tests, and some AssertJ usage that could be done throughout to improve the test readability.

Ceiu (approved):

PR #3642, opened 2022-10-27
[M] Fix flaky activation key list test

nikosmoum:
- The test would sometimes fail because the comparison of the activation key lists was taking into account the order of the returned list, which shouldn't matter.


wottop:
ack

PR #3645, opened 2022-10-28
Use Containerfile to build image, podman-compose to test

redsaz:


nikosmoum:
Thanks for opening the PR; we'll definitely consider the approach, but closing this for now since there's a lot of changes we agreed on (e.g. using a UBI8 minimal as base) and even more things to investigate (e.g how to deal with config files and certs) before moving on with an implementation

PR #3655, opened 2022-11-01
[M] Migrate content_access_spec to Java (Part 1) (ENT-5465)

JoshAlbrecht23:
- migrated half of the spec tests in content_access_spec.rb
- removed the migrated tests from content_access_spec.rb


spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Owners.java
```diff
@@ -28,7 +28,7 @@
  */
 public final class Owners {
 
-    public static final String ACCESS_MODE_LIST = "org_environment,entitlement";
+    public static final String ACCESS_MODE_LIST = "entitlement,org_environment";
```
JoshAlbrecht23, line 31: Updated the ordering to match the default db content access mode string.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -0,0 +1,860 @@
...
+        ApiClient adminClient = ApiClients.admin();
+        OwnerDTO scaOwner = adminClient.owners().createOwner(Owners.randomSca());
+        assertThat(scaOwner)
+            .returns(Owners.SCA_ACCESS_MODE, OwnerDTO::getContentAccessMode)
+            .returns(Owners.ACCESS_MODE_LIST, OwnerDTO::getContentAccessModeList);
+
+        // If we remove SCA mode from the list, the mode should also be forced to the default (entitlement)
+        ObjectNode objNode = ApiClient.MAPPER.createObjectNode();
+        objNode.put("contentAccessModeList", Owners.ENTITLEMENT_ACCESS_MODE);
+        Request.from(adminClient)
```
JoshAlbrecht23, line 235: This was interesting. In the original test ('sets mode to default when the list is updated to no longer have the original mode value') the body of the request is just `{'contentAccessModeList' : 'entitlement'}`. If I try and update the modified OwnerDTO using adminClient.owners().updateOwner(), and exception is thrown because the owner's contentAccessMode (org_environment) is no longer in the contentAccessModeList.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -0,0 +1,861 @@
...
+        String updatedContent = listings.get(1).asText();
+        String updatedLastUpdate = contentAccessBody.get("lastUpdate").asText();
+
+        assertEquals(originalX509, updatedX509);
+        assertNotEquals(originalContent, updatedContent);
+        assertNotEquals(originalLastUpdate, updatedLastUpdate);
+    }
+
+    @Test
+    public void shouldReturnANotModifiedReturnCodeWhenTheDataHasNotBeenUpdatedSinceDate() throws Exception {
```
JoshAlbrecht23, line 746: I'm seeing some sporadic results when running this method multiple times. Looking into this, it looks like this may be an issue with the actual logic with the implementation in the ContentAccessManager:

https://github.com/candlepin/candlepin/blob/master/src/main/java/org/candlepin/controller/ContentAccessManager.java#L561-L571

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -0,0 +1,864 @@
...
+        String certContentName = content.get(0).get("name").asText();
+        assertNotNull(certContentName);
+        JsonNode certContent1 = certContentName.equals(content1.getName()) ? content.get(0) : content.get(1);
+        JsonNode certContent3 = certContentName.equals(content3.getName()) ? content.get(0) : content.get(1);
+
+        assertEquals(content1.getType(), certContent1.get("type").asText());
+        assertEquals(content1.getName(), certContent1.get("name").asText());
+        assertEquals(content1.getLabel(), certContent1.get("label").asText());
+        assertEquals(content1.getVendor(), certContent1.get("vendor").asText());
+        assertEquals(content1.getArches(), certContent1.get("arches").get(0).asText());
```
Januson, line 116: There are a few pieces that might be worth extracting to make tests a bit more readable. For example this might be extracted as an assertion method for comparing content/certContent.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -0,0 +1,864 @@
...
+
+    @Test
+    public void shouldAssignTheDefaultModeAndListWhenEmptyStringsAreSpecified() throws Exception {
+        ApiClient adminClient = ApiClients.admin();
+        OwnerDTO owner = new OwnerDTO()
+            .id(StringUtil.random("id-"))
+            .key(StringUtil.random("key-"))
+            .displayName(StringUtil.random("display-name-"))
+            .contentAccessMode("")
+            .contentAccessModeList("");
```
Januson, line 172: It might be better to just use random owner and unset its content access.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -0,0 +1,864 @@
...
+        List<String> payloadCerts = extractCertsFromPayload(export);
+        assertThat(payloadCerts).singleElement();
+        byte[] value = CertificateUtil
+            .compressedContentExtensionValueFromCert(payloadCerts.get(0), "1.3.6.1.4.1.2312.9.7");
+        X509HuffmanDecodeUtil decode = new X509HuffmanDecodeUtil();
+        List<String> urls = decode.hydrateContentPackage(value);
+        assertThat(urls).contains(CandlepinMode.isStandalone() ? "/" + ownerKey : "/sca/" + ownerKey);
+
+        String extVal = CertificateUtil
+            .standardExtensionValueFromCert(payloadCerts.get(0), "1.3.6.1.4.1.2312.9.8");
```
Januson, line 289: It might be worth extracting the magic strings like `1.3.6.1.4.1.2312.9.8` into named constants to make it obvious what we are extracting.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
Januson: You can do this via `owners().updateOwner()` but you have to set `contentAccessMode` to null to not sent it.

Januson (changes requested):
Changes look good. There are just a couple places that could use a bit of cleanup.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -0,0 +1,860 @@
...
+        ProductDTO prod = adminClient.ownerProducts().createProductByOwner(ownerKey, Products.random());
+        ContentDTO content = adminClient.ownerContent().createContent(ownerKey, Content.random());
+        adminClient.ownerProducts().addContent(ownerKey, prod.getId(), content.getId(), true);
+        adminClient.owners().createPool(ownerKey, Pools.random(prod));
+
+        ConsumerDTO consumer = adminClient.consumers()
+            .createConsumer(Consumers.random(owner).facts(Map.of("system.certificate_version", "1.0")));
+        ApiClient consumerClient = ApiClients.ssl(consumer);
+
+        Thread.sleep(1000);
```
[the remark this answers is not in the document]
JoshAlbrecht23: We shouldn't need to sleep here. I removed the sleep and ran the test 30 times to help verify and all the tests passed.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
JoshAlbrecht23: Awesome! I updated the test to use `owners().updateOwner()` instead of building the Request object.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
JoshAlbrecht23: I agree! Updated here and throughout the other tests

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
JoshAlbrecht23: Good idea, I added the `compareCertContent` method to handle these assertions

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
JoshAlbrecht23: Added ENT_DATA_OID and ENT_TYPE_OID as named constants. If I'm understanding the OIDs correctly, I think the names make sense.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -0,0 +1,860 @@
...
+
+        List<JsonNode> certs = consumerClient.consumers().exportCertificates(consumer.getUuid(), null);
+        assertThat(certs).hasSize(0);
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -0,0 +1,860 @@
...
+        List<JsonNode> updatedCerts = CertificateUtil
+            .extractEntitlementCertificatesFromPayload(export, ApiClient.MAPPER);
+        assertThat(updatedCerts).singleElement();
+        payloadCerts = extractCertsFromPayload(export);
+        assertThat(payloadCerts).singleElement();
+        CertificateUtil.standardExtensionValueFromCert(payloadCerts.get(0), "1.3.6.1.4.1.2312.9.8");
+        assertThat(extVal).isEqualTo("OrgLevel");
+
+        // verify certificate serials were updated.
+        assertThat(updatedSerialIds).doesNotContainAnyElementsOf(originalSerialIds);
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good catch, updated

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -0,0 +1,860 @@
...
+        envName = envName != null ? "/" + envName : "";
+        String expected = CandlepinMode.isStandalone() ? "/" + ownerKey + envName + contentUrl : contentUrl;
+        assertEquals(expected, cerContentPath);
+    }
+
+    private List<String> getSerialIdsFromCertExport(Object export) {
+        List<Map<String, LinkedTreeMap>> certsToKeyVal = ((List<Map<String, LinkedTreeMap>>) export);
+        List<String> serialIds = new ArrayList<>();
+        certsToKeyVal.stream()
+            .forEach(entry -> {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good catch, updated both spots

PR #3667, opened 2022-11-09
CANDLEPIN-83: Migrate system_admin_spec to Java

wottop:


spec-tests/src/test/java/org/candlepin/spec/SystemAdminSpecTest.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    private UserDTO createUserType2(OwnerDTO owner) {
+        return UserUtil.createWith(client,
+            new PermissionBlueprintDTO().type("USERNAME_CONSUMERS").access("ALL")
+                .owner(new NestedOwnerDTO().id(owner.getId()).key(owner.getKey())),
+            new PermissionBlueprintDTO().type("OWNER").access("READ_ONLY")
+                .owner(new NestedOwnerDTO().id(owner.getId()).key(owner.getKey())),
+            new PermissionBlueprintDTO().type("ATTACH")
+                .owner(new NestedOwnerDTO().id(owner.getId()).key(owner.getKey())));
```
Januson, line 351: You can use Permissions utility to simplify creation of permissions. E.G. This could be written as
```java
return UserUtil.createWith(client,
    Permissions.USERNAME_CONSUMERS.all(owner),
    Permissions.OWNER.readOnly(owner),
    Permissions.ATTACH.all(owner));
```

spec-tests/src/test/java/org/candlepin/spec/SystemAdminSpecTest.java
```diff
@@ -0,0 +1,358 @@
...
+    public void shouldCreateEntitlementsAsynchronously() throws Exception {
+        OwnerDTO owner = ownerApi.createOwner(Owners.random());
+        UserDTO user = createUserType1(owner);
+        ApiClient userClient = ApiClients.basic(user);
+
+        ConsumerDTO consumer1 = userClient.consumers().createConsumer(
+            Consumers.random(owner).username(user.getUsername()));
+
+        PoolDTO pool = createPool(owner);
+        AsyncJobStatusDTO job = AsyncJobStatusDTO.fromJson(userClient.consumers().bind(consumer1.getUuid(),
```
Januson, line 142: You can use bindPool to get rid of the nulls. There is no need to deserialize the job. You can wait for it by id. Also we should probably assert that entitlements were created.

spec-tests/src/test/java/org/candlepin/spec/SystemAdminSpecTest.java
```diff
@@ -0,0 +1,358 @@
...
+        return UserUtil.createWith(client,
+            new PermissionBlueprintDTO().type("USERNAME_CONSUMERS")
+                .owner(new NestedOwnerDTO().id(owner.getId()).key(owner.getKey())),
+            new PermissionBlueprintDTO().type("OWNER_POOLS")
+                .owner(new NestedOwnerDTO().id(owner.getId()).key(owner.getKey())),
+            new PermissionBlueprintDTO().type("ATTACH")
+                .owner(new NestedOwnerDTO().id(owner.getId()).key(owner.getKey())));
+    }
+
+    private UserDTO createUserType2(OwnerDTO owner) {
```
Januson, line 344: Name could use more details to distinguish the two users. Similar to how ruby describes the scenario as read-only org.

Januson (changes requested):
Changes look good. I have just a few minor comments.

spec-tests/src/test/java/org/candlepin/spec/SystemAdminSpecTest.java
wottop: the bindPool in ConsumerClient is async false. I need it to be true.

nikosmoum (approved):

Januson (approved):

PR #3669, opened 2022-11-14
[M] Migrated conditional_content_spec to Java (CANDLEPIN-212)

JoshAlbrecht23:
- created ConditionalContentSpecTest.java
- removed conditional_content_spec.rb


spec-tests/src/test/java/org/candlepin/spec/content/ConditionalContentSpecTest.java
```diff
@@ -0,0 +1,759 @@
...
+            .isNotNull()
+            .singleElement();
+        JsonNode bundledProdCerts = ents.get(0).get("certificates");
+        Long bundledProdSerial = bundledProdCerts.get(0).get("serial").get("serial").asLong();
+
+        Response response = Request.from(consumerClient)
+            .setPath("/consumers/{consumer_uuid}/certificates")
+            .setPathParam("consumer_uuid", consumer.getUuid())
+            .addHeader("accept", "application/json")
+            .execute();
```
JoshAlbrecht23, line 298: I resorted to using the request and response class because I was was running into an issue with `consumerApi.exportCertificates(consumer.getUuid(), null).` returning certificates with incorrect serials. I verified that Candlepin is returning the correct serials with debugging, and some further investigation found that the issue is with the Gson de-serialization of the response in org.candlepin.invoker.client.JSON. 

I made a simple quick test to replicate how the json is being de-serialized with an actual serial that was from a test and I found that it is changing the serial value. 

```
  public static void main(String[] args) {
      JSON json = new JSON();
      Object o = json.deserialize("{\"serial\":1613091661373292420}", Object.class);

      // Output: 1.61309166137329254E18
      System.out.println(o);
  }
```

JoshAlbrecht23: Stepping through the gson de-serialization, it looks like gson is using a double as the default object to number strategy and this is causing the inaccuracies.

https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/Gson.java#L152

spec-tests/src/test/java/org/candlepin/spec/content/ConditionalContentSpecTest.java
```diff
@@ -0,0 +1,759 @@
...
+            .providedProducts(Set.of(dependentProvProd)));
+        adminClient.owners().createPool(ownerKey, Pools.random(dependentProd));
+
+        // Bind to the dependent subscription without being entitled to any of the required products
+        ConsumerDTO consumer = adminClient.consumers().createConsumer(Consumers.random(owner)
+            .facts(Map.of("system.certificate_version", "1.0")));
+        ApiClient consumerClient = ApiClients.ssl(consumer);
+        JsonNode ents = consumerClient.consumers().bindProduct(consumer.getUuid(), dependentProd.getId());
+        String cert = getCertFromEnt(ents.get(0));
+        assertThat(getContentRepoType(cert, conditionalContent1.getId())).isNull();
```
Ceiu, line 225: For one-liners like this, it might make sense to just use the junit base assertions for equality/null checks.

Ceiu (approved):
Looks pretty good. I'll leave it up for a bit if you wanted to make any stylistic changes, but nothing is required.

spec-tests/src/test/java/org/candlepin/spec/content/ConditionalContentSpecTest.java
Ceiu: Wonderful. I imagine our options at configuring that globally are fairly limited as well, since the GSON bits are buried so deep in the generated code.

What you have is fine for the interim, though I would argue adding such details as a comment block for future maintainers.

JoshAlbrecht23: Yeah, that's pretty frustrating that it's in generated code that we can't control easily. I added a comment block!

spec-tests/src/test/java/org/candlepin/spec/content/ConditionalContentSpecTest.java
JoshAlbrecht23: Yeah, that makes sense. Updated throughout

PR #3681, opened 2022-11-16
[4.1] 1931923: API-level product/content updates no longer occur in parallel (CANDLEPIN-440)

Ceiu:
- Changed the ProductManager and ContentManager to obtain pessimistic write locks before making any change or removal of products or content
- Moved several org-less content and product queries from the OwnerContentCurator or OwnerProductCurator to the ContentCurator or ProductCurator as appropriate
- Changed the output from several curator methods from lists of tuples to maps of collections to better convey exactly what was being returned and to make it easier to follow during analysis
- The OrphanCleanupJob no longer removes content that is referenced by a non-orphaned product or an environment, even in cases where the content is technically orphaned (bad content mapping)
- Added a DB-level delete cascade on Content.modifiedProductIds, as JPA is inexplicably unwilling or unable to cascade a deletion on the parent entity to an element collection when using JPA bulk deletions
- Removed some unused curator methods

### DIFF
--- a/src/main/java/org/candlepin/async/tasks/OrphanCleanupJob.java
+++ b/src/main/java/org/candlepin/async/tasks/OrphanCleanupJob.java
@@ -19,21 +19,19 @@ import org.candlepin.async.JobExecutionContext;
 import org.candlepin.async.JobExecutionException;
 import org.candlepin.controller.ContentManager;
 import org.candlepin.controller.ProductManager;
-import org.candlepin.model.CandlepinQuery;
-import org.candlepin.model.Content;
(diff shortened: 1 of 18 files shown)


Ceiu:
This is cherry-picked from the 4.2 version of this change, so reviewers should look for things that may be unintentional changes from merge/cherry-pick artifacts.

Ceiu:
This is blocked pending on details of the deadlock issue discovered in the master version of this patch. Once that update is accepted, this PR will be updated accordingly.

nikosmoum (approved):

PR #3682, opened 2022-11-16
CANDLEPIN-109: Migrate one_sub_pool_per_stack_spec to Java

wottop:


spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,805 @@
...
+        }
+        public ApiClient getHostClient() {
+            return hostClient;
+        }
+        public String getStackId1() {
+            return stackId1;
+        }
+        public String getStackId2() {
+            return stackId2;
+        }
```
JoshAlbrecht23, line 803: Since the member variables have package-private modifiers, you don't need getters for the them. You should be able to just `TestData.owner`

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,805 @@
...
+    private EntitlementDTO bindHostPool(TestData data, List<PoolDTO> pools, String contract) {
+        PoolDTO pool = pools.stream()
+            .filter(x -> contract.equals(x.getContractNumber()))
+            .collect(Collectors.toList()).get(0);
+        return ApiClient.MAPPER.convertValue(
+            data.getHostClient().consumers().bindPool(data.getHost().getUuid(), pool.getId(), 1).get(0),
+            EntitlementDTO.class);
+    }
+
+    private PoolDTO getSubPool(TestData data) {
```
JoshAlbrecht23, line 743: Instead of using TestData as a parameter here, you could use the consume uuid String and make the method more versatile.

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,805 @@
...
+    }
+
+    private String getAttributeValue(String name, List<AttributeDTO> attributes) {
+        List<AttributeDTO> matches = attributes.stream()
+            .filter(x -> name.equals(x.getName()))
+            .collect(Collectors.toList());
+        if (matches.size() > 0) {
+            return matches.get(0).getValue();
+        }
+        return null;
```
JoshAlbrecht23, line 724: Optional: You can use a ternary operator here:
`return matches.size() > 0 : matches.get(0).getValue() ? null; `

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,805 @@
...
+            .multiplier(1L)
+            .attributes(List.of(new AttributeDTO().name("host_limited").value("true"))));
+
+        ProductDTO ownerSku = ownerProductApi.createProductByOwner(owner.getKey(),
+            Products.random()
+            .multiplier(1L)
+            .attributes(List.of(
+                new AttributeDTO().name("host_limited").value("true"),
+                new AttributeDTO().name("stacking_id").value(stackingId),
+                new AttributeDTO().name("virt_limit").value("unlimited"))));
```
JoshAlbrecht23, line 704: We should probably use the ProductAttributes enums here. That way if we were to have to change the key for whatever reason, then this test would receive that change as well:
`ProductAttributes.HostLimited.withValue("true");`

It looks like this will also need to be done in some other methods as well:
createStackedDatacenterProduct
createRegularStackedProduct
createVirtLimitProduct2
createDerivedProduct

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,805 @@
...
+        return data.getHostClient().pools().listPoolsByConsumer(data.getHost().getUuid()).stream()
+            .filter(item -> contracts.contains(item.getContractNumber()))
+            .sorted(Comparator.comparing(item -> contracts.indexOf(item.getContractNumber())))
+            .collect(Collectors.toList());
+    }
+
+    private EntitlementDTO bindHostPool(TestData data, List<PoolDTO> pools, String contract) {
+        PoolDTO pool = pools.stream()
+            .filter(x -> contract.equals(x.getContractNumber()))
+            .collect(Collectors.toList()).get(0);
```
JoshAlbrecht23, line 737: You may get an index out of bounds exception here if the list of pools doesn't have an element that has a contract number that matches the provided contract. You may need to check if the filtered list of pools is empty first and if it is, maybe return null or maybe do an assertion if you expect just one element to always be found.

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,805 @@
...
+        List<PoolDTO> pools = getHostPools(data,
+            List.of(STACKED_VIRT_POOL_1, REGULAR_STACKED_WITH_DIFF_STACK_ID));
+        EntitlementDTO ent1 = bindHostPool(data, pools, STACKED_VIRT_POOL_1);
+        EntitlementDTO ent2 = bindHostPool(data, pools, REGULAR_STACKED_WITH_DIFF_STACK_ID);
+
+        assertThat(data.getGuestClient().pools().listPoolsByConsumer(data.getGuest().getUuid()).stream()
+            .filter(x -> data.getStackId1().equals(x.getSourceStackId()))
+            .collect(Collectors.toList()))
+            .isNotNull()
+            .singleElement()
```
JoshAlbrecht23, line 159: `.singleElement()` also check for null, so you can remove the `.isNotNull()`

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,805 @@
...
+            .collect(Collectors.toList())).hasSize(1);
+        assertThat(createdPools.stream().filter(x -> x.getContractNumber().equals(STACKED_NON_VIRT_POOL))
+            .collect(Collectors.toList())).hasSize(1);
+        assertThat(createdPools.stream().filter(x -> x.getContractNumber().equals(NON_STACKED_POOL))
+            .collect(Collectors.toList())).hasSize(1);
+        assertThat(createdPools.stream().filter(x -> x.getContractNumber().equals(DATACENTER_POOL))
+            .collect(Collectors.toList())).hasSize(1);
+        assertThat(createdPools.stream()
+            .filter(x -> x.getContractNumber().equals(REGULAR_STACKED_WITH_DIFF_STACK_ID))
+            .collect(Collectors.toList())).hasSize(1);
```
JoshAlbrecht23, line 121: You can flip all of these equal checks around to avoid a NPE.
Example: 
```
assertThat(createdPools.stream().filter(x -> STACKED_VIRT_POOL_1.equals(x.getContractNumber()))
            .collect(Collectors.toList())).hasSize(1);
```

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,805 @@
...
+            .createConsumer(Consumers.random(data.getOwner())
+            .guestIds(List.of(new GuestIdDTO().guestId(data.getGuestVirtUuid()))));
+        ApiClient host2Client = ApiClients.ssl(host2);
+        host2Client.consumers().updateConsumer(host2.getUuid(),
+            new ConsumerDTO().guestIds(List.of(new GuestIdDTO().guestId(data.getGuestVirtUuid()))));
+        // Guest entitlement should now be revoked.
+        assertThat(data.getGuestClient().consumers().listEntitlements(data.getGuest().getUuid()))
+            .hasSize(1);
+    }
+
```
JoshAlbrecht23, line 383: Extra new line

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,805 @@
...
+        // quantity should not have changed since there was no entitlement
+        // specifying virt_limit -- use the last instead.
+        assertThat(data.getGuestClient().pools().listPoolsByConsumer(data.getGuest().getUuid())
+            .stream()
+            .filter(x -> data.getStackId1().equals(x.getSourceStackId()))
+            .collect(Collectors.toList()).get(0))
+            .returns(6L, PoolDTO::getQuantity)
+            .returns(pools.get(1).getProductId(), PoolDTO::getProductId);
+    }
+
```
JoshAlbrecht23, line 494: Extra new line

JoshAlbrecht23 (changes requested):

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
wottop: I use more than just the uuid.

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
```diff
@@ -0,0 +1,771 @@
...
+        ContentDTO ownerContent2 = ownerContentApi.createContent(owner.getKey(), Content.random());
+        ProductDTO ownerDerivedEngProduct = ownerProductApi.createProductByOwner(
+            owner.getKey(), Products.random());
+        ownerProductApi.addBatchContent(owner.getKey(), ownerDerivedEngProduct.getId(),
+            Map.of(ownerContent1.getId(), true, ownerContent2.getId(), true));
+
+        ProductDTO ownerDerivedSku = ownerProductApi.createProductByOwner(owner.getKey(),
+            Products.random()
+            .multiplier(1L)
+            .attributes(List.of(new AttributeDTO().name("host_limited").value("true"))));
```
JoshAlbrecht23, line 692: Update to `ProductAttributes.HostLimited.withValue("true")`

spec-tests/src/test/java/org/candlepin/spec/pools/OneSubPoolPerStackSpecTest.java
JoshAlbrecht23: That is my bad, I didn't see that you were also using the stackId1 as well

JoshAlbrecht23 (changes requested):
There was one line that was missed for using the ProductAttributes enum, but then it should be good to go!

JoshAlbrecht23 (approved):

PR #3689, opened 2022-11-18
[4.0] 2142944: API-level product/content updates no longer occur in parallel (CANDLEPIN-440)

Ceiu:
- Changed the ProductManager and ContentManager to obtain pessimistic write locks before making any change or removal of products or content
- Moved several org-less content and product queries from the OwnerContentCurator or OwnerProductCurator to the ContentCurator or ProductCurator as appropriate
- Changed the output from several curator methods from lists of tuples to maps of collections to better convey exactly what was being returned and to make it easier to follow during analysis
- The OrphanCleanupJob no longer removes content that is referenced by a non-orphaned product or an environment, even in cases where the content is technically orphaned (bad content mapping)
- Added a DB-level delete cascade on Content.modifiedProductIds, as JPA is inexplicably unwilling or unable to cascade a deletion on the parent entity to an element collection when using JPA bulk deletions
- Removed some unused curator methods

### DIFF
--- a/server/src/main/java/org/candlepin/async/tasks/OrphanCleanupJob.java
+++ b/server/src/main/java/org/candlepin/async/tasks/OrphanCleanupJob.java
@@ -19,21 +19,19 @@ import org.candlepin.async.JobExecutionContext;
 import org.candlepin.async.JobExecutionException;
 import org.candlepin.controller.ContentManager;
 import org.candlepin.controller.ProductManager;
-import org.candlepin.model.CandlepinQuery;
-import org.candlepin.model.Content;
 import org.candlepin.model.ContentCurator;
-import org.candlepin.model.OwnerContentCurator;
-import org.candlepin.model.OwnerProductCurator;
 import org.candlepin.model.ProductCurator;
 
 import com.google.inject.Inject;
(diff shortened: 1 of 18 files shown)


Ceiu:
This backport was a bit spicy -- some things needed to be manually merged, and while everything passes locally, I worry that the manual merge step has introduced artifacts. Please evaluate this PR with that in mind.

JoshAlbrecht23 (approved):
> This backport was a bit spicy -- some things needed to be manually merged, and while everything passes locally, I worry that the manual merge step has introduced artifacts. Please evaluate this PR with that in mind.

I went through and compared the changes with the commit on master and the only difference I spotted were some formatting changes in OwnerProductResource.java. I don't think it's super critical haha but those are some changes we could add to this PR if you wanted to. 

Master commit:
https://github.com/candlepin/candlepin/commit/e981b712b8b4321944658cb838bc4dc6fedd8977#diff-27c5c1e72361a7ab27a9e8eefa85ddd949dc886a51c8d6e5cf1ebb7e282fcbbf

Ceiu:
Nah, I'm good with leaving stuff like that out in this case. The backport was complicated enough without it.

PR #3690, opened 2022-11-21
Copy the candlepin ansible role into the Candlepin repo

Januson:


Ceiu:
Pre-review comment:

I'm not really that concerned about maintaining the molecule stuff here. It's such a pain to deal with at present with the weird container-in-container junk that it's not worth the dev time or overhead for something we'll be using ourselves daily and we won't be shipping.

I'm fine with keeping the linter bits, but the rest of that can be left behind.

ansible/roles/candlepin/meta/main.yml
```diff
@@ -0,0 +1,41 @@
+galaxy_info:
+  author: jmolet
```
Ceiu, line 2: Should probably update this to one of our usernames, so JC doesn't get hammered with questions about our in-repo role.

If there's a syntax for having multiple maintainers, it'd be nice to keep his name in for legacy purposes so his work is still properly credited, but otherwise we can keep it in a comment for the time being.

ansible/roles/candlepin/tasks/el7/el7_base_deps.yml
```diff
@@ -0,0 +1,93 @@
+---
+- name: "Disable SELinux"
```
Ceiu, line 2: If we're adding in support for el9 in this branch, we can also drop el7 support and not need any specific workarounds for it anymore.

ansible/roles/candlepin/vars/f35.yml
```diff
@@ -0,0 +1,19 @@
...
+  - python3-PyMySQL
+mariadb_config_file: /etc/my.cnf.d/mariadb-server.cnf
+
+postgresql_packages:
+  - postgresql-server
+  - postgresql
+  - postgresql-jdbc
+  - python3-psycopg2
+postgresql_init_user: postgres
+postgresql_init_cmd: pg_ctl initdb -D /var/lib/pgsql/data
```
Ceiu, line 17: The init command should be common between all the versions of postgresql we use from EL8 and beyond. In fact, without el7, the DB setup should be the same for everything.

If we're hitting a situation where we need to var out the entire thing, we're better off leaving the DB setup as individual tasks scoped for the OS.

Ceiu (changes requested):
Looks like the el9 stuff is starting to move in as part of this PR? I was under the impression it was going to be two PRs for that.

In any event, holding off on the remainder of the review until everything is ready to go. In the interim, we should move this back to a draft PR.

.gitignore
```diff
@@ -51,7 +51,6 @@ dataSources/
 **/bin/main/
 **/bin/test/
 **/bin/__init__.py
-ansible/roles
```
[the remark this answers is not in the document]
Januson: ! doesn't seem to work. My ansible is downloading roles to ~/.ansible/roles so we could add config to do that for everyone and only keep our roles in candlepin/ansible.

ansible/roles/candlepin/meta/main.yml
Januson: It might be better to just delete the author completely. Meta is optional and we won't be publishing this role to galaxy.

ansible/roles/candlepin/vars/f35.yml
Januson: I accidentally copied vars files meant for EL9. Removed.

.gitignore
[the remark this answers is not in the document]
Januson: Done.

ansible/roles/candlepin/meta/main.yml
Ceiu: Without the meta/main.yml, do we still have a means of automatically triggering the roles ours depends on? I guess we can always use the role invocation task in our playbooks, but it was convenient having it run as a prereq.

ansible/roles/candlepin/tasks/el7/el7_base_deps.yml
Ceiu: ack

ansible/roles/candlepin/vars/f35.yml
Ceiu: ack

Ceiu (changes requested):
Updates look good -- just need to update the readme to be focused less on describing and running the role as a standalone entity, and more on how the vagrant bits work.

Unsure if it's best for the readme to be split into two: one for describing the role and how its used, and one for vagrant and how to configure things while using it.

ansible/roles/candlepin/tasks/el8/el8_base_deps.yml
```diff
@@ -0,0 +1,113 @@
...
+  tags:
+    - system
+    - system_update
+    - candlepin
+
+- name: "Enable required DNF modules"
+  become: true
+  ansible.builtin.command:
+    cmd: dnf module enable -y pki-core pki-deps
+    warn: false
```
Ceiu, line 26: I just hit this on F37: Leaving the "warn" parameter in place actually breaks this task on newer versions of ansible, so we should probably drop these as we migrate over.

```
fatal: [el8]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (ansible.legacy.command) module: warn. Supported parameters include: stdin_add_newline, _uses_shell, creates, removes, executable, argv, stdin, _raw_params, strip_empty_ends, chdir."}
```

There are a handful of these throughout that need to be trimmed.

ansible/roles/candlepin/meta/main.yml
Januson: Well it is there in this PR. It needs to get replaced by task in the followup PR to get ruby working on EL9. But I agree it would be much better if we could leave it as a prereq.

ansible/roles/candlepin/tasks/el8/el8_base_deps.yml
Januson: Done.

Regarding the readme. I copied the role readme but there's also a playbook readme in ansible folder.

Ceiu (approved):

PR #3698, opened 2022-11-25
CANDLEPIN-209: Enable RPM build on RHEL 9

Januson:
- Update pom generation to generate annotation processor paths
- Update script headers to fix python version ambiguity
- Remove unuser selinux_policyver

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,12 +26,6 @@
       <systemPath>${jssLocation}</systemPath>
     </dependency>
     <dependency>
-      <groupId>org.hibernate</groupId>
-      <artifactId>hibernate-jpamodelgen</artifactId>
-      <version>5.6.12.Final</version>
-      <scope>compile</scope>
-    </dependency>
-    <dependency>
       <groupId>commons-codec</groupId>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -529,6 +523,13 @@
           <debug>true</debug>
           <debuglevel>lines,vars,source</debuglevel>
           <compilerArgument>-proc:none</compilerArgument>
+          <annotationProcessorPaths>
+            <annotationProcessorPath>
+              <groupId>org.hibernate</groupId>
+              <artifactId>hibernate-jpamodelgen</artifactId>
+              <version>5.6.12.Final</version>
+            </annotationProcessorPath>
+          </annotationProcessorPaths>
```
nikosmoum, line 532: Can you give some context around this change? How does it relate to RHEL9 / rpm build?

Januson: Maven build was throwing a duplicate dependency error. Our current setup was generating compile dependencies for annotation processors so I took a look what is a Maven preferred way of defining annotation processors.

PR #3700, opened 2022-11-29
[M] Migrate consumer_checkin_spec to Java (CANDLEPIN-266)

JoshAlbrecht23:
- migrated tests to ConsumerCheckinSpecTest.java
- removed consumer_checkin_spec.rb


spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerCheckinSpecTest.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        return resultData;
+    }
+
+    private JsonNode getAsyncHostGuestMapping(String name, String id, List<String> expectedGuestIds,
+        Map<String, String> facts) {
+        ObjectMapper om = ApiClient.MAPPER;
+        ObjectNode root = om.createObjectNode();
+        ArrayNode hypervisors = root.putArray("hypervisors");
```
Ceiu, line 184: The various Jackson nodes here support chaining, so this could be made a bit easier to read by rearranging certain ops and chaining calls:

```
ArrayNode guestIds = mapper.createArrayNode();
expectedGuestIds.stream()
  .map(gid -> mapper.createObjectNode().put("guestId", gid))
  .forEach(guestIds::add);

ObjectNode hypervisor = mapper.createObjectNode()
  .put("name", name)
  .set("guestIds", guestIds)
  .set("facts", mapper.valueToTree(facts))
  .putObject("hypervisorId") // context change occurs here
  .put("hypervisorId", id);

return mapper.createArrayNode()
  .add(hypervisor);
```

JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerCheckinSpecTest.java
```diff
@@ -0,0 +1,209 @@
...
+        hypervisor.set("facts", factsNode);
+        hypervisors.add(hypervisor);
+
+        return root;
+    }
+
+    private HypervisorUpdateResultDTO getResultData(AsyncJobStatusDTO status) throws IOException {
+        Map resultData = (Map) status.getResultData();
+        Gson gson = new GsonBuilder().create();
+        String json = gson.toJson(resultData);
```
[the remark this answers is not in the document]
JoshAlbrecht23: I'm not sure why I used Gson here, I might have seen this method in another spec test honestly. We should be using Jackson though. Updated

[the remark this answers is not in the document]
JoshAlbrecht23: Looks like I got this method from HypervisorCheckInSpecTest. It looks like there are two issues with deserializing the result data using Jackson. First is that the key and values are not in quotes, and this can be fixed by enabling the 
JsonReadFeature.ALLOW_UNQUOTED_FIELD_NAMES.mappedFeature(), but the second issue is that it uses equal signs instead of colons.

ResultData:
`{created=[{uuid=3b7c9add-66b6-4693-adab-5ccc708252ec, name=name-7BF284DA, owner={id=null, key=test_owner-E3171A71, displayName=null, href=null, contentAccessMode=null}}]`

Ceiu (changes requested):

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerCheckinSpecTest.java
```diff
@@ -0,0 +1,213 @@
...
+            .forEach(guestIds::add);
+
+        ObjectNode root = om.createObjectNode();
+        ArrayNode hypervisors = root.putArray("hypervisors");
+        ObjectNode hypervisor = om.createObjectNode()
+            .put("name", name)
+            .set("hypervisorId", om.createObjectNode()
+            .put("hypervisorId", id));
+
+        hypervisor.set("guestIds", guestIds);
```
Ceiu, line 195: Looks like this got partially reverted...?

The object->array for the root can also be chained. The "special" `putArray` and `putObject` function by adding the property and the switching to the new container, so the final wrapper of wrappers can be created in-line:

```
return mapper.createObjectNode()
  .putArray("hypervisors")
  .add(hypervisor);
```

When it comes to building objects like this, it's better for maintenance to ensure the order is as linear as possible; either building it in full reverse order, or top-down. The more it jumps around, the harder it is to visualize the structure as it's being constructed.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerCheckinSpecTest.java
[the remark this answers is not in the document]
JoshAlbrecht23: I apologize, I was passing in the `.toString()` value into the convertValue which didn't allow the de-serialization based on that String. 

` ApiClient.MAPPER.convertValue(status.getResultData().toString(), HypervisorUpdateResultDTO.class);`

Passing in the Object allows the `convertValue` to deserialize properly. Stupid mistake on my part *sigh* 

` ApiClient.MAPPER.convertValue(status.getResultData(), HypervisorUpdateResultDTO.class);`

I also removed the use of Gson in HypervisorCheckInSpecTest as well

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerCheckinSpecTest.java
JoshAlbrecht23: That makes sense! I did have to make a couple changes. I couldn't chain the `.set()` calls when building the hypervisor ObjectNode as the `.set()` method returns a JsonNode and not an ObjectNode which is strange, and a quick search online shows others aren't too happy about this as well. 

Also, I used `.putPOJO("hypervisors", mapper.createArrayNode().add(hypervisor))` because it returns the outer wrapper object. The example you gave returns the ArrayNode instead of the Object Node and you get something like: 
`[{"name":"name-35008A87","hypervisorId":{"hypervisorId":"id-AF9BA895"},"guestIds":[{"guestId":"guest-1"},{"guestId":"guest-2"},{"guestId":"guest-3"}],"facts":{"test_fact":"fact_value"}}]`

Ceiu: > That makes sense! I did have to make a couple changes. I couldn't chain the `.set()` calls when building the hypervisor ObjectNode as the `.set()` method returns a JsonNode and not an ObjectNode which is strange, and a quick search online shows others aren't too happy about this as well.

Hrmm... that's pretty odd. Especially since ObjectNode is what defines `.set` to begin with.

I'm beginning to wonder if building it with Jackson objects is the way to go here, then. We could just as easily build it up with a bunch of maps and lists, and then throw that at Jackson:

```
  List<Map<String, String>> guestIds = expectedGuestIds.stream()
      .map(gid -> Map.of("guestId", gid))

  Map<String, Object> hypervisor = Map.of(
      "name", name,
      "hypervisorId", Map.of("hypervisorId", id),
      "guestIds", guestIds,
      "facts", facts);

  Object node = Map.of("hypervisors", List.of(hypervisor));

  return mapper.valueToTree(node);
```


> The example you gave returns the ArrayNode instead of the Object Node and you get something like:

Oops, good catch.

In this case, however, since the output specifies a JsonNode, the detrimental behavior of `.set` isn't an issue as we no longer need the ObjectNode specification.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerCheckinSpecTest.java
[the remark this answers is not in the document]
Ceiu: Interesting. Given that it's a string already, I wouldn't have expected that to make a difference.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerCheckinSpecTest.java
JoshAlbrecht23: +1 to building with maps and lists! I like that so much more and is easy to read as well. Updated

Ceiu (approved):

PR #3707, opened 2022-12-06
CANDLEPIN-176: Migrate autobind_spec to Java, parts 1 and 2

wottop:
Also includes CANDLEPIN-137

Got on a roll and was able to do the entire file under the 8 points for part 1.


spec-tests/src/test/java/org/candlepin/spec/entitlements/AutobindSpecTest.java
```diff
@@ -0,0 +1,1120 @@
...
+        ApiClient userClient = ApiClients.basic(UserUtil.createUser(client, owner));
+        ProductDTO product1 = ownerProductApi.createProductByOwner(owner.getKey(),
+            Products.randomEng()
+            .attributes(List.of(
+            ProductAttributes.Addons.withValue("addon1"))));
+        PoolDTO pool1 = ownerApi.createPool(owner.getKey(), Pools.random(product1));
+        ProductDTO product2 = ownerProductApi.createProductByOwner(owner.getKey(), Products.randomEng());
+        PoolDTO pool2 = ownerApi.createPool(owner.getKey(), Pools.random(product2));
+
+        ConsumerDTO consumer = userClient.consumers().createConsumer(
```
Ceiu, line 164: This block, too. There are three operations going on here, but the indentation implies only one.

Ceiu (changes requested):
The code itself seems fine enough, but there are a lot of stylistic things here that make it very difficult to read without knowledge of the operations here and specific IDE code highlighting. Better usage of whitespace, and decoupling long method chains into the separate operations would go a long way toward helping consistency and readability.

wottop:
Also cleaned up a similar issue with ".installedProducts"

PR #3709, opened 2022-12-08
Updated Vagrant networking configuration

wottop:
- Updated the networking configuration in the Vagrantfile
  to use the correct attributes when running on newer
  versions of Vagrant


Ceiu:
Change is fine with me, but can we update the commit to be a bit more clear? Something to the effect of:

```
Updated Vagrant networking configuration

- Updated the networking configuration in the Vagrantfile
  to use the correct attributes when running on newer
  versions of Vagrant
```

Ceiu (approved):

PR #3710, opened 2022-12-08
Prevent layer 2 cache eviction & improve caching of branding

barnabycourt:
* Set the NATIVE_SPACES query hint on native queries in order to prevent hibernate from unnecessarily clearing caches of immutable objects. If a hint is not set then all cache regions are cleared during native query execution.

* Set the appropriate cache hints on Branding objects so that they are cached as immutable.

* Update the cache definition to standard 24 hours instead of using seconds in order to make it more clear for humans reading the file in the future.

* Add a flush to the OwnerProductCurator in order to ensure all products exist in the database before updating references.


src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -853,6 +855,10 @@ public void updateOwnerProductReferences(Owner owner, Map<String, String> produc
             return;
         }
 
+        // Flush because any newly created products MUST be sent to the DB before the next calls
+        // as newly created products
+        this.getEntityManager().flush();
```
Ceiu, line 860: Couple things here:
First, we can call `this.flush()` instead of fetching the entity manager ourselves.

When it was first brought up that we needed to flush, I assumed that was meaning *after* the operations in this method. If the flush needs to occur *before*, that's still fine, but it means we can actually drop a flush in the `ProductNodeVisitor` (and `ContentNodeVisitor` right before this is called.

src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -951,7 +959,8 @@ private void updateOwnerProductActivationKeys(Owner owner, Map<String, String> p
             sql = "UPDATE cp2_activation_key_products SET product_uuid = :updated " +
                 "WHERE key_id = :key_id AND product_uuid = :current";
 
-            query = entityManager.createNativeQuery(sql);
+            query = entityManager.createNativeQuery(sql)
```
Ceiu, line 962: The `updateOwnerProductActivationKeys` method is dead and uncalled. We can just remove it entirely rather than update it.

src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -1119,7 +1128,8 @@ public void rebuildOwnerProductMapping(Owner owner, Map<String, String> productI
 
             Query query = this.getEntityManager()
                 .createNativeQuery(sql)
-                .setParameter("owner_id", owner.getId());
+                .setParameter("owner_id", owner.getId())
+                .setHint(QueryHints.NATIVE_SPACES, EnvironmentContent.class.getName());
```
Ceiu, line 1132: If I'm understanding the NATIVE_SPACES stuff correctly, this should be using the `OwnerProduct` class rather than `EnvironmentContent`.

Ceiu (changes requested):
Few changes needed, but do we also want to hit the content side of things? `ContentNodeVisitor` and `OwnerContentCurator` use a very similar flow and, if my understanding is correct, will also end up blasting the cache in the same way without these additions.

src/main/java/org/candlepin/model/OwnerProductCurator.java
barnabycourt: Good call, also, I can't find anybody calling rebuildOwnerProductMapping(). What would you think about removing the method entirely?

src/main/java/org/candlepin/model/OwnerProductCurator.java
barnabycourt: :+1:

src/main/java/org/candlepin/model/OwnerProductCurator.java
barnabycourt: Hmm... it's fair, I like the flush being handled outside of this as the visitors can call flush once before iterating over the maps. One concern with keeping the flush here would be that it is going to be called many times in a row by the ProductNodeVisitor.complete() . It's not horrible as a bunch of the flush calls should no-op but it's an odd contention in how the method is used and where we want to be calling flush from. Seems like either we should remove the flush from product node visitor and keep it here. Or, remove the flush from here and add it to ProductManager.updateProduct() so that both places that call updateOwnerProductReferences() have flushed beforehand. Thoughts?

src/main/java/org/candlepin/model/OwnerProductCurator.java
barnabycourt: keeping per separate conversation, and yes, that should be OwnerProduct instead of EnvironmentContent.

src/main/java/org/candlepin/model/OwnerProductCurator.java
Ceiu: I'm leaning toward adding it to the places in the managers where we call it directly, but we would still want to throw that note in the documentation on the `updateOwner-entity-References` so it's not a gotcha months from now.

As a note, `ProductNodeVisitor.complete` is called once per refresh, so we shouldn't get a ton of repeat flushes no matter where they end up. The confusing bit might be the loop over the map, but that should be a single-entry map in every case I can think of.

Ceiu (approved):

PR #3712, opened 2022-12-12
CANDLEPIN-292: Introduce Centos Stream 9 vagrant setup

Januson:
- Add EL9 support
- Drop EL7 support
- Drop Fedora 35 support


ansible/roles/candlepin/tasks/main.yml
```diff
@@ -5,32 +5,33 @@
...
+    file: el8/el8_tasks.yml
   when:
-    - ansible_facts['distribution'] in ["CentOS", "RedHat"]
-    - ansible_facts['distribution_major_version'] == "7"
+    - ansible_distribution in ["CentOS", "RedHat"]
+    - ansible_distribution_major_version == "8"
   tags:
     - always
 
 - name: "Include EL8 tasks"
```
Ceiu, line 17: EL9 task still labeled as EL8

ansible/roles/candlepin/tasks/common/cp_mariadb.yml
```diff
@@ -88,6 +91,20 @@
     name: candlepin
     priv: "candlepin.*:ALL,GRANT"
     state: present
+  when:
```
Ceiu, line 94: If we're stuck doing part of these DB changes differently based on the OS, I think I'd rather leave the whole thing in each directory as it was -- especially if we're going to be staying at only 2 or 3 different machines.

Ceiu (changes requested):
Couple very minor changes, but otherwise looks good.

Unfortunately, my hope for the DB changes was not as clean with the removal of the other VMs (specifically EL7) -- I was hoping that would just be identical going forward. But if we're stuck having to still look at the underlying OS and branch behavior, we're still better off in terms of readability leaving the tasks in their OS-specific variants.

ansible/roles/candlepin/tasks/main.yml
Januson: Done.

ansible/roles/candlepin/tasks/common/cp_mariadb.yml
Januson: Done.

Ceiu (approved):

PR #3717, opened 2022-12-15
[M] Added remote debug configurations for Gradle

JoshAlbrecht23:
- Added a block in the cp_debugging task that forwards port 8001
- Added a check for the 'remote-debug' system property in the spec gradle task

Running `vagrant up` with ansible.roles.candlepin.defaults.cp_configure_debugging set to 'true' will now forward the port 8001 that can be attached to for remote debugging Java spec tests. I also added a check for the system property 'remote-debug' in the 'spec' gradle task that will enable the remote debugging.

I figured that we should use a different port than what is already being forwarded for Candlepin debugging (8000) so you don't have to remove the debug arguments from Tomcat and restart the service so that port 8000 is open before trying to debug a spec test.

Example of running a spec test for remote debugging:
`./gradlew clean spec -Dremote-debug=true --tests OwnerResourceSpecTest.shouldCreateScaOwner`

### DIFF
--- a/ansible/roles/candlepin/tasks/common/cp_debugging.yml
+++ b/ansible/roles/candlepin/tasks/common/cp_debugging.yml
@@ -15,13 +15,16 @@
         - candlepin
         - debugging
 
-    - name: "Enable Tomcat debugging port in firewalld"
+    - name: "Enable debugging ports in firewalld"
       become: true
       ansible.posix.firewalld:
-        port: "8000/tcp"
+        port: "{{ item }}/tcp"
         state: enabled
         permanent: true
         offline: false
+      with_items:
+        - 8000 # Tomcat remote debugging port
+        - 5005 # Gradle remote debugging port
       when:
         - "'services' in ansible_facts"
         - "'firewalld.service' in ansible_facts.services"
--- a/build.gradle
+++ b/build.gradle
@@ -307,6 +307,10 @@ msgfmt {
 }
 
 test {
+    debugOptions {
+        host = '*'
+    }
+
     useJUnitPlatform ()
 
     // Sometimes causes out of memory on vagrant
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-all.zip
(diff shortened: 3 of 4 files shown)


Ceiu:
nack on this.

While cracking the port might be apt for "remote" debugging from the host system, the gradle bits can be better achieved with the native `--debug-jvm` option within gradle itself. We can also use the gradle config to set the port or debug defaults (suspend on start might be apt) rather than trying to do that ourselves.

See: https://docs.gradle.org/current/userguide/java_testing.html#sec:debugging_java_tests

ansible/roles/candlepin/tasks/common/cp_debugging.yml
```diff
@@ -18,10 +18,13 @@
     - name: "Enable Tomcat debugging port in firewalld"
```
Ceiu, line 18: Let's update the task name to something more general, like "Enable debugging ports in firewalld"

ansible/roles/candlepin/tasks/common/cp_debugging.yml
```diff
@@ -18,10 +18,13 @@
...
       become: true
       ansible.posix.firewalld:
-        port: "8000/tcp"
+        port: "{{ item }}/tcp"
         state: enabled
         permanent: true
         offline: false
+      with_items:
+        - 8000
```
Ceiu, line 26: Let's add some comments on these so future maintainers don't need to look up port numbers to find it. We can use the same format we do in the base_deps task:

```
with_items:
  - 8000     # Tomcat remote debugging port
  - 5005     # Gradle remote debugging port
```

Ceiu (changes requested):
Let's also add the debugOptions to the main build.gradle so we can debug unit tests the same way.

ansible/roles/candlepin/tasks/common/cp_debugging.yml
JoshAlbrecht23: Good catch, updated

ansible/roles/candlepin/tasks/common/cp_debugging.yml
JoshAlbrecht23: Yeah, that is a good idea! Updated

Ceiu (changes requested):
Looks good. Just need to update the commit title now to reflect that it's adding remote debugging configuration for Gradle as a whole, rather than just the Java spec tests.

Ceiu (approved):

PR #3719, opened 2022-12-20
CANDLEPIN-230: Migrate environment_spec to Java

Januson:
- Ported first part of environment_spec
- Pluralized Content and Environment builders to distinquish them from ordinary classes
- Added custom assertion for verifying certificates

### DIFF
--- a/spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Contents.java
+++ b/spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Contents.java
@@ -27,7 +27,7 @@ import java.util.stream.Collectors;
 /**
  * Class providing factory functions for BrandingDTO instances.
  */
-public final class Content {
+public final class Contents {
 
     /**
      * Throws an UnsupportedOperationException; individual instantiation of this class is not
@@ -35,7 +35,7 @@ public final class Content {
      *
      * @throws UnsupportedOperationException
      */
-    private Content() {
+    private Contents() {
         throw new UnsupportedOperationException();
     }
 
@@ -78,30 +78,15 @@ public final class Content {
      *  a ContentDTO instance with the required fields populated with semi-random values
      */
     public static ContentDTO random() {
-        return random("test_content");
-    }
-
-    /**
-     * Builds a ContentDTO instance with the required fields populated with semi-random values,
-     * using the provided prefix for the ID field.
-     *
-     * @param idPrefix
-     *  a prefix string to use for the ID field
-     *
-     * @return
-     *  a ContentDTO instance with the required fields populated with semi-random values, using the
-     *  provided prefix for the ID field.
-     */
-    public static ContentDTO random(String idPrefix) {
-        String cid = StringUtil.random(8, StringUtil.CHARSET_NUMERIC_HEX);
+        String cid = StringUtil.random(18, StringUtil.CHARSET_NUMERIC);
 
         return new ContentDTO()
-            .id(idPrefix + '-' + cid)
-            .name("test content " + cid)
+            .id(cid)
+            .name("test_content" + cid)
             .label("test label " + cid)
-            .type("test")
+            .type("yum")
             .vendor("test vendor")
-            .contentUrl("/url " + cid);
+            .contentUrl("/url_" + cid);
 
         // perhaps add optional fields here?
     }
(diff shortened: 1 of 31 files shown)


spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Contents.java
```diff
@@ -27,15 +27,15 @@
 /**
  * Class providing factory functions for BrandingDTO instances.
  */
-public final class Content {
+public final class Contents {
```
Ceiu, line 30: If the goal here is to differentiate it from the constructed instances, we're better off calling these what they are: builders or factories. ContentBuilder, OwnerBuilder, etc..

Specifically for this class, "content" in English is an unnecessarily complex word when it comes to pluralization, and modern usage is narrowing on just using the unsuffixed variant for both singular and plural usage; since the strict guidelines for which to use are pretty subjective (whether or not the contents are countable, individual things, and what constitutes individuality or whether or not something is countable is highly context-sensitive).

As a native speaker who tries to follow proper grammar rules when it comes to selecting terminology or writing documentation, "contents," for our purposes, almost always feels clunky and reads awkwardly. Probably why modern usage is gradually deviating from the countable rule as more and more content moves to digital mediums.

tl;dr: English is a silly language.

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -354,7 +354,7 @@ public AsyncJobStatusDTO demoteContent(@Verify(Environment.class) String envId,
     @SecurityHole(activationKey = true)
     @Transactional
     public ConsumerDTO createConsumerInEnvironment(String envId, ConsumerDTO consumer,
-        String userName, String activationKeys) throws BadRequestException {
+        String userName, String activationKeys, String owner) throws BadRequestException {
```
Ceiu, line 357: The new input doesn't appear to be used anywhere outside of one test. What is the intent for this?

Ceiu (changes requested):
First pass through this is pretty good, but I had a few questions on the API change, and a builder change specifically.

This probably deserves another pass on a day that's not the first back from a long vacation.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Contents.java
```diff
@@ -96,12 +96,12 @@ public static ContentDTO random(String idPrefix) {
         String cid = StringUtil.random(8, StringUtil.CHARSET_NUMERIC_HEX);
 
         return new ContentDTO()
-            .id(idPrefix + '-' + cid)
+            .id(StringUtil.random(8, StringUtil.CHARSET_NUMERIC))
```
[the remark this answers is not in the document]
Januson: Apparently content id has to be numeric or else we fail to create it's OID. This happens only for v1 certs, but I thought we should probably switch them all to numeric. Also the prefix is never used. There was one test calling it, but it just replicated what parameterless version was doing.

The hard restriction might be a good idea. We should investigate if we can do it without breaking anyone.

src/main/java/org/candlepin/resource/EnvironmentResource.java
Januson: I wasn't sure about this one. Just using our Request/Response would have been enough, but the owner here is necessary for activation key auth so I thought it might be better to have it in the API explicitly.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Contents.java
Ceiu: Fair enough.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Contents.java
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/resource/EnvironmentResource.java
Ceiu: Looking at this: https://spec.openapis.org/oas/latest.html#securitySchemeObject

Ceiu (approved):

PR #3720, opened 2022-12-20
CANDLEPIN-126: Migrate consumer_bonus_pool_provided_spec and consumer_facts_spec to Java

Januson:


spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerFactsSpecTest.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    @SuppressWarnings("indentation")
+    private ConsumerDTO createConsumer() {
+        return ownerClient.consumers()
+            .createConsumer(Consumers.random(this.owner)
+                .putFactsItem(Facts.UnameMachine.key(), "i686")
+                .putFactsItem(Facts.UnameSystem.key(), "Linux")
+            )
+            .addInstalledProductsItem(
```
Ceiu, line 189: This bit should probably be included in the part that's sent up to the server, so this util method isn't returning an already-modified consumer DTO.

In general it's kind of rough to read due to the indentation+chaining shenanigans.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerFactsSpecTest.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    @Test
+    public void shouldUpdateConsumerUpdatedDateWhenFactsAreUpdated() {
+        ConsumerDTO consumer = createConsumer();
+        ApiClient consumerClient = ApiClients.ssl(consumer);
+        OffsetDateTime initialDate = consumer.getUpdated();
+
+        // MySQL drops millis, we need to wait a bit longer
+        sleepForMs(1000);
```
Ceiu, line 106: Could probably just throw the InterruptedException from here instead of abstracting the whole thing out

Ceiu (changes requested):

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerFactsSpecTest.java
Januson: Good catch. I shuffled the chain a bit to remove the indentation.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerFactsSpecTest.java
Januson: Done.

Ceiu (approved):

PR #3721, opened 2022-12-21
[M] CANDLEPIN-151: Introduce Centos Stream 9 container image

bakajstep:


docker/candlepin-base-cs9/setup-devel-env.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/sh
+#
+# Sets a system up for a candlepin development environment (minus a db,
+# handled separately), and an initial clone of candlepin.
+
+set -ve
+
+source /root/dockerlib.sh
+
+export JAVA_VERSION=1.8.0
```
nikosmoum, line 10: Set this to 11. We don't have any supported branches that require 1.8 any more.

docker/candlepin-base-cs9/setup-devel-env.sh
```diff
@@ -0,0 +1,88 @@
...
+dnf install -y epel-release
+
+PACKAGES=(
+    createrepo_c
+    expect
+    gettext
+    git
+    hostname
+    java-11-openjdk-devel
+    java-$JAVA_VERSION-openjdk-devel
```
nikosmoum, line 23: Remove one of these. We no longer have a need for java 1.8, only 11

nikosmoum (changes requested):

docker/candlepin-base-cs9/setup-devel-env.sh
bakajstep: done

docker/candlepin-base-cs9/setup-devel-env.sh
bakajstep: done

docker/test
```diff
@@ -75,29 +74,51 @@ then
...
-if [ "$DETACHED" == "1" ]; then
+if [ "$INIT_CONTAINER" == "1" ]; then # For CS9
+  podman-compose $PROJ_NAME stop
+  podman-compose $PROJ_NAME rm -f
+
+  if [ "$USE_CACHE" != "1" ]; then
+    podman-compose $COMPOSE_ARGS pull
+  fi
+  podman-compose $PROJ_NAME $COMPOSE_ARGS up -d
+  sleep # Waiting to start systemd
```
nikosmoum, line 99: how does this work? doesn't this need an argument for duration? (e.g. 2 seconds)

docker/test
```diff
@@ -75,29 +74,51 @@ then
...
+  podman-compose $PROJ_NAME $COMPOSE_ARGS exec candlepin $TEST_CMD
```
nikosmoum, line 100: if DETACHED is set, then the `exec` needs to run with `-d`

docker/test
nikosmoum: Also, `wait <job_id>` may be a better option than `sleep`, since it will wait exactly until `podman-compose up` is done, instead of having to guess how long it may take

docker/test
nikosmoum: I just tried this, and it doesn't seem to work. podman still runs the container in the foreground

nikosmoum: Let's use podman direct commands (e.g. `podman pod create`, `podman start/exec`)

docker/candlepin-base-cs9/setup-devel-env.sh
```diff
@@ -0,0 +1,87 @@
...
+source /etc/profile.d/rvm.sh || true
+
+rvm pkg install openssl
+rvm install ruby-2.5.3 --with-openssl-dir=/usr/local/rvm/usr
+rvm use --default 2.5.3
+set -v
+
+# Install all ruby deps
+gem install bundler -v 1.16.1
+bundle install --without=proton --retry=5 --verbose --full-index
```
nikosmoum, line 77: Lets remove all this. We no longer need ruby

bakajstep: done

docker/test

docker/test
ptoscano: TL;DR: systemd takes some time to start :)

Few months ago, I updated the deployment of Candlepin in the 'services' image of the cockpit CI, and the deployment is done as container in a Fedora CoreOS distro. When you start the new container using (directly or indirectly) `podman run`, the `podman` command returns when the container is started; however, the entry point (systemd in this case) may take a while to start and be fully operational, so doing
```
$ podman run ...
$ podman exec systemctl
```
may fail because systemd is not ready yet. If the host that runs the container is not a baremetal system (e.g. a VM), the systemd startup may take longer. Hence, sleeping few seconds (I chose 5 for cockpit, and so far has been working fine) is "good enough" to ensure systemd is operational and responsive. Yes, it feels a bit of a hack, however not that bad of a hack.

`podman wait` is not actually useful here, since it waits for the container to stop, which in practice it means when systemd is stopped (and thus not what you want).

Reference: https://github.com/cockpit-project/bots/pull/3981

bakajstep:
Uvedomil jsem si ze pro odstraneni dockeru uplne bude potreba jeste vyresit vyber databaze, protoze pro postgrea a pro mariadb bude vytvoreni kontejneru trochu jine. Mam to udelat v tomhle ukolu nebo vytvorit pro nej uplne novej?

nikosmoum:
@bakajstep Let's do this as part of this task and get rid of docker completely

docker/test
```diff
@@ -15,41 +17,45 @@ OPTIONS:
...
        sudo chown 999:999 $DIR/mysql.cnf;;
-    n) PROJ_NAME="-p $OPTARG";;
+    n) PROJ_NAME=$OPTARG;;
     o) OPERATING_SYSTEM=$OPTARG;;
-    p) COMPOSE_ARGS="-f $DIR/docker-compose-postgres";;
-    x) COMPOSE_ARGS="-f $DIR/docker-compose-no-db";;
-    q) COMPOSE_ARGS="-f $DIR/docker-compose-sonarqube";;
+    p) DB="postgres:12";;
+    x) DB="";;
+    q) SONARQUBE="1";;
```
nikosmoum, line 58: We don't need options x and q in master branch any more (removed by Ondrej's PR)

nikosmoum: (but cp-test needs then for old branches to still function)

bakajstep: i see i am just rebasing it

bakajstep:
the migration to podman with spinning containers with commands bring feature that we can choose OS even with mysql

docker/candlepin-base-cs8/Dockerfile
```diff
@@ -24,5 +24,4 @@ RUN echo 'alias python="python3"' >> ~/.bashrc
 
 EXPOSE 8443 22
 
-#CMD ["/usr/bin/cp-test", "-t", "-u", "-r"]
-CMD ["true"]
\ No newline at end of file
+ENTRYPOINT ["/usr/sbin/init"]
```
nikosmoum, line 27: Is this change part of this task? if the cs8 container still uses supervisord then will this not break it?

bakajstep: because how the containers are started i need to be still running after podma start command and this is doing that and now i would still add that because of the task migrate cs8 to systemd

docker/candlepin-base-cs9/setup-devel-env.sh
```diff
@@ -0,0 +1,69 @@
...
+    python3-requests
+    python3-libxml2
+    python-unversioned-command
+    rpm-build
+    rpm-sign
+    rsyslog
+    wget
+    which
+    zlib
+    zlib-devel
```
nikosmoum, line 40: I think these 2 were only dependencies for ruby/rspec. Can you test if everything works if you remove them from the image?

docker/test
```diff
@@ -59,38 +59,92 @@ while getopts ":c:mn:o:plde" opt; do
...
+podman pod start $PROJ_NAME
+sleep 5 # Waiting to start systemd
+
 if [ "$DETACHED" == "1" ]; then
     # Run detached, and don't remove the containers
-    docker-compose $PROJ_NAME $COMPOSE_ARGS run -d $SERVICE_PORTS_ARG candlepin $TEST_CMD
+    podman exec -d $PROJ_NAME-candlepin $TEST_CMD
     RETVAL=$?
-else
-    docker-compose $PROJ_NAME $COMPOSE_ARGS run --rm $SERVICE_PORTS_ARG candlepin $TEST_CMD
```
nikosmoum, line 91: since we don't use docker-compose*.yml files anymore, we should remove the `docker/switch-defaults` file (it is useless)

nikosmoum (changes requested):
A few more changes needed, but this is a great long-awaited change

docker/README.mkd
```diff
@@ -19,6 +19,8 @@ Basic invocations of `build-images` are as follows:
 # build only one image
 # see the top level of docker-compose-build.yml under 'services:' for a list of image names
 ./build-images <image_name>
+# build only one image and push to a registry
+./build-images -p <image_name>
```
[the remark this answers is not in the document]
bakajstep: done

docker/test
bakajstep: done

docker/candlepin-base-cs9/setup-devel-env.sh
bakajstep: I tried it both with postgres and mysql and it works in both cases without the two packages

nikosmoum:
The code looks good; the only thing left to do is move the CS8 image from supervisord to systemd

nikosmoum:
This needs the updated CS8 image pushed to quay to pass jenkins, but that will only happen once the candlepin-4.2-HOTFIX backport pr is ready. Then both PRs can be merged

nikosmoum (approved):

PR #3723, opened 2022-12-23
CANDLEPIN-152: Migrate compliance_reasons_spec to Java

Januson:


src/main/java/org/candlepin/resource/util/ResourceDateParser.java
```diff
@@ -68,7 +68,7 @@ public static Date parseDateString(String activeOn) {
             d = DatatypeConverter.parseDateTime(activeOn).getTime();
         }
         catch (IllegalArgumentException e) {
-            throw new BadRequestException("Invalid date, must use ISO 8601 format", e);
+            throw new BadRequestException("Invalid date, \"" + activeOn + "\" must use ISO 8601 format", e);
```
Ceiu, line 71: :+1:

Side note, though: if this string is being returned to the user, we should probably be translating it.

Ceiu (approved):
General PR looks good, just one ancillary thing to address, maybe.

src/main/java/org/candlepin/resource/util/ResourceDateParser.java
Januson: Done.

PR #3739, opened 2023-01-10
CANDLEPIN-183: Migrate distributor_capability_spec to Java

wottop:


spec-tests/src/test/java/org/candlepin/spec/DistributorCapabilitySpecTest.java
```diff
@@ -0,0 +1,309 @@
...
+        OwnerDTO owner = ownerClient.createOwner(Owners.random());
+        String name = StringUtil.random("name");
+        DistributorVersionDTO distVersion = new DistributorVersionDTO()
+            .name(name)
+            .displayName(StringUtil.random("displayName"))
+            .addCapabilitiesItem(new DistributorVersionCapabilityDTO().name("midas touch"))
+            .addCapabilitiesItem(new DistributorVersionCapabilityDTO().name("telepathy"))
+            .addCapabilitiesItem(new DistributorVersionCapabilityDTO().name("lightning speed"));
+        client.distributorVersions().create(distVersion);
+        Map facts = Map.of("distributor_version", name);
```
JoshAlbrecht23, line 119: Optional: You can move this into the .fact() method so that it is all in one line since the facts variable is only being used in that one spot:
`.facts(Map.of("distributor_version", name))`. This could also be done in some other spots in this test file as well.

JoshAlbrecht23 (approved):
Looks good, I left an optional comment if you wanted to make an update.

PR #3742, opened 2023-01-11
Limit bundler version to 2.3.26

Januson:
- Bundler 2.4 dropped support of our ruby version. Bundler 2.3.26 is the last version that supports it.

### DIFF
--- a/ansible/roles/candlepin/tasks/el9/el9_ruby.yml
+++ b/ansible/roles/candlepin/tasks/el9/el9_ruby.yml
@@ -43,6 +43,6 @@
     - ruby
 
 - name: Ensure bundler is installed
-  ansible.builtin.command: "bash -lc 'gem install bundler'"
+  ansible.builtin.command: "bash -lc 'gem install bundler -v 2.3.26'"
   tags:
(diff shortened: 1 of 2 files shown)


ansible/roles/candlepin/tasks/el9/el9_ruby.yml
```diff
@@ -43,6 +43,6 @@
     - ruby
 
 - name: Ensure bundler is installed
-  ansible.builtin.command: "bash -lc 'gem install bundler'"
+  ansible.builtin.command: "bash -lc 'gem install bundler -v 2.3.26'"
```
nikosmoum, line 46: Is this an issue just in rhel9, or does it also need to be set in other OSes?

nikosmoum (approved):

ansible/roles/candlepin/tasks/el9/el9_ruby.yml
Januson: Good question. RVM role is not smart enough to deal with this so I updated EL8 as well.

PR #3754, opened 2023-01-20
CANDLEPIN-70: Migrate consumer_resource_spec to Java

wottop:
Divided into multiple files:
ConsumerResourceSpecTest
ConsumerResourceContentAccessSpecTest
ConsumerResourceEntitlementSpecTest
ConsumerResourceEnvironmentSpecTest
ConsumerResourceHypervisorIdSpecTest
ConsumerResourceSystemPurposeSpecTest

This is part 2 and 3.


spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceEntitlementSpecTest.java
```diff
@@ -0,0 +1,771 @@
...
+            .installedProducts(installed);
+        consumer = userClient.consumers().createConsumer(consumer);
+        ApiClient consumerClient = ApiClients.ssl(consumer);
+        consumer = consumerClient.consumers().getConsumer(consumer.getUuid());
+        assertThat(consumer).returns(serviceLevel1, ConsumerDTO::getServiceLevel);
+
+        // dry run against the set service level:
+        // should return both pools because we no longer filter on consumer's sla
+        // (unless the consumer has existing entitlements, and in this case we don't).
+        List<PoolQuantityDTO> pools = adminClient.consumers().dryBind(consumer.getUuid(), null);
```
Ceiu, line 424: Do we have any guarantee that the pools will be returned in this specific order every time? If not, the assertions following can use `.map` to map the pools collection to the IDs, then `.containsExactlyInAnyOrder` to specify the list of expected pool IDs:

```
assertThat(pools)
  .hasSize(2)
  .map(PoolQuantityDTO::getPool)
  .map(PoolDTO::getId)
  .containsExactlyInAnyOrder(pool1.getId(), pool2.getId());
```

Looks like this kinda thing potentially occurs in many places. I'll omit explicit comments on each one.

Ceiu (changes requested):
Most of this looks pretty good. Just a few concerns over the tests which target pools in a way that looks like output order needs to be guaranteed for test stability.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceEntitlementSpecTest.java
wottop: From the previous tests, I would say that the proper order is expected.
I will make the change to the contains exactly.

Ceiu (approved):

PR #3756, opened 2023-01-23
CANDLEPIN-151: Introduce Centos Stream 9 container image

bakajstep:
- Create Dockerfile for Centos stream 9 with candlepin
- Update test script that you can choose on which os will be running postgres test in Jenkins


Ceiu:
Is #3721 obsoleted by this PR?

bakajstep:
> Is #3721 obsoleted by this PR?

I was just trying if i broke cs8 container. But I think not, because Pino having same issue in his PR.

PR #3757, opened 2023-01-23
test data importers: use less shell

ptoscano:
Do not use a shell where there is no need for a shell: this is basically the case for the majority of the commands run. For this:
- change the non-shell command invocations to be a list of parameters, disabling the shell in that case
- disable all the manual quoting of parameters for non-shell command

This gives a noticeable speed improvement, as there are way less shell invocations.

### DIFF
--- a/bin/deployment/create_test_repos.py
+++ b/bin/deployment/create_test_repos.py
@@ -117,7 +117,8 @@ def run_command(command, verbose=False):
     """
 
     # Run command in subprocess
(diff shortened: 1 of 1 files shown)


Ceiu:
@jirihnidek
Added you, as this needs someone from the subman world to clear it. Feel free to delegate review responsibilities as necessary.

jirihnidek (approved):
Good point. No clue why I called `subprocess.Popen()` with `shell=True`. Maybe it was necessary for something long time ago, when Python 2 was used.

Ceiu (approved):

PR #3758, opened 2023-01-23
[M] CANDLEPIN-455: Normalized consumer fact constants and usage

Ceiu:
- Updated the Consumer.Facts pseudo-class to contain all of the facts Candlepin uses either directly or indirectly
- Updated the CloudProfileFacts to reference Consumer.Facts rather than hard-coding its own copies of the values
- Updated several places in Candlepin code which were using hard-coded strings for facts to use the new values in Consumer.Facts

### DIFF
--- a/src/main/java/org/candlepin/model/CloudProfileFacts.java
+++ b/src/main/java/org/candlepin/model/CloudProfileFacts.java
@@ -16,23 +16,25 @@
 package org.candlepin.model;
 
 /**
- * The subscription profile time stamp should also get updated
- * on the specific Facts updates.
- * This Enum helps to configure more facts if required in the future.
+ * The CloudProfileFacts enum is a collection of consumer facts which are critical for maintaining
+ * a cloud profile in some client applications such as Subscription Watch.
+ *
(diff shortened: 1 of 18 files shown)


src/main/java/org/candlepin/model/CloudProfileFacts.java
```diff
@@ -16,23 +16,25 @@
 package org.candlepin.model;
 
 /**
- * The subscription profile time stamp should also get updated
- * on the specific Facts updates.
- * This Enum helps to configure more facts if required in the future.
+ * The CloudProfileFacts enum is a collection of consumer facts which are critical for maintaining
+ * a cloud profile in some client applications such as Subscription Watch.
+ *
+ * The facts in this enum come from the consumer facts, and are primarily used to determine whethr
```
wottop, line 22: typo

wottop (changes requested):
Just the one line

src/main/java/org/candlepin/model/CloudProfileFacts.java
Ceiu: Fixed

wottop (approved):

PR #3761, opened 2023-01-24
[M] CANDLEPIN-456: Fixed input validation on content creation endpoint

Ceiu:
- Corrected the input validation on the POST /owners/{key}/content endpoint to match the request handler's logic: the id field is no longer required, but the label, name, type, and vendor fields are
- Updated the ValidationExceptionMapper to separate constraint violations when building the response error message
- Updated a few older bits of code with to fit with modern practices and stylistic choices
- Updated a couple of the OpenAPI generator templates to produce slightly more readable code

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -8566,26 +8566,36 @@ components:
         - $ref: '#/components/schemas/TimestampedEntity'
         - type: object
           required:
-            - id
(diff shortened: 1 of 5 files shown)


Ceiu:
Nope, can't put those requirements on the objects, or it breaks our update semantics. Rejecting PR and will try again shortly.

PR #3764, opened 2023-01-24
[M] CANDLEPIN-456: Fixed input validation on content creation endpoint

Ceiu:
- Corrected the input validation on the POST /owners/{key}/content endpoint to no longer require the content ID on creation
- Updated the input validation on content creation to throw 400s when the label, name, type, or vendor fields are not provided during content creation, rather than erroring with a 500
- Updated the ValidationExceptionMapper to separate constraint violations when building the response error message
- Updated a few older bits of code with to fit with modern practices and stylistic choices
- Updated a couple of the OpenAPI generator templates to produce slightly more readable code

### DIFF
--- a/buildSrc/src/main/resources/templates/api.mustache
+++ b/buildSrc/src/main/resources/templates/api.mustache
@@ -24,7 +24,8 @@ import javax.validation.Valid;{{/useBeanValidation}}
 @Api(description = "the {{{baseName}}} API"){{/useSwaggerAnnotations}}{{#hasConsumes}}
 @Consumes({ {{#consumes}}"{{{mediaType}}}"{{^-last}}, {{/-last}}{{/consumes}} }){{/hasConsumes}}{{#hasProduces}}
 @Produces({ {{#produces}}"{{{mediaType}}}"{{^-last}}, {{/-last}}{{/produces}} }){{/hasProduces}}
-{{>generatedAnnotation}}public {{#interfaceOnly}}interface{{/interfaceOnly}}{{^interfaceOnly}}class{{/interfaceOnly}} {{classname}} {
+{{>generatedAnnotation}}
+public {{#interfaceOnly}}interface{{/interfaceOnly}}{{^interfaceOnly}}class{{/interfaceOnly}} {{classname}} {
 {{#operations}}
 {{#operation}}
 
--- a/spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
@@ -15,7 +15,7 @@
 package org.candlepin.spec.content;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.candlepin.spec.bootstrap.assertions.StatusCodeAssertions.assertBadRequest;
+import static org.assertj.core.api.Assertions.assertThatObject;
 import static org.candlepin.spec.bootstrap.assertions.StatusCodeAssertions.assertForbidden;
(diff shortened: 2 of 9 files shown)


src/main/java/org/candlepin/resource/OwnerContentResource.java
```diff
@@ -72,12 +73,41 @@ public OwnerContentResource(OwnerCurator ownerCurator,
...
+     *  the content DTO to validate
+     *
+     * @throws BadRequestException
+     *  if the DTO does not pass validation
+     */
+    private void validateContentForCreation(ContentDTO dto) {
+        this.validator.validateCollectionElementsNotNull(dto::getModifiedProductIds);
+
+        if (dto.getLabel() == null || dto.getLabel().isEmpty()) {
+            throw new BadRequestException("content label cannot be null or empty");
```
nikosmoum, line 89: These strings should probably be translated

nikosmoum (changes requested):
A couple of minor things, otherwise good

buildSrc/src/main/resources/templates/api.mustache
```diff
@@ -24,11 +24,12 @@ import javax.validation.Valid;{{/useBeanValidation}}
 @Api(description = "the {{{baseName}}} API"){{/useSwaggerAnnotations}}{{#hasConsumes}}
 @Consumes({ {{#consumes}}"{{{mediaType}}}"{{^-last}}, {{/-last}}{{/consumes}} }){{/hasConsumes}}{{#hasProduces}}
 @Produces({ {{#produces}}"{{{mediaType}}}"{{^-last}}, {{/-last}}{{/produces}} }){{/hasProduces}}
-{{>generatedAnnotation}}public {{#interfaceOnly}}interface{{/interfaceOnly}}{{^interfaceOnly}}class{{/interfaceOnly}} {{classname}} {
+{{>generatedAnnotation}}
+public {{#interfaceOnly}}interface{{/interfaceOnly}}{{^interfaceOnly}}class{{/interfaceOnly}} {{classname}} {
```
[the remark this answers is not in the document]
nikosmoum: Looking at the specific upstream files, it looks like others have made the same formatting changes as you, so ignore my previous comment

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -141,58 +149,91 @@ public void shouldAllowSuperAdminsToCreateContent() throws Exception {
         assertEquals(content, fetched);
     }
 
-    @Nested
-    @TestInstance(Lifecycle.PER_CLASS)
-    public class JsonManipulationTests {
+    private static Stream<Arguments> critialContentStringFieldsAndValues() {
```
nikosmoum, line 152: also typo: critial vs critical

buildSrc/src/main/resources/templates/api.mustache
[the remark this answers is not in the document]
Ceiu: Good to know for future things that catch my eye.

src/main/java/org/candlepin/resource/OwnerContentResource.java
Ceiu: Yup. Fixed.

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
Ceiu: fixed

nikosmoum (approved):

PR #3765, opened 2023-01-25
Fix date-time parsing of query params

Januson:
Candlepin was accepting some query parameters meant to represent dates as string and doing the parsing later. This parsing was using a different parser that did not handle all edge cases of ISO 8601. This led to random date-time parsing errors when optional parts of the date-time string were missing.

- Updated API spec file to deserialize incoming date-time strings into OffsetDateTime
- Removed the second parser


nikosmoum (approved):
Some issues to resolve in Jenkins, but looks good in general

PR #3767, opened 2023-01-25
Removing supervisord

Lorquas:
This should also work for the cs8 container if we want to port it to that.

One caveat is I removed the ssh server from the container, I'm not sure who actually uses this with the ./test workflow.

Also the test script has gotten pretty ugly I think this would be a good chance to fix up some of the args.


docker/candlepin-base-cs9/setup-supervisord.sh
```diff
@@ -1,77 +0,0 @@
...
-[program:tomcat]
-user=tomcat
-environment=CATALINA_PID="/tmp/tomcat.pid",JAVA_HOME="/usr/lib/jvm/jre"
-command=/usr/libexec/tomcat/server start
-stopsignal=TERM
-redirect_stderr=true
-TOMCAT_SUPERVISOR
-}
-
-setup_ssh() {
```
nikosmoum, line 46: does removing this means we can't ssh into the container (`docker exec -it`)? If so, that's a problem because need to be able to debug when issues come up

docker/test
```diff
@@ -78,14 +82,21 @@ if [ "$EXPOSE_PORTS" == "1" ]; then
   SERVICE_PORTS_ARG="--service-ports"
 fi
 
-if [ "$DETACHED" == "1" ]; then
-    # Run detached, and don't remove the containers
-    docker-compose $PROJ_NAME $COMPOSE_ARGS run -d $SERVICE_PORTS_ARG candlepin $TEST_CMD
-    RETVAL=$?
+if [ "$INIT_CONTAINER" == "1" ]; then
+  docker-compose $PROJ_NAME $COMPOSE_ARGS up -d
```
nikosmoum, line 86: If it's not too much trouble, lets move this straight to podman-compose for cs9 for now, and we'll do cs8 in a separate PR

docker/candlepin-base-cs9/setup-supervisord.sh
bakajstep: you still can

bakajstep: Example 
"""
podman-compose -p cs9test -f docker-compose-postgres-cs9.yml up
"""

Then you can enter it in new terminal with:

"""
podman exec -it <container_ID> /bin/bash
"""

nikosmoum: alright, then lets remove it

Lorquas:
@bakajstep   Your last commit looks good to me.

Lorquas:
@bakajstep   might need your help with rebasing on sbakaj/docker_cs9 , you've pushed to both branches since this PR and I'm not sure which parts you want to keep.

bakajstep:
> @bakajstep might need your help with rebasing on sbakaj/docker_cs9 , you've pushed to both branches since this PR and I'm not sure which parts you want to keep.

I will do it and merge it. Thank you for your help.

docker/test

PR #3768, opened 2023-01-25
[M] CANDLEPIN-122: Move Jenkins PR mariadb pipeline stages to use Centos Stream 8

bakajstep:
MariaDB pipeline stages are set to run on Centos Stream 8.
Adding support to run MariaDB stages on different OS.


docker/docker-compose-mysql-cs8.yml
```diff
@@ -8,7 +8,7 @@ services:
     volumes:
         - ./mysql.cnf:/etc/mysql/conf.d/mysql.cnf
   candlepin:
-    image: ${REGISTRY}/candlepin-base
+    image: ${REGISTRY}/candlepin-base-cs8
```
nikosmoum, line 11: Wouldn't this change be enough to make this work? Why the rest?

docker/test
```diff
@@ -11,7 +11,7 @@ USAGE: test [OPTIONS]
 OPTIONS:
   -c CMD    sets the command to be run by the container
               default: $TEST_CMD
-  -m        use mysql
+  -m [OS]   use mysql, we can specify OS where candlepin will be running. Default is Centos Stream 8: cs8.
```
nikosmoum, line 14: If we need to specify the OS/image version, It would be much better to use a completely new argument (e.g. `-a [image]` instead of making this an argument on the mysql flag

jenkins/Jenkinsfile
```diff
@@ -108,6 +108,7 @@ pipeline {
                     environment {
                         CANDLEPIN_DATABASE = 'mysql'
                         CP_TEST_ARGS = '-r'
+                        OS_IMAGE = 'cs8'
```
nikosmoum, line 111: why is this only set at the standalone-mysql? should it also not be on the hosted-mysql stage?

nikosmoum (changes requested):

jenkins/Jenkinsfile
bakajstep: I want to just show how you can specify that OS, so i just set it to one job. So next person can see you could select OS or not and use default.

docker/docker-compose-mysql-cs8.yml
bakajstep: yes, but i was thinking if we decide in future to run different OS like with postgres, so I thought it would be ready.

docker/test
bakajstep: I know we talked about it, but then I thought about it and I thought that there are a lot of edge cases that we'll have to cover(both parameters need to be set, different OS for mariadb and postgres and more ) and I don't know who all uses the script, so this seemed like a safe choice, but I can redo it if you want.

nikosmoum: I think we should re-do it by using a separate flag. This is not intuitive and will confuse people, because one flag is used for specifying two arguments (db, and image)

jenkins/Jenkinsfile
nikosmoum: If it is not functional, we should remove it. Anyone looking to do something like this will be able to tell by checking `docker/test --help`

docker/test

jenkins/Jenkinsfile
bakajstep: it has functionality, but same as default. Ok, I remove it.

docker/test
```diff
@@ -2,6 +2,7 @@
 
 TEST_CMD="/usr/bin/cp-test -t -u -r"
 DIR="$(git rev-parse --show-toplevel)/docker/"
+OPERATION_SYSTEM="cs8" # Changing this variable, change default OS
```
nikosmoum, line 5: nitpick: OPERATING instead of OPERATION

docker/test
```diff
@@ -11,6 +12,15 @@ USAGE: test [OPTIONS]
 OPTIONS:
   -c CMD    sets the command to be run by the container
               default: $TEST_CMD
+  -o OS     Specify operation system where candlepin will be running
```
nikosmoum, line 15: same here: operating

docker/candlepin-base-cs8/setup-devel-env.sh
```diff
@@ -81,6 +81,9 @@ bundle install --without=proton --retry=5 --verbose --full-index
 # Installs all Java deps into the image, big time saver
 ./gradlew --no-daemon dependencies
 
+# Fix issue with owner on mount volume
```
nikosmoum, line 84: Does this issue happen only on cs8 and cs9? Or does it also need to be fixed in c7 (because older branches still need that)?

docker/test
```diff
@@ -54,6 +66,15 @@ while getopts ":c:mn:pxqlde" opt; do
   esac
 done
 
+COMPOSE_ARGS="$COMPOSE_ARGS-$OPERATION_SYSTEM.yml"
+
+if ! [ $COMPOSE_ARGS ];
+then
+  echo "Docker compose file does not exist"
```
nikosmoum, line 73: Since this error is the result of someone choosing an operating system that isn't supported, like `-o cs10`, can we change this to something like `Operating system not supported`?

nikosmoum (changes requested):

docker/test

docker/test
bakajstep: Done, I also print that bad os what user enter.

nikosmoum (approved):

PR #3770, opened 2023-01-25
CANDLEPIN-375: Remove synchronous hypervisor checkin and tests

wottop:


src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -107,132 +99,6 @@ public HypervisorResource(ConsumerResource consumerResource, ConsumerCurator con
...
-                if (hypervisorConsumersMap.get(hypervisorId) == null) {
-                    if (!createMissing.booleanValue()) {
-                        log.info("Unable to find hypervisor with id {} in org {}", hypervisorId, ownerKey);
-                        result.setFailedUpdate(addFailed(result.getFailedUpdate(),
-                            hypervisorId + ": " +
-                            i18n.tr("Unable to find hypervisor in org \"{0}\"", ownerKey)));
-                        continue;
-                    }
-                    log.debug("Registering new host consumer for hypervisor ID: {}", hypervisorId);
-                    consumer = createConsumerForHypervisorId(hypervisorId, owner, principal);
```
JoshAlbrecht23, line 198: I think we should be good to remove the `createConsumerForHypervisorId` method. It looks like the hypervisorUpdate method is the only one using that private method.

src/main/java/org/candlepin/resource/HypervisorResource.java
```diff
@@ -107,132 +99,6 @@ public HypervisorResource(ConsumerResource consumerResource, ConsumerCurator con
...
-                List<GuestId> guestIds = new ArrayList<>();
-                consumerResource.populateEntities(guestIds, hostEntry.getValue());
-                boolean guestIdsUpdated = addGuestIds(consumer, guestIds);
-
-                Date now = new Date();
-                consumerCurator.updateLastCheckin(consumer, now);
-                consumer.setLastCheckin(now);
-                // Populate the result with the processed consumer.
-                if (hostConsumerCreated) {
-                    result.setCreated(addHypervisorConsumerDTO(result.getCreated(), consumer));
```
JoshAlbrecht23, line 217: I think that we should be able to remove the `addHypervisorConsumerDTO` method as well.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerCheckinSpecTest.java
```diff
@@ -107,22 +106,6 @@ public void shouldUpdateLastCheckinTimeWhenFetchingEntitlementCertSerials() thro
...
 
-    @Test
-    public void shouldUpdateLastCheckinTimeOnHypervisorCheckin() throws Exception {
-        ApiClient adminClient = ApiClients.admin();
-        OwnerDTO owner = adminClient.owners().createOwner(Owners.random());
-        ConsumerDTO consumer = adminClient.consumers().createConsumer(Consumers.random(owner));
-        ApiClient consumerClient = ApiClients.ssl(consumer);
-        OffsetDateTime initialCheckin = consumer.getLastCheckin();
-
-        consumerClient.hypervisors().hypervisorUpdate(owner.getKey(), new HashMap<>(), false);
```
JoshAlbrecht23, line 118: Instead of fully removing the test, would it be valid to just replace the `hypervisorUpdate` method with the `hypervisorUpdateAsync` method?

JoshAlbrecht23 (changes requested):
Looks good, I just noticed some additional methods that can be removed and had a question.

api/candlepin-api-spec.yaml
```diff
@@ -3595,65 +3595,6 @@ paths:
...
-          schema:
-            type: boolean
-            default: true
-      responses:
-        200:
-          description: A successful operation
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/HypervisorUpdateResultDTO'
```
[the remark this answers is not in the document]
wottop: The async job returns this data structure also. It is returned as an object but then converted with a mapper.
We still need this class generated.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerCheckinSpecTest.java
wottop: The next method in the class is the async one so this can be cut.

api/candlepin-api-spec.yaml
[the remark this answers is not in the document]
JoshAlbrecht23: Good catch, you are right! We can't remove then!

JoshAlbrecht23 (approved):

PR #3773, opened 2023-01-26
[M] CANDLEPIN-459: Updated Util.getHostname to no longer use reflection

Ceiu:
- Updated Util.getHostname to drop the use of reflection, as this method has been blocked with new restrictions as of Java 17
- Added a new configuration for specifying the name of a node to use while setting the origin and executor for jobs within the job system


wottop:
lgtm

PR #3774, opened 2023-01-27
CANDLEPIN-261: Migrate environment_spec to Java (part 2)

Januson:


spec-tests/src/test/java/org/candlepin/spec/environments/EnvironmentSpecTest.java
```diff
@@ -466,7 +470,7 @@ public void shouldGiveConsumerContentFromMultipleEnvironments() {
         ConsumerDTO consumer = ownerClient.consumers().createConsumer(Consumers.random(owner)
             .environments(List.of(env1, env2, env3)));
         ApiClient consumerClient = ApiClients.ssl(consumer);
-        consumerClient.consumers().bindPoolSync(consumer.getUuid(), pool.getId(), 1);
+        consumerClient.consumers().bindPool(consumer.getUuid(), pool.getId(), 1);
```
wottop, line 473: Every other invocation in the two spec files is to bindPoolSync. Why is this one different?

wottop (changes requested):

wottop:
Looks good. Just one question.

spec-tests/src/test/java/org/candlepin/spec/environments/EnvironmentSpecTest.java
Januson: Because it doesn't care about the response. `bindPoolSync` deserializes the synchronous version of the response. Since the test doesn't use the response I used `bindPool` to skip the deserialization.

wottop: good enough

wottop:
ack

PR #3775, opened 2023-01-27
CANDLEPIN-462: Update spec tests that modify Gson object

JoshAlbrecht23:


spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -653,82 +653,39 @@ public void shouldAllowDeletingAProductAssociatedWithAnActivationKey() {
...
-            // and ensuring we're actually testing fields that are explicitly set to null
-            this.client = ApiClients.admin();
-
-            this.unmodified = this.client.getApiClient().getJSON().getGson();
-            assertNotNull(this.unmodified);
-
-            Gson modified = this.unmodified.newBuilder()
-                .serializeNulls()
-                .create();
+        ObjectNode productNode = (ObjectNode) ApiClient.MAPPER.readTree(Contents.random().toJson());
```
Ceiu, line 675: Couple things here:
1) this is building a content DTO from Contents.random(), rather than a product. Given the nature of this test, we should be manually building a fully-populated product so we can guarantee the target field is the one that's generating the error code we're expecting. Should also check the message includes the field in question.

2) The conversion to `ObjectNode` can be done such that you don't need a cast if you use `readValue` instead.

You can see examples of this in `OwnerContentSpecTest.shouldRequireCriticalStringFieldsAreValidWhenCreatingContent`

Ceiu (changes requested):
We should also add a test to capture the "not present" state for the critical fields.

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
JoshAlbrecht23: Good catch!! 
- Updated the test to use a fully populated Product
- Updated to use `ApiClient.MAPPER.readValue` instead of `ApiClient.MAPPER.readTree`
- Added shouldRequirePopulatedCriticalStringFieldsWhenInsertingProduct to test that required fields are present

src/main/java/org/candlepin/resource/OwnerProductResource.java
```diff
@@ -154,7 +154,7 @@ public ProductDTO createProductByOwner(String ownerKey, ProductDTO dto) {
         }
 
         if (dto.getName() == null || dto.getName().matches("^\\s*$")) {
-            throw new BadRequestException(i18n.tr("product has null or invalid name"));
+            throw new BadRequestException(i18n.tr("product has a null or invalid name"));
```
JoshAlbrecht23, line 157: Updated this to match the first BadRequest message on line 153 so we can assert against both messages in the same spec test.

Ceiu (approved):

PR #3778, opened 2023-01-30
CANDLEPIN-461: Update openapi-generator to 6.3.0

wottop:
Now goes to 6.3.0

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -26,7 +26,7 @@ ext.plugins = [
     "com.adarshr:gradle-test-logger-plugin:3.2.0",
     "com.github.ben-manes:gradle-versions-plugin:0.45.0",
     "com.netflix.nebula:gradle-lint-plugin:18.0.3",
-    "org.openapitools:openapi-generator-gradle-plugin:6.0.1",
+    "org.openapitools:openapi-generator-gradle-plugin:6.3.0",
     "org.owasp:dependency-check-gradle:8.1.0",
     "org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:3.5.0.2730",
 ]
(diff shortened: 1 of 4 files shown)


dependencies.gradle
```diff
@@ -26,7 +26,7 @@ ext.plugins = [
     "com.adarshr:gradle-test-logger-plugin:3.2.0",
     "com.github.ben-manes:gradle-versions-plugin:0.44.0",
     "com.netflix.nebula:gradle-lint-plugin:17.8.0",
-    "org.openapitools:openapi-generator-gradle-plugin:6.0.1",
```
Januson, line 29: Our Templates and customized AbstractJavaJAXRSServerCodegen have to be updated as well while preserving changes we made.

Also version 6.3 was released a few days ago so we should probably update to that version while we are at it.

wottop: AbstractJavaJAXRSServerCodegen was updated when we went to 6.0. I don't believe it needs updating. Let me know if you see something.

wottop:
Note: With 6.2.1 we don't seem to need changes to the template files or AbstractJavaJAXRSServerCodegen
With 6.3.0 we get errors such that we need to change client/buildSrc/src/main/resources/templates/pojo.mustache to import both ApiModel and ApiModelProperty. Plus there is an error in HttpUrl.kt about the URL not starting with http or https. I imagine this means a change to AbstractJavaJAXRSServerCodegen. Still worth going up to 6.3.0?

Januson:
I think it might be worth updating to the latest. @Ceiu @nikosmoum @JoshAlbrecht23 What do you think?

nikosmoum:
> Note: With 6.2.1 we don't seem to need changes to the template files or AbstractJavaJAXRSServerCodegen With 6.3.0 we get errors such that we need to change client/buildSrc/src/main/resources/templates/pojo.mustache to import both ApiModel and ApiModelProperty. Plus there is an error in HttpUrl.kt about the URL not starting with http or https. I imagine this means a change to AbstractJavaJAXRSServerCodegen. Still worth going up to 6.3.0?

If by "we need changes in templates and AbstractJavaJAXRSServerCodegen" you mean merging in the upstream changes that were made in 6.3.0, then we should do them, since we will eventually have to once we upgrade to that version or above.

Also, what is the root cause of that HttpURL error? Is it a regression bug in 6.3.0 that we have to add a workaround fix for? If so, does it have an issue in their github that we can track?

wottop:
A change to the buildUrl method in the generated ApiClient class required the change to ApiClientFactory

wottop:
All issues mentioned above have been remedied. This can be reviewed again.

Januson (approved):

PR #3780, opened 2023-01-31
CANDLEPIN-239: Upgraded hibernate-validator to 6.x

JoshAlbrecht23:
- Also updated hibernate-validator-annotation-processor to 6.x


dependencies.gradle
```diff
@@ -53,8 +53,8 @@ libraries["hibernate"] = "org.hibernate.javax.persistence:hibernate-jpa-2.1-api:
 libraries["hibernateC3p0"] = "org.hibernate:hibernate-c3p0:5.6.7.Final"
 libraries["hibernateJcache"] = "org.hibernate:hibernate-jcache:5.6.7.Final"
 libraries["hibernateJpamodelgen"] = "org.hibernate:hibernate-jpamodelgen:5.6.12.Final"
-libraries["hibernateValidator"] = "org.hibernate:hibernate-validator:5.4.3.Final"
-libraries["hibernateValidatorProcessor"] = "org.hibernate:hibernate-validator-annotation-processor:5.4.3.Final"
+libraries["hibernateValidator"] = "org.hibernate:hibernate-validator:6.2.5.Final"
+libraries["hibernateValidatorProcessor"] = "org.hibernate:hibernate-validator-annotation-processor:6.2.5.Final"
```
nikosmoum, line 57: Let's also update the namespace from `org.hibernate:` to `org.hibernate.validator:`, because apparently they have moved to that one (as per the note [here](https://mvnrepository.com/artifact/org.hibernate/hibernate-validator)).

pom.xml
```diff
@@ -232,13 +232,13 @@
     <dependency>
       <groupId>org.hibernate</groupId>
       <artifactId>hibernate-validator</artifactId>
-      <version>5.4.3.Final</version>
+      <version>6.2.5.Final</version>
```
nikosmoum, line 235: Same for these 2: groupId should be `org.hibernate.validator`

nikosmoum (changes requested):

dependencies.gradle
JoshAlbrecht23: Good catch! I missed that note when I was looking up the versions

pom.xml
JoshAlbrecht23: Updated

nikosmoum (approved):

PR #3782, opened 2023-01-31
test data importers: simplify productid addition to repos

ptoscano:
The current approach to add the productid certificate to each created repositories is:
- get the content of the certificate:
  - if it exists locally, read it fully, and return it
  - if it does not exists, fetch it from Candlepin, save it, and return it
- write it as "productid"
- add it to the repository
- delete this temporary "productid" file

Simplify all the above process a bit:
- always return the path to the certificate file, no matter whether it existed already or it was fetched from Candlepin
- when adding it to the repository, create "productid" as temporary link to the file

This has few improvements:
- existing certificates are not read at all
- creating & removing the temporary "productid" is a cheap link/unlink operation
- the internal get_productid_cert() does not return bytes anymore


wottop:
ack

PR #3784, opened 2023-02-01
[M] Set default compose file in docker/test script

nikosmoum:
- This solves an error when the user does not explicitly set -m, -p, -q, -x to choose a compose profile. The default is now set to postgresql on Centos Stream 8.


nikosmoum:
Notes on how to test this: try to run `./docker/test -c 'cp-test -d'`.
At the moment this fails because no -m/p/q/x was specified. It should work by defaulting to postgres (as if -p was passed)

nikosmoum:
Closing this, since we're removing compose files in favor of direct podman commands

PR #3791, opened 2023-02-02
CANDLEPIN-468: R2 CA cert with new Hosted CA cert

JoshAlbrecht23:
After the new CA cert was moved into Stage, I generated a manifest from Stage to import against my changes in Vagrant VM. I copied the new candlepin-authority.crt into the /etc/candlepin/certs/upstream directory and then imported the manifest and verified that the import job was successful.

**Import asyn job creation:**
```
curl -k -X POST -F "application/zip=@manifest.zip" -u admin:admin https://localhost:8443/candlepin/owners/donaldduck/imports/async
{
  "created" : "2023-02-02T20:32:54+0000",
  "updated" : "2023-02-02T20:32:54+0000",
  "id" : "ff8081818613d33a018613d50acf0001",
  "key" : "ImportJob",
  "name" : "Import Manifest",
  "group" : null,
  "origin" : "candlepin-el8.example.com",
  "executor" : null,
  "principal" : "admin",
  "state" : "CREATED",
  "previousState" : "CREATED",
  "startTime" : null,
  "endTime" : null,
  "attempts" : 0,
  "maxAttempts" : 1,
  "statusPath" : "/jobs/ff8081818613d33a018613d50acf0001",
  "resultData" : null
}
```

**Successful job:**
```
curl -k -u admin:admin https://localhost:8443/candlepin/jobs/ff8081818613d33a018613d50acf0001?result=true
{
  "created" : "2023-02-02T20:32:54+0000",
  "updated" : "2023-02-02T20:33:05+0000",
  "id" : "ff8081818613d33a018613d50acf0001",
  "key" : "ImportJob",
  "name" : "Import Manifest",
  "group" : null,
  "origin" : "candlepin-el8.example.com",
  "executor" : "candlepin-el8.example.com",
  "principal" : "admin",
  "state" : "FINISHED",
  "previousState" : "RUNNING",
  "startTime" : "2023-02-02T20:32:54+0000",
  "endTime" : "2023-02-02T20:33:05+0000",
  "attempts" : 1,
  "maxAttempts" : 1,
  "statusPath" : "/jobs/ff8081818613d33a018613d50acf0001",
  "resultData" : {"created":1675369985563,"updated":1675369985563,"id":"ff8081818613d33a018613d5361b4277","status":"SUCCESS","statusMessage":"donaldduck file imported successfully.","fileName":"manifest.zip","generatedBy":"jalbrech-CANDLEPIN-468","generatedDate":1675368617772,"upstreamConsumer":{"created":1675369985563,"updated":1675369985563,"id":"ff8081818613d33a018613d5361b4278","uuid":"b1a9b661-8309-460e-9a03-474056139c63","name":"sub-alloc-01","type":{"created":1675369975070,"updated":1675369975070,"id":"ff8081818613d33a018613d50d1e0005","label":"sam","manifest":true},"ownerId":"ff8081818613c813018613c829b70004","contentAccessMode":"org_environment","webUrl":"access.stage.redhat.com/management/subscription_allocations/","apiUrl":"https://subscription.rhsm.stage.redhat.com/subscription/consumers/"}}
```


candlepin.spec.tmpl
```diff
@@ -116,7 +116,7 @@ rm -rf %{buildroot}
 ## common
 %{__install} -d -m 755 %{buildroot}/%{_sysconfdir}/%{name}/certs/upstream/
 %{__install} -m 644 config/candlepin/candlepin-redhat-ca.crt %{buildroot}%{_sysconfdir}/%{name}/certs/upstream/
-%{__install} -m 644 config/candlepin/candlepin-redhat-r2-ca.crt %{buildroot}%{_sysconfdir}/%{name}/certs/upstream/
+%{__install} -m 644 config/candlepin/candlepin-authority.crt %{buildroot}%{_sysconfdir}/%{name}/certs/upstream/
```
JoshAlbrecht23, line 119: @nikosmoum I was uncertain on what the name of the new certificate should be? Should we establish a naming convention the certificates?

nikosmoum: Good question, but that brings something else to mind: If I remember correctly, the new CA cert should be able to validate signatures coming from both the old and new certs. So we can entirely replacing the old certificate with the new one (and keep the name the same). But that needs testing, which would be: Export manifest from Prod (using the old cert atm), and try to import to local candlepin that **only** has the new cert (no old, no r2). The expected outcome would be it works and we go forward with only the new cert. If it fails, then we need both certs and we'll think of a name for the new one then.

I'd also throw in a negative test just to be sure: Export manifest from Stage (using the new cert atm), and try to import to local candlepin _without_ this change. The expected outcome would be the import fails.

nikosmoum (approved):

PR #3795, opened 2023-02-03
[M] Updated the refresh_data_extractor tool to support .gz files

Ceiu:
- Added basic support for reading gzipped logs when extracting
  refresh data using the refresh_data_extractor.sh script. It
  will now automatically decompress any such input file using
  the 'gunzip' command, if `gunzip -t` on the file returns
  successfully


wottop:
ack

PR #3798, opened 2023-02-06
CANDLEPIN-207: Remove support of RHEL7 JSS version

Januson:


src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
```diff
@@ -122,43 +74,20 @@ public static synchronized void initialize() {
...
-                e.getCause() instanceof CertDatabaseException ||
-                e.getCause() instanceof GeneralSecurityException) {
-                throw new JSSLoaderException("Could not initialize CryptoManager!", e.getCause());
-            }
-            else {
-                throw new JSSLoaderException("Could not initialize CryptoManager!", e);
-            }
+        catch (AlreadyInitializedException | CertDatabaseException | GeneralSecurityException |
+            KeyDatabaseException e) {
+            throw new JSSLoaderException("Could not initialize CryptoManager!", e);
```
nikosmoum, line 90: `AlreadyInitializedException` shouldn't result in an exception, only a warning

nikosmoum (changes requested):
One change requested, and the main question:
- Has this been tested and works in both RHEL8 and RHEL9?

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
```diff
@@ -109,7 +61,7 @@ public static String getJSSVersion() {
      * Initializes the JSS CryptoManager (which in turn initializes the NSS DB), and loads the JSS provider
      * into the JVM. The method should be called during context initialization. Can be called more than once
      * without ill effect (-1 will be returned if the provider is already installed).
-     *
+     * <p>
```
Ceiu, line 64: Need to close the paragraph tag

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
Ceiu: Should also use this opportunity to move the InitializationValues bits into the try block for better organization.

Ceiu (changes requested):

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
Januson: Yes I tested in both 8 and 9 (both downgrade and Java17).

I think the exception here is correct. `AlreadyInitializedException` should never happen because we handle that case at the beginning of the method. Warning would leave us in invalid state.

@Ceiu Why? Putting simple assignmnets into try block will only obscure the single line that actually throws something.

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
Januson: Javadoc uses only opening tags for paragraphs. But this is something I wanted to bring up. Some of our javadoc is not formatted correctly (such as paragraphs as javadoc ignores blank lines) and because of that the output gets squished into a single wall of text which makes documentation pop-ups in IDE less useful.

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
nikosmoum: Technically it should never happen at the moment, but code changes, so if for whatever reason that check is removed in the future, and something tries to re-initialize the cryptomanager, `AlreadyInitializedException` will be thrown. And if that happens there is no invalid state afaik. Something tried to re-initialize, it failed, but everything still works, so a warning is better instead of throwing a catastrophic error that will stop candlepin from deploying

Januson: Even in that case an error would make sense. Candlepin would would fail according to fail-fast phylosophy as broken code was introduced.

nikosmoum: Fair enough, failing fast sounds good

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
Ceiu: +1 to addressing of paragraph formatting

The part I was mentioning here has something to do with how some parsers handle that dangling paragraph tag. I don't recall the details, and it may not be important (anymore?), but a number of years ago I remember we had to stop using the XHTML variant of the tag because it broke something or someone's IDE.

If you find anything on that which lines up with the memory fragment I have here, it'd be good to get a refresher as to what that issue is and whether or not we need to care about well-formed HTML in our docs.

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
Ceiu: @Januson 
Stylistic choice, I guess. In my view, you can't do the targeted op without those initialization values -- they're basically arguments for it -- so they logically live/belong with that statement like the arguments of any typical function call. Additionally, they're not used outside of it, so my preference is to block them up accordingly.

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
```diff
@@ -109,7 +61,7 @@ public static String getJSSVersion() {
      * Initializes the JSS CryptoManager (which in turn initializes the NSS DB), and loads the JSS provider
      * into the JVM. The method should be called during context initialization. Can be called more than once
      * without ill effect (-1 will be returned if the provider is already installed).
```
nikosmoum, line 63: Can we update the last sentence here? This method isn't returning an integer, so it should say something like `the method call will be a no-op` instead if `-1 will be returned`

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
```diff
@@ -109,7 +61,7 @@ public static String getJSSVersion() {
...
-     *
+     * <p>
      * Note: This method uses reflection to initialize the CryptoManager, due to dynamically loading the
      * InitializationValues class, depending on which JSS version is on the classpath. This is because of
      * breaking changes introduced between JSS 4.4.X (RHEL 7) and 4.5.0+ (RHEL 8).
```
nikosmoum, line 67: Lets remove this now since it is irrelevant

nikosmoum (changes requested):

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
Januson: Done.

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
Januson: Done.

src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java
Januson: I checked how it works and it's a mess. Attempt to generate our javadoc results in ~100 errors and ~100 warnings. XHTML tags such as <br /> break it. Empty tags such as <p></p> break it and multitude of other issues.

I also found the JEP about HTML5, but unfortunately that makes only the output HTML5 compliant not the javadoc itself. But the solution is on the way. It looks like OpenJDK is working on supporting markdown in javadoc.

Ceiu: It's both amusing and frustrating that the only syntax that works correctly is the one that's invalid and only in existence due to the legacy of everyone doing it wrong at the dawn of the web.

But good to know, and the move to markdown would be a welcome one -- even though we know we'll get yet another weird implementation that follows none of the existing standards.

Thanks for looking into this.

PR #3804, opened 2023-02-07
[M] CANDLEPIN-480: Fixed product and content attribute referencing

Ceiu:
- Added entity class guidelines and documentation to the
  AbstractHibernateObject
- Changed several methods in Product and Content to no longer
  reference model attributes outside of mutators and accessors
- Added entity version to the string representation of Product
  and Content
- Improved output of entity version collision error message in
  ProductNodeVisitor and ContentNodeVisitor
- Changed how content processes modified product IDs during
  entity version calculation
- Removed XML serialization annotations from AbstractHibernateObject,
  Content, and Product
- AbstractHibernateObject now returns a typed self-reference from
  its mutators
- Deprecated the AbstractHibernateObject.populate method, and the
  AbstractHibernateObject.DEFAULT_SORT_FIELD property

### DIFF
--- a/src/main/java/org/candlepin/controller/refresher/visitors/ContentNodeVisitor.java
+++ b/src/main/java/org/candlepin/controller/refresher/visitors/ContentNodeVisitor.java
@@ -246,8 +246,8 @@ public class ContentNodeVisitor implements NodeVisitor<Content, ContentInfo> {
                 // Impl note:
                 // We've already implicitly checked the ID above by how we're pulling candidate
                 // entities from the map
-                log.error("Entity version collision detected; attempting resolution... {} != {}",
-                    entity, candidate);
+                log.error("Entity version collision detected; attempting resolution..." +
+                    "\nConflicting entities:\n{}\n{}", entity, candidate);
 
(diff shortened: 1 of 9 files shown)


nikosmoum (commented):
The changes look good, but we should document
1.  In the commit message: the reason for not referencing fields directly (what quirk does Hibernate do and what it results in, etc.) 
2. Somewhere in code (perhaps the class javadoc of Content and Product?): The logic behind when to use methods instead of field references. Which types of methods need it, and which dont, what entities (only Product and Content, or?)

Ceiu:
@nikosmoum 
It's a general Java guideline in many cases (let subclasses do stuff), but I can also document the general approach and reasoning in either Content/Product or AbstractHibernateObject or some such. Do you have a preference for where it should live?

nikosmoum:
@Ceiu AbstractHibernateObject sounds good

Ceiu:
@nikosmoum done

nikosmoum:
There is a weird git checkout error when jenkins is trying to run for this PR. Trying to figure this out

PR #3808, opened 2023-02-09
Draft pipeline

Januson:


build.gradle
```diff
@@ -321,6 +321,33 @@ test {
             '-XX:+HeapDumpOnOutOfMemoryError'
     ]
 
+    filter {
+        //include specific method in any of the tests
+        excludeTestsMatching("*Curator*")
```
Ceiu, line 326: Does this match the fqcn? If so, this might be the point at which we actually namespace our tests properly and just filter on it.

Januson: Yes. It should understand the same patterns as `--tests` filter. https://docs.gradle.org/current/userguide/java_testing.html#test_filtering

jenkins/Jenkinsfile
```diff
@@ -19,68 +19,79 @@ pipeline {
...
-                        sh 'sudo chown -R jenkins:jenkins $WORKSPACE'
-                        checkout scm
-                        sh 'sh jenkins/unit-tests.sh'
-                    }
-                    post {
-                        always {
-                            sh 'sh jenkins/cleanup.sh'
-                            archiveArtifacts artifacts: 'unit-artifacts/**',
-                                             fingerprint: true,
-                                             onlyIfSuccessful: false
```
nikosmoum, line 35: Why was this removed? Without it, the unit test logs no longer appear in the "Artifacts" tab in jenkins

build.gradle
```diff
@@ -307,11 +307,13 @@ msgfmt {
 }
 
 test {
+    finalizedBy jacocoTestReport
```
nikosmoum, line 310: I don't think we want to directly create either a `dependsOn` or a `finalizedBy` relationship between the two tasks. This makes it so that no matter which task we run both will always run. I can see that a developer might want to only run unit tests locally, but not coverage, or, similarly, run coverage on an existing unit test report, without having to re-run the unit tests.

The best would probably be to remove both of these, and create a third task called `testAndCoverage`, that dependsOn test, and is finalyzedBy jacocoTestReport. And that task is called in the Jenkinsfile

nikosmoum (changes requested):
There's an issue with archiving unit test logs, and a few other minor issues and suggestions, but overall great change

jenkins/Jenkinsfile
Januson: Because those artifacts are collected from docker and so are only relevant for spec test runs. For unit tests i replaced this with junit report collector. Check tests panel next to artifacts in Jenkins.

nikosmoum: Ok, that looks good. I assume any failing tests will be shown up first, to avoid having to scroll through thousands of tests and click 'Show More' at the bottom of the page hundreds of times?

nikosmoum: Maybe you can purposely introduce a broken unit test to see if it works like that?

build.gradle
Januson: I think we should leave at least one direction of the relationship there. Tests can run without coverage, but running coverage implies running tests.

build.gradle
```diff
@@ -324,6 +326,8 @@ test {
 }
 
 jacocoTestReport {
```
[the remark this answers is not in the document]
Januson: Done.

jenkins/Jenkinsfile
```diff
@@ -19,68 +19,79 @@ pipeline {
         }
         stage('Test') {
             parallel {
-                stage('unit') {
-                    // ensures that this stage will get assigned its own workspace
-                    agent { label 'candlepin' }
-                    steps {
-                        sh 'sudo chown -R jenkins:jenkins $WORKSPACE'
-                        checkout scm
-                        sh 'sh jenkins/unit-tests.sh'
```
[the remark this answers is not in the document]
Januson: Done.

docker/cp-test
```diff
@@ -27,87 +27,17 @@ export AUTOCONF=1
 export FORCECERT=1
 export UNITTEST=0
 
-# functions to abstract away buildr vs. gradle differences
-build_clean() {
```
nikosmoum, line 31: We can't remove this stuff from cp-test because its packaged in the container images, which are used by older branches

build.gradle
nikosmoum: ok, sounds reasonable. The 'run coverage on existing unit test report' use case is probably extremely rare if ever needed

docker/cp-test
nikosmoum: You can at least backport this change to candlepin-4.2-HOTFIX (which was cut just yesterday from master), and maybe even candlepin-4.1-HOTFIX, which will make us get rid of this stuff earlier in the future (in about ~2 months when 4.0 goes out of support).

jenkins/Jenkinsfile
Januson: Done.

docker/cp-test
```diff
@@ -236,28 +160,17 @@ OPTIONS:
...
-                    1) SONAR_HOST_URL
-                    2) BRANCH_NAME, branch to be analyzed
-                    3) CHANGE_ID, pull request ID (when Uploading PR)
-                    4) CHANGE_TARGET, pull request target branch (when Uploading PR)
   -a <arguments>    extra arguments to pass to the Candlepin deploy script (implies -d)
 HELP
 }
 
 ARGV=("$@")
-while getopts ":dtRqrHulskib:c:a:vj:n" opt; do
```
nikosmoum, line 260: A couple issues with the colons here: a letter is supposed to be followed by a colon when that option requires an argument (OPTARG). So a) by removing `b`, you should also remove the colon after it, because now it is right after option `k`, which doesn't need it. Also, when removing `n`, you also removed the colon before it, which belonged to `j`, so that needs to be added back

nikosmoum (approved):

PR #3810, opened 2023-02-09
[M] CANDLEPIN-480: Fixed incorrect variable reference in Content.getEntityVersion

Ceiu:
- Updated Content.getEntityVersion to reference the correct variable when adding modified product IDs to the hash


wottop:
ack

PR #3833, opened 2023-02-10
[M] CANDLEPIN-481: Updated the Consumer model object and related children

Ceiu:
- Updated Consumer and consumer-adjacent model objects to follow stricter encapsulation and field referencing rules
- Removed extraneous constructors that were primarily used for test logic and were never used in production code
- Normalized translation of consumer objects to explicitly set fields when no data is available for a given field
- Fixed issue in consumer creation that could cause some consumer data to be lost
- Fixed issue in some spec tests that was preventing distributor capabilities to not be assigned to consumers during import/export testing
- Fixed issue with consumer update that could trigger extraneous field updates when a consumer is updated to itself

### DIFF
--- a/src/main/java/org/candlepin/model/Consumer.java
+++ b/src/main/java/org/candlepin/model/Consumer.java
@@ -15,16 +15,8 @@
 package org.candlepin.model;
 
 import org.candlepin.exceptions.DuplicateEntryException;
-import org.candlepin.jackson.HateoasArrayExclude;
-import org.candlepin.jackson.HateoasInclude;
-import org.candlepin.jackson.StringTrimmingConverter;
 import org.candlepin.service.model.ConsumerInfo;
 import org.candlepin.util.Util;
-
-import com.fasterxml.jackson.annotation.JsonFilter;
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 
 import org.apache.commons.lang3.StringUtils;
 import org.hibernate.annotations.BatchSize;
@@ -35,6 +27,8 @@ import org.hibernate.annotations.GenericGenerator;
 import org.hibernate.annotations.Type;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
@@ -43,6 +37,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -66,10 +61,6 @@ import javax.persistence.Transient;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
-import javax.xml.bind.annotation.XmlAccessType;
-import javax.xml.bind.annotation.XmlAccessorType;
-import javax.xml.bind.annotation.XmlRootElement;
-import javax.xml.bind.annotation.XmlTransient;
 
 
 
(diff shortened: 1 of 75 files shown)


Ceiu:
For reviewers:
This is a preview of the larger model normalization changes, and is one of the bigger such changes due the age and reach of the consumer object. 

Most of the changes occur in testing, since so many tests were reliant upon these custom constructors rather than local utility methods, but some of the scarier changes are around the encapsulation of the child collections. If we don't have 100% test/spec coverage of flows that use the strange `.get<collection>().add(<object>)` pattern, I may not have caught it in this PR.

This draft will be up for some time for evaluation due to the sheer size and number of mind-numbing changes that are flanking critical updates.

wottop (changes requested):
Sorry, left it as a comment.

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -867,27 +1010,29 @@ public Release getReleaseVer() {
...
-            for (ConsumerCapability cc : this.capabilities) {
-                cc.setConsumer(this);
-            }
+        this.capabilities.clear();
+
+        if (capabilities != null) {
+            capabilities.stream()
+                .filter(Objects::nonNull)
+                .peek(cc -> cc.setConsumer(this))
+                .forEach(this.capabilities::add);
```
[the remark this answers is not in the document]
Ceiu: All objects extending `AbstractHibernateObject` automatically update their last updated time on every persist. Even if we set the value here, it'll be overwritten by that persist hook.

src/test/java/org/candlepin/test/TestUtil.java
```diff
@@ -324,13 +222,123 @@ public static String randomString() {
...
+            .username(userName)
+            .type(type)
+            .autoheal(true)
+            .serviceLevel("")
+            .entitlementCount(0L)
+            .facts(new HashMap<>())
+            .installedProducts(new HashSet<>())
+            .guestIds(new ArrayList<>())
+            .environments(new ArrayList<>());
+
```
JoshAlbrecht23, line 297: Extra new line

src/test/java/org/candlepin/test/TestUtil.java
```diff
@@ -324,13 +222,123 @@ public static String randomString() {
...
+        return createDistributor(createOwner());
+    }
+
+    public static Consumer createDistributor(Owner owner) {
+        ConsumerType ctype = new ConsumerType(ConsumerType.ConsumerTypeEnum.CANDLEPIN);
+        ctype.setId("test-ctype-" + randomInt());
+
+        return createConsumer(ctype, owner);
+    }
+
```
JoshAlbrecht23, line 323: Extra new line

src/test/java/org/candlepin/test/TestUtil.java
```diff
@@ -324,13 +222,123 @@ public static String randomString() {
...
+     * Create a consumer with a new owner
+     *
+     * @return Consumer
+     */
+    public static Consumer createConsumer(Owner owner) {
+        Consumer consumer = createConsumer(createConsumerType(), owner);
+
+        // TODO: FIXME: These *should* live on the main createConsumer for consistency and sanity,
+        // but for whatever reason we decided rewriting the same thing 9001 with slight variations
+        // would be more spicy. Move these if they don't break too many tests.
```
JoshAlbrecht23, line 263: is the '9001' a mistake in this comment?

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -528,22 +508,46 @@ public Consumer setOwner(Owner owner) {
...
-    @Override
-    public String toString() {
-        return String.format("Consumer [id: %s, uuid: %s, name: %s]",
-            this.getId(), this.getUuid(), this.getName());
+    /**
+     * Check if the consumers facts have changed and contains the cloud profile facts.
+     * It returns true if incoming facts are cloud profile facts
+     *
+     * @param incomingFacts incoming facts
+     * @return a boolean
```
JoshAlbrecht23, line 516: We could add a better description of the returned value here

JoshAlbrecht23 (changes requested):
Looks good!! There were some small formatting changes I noticed and a typo

src/test/java/org/candlepin/test/TestUtil.java
Ceiu: No, but the word "times" is missing. Basically, I was referencing that our implementation of all those methods are independent for some reason, rather than just being slight overloads that add more or less data (or chaining fluent methods).

This call in particular behaving differently wasn't something I wanted to deal with given the already huge amount of changes made elsewhere just to get things to work. At some point we should remove this method and have it be largely passthrough to the base util method.

src/test/java/org/candlepin/test/TestUtil.java
Ceiu: Fixed

src/test/java/org/candlepin/test/TestUtil.java
Ceiu: Fixed

src/main/java/org/candlepin/model/Consumer.java
Ceiu: Fixed.

src/test/java/org/candlepin/test/TestUtil.java
JoshAlbrecht23: Okay, that makes sense!

JoshAlbrecht23 (approved):

PR #3834, opened 2023-02-10
CANDLEPIN-121: Update servlet API dependency to latest version

wottop:
javax.servlet:servlet-api:2.5 -> javax.servlet:javax.servlet-api:4.0.1


Ceiu (changes requested):

src/main/java/org/candlepin/servlet/filter/logging/TeeHttpServletRequest.java
```diff
@@ -51,9 +52,29 @@ public TeeHttpServletRequest(HttpServletRequest request) throws IOException {
     public ServletInputStream getInputStream() throws IOException {
         final ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(body);
         ServletInputStream servletInputStream = new ServletInputStream() {
+            ReadListener readListener;
             @Override
             public int read() throws IOException {
-                return byteArrayInputStream.read();
+                int count = byteArrayInputStream.read();
```
Ceiu, line 58: The no-arg `read` actually reads a single byte from the stream. If there is no data remaining, it will return -1.

We can use that to optimize the checks a bit:
```
int output = byteArrayInputStream.read();

if (output == -1 && !this.finished) { // This will short-circuit in most cases
  this.finished = true;
  if (this.readListener != null) {
    // This event probably shouldn't occur more than once
    this.readListener.onAllDataRead();
  }
}
  
...


public boolean isFinished() {
  return this.finished;
}
```

Ceiu (changes requested):
One more quick correction, then this is good

Ceiu (approved):

PR #3838, opened 2023-02-14
CANDLEPIN-402: Update hypervisor logic to not use model objects for external API

wottop:


src/main/java/org/candlepin/service/impl/HypervisorUpdateAction.java
```diff
@@ -134,10 +136,15 @@ public Result update(
...
         boolean create, String principal, String jobReporterId) {
-        String systemUuid = incomingHost.getFact(Consumer.Facts.DMI_SYSTEM_UUID);
-        String hypervisorId = incomingHost.getHypervisorId().getHypervisorId();
+        String systemUuid = incomingHost.getFacts() == null ? null :
+            incomingHost.getFacts().get(Consumer.Facts.DMI_SYSTEM_UUID);
+        // When the data comes in via a HypervisorIdDTO the lowercasing is not applied as per
+        // the HypervisorId class method setHypervisorId. The hypervisor id lookup is the only
+        // place where it will matter as the DTO data will be put through the model object before
+        // any other operations.
+        String hypervisorId = incomingHost.getHypervisorId().getHypervisorId().toLowerCase();
```
JoshAlbrecht23, line 147: Would it be better to apply the `toLowerCase()` in the getExistingConsumerByHypervisorIdOrUuid method? I imagine that this can be more helpful so that in the future if getExistingConsumerByHypervisorIdOrUuid is used somewhere else, the implementer doesn't have to worry about applying the `toLowerCase()` to the hypervisorId value before using the method. Also, it looks like this is already done for the systemUuid value in that method.

JoshAlbrecht23 (changes requested):
Looks good! I just had a question regarding the `toLowerCase()` of the hypervisorId being used in the getExistingConsumerByHypervisorIdOrUuid method.

src/main/java/org/candlepin/service/impl/HypervisorUpdateAction.java
wottop: Good point

nikosmoum (commented):
Can we test this with virt-who, to make sure all the fields from the DTO match 1:1 with what it expects still?

JoshAlbrecht23 (approved):

wottop:
While the success of HypervisorCheckInSpecTest should be sufficient, I have manually tested with virt-who and confirmed that all the data has landed in the database.

PR #3844, opened 2023-02-16
[M] CANDLEPIN-128: Upgrade to RESTEasy 4.7.7

nikosmoum:
- We cannot upgrade to 5+ at the moment, because the Guice module was deprecated and removed at that version.


nikosmoum:
Note: We can't upgrade to 4.7.8 because of this bug: https://issues.redhat.com/browse/RESTEASY-3290

wottop:
ack

PR #3845, opened 2023-02-16
CANDLEPIN-476: Updated Pools model to avoid immutability warning

JoshAlbrecht23:
- Updated OwnerProductCurator.clearProductEntityVersion to use native query

The Hibernate warning messages that this PR addresses are the following:
1. HHH000502: The [product] property of the [org.candlepin.model.Pool] entity was modified, but it won't be updated because the property is immutable.
2. HHH000487: The query: [UPDATE Product SET entityVersion = NULL WHERE uuid = :product_uuid] attempts to update an immutable entity: [cp2_products]

The first error message is reproduced by running the spec tests in the RefreshPoolsSpecTest suite. I ran the same tests to validate that the changes cause the warning messages to no longer occur. The issue is that the annotation parameters `insertable = false` and `updatable = false` was marking the Pool.Product field as non updatable, but we are updating that field in the in refresh pools job.

I used the `RefreshPoolsSpecTest.shouldRegenerateEntsWhenModifiedProductIdsOfContentChange` spec test to remote debug the field access of the Pool's product uuid in the RefreshWorker.mapExistingPools method. I verified that the productuuid field is properly used instead of hydrating the whole product field. I also verified no additional database calls are made by `Hibernate.isInitialized` in Pools.getProductUuid by turning on the Hibernate debug logging and stepping over check while tailing the logs.

https://github.com/candlepin/candlepin/blob/f1fa1dbaef691029495ca893e42aed1d5c0b36dd/src/main/java/org/candlepin/controller/refresher/RefreshWorker.java#L394

I wasn't able to recreate the second error regarding the update statement, but looking at where the hibernate error message is thrown, changing our OwnerProductCurator.clearProductEntityVersion method to use native sql instead of hsql should resolve the issue.

https://github.com/hibernate/hibernate-orm/blob/7d30b57f15617f679a20aa1389c9385433e45b2c/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java#L428-L434

### DIFF
--- a/src/main/java/org/candlepin/model/OwnerProductCurator.java
+++ b/src/main/java/org/candlepin/model/OwnerProductCurator.java
@@ -1148,9 +1148,13 @@ public class OwnerProductCurator extends AbstractHibernateCurator<OwnerProduct> 
             return;
         }
 
+        String sql = "UPDATE " + Product.DB_TABLE + " SET entity_version = NULL " +
+            "WHERE uuid = :product_uuid";
+
         this.getEntityManager()
-            .createQuery("UPDATE Product SET entityVersion = NULL WHERE uuid = :product_uuid")
+            .createNativeQuery(sql)
             .setParameter("product_uuid", productUuid)
+            .setHint(QueryHints.NATIVE_SPACES, Product.class.getName())
             .executeUpdate();
     }
 
--- a/src/main/java/org/candlepin/model/Pool.java
+++ b/src/main/java/org/candlepin/model/Pool.java
@@ -19,6 +19,7 @@ import org.candlepin.service.model.SubscriptionInfo;
 import org.candlepin.util.DateSource;
 
 import org.apache.commons.lang3.StringUtils;
+import org.hibernate.Hibernate;
 import org.hibernate.annotations.BatchSize;
 import org.hibernate.annotations.Cascade;
 import org.hibernate.annotations.Fetch;
@@ -241,11 +242,11 @@ public class Pool extends AbstractHibernateObject<Pool> implements Owned, Named,
     @NotNull
     private Date endDate;
 
-    @Column(name = "product_uuid", nullable = false)
+    @Column(name = "product_uuid", insertable = false, updatable = false, nullable = false)
     private String productUuid;
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "product_uuid", insertable = false, updatable = false)
+    @JoinColumn(name = "product_uuid", nullable = false)
     private Product product;
 
     @ElementCollection
@@ -838,7 +839,20 @@ public class Pool extends AbstractHibernateObject<Pool> implements Owned, Named,
      *  otherwise
      */
     public String getProductUuid() {
-        return this.productUuid;
+        Product product = this.getProduct();
+
+        // If we don't have a product instance, it doesn't matter what the
+        // productUuid field contains -- we've desync'd and it's likely wrong.
+        if (product == null) {
+            return null;
+        }
+
+        // Return the productUuid if, and only if, our product instance is an
+        // uninitialized Hibernate proxy and we have a non-null UUID value.
+        // Otherwise, defer to the product itself.
+        return !Hibernate.isInitialized(product) && this.productUuid != null ?
+            this.productUuid :
+            product.getUuid();
     }
 
     /**
@@ -864,22 +878,9 @@ public class Pool extends AbstractHibernateObject<Pool> implements Owned, Named,
      */
     public Pool setProduct(Product product) {
         this.product = product;
-        this.productUuid = product != null ? product.getUuid() : null;
+        this.productUuid = null;
 
         return this;
-    }
-
-    /**
-     * Sets the product UUID from the product instance if, and only if, the product UUID is
-     * currently null, but the product is not.
-     *
-     * This method is used by the persist and update hooks to ensure the productUuid field is
-     * populated even in instances where the product was set before it was persisted.
-     */
-    private void synchronizeProductUuid() {
-        if (this.productUuid == null && this.product != null) {
-            this.productUuid = this.product.getUuid();
-        }
     }
 
     /**
@@ -1309,15 +1310,13 @@ public class Pool extends AbstractHibernateObject<Pool> implements Owned, Named,
     protected void onCreate() {
         super.onCreate();
 
-        this.synchronizeProductUuid();
         this.type = this.getType();
     }
 
     @Override
     protected void onUpdate() {
-        super.onCreate();
+        super.onUpdate();
 
-        this.synchronizeProductUuid();
         this.type = this.getType();
     }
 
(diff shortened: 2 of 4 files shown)


src/main/java/org/candlepin/model/OwnerProductCurator.java
```diff
@@ -1148,9 +1148,13 @@ public void clearProductEntityVersion(String productUuid) {
             return;
         }
 
+        String sql = "UPDATE " + Product.DB_TABLE + " SET entity_version = NULL " +
```
Ceiu, line 1151: This change also needs to occur in `OwnerContentCurator.clearContentEntityVersion`

JoshAlbrecht23: Good call! Updated `OwnerContentCurator.clearContentEntityVersion`

src/main/java/org/candlepin/model/Pool.java
```diff
@@ -241,11 +244,11 @@ public String getDescription() {
     @NotNull
     private Date endDate;
 
-    @Column(name = "product_uuid", nullable = false)
+    @Column(name = "product_uuid", insertable = false, updatable = false, nullable = false)
```
Ceiu, line 247: We should move the nullable flag to the product field we'll actually be maintaining

src/main/java/org/candlepin/model/Pool.java
```diff
@@ -1309,15 +1303,13 @@ public String toString() {
     protected void onCreate() {
         super.onCreate();
 
-        this.synchronizeProductUuid();
         this.type = this.getType();
     }
 
     @Override
     protected void onUpdate() {
         super.onCreate();
```
Ceiu, line 1311: Not your bug, but we should fix this while we're here: that should be calling `super.onUpdate()`, not onCreate as it's doing now.

src/main/java/org/candlepin/model/Pool.java
```diff
@@ -838,6 +841,10 @@ public Pool setActiveSubscription(Boolean activeSubscription) {
      *  otherwise
      */
     public String getProductUuid() {
+        if (this.productUuid == null || Hibernate.isInitialized(this.product)) {
```
Ceiu, line 844: Few things here:
- While we do have the UUID available as a field, the deferment to product makes this more of a utility method. Fetching of the product should be done through `.getProduct()` not through a direct field reference.
- Hibernate.isInitialized will return `true` on a null field, and the usage of `this.product` into `getProduct()` could lead to a NPE or otherwise inconsistent output.
- Minorist of nano-optimizations: given that our primary concern is the presence and initialization of product, that should be the first thing we check; the product_uuid field is now our secondary. Granted, I'd expect a JIT compiler to notice this and rearrange stuff as appropriate, it's still good habit to be aware of that kind of thing, and it makes the code a bit more obvious in where the priorities lie.

Given that we need to be concerned with the nullness of our product field, the checks should look something like this:

```
Product product = this.getProduct();

// If we don't have a product instance, it doesn't matter what the 
// productUuid field contains -- we've desync'd and it's likely wrong.
if (product == null) {
  return null;
}

// Return the productUuid if, and only if, our product instance is an
// uninitialized Hibernate proxy and we have a non-null UUID value.
// Otherwise, defer to the product itself.
return !Hibernate.isInitialized(product) && this.productUuid != null ?
  this.productUuid :
  product.getUuid();
```

src/main/java/org/candlepin/model/Pool.java
```diff
@@ -864,24 +871,11 @@ public Product getProduct() {
      */
     public Pool setProduct(Product product) {
         this.product = product;
-        this.productUuid = product != null ? product.getUuid() : null;
+        this.productUuid = null;
```
Ceiu, line 874: While this is correct behavior imo, I wonder if this will trigger the exact warning we're aiming to address.

Ceiu (changes requested):

src/main/java/org/candlepin/model/Pool.java
JoshAlbrecht23: Yeah, I would expect to get that warning as well. However, I ran the changes against the RefreshPoolsSpecTest tests and verified we are hitting the setProduct method, but we are not getting the warning in the logs. I don't fully understand why that is.

src/main/java/org/candlepin/model/Pool.java
JoshAlbrecht23: Oh haha I looked at the multiple times and didn't realize that was an `onCreate` method being used in the `onUpdate` method. Updated

src/main/java/org/candlepin/model/Pool.java
JoshAlbrecht23: Good point! Moved that back

src/main/java/org/candlepin/model/Pool.java
JoshAlbrecht23: This makes sense! I also added those comments as they will be very helpful for others looking at this code.

src/main/java/org/candlepin/model/Pool.java
Ceiu: ack

src/test/java/org/candlepin/model/PoolTest.java
```diff
@@ -295,4 +303,128 @@ public void testIsNotDerivedPool() {
...
+        @Test
+        public void testGetProductUuidWithDesyncedProduct() throws Exception {
+            Owner owner = this.createOwner();
+            Product product = this.createProduct(TestUtil.randomString("id-"),
+                TestUtil.randomString("name-"),
+                owner);
+            Pool pool = this.createPool(owner, product);
+
+            // We don't want to persist the upcoming changes to the pool, so commit now
+            this.getCurrentSession().getTransaction().commit();
```
Ceiu, line 320: You can use `this.commitTransaction()` here, utilizing the declaration from `DatabaseTestFixture` instead of having to fetch the session manually.

Also, in general, we should use JPA interfaces over Hibernate interfaces wherever reasonably possible.

src/test/java/org/candlepin/model/PoolTest.java
```diff
@@ -295,4 +303,128 @@ public void testIsNotDerivedPool() {
...
+            this.getCurrentSession().clear();
```
Ceiu, line 321: You can use `this.getEntityManager().clear()` instead, or the `.clear()` method in any curator -- many of which are provided by the test fixture. For instance, `poolCurator.clear()`

Ceiu (changes requested):

src/test/java/org/candlepin/model/PoolTest.java
JoshAlbrecht23: Updated! Also, it is good to know to use the JPA interfaces over the Hibernate ones. I was honestly not sure which one to use.

src/test/java/org/candlepin/model/PoolTest.java
JoshAlbrecht23: Updated

Ceiu:
Is this ready for review/merge now?

Ceiu (approved):

PR #3850, opened 2023-02-20
CANDLEPIN-443: drop obsolete tables

wottop:
- cp2_activation_key_products
- cp2_environment_content tables

### DIFF
--- a/src/main/resources/db/changelog/20230220150433-drop-act-key-env-tables.xml
+++ b/src/main/resources/db/changelog/20230220150433-drop-act-key-env-tables.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <changeSet id="20230220150433-1" author="wpoteat">
+        <preConditions onFail="MARK_RAN">
+            <tableExists tableName="cp2_activation_key_products"/>
+        </preConditions>
+        <dropTable tableName="cp2_activation_key_products"/>
+    </changeSet>
+
+    <changeSet id="20230220150433-2" author="wpoteat">
+        <preConditions onFail="MARK_RAN">
+            <tableExists tableName="cp2_environment_content"/>
+        </preConditions>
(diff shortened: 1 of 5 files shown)


src/main/resources/db/changelog/20230220150433-drop-act-key-env-tables.xml
```diff
@@ -0,0 +1,20 @@
...
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="20230220150433-1" author="wpoteat">
+        <comment>
+            Remove old activation key and environment tables
+        </comment>
+
+        <sql>
+            DROP TABLE cp2_activation_key_products;
```
Januson, line 15: I guess it doesn't matter much in this case, but generally it might be better to use liquibase operations like `<dropTable>` rather than sql.

src/main/resources/db/changelog/20230220150433-drop-act-key-env-tables.xml
```diff
@@ -0,0 +1,20 @@
...
+            DROP TABLE cp2_environment_content;
```
Januson, line 16: It might be a good idea to add preconditions to make sure the current state of the db is what we expect.

Januson (changes requested):

src/main/resources/db/changelog/20230220150433-drop-act-key-env-tables.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
```
nikosmoum, line 7: We're currently on liquibase 4.19, so let's link to the `dbchangelog-4.19.xsd` and make it a habit to use that from now on. It will help with validation if/whenever we might use newer liquibase xml tags

src/main/resources/db/changelog/20230220150433-drop-act-key-env-tables.xml
Ceiu: +1 to this

nikosmoum (approved):

nikosmoum:
Looks like a unit test related to the xsd version failed and needs an update

Januson (approved):

PR #3862, opened 2023-02-22
CANDLEPIN-497: Remove Ruby setup from ansible and add Gemfiles back

wottop:
The best way to test the docker image build is to apply this diff to the branch:
http://pastebin.test.redhat.com/1093124
You have to copy the Gemfile and Gemfile.lock to the docker/ directory before building images locally.
The real image building only happens after it is all in master.


docker/candlepin-base-cs8/setup-devel-env.sh
```diff
@@ -61,23 +61,6 @@ BASHRC
 git clone https://github.com/candlepin/candlepin.git /candlepin
 cd /candlepin
 
-# Setup and install rvm, ruby and pals
```
nikosmoum, line 64: We can't remove ruby from our c7 and cs8 containers. We need it to run rspec in older branches.

ansible/roles/candlepin/README.md
```diff
@@ -13,14 +13,6 @@ Candlepin has many requirements for its developers. You will need a Fedora or Ce
 enabled to use this role. It is recommmended that that box has a non-root user to run Candlepin on (see
 variables below).
 
-Dependencies
-------------
-
-This role depends on the `rvm1.ruby` role for installing and configuring RVM. At the time of writing, Candlpin
```
nikosmoum, line 19: we also need to remove the reference of the role from `ansible/requirements.yml`

nikosmoum (changes requested):
ACK on the removal of ruby from ansible scripts
NACK on its removal from the containers. We still need them there to run rspec on older supported branches.

nikosmoum:
@wottop If the changes in the pastebin need are needed (copying of the Gemfiles), should they not be applied on this PR?

wottop:
> @wottop If the changes in the pastebin need are needed (copying of the Gemfiles), should they not be applied on this PR?

That is the workaround for the fact that a local build will retrieve the master branch to build the image.

nikosmoum (approved):

PR #3863, opened 2023-02-23
2172538: Fix setting of metadata expiration during import

Januson:


src/test/java/org/candlepin/sync/ProductImporterTest.java
```diff
@@ -409,4 +317,170 @@ public void testImportProductMapReturnsEmptyMapWithNoData() throws IOException {
         assertTrue(output.isEmpty());
     }
 
+    @Test
+    public void importShouldSetMetadataExpireTo1() throws Exception {
```
nikosmoum, line 321: Can we add a spec test that checks all content (from all 4 types of product) imported has this set to 1? Unit tests are good, but for business logic like this a unit test can get lost/broken during refactoring very easily

nikosmoum (changes requested):
Just need to add a spec test for the specific scenario.
Also, let's amend the commit msg to start with the cloned 4.1 BZ id to make sure automation can update bugzillas during release

src/test/java/org/candlepin/sync/ProductImporterTest.java
Januson: I added spec test to master.

nikosmoum (approved):

PR #3869, opened 2023-02-28
[4.1] Revert accidental openapi maven plugin upgrade

nikosmoum:


nikosmoum:
Reason for reverting is that while running a release, the openapi maven plugin hit this bug: https://github.com/OpenAPITools/openapi-generator/issues/11573 which exists in version 5.4.0, but not before.

Ceiu (approved):

PR #3875, opened 2023-03-01
CANDLEPIN-509: [M] Link to source cp-test

wottop:
This allows the cp-test file in the branch to be used instead of one baked into
 the docker image.

to test, build locally and then run the test command against it. See https://github.com/candlepin/candlepin/pull/3876 if you need the full instructions.

Note that once this makes it into the images on quay.io, none of the branches prior to 4.2 will run CI correctly until they have the change in PR https://github.com/candlepin/candlepin/pull/3876


wottop:
Might want to wait for the Centos 9 branch to get merged so that this change can land there too.

Januson (approved):

PR #3878, opened 2023-03-01
CANDLEPIN-512: Update org_migrator for schema changes

wottop:


bin/scripts/org_migrator/org_migrator.py
```diff
@@ -455,7 +455,7 @@ def do_export(self):
 
         content_uuid_query1 = 'SELECT oc.content_uuid FROM cp2_owner_content oc WHERE oc.owner_id = %s' + \
             'UNION ' + \
-            'SELECT ec.content_uuid FROM cp2_environment_content ec JOIN cp_environment e on e.id = ec.environment_id WHERE e.owner_id = %s'
+            'SELECT ec.content_id FROM cp_environment_content ec JOIN cp_environment e on e.id = ec.environment_id WHERE e.owner_id = %s'
```
Ceiu, line 458: This chunk here is for pulling all content UUIDs to ensure we have all the referenced content. With the env content no longer being a hard ref to a specific content instance, we no longer need this query here.

bin/scripts/org_migrator/org_migrator.py
```diff
@@ -211,7 +211,7 @@ def fetch_product_uuids(db, org_id):
         'UNION ' + \
         'SELECT pool.product_uuid FROM cp_pool pool WHERE pool.owner_id = %s ' + \
         'UNION ' + \
-        'SELECT akp.product_uuid FROM cp_activation_key ak JOIN cp2_activation_key_products akp ON akp.key_id = ak.id WHERE ak.owner_id = %s'
+        'SELECT akp.product_id FROM cp_activation_key ak JOIN cp_activation_key_products akp ON akp.key_id = ak.id WHERE ak.owner_id = %s'
```
Ceiu, line 214: Like the content lookup, this is to ensure we have all the referenced products by UUID. Injecting product IDs into the mix won't work. This needs to be removed, and, potentially, changed to a lookup that fetches the product by its ID instead.

Ceiu (changes requested):

wottop:
In going back over this I noticed something about how we are collecting the uuids for products.

Pulling product uuids from the cp_pool table adds them to the collection if they are not already in the cp2_owner_products table. This will complete the connection in the destination db, correct?

It appears that the original queries were also achieving the same outcome, so I duplicated it here.

My question now is, are we trying to replicate the data AND fix it on the fly?

bin/scripts/org_migrator/org_migrator.py
```diff
@@ -455,7 +456,8 @@ def do_export(self):
 
         content_uuid_query1 = 'SELECT oc.content_uuid FROM cp2_owner_content oc WHERE oc.owner_id = %s' + \
             'UNION ' + \
-            'SELECT ec.content_uuid FROM cp2_environment_content ec JOIN cp_environment e on e.id = ec.environment_id WHERE e.owner_id = %s'
+            'SELECT cont.uuid FROM cp_environment e JOIN cp_environment_content ec ON e.id = ec.environment_id  ' + \
```
Ceiu, line 459: We still don't want to do this query at all anymore -- It's no longer required for pulling the data, and this query doesn't guarantee it'll pull the same thing on every export.

Same with the product version

Ceiu (changes requested):

Ceiu:
> In going back over this I noticed something about how we are collecting the uuids for products.

> Pulling product uuids from the cp_pool table adds them to the collection if they are not already in the cp2_owner_products table. This will complete the connection in the destination db, correct?

It will not. It simply pulls the product row that is referenced. It does not add an additional row in cp2_owner_products (or cp2_owner_content). This is required for two reasons: first and foremost, in the event we had an environment or activation key pointing to an out-of-org product or content, that we'll be able to import the org dump without needing to explicitly track down which org, if any, owns it and export them as well. Secondly, in the case of an operation misbehaving, we want to be sure we pull the exact data a given org is working with so we can reproduce any issues.

> It appears that the original queries were also achieving the same outcome, so I duplicated it here.

> My question now is, are we trying to replicate the data AND fix it on the fly?

The org migrator should not be performing any manipulation of the data at all, nor should it be fixing anything it sees. It simply exports the data as it exists -- no matter how broken it may be -- so it can be imported locally for reproduction.

The original queries were fetching the product and content references off environments and activation keys because they had to. Referential integrity forced those explicit references to exist. Now that we've reverted back to soft references via ID, we no longer need the product to be pulled if it hasn't been already.

bin/scripts/org_migrator/org_migrator.py
```diff
@@ -513,7 +508,7 @@ def do_export(self):
         # then recursively sift through the products we've pulled to check for children products
         #
         # Product references:
-        #   - cp2_activation_key_products.product_uuid
+        #   - cp_activation_key_products.product_id
```
Ceiu, line 511: This comment line is obsolete, and no longer needed

PR #3882, opened 2023-03-03
CANDLEPIN-513: Update usages of Docker to Podman

wottop:


nikosmoum (commented):
Just a general note that this task is dependent on https://github.com/candlepin/candlepin/pull/3721, and a lot of the changes here won't be relevant (e.g. removed files). Also, we're not moving from docker-compose to podman-compose, but to podman direct commands (podman pod create ,etc.), so the README changes will have to reflect that

wottop:
ok. I'll update as needed when the other is done.

nikosmoum:
Amongst other things, the new cp-test symlink creation in Dockerfiles will be updated too (candlepin-dev/docker/cp-test)

PR #3892, opened 2023-03-09
[M] RPM spec file cleanup

nikosmoum:
- Remove conditional statements used for RHEL 7 or earlier since now candlepin supports RHEL 8 as the minimum.
- Remove long unneeded obsoletes directives.
- Don't depend on tomcatjss, but instead directly on pki-servlet-engine and jss. The tomcatjss dependency was a hack to work around the renaming of the tomcat package from pki-servlet-container to pki-servlet-engine on RHEL 8.1.

### DIFF
--- a/candlepin.spec.tmpl
+++ b/candlepin.spec.tmpl
@@ -26,25 +26,21 @@ License: GPLv2
 Source0: %{name}-%{version}-complete.tar.gz
 Source1: %{name}-%{version}.war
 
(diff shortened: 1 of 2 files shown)


nikosmoum:
Note: I've tested the RPM build, install and deploy of candlepin on Fedora 37, RHEL8 and RHEL9

JoshAlbrecht23 (approved):

PR #3898, opened 2023-03-14
CANDLEPIN-469: Require compiling and running with Java 17

Januson:


nikosmoum (approved):
Changes look simple enough. Just a question if rpm build/install/deploy was tested with both RHEL8/9?

Januson:
@nikosmoum Yes. I tested on both RHEL 8 and 9.

PR #3899, opened 2023-03-14
CANDLEPIN-471: Compile and run Candlepin with Java 17

Januson:

PR #3905, opened 2023-03-15
CANDLEPIN-532: Update build target for Java 17

wottop:
We need to make sure the target is ready to go before merging.


nikosmoum (changes requested):
One thing this PR also needs is the change maven change in @Januson's PR here: https://github.com/candlepin/candlepin/pull/3899/files#diff-dbd40a3d6bbf2a0e4b6b199c778f60d3bf42218ba7a1b47bf145ecbe02c48c67
and then doing a `./gradlew pom` to generate the pom.xml with the new compiler plugin config to target java 17

nikosmoum (approved):

PR #3911, opened 2023-03-17
Fix vagrant up

Januson:
- Switched to generic/centos8s box. The box we currently use, which got it's last update two years ago seems no longer updatable.
- Updated dnf update task to use dnf module instead of command.

### DIFF
--- a/ansible/roles/candlepin/tasks/el8/el8_base_deps.yml
+++ b/ansible/roles/candlepin/tasks/el8/el8_base_deps.yml
@@ -7,12 +7,12 @@
     - system
     - candlepin
 
-# FIXME: Update this task to use the dnf module once Ansible 2.10 or newer is readily available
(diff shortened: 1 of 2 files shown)


ansible/roles/candlepin/tasks/el8/el8_base_deps.yml
```diff
@@ -7,12 +7,12 @@
...
-# FIXME: Update this task to use the dnf module once Ansible 2.10 or newer is readily available
 - name: "Update system packages"
   become: true
-  ansible.builtin.command:
-    cmd: dnf update -y --allowerasing
-  changed_when: false
+  ansible.builtin.dnf:
+    name: "*"
+    state: latest
+    allowerasing: true
```
Ceiu, line 15: :+1:

PR #3931, opened 2023-03-23
CANDLEPIN-536: Update JSON time serializer for ISO-8601

wottop:
Update JSON time serializer to follow more strict ISO-8601 derivatives
- When we specify the format as date-time for a given field, the OpenAPI spec
 expects it to be an RFC-3339 date-time, which while close to the basic
 ISO-8601 format, does not permit omitting the colon in the offset.
- This is handled in code with the DateTimeFormatter.ISO_OFFSET_DATE_TIME
 which claims to be ISO-8601 lenient, but still has a requirement of the
 colon in the zone offset.

Ensure Status output has proper UTC time
- Previously we relied on the value to come from a system set to
  UTC already. This will ensure it is in UTC regardless of the system
  setting.

### DIFF
--- a/src/main/java/org/candlepin/jackson/OffsetDateTimeSerializer.java
+++ b/src/main/java/org/candlepin/jackson/OffsetDateTimeSerializer.java
@@ -36,7 +36,7 @@ public class OffsetDateTimeSerializer extends JsonSerializer<OffsetDateTime> {
 
     public OffsetDateTimeSerializer() {
         // This DateTimeFormatter pattern is ISO 8601 without milliseconds
-        this.dateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ssZ");
+        this.dateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ssxxx");
     }
 
     @Override
--- a/src/main/java/org/candlepin/resource/StatusResource.java
+++ b/src/main/java/org/candlepin/resource/StatusResource.java
@@ -37,6 +37,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.time.OffsetDateTime;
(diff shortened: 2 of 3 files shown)


Ceiu:
While this patch fixes the immediate problem and corrects the output of the status endpoint, I'm not sure it fixes the underlying root cause of the parsing error. Do we know why non-UTC timestamps were causing problems? Are we confident that this won't occur when parsing any other timestamps we might return?

wottop:
We seem to be at the mercy of a generated class in the client:
client/build/generated/src/main/java/org/candlepin/invoker/client/JSON.java
it only knows how to deal with dates ending with "+0000"

Ceiu:
The comments in the generated code have the answer here:
When we specify the format as `date-time` for a given field, the OpenAPI spec expects it to be an RFC-3339 date-time, which while close to the basic ISO-8601 format, does not permit omitting the colon in the offset. Amusingly, this is handled in code with the `DateTimeFormatter.ISO_OFFSET_DATE_TIME` which claims to be ISO-8601 lenient, but still has a requirement of the colon in the zone offset (see: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/ZoneOffset.html#getId())

We can fix this in our code by adding the colon in the OffsetDateTimeSerializer and ensuring we're following the more strict ISO-8601 derivatives. Since we're still staying within spec by adding it, we shouldn't break any existing clients.

Ceiu:
We should also add these details as a commit comment

Ceiu (changes requested):
See previous comments for change request

Ceiu (changes requested):
These look good, but the change to ensure the timeUTC field is always in UTC is still something we should do. So that should either be brought back in this PR, added as another commit on this PR, or another PR entirely.

Also the commit message should be formatted closer to our guidelines for that.

Ceiu (approved):

PR #3932, opened 2023-03-27
CANDLEPIN-541: Spec test improvements for Jenkins runs

wottop:
- Extended the network timeout in the gradle-wrapper.properties to
  fix the issue of retrieving the gradle zip and Java dependencies
    
- Updated the build.gradle for the spec tests to force the use of
  UTF-8 encoding for compiling


wottop:
I will run this for a couple of days to see if I can get it to fail with the new timeout setting.

Januson:
@wottop Can you configure this for the download of dependencies as well? They were also failing with this issue.

PR #3937, opened 2023-03-30
CANDLEPIN-544: Run liquibase update on startup [POC]

wottop:
- Uses the existing lookup for unrun liquibase change sets to determine the need for executing the update.

### DIFF
--- a/config/candlepin/candlepin.conf.template
+++ b/config/candlepin/candlepin.conf.template
@@ -34,7 +34,7 @@ candlepin.pretty_print=true
 candlepin.async.scheduler.enabled=<%= candlepin.async_scheduler_enabled %>
 
 candlepin.hidden_resources=
-candlepin.db.halt_on_liquibase_desync=true
+candlepin.db.database_manage_on_startup=Manage
 
 <% if (candlepin.hidden_capabilities) { %>\
 candlepin.hidden_capabilities=<%= candlepin.hidden_capabilities %>
--- a/bin/deployment/deploy
+++ b/bin/deployment/deploy
@@ -26,7 +26,6 @@ gendb() {
 
     if [ "$GENDB" == "1" ]; then
         MESSAGE="Generating New Database"
-        CHANGELOG="changelog-create.xml"
         if [ "$USE_MYSQL" == "1" ]; then
             recreate_mysql $APP_DB_NAME $DBUSER
         else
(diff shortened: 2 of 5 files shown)


config/candlepin/candlepin.conf.template
```diff
@@ -35,6 +35,7 @@ candlepin.async.scheduler.enabled=<%= candlepin.async_scheduler_enabled %>
 
 candlepin.hidden_resources=
 candlepin.db.halt_on_liquibase_desync=true
+candlepin.db.liqubase_update_on_start=false
```
nikosmoum, line 38: this template is only used for candlepin.conf on dev deployments, so why not set to true?

nikosmoum: and at the same time, stop running liquibase.sh in the deploy script

wottop: Do you mean stop running the update in the deploy script and keep the create? Is this the point where we use the -create.xml only regardless of create/update and only use the on start method in this PR?

Ceiu: I'm not sure we want two separate configs for this. That could work, but it ends up having weird or non-functional pairs [1].

Given the last thing we added in this area, and what we're looking at here, a single configuration with multiple values similar to a logging level seems apt. I.e.
`candlepin.db.maintenance = (NONE|CHECK|MANAGE)`. This would allow indicating that the system admin wants to fully deal with everything themselves (NONE), wants to be warned if a desync occurs (CHECK), or wants to have Candlepin fully manage its own DB (MANAGE).

(Note that the names here are all for example purposes, not a suggestion of actual usage)

Ideally we'll convert this to a design so stuff like this can get hammered out before final implementation.

[1]
(false, false) = no maintenance at all; equiv to NONE above
(false, true) = update DB, but do not warn if it's desync'd; DB update failure acts as a desync warning; equiv to MANAGE above
(true, false) = do not update DB, but warn if it appears desync'd; equiv to CHECK above
(true, true) = update DB and warn if it's desync'd; loosely equiv to MANAGE, but with redundancy on the check option

Januson: One thing to consider is compatibility with other frameworks.  If we use separate properties we can map liquibase update to whichever framework we'll end up using. As far as I know just checking if the db is up to date is not supported by default by Quarkus nor Spring so we'll have to keep doing that one ourselves.

wottop:
Link to design document:

https://docs.google.com/document/d/16FE5rfFmy1Gcwnr3tyJTZp_BqszK01ZF4ukXrYraUqI

ekohl:
FYI: the document is only readable for Red Hat employees.

If I read it correctly, then the goal is to have a single "migrate" step that replaces both create and update. This can even run on application start up, meaning that as long as the DB exists and the app has valid credentials configured, you can just run the service. No more cpdb needed.

Speaking from the Foreman/Katello installer (and implicitly Red Hat Satellite), I can say that it would make things easier. We don't have clustered deployments, so no need for us to synchronize state between multiple servers.

nikosmoum (changes requested):
Somehow the 'CANDLEPIN-494: Update resteasy packages' commit snuck/merged its way into this commit?

Ceiu:
PoC has cleared review. Next steps will be formalizing the code and adding tests. Thanks for the demo on this!

PR #3940, opened 2023-04-04
CANDLEPIN-522: Move spec test execution outside of container

bakajstep:


containers/cp-test
```diff
@@ -357,79 +180,15 @@ if [ -f /usr/bin/supervisord ]; then
...
     CLEAN_CP=1
 
     DEPLOY_FLAGS="-g"
 
-    if [ "$SPEC_TESTS" == "1" ] && [ "$HOSTED" == "1" ] \
-        && [ "$CP_NEW_STRUCTURE" == false ] && ($PROJECT_DIR/bin/deploy '-?' | grep -q -- '-H'); then
-        DEPLOY_FLAGS="$DEPLOY_FLAGS -H -a"
-    elif [ "$SPEC_TESTS" == "1" ] && [ "$HOSTED" == "1" ] \
-        && [ "$CP_NEW_STRUCTURE" == true ] && ($PROJECT_DIR/bin/deployment/deploy '-?' | grep -q -- '-H'); then
+    if [ "$HOSTED" == "1" ] && ($PROJECT_DIR/bin/deployment/deploy '-?' | grep -q -- '-H'); then
```
JoshAlbrecht23, line 191: I don't think this is working properly. Looking at the Jenkins logs for this PR we can see that the message `This deployment is not for HOSTED` is displayed in the `sh jenkins/spec-test.sh` step for hosted spec tests. Also, if you look at the spec tests that are being run, you can see that only the standalone tests are being run because Candlepin is set to standalone mode.

JoshAlbrecht23 (changes requested):

containers/cp-test
bakajstep: Oh, i missed that. done

Januson (approved):

PR #3943, opened 2023-04-06
CANDLEPIN-529: Add Conscious Language Initiative guidelines

wottop:


nikosmoum (approved):
Merging this regardless of jenkins failure because it is unrelated (this is just a markdown change)

PR #3945, opened 2023-04-10
CANDLEPIN-530: Added Conscious Language Initiative check to pipeline

wottop:
The description on rule changes here:
https://github.com/get-woke/woke/blob/main/docs/rules.md

Listing of the default rules:
https://github.com/get-woke/woke/blob/main/pkg/rule/default.yaml

### DIFF
--- a/.github/workflows/woke.yaml
+++ b/.github/workflows/woke.yaml
@@ -1,0 +1,16 @@
+name: woke
+on:
+  - pull_request
+jobs:
+  woke:
+    name: woke
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: woke
+        uses: get-woke/woke-action@v0
+        with:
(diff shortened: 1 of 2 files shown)


Ceiu (changes requested):
I have no love for this kind of change, but it seems to achieve its intended goal well enough.

However, one thing I'm a bit curious on: our configuration looks to be inheriting the default rules and then overriding most of them to do nothing. Given that we have our own set of off-limits terms based on a different set of criteria and reasoning, it feels like we'd be better off not using the default at all and having to deal with any changes they make or decide to add/remove from time to time. We could just as easily copy/paste the existing rules we care about, and then rather than regurgitate misguided reasoning from a twitter post as the justification for a rule, we could, instead, reference the Red Hat initiatives with a link to them for those interested in further details.

Ceiu:
Let's update the verbiage on the commit message to be past-tense, and this is good to go.

PR #3947, opened 2023-04-10
Test sleep time for spec tests on Jenkins

wottop:


containers/test
```diff
@@ -125,7 +125,7 @@ fi
 
 
 podman pod start $PROJ_NAME
-sleep 5 # Waiting to start systemd
+sleep 60 # Waiting to start systemd
```
bakajstep, line 128: i dont think this is a problem. I take this time from cockpit folks and they dont have problem with it. 

https://github.com/cockpit-project/bots/blob/78ba58fed66ec88d0298a4271f77ce0b6251fccf/images/scripts/services.setup#L148

PR #3950, opened 2023-04-13
[M] CANDLEPIN-557: Added support for entitlement cleanup in SCA mode

Ceiu:
- Added an option to enable revoking consumer entitlements during an
  entitlement certificate regeneration operation when the consumer is
  running in SCA mode. This feature is enabled by default, but can be
  disabled by setting the configuration
  "candlepin.sca_entitlement.cleanup" to "false".
- Miscellaneous comment, logging, and documentation cleanup

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -3179,7 +3179,13 @@ paths:
 
   /environments/{env_id}/content:
     post:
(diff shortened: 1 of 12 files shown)


Januson (approved):
Changes look good, but tests are failing.

Ceiu:
I've fixed most of the test failures, but the more I address, the more this feature feels *wrong*. Do we have a spec for how SCA should behave with regards to import/export wrt how SCA has evolved over the years? I'm not confident in the idea of just throwing entitlements out on refresh. I feel like there are side effects that the RFE isn't considering or accounting for here.

Going to convert this to a draft while I try to track down more information on both sides of the equation here.

PR #3952, opened 2023-04-14
CANDLEPIN-501: Migrate checkstyle to GH Actions

Januson:
- Created checkstyle job
- Created workflow for building CI images
- Moved woke job into PR workflow

@Ceiu @wottop @JoshAlbrecht23 @nikosmoum @bakajstep : We have a choice here. Our build requires gettext and jss installed. We can either periodically build a runner image on which our build executes or we can install them as a job execution step. Doing so in the job adds about ~60 seconds to the job run time, but saves us from managing the runner images and keeps job/branches isolated. What do you think? I included both options in the PR as checkstyle and checkstyle_img jobs.


ptoscano:
Humble note from a subscription-manager person: we recently switched exclusively to GitHub Actions, and so far we are quite happy with them. We did not really have container images for testing, so for us the choice was simple: each job installs whatever it needs. While this seems to add maintenance, in reality it makes things easier for us:
- each job installs whatever is needed, so jobs that need to do only few/easy things installs the few packages needed
- each job is fully independent, it does not need anything else than the base container image of the OS used for testing
  - this also means that, in case you want to change anything in the way tests/jobs run (for example install an extra package), you can simply do the changes in each PR, and you see them reflected immediately on the jobs
- no need to maintain other artifacts

The only real downside is that it adds some time to the job execution; overall, it isn't an extra burden for us (the runtime of each of our jobs is still way below 5 minutes).

One more extra suggestion I can provide is to install packages without weak dependencies (recommends), so you don't rely on packages you need installed "by dependency" of others; in practice, it means calling `dnf` with `--setopt install_weak_deps=False`.

You can find the jobs of subscription-manager here: https://github.com/candlepin/subscription-manager/tree/main/.github/workflows

.github/workflows/pr_verification.yml
```diff
@@ -0,0 +1,52 @@
+---
+name: PR verification workflow
+
+on:
+  pull_request:
+    branches:
+      - main
```
Ceiu, line 7: We'd want this to run for any PR, not just those targeting main, right? I know it's rare for our project, but in the past we had rather large features get PRs into their own feature branch. Plus we have the individual CP maintenance branches.

.github/workflows/pr_verification.yml
```diff
@@ -0,0 +1,52 @@
...
+  workflow_dispatch:
+
+jobs:
+  checkstyle:
```
Ceiu, line 11: Looking at the two options and giving it a couple days to stew on it, I think I'm ultimately in favor of this method here where we build everything on a given invocation of the job.

The extra time isn't significant for us, as our other tests are going to dominate the overall runtime by an order of magnitude or two, and addressing problems on a per-run basis seems like it'll be easier than having to go through the mental jogging of remember where the image is and rebuilding it in cases where that is the resolution.

Ideally we can arrive at a place in the future where we're no longer tied to dependencies that have to be installed at the system level to run our various tasks, eliminating the need to make considerations and trade-offs like this.

.github/workflows/pr_verification.yml
Januson: Good point. Removed branch filter.

.github/workflows/pr_verification.yml
Januson: I agree. Extra time isn't that bad and it will save us a lot of maintenance overhead.

JoshAlbrecht23: I'm also in agreement here. The maintenance overhead for the runner image doesn't seem worth the benefit of saving a small amount of time! 

If I understand the cache action properly, we could also cache these dependencies so that they are only downloaded once and then pulled from the cache for each job [0].  However, this adds more complexity to our jobs for not much gain.

[0] https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

Januson: @JoshAlbrecht23 Caching should be handled automatically by the gradle action. https://github.com/gradle/gradle-build-action

JoshAlbrecht23: Perfect!! Thanks for linking that.

.github/workflows/pr_verification.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:37
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        shell: bash
+        run: sudo dnf install -y gettext jss
```
ptoscano, line 18: - `sudo` is not needed here, you are already "root" in the container
- pass also `--setopt install_weak_deps=False` to `dnf` to avoid installing weak dependencies (ie recommends) by default; right now it won't make a difference, however it is more future-proof
```suggestion
        run: dnf --setopt install_weak_deps=False install -y gettext jss
```

Januson: Thanks. Done.

.github/workflows/pr_verification.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        shell: bash
+        run: dnf --setopt install_weak_deps=False install -y gettext jss
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+      - name: Setup Gradle
```
Ceiu, line 23: I think we may be able to roll the setup and run actions into one action:

```
- name: Run checkstyle via Gradle
        with:
          arguments: checkstyle
```

PR #3954, opened 2023-04-17
[M] CANDLEPIN-557: Added support for entitlement cleanup in SCA mode

Ceiu:
- Added a new query parameter, cleanup_entitlements, to the consumer entitlement certificate regeneration endpoint. When set, if the consumer is operating in SCA mode, unused entitlements will be removed before certificate regeneration occurs
- Miscellaneous comment, logging, and documentation cleanup

### DIFF
--- a/src/main/java/org/candlepin/controller/ContentAccessManager.java
+++ b/src/main/java/org/candlepin/controller/ContentAccessManager.java
@@ -67,6 +67,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Predicate;
 
 import javax.inject.Inject;
 import javax.naming.ldap.Rdn;
@@ -85,7 +86,10 @@ public class ContentAccessManager {
      * utility methods for resolving, matching, and converting to database-compatible values.
      */
     public enum ContentAccessMode {
+        /** traditional entitlement mode requiring clients to consume pools to access content */
         ENTITLEMENT,
+
+        /** simple content access (SCA) mode; clients have access to all org content by default */
         ORG_ENVIRONMENT;
 
         /**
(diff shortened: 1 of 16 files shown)


src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -715,9 +759,52 @@ else if (listUpdated) {
...
+     * @param updated
+     *  the updated content access mode name
+     *
+     * @param target
+     *  the targeted content access mode to check
+     *
+     * @return
+     *  true if the content access mode is transitioning to the target mode; false otherwise
+     */
+    private boolean isTransitioningTo(String current, String updated, ContentAccessMode target) {
```
JoshAlbrecht23, line 779: This method doesn't appear to be used, just wanted to double check and see if you meant to include it. I can see that it could be helpful for future implementations where we need to check that we are updating to the target ContentAccessMode.

JoshAlbrecht23 (commented):
Looks good! The cleanup_entitlements query parameter on the /consumers/{consumer_uuid}/certificates is a good and flexible way of a adding this functionality. I just had a couple of small questions.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2609,17 +2602,53 @@ public void unbindByPool(@Verify(Consumer.class) String consumerUuid, String poo
...
+        // revoke all entitlements. They won't be needing them anymore, apparently.
+        try {
+            ContentAccessMode caMode = ContentAccessManager.resolveContentAccessMode(consumer);
+            if (caMode != ContentAccessMode.ORG_ENVIRONMENT) {
+                cleanupEntitlements = false;
+            }
+        }
+        catch (IllegalStateException e) {
+            // Consumer or org have an invald mode set. Assume it's not SCA.
+            cleanupEntitlements = false;
```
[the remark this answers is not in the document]
Ceiu: Yeah, that's a good point. This whole thing could use some logging given how serious and irreversible it all is.

src/main/java/org/candlepin/controller/ContentAccessManager.java
Ceiu: Oh, yup, it ended up not being used. The first iteration of this revoked all entitlements on the changeover to SCA, but performance testing proved that to be non-viable in many cases so it got removed. The refactoring of the original method and the new companion method were left over as a result of that.

I'll remove the unused method but leave the other refactor in place.

Ceiu: done

src/main/java/org/candlepin/resource/ConsumerResource.java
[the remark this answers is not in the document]
Ceiu: done

JoshAlbrecht23 (approved):

PR #3958, opened 2023-04-19
[M] CANDLEPIN-510: Added unit tests to the Github actions

Ceiu:
- Unit tests will now run in Github actions along with the other PR verification jobs

### DIFF
--- a/.github/workflows/pr_verification.yml
+++ b/.github/workflows/pr_verification.yml
@@ -1,5 +1,5 @@
 ---
-name: PR verification workflow
+name: Pull request verification
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


Ceiu:
This largely follows the design set forth by #3952 to migrate checkstyle to Github actions, and is blocked by it (as it copies its two jobs).

I spent a bunch of time looking into options for creating the container a single time as a one-off and running it, but none of them were very good for maintenance and/or usability. We're currently left with very few options: using a custom, self-hosted runner, creating a container periodically and using that as the basis for our job containers as Ondrej demonstrated, or abusing GH cache/artifacts to move an image file around and manually invoking everything with `docker run` instructions with custom commands per job.

Ultimately, the only reason we even need a container is because of two system-level dependencies that, honestly, are kind of a pain to depend on in this way at the project level and need to go away anyway. Rather than spend time and add complexity working around that, I ended up arriving at the same conclusion I did when looking at Ondrej's PR in that the better answer is to just deal with the extra 30-60s of setup time overhead to keep the individual jobs nice and easy to manage, and then work toward moving to better dependencies that we can ship or deliver via Maven/Gradle repos and entirely eliminate the need for a nested container for these jobs.

We'll still need to sort out some kind of container build step for the spec tests once we get Candlepin properly containerized, but we may have better options there using GH action's service containers.

JoshAlbrecht23 (approved):
I double checked that the checkstyle job ran in Github actions and it looks good. Thanks for standardizing the step names!

PR #3971, opened 2023-04-25
Fixed i18n puzzles in few error messages

ptoscano:
Joining two parts of a message to make a single translatable message is
a bad idea:
- the message will be harder to translate
- in some cases, parts were even not marked as translatable (so the
  resulting message always resulted in "half-English" text)
             
Hence, to avoid this:
- some code blocks were simplified to avoid checking things multiple
  times
- error messages were rolled out as single messages, making them
  properly translatable
- one description text was used as-is, without wrapping it in a 
  redundant string
- a couple of messages got quotes around the placeholders in them, so 
  they now are either the same or similar to existing versions of the
  same messages

### DIFF
--- a/src/main/java/org/candlepin/controller/Entitler.java
+++ b/src/main/java/org/candlepin/controller/Entitler.java
@@ -234,9 +234,11 @@ public class Entitler {
                 consumer.getUuid(), owner.getKey(), caMessage, hypMessage);
 
             if (autobindHypervisorDisabled) {
-                throw new AutobindHypervisorDisabledException(i18n.tr(
-                    "Auto-attach is disabled for owner \"{0}\"{1}.",
-                    owner.getKey(), hypMessage));
+                String message = owner.isAutobindHypervisorDisabled() ?
+                    i18n.tr("Auto-attach is disabled for owner \"{0}\" because of the hypervisor " +
+                    "autobind setting.", owner.getKey()) :
+                    i18n.tr("Auto-attach is disabled for owner \"{0}\".", owner.getKey());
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/controller/Entitler.java
```diff
@@ -234,9 +234,11 @@ public List<Entitlement> bindByProducts(AutobindData data, boolean force)
                 consumer.getUuid(), owner.getKey(), caMessage, hypMessage);
 
             if (autobindHypervisorDisabled) {
-                throw new AutobindHypervisorDisabledException(i18n.tr(
-                    "Auto-attach is disabled for owner \"{0}\"{1}.",
-                    owner.getKey(), hypMessage));
+                String message = owner.isAutobindHypervisorDisabled() ?
```
Ceiu, line 237: Same deal here. The whole outer block is just a complicated way to avoid separate smaller blocks. We should just rewrite the whole thing rather than trying to work around the madness.

Ceiu (changes requested):
The change itself has merit, but should be expanded a bit to better avoid the core problem leading to the string concatenation shenanigans.

Also, the commit message should be updated to follow our format. Commits should be written in past tense, with relevant explanatory details in bulleted entries in the body. The header/title should also be a bit more clear that it's addressing/removing translation string concatenation.

src/main/java/org/candlepin/controller/Entitler.java
Ceiu: Has there been any update or difficulties on this item?

PR #3986, opened 2023-04-28
[M] Use shell script for source translation generation

nikosmoum:


nikosmoum:
Note: This is a replacement (workaround) to avoid maven builds failing when trying to download gradle during running the Msgfmt gradle plugin.

bin/scripts/generate_translations.sh
```diff
@@ -0,0 +1,44 @@
...
+  # a previous build already exist when building sources
+  rm -rf $BUILD_DIR/*
+
+  locale="$(basename $po_file .po)"
+  set -x # echo out the commands as you run them
+  msgfmt --java2 --source -r org.candlepin.common.i18n.Messages -l $locale -d $BUILD_DIR $po_file
+  retVal=$?
+  set +x # Stop echoing
+
+  # Validate that the msgfmt command was run successfully. If not, fail fast.
```
Ceiu, line 34: We could throw a `set -e` at the beginning of the script and eliminate all of the additional checking here, while also catching any weird errors that might happen while setting up the environment.

Ceiu (approved):
Just one small, non-blocking comment, but this otherwise looks good

bin/scripts/generate_translations.sh
```diff
@@ -0,0 +1,41 @@
...
+  set -x # echo out the commands as you run them
+  msgfmt --java2 --source -r org.candlepin.common.i18n.Messages -l $locale -d $BUILD_DIR $po_file
+  set +x # Stop echoing
+
+  # Copy everything over to the generated-sources directory
+  cp -r $BUILD_DIR/* $OUTPUT/
+done
+
+# Always delete the build directory on exit
+trap 'rm -rf -- "$BUILD_DIR"' EXIT
```
Ceiu, line 41: This should be set up earlier, preferably right after the directory is created so we can catch earlier exits

```
# Create a temporary directory that we can freely munge with temporary build artifacts
BUILD_DIR=$(mktemp -d -q msgfmt.XXXXXX)

# Always delete the build directory on exit
trap 'rm -rf -- "$BUILD_DIR"' EXIT
```

bin/scripts/generate_translations.sh
```diff
@@ -0,0 +1,41 @@
...
+translation_files=`ls $INPUT/*.po`
+for po_file in $translation_files
+do
+  echo "Processing file: $po_file"
+  # Ensure the the build directory is empty, as msgfmt complains if any artifacts
+  # a previous build already exist when building sources
+  rm -rf $BUILD_DIR/*
+
+  locale="$(basename $po_file .po)"
+  set -x # echo out the commands as you run them
```
Ceiu, line 32: If we block this off in a subshell, we don't need to turn it back off, and we don't get the spam that comes with `set +x`:

```
(
  set -x
  msgfmt --java2 ...
)

File: bin/scripts/generate_translations.sh [hunk 4]
```diff
@@ -0,0 +1,41 @@
...
+
+# Ensure our output directory exists
+mkdir -p $OUTPUT
+echo "Writing i18n catalog source files to: $OUTPUT"
+
+# Impl note:
+# msgfmt does not allow us to convert more than one .po file at a time due to the need to
+# specify the locale directly for the Java command. For this reason we do the initial
+# generation in the build directory and then copy the artifact the output directory.
+translation_files=`ls $INPUT/*.po`
```
Ceiu, line 23: We can just use the directory/path directly for the for loop here:

```
for po_file in $INPUT/*.po
do
  ...
```

Ceiu (changes requested):
Sorry, did some local testing and found a few cleanup things. These aren't massively critical (especially for a temporary script), but more exercises in bash scripting.

bin/scripts/generate_translations.sh
nikosmoum: done

bin/scripts/generate_translations.sh
nikosmoum: done

bin/scripts/generate_translations.sh
nikosmoum: done

Ceiu: This still has the ls bit in there. The subshell and ls can be dropped by just specifying the path glob: `for po_file in $INPUT/*.po`

Note that if you quote the glob it no longer gets processed and the loop treats it as a single item.

nikosmoum: fixed

PR #3987, opened 2023-04-28
CANDLEPIN-503: Migrated bugzilla reference to Github actions

JoshAlbrecht23:
Created Bugzilla reference check step for PR verification. I migrated the check_pr_branch.py script that is used for the Jenkins job to the .github/scripts directory and renamed the file to bugzilla_reference_check.py to more accurately describe the functionality of the script.

### DIFF
--- a/.github/scripts/bugzilla_reference_check.py
+++ b/.github/scripts/bugzilla_reference_check.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python3
+from __future__ import unicode_literals
+
+import os
+import re
+import sys
+
+from bugzilla import Bugzilla
+import requests
+
+UH_OH = '''
+       _                 _     _
+ _   _| |__         ___ | |__ | |
+| | | | '_ \ _____ / _ \| '_ \| |
+| |_| | | | |_____| (_) | | | |_|
+ \__,_|_| |_|      \___/|_| |_(_)
+'''
+
+bugzilla_token = os.environ.get('BUGZILLA_TOKEN')
+if not bugzilla_token:
+    raise EnvironmentError('BUGZILLA_TOKEN not specified')
+
+github_token = os.environ.get('GITHUB_TOKEN_PSW')
+if not github_token:
+    raise EnvironmentError('GITHUB_TOKEN not specified')
+
+pr_number = os.environ.get('PR_NUMBER')
+if not github_token:
+    raise EnvironmentError('PR_NUMBER not specified')
+
+pr = requests.get('https://api.github.com/repos/candlepin/candlepin/pulls/{pr}'.format(pr=pr_number),
+                  headers={'Authorization': 'token {}'.format(github_token)}).json()
+target = pr['base']['ref']
+
+bz = Bugzilla('bugzilla.redhat.com', api_key=bugzilla_token)
+
+main_version = None
+def fetch_main_version():
+    global main_version
+    if main_version:
+        return main_version
+    spec_file = requests.get('https://raw.githubusercontent.com/candlepin/candlepin/main/candlepin.spec.tmpl').text
+    for line in spec_file.split('\n'):
+        if line.startswith('Version:'):
+            match = re.search('^Version: (\d+\.\d+)\.\d+$', line)
+            version = match.group(1)
+            main_version = version
+            return version
+
+version = None
+if target == 'main':
+    # fetch main spec and then parse from it
+    version = fetch_main_version()
+else:
+    version_match = re.search('^candlepin-(.*)-HOTFIX$', target)
+    if version_match:
+        version = version_match.group(1)
+if not version:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_verification.yml
```diff
@@ -92,3 +92,26 @@ jobs:
...
+
+  bugzilla_check:
+    name: bugzilla reference check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Mask secrets
+        shell: bash
+        run: |
+          echo "::add-mask::${{ secrets.BUGZILLA_API_TOKEN }}"
+          echo "::add-mask::${{ secrets.GITHUB_TOKEN }}"
```
JoshAlbrecht23, line 104: Added masking for the secrets that we are using so that we don't potentially leak them in visible logs.

.github/scripts/bugzilla_reference_check.py
```diff
@@ -0,0 +1,88 @@
...
+if not bugzilla_token:
+    raise EnvironmentError('BUGZILLA_TOKEN not specified')
+
+github_token = os.environ.get('GITHUB_TOKEN_PSW')
+if not github_token:
+    raise EnvironmentError('GITHUB_TOKEN not specified')
+
+pr_number = os.environ.get('PR_NUMBER')
+if not github_token:
+    raise EnvironmentError('PR_NUMBER not specified')
```
JoshAlbrecht23, line 30: I modified this from the original check_pr_branch.py script. I'm setting the PR number through an environment variable instead of an argument in the script. I was having issues with passing in ${{ github.event.pull_request.number }} as an argument.

.github/workflows/pr_verification.yml
```diff
@@ -92,3 +92,26 @@ jobs:
...
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
ptoscano, line 108: this checks runs on "ubuntu-latest", which is currently 22.04 LTS [1]: this means it has python 3.10 has default version, so you don't need to install a custom python; simply install `python3` and `python3-pip`


[1] https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

.github/workflows/pr_verification.yml
```diff
@@ -92,3 +92,26 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: pip install Bugzilla requests
```
ptoscano, line 111: if you install python from the distro repository, then install also `python3-requests` from there too, it should be more than enough

.github/scripts/bugzilla_reference_check.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python
+from __future__ import unicode_literals
+
+import argparse
```
ptoscano, line 4: this is no more needed now, since you changed the way the PR number is passed to this script

.github/workflows/pr_verification.yml
ptoscano, line 1: nit: missing empty line at the end of the file

ptoscano (commented):
while I notice that this is a (almost) straight port of the existing script from jenkins to GHA, please note that the "Bugzilla" library is super-old, updated last in 2015: https://pypi.org/project/bugzilla/; the next would be to convert this script to "python-bugzilla", more regularly updated: https://pypi.org/project/python-bugzilla/

.github/workflows/pr_verification.yml
JoshAlbrecht23: Good catch! Thanks for linking the github hosted runner docs. I added an 'Install packages' step where I install the `python3` and `python3-pip` package. I also updated the bugzilla_reference_check.py to use python3 instead of python.

.github/workflows/pr_verification.yml
JoshAlbrecht23: Updated to install `python3-requests`  in the 'Install packages' step.

.github/scripts/bugzilla_reference_check.py
JoshAlbrecht23: Good catch! Removed

.github/workflows/pr_verification.yml
JoshAlbrecht23: added empty line

Januson (approved):
Changes look good. Just a note here. Although the runner has correct Python version setup-python action also provides [dependency caching](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages) which could save some time on reruns, but the difference will probably be insignificant in this case given it's our only python job and it runs for only a minute.

.github/workflows/pr_verification.yml
Ceiu: This has to be some of the crazier syntax I've ever seen for configuring log output, but it works. Good thinking on preempting this kind of leak. :+1:

PR #3988, opened 2023-04-28
CP-561: POC to use small-rye configuration

wottop:
I did not remove all of the old configuration classes.
It should be running completely with the small-rye config.
Unit tests will need to be added on implementation.
Moved the default database dialect setting to regular defaults.

The only real constraint I saw that could not be overcome in the wrapper I wrote:
You cannot create a parameter name/value with a new name after the config is built. You must put in a default value to "reserve" a spot.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -178,6 +178,10 @@ dependencies {
     implementation libraries["hibernateJpamodelgen"]
     annotationProcessor libraries["hibernateJpamodelgen"]
 
+    //Small Rye Config
+    implementation libraries["smallrye"]
+    implementation libraries["annotations"]
+
     // Use wildcard because this could be called jss4.jar or jss.jar
     // (depending on if we are before Fedora35/RHEL9 or after)
     providedCompile fileTree(dir: '/usr/lib64/jss', include: '*.jar')
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -31,6 +31,7 @@ ext.plugins = [
     "org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:4.0.0.2929",
 ]
 
+libraries["annotations"] = "org.apache.tomcat:tomcat-annotations-api:10.0.0"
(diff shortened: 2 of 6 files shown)


Ceiu:
There are parts of this which make a fairer evaluation difficult. I'm not entirely thrilled with the idea of maintaining a config middleware given the primary argument for using a library was so we didn't have code to maintain. I'd be interested to see how wide spread and invasive the changes would be if we just use the new configuration bits straight up.

The implementation for the encryptedconfig seems viable with a few more optimizations. I'll need to spend a bit more time and look through smallrye-config to see if there's an alternative way to do this that might be a better fit.

There are some other things which I didn't call out yet because it's still PoC code, but a number of things should be given a bit more care during formalization -- specifically surrounding the utility methods like `get<type>` which could have a very uniform and unidirectional call stack/flow order, but don't in the PoC.

Januson:
POC looks alright. There's still a lot of legacy code surrounding the configuration, but I understand that for the purpose of POC it's easier to just interface into it rather than fix everything. We can fix that during the migration and trim down the capabilities of the wrapper. After that a lot of the problems will disappear, like the inability to modify the config. We shouldn't be doing that anyway.

Ceiu:
I agree with @Januson on the modifying config bits -- we shouldn't be doing that once the config is being passed around. I imagine most of the areas that's occurring are in our unit tests, so a test wrapper may be apt; but it'd be good to identify where that's occurring.

There's probably a similar argument surrounding the subset stuff. I think we only do that in the job system to avoid a bunch of extra string building. But it was pre-existing at the time, so it may be used elsewhere. 

Ideally we can drop all of that and the need for a config wrapper fades entirely.

wottop:
The most notable place was in some initialization code where it sets the db dialect to postgres if none is configured.
I could not determine why the usual default method wouldn't serve the same purpose, so I changed it over.

wottop:
Completed POC process

PR #3990, opened 2023-05-02
[M] CANDLEPIN-565: Rewrote dev certificate generation process

Ceiu:
- Added a new pair of scripts to generate CA certificates for Candlepin, replacing the previous script which only generated the local server CA certificate
- Moved the location of cert generation and root dev CA to the "certs" directory
- Removed the now-obsolete gen_certs.sh script


Ceiu:
There are a couple things I'm still unsure about here:
- The primary goal of doing all of this was so we could locally trust a single CA to avoid having to import a new cert every time we spin up a new container/vagrant image. However, due to how Java's CA trust traversal works, this doesn't seem like it's possible without explicitly trusting the whole chain anyway. Maybe this will still be fruitful for the client side of things (which will make the effort worth it), but it certainly doesn't make anything notably better on the server side.
- The scripts living in the `certs` directory is a questionable one. It made sense at the time, but I'm not sure if this is a good move. I'm not a huge fan of the way the bin directory simultaneously pretends to have organization with the `deployment` and `scripts` directories, while still largely just being a dumping ground for misc scripts and whatnot, but maybe it makes sense to live in `bin/certs` instead. The certs themselves still feel at home at the top-level in the certs directory there, though, which is what makes it hard. Do we break it up so that the generators live in the bin directory somewhere and the generated root CA cert+key lives in the root? Very open to input here.
- I still need to write up a simple-ish README doc for the certs directory. After playing with it, I don't want to include any trust script, as I want everyone who is considering trusting the root CA to understand all the implications of doing so (Namely, that since the private key will be present in the repo, literally anyone will be able to use it to sign whatever certs they want and they would be implicitly trusted).

Ceiu:
Closing this PR and rethinking the approach. This basically falls into the same problem the original script has in that it's not flexible enough for general consumption, and doesn't really solve the original problem as nicely as initially hoped.

PR #3991, opened 2023-05-03
CANDLEPIN-570: Consolidate database changelog scripts

wottop:
Had to add dbms indicators to changesets so that a single unified script could be created.

To test you need to [for both postgresql and mysql/mariadb]:
1. run create on this branch
2. run create on main then run update with this branch

In addition, run the unit tests so that the hsqldb changesets run.


bin/deployment/deploy
```diff
@@ -40,7 +40,7 @@ gendb() {
         $SUDO rm -rf /var/lib/candlepin/activemq-artemis/*
     else
         MESSAGE="Updating Database"
-        CHANGELOG="changelog-update.xml"
+        CHANGELOG="changelog-create.xml"
```
Ceiu, line 43: Let's keep the "updated" name, as it better aligns with the verbiage used/expected for this type of operation.

Ceiu (changes requested):
We should use this opportunity to pull all the changesets out of the collection that is the `changelog-update.xml` so they match the individual file format we've used for the last 10 years or so.
This also lets us match up the necessary changes to the hsqldb stuff so we don't need a separate script that runs there. For bonus points, we could also straight up remove stuff where we've added and then removed a given column/table/whatever before we get to the "newer" individual changesets.

bin/deployment/deploy
wottop: Liqubase will not allow it. the changelog-create name is in the DB for determining if the set has been run:
id + author + filename is the key.

Ceiu (approved):

Ceiu:
This looks a lot better, especially in the context of the limitations imposed on us by the file name being part of the changeset hash. It's good to see that it's just the file's base name and not the whole path, but still a little unfortunate.

wottop:
Still a bit of cleaning where the dbms list is just the entire list of choices. No allowlist is the same as a full allowlist.

Ceiu:
Will re-run the tests after Ondrej's PR gets merged to address the sporadic test failures.

PR #3996, opened 2023-05-05
CANDLEPIN-355: [POC] Switch style check to spotless

wottop:


build.gradle
```diff
@@ -85,6 +85,21 @@ ext {
     junit_version = "5.8.2"
 }
 
+allprojects {
+    apply plugin: "com.diffplug.spotless"
+    spotless {
+        java {
```
JoshAlbrecht23, line 91: I think that we could also add the `ratchetFrom 'origin/main'` feature. If I understand the ratchet feature correctly, this make Spotless apply to only files that differ from `origin/main`. This could make our transition easier as we would only need fix the formatting of files as we work on them instead of pushing a massive commit to **all** the files that are not formatted correctly.

https://github.com/diffplug/spotless/tree/main/plugin-gradle#ratchet

JoshAlbrecht23 (approved):

wottop:
POC process complete. Will need the rules xml file updated with our rules in the implementation step.

PR #3997, opened 2023-05-05
CANDLEPIN-569: Run liquibase update on startup

wottop:
- Uses the existing lookup for unrun liquibase change sets to determine  the need for executing the update
- Has 4 levels of action
- Defaults to bypassing the process
- Update for liquibase package use to avoid deprecated methods

### DIFF
--- a/src/main/java/org/candlepin/guice/CandlepinContextListener.java
+++ b/src/main/java/org/candlepin/guice/CandlepinContextListener.java
@@ -15,8 +15,8 @@
 package org.candlepin.guice;
 
 import static org.candlepin.config.ConfigProperties.ACTIVEMQ_ENABLED;
+import static org.candlepin.config.ConfigProperties.DB_MANAGE_ON_START;
 import static org.candlepin.config.ConfigProperties.ENCRYPTED_PROPERTIES;
-import static org.candlepin.config.ConfigProperties.HALT_ON_LIQUIBASE_DESYNC;
 import static org.candlepin.config.ConfigProperties.PASSPHRASE_SECRET_FILE;
 
 import org.candlepin.async.JobManager;
@@ -40,14 +40,18 @@ import com.google.inject.Stage;
 import com.google.inject.persist.PersistService;
 import com.google.inject.util.Modules;
 
-import liquibase.Liquibase;
+import liquibase.changelog.ChangeLogParameters;
 import liquibase.changelog.ChangeSet;
+import liquibase.command.CommandScope;
+import liquibase.command.core.StatusCommandStep;
+import liquibase.command.core.UpdateCommandStep;
+import liquibase.command.core.helpers.DbUrlConnectionCommandStep;
 import liquibase.database.Database;
 import liquibase.database.DatabaseFactory;
 import liquibase.database.jvm.JdbcConnection;
 import liquibase.exception.LiquibaseException;
+import liquibase.parser.core.xml.XMLChangeLogSAXParser;
 import liquibase.resource.ClassLoaderResourceAccessor;
-import liquibase.resource.ResourceAccessor;
 
 import org.apache.commons.lang3.StringUtils;
 import org.hibernate.cfg.beanvalidation.BeanValidationEventListener;
@@ -72,6 +76,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.cache.CacheManager;
 import javax.persistence.EntityManagerFactory;
@@ -91,6 +97,7 @@ import javax.servlet.ServletContextEvent;
  */
 public class CandlepinContextListener extends GuiceResteasyBootstrapServletContextListener {
     public static final String CONFIGURATION_NAME = Configuration.class.getName();
+    public static final String CHANGELOG_FILE_NAME = "db/changelog/changelog-update.xml";
 
     /**
      * The (rough) state of this listener's lifecycle, not including transitional states.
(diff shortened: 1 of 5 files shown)


wottop:
Needs to wait for https://github.com/candlepin/candlepin/pull/3991 so it can be rebased.

src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -405,11 +424,16 @@ private void insertValidationEventListeners(Injector injector) {
      * @throws RuntimeException if there are missing changesets or a LiqubaseException
      */
     protected void checkDbChangelog() {
-        if (!config.getBoolean(HALT_ON_LIQUIBASE_DESYNC)) {
+        log.info(String.format("Liquibase startup management set to %s",
```
Ceiu, line 427: format called inside the log call which already has formatting

This could also come after the string below to use the fetched value instead of looking it up twice.

src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -426,17 +450,46 @@ protected void checkDbChangelog() {
...
-
-                log.error(builder.toString());
-                log.error("Please update the database");
-                log.error("Aborting Candlepin initialization...");
-
-                throw new RuntimeException("The database is missing Liquibase changeset(s)");
+                // all levels above None need this
+                log.warn(builder.toString());
+
+                if (liquibaseOnStart.equals(DBManageLevel.REPORT.getName())) {
```
Ceiu, line 456: This logic should live in the enum itself (if not using the enum's native `.valueOf` call), and then can be moved to the top of the method to avoid a bunch of work when we have an invalid argument.

```
DBManagementLevel dbmLevel = DBManagementLevel.valueOf(config.getString(DB_MANAGE_ON_START, "none"))

switch (dbmLevel) {
  case MANAGE:
    // do update bits here
    break;

  case HALT:
  case CHECK:
    // do status check & potentially halt here
    break;

  default:
  case NONE:
    // Do nothing, basically
}
```

You could also do individual blocks to have shared logic and logging in the case where we do want to manage it, it might be nice to see in the logs the changesets that appear to be missing before applying any update automatically.

src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -426,17 +450,46 @@ protected void checkDbChangelog() {
...
         catch (LiquibaseException e) {
+            log.error("Liquibase exception: " +  e.getMessage());
             throw new RuntimeException(e);
         }
+        finally {
+            if (liquibase != null && liquibase.getDatabase() != null) {
+                DatabaseConnection dc = liquibase.getDatabase().getConnection();
+                if (dc != null) {
+                    try {
+                        dc.rollback();
```
Ceiu, line 484: Do we always want to roll back in this case?

src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -450,12 +503,12 @@ protected Liquibase getLiquibase() {
         Liquibase liquibase = null;
         try {
             // this ensures that the driver class has been loaded
-            Class.forName(config.getString("jpa.config.hibernate.connection.driver_class"))
+            Class.forName(config.getString(ConfigProperties.DB_DRIVER_CLASS))
```
Ceiu, line 506: :+1:

Ceiu (changes requested):
Looks like a few areas of prototype code still hanging out. Also, this is starting to edge in on complex enough to justify being abstracted out into its own class. I'll reserve full judgement until after a bit of refactor around the enum usage, but it seems like this could candidate for abstraction.

src/main/java/org/candlepin/guice/CandlepinContextListener.java
wottop: I can see keeping sets that ran successfully before an error.

Ceiu (changes requested):
Something I noticed that's a bit concerning is that all of the methods we're calling in Liquibase's API are flagged as deprecated. Worse, in the case of both update and listUnrunChangesets, they don't even provide a functional alternative.

This may end up being a source of some pain in the near-ish future.

Ceiu (approved):
The changes mostly line up with where I'd expect this to land, but I have a nagging feeling that it should be abstracted out into its own object. We're at roughly 20% of the total size of the class just in the Liquibase logic, and especially now that we have to build the command objects instead of just using a call, we could easily do that ourselves:

```
class CPLiquibaseWrapper implements AutoCloseable {

  public CPLiquibaseWrapper(Configuration config) {
    ...
  }

  public void close() {
    if (this.database != null) {
      this.database.close();
    }
  }

  public List<ChangeSet> getUnrunChangesets()
  public <void? status out?> updateDatabase()
}

...

try (CPLiquibaseWrapper liquibase = new CPLiquibaseWrapper(config)) {
  ...

```

At this point it doesn't buy us much other than compartmentalizing the code and making it easier to mock in the tests (since you can actually use mocks instead of spying on the context listener); but it's something I've been feeling for the duration of this task.

src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -405,73 +429,115 @@ private void insertValidationEventListeners(Injector injector) {
...
+            .addArgumentValue(UpdateCommandStep.CHANGELOG_FILE_ARG, CHANGELOG_FILE_NAME);
+        commandScope.execute();
+    }
+
+    /**
+     * Reads the list of unrun changesets from the database supplied based on the changelog.
+     *
+     * @param database the object which defines the database connection
+     * @return List of unrun changesets
+     * @throws RuntimeException if there is an issue with reading the changesets.
```
JoshAlbrecht23, line 502: Update "RuntimeException" to "LiquibaseException"

JoshAlbrecht23 (changes requested):
Looks good! I just noticed a couple quick changes to a couple java docs

JoshAlbrecht23 (approved):

PR #3999, opened 2023-05-09
Fix Bugzilla check for PRs from public forks

Januson:
Workflows triggered by PRs from public forks do not have access to repo secrets. This causes failures for jobs that require secrets to work.
    
- Updated bugzilla library
- Removed secret as we use it to only read public bugs
- Reordered GH jobs


nikosmoum (approved):

nikosmoum:
@Januson Can we also fix this for 4.2 and 4.1 branches?

PR #4000, opened 2023-05-09
[M] CANDLEPIN-580: Fixed various remapping issues during refresh

Ceiu:
- Fixed an issue where a "dirty" local reference to one or more existing entities would trigger a global org-entity remapping, but would not fix the mapping error in the entity model
- Reduced the conditions in which the global org-entity remapping logic would be invoked to only cases where an org has one or more missing org-entity mappings


Ceiu:
This has the issue of potentially disrupting refresh entirely in some cases:

org mapping: p1, p2, p3b
prod mapping: p1->p2, p1->p3a

This logic would detect that p1->p3a needs to become p1->p3b but would hit a duplicate entry at the point where the entity refs would be updated to point to p3b.

PR #4002, opened 2023-05-11
[M] CANDLEPIN-580: Fixed various remapping issues during refresh

Ceiu:
- Fixed an issue where a "dirty" local reference to one or more
  existing entities would trigger a global org-entity remapping,
  but would not fix the mapping error in the entity model
- Reduced the conditions in which the global org-entity remapping
  logic would be invoked to only cases where an org has one or
  more missing org-entity mappings
- Entity trees which contain one or more dirty nodes will now be
  entirely rebuilt, clearing the versioning of any dirty node and
  all nodes above any dirty nodes in the tree
- Versioning lookup will now be skipped for dirty nodes to avoid a
  case where the rebuild process will resolve to an existing
  entity, undoing the work of the rebuild step
- Extremely rare cases where entity version resolution would
  result in an entity mapping back to itself will now be skipped
  from the mapping operation, and a warning will be issued instead
- Certain UUID-based entity lookups which used sets have been
  converted to use lists or generic collections to avoid a case
  where two entities with different UUIDs but are otherwise equal
  would cause one of them to be overwritten

### DIFF
--- a/src/main/java/org/candlepin/model/ContentCurator.java
+++ b/src/main/java/org/candlepin/model/ContentCurator.java
@@ -158,18 +158,18 @@ public class ContentCurator extends AbstractHibernateCurator<Content> {
     }
 
     /**
-     * Fetches a set consisting of the content of the  products specified by the given UUIDs. If the
-     * given products do not have any content or no products exist with the provided UUIDs, this
-     * method returns an empty set.
+     * Fetches a collection consisting of the content of the products specified by the given UUIDs.
+     * If the given products do not have any content or no products exist with the provided UUIDs,
+     * this method returns an empty collection.
      *
      * @param productUuids
      *  a collection of UUIDs of products for which to fetch children content
      *
      * @return
-     *  a set consisting of the content of the products specified by the given UUIDs
+     *  a collection consisting of the content of the products specified by the given UUIDs
      */
-    public Set<Content> getChildrenContentOfProductsByUuids(Collection<String> productUuids) {
-        Set<Content> output = new HashSet<>();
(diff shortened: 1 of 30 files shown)


Ceiu:
This is part one of two major issues affecting refresh times in production, as well as preventing the remapping logic from actually doing its job.

src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -158,7 +158,7 @@ public Map<String, Set<String>> getProductsReferencingContent(Collection<String>
     }
 
     /**
-     * Fetches a set consisting of the content of the  products specified by the given UUIDs. If the
+     * Fetches a set consisting of the content of the products specified by the given UUIDs. If the
```
JoshAlbrecht23, line 161: Update "Fetches a set" to "Fetches a collection" or maybe "Fetches a list".

It looks like this could also apply to the following methods:
ProductCurator.getChildrenProductsOfProductsByUuids

JoshAlbrecht23 (changes requested):
Looks good! I noticed some small changes in a couple Java docs, and I believe there is an extra 'a' character at the end of one of you commit bullets. I believe it "stepa" should be "step".

**Commit bullet:**
"Versioning lookup will now be skipped for dirty nodes to avoid a
case where the rebuild process will resolve to an existing
entity, undoing the work of the rebuild stepa"

src/main/java/org/candlepin/model/ContentCurator.java
Ceiu: fixed

JoshAlbrecht23 (approved):

wottop:
LGTM. Ack.

PR #4006, opened 2023-05-11
[M] CANDLEPIN-580: Fixed missing null-as-empty check in content equality test

Ceiu:
- ContentManager.isChangedBy now respects the null-as-empty property of the five content fields where null and empty have the same value: arches, content url, GPG url, release version, and required tags
- Refresh will no longer flag content for update when the only change appears to be setting a null-as-empty field to an empty string when the local value is null


Ceiu:
This is the second part of the performance-related issues found during the investigation of CANDLEPIN-580. Hopefully there will not need to be more changes.

PR #4032, opened 2023-05-23
CANDLEPIN-505: Migrate sonar to GH Actions

bakajstep:
- Migrate sonar to GH actions

### DIFF
--- a/.github/workflows/pr_verification.yml
+++ b/.github/workflows/pr_verification.yml
@@ -87,7 +87,15 @@ jobs:
       - name: Run unit tests
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: test
+          arguments: test coverage
+
+      - name: Archive reports
+        uses: actions/upload-artifact@v3
+        with:
+          name: jacocoReports
+          path: |
+            **/build/test-results/**/*.xml
+            **/build/jacoco/test.exec
 
   validate_translations:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_verification.yml
```diff
@@ -82,7 +82,15 @@ jobs:
...
-          arguments: test
+          arguments: test coverage
+
+      - name: Archive reports
+        uses: actions/upload-artifact@v3
+        with:
+          name: jacocoReports
+          path: |
+            **/build/test-results/**/*.xml
+            **/build/jacoco/test.exec
```
nikosmoum, line 93: I don't think we should be filtering anything here. The filtering should only happen on sonarcloud

nikosmoum: nevermind, this isn't exclusions.

nikosmoum (approved):

nikosmoum:
This is good to merge, however, I want to post my PR on getting the long-lived branch sonar analysis GH Action up and get that merged first, in order to not leave the sync issue hanging that can cause other PRs to fail for no reason of their own.

nikosmoum:
The related PR I mentioned earlier: https://github.com/candlepin/candlepin/pull/4038

nikosmoum (commented):
While working on my PR for the branch analysis, I checked how the Cache step you used works, and it doesn't seem to work. I can see `Cache not found for input keys: Linux-sonar, Linux-sonar` when `Cache SonarCloud packages` runs, and then `Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.` in the `Post Cache SonarCloud packages`

nikosmoum:
PR looks ready to merge, but first I would like to fix the 'long-lived branch analysis' workflow that has been failing. If we don't fix it first, then the main branch will have outdated sonar analysis which will cause headaches for all other PRs

nikosmoum:
One last (hopefully) PR needs to be reviewed & merged before this can also get merged: https://github.com/candlepin/candlepin/pull/4053

PR #4034, opened 2023-05-24
[M] CANDLEPIN-589: Added support for forced-update owner refresh

Ceiu:
- Added support for performing forced-update refresh operations, which will cause refreshed entities to be updated even in cases where an explicit change in data is not detected
- Added the "force" query parameter to the refresh pools API endpoint (PUT /owners/{owner_key}/subscriptions), which will put the refresh operation into forced-update mode
- Removed the trace-level logging of upstream subscriptions and product data from the refresh operation

### DIFF
--- a/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
+++ b/src/main/java/org/candlepin/controller/CandlepinPoolManager.java
@@ -81,7 +81,6 @@ import org.candlepin.util.Traceable;
 import org.candlepin.util.TraceableParam;
 import org.candlepin.util.Util;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.Iterables;
 import com.google.inject.persist.Transactional;
 
@@ -111,7 +110,6 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 import javax.inject.Provider;
-import javax.ws.rs.core.MediaType;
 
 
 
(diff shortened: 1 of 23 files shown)


src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -285,10 +284,11 @@ private String getVirtQuantity(String virtLimit, long primaryPoolQuantity) {
      * @param floatingPools pools with no subscription ID
```
wottop, line 284: Also needs force parameter

src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -623,27 +625,30 @@ private boolean checkForChangedProducts(Product incomingProduct, Pool existingPo
...
+        return false;
     }
 
-    private boolean checkForDateChange(Date start, Date end, Pool existingPool) {
+    private boolean checkForDateChange(Date start, Date end, Pool existingPool, boolean force) {
         boolean datesChanged = (!start.equals(existingPool.getStartDate())) ||
             (!end.equals(existingPool.getEndDate()));
 
-        if (datesChanged) {
+        if (datesChanged || force) {
```
wottop, line 640: Maybe we should check force and short circuit so we can skip the datesChanged calculation.

src/main/java/org/candlepin/policy/js/pool/PoolRules.java
```diff
@@ -655,11 +660,13 @@ private boolean checkForQuantityChange(Pool pool, Pool existingPool, Long origin
 
         boolean quantityChanged = !(expectedQuantity == existingPool.getQuantity());
 
-        if (quantityChanged) {
+        if (quantityChanged || force) {
```
wottop, line 663: short circuit here too?

wottop (changes requested):

src/main/java/org/candlepin/controller/CandlepinPoolManager.java
```diff
@@ -729,7 +698,9 @@ private Map<String, List<Entitlement>> groupByPoolId(Collection<Entitlement> ent
      * @return
      */
     @Transactional
-    void updateFloatingPools(List<Pool> floatingPools, boolean lazy, Map<String, Product> changedProducts) {
+    void updateFloatingPools(List<Pool> floatingPools, Map<String, Product> changedProducts, boolean lazy,
+        boolean force) {
```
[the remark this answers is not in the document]
Ceiu: Fixed

src/main/java/org/candlepin/policy/js/pool/PoolRules.java
Ceiu: Fixed

src/main/java/org/candlepin/policy/js/pool/PoolRules.java
Ceiu: Fixed

wottop (approved):

PR #4038, opened 2023-05-30
CANDLEPIN-505: Add Sonarcloud long-lived branch analysis GH Action

nikosmoum:
- We need this for when sonarcloud compares each PR code's analysis with the analysis of its target branch. If it's not there, then the PR analysis will show sonar issues that aren't actually in the PR code itself, but only exist in the target branch's commits that weren't analysed yet.

### DIFF
--- a/.github/workflows/sonar_branch_analysis.yml
+++ b/.github/workflows/sonar_branch_analysis.yml
@@ -1,0 +1,47 @@
+---
+name: Sonar analysis for long-lived branches
+
+on:
+  push:
+    branches:
+      - main
+      - candlepin-*-HOTFIX
+
+env:
+  JAVA_DISTRIBUTION: 'temurin'
+  JAVA_VERSION: '17'
+
+jobs:
+  branch_sonar_analysis:
+    name: long-lived branch sonar analysis
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:38
+    steps:
+      - name: Mask secrets
+        shell: bash
+        run: |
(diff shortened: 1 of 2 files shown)


Ceiu (approved):
Looks good.
I'm trusting the fact that several other PRs run almost identical jobs/steps, since we can't actually invoke this until it's merged.

bakajstep:
It looks good as well.

bakajstep (approved):
It looks good as well.

.github/workflows/sonar_branch_analysis.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: Archive reports
+        uses: actions/upload-artifact@v3
+        with:
+          name: jacocoReports
+          path: |
+            **/build/test-results/**/*.xml
+            **/build/jacoco/test.exec
+
+  branch_sonar_analysis:
```
Januson, line 47: This can be a single job. We don't really care about unit-test job results like we do in the PR workflow so separating the won't give us anything.

Januson (changes requested):
Looks good. I have just a small comment regarding the job structure.

.github/workflows/sonar_branch_analysis.yml
nikosmoum: I've consolidated this into a single job, which means we also don't need to archive/download the artifacts

Januson (approved):

PR #4050, opened 2023-06-02
[M] Increase gradle memory for sonar analysis reports

nikosmoum:
- Avoid errors in the long-lived branch sonar analysis around JVM heap space running out.

### DIFF
--- a/.github/workflows/sonar_branch_analysis.yml
+++ b/.github/workflows/sonar_branch_analysis.yml
@@ -50,3 +50,4 @@ jobs:
         with:
           arguments: test coverage sonar
             -Dsonar.branch.name=${{ github.ref_name }}
+            -Dorg.gradle.jvmargs='-Xmx1g -XX:MaxMetaspaceSize=512m'


nikosmoum:
Here's an example failure: https://github.com/candlepin/candlepin/actions/runs/5153348360/jobs/9280473183
with error:
```
The Daemon will expire after the build after running out of JVM heap space.
The project memory settings are likely not configured or are configured to an insufficient value.
The daemon will restart for the next build, which may increase subsequent build times.
These settings can be adjusted by setting 'org.gradle.jvmargs' in 'gradle.properties'.
The currently configured max heap space is '512 MiB' and the configured max metaspace is '384 MiB'.
For more information on how to set these values, visit the user guide at https://docs.gradle.org/8.1/userguide/build_environment.html#sec:configuring_jvm_memory
To disable this warning, set 'org.gradle.daemon.performance.disable-logging=true'.
Daemon will be stopped at the end of the build after running out of JVM heap space
```

Januson (approved):

PR #4055, opened 2023-06-07
2213197: Ensure pool lock order for activation keys

wottop:
This issue arises because the pool sorting and locking is happening in sub-sets based on key. The entitlement revoke sorts on all the pools that are attached. The difference in sort order while locking between the actions causes a deadlock when a force re-registration is done concurrently. Existing consumers that are all using the same combination (>1) of keys will have this issue if the earlier keys have "later" pool ids.

You can see the error if you comment out the new locking lines in ConsumerBindUtil.


src/main/java/org/candlepin/resource/util/ConsumerBindUtil.java
```diff
@@ -88,6 +93,15 @@ public void handleActivationKeys(Consumer consumer, List<ActivationKey> keys,
         boolean scaEnabledForAny = false;
         boolean isAutoheal = BooleanUtils.isTrue(consumer.isAutoheal());
 
+        // we need to lock all the pools in id order so that it won't deadlock if there are other
+        // processes in the same space at the same time. Current code sorts and locks
+        // per activation key which can lead to this deadlock when entitlement revocation is
+        // occurring on the same pools
+        Set<Pool> pools = new HashSet();
+        keys.stream().forEach(x -> pools.addAll(
```
Ceiu, line 101: We can reduce the number of explicitly written operations here and add some safety with the use of `flatMap` and stream filters:

```
Set<Pool> akPools = keys.stream()
    .filter(Objects::nonNull)
    .map(ActivationKey::getPools)
    .filter(Objects::nonNull)
    .flatMap(Set::stream)
    .filter(Objects::nonNull)
    .collect(Collectors.toSet());

this.poolCurator.lock(pools);
```

Additionally, we can then condense it down a bit by merging the map and flatMap ops into one line:
```
Set<Pool> akPools = keys.stream()
    .filter(Objects::nonNull)
    .flatMap(key -> key.getPools().stream())
    .filter(Objects::nonNull)
    .collect(Collectors.toSet());
```

Ceiu, line 101: We can reduce the number of explicitly written operations here and add some safety with the use of `flatMap` and stream filters:

```
Set<Pool> akPools = keys.stream()
    .filter(Objects::nonNull)
    .map(ActivationKey::getPools)
    .filter(Objects::nonNull)
    .flatMap(Set::stream)
    .filter(Objects::nonNull)
    .collect(Collectors.toSet());

this.poolCurator.lock(pools);
```

Additionally, we can then condense it down a bit by merging the map and flatMap ops into one line:
```
Set<Pool> akPools = keys.stream()
    .filter(Objects::nonNull)
    .flatMap(key -> key.getPools().stream())
    .filter(Objects::nonNull)
    .collect(Collectors.toSet());
```

Ceiu (changes requested):
Only a minor change request here. We can make the locking step a bit more readable (and potentially faster) with the stream ops instead of manually building multiple sets.

PR #4056, opened 2023-06-07
[M] CANDLEPIN-411: Removed obsolete serialization filters

Ceiu:
- Removed all JSON and XML filtering annotations from model objects
- Removed obsolete filters applied to the global Jackson ObjectMapper
- Removed the obsolete ResolverUtil object
- Updated a few areas that were serializing model objects to instead serialize appropriate DTOs


Januson (approved):
Looks good to me.

PR #4058, opened 2023-06-07
2213261: [M] Ensure pool lock order for activation keys

wottop:
This issue arises because the pool sorting and locking is happening in sub-sets based on key. The entitlement revoke sorts on all the pools that are attached. The difference in sort order while locking between the actions causes a deadlock when a force re-registration is done concurrently. Existing consumers that are all using the same combination (>1) of keys will have this issue if the earlier keys have "later" pool ids.

You can see the error if you comment out the new locking lines in ConsumerBindUtil.

Note that this does not solve an issue with the volume of concurrent force refresh operations that can successfully occur without lock wait timeouts.

### DIFF
--- a/.github/workflows/pr_verification.yml
+++ b/.github/workflows/pr_verification.yml
@@ -386,6 +386,8 @@ jobs:
 
       - name: Run Spec tests
         uses: gradle/gradle-build-action@v2
+        env:
+          MYSQL_IN_USE: ${{ matrix.database == 'mariadb' }}
         with:
           arguments: spec -Dspec.test.client.host=candlepin
 
--- a/spec-tests/src/main/java/org/candlepin/spec/bootstrap/assertions/NotWithMySql.java
+++ b/spec-tests/src/main/java/org/candlepin/spec/bootstrap/assertions/NotWithMySql.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.spec.bootstrap.assertions;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.condition.EnabledIf;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+
+/**
+ * This annotation marks a test or test suite to run only when not running against MariaDB/MySql.
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@EnabledIf(value = "org.candlepin.spec.bootstrap.assertions.DatabaseEmployed#notWithMySql", disabledReason = "Candlepin is running against MySQL")
+@Tag("notWithMySql")
+public @interface NotWithMySql {
+}
--- a/spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceActivationKeySpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceActivationKeySpecTest.java
@@ -15,7 +15,9 @@
 package org.candlepin.spec.consumers;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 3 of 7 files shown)


spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceActivationKeySpecTest.java
```diff
@@ -91,6 +99,139 @@ public void shouldAllowConsumerToRegisterWithActivationKeys() {
...
+        threads.forEach(Thread::start);
+        for (Thread thread : threads) {
+            while (thread.isAlive()) {
+                thread.join();
+            }
+        }
+        assertThat(exceptions).isEmpty();
+        assertThat(consumers).hasSize(consumerCount);
+        pools.forEach(pool -> assertThat(
+            client.pools().getPool(pool.getId(), null, null).getConsumed()).isEqualTo(consumerCount));
```
JoshAlbrecht23, line 167: We may want to assert that the pool we get back from `client.pools.getPool()` is not null before asserting against the consumer count.

JoshAlbrecht23 (changes requested):

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceActivationKeySpecTest.java
```diff
@@ -91,6 +100,131 @@ public void shouldAllowConsumerToRegisterWithActivationKeys() {
...
+                .addAttributesItem(ProductAttributes.MultiEntitlement.withValue("yes"));
+            prod = ownerProductApi.createProductByOwner(owner.getKey(), prod);
+            PoolDTO pool = ownerClient.createPool(owner.getKey(), Pools.random(prod).quantity(100L));
+            if (entry < 3) {
+                activationKeyApi.addPoolToKey(key1.getId(), pool.getId(), 1L);
+            }
+            else {
+                activationKeyApi.addPoolToKey(key2.getId(), pool.getId(), 1L);
+            }
+        });
```
JoshAlbrecht23, line 132: Maybe we could move this to private helper method and specify the ActivationKeyDTO that should be added to pool and the number of products and pools that should be created. I'm just trying to break up this large test a bit so it is more digestible.  I know that is not easy to do given the nature of what this test is testing!

I'm thinking the method could be something like:
```
    private void createProductsAndPoolsForActivationKey(ActivationKeyDTO key, int numberOfProductAndPools) {
        IntStream.range(0, numberOfProductAndPools).forEach(entry -> {
            ProductDTO prod = Products.random()
                .addAttributesItem(ProductAttributes.MultiEntitlement.withValue("yes"));
            prod = ownerClient.createProductByOwner(owner.getKey(), prod);
            PoolDTO pool = ownerClient.createPool(owner.getKey(), Pools.random(prod).quantity(100L));
            activationKeyApi.addPoolToKey(key.getId(), pool.getId(), 1L);
        });
    }
```

JoshAlbrecht23 (changes requested):
Found a part of `shouldAllowConsumerToRegisterThenUnregisterWithActKeysConcurrently` that can be broken out into a helper method to help with readability.

.github/workflows/pr_verification.yml
```diff
@@ -370,6 +370,13 @@ jobs:
           echo "MODULE_CONFIG_HOSTED_CONFIGURATION_MODULE=org.candlepin.hostedtest.AdapterOverrideModule" >> .env
           echo "CANDLEPIN_AUTH_CLOUD_ENABLE=true" >> .env
 
+      - if: matrix.database == 'mariadb'
+        name: Add mysql in use env variable
+        working-directory: ./.github/containers
+        shell: bash
+        run: |
+          echo "MYSQL_IN_USE=true" >> .env
```
JoshAlbrecht23, line 378: We don't need this as we are always setting "MYSQL_IN_USE" in mariadb.docker-compose.yaml

JoshAlbrecht23 (approved):

spec-tests/src/main/java/org/candlepin/spec/bootstrap/assertions/NotWithMySql.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+
+/**
+ * This annotation marks a test or test suite to run only in standalone mode.
```
bakajstep, line 28: I think comment needs to be updated, because its affected by mysql database not mode in which candlepin runs.

bakajstep (commented):
Just a smallthing. Otherwise it looks good to me.

bakajstep (approved):

.github/workflows/pr_verification.yml
```diff
@@ -386,6 +386,8 @@ jobs:
 
       - name: Run Spec tests
         uses: gradle/gradle-build-action@v2
+        env:
+          MYSQL_IN_USE: ${{ matrix.database == 'mariadb' && 'true' || 'false' }}
```
nikosmoum, line 390: I'm a bit confused: what is the purpose of the `&& 'true' || 'false'` ? wouldn't `${{ matrix.database == 'mariadb' }}` be enough?

JoshAlbrecht23: Yeah, that is definitely worth trying!

wottop: It works fine with the latter. Changed.

PR #4066, opened 2023-06-12
Run Sonar scan as a separate workflow

Januson:
Workflows triggered by PRs from forks do not have access to secrets, so jobs like sonar cannot run. Moving sonar job to a separate workflow triggered by workflow_run lets it access secrets.

### DIFF
--- a/.github/workflows/sonar_pr_analysis.yml
+++ b/.github/workflows/sonar_pr_analysis.yml
@@ -1,0 +1,79 @@
+---
+name: Sonar analysis for pull requests
+
+on:
+  workflow_run:
+    workflows:
+      - Pull request verification
+    types:
+      - completed
+env:
+  JAVA_DISTRIBUTION: 'temurin'
+  JAVA_VERSION: '17'
+
+# Cancel in-progress sonar branch analysis workflows. We only care about analyzing the latest commit.
+concurrency:
+  group: ${{ github.event.workflow.name }}-${{ github.event.workflow_run.pull_requests[0].number }}
+  cancel-in-progress: true
+
+jobs:
+  pr_sonar_analysis:
+    name: PR sonar analysis
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:38
+    steps:
+      - name: Mask secrets
+        shell: bash
+        run: |
+          echo "::add-mask::${{ secrets.GITHUB_TOKEN }}"
+          echo "::add-mask::${{ secrets.SONAR_TOKEN }}"
+
+      - name: Check out repository
+        uses: actions/checkout@v3
+
+      - name: Download test reports
+        uses: actions/github-script@v6
+        with:
+          script: |
+            let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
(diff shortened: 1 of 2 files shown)


.github/workflows/sonar_pr_analysis.yml
```diff
@@ -0,0 +1,80 @@
...
+  JAVA_VERSION: '17'
+
+# Cancel in-progress sonar branch analysis workflows. We only care about analyzing the latest commit.
+concurrency:
+  group: ${{ github.event.workflow.name }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  branch_sonar_analysis:
+    name: long-lived branch sonar analysis
```
nikosmoum, line 21: This job isn't about long-lived branches, it's about PRs, so we need a different name like:
```
pr_sonar_analysis:
  name: PR sonar analysis
```

.github/workflows/sonar_pr_analysis.yml
```diff
@@ -0,0 +1,80 @@
...
+      - Pull request verification
+    types:
+      - completed
+env:
+  JAVA_DISTRIBUTION: 'temurin'
+  JAVA_VERSION: '17'
+
+# Cancel in-progress sonar branch analysis workflows. We only care about analyzing the latest commit.
+concurrency:
+  group: ${{ github.event.workflow.name }}-${{ github.ref_name }}
```
nikosmoum, line 16: Is the github.ref_name variable available here? Since the workflow wasn't triggered by a push on a branch/tag or PR commit, how can it have a branch name in the context?

.github/workflows/sonar_pr_analysis.yml
```diff
@@ -0,0 +1,80 @@
...
+          java-version: ${{ env.JAVA_VERSION }}
+
+      - name: Run sonar
+        uses: gradle/gradle-build-action@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        with:
+          arguments: sonar -x coverage
+            -Dsonar.branch.name=${{ github.event.workflow_run.display_title }}
```
nikosmoum, line 76: why is the sonar.branch.name needed? aren't the `base` and `branch` variables enough?

.github/workflows/sonar_pr_analysis.yml
nikosmoum: Perhaps it should use the `github.event.workflow_run.pull_requests[0].number` variable like below?

.github/workflows/sonar_pr_analysis.yml
Januson: Done.

.github/workflows/sonar_pr_analysis.yml
Januson: Done. Switched to number.

.github/workflows/sonar_pr_analysis.yml
Januson: Just forgot to remove it. Done.

PR #4090, opened 2023-06-20
Fix test coverage in Sonar scan

Januson:
- Disable unnecessary html and csv reports
- Upload only coverage info in unit tests
- Generate test report from coverage info during sonar scan

### DIFF
--- a/.github/workflows/pr_verification.yml
+++ b/.github/workflows/pr_verification.yml
@@ -87,14 +87,13 @@ jobs:
       - name: Run unit tests
         uses: gradle/gradle-build-action@v2
         with:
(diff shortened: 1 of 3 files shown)


nikosmoum (commented):
The change seems good, but can you test it by temporarily making a production code change, to see if the test coverage works?

Januson:
@nikosmoum Unfortunately that won't work. There's a change to the sonar_pr_analysis so merge is the only way to test it. I ran it from my fork. Coverage got generated correctly but I couldn't run sonar from there.

nikosmoum (approved):

PR #4092, opened 2023-06-20
[M] CANDLEPIN-610: GH actions spec tests can't push images from forks

JoshAlbrecht23:
- updated postgres docker-compose health check
- updated spec test Dockerfile to use UBI 9 base image

I created a PR from a forked repository to validate that the spec tests properly run:
https://github.com/candlepin/candlepin/actions/runs/5338261121/jobs/9675387427

### DIFF
--- a/.github/containers/Dockerfile
+++ b/.github/containers/Dockerfile
@@ -25,19 +25,16 @@ WORKDIR /app/build
 RUN jar xf $(find . -name 'candlepin*.war' | head -n 1); \
     sed -i 's/jss4.jar/jss.jar/g' ./META-INF/context.xml
 
-FROM quay.io/centos/centos:stream9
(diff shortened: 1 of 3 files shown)


Januson (changes requested):
7 minutes. It's worse but not too bad. Nikos mentioned UBI repo got updated. Can you try what it would look like if we could use that? Preferably if we could get away with using a minimal one.

JoshAlbrecht23:
> 7 minutes. It's worse but not too bad. Nikos mentioned UBI repo got updated. Can you try what it would look like if we could use that? Preferably if we could get away with using a minimal one.

Good idea! I updated the Dockerfile to use the UBI 9 as a base image and it looks like it reduced the time from 3 min 15 sec to 2 min 12 seconds, so that definitely helped!

Januson (approved):

PR #4102, opened 2023-06-27
[M] CANDLEPIN-173: Created containerfile for production image

JoshAlbrecht23:
- Updated spec test containerfile to use entrypoint
- Added null check to RyeConfig.toProperties method

# Production Containerfile

Image size: 825 MB

## Builder Stage

In the builder stage of the release.Containerfile I am downloading and unzipping Tomcat and also unzipping the Candlepin WAR file. I am doing this in a 'builder' stage so that we do not have to include the wget and tar package in the actual production image. Also, it reduces the layers in the production image and further reduces the image size.

## Labels

For the labels of the production image, I went with the OCI convention and selected what seemed to be useful labels. Let me know if there are any modifications that you want to make to this, or if we should use our own convention.

https://github.com/opencontainers/image-spec/blob/main/annotations.md

## Entrypoint Script

I added an entrypoint script to the production image so that we can update the Java trust with the mounted certificates that we are expecting in the /etc/candlepin/certs/ directory. I'm not sure if there is a better way of handling this as you need to call `update-ca-trust` when you are attempting to run the container with the certs.

## Testing

I ran the spec tests against the production container locally to test the containerfile. To test locally you can use these commands to setup a  Candlepin container in standalone mode using Postgres DB on the host nework.

### Start Postgres Container

```
sudo docker run --name postgres -itd  \
	-e POSTGRES_USER=candlepin \
	-e POSTGRES_PASSWORD=candlepin \
	-e POSTGRES_DB=candlepin \
	-e POSTGRES_HOST_AUTH_METHOD=trust \
	--network host \
	postgres:latest
```

### Build Candlepin Image

```
sudo docker build -f ./containers/release.Containerfile -t <Name>:<Tag> . \
	--build-arg WAR_FILE=./build/libs/candlepin-4.3.6.war \
	--no-cache
```

### Run the Candlepin Image

```
sudo docker run --name candlepin -itd  \
	-e JPA_CONFIG_HIBERNATE_DIALECT=org.hibernate.dialect.PostgreSQL92Dialect \
	-e JPA_CONFIG_HIBERNATE_CONNECTION_DRIVER_CLASS=org.postgresql.Driver \
	-e JPA_CONFIG_HIBERNATE_CONNECTION_URL=jdbc:postgresql://localhost/candlepin \
	-e JPA_CONFIG_HIBERNATE_CONNECTION_USERNAME=candlepin \
	-e JPA_CONFIG_HIBERNATE_CONNECTION_PASSWORD=candlepin \
	-e CANDLEPIN_AUTH_TRUSTED_ENABLE="true" \
      	-e CANDLEPIN_AUTH_OAUTH_ENABLE="true" \
	-e CANDLEPIN_STANDALONE=true \
	-e CANDLEPIN_REFRESH_ORPHAN_ENTITY_GRACE_PERIOD=0 \
	-e CANDLEPIN_DB_DATABASE_MANAGE_ON_STARTUP=Manage \
	--mount type=bind,source=./.github/containers/server.xml,target=/opt/tomcat/conf/server.xml \
	--volume ./tmp/certs:/etc/candlepin/certs \
	--network host \
	<Name>:<Tag>
```


containers/release.Containerfile
```diff
@@ -0,0 +1,86 @@
+FROM quay.io/centos/centos:stream9 as builder
```
bakajstep, line 1: Shouldn't we use: Base image UBI 9?

JoshAlbrecht23: Yeah, we can use UBI 9 for the builder stage! Updated

Ceiu:
That entrypoint script has me slightly concerned. Before we continue forward with that, I would like to understand our goal with this image.

If the goal is to be able to run it as-is with arbitrary set of certs and server config on every run, then I would argue that it's probably fine. However, if the goal is for users to basically wrap/extend this container with their own to add the bits they need, then it's probably extraneous. I'm in a similar place with the server.xml file. I feel like anything we have to mount in is going to add to the pain of working with this. 

That being the case, I would like to know if our goal is for this container to be runnable straight away, or if we're expecting a lot of hand-configuration or rebuild.

JoshAlbrecht23:
> That entrypoint script has me slightly concerned. Before we continue forward with that, I would like to understand our goal with this image.
> 
> If the goal is to be able to run it as-is with arbitrary set of certs and server config on every run, then I would argue that it's probably fine. However, if the goal is for users to basically wrap/extend this container with their own to add the bits they need, then it's probably extraneous. I'm in a similar place with the server.xml file. I feel like anything we have to mount in is going to add to the pain of working with this.
> 
> That being the case, I would like to know if our goal is for this container to be runnable straight away, or if we're expecting a lot of hand-configuration or rebuild.

Those are some good points. I believe that this production image should be designed to be extended/wrapped by consumers to produce their own image with their own server.xml configurations, certs, and Candlepin configurations. So, it seems appropriate that the consumer includes the updating of the Java trust store as a part of their container image building process. If that is the case, then I think we can remove the entrypoint script as it won't be needed.

Opinions? @nikosmoum @wottop @Januson @bakajstep

JoshAlbrecht23:
@Ceiu I removed the entrypoint script from the release.Containerfile as we can expect the consumers who extend the production image to update the Java trust store as a part of their build process.

containers/release.Containerfile
```diff
@@ -0,0 +1,78 @@
+FROM registry.access.redhat.com/ubi9-minimal:9.2-691 as builder
+
+ARG WAR_FILE
```
Januson, line 3: A few questions:
- Do we want to build production image from war or from RPM? E.G. built in PNC?
- If we we want war. What's the benefit of getting the war as a param compared to building it in the dockerfile.

PR #4106, opened 2023-06-28
[M] Revert "CANDLEPIN-589: Added support for forced-update owner refresh"

Ceiu:
This reverts commit c9ed71f4b920857564f78e6cb5eeaf8fbf404b49.

### DIFF
--- a/src/main/java/org/candlepin/async/tasks/RefreshPoolsJob.java
+++ b/src/main/java/org/candlepin/async/tasks/RefreshPoolsJob.java
@@ -45,7 +45,6 @@ public class RefreshPoolsJob implements AsyncJob {
 
     protected static final String OWNER_KEY = "org";
     protected static final String LAZY_REGEN = "lazy_regen";
-    protected static final String FORCE_UPDATE = "force_update";
 
     private final OwnerCurator ownerCurator;
     private final RefresherFactory refresherFactory;
@@ -102,22 +101,6 @@ public class RefreshPoolsJob implements AsyncJob {
             return this;
         }
 
-        /**
-         * Sets the flag for performing a forced-update refresh, causing all entities to be updated
-         * even if no explicit change in data is detected.
-         *
-         * @param force
-         *  whether or not to perform a forced-update refresh
-         *
-         * @return
-         *  a reference to this job config
-         */
-        public RefreshPoolsJobConfig setForceUpdate(boolean force) {
-            this.setJobArgument(FORCE_UPDATE, force);
-
-            return this;
-        }
-
         @Override
         public void validate() throws JobConfigValidationException {
             super.validate();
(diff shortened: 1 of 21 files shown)


src/main/java/org/candlepin/async/tasks/RefreshPoolsJob.java
```diff
@@ -178,7 +159,6 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
             // Assume that we verified the request in the resource layer:
             this.refresherFactory.getRefresher(this.subAdapter, this.prodAdapter)
                 .setLazyCertificateRegeneration(lazy)
```
JoshAlbrecht23, line 161: Do we also want to revert the addition of `.setLazyCertificateRegeneration(lazy)`? It looks like this line was added by commit c9ed71f4b920857564f78e6cb5eeaf8fbf404b49 as well.

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1292,8 +1292,7 @@ public AsyncJobStatusDTO refreshPools(String ownerKey, Boolean autoCreateOwner,
 
         JobConfig config = RefreshPoolsJob.createJobConfig()
             .setOwner(owner)
-            .setLazyRegeneration(true)
-            .setForceUpdate(forceUpdate != null && forceUpdate);
+            .setLazyRegeneration(true);
```
JoshAlbrecht23, line 1295: Same as the other comment, just want to make sure we are intentionally keeping this new line

JoshAlbrecht23 (changes requested):
I noticed a couple of lines that were left from the commit we are reverting and just want to make sure it is intentional. 

It also looks like some missing changes in src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java were not reverted and I want to double check that it was intentional:
https://github.com/candlepin/candlepin/blob/c9ed71f4b920857564f78e6cb5eeaf8fbf404b49/src/main/java/org/candlepin/async/tasks/RefreshPoolsForProductJob.java#L77-L78

Also, should we add back the trace-level logging of upstream subscriptions and product data from the refresh operation?

src/main/java/org/candlepin/resource/OwnerResource.java
Ceiu: The lazy regeneration wasn't added as a result of this logic, it was simply a refactor to move it out of the refresher's constructor and into more of a builder pattern. That pattern has been retained and extended slightly with Ondrej's work to break some of those circular dependencies.

Ceiu:
The trace-level logging was a hack/workaround to get upstream data for duplicating a refresh when debugging issues in production. With the addition of direct fetching of upstream data, we no longer have a use for it here. That it was removed with the commit being reverted was happenstance due to it being in proximity of the other changes. It has no value even with the revert.

As far as the lazy regen thing goes -- that's no different than where it was originally, other than it moving to a separate method instead of being part of the refresher's constructor. The important bit is removing the forced-update param and mutator.

src/main/java/org/candlepin/async/tasks/RefreshPoolsJob.java
Ceiu: Commented elsewhere, but this was added as part of a refactor that we're retaining despite reverting the other surrounding logic.

PR #4110, opened 2023-06-29
Fix fork checks for dependabot

Januson:


.github/workflows/pr_verification.yml
```diff
@@ -15,6 +15,35 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+
+  dump_contexts_to_log:
```
nikosmoum, line 19: I assume this is just for debugging and will be removed before merge?

Januson: Yes. I forgot to remove it. Done.

nikosmoum (approved):

PR #4112, opened 2023-06-29
[M] CANDLEPIN-613: Migrated the i18n job to GH actions

JoshAlbrecht23:
GH actions job of the internationalization job running in on this branch: 
https://github.com/candlepin/candlepin/actions/runs/5523927019/jobs/10075523568

### DIFF
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -1,0 +1,64 @@
+---
+name: I18n
+on:
+  schedule:
+    - cron: '0 7 * * 0,4'
+  workflow_dispatch:
+
+env:
+  JAVA_DISTRIBUTION: 'temurin'
+
+jobs:
+  i18n:
+    name: i18n
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:38
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: [main, candlepin-4.1-HOTFIX, candlepin-4.2-HOTFIX]
+        include:
+          - branch: main
+            java-version: 17
+          - branch: candlepin-4.1-HOTFIX
+            java-version: 11
+          - branch: candlepin-4.2-HOTFIX
+            java-version: 11
+
+    steps:
+      - name: Mask secrets
+        shell: bash
+        run: |
+          echo "::add-mask::${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Install dependencies
+        shell: bash
+        run: dnf --setopt install_weak_deps=False install -y gettext jss git
+
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: ${{ env.JAVA_DISTRIBUTION }}
+          java-version: ${{ matrix.java-version }}
+
+      - name: Add safe directory
+        shell: bash
+        run: git config --system --add safe.directory $GITHUB_WORKSPACE
+
+      - name: Run internationlization
+        shell: bash
+        run: |
+          chmod +x ./.github/scripts/i18n.sh
+          ./.github/scripts/i18n.sh
+        env:
+          GIT_AUTHOR_EMAIL: '3229038+candlepin-bot@users.noreply.github.com'
+          GIT_AUTHOR_NAME: 'candlepin-bot'
+          GIT_BRANCH: ${{ matrix.branch }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
--- a/.github/scripts/i18n.sh
+++ b/.github/scripts/i18n.sh
@@ -1,0 +1,29 @@
+#!/bin/bash -x
+
+evalrc() {
+  if [ "$1" -ne "0" ]; then
+    echo "$2"
+    exit $1
+  fi
+}
+
+#Execute the gettext task
+./gradlew gettext
+evalrc $? "Gettext was not successful for branch $GIT_BRANCH."
+
+files=$(git diff | egrep -v -e '^( |\+#|\-#|@@|\+\+\+|\-\-\-|diff|index)' -e 'X-Generator' -e 'POT-Creation-Date')
+
+if [ ! -z "$files" ]; then
+
+#Code to commit the updated template file
+git add po/keys.pot
+evalrc $? "Git add file was not successful for branch $GIT_BRANCH."
+
+echo "Committing and pushing the template file."
+git -c "user.name=$GIT_AUTHOR_NAME" -c "user.email=$GIT_AUTHOR_EMAIL" commit -m "updated po/keys.pot template"
+evalrc $? "Git commit was not successful for branch $GIT_BRANCH."
+
+git push https://${GITHUB_TOKEN}@github.com/candlepin/candlepin $GIT_BRANCH
+evalrc $? "Git push was not successful for branch $GIT_BRANCH."
+
+fi


.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Set up Java 17
+        uses: actions/setup-java@v3
+        with:
+          distribution: ${{ env.JAVA_DISTRIBUTION }}
+          java-version: 17
+
+      - name: Add safe directory
+        shell: bash
+        run: git config --system --add safe.directory /__w/candlepin/candlepin
```
JoshAlbrecht23, line 47: This was added because of the following known issue actions/checkout issue with containers.
https://github.com/actions/checkout/issues/1169

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/setup-java@v3
+        with:
+          distribution: ${{ env.JAVA_DISTRIBUTION }}
+          java-version: 11
+
+      - name: Set up Java 17
+        uses: actions/setup-java@v3
+        with:
+          distribution: ${{ env.JAVA_DISTRIBUTION }}
+          java-version: 17
```
JoshAlbrecht23, line 43: Setting up Java 11 and Java 17 because the older code lines are still on Java 11

.github/workflows/continuous_integration.yml
ptoscano: better if you use only `/__w/*' (even only `*`), so forks are covered too

.github/workflows/continuous_integration.yml
ptoscano: wouldn't it be better if each branch had its own i18n job? this way:
- the job in each branch would only care about its own branch -- no need to support multiple java versions, or multiple layouts of the sources, etc
- disabling the i18n job for a branch means only changing/removing it only in that branch, with no risks to break others

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Add safe directory
+        shell: bash
+        run: git config --system --add safe.directory /__w/candlepin/candlepin
+
+      - name: Run internationlization
```
ptoscano, line 49:
```suggestion
      - name: Run i18n
```
(there was a typo in the name, and IMHO it can be shortened)

.github/workflows/continuous_integration.yml
nikosmoum: I agree with Pino. This multi-branch juggling is an artifact of the job source code living separately than the main candlepin repo. But now that the job is part of the main repo it should live naturally in each branch

.github/workflows/continuous_integration.yml
JoshAlbrecht23: Good catch! Updated

.github/workflows/continuous_integration.yml
JoshAlbrecht23: Updated!

.github/workflows/continuous_integration.yml
JoshAlbrecht23: I set the i18n jobs to run on the twice a week schedule. I also created separate jobs for each code line branch so that we can specify the setup, like Java version, that is specific to each branch.

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,135 @@
...
+        run: |
+          chmod +x ./.github/scripts/internationalization.sh
+          ./.github/scripts/internationalization.sh
+        env:
+          GIT_AUTHOR_EMAIL: '3229038+candlepin-bot@users.noreply.github.com'
+          GIT_AUTHOR_NAME: 'candlepin-bot'
+          GIT_BRANCH: 'main'
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  4.1_branch_i18n:
```
nikosmoum, line 53: I thought this would be done by simply adding the workflow in each separate branch, but it turns out scheduled workflows only run on the default branch (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule). So I guess this is the only way of doing this

.github/scripts/internationalization.sh
```diff
@@ -0,0 +1,55 @@
...
+./gradlew war
+evalrc $? "Compilation was not successful for branch $GIT_BRANCH."
+
+#Execute the gettext task
+./gradlew gettext
+evalrc $? "Gettext was not successful for branch $GIT_BRANCH."
+
+files=$(git diff | egrep -v -e '^( |\+#|\-#|@@|\+\+\+|\-\-\-|diff|index)' -e 'X-Generator' -e 'POT-Creation-Date')
+
+if [ ! -z "$files" ]; then
```
ptoscano, line 39: IMHO a simpler way could be:
- copy the `keys.pot` file before the message extraction
- after the extraction, compare the old with the new one excluding the creation date

I implemented something like that for virt-manager few years ago:
https://github.com/virt-manager/virt-manager/blob/7cd6151a212c4c477a947fe5a7f2b3363dd0dcbd/.github/workflows/translations.yml#L41-L58

.github/scripts/internationalization.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "Committing and pushing the template file."
+git -c "user.name=$GIT_AUTHOR_NAME" -c "user.email=$GIT_AUTHOR_EMAIL" commit -m "updated po/keys.pot template"
+evalrc $? "Git commit was not successful for branch $GIT_BRANCH."
+
+git push https://${GITHUB_TOKEN}@github.com/candlepin/candlepin $GIT_BRANCH
+evalrc $? "Git push was not successful for branch $GIT_BRANCH."
+
+else
+#To stash the unnecessary changes
+git stash
```
ptoscano, line 54: not sure what this is for:
- if nothing changed, there is nothing to stash anyway
- even if what changed is only the pot creation date, it does not matter anyway, as after this script runs the job ends

ptoscano (commented):
considering that in the past there were changes in how the messages are extracted (e.g. a different path of the `.pot` file), I wonder whether it would make sense to also have different per-branch versions of the `internationalization.sh` script, i.e. `internationalization-main.sh`, `internationalization-4.2.sh`, etc; this way, in case there are changes in `main`, other branches are really not affected

(also, please consider a shorter filename than `internationalization.sh`, which is long and sometimes it gets spelt incorrectly; maybe `i18n.sh`?)

.github/scripts/internationalization.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# This function converts candlepin version to a int value with (Zero) padding
+# which helps in comparing versions.
+# Examples -
+# 2.5.6  -> 2005006000
+# 3.2.0  -> 3002000000
+# 2.9.6  -> 2009006000
+get_simplified_version() {
+  echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }'
+}
```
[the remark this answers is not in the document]
JoshAlbrecht23: Removed

.github/scripts/internationalization.sh
```diff
@@ -0,0 +1,55 @@
...
+
+git fetch --all
+
+# Declare an associative array
+declare -A candlepin_java
```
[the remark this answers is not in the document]
JoshAlbrecht23: Removed

.github/scripts/internationalization.sh
```diff
@@ -0,0 +1,55 @@
...
+# which helps in comparing versions.
+# Examples -
+# 2.5.6  -> 2005006000
+# 3.2.0  -> 3002000000
+# 2.9.6  -> 2009006000
+get_simplified_version() {
+  echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }'
+}
+
+git fetch --all
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good point! Removed

.github/scripts/internationalization.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Declare an associative array
+declare -A candlepin_java
+
+#Clean the project
+./gradlew clean
+evalrc $? "Project cleaning was not successful."
```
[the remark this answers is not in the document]
JoshAlbrecht23: Removed

.github/scripts/internationalization.sh
JoshAlbrecht23: Yeah, I'm not sure what the purpose of this is either. It was in the original script and may have been used for stashing changes before changing to another branch since the original script would add a commit for all the code line branches.

JoshAlbrecht23:
> considering that in the past there were changes in how the messages are extracted (e.g. a different path of the `.pot` file), I wonder whether it would make sense to also have different per-branch versions of the `internationalization.sh` script, i.e. `internationalization-main.sh`, `internationalization-4.2.sh`, etc; this way, in case there are changes in `main`, other branches are really not affected
> 
> (also, please consider a shorter filename than `internationalization.sh`, which is long and sometimes it gets spelt incorrectly; maybe `i18n.sh`?)

I agree on renaming the script from `internationalization.sh` to `i18n.sh`. In regards to having different versions of the i18n.sh script per code line, I think that we should introduce a new version of the script when a branch requires it. If we were to do this today, we would have 3 identical scripts with different names being used by their respective GH actions job which seems like overkill for today's functionality.

.github/scripts/internationalization.sh
```diff
@@ -0,0 +1,55 @@
...
+
+#To generate the auto generated classes and files
+./gradlew war
+evalrc $? "Compilation was not successful for branch $GIT_BRANCH."
```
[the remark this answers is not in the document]
ptoscano: hm it seems still there?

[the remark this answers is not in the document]
JoshAlbrecht23: My bad, I missed this. Removed and retested!

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,135 @@
...
+        run: |
+          chmod +x ./.github/scripts/i18n.sh
+          ./.github/scripts/i18n.sh
+        env:
+          GIT_AUTHOR_EMAIL: '3229038+candlepin-bot@users.noreply.github.com'
+          GIT_AUTHOR_NAME: 'candlepin-bot'
+          GIT_BRANCH: 'candlepin-4.1-HOTFIX'
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  4.2_branch_i18n:
```
Januson, line 95: Just an idea. If we really wanted, we could have this as a reusable workflow in each branch. All that the main scheduled workflow would need to do is to call them from all supported branches.

Something like this.

```yaml
uses: {owner}/{repo}/.github/workflows/i18n.yml@{branch}
```

Januson (changes requested):
Looks good. Just a couple suggestions.

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,135 @@
...
+name: I18n
+on:
+  schedule:
+    - cron: '0 7 * * 0,4'
+
+env:
+  JAVA_DISTRIBUTION: 'temurin'
+
+jobs:
+  main_branch_i18n:
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good idea! Updated.

.github/workflows/i18n.yml
JoshAlbrecht23: I feel like this makes things harder than they need to be for not much gain as the code stands today. I could see this being pretty useful if a code line branch needs to handle i18n in a drastically different way. Then you could let that logic live in the code line branch and all the scheduled main branch would have to do is call it like you pointed out.

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,63 @@
+---
+name: I18n
+on:
```
nikosmoum, line 3: Can we add the `workflow_dispatch:` here, so that we can run it manually too?

nikosmoum (changes requested):

nikosmoum (approved):

nikosmoum:
Ignoring the jenkins infrastructure failure during the sonarqube task, since we already have sonarcloud for this check, and merging

PR #4115, opened 2023-07-03
[M] CANDLEPIN-531: Update RHEL dependency from pki-servlet-engine to tomcat

bakajstep:
- Change package from pki-servelet-engine to tomcat in candlepin.spec.tmpl and for cs9 images

### DIFF
--- a/candlepin.spec.tmpl
+++ b/candlepin.spec.tmpl
@@ -44,8 +44,9 @@ BuildRequires: selinux-policy-doc
 
 Requires: java-17 >= 1:17.0.0
 Requires: wget
-%if 0%{?fedora}
-Requires: tomcat
+# TODO: remove the version restriction on Tomcat once CP supports the Java EE namespace changes
+%if 0%{?fedora} || 0%{?rhel} >= 9
+Requires: tomcat < 1:10.0.0
 %else
 Requires: pki-servlet-engine
 %endif
--- a/ansible/roles/candlepin/tasks/el9/el9_base_deps.yml
+++ b/ansible/roles/candlepin/tasks/el9/el9_base_deps.yml
@@ -30,7 +30,7 @@
       - jss
       - make
       - openssl
-      - pki-servlet-engine
+      - tomcat
       - platform-python
       - python3
       - python3-libxml2
--- a/containers/candlepin-base-cs9/setup-devel-env.sh
+++ b/containers/candlepin-base-cs9/setup-devel-env.sh
@@ -23,7 +23,7 @@ PACKAGES=(
     jss
     mariadb
     openssl
-    pki-servlet-engine
+    tomcat
     postgresql
     procps
     python3-pip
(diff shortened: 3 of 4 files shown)


bakajstep:
I can only change the package for CS9 because CS8 still has the old name.

candlepin.spec.tmpl
```diff
@@ -44,11 +44,7 @@ BuildRequires: selinux-policy-doc
 
 Requires: java-17 >= 1:17.0.0
 Requires: wget
-%if 0%{?fedora}
-Requires: tomcat
-%else
-Requires: pki-servlet-engine
-%endif
+Requires: tomcat < 10
```
nikosmoum, line 47: After more discussion with Satellite/RHCS/tomcat folks, we decided not to make this update for RHEL8 or existing versions of Satellite. But we will make it for RHEL9.
Can we change this logic to be the following then?:
If (fedora or rhel >= 9)
  Requires: tomcat
else
  Requires: pki-servlet-engine

nikosmoum (changes requested):

candlepin.spec.tmpl
bakajstep: done

candlepin.spec.tmpl
```diff
@@ -44,8 +44,9 @@ BuildRequires: selinux-policy-doc
 
 Requires: java-17 >= 1:17.0.0
 Requires: wget
-%if 0%{?fedora}
-Requires: tomcat
+# TODO: remove the version restriction on Tomcat once CP supports the Java EE namespace changes
+%if (fedora or rhel >= 9)
```
ptoscano, line 48:
```suggestion
%if 0%{?fedora} or 0%{?rhel} >= 9
```

bakajstep: done

candlepin.spec.tmpl
```diff
@@ -44,8 +44,9 @@ BuildRequires: selinux-policy-doc
 
 Requires: java-17 >= 1:17.0.0
 Requires: wget
-%if 0%{?fedora}
-Requires: tomcat
+# TODO: remove the version restriction on Tomcat once CP supports the Java EE namespace changes
+%if 0%{?fedora} or 0%{?rhel} >= 9
```
nikosmoum, line 48: `or` is not an rpm spec operator. this should be `||`

nikosmoum (changes requested):
There's a change on the 'or' operator needed, but furthermore, this needs to be tested  on RHEL8/9 (both the building of the rpm, and installing it).

candlepin.spec.tmpl
bakajstep: done

bakajstep:
> There's a change on the 'or' operator needed, but furthermore, this needs to be tested on RHEL8/9 (both the building of the rpm, and installing it).

i just verify it that you can build it and install on RHEL8/9 using tito

nikosmoum (approved):

PR #4116, opened 2023-07-03
[4.2] CANDLEPIN-531: Update RHEL dependency from pki-servlet-engine to tomcat

bakajstep:
- Change package from pki-servelet-engine to tomcat in candlepin.spec.tmpl and for cs9 images


nikosmoum (commented):
After more discussion with Satellite/RHCS/tomcat folks, we decided not to make this update for RHEL8 or existing versions of Satellite. But we will make it for RHEL9.

PR #4120, opened 2023-07-04
CANDLEPIN-619: Re-run PR workflow when target branch gets pushed

bakajstep:
- Create a workflow that triggers when a PR is closed, and it will rerun all workflows associated with the affected branch

### DIFF
--- a/.github/workflows/closed_pr_workflow.yml
+++ b/.github/workflows/closed_pr_workflow.yml
@@ -1,0 +1,51 @@
+name: Workflow run on closed PR
+
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  rerun-workflows:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Mask secrets
+        shell: bash
+        run: |
+          echo "::add-mask::${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Trigger workflow runs
+        run: |
+          token="${{ secrets.GITHUB_TOKEN }}"
+          base="${{ github.base_ref }}"
+          
+          pr_head_shas=$(curl -L \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer $token"\
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/candlepin/candlepin/pulls |\
+            jq -r '.[] | select(.base.ref == $base) | .head.sha')
+          
+          for head_sha in $pr_head_shas; do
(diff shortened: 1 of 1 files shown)


.github/workflows/rerun_workflow_on_merge.yml
```diff
@@ -0,0 +1,28 @@
...
+  rerun-workflows:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Trigger workflow runs
+        run: |
+          token="${{ secrets.GITHUB_TOKEN }}"
```
JoshAlbrecht23, line 18: I think we should mask this so that it doesn't potentially leak in the logs

JoshAlbrecht23 (changes requested):

.github/workflows/rerun_workflow_on_merge.yml
```diff
@@ -0,0 +1,28 @@
+name: Rerun Workflows on PR Merge
```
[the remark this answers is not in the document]
bakajstep: that's a good idea. Done

.github/workflows/rerun_workflow_on_merge.yml
bakajstep: done

bakajstep:
We are not able to test this until we merge it to main so i write script to test it. So whoever's going to go through it can give it a try.
```shell
#!/bin/bash

pr_head_shas=$(gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" /repos/candlepin/candlepin/pulls | jq -r '.[] | select(.base.ref == "main") | .head.sha')

for head_sha in $pr_head_shas; do
    runner_id=$(gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" /repos/candlepin/candlepin/actions/workflows/pr_verification.yml/runs?head_sha=$head_sha | jq -r '.workflow_runs[] | .id')
    workflow_ids+=($runner_id)
done


for id in "${workflow_ids[@]}"
do
    gh api \
  --method POST \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  /repos/candlepin/candlepin/actions/runs/$id/rerun
done
``` 

You must be logged with account in candlepin organization to make request the gh api of the candlepin organization.
[https://cli.github.com/manual/gh_auth_login](url)

nikosmoum:
We need to discuss an alternative to this during retro

nikosmoum:
Closing this in favor of using GH Merge Queues and adding a step to merge the target branch in the feature branch before running the CI checks

PR #4122, opened 2023-07-06
CANDLEPIN-323: Update exception handling for adapters [M]

wottop:
- Brings the translated messages to Candlepin away from adapter implementation code.
- Exceptions from the adapters are qualified so that Candlepin can react to specific types.


wottop:
This will need to stay in draft until the accompanying changes are made in the service adapters. The specific exception changes in the adapters will need to wait until the adapters are first updated to CP 4.3.6.

Feedback on this branch in the meantime is appreciated.

Januson (changes requested):
Looks good. There are a few things to iron out before it's ready. Also two additional thoughts on the overall change.

- How thorough we want to be with testing? We should probably unit test all these mappings. 
- It might be better to use an exception mapper instead of mapping in place of occurrence. E.G. UserServiceExceptionMapper that would handle transformation of all UserServiceExceptions (See RuntimeExceptionMapper for an example). But that would probably require having those exceptions unchecked to avoid polluting the code with redundant throws.

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -68,12 +72,28 @@ public String cloudAuthorize(CloudRegistrationDTO cloudRegistrationDTO) {
...
-            throw new BadRequestException(e.getMessage(), e);
+        catch (CloudRegistrationServiceException e) {
+            if (e instanceof CloudRegistrationAuthorizationException) {
+                throw new NotAuthorizedException(e.getMessage());
+            }
+            else if (e instanceof CloudRegistrationMissingTypeException) {
+                throw new BadRequestException(this.i18n.tr(
+                    "Request is missing cloud provider type (e.g. amazon, gcp, azure)"));
+            }
+            else if (e instanceof CloudRegistrationBadTypeException) {
```
Januson, line 83: One more thought. Rather than catching and mapping, we could provide hosted with a set of exceptions that have the translation integrated inside. That way we wouldn't need to do any mapping on our side. All hosted exceptions would be translated. E.G.

```java
public class CloudRegistrationBadTypeException extends ProductServiceException {
    public CloudRegistrationBadTypeException(I18n i18n, String type) {
        super(i18n.tr("Unrecognized provider type in request: {}", type));
    }
}
```

src/main/java/org/candlepin/service/exception/user/UserMissingOwnerException.java
```diff
@@ -0,0 +1,74 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.user;
+
+
+/**
+ * The UserMissingOwnerException is used to reject a user validation action.
+ */
+public class UserMissingOwnerException extends UserServiceException {
```
redsaz, line 21: Adapters had (well, still has) been using an OwnerCurator in order to look up the full details of the Owner(s) that the user is a part of, as part of the user lookup. When the OwnerCurator returns back null for "nah, there's no owner by that key". then Adapters would either throw an error or return back an empty list of accessible owners depending on the method.

I posed a question in a google meet call with Candlepin folks, including @Ceiu about Adapters use of curators. It was determined that _*any use of curators by Adapters*_ is ill-informed, no good, bad, why-is-this-happening-ow-my-eyes.

So, at some point down the line, and hopefully soon, Adapters will no longer be using the owner curator to fill in the owner details when looking up a user, and will instead return an `OwnerInfo` (though if an `OwnerKeyInfo` was a thing I'd do that) which only has info for `key`. Which means that Adapters won't be erroring on missing owners, and it'll now be on core Candlepin to determine what to do when it gets back a user who belongs to an Owner that doesn't exist. Which means that this exception probably isn't needed, but not 100% sure on that. At the very least, eventually Adapters will never throw the exception, because in order to throw it, Adapters must be able to look up the Owner, and the only way it can do that is through Curator, which Adapters should not use.

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1277,11 +1278,17 @@ public List<SubscriptionDTO> getOwnerSubscriptions(String ownerKey) {
...
+                    owner = createOwnerFromDTO(new OwnerDTO().key(ownerKey).displayName(ownerKey));
+                }
+                else {
+                    throw new NotFoundException(
+                        i18n.tr("owner with key: {0} was not found.", ownerKey));
+                }
             }
-            else {
-                throw new NotFoundException(i18n.tr("owner with key: {0} was not found.", ownerKey));
+            catch (OwnerInvalidKeyFormatException e) {
```
redsaz, line 1290: `OwnerInvalidKeyFormatException` looks like it is intended for use by this Adapter code:

```java
  public boolean isOwnerKeyValidForCreation(String ownerKey) {
    try {
      LOGGER.debug("Checking if ownerkey={} is valid.", ownerKey);
      return userService.isOwnerKeyValid(Long.parseLong(ownerKey));
    } catch (NumberFormatException e) {
      LOGGER.debug("... ownerkey={} is invalid due to NumberFormatException.", ownerKey);
      throw new CandlepinAdapterException(ownerKey + " is not a valid OrgId", e);
    }
  }
```

I don't believe I ever was happy about the exception getting thrown by Adapters here. To me, it should have simply returned `false`, because that is the answer to the `isOwnerKeyValidForCreation` question.

With that in mind, are you okay if we get rid of OwnerInvalidKeyFormatException, and Adapters can not throw an exception here, and return `false` instead?

src/main/java/org/candlepin/service/exception/subscription/SubscriptionActivationNoTagException.java
```diff
@@ -0,0 +1,75 @@
...
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.subscription;
+
+
+/**
+ * The SubscriptionActivationNoTagException is used to reject an activation attempt for a
+ *  * subscription.
```
redsaz, line 20: Looks like a minor auto-formatter issue here, extra asterisk

src/main/java/org/candlepin/service/exception/subscription/SubscriptionSuccessfulActivationException.java
```diff
@@ -0,0 +1,75 @@
...
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.subscription;
+
+
+/**
+ * The SubscriptionSuccessfulActivationException is used to show the success of an activation attempt
+ * for a subscription.
+ */
+public class SubscriptionSuccessfulActivationException extends SubscriptionServiceException {
```
redsaz, line 22: I did a double-take here and asked "Wait, do the Adapters actually do this?" And sure enough, yes it does.

In fact, the `activateSubscription` method never returns, it always throws an exception. In this case, a successful exception. Which seems wacky to me.

We'll have more opportunities to improve this adapter later, I think we can keep all the exceptions. What are your thoughts on putting all these subscription activation exceptions in a sub package, say `org.candlepin.exception.subscription.activation` to keep them separate from the more traditional subscription exceptions?

(Another aside, what are thoughts on extracting the activation methods and exceptions out of the `SubscriptionServiceAdapter` and into a `SubscriptionActivationServiceAdapter`? This would be effort for a different, future ticket, not this one.)

src/main/java/org/candlepin/service/exception/subscription/SubscriptionUnknownActivationException.java
```diff
@@ -0,0 +1,75 @@
...
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.subscription;
+
+
+/**
+ * The SubscriptionUnknownActivationException is used to reject an activation attempt for a
+ *  * subscription.
+ */
+public class SubscriptionUnknownActivationException extends SubscriptionServiceException {
```
redsaz, line 22: This looks like another potential candidate to be replaced by a suggestion in an earlier comment, to have a generic `ServiceRuntimeException` instead. Then, in the SubscriptionResource block:

```java
           else if (e instanceof SubscriptionUnknownActivationException) {
                message = i18n.tr(
                    "The system is unable to redeem the requested subscription: {0}", dellAssetTag);
            }
```

Instead of catching `SubscriptionUnknownActivationException`. it can catch the `ServiceRuntimeException` and still provide the same error message.

src/main/java/org/candlepin/service/exception/subscription/SubscriptionUnknownRetrievalException.java
```diff
@@ -0,0 +1,75 @@
...
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.subscription;
+
+
+/**
+ * The SubscriptionUnknownRetrievalException is used to reject an activation attempt for a
+ *  * subscription.
+ */
+public class SubscriptionUnknownRetrievalException extends SubscriptionServiceException {
```
redsaz, line 22: Parroting earlier comments, this could be replaced by a potential `ServiceRuntimeException` or `UnknownServiceException` general case exception, and the catch block that receives the exceptions can still provide the correct logging for the case.

src/main/java/org/candlepin/service/exception/subscription/SubscriptionUnknownTermsException.java
```diff
@@ -0,0 +1,75 @@
...
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.subscription;
+
+
+/**
+ * The SubscriptionUnknownTermsException is used to reject an activation attempt for a
+ *  * subscription.
+ */
+public class SubscriptionUnknownTermsException extends SubscriptionServiceException {
```
redsaz, line 22: Dittoing earlier comments, this could be replaced by a potential ServiceRuntimeException or UnknownServiceException general case exception, and the catch block that receives the exceptions can still provide the correct logging for the case.

src/main/java/org/candlepin/service/exception/user/UserMissingException.java
```diff
@@ -0,0 +1,74 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.user;
+
+
+/**
+ * The UserMissingException is used to reject a user validation action.
+ */
+public class UserMissingException extends UserServiceException {
```
redsaz, line 21: For now, having both a UserInvalidException and UserMissingException is okay. Today, Adapters **does** distinguish between invalid users (users with incorrect login+password combos) and users that do not exist (login was not found) for certain calls. However, I wonder if we should be returning UserInvalidException in either case, for security purposes.

src/main/java/org/candlepin/service/exception/user/UserRetrievalException.java
```diff
@@ -0,0 +1,74 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.user;
+
+
+/**
+ * The UserRetrievalException is used to reject a user retrieval action.
+ */
+public class UserRetrievalException extends UserServiceException {
```
redsaz, line 21: It doesn't appear that this class gets caught anywhere, could this be deleted?

src/main/java/org/candlepin/service/exception/user/UserUnknownRetrievalException.java
```diff
@@ -0,0 +1,74 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.user;
+
+
+/**
+ * The UserUnknownRetrievalException is used to reject a user retrieval action.
+ */
+public class UserUnknownRetrievalException extends UserServiceException {
```
redsaz, line 21: Dittoing earlier comments, this could be replaced by a potential `ServiceRuntimeException` or `UnknownServiceException` general case exception, and the catch block that receives the exceptions can still provide the correct logging for the case.

redsaz (changes requested):
This does a good job covering all the different cases that Adapters was covering with its own custom exceptions and i18n translations. There were a couple of exceptions that might've let too much implementation details leak through, which I commented on. There were a few different `*UnknownException` exceptions that I thought could merge into a single `ServiceRuntimeException` since they would cover transient errors that are outside of Candlepin's control.

Out of curiosity, where did the i18n translations go? I don't see any .po files get changed/updated within this commit.

src/main/java/org/candlepin/resource/OwnerResource.java
Ceiu: That seems reasonable to me

src/main/java/org/candlepin/service/exception/product/ProductUnknownRetrievalException.java
```diff
@@ -0,0 +1,75 @@
...
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.product;
+
+
+/**
+ * The ProductUnknownRetrievalException is used to reject a product retrieval for an unexpected
+ * reason.
+ */
+public class ProductUnknownRetrievalException extends ProductServiceException {
```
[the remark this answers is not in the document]
Ceiu: I would agree with this. All of our service exceptions should derive from a common exception. Maybe a pair of common exceptions for both checked and unchecked variations: ServiceException (checked) and ServiceRuntimeException (unchecked).

src/main/java/org/candlepin/service/exception/user/UserMissingException.java
Ceiu: This is an interesting one. I guess the question comes down to whether or not Candlepin ever needs to care about such a differentiation and act accordingly. If not, then the adapters throwing a single "user not found" exception is probably okay. Otherwise, it may make sense to leave that up to CP to kick back the same exception in both paths.

But to the larger point, I agree that we could be more limiting on the info we give to users via error messages.

nikosmoum:
The new exceptions added by Stepan for autoreg 2.0 potentially need an update here to adhere to the pattern we chose

src/main/java/org/candlepin/service/exception/product/ProductUnknownRetrievalException.java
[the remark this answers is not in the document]
wottop: I did want to keep the "UnknownServiceException" at least be categorized as to the service that threw it. Since the  end goal is to stop passing messages through exceptions, then this differentiation is useful.

wottop:
The .po files are updated just before a tagged release.

src/main/java/org/candlepin/service/exception/subscription/SubscriptionSuccessfulActivationException.java
wottop: Updated both ends such that there is no exception returned on a successful action.

src/main/java/org/candlepin/auth/BasicAuth.java
```diff
@@ -74,6 +84,46 @@ public Principal getPrincipal(HttpRequest httpRequest) {
                 }
             }
         }
+        catch (UserServiceException e) {
+            log.error("Error getting principal " + e);
+            if (e instanceof UserInvalidException) {
```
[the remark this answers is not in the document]
Ceiu: They should be. Linting takes a backseat to better code structure.

wottop:
This can go to final review now. I talked with Shayne and we feel it should be put in a tag by itself so we can get it to stage with the updated adapter code. If there is a big problem (which we don't expect) then it can be reverted. There is not another way to effectively smoke test it and we should see any issues quickly.

src/main/java/org/candlepin/service/exception/product/ProductUnknownRetrievalException.java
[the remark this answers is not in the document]
Ceiu: A quick look at the adapter level changes, I drafted up an exception tree that looks something like this:

```
ServiceAdapterException
    CloudRegistrationServiceException
        CloudRegistrationAuthorizationException
        MalformedCloudRegistrationException

    EntitlementCertServiceException

    IdentityCertServiceException

    OwnerServiceException

    ProductServiceException

    SubscriptionServiceException
        ExpiredTagException
        ExhaustedTagException (tag already used)
        InvalidTagException

    UserServiceException
        InvalidCredentialsException
        UnacceptedTermsException
        UnauthorizedException
        UserDisabledException
```

This may not be a complete capture of everything (some of the calls were a bit difficult to follow), but it gives the outline of what the exception tree should look like. The remaining question is whether or not we consider these true error states or actionable and thus whether or not they should be checked or unchecked exceptions. Sadly with a tree like this, we're kinda forced into one or the other barring some shenanigans I'm not aware of.

I'm kinda leaning toward unchecked exceptions, since the only thing we really do with them is translate them to the appropriate error code, and not catching one should probably result in a 500 level as a signal that we need to catch it. In either case we're going to need the try/catch blocks around these calls, so the boilerplate is still going to be a thing in many cases, but I can be convinced either way yet.

Ceiu (changes requested):
If time permits, we should go over this again next week.

src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -173,9 +184,18 @@ public void run() {
...
-            poolManager.refreshPoolsWithRegeneration(this.subAdapter, this.prodAdapter, owner, this.lazy);
-            recalculatePoolQuantitiesForOwner(owner);
-            updateRefreshDate(owner);
+            try {
+                poolManager.refreshPoolsWithRegeneration(this.subAdapter, this.prodAdapter, owner, this.lazy);
+                recalculatePoolQuantitiesForOwner(owner);
+                updateRefreshDate(owner);
+            }
+            catch (SubscriptionServiceException e) {
+                throw new CandlepinException(null,
```
Ceiu, line 193: We shouldn't be using CandlepinExceptions at this level, as we're not API-facing. These should either be rethrown directly or wrapped in something like a RuntimeException or something more specific if applicable.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2316,13 +2324,25 @@ public Response bind(
...
+            try {
+                if (subAdapter.hasUnacceptedSubscriptionTerms(owner.getKey())) {
+                    message = new ExceptionMessage(
+                        i18n.tr("You must first accept Red Hat''s Terms and conditions. Please visit {0}",
+                        "https://www.redhat.com/wapps/tnc/ackrequired?site=candlepin" +
+                        "&event=attachSubscription"));
+                    return Response.serverError().entity(message).build();
+                }
+            }
+            catch (SubscriptionServiceException e) {
```
Ceiu, line 2337: Does this need to be handled in a nested try-catch? On the surface it looks like we can avoid this nesting and rearranging of code via finally without it.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2412,6 +2432,10 @@ public Response bind(
                     "It is disabled for org \"{0}\" because of the hypervisor autobind setting.",
                     owner.getKey()), e);
             }
+            catch (ProductServiceException e) {
+                throw new BadRequestException(i18n.tr("Unable to auto-attach. The products" +
+                    " are not retrievable by their ID's"));
```
Ceiu, line 2437: No apostrophe on "IDs".

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1291,7 +1291,8 @@ public AsyncJobStatusDTO refreshPools(String ownerKey, Boolean autoCreateOwner)
                 owner = createOwnerFromDTO(new OwnerDTO().key(ownerKey).displayName(ownerKey));
             }
             else {
-                throw new NotFoundException(i18n.tr("owner with key: {0} was not found.", ownerKey));
+                throw new NotFoundException(
```
Ceiu, line 1294: Is the line break necessary here?

src/main/java/org/candlepin/service/exception/subscription/SubscriptionServiceExceptionMapper.java
```diff
@@ -0,0 +1,48 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service.exception.subscription;
+
+import org.xnap.commons.i18n.I18n;
+
+
+
+public class SubscriptionServiceExceptionMapper {
```
Ceiu, line 21: While I'm not a huge fan of these mappers, if we're going to have them at all, they should all function the same way. This mapper returns a string for the pre-translated message to be, presumably, wrapped in another exception; rather than just throwing the exception like the other mappers do.

So far this seems to be the odd one out, and given the point of abstraction out the handling of specific exceptions is to avoid boilerplate, this class should follow in the convention of returning (not throwing!) the appropriate exception established by the other mappers.

As a side note, since the mappers are throwing Candlepin or API-facing exceptions, these should only ever be used from the API handlers (typically "resource" classes). Additionally, the mappers shouldn't be forced to accept a bunch of state to build the error messages needed for the mapped exceptions -- that bit should be contained within the exception itself. I would expect the mapper call to be nothing more than:

`throw SubscriptionServiceExceptionMapper.map(exception);`

For the exceptions that wrap additional state, that should be present in their constructors. E.g. `throw new SubscriptionExpiredTagException(tag)`

src/main/java/org/candlepin/service/exception/subscription/SubscriptionServiceExceptionMapper.java
```diff
@@ -0,0 +1,48 @@
...
+    private SubscriptionServiceExceptionMapper() { };
+
+    public static String map(SubscriptionServiceException e, String dellAssetTag, String ownerKey,
+        I18n i18n) {
+        String message = "";
+        if (e instanceof SubscriptionActivationException) {
+            message = i18n.tr(
+                "No subscription was able to be activated with Dell service tag '{0}'.", dellAssetTag);
```
Ceiu, line 29: Adding onto the comment above, this handler runs directly into the problem these mappers can hit without care. `SubscriptionActivationException` is pretty generic, but because of the state required to be passed in with the mapping function itself, there's an assumption that this only occurs when registering with an asset tag, and that is reflected in the output message. This limits the scope of the mapper to only the one or two places where tag-based activation is invoked, somewhat defeating the point of abstracting this logic out.

Ceiu (changes requested):
This is coming along, but the handling of exceptions still needs a bit of work

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -61,19 +62,21 @@ public String cloudAuthorize(CloudRegistrationDTO cloudRegistrationDTO) {
 
         Principal principal = ResteasyContext.getContextData(Principal.class);
         CloudRegistrationData registrationData = getCloudRegistrationData(cloudRegistrationDTO);
+        if (registrationData.getType() == null) {
+            throw new BadRequestException(i18n.tr(
+                "Request is missing cloud provider type (e.g. amazon, gcp, azure)"));
+        }
```
Ceiu, line 68: Line break between the end of this block and start of the next

Ceiu (changes requested):
Getting there; just need a bit of refinement on the exception properties (and likely translation).

I imagine one of the reasons it was implemented the way it was is to avoid the tedium of all the conversion with the instanceof spam in the mappers. If that's accurate, we have a few options:

1) Java 17 gives us the instanceof "pattern matching" syntax to avoid the need for explicit conversion:
```
if (exception instanceof UserDisabledException ude) {
  System.out.println("User disabled: " + ude.getUsername());
}
```

2) Something a bit jank we could do as well -- though I'm not sure how to feel about it -- would be to use try-catch to cast the exceptions in the mapper:
```
public CandlepinException map(UserServiceException exception) {
  try {
    throw exception;
  }
  catch (UserDisabledException ude) {
    return new BadRequestException("user disabled: " + ude.getUsername());
  }
  ...
  catch (UserServiceException use) {
    return new BadRequestException("generic error msg");
  }
}
```

I think either method is going to come with some degree of pain and ugliness, since it's honestly a pretty janky way of dealing with exceptions. It really should be done on a case-by-case basis rather than abstracted out; but so long as we're going to be doing it this way, then we'll have to live with the pain that comes along with it.

Ceiu (changes requested):
The core logic changes look to be complete. At this point it's just stylistic and documentation cleanup.

The main comment of my last review is still pending as well -- we could be using newer Java features to avoid needing to explicitly cast the exceptions in the mappers.

Also, we should get another set of eyes on this to chime in on the mappers. For what it's worth, I did a quick count of their use and my results were:
- user service: 7
- cloud registration: 1
- subscription service: 1

Those mappers which are only used once could be removed and have the places using them manually handle the exceptions directly.

wottop:
I will squash down after approval.

src/main/java/org/candlepin/service/exception/ServiceException.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param cause
+     *  the cause (which is saved for later retrieval by the Throwable.getCause() method). A null
+     *  value is permitted, and indicates that the cause is nonexistent or unknown.
+     */
+    public ServiceException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public ServiceException setStatus(int status) {
```
Ceiu, line 75: Mutable properties on the exception again. Also what was this intended to be used for? Looks to be extraneous with the mapper.

Ceiu (changes requested):

src/main/java/org/candlepin/service/exception/ServiceException.java
```diff
@@ -72,12 +71,12 @@ public ServiceException(String message, Throwable cause) {
...
-    }
-
-    public int getStatus() {
-        return this.status;
+    /**
+     * Information needed for the generation of a translated message
+     *
+     * @return Object[] array of substitution data
+     */
+    public Object[] getDetails() {
```
Ceiu, line 79: Hrmm... if the mapper is reliant on stuff like this, we're better off with the gigabranch that the original had. Definitely would like to see something slightly more involved like a function or consumer so we don't need to add generic abstractions like this to the exception API. Something closer to this:

```
mappings.add(ProductServiceException.class, pse -> {
  return Pair.of(
    Status.BAD_REQUEST,
    i18n.tr("Unable to retrieve product \"{0}\" from ProductService.", pse.getProductId()));
})
```

We can then also use `.getOrDefault` to wrap up the generic message stuff as well:

```
Pair<Status, String> result = this.mappings.getOrDefault(exception.getClass(), 
  e -> Pair.of(Status.BAD_REQUEST, i18n.get().tr("Unknown service exception"));

return this.getDefaultBuilder(..., result.getLeft());
```

src/main/java/org/candlepin/exceptions/mappers/ServiceExceptionMapper.java
```diff
@@ -46,91 +48,84 @@
...
-        }
-        else if (exception instanceof UserServiceException) {
-            status = Status.UNAUTHORIZED;
-            message = i18n.get().tr("Unexpected error from User Service: {0}", exception.getMessage());
-        }
-        else if (exception instanceof ProductServiceException pse) {
-            status = Status.BAD_REQUEST;
-            message = i18n.get().tr("Unable to retrieve proudct \"{0}\" from ProductService",
-                pse.getProductId());
+        Pair<Status, String> result = mappings.get(exception.getClass());
```
Ceiu, line 118: Note that this lookup doesn't work with subclasses.

When/if we move to Java 21, we can start using the switch expressions to best wrap all of this up in a single mapping function, but for now we may need to either do a semi-recursive lookup, or at least document that subclasses of registered mappings aren't implicitly handled as one might expect.

@Januson thoughts on this?

Januson: I agree. Switch would make it a bit more concise, but until then, pattern matching with instanceOf should be good enough.

PR #4126, opened 2023-07-13
[M] CANDLEPIN-631: CloudRegistrationSpecTest not running in GH actions

JoshAlbrecht23:
The CloudRegistrationSpecTest requires that cloud auth be enabled to run the spec tests. Since the docker-compose files do not set this configuration, the capability is not added and the tests are skipped. We need to ensure that the following configuration is set for Hosted spec tests:

candlepin.auth.cloud.enable=true


JoshAlbrecht23:
I confirmed that the CloudRegistrationSpecTest are now being run for Hosted spec tests.

PR #4131, opened 2023-07-18
[M] Upgrade gradle-build-action to 2.6.1

nikosmoum:


ptoscano:
hm if `v2` actually points to latest 2.x tag, wouldn't that use `v2.6.1` automatically already?

nikosmoum:
> hm if `v2` actually points to latest 2.x tag, wouldn't that use `v2.6.1` automatically already?

Does setting the major version only point to the latest x.y.z? I couldn't find any documentation on this.
The reason I did this is because I keep getting a notification email called 'GH security alert digest', that points out:
```
gradle/gradle-build-action | Version < 2.4.2 | Upgrade to ~> 2.4.2
```
so I assumed when setting `@v2`, means `@v2.0.0`

ptoscano:
> > hm if `v2` actually points to latest 2.x tag, wouldn't that use `v2.6.1` automatically already?
> 
> Does setting the major version only point to the latest x.y.z? I couldn't find any documentation on this.

It seems to be a common practice for actions, from what I saw so far. Some actions (eg the one for Python's `black` tool) have even a `stable`/`latest` tag that is kept up-to-date with whichever is the latest or stable version.

In case of `gradle/gradle-build-action`:
```bash
$ git clone https://github.com/gradle/gradle-build-action.git
Cloning into 'gradle-build-action'...
[...]
$ cd gradle-build-action
gradle-build-action$ PAGER= git show v2
tag v2
Tagger: daz <daz@gradle.com>
Date:   Mon Jul 17 16:05:12 2023 -0600

v2.6.1

commit 915a66c096a03101667f9df2e56c9efef558b165 (tag: v2.6.1, tag: v2)
Author: daz <daz@gradle.com>
Date:   Mon Jul 17 15:46:14 2023 -0600

    Bump dependency-graph version number

```

> The reason I did this is because I keep getting a notification email called 'GH security alert digest', that points out:
> 
> ```
> gradle/gradle-build-action | Version < 2.4.2 | Upgrade to ~> 2.4.2
> ```
> 
> so I assumed when setting `@v2`, means `@v2.0.0`

Yes, I noticed that too; I think the GH security scanner cannot know that the action actually used is already a fixed one, since it will depend on that `v2` represents when a workflow runs.

I also see that in some projects people pin to exact versions rather than these "generic" tags, or even pin to git commit hashes. Considering that there is dependabot in use for updating also the actions, this has the downside of adding more dependabot PRs.

nikosmoum:
ack, thanks for pointing this out. Closing this PR then!

PR #4134, opened 2023-07-20
CANDLEPIN-598: Remove uses of property injection

wottop:


src/main/java/org/candlepin/resteasy/filter/StoreFactory.java
```diff
@@ -59,14 +60,29 @@ public class StoreFactory {
...
+        ));
+        storeMap.put(Product.class, new ProductStore(
+            injector.getInstance(ProductCurator.class)
+        ));
+        storeMap.put(AsyncJobStatus.class, new AsyncJobStatusStore(
+            injector.getInstance(AsyncJobStatusCurator.class)
+        ));
 
         for (EntityStore<? extends Persisted> store : storeMap.values()) {
             injector.injectMembers(store);
```
Ceiu, line 88: This is no longer necessary (or recommended) if we're removing property injection.

src/main/java/org/candlepin/resteasy/filter/StoreFactory.java
```diff
@@ -59,14 +60,29 @@ public class StoreFactory {
     @Inject
     public StoreFactory(Injector injector) {
         storeMap.put(Owner.class, new OwnerStore());
-        storeMap.put(Environment.class, new EnvironmentStore());
-        storeMap.put(Consumer.class, new ConsumerStore());
-        storeMap.put(Entitlement.class, new EntitlementStore());
-        storeMap.put(Pool.class, new PoolStore());
+        storeMap.put(Environment.class, new EnvironmentStore(
```
Ceiu, line 63: If we've updated the classes below to use constructor injection, we could have the injector provide the store classes directly rather than still building them ourselves.

src/main/java/org/candlepin/resteasy/filter/StoreFactory.java
```diff
@@ -15,6 +15,7 @@
 package org.candlepin.resteasy.filter;
```
Ceiu, line 15: OwnerStore is still using member injection on line 107.

(Sorry for weird comment placement. Github won't let me comment on the line directly)

src/test/java/org/candlepin/controller/ContentAccessManagerDBTest.java
```diff
@@ -62,15 +59,22 @@ public class ContentAccessManagerDBTest extends DatabaseTestFixture {
...
+    private ContentAccessCertificateCurator caCertCurator;
+    private KeyPairDataCurator keyPairDataCurator;
 
     private PKIUtility pkiUtility;
     private ObjectMapper objMapper;
     private X509V3ExtensionUtil x509V3ExtensionUtil;
 
     private EventSink mockEventSink;
 
+    @BeforeEach
```
Ceiu, line 71: Have you verified if this is triggering a double-invocation of it? It being an override of a method that's already flagged with `@BeforeEach` makes this redundant or dangerous.

src/test/java/org/candlepin/model/CertificateTest.java
```diff
@@ -19,16 +19,24 @@
...
 
-import javax.inject.Inject;
 
 public class CertificateTest extends DatabaseTestFixture {
-    @Inject private SubscriptionsCertificateCurator certificateCurator;
-    @Inject private CertificateSerialCurator certSerialCurator;
+    private SubscriptionsCertificateCurator certificateCurator;
+    private CertificateSerialCurator certSerialCurator;
+
+    @BeforeEach
```
Ceiu, line 33: Same question here, and in several other files

Ceiu (changes requested):

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -178,19 +178,32 @@ public Map<String, Collection<String>> getFacts() {
...
-    @Inject private PrincipalProvider principalProvider;
-
-    public ConsumerCurator() {
+    private EntitlementCurator entitlementCurator;
+    private ConsumerTypeCurator consumerTypeCurator;
+    private ContentAccessCertificateCurator contentAccessCertificateCurator;
+    private CertificateSerialCurator certificateSerialCurator;
+    private DeletedConsumerCurator deletedConsumerCurator;
+    private FactValidator factValidator;
+    private IdentityCertificateCurator identityCertificateCurator;
```
Ceiu, line 187: Sonarqube is identifying this and a couple other values as unused and safe for removal. Given that we're updating the constructor's input, this seems like a good time to remove them entirely.

src/main/java/org/candlepin/resteasy/filter/StoreFactory.java
wottop: I believe the use of static class OwnerStore (and UserStore) keeps us from using constructor injection.

src/test/java/org/candlepin/model/CertificateTest.java
wottop: I have put all getInstance for curators in the DatabaseTestFixture (plus some other common ones). The get for the Resources and lower use stuff is per test.

src/main/java/org/candlepin/resteasy/filter/StoreFactory.java
wottop: I tried this and got "[Guice/InjectInnerClass]: Injecting into inner classes is not supported."

Ceiu: gross.

Ceiu: If we change these to static classes (as they probably should be), does that make Guice happier with the situation, or does it still get angry and refuse to help?

wottop: It looks like that is it.

src/test/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapterTest.java
```diff
@@ -995,8 +1002,8 @@ public void testPrepareV1ExtensionsBrandedProduct() {
         products.add(product);
         setupEntitlements(ARCH_LABEL, "1.0");
 
-        Set<X509ExtensionWrapper> extensions =
-            certServiceAdapter.prepareV1Extensions(products, pool, consumer,
+        Set<X509ExtensionWrapper> extensions = certServiceAdapter.prepareV1Extensions(products, pool,
```
Ceiu, line 1005: Another example of a busted spotless rule

src/test/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapterTest.java
```diff
@@ -1268,8 +1272,8 @@ public void testPrepareV3EntitlementData() throws IOException {
         assertEquals(((Integer) order.get("quantity")).intValue(), (long) subscription.getQuantity());
         assertNotNull(order.get("start"));
         assertNotNull(order.get("end"));
-//        assertEquals(order.get("contract"), subscription.getContractNumber());
-//        assertEquals(order.get("account"), subscription.getAccountNumber());
+// assertEquals(order.get("contract"), subscription.getContractNumber());
```
Ceiu, line 1275: and another

src/test/java/org/candlepin/service/impl/DefaultUserServiceAdapterTest.java
```diff
@@ -210,115 +205,114 @@ public void testUpdating() {
...
-    //     Set<Owner> owners = user.getOwners(null, Access.ALL);
-    //     assertEquals(0, owners.size());
-    //     user.addPermissions(new TestPermission(owner1));
-    //     user.addPermissions(new TestPermission(owner2));
-
-    //     // Adding the new permissions should give us access
-    //     // to both new owners
-    //     owners = user.getOwners(null, Access.ALL);
-    //     assertEquals(2, owners.size());
+    // String username = "TESTUSER";
```
Ceiu, line 210: Utterly destroyed comment block

Ceiu (changes requested):
By and large I'm okay with the logic changes here. The linter is doing some very dumb things, however, and there was one other thing to sort out with Guice in terms of inner class compatibility.

src/test/java/org/candlepin/async/tasks/UndoImportsJobTest.java
```diff
@@ -177,22 +169,22 @@ protected Pool createPool(String name, Owner owner, boolean keepSourceSub, PoolT
...
 
         switch (type) {
-            // TODO: Others as necessary
+        // TODO: Others as necessary
 
-            case ENTITLEMENT_DERIVED:
-                pool.setAttribute(Pool.Attributes.DERIVED_POOL, "true");
-                consumer = this.createConsumer(owner);
-                entitlement = this.createEntitlement(owner, consumer, pool, null);
+        case ENTITLEMENT_DERIVED:
```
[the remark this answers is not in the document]
Januson: Instead of ignoring the linter, we should be fixing the rules. In this case "indent_switchstatements_compare_to_switch" should be set to true.

src/test/java/org/candlepin/model/EntitlementCuratorTest.java
```diff
@@ -128,17 +126,15 @@ public void init() throws Exception {
...
-     *     (1) ent.getPool().getEntitlements().remove(ent);
-     *     (2) Hibernate.initialize(ent.getPool().getEntitlements());
-     * The problem is that remove() in (1) will not initialize the collection because
-     * it is EXTRA LAZY. Then (2) will initialize without removed entitlement
-     * Then when owning consumer c is being deleted, hibernate will recreate
-     * the entitlements from the initialized collection Pool.entitlements and
-     * subsequent delete of Consumer will cause foreign key exception.
+     * Reproducer of EXTRA Lazy problem. This test is here mainly to demonstrate the problem. The
+     * problem is that Pool.entitlements is an extra lazy collection and at the same time it is
+     * cascading CREATE. The most important lines in this method are: (1)
```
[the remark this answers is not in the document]
Januson: While the existing formatting looks better, it is not a valid Javadoc formatting. We should fix the formatting not ignore the linter. If we want to keep invalid formatting, I think we can turn off the comment formatting.

src/test/java/org/candlepin/model/PoolCuratorTest.java
```diff
@@ -1988,7 +1981,7 @@ public void testUpdateQuantityColumnsOnPool() {
 
         Pool pool = createPool(owner, product, 20L,
             TestUtil.createDate(2010, 3, 2), TestUtil.createDate(
-            Calendar.getInstance().get(Calendar.YEAR) + 1, 3, 2));
+                Calendar.getInstance().get(Calendar.YEAR) + 1, 3, 2));
```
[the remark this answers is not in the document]
Januson: This one is correct. Each line continuation should be indented. This rule differs from checkstyle, because checkstyle couldn't handle more then one line break for a single row of code.

src/test/java/org/candlepin/service/impl/DefaultUserServiceAdapterTest.java
Januson: Tests should be disabled, not commented out.

src/test/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapterTest.java
Januson: To prevent this one, we can enable joining lines, but we would loose all control over line splitting. I disabled the rule to give us option to have custom line breaks.

src/test/java/org/candlepin/async/tasks/UndoImportsJobTest.java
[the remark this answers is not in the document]
Ceiu: Indeed, but that's far outside the scope of this PR. Those corrections should be done in another PR and then we can either rebase this against that once it's merged, or ignore it and manually evaluate it for the purposes of this review.

src/test/java/org/candlepin/model/PoolCuratorTest.java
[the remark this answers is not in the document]
Ceiu: Why would we want an additional intent to process the continuation of an arbitrary parameter that itself is only being line-broken due to the same processing rule leading to the first line break?

The fact that we're in the middle of a nested call is entirely incidental, and doesn't make this blob of code any more readable. It's no more or less informative or arbitrary than the `createDate` on the untouched line that *isn't* broken up, or the subsequent call to `.getInstance().get(...)` and the math that follows. It's just another indentation because a line break happened to occur where it did.

Fun thought: would it still require an indent if the line were to change to read `TestUtil\n.createDate(`?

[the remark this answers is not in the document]
Ceiu: As a follow-on here, a better approach is to break this up into individual named parameters, or build the pool directly instead of using a long multi-parameter call that tends to run into this formatting issue.

src/test/java/org/candlepin/model/EntitlementCuratorTest.java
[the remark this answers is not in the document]
Ceiu: What Spotless is trying to do is no more or less valid in terms of Javadoc syntax [1] than what we already have. If the argument is that our lists and whatnot should be properly HTML formatted, fine, but we're not getting that out of Spotless, either. The output may be the same, but as-written it's at least readable in code and to many IDE parsers.

What I absolutely will not accept is reformatting of comments from one arbitrary syntax or style to another arbitrary syntax or style that doesn't impart at least some objectively tangible improvement. This particular example, and several others I've called out, does not improve the general output when rendered as HTML, but makes it worse to read for those looking at the code directly or interpreters or toolings that process the text literally.

[1] https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html

Ceiu (approved):

PR #4135, opened 2023-07-20
CANDLEPIN-550: Created build stage for dev Candlepin image

JoshAlbrecht23:
The following commands are an example on how you can build and stand up the new Candlepin dev container.

To build the dev image run:
```
docker build --target development -f ./containers/release.Containerfile -t <image name>:<tag> . \
	--build-arg WAR_FILE=<add location of the war file>
	--no-cache
```

Start Postgres container:
```
docker run --name postgres -itd  \
	-e POSTGRES_USER=candlepin \
	-e POSTGRES_PASSWORD=candlepin \
	-e POSTGRES_DB=candlepin \
	-e POSTGRES_HOST_AUTH_METHOD=trust \
	--network host \
	postgres:latest
```

Start Candlepin container:
```
docker run --name candlepin -itd  \
	--network host \
	<image name>:<tag>
```


.github/workflows/pr_verification.yml
```diff
@@ -327,7 +327,6 @@ jobs:
         shell: bash
         run: |
           echo "CANDLEPIN_IMAGE=ghcr.io/candlepin/gha_build:pr-${{ github.event.pull_request.number }}" >> ./.github/containers/.env
-          echo "SOURCE_LOCATION=${{ github.workspace }}" >> ./.github/containers/.env
```
JoshAlbrecht23, line 330: This was unused

wottop:
I see the following 7 times during the first step. Cause for concern?

WARN[0078] Failed to mount subscriptions, skipping entry in /usr/share/containers/mounts.conf: getting host subscription data: failed to read subscriptions from "/usr/share/rhel/secrets": open /usr/share/rhel/secrets/redhat.repo: permission denied

JoshAlbrecht23:
> I see the following 7 times during the first step. Cause for concern?

> WARN[0078] Failed to mount subscriptions, skipping entry in /usr/share/containers/mounts.conf: getting host subscription data: failed to read subscriptions from "/usr/share/rhel/secrets": open /usr/share/rhel/secrets/redhat.repo: permission denied

So, I think this is a rootless podman configuration issue and not an issue with the Dockerfile. From the following Bugzilla, it looks like it could be resolved by running the command `touch ~/.config/containers/mounts.conf`.

https://bugzilla.redhat.com/show_bug.cgi?id=1874621

Januson (commented):
Image looks alright. Just one question. It surprised me that development stage is independent of production stage. Couldn't we use production as a base for development and only override the changes instead of pretty much duplicating the setup?

JoshAlbrecht23:
> Image looks alright. Just one question. It surprised me that development stage is independent of production stage. Couldn't we use production as a base for development and only override the changes instead of pretty much duplicating the setup?

Thanks for pointing this out! You are right, it makes a lot more sense to use the production stage as the base for the development stage. I updated the development stage with the changes!

PR #4139, opened 2023-07-21
[M] Added configuration for gitleaks

Ceiu:
- Added the .gitleaks.toml file for configuring gitleaks scans
- Added a number of paths to the configuration to ignore some erroneous detections of "secrets" from our tests or obsoleted code


JoshAlbrecht23 (approved):
Looks good! I verified that leaks are no longer detected using gitleaks

PR #4150, opened 2023-07-25
CANDLEPIN-619: Run workflow against actual base branch

bakajstep:


.github/workflows/pr_verification.yml
```diff
@@ -26,14 +27,26 @@ jobs:
         shell: bash
         run: |
           echo "::add-mask::${{ secrets.GITHUB_TOKEN }}"
+          
+      - name: Install git
+        shell: bash
+        run: dnf --setopt install_weak_deps=False install -y git
```
nikosmoum, line 33: there is already a 'dnf install' step later on. can't we re-use that one? or if it needs to be before the `checkout@v3` step, then move it before it, and add git to the list of packages to install?

bakajstep: yes, it just need to be before checkout, because if git is not installed it will downoald just files and it is not repository

bakajstep:
You can check that rebase works by looking at #4165 i create that branch before i broke test. After that i open 
PR againts this branch and the test is also failing there.

bakajstep:
I didn't put git install in the composite action because we have to install git before using actions/checkout. Otherwise, we'll just have the downloaded files, but it won't be a git repository. And we can't add actions/checkout to the composite action because we can't see it otherwise.

.github/actions/rebase/action.yml
```diff
@@ -0,0 +1,22 @@
...
+        git config --global user.email "ghaction@random.com"
+        git config --global user.name "ghaction"
+        
+        base_branch="${{ github.base_ref }}"
+        
+        # Check if this is a pull request and if the base branch has changed
+        if [ -n "$GITHUB_HEAD_REF" ] && [ "$GITHUB_HEAD_REF" != "$base_branch" ]; then
+          git remote add base_branch https://github.com/${GITHUB_REPOSITORY}.git
+          git fetch base_branch
+          git rebase base_branch/$base_branch
```
nikosmoum, line 19: this is a bit confusing to read here. can we leave the $base_branch variable as is, which will contain the value of the target branch (e.g. main), but rename the remote branch alias as `remote_base_branch` or `origin_base_branch`?.

Then this command will read a bit better: `git rebase remote_base_branch/$base_branch`

bakajstep: done

PR #4154, opened 2023-07-26
[M] CANDLEPIN-621: Added auto reg 2.0 instance metadata authentication

JoshAlbrecht23:


src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -271,19 +278,114 @@ public String generateRegistrationToken(Principal principal, CloudRegistrationIn
...
+        String productId = result.getProductId();
+        if (productId == null || productId.isEmpty() || productId.isBlank()) {
+            String errmsg = this.i18nProvider.get()
+                .tr("product id could not be resolved");
+
+            throw new CloudRegistrationAuthorizationException(errmsg);
+        }
+
+        String ownerKey = result.getOwnerKey();
+        boolean ownerExistsAndPoolExists = false;
```
Ceiu, line 349: nit: you could express this variable's purpose with the phrase `ownerAndPoolExist` and save the doubling of the word 'exists'.

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -271,19 +278,114 @@ public String generateRegistrationToken(Principal principal, CloudRegistrationIn
...
+        if (ownerKey != null && !ownerKey.isEmpty() && result.isEntitled()) {
+            Owner owner = ownerCurator.getByKey(ownerKey);
+            if (owner != null) {
+                ownerExistsAndPoolExists = !poolCurator.listByOwnerAndProduct(owner, productId).isEmpty();
```
Ceiu, line 353: It's not strictly necessary to do both lookups in this case. If the org doesn't exist, the pool lookup will fail even if the product does, making the owner fetching redundant.

Ceiu (changes requested):
Found some things on the first pass; will make another pass through it after this round of updates is done.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/CloudInstanceMetadata.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public String getAccountId() {
+        return this.accountId;
+    }
+
+    public String getOfferingId() {
+        return this.offeringId;
+    }
+
+    public String toJson(ObjectMapper mapper) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: This method is used in the CloudRegistrationSpecTest tests, specifically the tests for v2 auth

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -224,31 +232,30 @@ private UserPrincipal createCloudUserPrincipal(String subject, String ownerKey)
      * valid
      *
      * @param principal
-     *  the principal for which to generate the token
+     *     the principal for which to generate the token
```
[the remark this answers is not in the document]
JoshAlbrecht23: So, this additional spacing is being introduced by Spotless apply, and so we may need to adjust the Spotless configuration for this offset space count!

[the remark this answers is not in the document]
Ceiu: That was my guess, since a number of these were updated on methods you weren't otherwise touching or likely even looking at. And yeah, if that's the case, our spotless config needs an update.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/CloudInstanceMetadata.java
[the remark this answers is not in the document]
Ceiu: Unless there's a very compelling and specific reason, we should not have code for tests leak into the production side of things.

src/main/resources/db/changelog/20230710000000-create-anonymous-consumers.xml
```diff
@@ -0,0 +1,39 @@
...
+            <column name="id" type="varchar(32)">
+                <constraints primaryKey="true" />
+            </column>
+            <column name="uuid" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cloud_account_id" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cloud_provider_short_name" type="varchar(255)">
```
nikosmoum, line 22: The short name is always 3 or 4 characters long (AWS/ACN/AGOV/MSAZ/GCE). Let's make this column length of 10 or 15 to be sure, but 255 would be a waste

api/candlepin-api-spec.yaml
```diff
@@ -8293,6 +8302,19 @@ components:
         signature:
           type: string
 
+    CloudAuthenticationResultDTO:
+      description: Contains the result of authenticating the provided cloud registration information.
+      required:
+        - token
+      properties:
+        ownerId:
```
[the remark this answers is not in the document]
nikosmoum: Why are we adding an owner id/key here in the first place? We won't always have it to add it here, and even if we do, the clien't won't be able to do anything with it yet.

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -267,23 +275,122 @@ public String generateRegistrationToken(Principal principal, CloudRegistrationIn
...
+            // TODO: Schedule anonymous org creation job
+
+            AnonymousCloudConsumer anonConsumer = anonymousCloudConsumerCurator
+                .create(new AnonymousCloudConsumer(cloudAccountId, productId, cloudProviderShortName));
+            anonymousConsumerUuid = anonConsumer.getUuid();
+        }
+
+        String token = anonymousConsumerUuid == null ?
+            this.buildRegistrationToken(principal, ownerKey, TOKEN_TYPE) :
+            this.buildRegistrationToken(principal, cloudAccountId, ANON_TOKEN_TYPE);
```
nikosmoum, line 370: Can we abstain from adding the cloudAccountId here, and leaving the audience field empty? The client has no need for it, so the less data we expose the better.

src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return String.format("Anonymous Consumer [id: %s, uuid: %s, cloudAccountId: %s, productId: %s]",
+            this.getId(), this.getUuid(), this.getCloudAccountId(), this.getProductId());
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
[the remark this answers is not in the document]
Ceiu: I'm realizing the day after writing this that these are literally copy/pasted from the most recent refactor that I wrote.

Disregard.

api/candlepin-api-spec.yaml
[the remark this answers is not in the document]
JoshAlbrecht23: Updated to ownerKey

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/CloudInstanceMetadata.java
```diff
@@ -0,0 +1,101 @@
...
+        public Builder(CloudInstanceMetadata metadata) {
+            this.accountId = metadata.getAccountId();
+            this.offeringId = metadata.getOfferingId();
+        }
+
+        public String getAccountId() {
+            return this.accountId;
+        }
+
+        public CloudInstanceMetadata.Builder setAccountId(String accountId) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Using `public Builder` would work here! I removed the CloudInstanceMetadata class and added a helper method in the CloudRegistrationSpecTest class for generating the json for the cloud instance metadata.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/CloudInstanceMetadata.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Representation of a cloud instance metadata information provided by cloud providers.
+ */
+public class CloudInstanceMetadata {
+
+    public static final String ACCOUNT_ID_FIELD_NAME = "accountId";
+    public static final String CLOUD_OFFERING_ID_FIELD_NAME = "cloudOfferingId";
+
+    private String accountId;
```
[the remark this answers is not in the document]
JoshAlbrecht23: Yeah, I was making this class immutable with the builder pattern, but I left out some details to make the class truly immutable like you pointed out. However, the class didn't really need to be immutable!

I removed the CloudInstanceMetadata class and added a helper method in the CloudRegistrationSpecTest class for generating the json for the cloud instance metadata.

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -271,19 +278,114 @@ public String generateRegistrationToken(Principal principal, CloudRegistrationIn
...
+        String cloudAccountId = result.getCloudAccountId();
+        if (cloudAccountId == null || cloudAccountId.isEmpty() || cloudAccountId.isBlank()) {
+            String errmsg = this.i18nProvider.get()
+                .tr("cloud provider account could not be resolved");
+
+            throw new CloudRegistrationAuthorizationException(errmsg);
+        }
+
+        String cloudProviderShortName = result.getCloudProviderShortName();
+        if (cloudProviderShortName == null || cloudProviderShortName.isEmpty() ||
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good catch, removed throughout the class

src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+    @Size(max = CLOUD_ACCOUNT_ID_MAX_LENGTH)
+    @NotNull
+    private String cloudProviderShortName;
+
+    @Column(name = "product_id")
+    @Size(max = PRODUCT_ID_MAX_LENGTH)
+    @NotNull
+    private String productId;
+
+    public AnonymousCloudConsumer(String cloudAccountId, String productId, String cloudProviderShortName) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Removed

src/main/java/org/candlepin/model/AnonymousCloudConsumerCurator.java
```diff
@@ -0,0 +1,59 @@
...
+public class AnonymousCloudConsumerCurator extends AbstractHibernateCurator<AnonymousCloudConsumer> {
+
+    public AnonymousCloudConsumerCurator() {
+        super(AnonymousCloudConsumer.class);
+    }
+
+    @Transactional
+    @Override
+    public AnonymousCloudConsumer create(AnonymousCloudConsumer entity, boolean flush) {
+        entity.ensureUUID();
```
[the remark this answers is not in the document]
JoshAlbrecht23: Yeah, we shouldn't need this with the prepersist method! Removed

src/main/java/org/candlepin/model/AnonymousCloudConsumerCurator.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.inject.Singleton;
+
+@Singleton
+public class AnonymousCloudConsumerCurator extends AbstractHibernateCurator<AnonymousCloudConsumer> {
+
+    public AnonymousCloudConsumerCurator() {
+        super(AnonymousCloudConsumer.class);
+    }
+
+    @Transactional
```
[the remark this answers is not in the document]
JoshAlbrecht23: Removed

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -54,15 +53,23 @@ public CloudRegistrationResource(I18n i18n, CloudRegistrationAuth cloudRegistrat
...
 
         Principal principal = ResteasyContext.getContextData(Principal.class);
         CloudRegistrationData registrationData = getCloudRegistrationData(cloudRegistrationDTO);
         try {
-            return this.cloudRegistrationAuth.generateRegistrationToken(principal, registrationData);
+            if (version == null || version == 1) {
+                return this.cloudRegistrationAuth.generateRegistrationToken(principal, registrationData);
+            }
+            else if (version == 2) {
+                return this.cloudRegistrationAuth.generateRegistrationTokenV2(principal, registrationData);
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/service/model/CloudAuthenticationResult.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * The cloud authentication result contains the results of authenticating
+ * {@link CloudRegistrationInfo}.
+ */
+public interface CloudAuthenticationResult {
+
+    /**
+     * Retrieves the cloud account ID that is found in the {@link CloudRegistrationInfo} metadata.
+     *
+     * @return the cloud account ID that is found in the {@link CloudRegistrationInfo} metadata
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated throughout

src/main/resources/db/changelog/20230710000000-create-anonymous-consumers.xml
```diff
@@ -0,0 +1,39 @@
...
+        </createTable>
+    </changeSet>
+
+    <changeSet id="20230710000000-2" author="jalbrech">
+        <createIndex tableName="cp_anonymous_cloud_consumers" indexName="cp_anonymous_cloud_consumers_idx1">
+            <column name="uuid"/>
+        </createIndex>
+    </changeSet>
+
+</databaseChangeLog>
```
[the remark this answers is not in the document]
JoshAlbrecht23: Added

src/main/resources/db/changelog/20230710000000-create-anonymous-consumers.xml
```diff
@@ -0,0 +1,39 @@
...
+            <column name="cloud_account_id" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cloud_provider_short_name" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="product_id" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="created" type="DATETIME"/>
```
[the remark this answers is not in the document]
JoshAlbrecht23: By the `AnonymousCloudConsumer` model class extending `AbstractHibernateObject`, it appears that the DB table needs the created and updated column. Database operations fail when creating or updating AnonymousCloudConsumers if I remove the columns.  I did move the column definitions to the front of the listing.

src/main/resources/db/changelog/20230710000000-create-anonymous-consumers.xml
```diff
@@ -0,0 +1,39 @@
...
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+<!-- TODO: need to add the provider short name here and in the model/curator -->
+
+    <changeSet id="20230710000000-1" author="jalbrech">
+        <createTable tableName="cp_anonymous_cloud_consumers">
+            <column name="id" type="varchar(32)">
+                <constraints primaryKey="true" />
+            </column>
+            <column name="uuid" type="varchar(255)">
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
```diff
@@ -66,4 +76,99 @@ public String resolveCloudRegistrationData(CloudRegistrationInfo cloudRegInfo)
...
+
+        Set<String> subIds = this.datastore.productSubscriptionMap.get(productId);
+        if (subIds == null) {
+            return false;
+        }
+
+        boolean isEntitled = false;
+        for (String subId : subIds) {
+            SubscriptionInfo sub = this.datastore.getSubscription(subId);
+            isEntitled = ownerKey.equals(sub.getOwner().getKey());
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
```diff
@@ -66,4 +76,99 @@ public String resolveCloudRegistrationData(CloudRegistrationInfo cloudRegInfo)
...
+        }
+
+        if (cloudRegInfo.getMetadata() == null) {
+            throw new MalformedCloudRegistrationException(
+                "No metadata provided with the cloud registration info");
+        }
+
+        JsonNode metadata;
+        try {
+            metadata = OBJ_MAPPER.readTree(cloudRegInfo.getMetadata());
```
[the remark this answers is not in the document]
JoshAlbrecht23: Good catch, this does make life easier!! Updated

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
JoshAlbrecht23: Yeah, that is a much better name! Updated

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
JoshAlbrecht23: Nice!! I'd rather just make the one curator call. Updated

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/CloudInstanceMetadata.java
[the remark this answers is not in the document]
JoshAlbrecht23: Removed the CloudInstanceMetadata object for a helper method in CloudRegistrationSpecTest

src/main/resources/db/changelog/20230710000000-create-anonymous-consumers.xml
JoshAlbrecht23: Yeah, that is a good idea! Updated

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
JoshAlbrecht23: That makes sense, I updated to use an empty string for the audience field

src/main/java/org/candlepin/hostedtest/HostedTestDataStore.java
```diff
@@ -968,7 +968,21 @@ protected SubscriptionsCertificate convertSubscriptionCertificate(CertificateInf
         return null;
     }
 
+    protected void addCloudOfferIdToProductId(String cloudOfferId, String productId) {
```
Ceiu, line 971: Are offering IDs always 1:1 with product IDs?

Also as a nit: Cloud offering is the owner of this relationship, so the naming should be reversed. We're adding a product to a cloud offering, not the other way around. Additionally, so long as we're maintaining a 1:1 mapping, we're not really "adding" so much as we're setting or mapping.

src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
```diff
@@ -0,0 +1,226 @@
...
+
+    @Id
+    @GeneratedValue(generator = "system-uuid")
+    @GenericGenerator(name = "system-uuid", strategy = "uuid")
+    @Column(length = 32)
+    @NotNull
+    private String id;
+
+    @Column(unique = true)
+    @Size(max = 36)
```
Ceiu, line 60: Size literal in the declaration.

As a side note here, these annotations trigger hard stops rather than autofill or truncate, if they even work (we've done some work to pull them out IIRC, and I can't find the annotation validator anymore); and since we use Liquibase, they're not used to build the schema, either.

if we care about these restrictions, it might make sense to enforce them at the mutator stage rather than waiting until persist.

src/test/java/org/candlepin/model/AnonymousCloudConsumerCuratorTest.java
```diff
@@ -0,0 +1,171 @@
...
+            .returns(expected.getUuid(), AnonymousCloudConsumer::getUuid)
+            .returns(expected.getCloudAccountId(), AnonymousCloudConsumer::getCloudAccountId)
+            .returns(expected.getCloudInstanceId(), AnonymousCloudConsumer::getCloudInstanceId)
+            .returns(expected.getCloudProviderShortName(), AnonymousCloudConsumer::getCloudProviderShortName)
+            .returns(expected.getProductId(), AnonymousCloudConsumer::getProductId);
+    }
+
+    @ParameterizedTest(name = "{displayName} {index}: {0}")
+    @NullSource
+    @ValueSource(strings = { "", "  " })
```
Ceiu, line 112: This won't eliminate the need for the additional ValueSource here, but jUnit has a composite NullAndEmptySource that will cover the need to use a value source for the first value here.

api/candlepin-api-spec.yaml
```diff
@@ -557,7 +557,16 @@ paths:
       tags:
         - cloud_registration
       operationId: cloudAuthorize
```
nikosmoum, line 559: I don't know if openapi-generator allows it, but is it possible to add both the content types in the 200 response here to show the 2 different types of response? And maybe even update the `description` node to explain that when using autoreg 1 you get text/plain, and for autoreg 2 you get application/json. So then the `responses:` node would look like this:
```
      responses:
        200:
          description: Cloud authorization successful. When the version of authorization requested is 2, application/json is returned that contains a JWT token amongst other data. Otherwise, only a JWT token in text/plain format is returned.
          content:
            text/plain:
              schema:
                type: string
            application/json:
              schema:
                $ref: '#/components/schemas/CloudAuthenticationResultDTO'
```

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -267,23 +275,142 @@ public String generateRegistrationToken(Principal principal, CloudRegistrationIn
...
+                    .setCloudProviderShortName(cloudProviderShortName);
+
+                anonConsumer = anonymousCloudConsumerCurator.create(anonConsumer);
+            }
+
+            anonymousConsumerUuid = anonConsumer.getUuid();
+        }
+
+        String tokenType = anonymousConsumerUuid == null ? TOKEN_TYPE : ANON_TOKEN_TYPE;
+        String token = anonymousConsumerUuid == null ?
```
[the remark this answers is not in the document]
nikosmoum: If we're introducing a new token type, we have to validate/verify it separately than the current one. It's inevitable. For example in addition to validating the token type itself, we'd have to check if we actually have an AnonymousCloudConsumer with the uuid specified in it. I assume this change is not added yet because the PR is still a draft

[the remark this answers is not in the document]
nikosmoum: Actually, now I remember why that code isn't added yet. It's because it's part of the next jira card

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -267,23 +275,142 @@ public String generateRegistrationToken(Principal principal, CloudRegistrationIn
...
+            this.buildRegistrationToken(principal, ownerKey, tokenType) :
+            this.buildRegistrationToken(principal, "", tokenType);
```
nikosmoum, line 389: One thing that's missing in the anonymous token use case: the expiration of the anonymous token should be set at 2 days; not at whatever the current token is using. The current token has a config property (ConfigProperties.JWT_TOKEN_TTL) that gets fed into this class. Perhaps we need a separate property called something like ANON_JWT_TOKEN_TTL, set to a default of 172800 seconds (2 days).

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -267,23 +275,142 @@ public String generateRegistrationToken(Principal principal, CloudRegistrationIn
...
+
+            if (anonConsumer == null) {
+                anonConsumer = new AnonymousCloudConsumer()
+                    .setCloudAccountId(cloudAccountId)
+                    .setCloudInstanceId(cloudInstanceId)
+                    .setProductId(productId)
+                    .setCloudProviderShortName(cloudProviderShortName);
+
+                anonConsumer = anonymousCloudConsumerCurator.create(anonConsumer);
+            }
```
nikosmoum, line 381: Something tricky here: In the case where the anonConsumer is not null here, it means the client already has an anon cloud consumer record, so it was already given an anonymous token beforehand. But the code goes on to generate a brand new anonymous token (with a renewed expiration date). This could potentially be abused for avoiding perpetually generating newer and newer anonymous tokens (and in extent, anonymous SCA certs). 
I wonder what should be done in this case.

One option would be to throw an exception at them that more or less says 'you already got an anonymous token, you aren't allowed another one'.

Another would be to send them a new anonymous token, but with a reduced expiration time, depending on the amount of time that has passed since the creation of the AnonCloudConsumer record (so if the original token is supposed to have 2 days of total duration until expiration, we would set the same expiration date on the new token: AnonCloudConsumer.created + 2 days). If 2 days have already passed, then we throw the exception. 

If they haven't properly registered within 2 days, something has gone really wrong, and they should open a customer case.

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -267,23 +275,142 @@ public String generateRegistrationToken(Principal principal, CloudRegistrationIn
...
+                .getByCloudInstanceId(cloudInstanceId);
+
+            if (anonConsumer == null) {
+                anonConsumer = new AnonymousCloudConsumer()
+                    .setCloudAccountId(cloudAccountId)
+                    .setCloudInstanceId(cloudInstanceId)
+                    .setProductId(productId)
+                    .setCloudProviderShortName(cloudProviderShortName);
+
+                anonConsumer = anonymousCloudConsumerCurator.create(anonConsumer);
```
nikosmoum, line 380: Something we need to think about here is where should the transactional boundary be. We aren't within a Resource method (all of them which have @transactional set on). This is the first time that we actually persist new data in an authentication method. 

What happens if we create the consumer here, but we fail later on (e.g. when we generate the token)? Probably we need a @Transactional on the `generateRegistrationTokenV2` method itself

nikosmoum (changes requested):
Found a few things to change. I may do another pass tomorrow, but these were the ones that stood out

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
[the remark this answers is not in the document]
JoshAlbrecht23: I think splitting up the token creation into two different methods makes sense. One for the regular JWT token and another for the anonymous JWT token. I also think it makes sense to put the different token types into an enum!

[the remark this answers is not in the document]
Ceiu: Yeah, but I feel like we should probably go one step further: a completely separate auth class (`AnonymousRegistrationAuth`?). There's a lot of overlap, but where the deviations occur make it very difficult to properly utilize all the same paths without getting into a case where we're constantly checking the token type, what "mode" we're in, or some other state validation and hand-off.

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
Ceiu: Given how the base design this is being built on works, it *should* be in the context of a resource method. The token generation is done as a specific request from a method which currently doesn't do any transaction management, but could easily enough.

The more I look at this and the other comment regarding the splitting up of classes, I wonder if this needs to be broken up even further into not just two auth classes as suggested in the other comment, but isolating the auth bits into *only* the token verification, and moving token generation to the respective resource or some other shared utility. Thoughts on this?

nikosmoum: I got confused for a moment because we're in an *Auth.java method, but you're right. We are in the context of the /cloud/authorize resource method, so that's where the transactional annotation should be set.

And me being confused by this, is proof that we should be splitting this class even further. So something like:
CloudRegistrationAuthV1: handles only the v1 autoreg verification
CloudRegistrationAuthV2: handles only the v2 autoreg verification
Some other class that deals with generating the v1 token, and is called by the Resource method
Some other class that deals with generating the v2 token, and is called by the Resource method

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
nikosmoum: A third option here, which I think would be the most elegant, is to have the jwt token stored at the time of the initial generation, and if a consumer for which we already have a record for tries to autoreg again, we just send back that original jwt token (which conveniently has the expiration date we want: 2 days after the original autoreg request).

The disadvantage is obviously that we would have to store the token, which takes extra cycles and DB storage for all autoreg requests. And I don't think this is a good trade, since this would only be useful for handling this edge case of someone doing stupid things using a custom script.

nikosmoum: I think I prefer the second option here. Any thoughts on it? @JoshAlbrecht23 @Ceiu

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
JoshAlbrecht23: I agree with having two cloud registration auth classes (CloudRegistrationAuth and AnonymousCloudRegistrationAuth) that only handles verifying the http request authorization handler and producing a Principal as that is their responsibility as a AuthProvider. Since this authorization is done differently between the two token types, it is nice to separate that logic into two different classes instead of having one class with bunch of conditionals based on token type (as Chris mentioned).

It is a little strange to be using an AuthProvider class for token generation and so I agree that moving the token generation logic something like a utility method or maybe to CloudRegistrationResource makes sense to me.

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
nikosmoum: Adding note here: We need feedback from PM before deciding the best way forward, but for now we leave the code as is (generate new token every time, with renewed expiration time). If that feedback arrives after this PR is merged, then it will be worked on in a separate task

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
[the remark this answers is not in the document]
JoshAlbrecht23: My bad on this, I misunderstood what you were suggesting on separating. I agree that it makes sense to seperate the auth for the anonymous token type into a new class!

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
JoshAlbrecht23: I added a commit showing the restructuring of the cloud registration auth classes and the token generation. I moved the `CloudRegistrationAdapter` calls and the processing of the result to the `CloudRegistrationResource` class. I want to check and make sure this aligns with the separation that you had in mind!

src/main/java/org/candlepin/auth/CloudAuthTokenType.java
```diff
@@ -0,0 +1,18 @@
+package org.candlepin.auth;
+
+// TODO: Java Docs, does this enum need to be generated by open api spec?
```
Ceiu, line 3: Nope. We probably also don't need the enum at all if we're separating out the two auth handlers. We likely won't need to be making the comparison enough to justify parsing it into a full object.

src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
```diff
@@ -78,4 +179,18 @@ public CloudAccountData setupCloudAccountOrg(String cloudAccountID, String cloud
         return new CloudAccountData("owner_key", false);
     }
 
+    private boolean isOwnerEntitledToProduct(String ownerKey, String productId) {
+        if (ownerKey == null || productId == null) {
+            return false;
+        }
+
+        Set<String> subIds = this.datastore.productSubscriptionMap.getOrDefault(productId, new HashSet<>());
```
Ceiu, line 187: micro-optimization: the new `HashSet<>()` can be changed to Set.of() for fewer characters, or `Collections.emptySet()` can be used to avoid the overhead of creating the new instance entirely.

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -267,23 +275,142 @@ public String generateRegistrationToken(Principal principal, CloudRegistrationIn
...
+            Owner owner = ownerCurator.getByKey(ownerKey);
+            if (owner != null) {
+                ownerAndPoolExists = !poolCurator.listByOwnerAndProduct(owner, productId).isEmpty();
+            }
+        }
+
+        String anonymousConsumerUuid = null;
+        if (ownerKey == null || ownerKey.isBlank() || !result.isEntitled() || !ownerAndPoolExists) {
+
+            // TODO: Schedule anonymous org creation job
```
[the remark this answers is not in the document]
JoshAlbrecht23: Unfortunately I don't see a way of verifying that the job was scheduled through the spec tests. Ideally we would be able to use the GET `/jobs` endpoint and specify the owner key that is unique to the spec test and the job key. However, looking at cp_async_jobs table we don't actually store the owner id for these jobs in the `owner_id` column.

So, without being able to filter the jobs down by owner, I don't think it's possible to reliably tell if a job was scheduled by that particular test because there will be concurrently running spec tests also creating jobs for the job key `CloudAccountOrgSetupJob`. However, I did add assertions for scheduling the job in `CloudRegistrationResourceTest.java`.

src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
```diff
@@ -66,4 +79,102 @@ public String resolveCloudRegistrationData(CloudRegistrationInfo cloudRegInfo)
...
+
+        String offerId = metadata.get(OFFERING_ID_FIELD_NAME);
+        if (offerId == null || offerId.isBlank()) {
+            throw new MalformedCloudRegistrationException(
+                "missing offer ID in registration information metadata");
+        }
+
+        String ownerKey = this.datastore.getOwnerKeyForCloudAccountId(accountId);
+        String productId = this.datastore.getProductIdForOfferId(offerId);
+        Boolean isEntitled = isOwnerEntitledToProduct(ownerKey, productId);
```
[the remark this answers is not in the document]
JoshAlbrecht23: Yeah, this should be a primitive, this was the remains of removing some logic in `isOwnerEntitledToProduct`

src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
```diff
@@ -66,4 +79,102 @@ public String resolveCloudRegistrationData(CloudRegistrationInfo cloudRegInfo)
...
+    }
+
+    private Boolean isOwnerEntitledToProduct(String ownerKey, String productId) {
+        if (ownerKey == null || productId == null) {
+            return false;
+        }
+
+        Set<String> subIds = this.datastore.productSubscriptionMap.getOrDefault(productId, new HashSet<>());
+
+        return subIds.stream().anyMatch(subId -> {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/hostedtest/HostedTestDataStore.java
```diff
@@ -82,23 +81,28 @@ public class HostedTestDataStore {
...
+    protected Map<String, String> cloudOfferIdToProductId;
+    protected Map<String, String> cloudAccountIdToOwnerKey;
+
     /**
      * Creates a new HostedTestDataStore instance
      */
     @Inject
     public HostedTestDataStore() {
-        this.ownerMap = new ConcurrentHashMap<>();
+        this.ownerMap = new ConcurrentHashMap<String, Owner>();
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated. I think Sonarqube flagged me on this one. I wonder if we can adjust some of the configurations

src/main/java/org/candlepin/hostedtest/HostedTestDataStore.java
JoshAlbrecht23: From my understanding the offering IDs are 1:1 with product IDs. I updated the name

src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
JoshAlbrecht23: Updated, and moved the enforcement of the sizing to the mutators

src/test/java/org/candlepin/model/AnonymousCloudConsumerCuratorTest.java
JoshAlbrecht23: Updated throughout

api/candlepin-api-spec.yaml
JoshAlbrecht23: It appears that this is possible! Updated

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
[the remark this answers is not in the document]
JoshAlbrecht23: Updated the logic for this

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
JoshAlbrecht23: Added

src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
JoshAlbrecht23: Updated

src/main/java/org/candlepin/hostedtest/HostedTestDataStore.java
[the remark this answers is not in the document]
Ceiu: I took a look at the latest run and it seems to be fine with this section now. No idea what it could have had a problem with before.

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -83,4 +156,149 @@ private CloudRegistrationData getCloudRegistrationData(CloudRegistrationDTO clou
...
+     * @return
+     *  true if the owner exists and has a pool for the product ID, or false if the owner
+     *  does not exist or does not have a pool for the product ID.
+     */
+    private boolean doesOwnerAndPoolExist(String ownerKey, String productId) {
+        boolean ownerAndPoolExists = false;
+        if (ownerKey != null && !ownerKey.isBlank()) {
+            Owner owner = ownerCurator.getByKey(ownerKey);
+            if (owner != null) {
+                ownerAndPoolExists = !poolCurator.listByOwnerAndProduct(owner, productId).isEmpty();
```
Ceiu, line 297: We had talked about this before as a double query optimization, but looking at the actual underlying query here, we should probably write something custom specifically for this lookup rather than doing two individual queries and all the builder pain that goes into that `listByOwnerAndProduct` call.

Specifically, I believe the product ID in this case is going to be a SKU-level product, so we can really optimize the query down to a couple of joins. Given that this will probably be getting a lot of usage, it'll be better than hitting one of the slowest queries in CP repeatedly. Something on the order of:

```
// rename me pls
PoolCurator.poolWithProductExistsInOrg(owner_key, product_id):

  String jpql = "SELECT pool.id FROM Pool pool JOIN pool.owner owner JOIN pool.product prod WHERE owner.key = :owner_key AND prod.id = :product_id";

  // remembering to set the result limit to 1 to short circuit out
  return !this.getEntityManager()
    .createQuery(jpql)
    .setParameter("owner_key", owner_key)
    .setParameter("product_id", product_id)
    .setMaxResults(1)
    .getResultList()
    .isEmpty();
```

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -83,4 +156,149 @@ private CloudRegistrationData getCloudRegistrationData(CloudRegistrationDTO clou
...
+
+            createdAnonConsumer = anonymousCloudConsumerCurator.create(createdAnonConsumer);
+            anonymousConsumerUuid = createdAnonConsumer.getUuid();
+
+            log.info("Anonymous consumer created for instance {} using cloud account {} {}",
+                authResult.getCloudInstanceId(), authResult.getCloudProvider().shortName(),
+                authResult.getCloudAccountId());
+        }
+
+        String token = anonymousConsumerUuid == null ?
```
Ceiu, line 218: This bit confuses me somewhat. This flow is for requesting a v2 anon auth token, but it's possible we kick back a v1 token which is only good for doing passwordless registration.

What's the client flow expected to be in this case? Depending on the output they will go into normal registration or request the anon content cert?

src/main/java/org/candlepin/service/CloudRegistrationAdapter.java
```diff
@@ -44,6 +45,24 @@ public interface CloudRegistrationAdapter {
...
+     * @param cloudRegInfo
+     *     A CloudRegistrationInfo instance which contains the cloud provider details to process
+     *
+     * @throws MalformedCloudRegistrationException
+     *     if the cloud registration info is null, incomplete, or invalid
+     *
+     * @throws CloudRegistrationAuthorizationException
+     *     if cloud registration is not permitted for the provider or account holder
+     *
+     * @return the cloud authentication result which contains the owner ID, cloud account ID, and
```
Ceiu, line 60: Something very critical to note here: we will be getting back an owner *key*, not an ID. The owner ID is internal to CP and nothing outside of us uses it. Other teams in the flow may call the value an owner ID, but anywhere we have an external reference to an owner, it's the key.

src/main/resources/db/changelog/changelog-update.xml
```diff
@@ -186,4 +186,5 @@
     <include file="db/changelog/20230220150433-drop-act-key-env-tables.xml"/>
     <include file="db/changelog/20230322103117-reduce-consumer-name.xml"/>
     <include file="db/changelog/20230725171517-add-anonymous-claimed_column.xml"/>
-</databaseChangeLog>
+    <include file="db/changelog/20230803000000-create-anonymous-consumers.xml" />
+</databaseChangeLog>
```
Ceiu, line 190: missing line break at EOF

src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
```diff
@@ -0,0 +1,271 @@
...
+        if (uuid.length() > UUID_MAX_LENGTH) {
+            throw new IllegalArgumentException("uuid exceeds the max length");
+        }
+
+        this.uuid = uuid;
+        return this;
+    }
+
+    @PrePersist
+    public void ensureUUID() {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Removed the mutators for UUID and the pre-persist hook, and moved the generation of the UUID to the constructor!

I can create a separate PR to make the same changes for Consumer. It will be much easier to review them in a separate PR!

src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /**
+     * @param uuid
+     *     the UUID to set
+     *
+     * @return a reference to this AnonymousCloudConsumer instance
+     */
+    public AnonymousCloudConsumer setUuid(String uuid) {
+        if (uuid == null) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Removed this mutator when moving the UUID generation to the constructor

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
JoshAlbrecht23: Added this new method `hasPoolForProduct` in the PoolCurator for this!

src/main/java/org/candlepin/service/CloudRegistrationAdapter.java
JoshAlbrecht23: Good catch! Updated to owner key

src/main/resources/db/changelog/changelog-update.xml
JoshAlbrecht23: Added

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
JoshAlbrecht23: So this isn't requesting a v2 anon auth token, it is only asking to authenticate the cloud instance metadata. So, we still have to handle the case already existing workflows where we don't need to generate an anonymous token because we don't need to do any additional processing on the backend (create orgs, subscriptions, etc). So, we send the standard token because the client can use it for the passwordless registration. 

> What's the client flow expected to be in this case? Depending on the output they will go into normal registration or request the anon content cert?

Correct! The client will now have two different workflows based on the token type. If the token type is standard then they can use the existing workflow and do a standard registration with the token. If the token type is anonymous then they will have to get the anonymous content certificate until they can fully register after all backend processing is completed.

src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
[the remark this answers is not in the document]
JoshAlbrecht23: Discussed offline, but we decided to keep the UUID mutators and the validation check in the pre-persist hook.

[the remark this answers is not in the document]
Ceiu: Our wires got crossed somewhere along the line. My intent was to:
- generate a UUID at construction time
- keep the setter with its validators (adding the isBlank check)
- remove the pre-persist hook for UUID generation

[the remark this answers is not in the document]
JoshAlbrecht23: That is my bad, I misunderstood your message! I have made the changes

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
Ceiu: > So this isn't requesting a v2 anon auth token, it is only asking to authenticate the cloud instance metadata. 

I don't follow. This particular end point is used to fetch a token, when is then used to either register directly (v1) or fetch an anon SCA cert while waiting to register (v2) (feel free to correct or add context there); providing the cloud instance metadata is just a means to that end.

What struck me as odd is that we opted to have the client specify which version of the logic they're invoking via the version parameter, but depending on the result of the upstream processing of it, would kick back a different token entirely. That looks to have been corrected in the latest update, so the point is probably moot now.

> The client will now have two different workflows based on the token type.

I would hope that the flow has already been determined by the client before the request is made. These feel different enough that the difference in token should occur at the behest of the client, rather than making the client react to our output. Again, this looks to have been addressed, so my initial confusion or concern may no longer be warranted.

Ceiu: ARRRGH, Githuuuuub! The link went to the `cloudRegister` method, not the `processAdapterAuthResult` method, and I based the entirety of my response on what I tricked myself into thinking was a refactor of the logic.

Looking at the *actual* line in question again, I'm still a bit confused as to the intended usage of both tokens here. Given that their scope and usage are completely different, is the client already aware of what token they expect to get back? Are we setting the client up for failure here if the client is expecting an anon token and we give them back the standard token?

I wasn't expecting this functionality to have overlap with the v1/standard tokens at all.

src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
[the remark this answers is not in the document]
Ceiu: Pinging this comment again given the confusion my bad communication skills caused surrounding the mutator and persist hook. Still should have that check.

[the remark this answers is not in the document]
JoshAlbrecht23: Added and I also added a isBlank check to `setId`

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -777,6 +774,36 @@ public boolean hasActiveEntitlementPools(String ownerId, Date date) {
...
+     *
+     * @param ownerKey
+     *  the key to the owner who owns the pool
+     *
+     * @param productId
+     *   the ID of the product to check for
+     *
+     * @return true if the owner has a pool for the product, or false if there is no pool
+     */
+    @Transactional
```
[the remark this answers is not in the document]
JoshAlbrecht23: removed

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -902,7 +929,7 @@ public List<Pool> getOversubscribedBySubscriptionIds(String ownerId, Map<String,
 
             Junction subscriptionJunction = Restrictions.and(subscriptionExpr).add(
                 Restrictions.or(Restrictions.isNull("sourceEntitlement"),
-                Restrictions.eqOrIsNull("sourceEntitlement", entry.getValue())));
+                    Restrictions.eqOrIsNull("sourceEntitlement", entry.getValue())));
```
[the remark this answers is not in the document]
JoshAlbrecht23: updated

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
JoshAlbrecht23: When the client makes a request to GET /cloud/authorize the client can't know what type of token to expect because that is dependent on the state of the server. However, the future requests after the call to GET /cloud/authorize is dependent on the token type that is returned. This is what I believe the differences in the workflows:

**Standard Token Workflow (Auto reg v1)**
1. Request to GET /cloud/authorize and receives a standard token with owner key
2. Request to POST /consumers to register system to the owner and get identity certificate
3. Request to GET /consumers/{consumer_uuid}/content_access to get content access certificate

**Anonymous Token Workflow**
1. Request to GET /cloud/authorize and receives an anonymous token with anonymous consumer uuid
2. Request to GET /consumers/{consumer_uuid}/content_access to get anonymous content access certificate

Ceiu (approved):

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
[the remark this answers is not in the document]
nikosmoum: As I mentioned in one of my other comments, when queueing the job, we should be also setting the owner key (if it is not null, when sent by the adapters). so at least in 1 use case (when we're running this job and the org exists, but isn't entitled upstream yet), we could check for the specific job being completed

src/main/java/org/candlepin/resource/util/CloudAuthTokenUtil.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.Objects;
+
+import javax.inject.Inject;
+
+
+
+/**
+ * A utility class for generating cloud authentication tokens
+ */
+public class CloudAuthTokenUtil {
```
nikosmoum, line 46: This class should probably be in the org.candlepin.auth package, and called `CloudAuthTokenGenerator`, since it's not a utility, but key to cloud authentication. I know it's called by a *Resource class, but the only reason for that being a Resource is technical limitations (that the instance metadata is too large to fit in HTTP headers, so we had to create a new API endpoint for it).

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
nikosmoum: The general separation looks good. For more detailed feedback I'll add new comments

src/main/java/org/candlepin/auth/AnonymousCloudRegistrationAuth.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Objects;
+
+import javax.inject.Inject;
+
+
+
+/**
+ * AuthenticationProvider that authenticates anonymous access tokens generated from the CloudRegistration endpoint
+ */
+public class AnonymousCloudRegistrationAuth implements AuthProvider {
```
nikosmoum, line 43: This whole class should have probably been written as part of CANDLEPIN-625, but since you've already started it it's fine. Just anote for that task: for any new AuthProvider to work, i needs to be registered in `AuthenticationFilter.setupAuthStrategies()`.

nikosmoum (changes requested):

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -83,4 +155,124 @@ private CloudRegistrationData getCloudRegistrationData(CloudRegistrationDTO clou
...
+
+        String ownerKey = authResult.getOwnerKey();
+
+        // verify that the owner exists upstream and is entitled
+        boolean createAnonConsumer = false;
+        if (ownerKey == null || ownerKey.isBlank() || !authResult.isEntitled()) {
+            CloudAccountOrgSetupJobConfig jobConfig = CloudAccountOrgSetupJob.createJobConfig()
+                .setCloudAccountId(authResult.getCloudAccountId())
+                .setCloudOfferingId(authResult.getOfferId())
+                .setCloudProvider(authResult.getCloudProvider());
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -83,4 +155,124 @@ private CloudRegistrationData getCloudRegistrationData(CloudRegistrationDTO clou
...
+                authResult.getCloudProvider().shortName(), authResult.getCloudAccountId(),
+                authResult.getOfferId());
+        }
+
+        // Impl note: It is possible the owner and subscription exists through the adapters, but
+        // the owner and pool do not yet exist in Candlepin. If that is the case we will want to create
+        // an anonymous cloud consumer record and provide an anonymous registration token.
+        String anonymousConsumerUuid = existingAnonConsumer == null ? null : existingAnonConsumer.getUuid();
+        boolean ownerAndPoolExists = poolCurator.hasPoolForProduct(ownerKey, authResult.getProductId());
+        if ((!ownerAndPoolExists && existingAnonConsumer == null) || createAnonConsumer) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated. 

There was a problem with that algorithm not covering the case where a VM is authenticating again and has already created an anonymous consumer record and the pools and owner are not yet synced to Candlepin. This results in an anonymous consumer not needing to be created (`createAnonConsumer=false`) and we need to return an anonymous token, but since the token type is dictated by the `createAnonConsumer`, we return a standard token type.

I moved the token and token type to be based on `anonymousConsumerUuid` again.

This test case is covered by the spec test:
`CloudRegistrationSpecTest.shouldReceiveSameAnonConsumerUuidWhenReAuthenticatingForV2Auth`

src/main/java/org/candlepin/resource/util/CloudAuthTokenUtil.java
JoshAlbrecht23: Updated

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
```diff
@@ -14,85 +14,469 @@
...
+        assertTokenType(adminClient.MAPPER, result.getToken(), ANON_TOKEN_TYPE);
+        assertThat(result)
+            .isNotNull()
+            .returns(null, CloudAuthenticationResultDTO::getOwnerKey)
+            .doesNotReturn(null, CloudAuthenticationResultDTO::getAnonymousConsumerUuid)
+            .returns(ANON_TOKEN_TYPE, CloudAuthenticationResultDTO::getTokenType);
+    }
+
+    @Test
+    public void shouldAllowV2AuthWithExistingEntitledAnonymousOwnerForCloudAccountId() throws Exception {
```
nikosmoum, line 199: This test seems identical with the first one here `shouldAllowV2AuthWithExistingEntitledOwnerForCloudAccountId`, however they do have different names. Was it by accident, or is one supposed to test with an anonymous owner and the other with a normal owner?

JoshAlbrecht23: On second thought it probably makes more sense to merge CANDLEPIN-646 first and then update the test

JoshAlbrecht23: Got CANDLEPIN-646 merged in and rebased this branch with the changes and updated the `shouldAllowV2AuthWithExistingEntitledAnonymousOwnerForCloudAccountId` test!

nikosmoum (approved):

PR #4162, opened 2023-08-01
Drop ruby/rvm/bundle leftovers

ptoscano:
ruby was mostly dropped few months ago, and few related bits of it are still there; since they are not actually of any use, drop them:
- drop the custom `gai.conf` for glibc in containers: it was used only to workaround network issues when connecting to rubygems.org
- stop installing `ruby-devel` in containers
- stop installing rvm and ruby dependencies in containers
- drop the `|| true` workaround in `cp-test`: apparently needed only for the rvm.sh shell profile script
- drop `Gemfile` & `Gemfile.lock`, no more needed since bundle is no more called from anywhere (hopefully)
- drop bits & comments in other places


bakajstep:
We can not remove ruby staff from main, we are building image for jenkins for 4.1 branch from main. After this [task](https://issues.redhat.com/browse/CANDLEPIN-637) will be done, we should drop jenkins support and we can delete them.

PR #4181, opened 2023-08-11
2231503: CertificateCleanupJob execution takes a very long time; CANDLEPIN-643

wottop:
- When there was a lot of old data in the Content Access Certificate table, it to hours for the cleanup to process
- Indexes added to tables that have a serial id column that were not indexed like other similar tables
- Index added to consumer table for content access cert id column


Ceiu (changes requested):
We should scope these index additions to PostgreSQL. MariaDB/MySQL automatically adds them (under InnoDB), so it'd be redundant on that side of things.

Ceiu (approved):

PR #4183, opened 2023-08-14
CANDLEPIN-431: Reorganize creation of Object Mappers

wottop:
Also some linter fixes

### DIFF
--- a/src/main/java/org/candlepin/guice/CandlepinModule.java
+++ b/src/main/java/org/candlepin/guice/CandlepinModule.java
@@ -165,23 +165,12 @@ import org.candlepin.util.DateSource;
 import org.candlepin.util.DateSourceImpl;
 import org.candlepin.util.ExpiryDateFunction;
 import org.candlepin.util.FactValidator;
+import org.candlepin.util.ObjectMapperFactory;
 import org.candlepin.util.Util;
 import org.candlepin.util.X509ExtensionUtil;
 import org.candlepin.validation.CandlepinMessageInterpolator;
 
-import com.fasterxml.jackson.annotation.JsonAutoDetect;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.PropertyAccessor;
-import com.fasterxml.jackson.databind.AnnotationIntrospector;
-import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.introspect.AnnotationIntrospectorPair;
-import com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector;
-import com.fasterxml.jackson.databind.ser.impl.SimpleFilterProvider;
-import com.fasterxml.jackson.datatype.hibernate5.Hibernate5Module;
-import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
-import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
-import com.fasterxml.jackson.module.jaxb.JaxbAnnotationIntrospector;
 import com.google.common.base.Function;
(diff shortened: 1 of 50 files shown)


Januson (changes requested):
Changes look great. It will be much easier to manage mappers moving forward.

Januson (approved):

src/main/java/org/candlepin/sync/SyncUtils.java
```diff
@@ -63,34 +55,9 @@ File makeTempDir(String baseName) throws IOException {
...
-
-        if (config != null) {
-            this.mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES,
-                config.getBoolean(ConfigProperties.FAIL_ON_UNKNOWN_IMPORT_PROPERTIES));
-        }
-    }
-
-    public ObjectMapper getObjectMapper() {
-        return this.mapper;
+        this.mapper = objectMapper;
```
Ceiu, line 61: The removal of the `getObjectMapper` method invalidates the need for the mapper at all in this class.

src/main/java/org/candlepin/sync/Exporter.java
```diff
@@ -118,6 +119,7 @@ public Exporter(ConsumerTypeCurator consumerTypeCurator, OwnerCurator ownerCurat
         CdnCurator cdnCurator,
         CdnExporter cdnExporter,
         SyncUtils syncUtils,
+        @Named("SyncUtilsObjectMapper") ObjectMapper mapper,
```
Ceiu, line 122: This should be named something more specific, since the mapper isn't associated with the sync utility, but with the sync (import/export) operation itself. SyncObjectMapper, ImportObjectMapper, ManifestObjectMapper, etc..

src/main/java/org/candlepin/policy/js/RulesObjectMapper.java
```diff
@@ -62,29 +56,7 @@ public class RulesObjectMapper {
...
-        // Very important for deployments so new rules files can return additional
-        // properties that this current server doesn't know how to serialize, but still
-        // shouldn't fail on.
-        this.mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
-
-        AnnotationIntrospector primary = new JacksonAnnotationIntrospector();
-        AnnotationIntrospector secondary = new JaxbAnnotationIntrospector(mapper.getTypeFactory());
-        AnnotationIntrospector pair = new AnnotationIntrospectorPair(primary, secondary);
-        this.mapper.setAnnotationIntrospector(pair);
+        this.mapper = ObjectMapperFactory.getObjectMapper();
```
Ceiu, line 59: If we're going to tie the object mapper to the DI framework in general, we should be doing that here as well.

src/main/java/org/candlepin/guice/CandlepinModule.java
```diff
@@ -481,52 +471,38 @@ protected void configureModelTranslator() {
...
-        return mapper;
+        return ObjectMapperFactory.getHypervisorUpdateJobObjectMapper();
+    }
+
+    @Provides
+    @Singleton
+    @Named("SyncUtilsObjectMapper")
+    private ObjectMapper configureSyncUtilsObjectMapper() {
+        return ObjectMapperFactory.getSyncUtilsObjectMapper(this.config);
     }
```
Ceiu, line 507: We probably want to bind the ObjectMapperFactory as an unnamed provider to handle the cases where we don't need any special, named configuration (i.e. RuleObjectMapper). Then we can eliminate any direct static invocation of the factory method(s).

Ceiu (changes requested):
A number of the mappers changed from plain object mapper instances to our full "standard" mapper with all the extra filters and translators. Are we sure the output is going to be close enough to not cause any problems in those instances?

PR #4190, opened 2023-08-25
[M] Collecting logs from spec tests runners in GH action

bakajstep:
During backporting GH action I need to get logs from runners with spec test. So I write this to get logs and upload them as artifacts. I think it could be useful also in main, if not just close it. Or if you want more logs to collect just write me. Now its collecting docker log, candlepin logs and tomcat logs. It collects them only when spec test fail, so we save time on successful run.

### DIFF
--- a/.github/workflows/pr_verification.yml
+++ b/.github/workflows/pr_verification.yml
@@ -395,6 +395,38 @@ jobs:
         uses: gradle/gradle-build-action@v2
         with:
           arguments: spec -Dspec.test.client.host=candlepin
+          
+      - name: Collect docker logs on failure
+        if: failure()
+        uses: jwalton/gh-docker-logs@4599896f4f5414b2444c1512096809da191c9e81
+        with:
+          dest: './logs-${{ matrix.database }}-${{ matrix.mode }}'
+
+      - name: Collect candlepin and Tomcat logs on failure
+        if: failure()
+        shell: bash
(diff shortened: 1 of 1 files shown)


bakajstep:
[Here](https://github.com/candlepin/candlepin/actions/runs/5976654932?pr=4190) when you scroll on artifacts you can downoald the logs.

bakajstep:
Added support for getting logs from database container

.github/workflows/pr_verification.yml
```diff
@@ -395,6 +395,38 @@ jobs:
         uses: gradle/gradle-build-action@v2
         with:
           arguments: spec -Dspec.test.client.host=candlepin
+          
+      - name: Collect docker logs on failure
+        if: failure()
+        uses: jwalton/gh-docker-logs@v2
```
JoshAlbrecht23, line 401: @nikosmoum what is the policy for incorporating unverified actions/dependencies like this? From what I am seeing, this looks to be the only action that will pull the logs from docker containers, but I'm not sure what kind of risk we are taking here by using this action.

bakajstep: Done, with this [solution](https://github.com/candlepin/candlepin/pull/4197#discussion_r1321662594)

JoshAlbrecht23 (approved):

PR #4197, opened 2023-08-29
[4.2] CANDLEPIN-637: Backport GH Actions

bakajstep:


.github/containers/entrypoint.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Function to write variables to the configuration file
+write_config() {
+    mkdir -p /etc/candlepin
+    touch "$CONF_FILE"
+    chmod 600 "$CONF_FILE"
+    echo "# Configuration generated from env variables" > "$CONF_FILE"
+    for key in "${!conf_candlepin[@]}"; do
+        value="${conf_candlepin[$key]}"
+        key_value=$(env | awk -F '=' -v k="$key" '{ if ($1 == k) print $2 }')
```
JoshAlbrecht23, line 45: Maybe `env_value` might be a better name instead of `key_value`. It's a little confusing there is also the `key` and `value` variables and I think indicating that the value come from env might be helpful

JoshAlbrecht23 (changes requested):
Just some minor changes, but for the most part it looks good!

One thing we could consider for the candlepin.conf file is dynamically building the candlepin.conf file in pr_verification.yml `spec_tests` task and then mounting the candlepin.conf file to the /etc/candlepin directory when running docker compose. That may be more straight forward then parsing the env variables and building the candlepin.conf file in the entrypoint script.

.github/workflows/pr_verification.yml
```diff
@@ -0,0 +1,364 @@
...
+          distribution: ${{ env.JAVA_DISTRIBUTION }}
+          java-version: ${{ env.JAVA_VERSION }}
+
+      - name: Run validate translation
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: validate_translation
+
+  build_image:
+    name: Build Candlepin image
```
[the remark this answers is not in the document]
bakajstep: done

.github/containers/entrypoint.sh
```diff
@@ -0,0 +1,66 @@
...
+write_config
+
+#Run liquibase
+export LIQUIBASE_HOME="/opt/tomcat/webapps/candlepin/WEB-INF"
+liquibase=$(find /opt/tomcat/webapps/candlepin/WEB-INF/lib/ -maxdepth 1 -type f -name "liquibase-core*" -printf "%f\n")
+
+java -jar "$LIQUIBASE_HOME/lib/$liquibase" \
+  --classpath=$LIQUIBASE_HOME/classes \
+  --changeLogFile=db/changelog/changelog-create.xml \
+  --url="$JPA_CONFIG_HIBERNATE_CONNECTION_URL" --username="$JPA_CONFIG_HIBERNATE_CONNECTION_USERNAME" \
```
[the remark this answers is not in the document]
bakajstep: done

bakajstep:
@JoshAlbrecht23 I also made these changes in 4.1 branch

bakajstep:
> Just some minor changes, but for the most part it looks good!
> 
> One thing we could consider for the candlepin.conf file is dynamically building the candlepin.conf file in pr_verification.yml `spec_tests` task and then mounting the candlepin.conf file to the /etc/candlepin directory when running docker compose. That may be more straight forward then parsing the env variables and building the candlepin.conf file in the entrypoint script.

I will do that change in separate commit, so we can decide which solution we want.

.github/workflows/pr_verification.yml
```diff
@@ -0,0 +1,355 @@
...
+          java-version: ${{ env.JAVA_VERSION }}
+
+      - name: Run Spec tests
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: spec -Dspec.test.client.host=candlepin
+          
+      - name: Collect docker logs on failure
+        if: failure()
+        uses: jwalton/gh-docker-logs@v2
```
JoshAlbrecht23, line 322: Just a note on this, we have another comment on the use of this GH action and we are investigating on how to proceed on this.

https://github.com/candlepin/candlepin/pull/4190/files#r1310527992

JoshAlbrecht23: I took a look at jwalton/gh-docker-logs and it looks good to me to use. We should pin the action to the current SHA to prevent additional security risk!

JoshAlbrecht23 (approved):

PR #4201, opened 2023-08-31
[M] CANDLEPIN-619: Re-run PR workflow when target branch gets pushed

bakajstep:
- Create a workflow that triggers when a PR is closed, and it will rerun all workflows associated with the affected branch
- Added feature when you mark PR with label `skip CI` it will not rerun affected workflows
- **We need to start again using merge commits, otherwise it doesn't work**

### DIFF
--- a/.github/workflows/rerun_workflow_on_push.yml
+++ b/.github/workflows/rerun_workflow_on_push.yml
@@ -1,0 +1,84 @@
+---
+name: Rerun affected workflows in long lived branches on push
+
+on:
+  push:
+    branches:
+      - main
+      - candlepin-*-HOTFIX
+
+jobs:
+  rerun-workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Mask secrets
+        shell: bash
+        run: |
+          echo "::add-mask::${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Get PR info
+        uses: 8BitJonny/gh-get-current-pr@fa2160116055f6fe10376375892b255f8b2ed8f6
+        id: PR
+        with:
+          github-token: ${{ github.token }}
+
+      - name: Rerun workflows
+        if: ${{ !contains(steps.PR.outputs.pr_labels, 'skip CI') }}
+        shell: bash
+        run: |
+          base_branch="${GITHUB_REF#refs/heads/}"
+          
+          # List affected PRs exclude dependabots, weblate and draft PRs
+          pr_head_shas=$(gh api \
+          -H "Accept: application/vnd.github+json" \
+          /repos/candlepin/candlepin/pulls | \
+          jq --arg base "$base_branch" -r '.[] | select(.base.ref == $base and .draft == false and .user.login != "weblate" and .user.login != "dependabot[bot]") | .head.sha')
+
+          # List runner ids from PRs(only for pr_verification.yml)
+          for head_sha in $pr_head_shas; do
+            runner_id=$(gh api \
+            -H "Accept: application/vnd.github+json" \
+            /repos/candlepin/candlepin/actions/workflows/pr_verification.yml/runs?head_sha=$head_sha | \
+            jq -r '.workflow_runs[] | .id')
+            runner_ids+=($runner_id)
+          done
+          
+          # Cancel running workflows
+          for id in "${runner_ids[@]}"; do
+            # Get status of running workflow
+            status=$(gh api \
+            -H "Accept: application/vnd.github+json" \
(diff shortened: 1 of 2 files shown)


.github/workflows/rerun_workflow_on_push.yml
```diff
@@ -0,0 +1,79 @@
...
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          /repos/candlepin/candlepin/pulls | \
+          jq --arg base "$base_branch" -r '.[] | select(.base.ref == $base and .draft == false and .user.login != "weblate" and .user.login != "dependabot[bot]") | .head.sha')
+
+          # List runner ids from PRs(only for pr_verification.yml)
+          for head_sha in $pr_head_shas; do
+            runner_id=$(gh api \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            /repos/candlepin/candlepin/actions/workflows/pr_verification.yml/runs?head_sha=$head_sha | \
```
nikosmoum, line 43: This will need to be updated after the pr_verification.yml workflow gets split into many files.

.github/workflows/rerun_workflow_on_push.yml
```diff
@@ -0,0 +1,79 @@
...
+              gh api \
+              --method POST \
+              -H "Accept: application/vnd.github+json" \
+              -H "X-GitHub-Api-Version: 2022-11-28" \
+              /repos/candlepin/candlepin/actions/runs/$id/cancel
+            fi
+          done
+          
+          # We need to sleep here because canceling workflows take some time
+          sleep 15
```
nikosmoum, line 68: Sleeping usually leads to intermittent failures. what happens if one of the workflows takes longer than 15 to get cancelled? Perhaps a do-while loop over the runner_ids where you check that all of them have been cancelled is better here?

nikosmoum: Or another approach would be to keep the sleep here, but in the next step (rerun workflows), you use some kind of try/catch, around the API call, so that if one call fails, the loop continues to call all the rest of them, instead of the whole workflow failing?

nikosmoum (changes requested):
Looks good in total, just have some suggestions

.github/workflows/rerun_workflow_on_push.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Rerun workflows
+        if: ${{ !contains(steps.PR.outputs.pr_labels, 'skip CI') }}
+        shell: bash
+        run: |
+          base_branch="${GITHUB_REF#refs/heads/}"
+          
+          # List affected PRs exclude dependabots, weblate and draft PRs
+          pr_head_shas=$(gh api \
+          -H "Accept: application/vnd.github+json" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
```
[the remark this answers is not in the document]
bakajstep: Not really. I locked the API version if something change so it doesn't break our action. But I can remove it.

.github/workflows/rerun_workflow_on_push.yml
bakajstep: yes

.github/workflows/rerun_workflow_on_push.yml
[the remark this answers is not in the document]
bakajstep: done

.github/workflows/rerun_workflow_on_push.yml
bakajstep: done

.github/workflows/rerun_workflow_on_push.yml
```diff
@@ -0,0 +1,84 @@
...
+          
+          # We need to wait here because canceling workflows take some time
+          for id in "${cancelled_ids[@]}"; do
+            while : ; do
+              sleep 2 # We sleep here so we don't overwhelm the GH API
+              status=$(gh api \
+              -H "Accept: application/vnd.github+json" \
+              /repos/candlepin/candlepin/actions/runs/$id | \
+              jq -r '.status')
+              [[ $status = "completed" ]] && break
```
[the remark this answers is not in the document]
bakajstep: Yes, i was suprised too and get block for overhelming GH API with requests. In UI they have `Cancelled` but from endpoint you will get status field set as `completed`. [Here](https://github.com/bakajstep/candlepin/actions/runs/6172549865/job/16753033189) you can see output from testing

PR #4220, opened 2023-09-06
[F] CANDLEPIN-625: Added anonymous bearer token authentication

JoshAlbrecht23:


JoshAlbrecht23:
Moving to draft to address SonarQube comments

src/main/java/org/candlepin/resteasy/filter/VerifyAuthorizationFilter.java
```diff
@@ -251,40 +254,39 @@ protected List<Persisted> getAccessedEntities(Verify verify, Object requestValue
...
-                // doesn't seem to exist in the DB. Error will be thrown in
-                // invoke though.
-                String typeName = Util.getClassName(verifyType);
-                if (typeName.equals("Owner")) {
-                    typeName = i18nProvider.get().tr("Organization");
+        Class<? extends Persisted>[] verifyTypes = verify.value();
+
+        for (Class<? extends Persisted> verifyType : verifyTypes) {
+            if (requestValue instanceof String) {
+                String verifyParam = (String) requestValue;
```
Januson, line 261: You can simplify this with a pattern variable to avoid manual casting.

```java
if (requestValue instanceof String verifyParam) {
```

src/main/java/org/candlepin/resteasy/filter/VerifyAuthorizationFilter.java
```diff
@@ -251,40 +254,39 @@ protected List<Persisted> getAccessedEntities(Verify verify, Object requestValue
...
+
+                entities.add(entity);
             }
+            else if (requestValue instanceof Collection) {
+                Collection<String> verifyParams = (Collection<String>) requestValue;
 
-            entities.add(entity);
+                // If the request is for a list of items, we'll leave it
+                // up to the requester to determine if something is missing or not.
+                if (verifyParams != null && !verifyParams.isEmpty()) {
```
Januson, line 280: You can just check for emptiness, `verifyParams` cannot be null.

src/main/java/org/candlepin/auth/AnonymousCloudConsumerPrincipal.java
```diff
@@ -0,0 +1,82 @@
...
+        return this.consumer.getUuid();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (!(obj instanceof AnonymousCloudConsumerPrincipal)) {
```
Januson, line 65: Another pattern variable possibility. It should also check for null.

src/main/java/org/candlepin/auth/AnonymousCloudConsumerPrincipal.java
```diff
@@ -0,0 +1,82 @@
...
+
+        final AnonymousCloudConsumerPrincipal other = (AnonymousCloudConsumerPrincipal) obj;
+
+        return this.consumer.getUuid().equals(other.consumer.getUuid());
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 3;
+        hash = 83 * hash + (this.consumer != null ? this.consumer.hashCode() : 0);
```
Januson, line 77: You can use Objects.hash to avoid calculating hash manually. Also please note that there is an inconsistency between equals and hashCode. HashCode uses the whole consumer while equals uses just it's uuid.

Januson (changes requested):
Changes look good. I have just a few nitpicks.

src/main/java/org/candlepin/resteasy/filter/VerifyAuthorizationFilter.java
JoshAlbrecht23: Good idea, that is much better. Updated

src/main/java/org/candlepin/resteasy/filter/VerifyAuthorizationFilter.java
JoshAlbrecht23: Updated

src/main/java/org/candlepin/auth/AnonymousCloudConsumerPrincipal.java
JoshAlbrecht23: Updated

src/main/java/org/candlepin/auth/AnonymousCloudConsumerPrincipal.java
JoshAlbrecht23: Good catch! Updated

nikosmoum (approved):

PR #4231, opened 2023-09-11
[M] CANDLEPIN-652: Added documentation for using the new container images

JoshAlbrecht23:


Ceiu:
What's here looks pretty good. But we should also add a section on importing test data, since I'm pretty sure we don't do that out of the box. It's just a matter of invoking the script, but especially since we're trying to move away from that tool, we should have something for prepping our users for hitting the API. Maybe that's a separate document entirely for using the container for testing and adding test data?

Another thing that this highlighted for me is that we should include a basic test compose file that will pull a postgresql container and the dev container and just *go*. Again, uncertain if that should be part of this PR (probably not), but we should definitely provide that.

JoshAlbrecht23:
> But we should also add a section on importing test data, since I'm pretty sure we don't do that out of the box. It's just a matter of invoking the script, but especially since we're trying to move away from that tool, we should have something for prepping our users for hitting the API. Maybe that's a separate document entirely for using the container for testing and adding test data?

I agree, having documentation on using the dev container for testing is a good idea,  and I also agree that it should be in a different location than the CONTRIBUTING.md. Maybe that should live in candlepinproject.org and we have a link to the page in the CONTRIBUTING.md?

> Another thing that this highlighted for me is that we should include a basic test compose file that will pull a postgresql container and the dev container and just go. Again, uncertain if that should be part of this PR (probably not), but we should definitely provide that at some point.

I completely agree!! It is nice to have a compose file that just sets up your dev environment for you and is especially nice for people viewing the project to easily get Candlepin running with little effort. I logged a jira CANDLEPIN-665 for this work.

containers/released-containers.md
```diff
@@ -0,0 +1,61 @@
...
+```bash
+curl -k -X POST https://localhost:8443/candlepin/owners/test-owner/products \
+    -u admin:admin \
+    -H 'Content-Type: application/json' \
+    -d '{"id":"test-prod", "name":"test product", "productContent":[{"content":{"id":"5001", "type":"yum", "name":"test content", "contentUrl":"/test/path", "arches":"x86_64"}}], "href":"/products/123456"}'
+```
+
+#### Using test_data_importer.py
+
+The `test_data_importer.py` script located in the `/bin/deployment` directory reads in predefined data from a json file and makes HTTP requests to the Candlepin API to generate the data.
```
Ceiu, line 55: Very happy to see it getting called out that the test data importer is just calling the API endpoints and isn't doing anything magical. :+1:

Ceiu (approved):

PR #4257, opened 2023-09-22
[M] Disallow creating consumers in org without permission

nikosmoum:
- This issue allowed a user that has no 'create consumer' permission in a certain org, to create consumers in it, if they could guess the organization key.


nikosmoum:
This is cherry-picked from https://github.com/candlepin/candlepin/pull/4244

Ceiu (approved):

PR #4262, opened 2023-09-26
[F] CANDLEPIN-657: Record authentication method on consumer registration

wottop:


nikosmoum:
This PR brings in 5 commits from main. Instead only the main commit should be here "CANDLEPIN-657: Record authentication method on consumer registration", and rebased against the feature/autoreg_v2 branch

PR #4263, opened 2023-09-27
[F] Migrate autoregv2 commits from main branch to feature branch

nikosmoum:


JoshAlbrecht23:
Just have to fix some checkstyle issues. I think some of these commits were merged in when we were using Spotless

PR #4268, opened 2023-09-28
[F] CANDLEPIN-675: Update anon cloud consumers to store multiple SKUs

JoshAlbrecht23:
- updated cloud offering hosted test endpoint to use multiple products
- PoolCurator.hasPoolForProduct updated to check pools for multiple products

### DIFF
--- a/src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
+++ b/src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
@@ -18,7 +18,11 @@ import org.candlepin.util.Util;
 
 import org.hibernate.annotations.GenericGenerator;
 
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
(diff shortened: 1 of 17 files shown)


nikosmoum:
This stopped compiling after I merged the other PR

src/main/java/org/candlepin/model/AnonymousCloudConsumer.java
```diff
@@ -50,8 +54,8 @@ public class AnonymousCloudConsumer extends AbstractHibernateObject<AnonymousClo
     public static final int CLOUD_INSTANCE_ID_MAX_LENGTH = 85;
     /** Max length for a value in the cloud provider short name field */
     public static final int CLOUD_PROVIDER_SHORT_NAME_MAX_LENGTH = 15;
-    /** Max length for a value in the product ID field */
-    public static final int PRODUCT_ID_MAX_LENGTH = 32;
+    /** Max length for a value in the product IDs field */
+    public static final int PRODUCT_ID_MAX_LENGTH = 200;
```
nikosmoum, line 58: So this doesn't seem to be used any more. Especially with the db column having changed to TEXT, it can be removed right? Alternatively, we can keep the column as VARCHAR with a large enough value.

JoshAlbrecht23: Good catch! Yeah, this should be removed. I missed this when I switched the column from VARCHAR to TEXT. Removed

nikosmoum (approved):

PR #4271, opened 2023-10-02
[F] CANDLEPIN-657: Record authentication method on consumer registration

wottop:


src/main/java/org/candlepin/auth/UserPrincipal.java
```diff
@@ -65,10 +66,37 @@ public UserPrincipal(String username, Collection<? extends Permission> permissio
         addPermission(new AsyncJobStatusPermission(getData(), this.getOwnerIds()));
     }
 
+    public UserPrincipal(String username, Collection<? extends Permission> permissions, boolean admin,
+                         OwnerInfo primaryOwner, AuthenticationMethod authMethod) {
+        this.username = username;
+        this.admin = admin;
+        this.primaryOwner = primaryOwner;
+        this.authMethod = authMethod == null ? AuthenticationMethod.BASIC : authMethod;
```
nikosmoum, line 74: I think if authMethod is null, it is better to leave it as null, since setting BASIC as default would be misleading.

src/main/resources/db/changelog/20230921144200-add_reg_auth_method_to_cp_consumers.xml
```diff
@@ -0,0 +1,20 @@
...
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <changeSet id="20230021144200-1" author="wpoteat">
+        <comment>
+            Add registration authorization method to the cp_consumer table.
+        </comment>
+
+        <addColumn tableName="cp_consumer">
+            <column name="reg_auth_method" type="VARCHAR(255)"/>
```
nikosmoum, line 15: This column should probably be nullable (all the existing rows will need to have null as the default value).

nikosmoum: Also, we can probably save some space/performance by reducing the varchar from 255 chars to 30 or so. Its an enum, and I doubt we will ever add a new auth type with a name longer than that. If we do, we can always abbreviate it.

src/main/java/org/candlepin/auth/UserPrincipal.java
```diff
@@ -65,10 +66,37 @@ public UserPrincipal(String username, Collection<? extends Permission> permissio
...
         return username;
     }
 
+    public UserPrincipal setAuthenticationMethod(AuthenticationMethod method) {
+        this.authMethod = method;
+        return this;
+    }
+
+    public AuthenticationMethod getAuthenticationMethod() {
+        return authMethod ==  null ? AuthenticationMethod.BASIC : authMethod;
```
nikosmoum, line 97: same here: let this be nullable

src/main/java/org/candlepin/dto/manifest/v1/ConsumerDTO.java
```diff
@@ -33,6 +33,7 @@ public class ConsumerDTO extends CandlepinDTO<ConsumerDTO> {
     protected ConsumerTypeDTO type;
     protected String urlWeb;
     protected String urlApi;
+    protected String regAuthMethod;
```
nikosmoum, line 36: I really dont think we need to be exporting this field to the manifest. This is only valueable for statistics/reporting of auth methods used in Hosted.

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -319,6 +319,9 @@ public static final class Facts {
     @Fetch(FetchMode.SELECT)
     private Map<String, String> environmentIds;
 
+    @Column(name = "reg_auth_method")
+    private String registrationAuthenticationMethod;
```
nikosmoum, line 323: You an use Hibernate's `@Enumerated(EnumType.STRING)` annotation here, and make the field type be the enum class iself

src/main/java/org/candlepin/auth/AuthenticationMethod.java
```diff
@@ -0,0 +1,37 @@
...
+ * in this software or its documentation.
+ */
+
+package org.candlepin.auth;
+
+public enum AuthenticationMethod {
+    BASIC("User"),
+    ACTIVATION_KEY("ActivationKey"),
+    ANONYMOUS_CLOUD("AnonymousCloud"),
+    CLOUD("Cloud"),
```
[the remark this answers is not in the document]
wottop: Still unchanged until we have this conversation.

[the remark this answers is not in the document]
wottop: Updated

PR #4275, opened 2023-10-03
Gradle Wrapper auto-update workflow

Januson:
- Add a workflow that periodically checks for Gradle Wrapper updates and opens a PR if update is available.
- Switch to wrapper distribution to bin for faster setup. It can be switched to full locally.


Januson:
Here's an example [PR](https://github.com/Januson/candlepin/pull/58) the action opens.

nikosmoum (approved):

PR #4283, opened 2023-10-04
[F] CANDLEPIN-626: Added anonymous content access certificate

JoshAlbrecht23:
- added support for fetching anonymous content access certificates

### DIFF
--- a/src/main/java/org/candlepin/controller/ContentAccessManager.java
+++ b/src/main/java/org/candlepin/controller/ContentAccessManager.java
@@ -17,10 +17,15 @@ package org.candlepin.controller;
 import org.candlepin.audit.EventSink;
 import org.candlepin.config.ConfigProperties;
 import org.candlepin.config.Configuration;
+import org.candlepin.controller.util.AnonymousContentPrefix;
 import org.candlepin.controller.util.ContentPrefix;
 import org.candlepin.controller.util.PromotedContent;
 import org.candlepin.controller.util.ScaContainerContentPrefix;
 import org.candlepin.controller.util.ScaContentPrefix;
+import org.candlepin.model.AnonymousCloudConsumer;
+import org.candlepin.model.AnonymousCloudConsumerCurator;
+import org.candlepin.model.AnonymousContentAccessCertificate;
+import org.candlepin.model.AnonymousContentAccessCertificateCurator;
 import org.candlepin.model.CertificateSerial;
 import org.candlepin.model.CertificateSerialCurator;
 import org.candlepin.model.Consumer;
@@ -43,6 +48,11 @@ import org.candlepin.model.dto.Content;
 import org.candlepin.pki.PKIUtility;
 import org.candlepin.pki.X509ByteExtensionWrapper;
 import org.candlepin.pki.X509ExtensionWrapper;
+import org.candlepin.service.ProductServiceAdapter;
+import org.candlepin.service.model.ContentInfo;
+import org.candlepin.service.model.ProductContentInfo;
+import org.candlepin.service.model.ProductInfo;
+import org.candlepin.util.Arch;
 import org.candlepin.util.OIDUtil;
 import org.candlepin.util.Util;
 import org.candlepin.util.X509V3ExtensionUtil;
@@ -60,11 +70,14 @@ import java.math.BigInteger;
 import java.security.GeneralSecurityException;
 import java.security.KeyPair;
 import java.security.cert.X509Certificate;
+import java.time.OffsetDateTime;
 import java.util.ArrayList;
-import java.util.Calendar;
+import java.util.Collection;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
@@ -192,6 +205,8 @@ public class ContentAccessManager {
             ContentAccessMode.ORG_ENVIRONMENT.toDatabaseValue());
     }
 
+    private static final String BASIC_ENTITLEMENT_TYPE = "basic";
+    private static final String SCA_ENTITLEMENT_TYPE = "OrgLevel";
 
     private final Configuration config;
     private final PKIUtility pki;
@@ -205,6 +220,9 @@ public class ContentAccessManager {
     private final EnvironmentCurator environmentCurator;
     private final ContentAccessCertificateCurator contentAccessCertCurator;
     private final EventSink eventSink;
+    private final AnonymousCloudConsumerCurator anonCloudConsumerCurator;
+    private final AnonymousContentAccessCertificateCurator anonContentAccessCertCurator;
+    private final ProductServiceAdapter prodAdapter;
 
     private final boolean standalone;
 
@@ -221,7 +239,10 @@ public class ContentAccessManager {
         ConsumerTypeCurator consumerTypeCurator,
         EnvironmentCurator environmentCurator,
         ContentAccessCertificateCurator contentAccessCertCurator,
-        EventSink eventSink) {
+        EventSink eventSink,
+        AnonymousCloudConsumerCurator anonCloudConsumerCurator,
+        AnonymousContentAccessCertificateCurator anonContentAccessCertCurator,
+        ProductServiceAdapter prodAdapter) {
 
         this.config = Objects.requireNonNull(config);
         this.pki = Objects.requireNonNull(pki);
@@ -235,6 +256,9 @@ public class ContentAccessManager {
         this.environmentCurator = Objects.requireNonNull(environmentCurator);
         this.contentAccessCertCurator = Objects.requireNonNull(contentAccessCertCurator);
         this.eventSink = Objects.requireNonNull(eventSink);
+        this.anonCloudConsumerCurator = Objects.requireNonNull(anonCloudConsumerCurator);
+        this.anonContentAccessCertCurator = Objects.requireNonNull(anonContentAccessCertCurator);
+        this.prodAdapter = Objects.requireNonNull(prodAdapter);
         this.standalone = this.config.getBoolean(ConfigProperties.STANDALONE);
     }
 
@@ -279,19 +303,32 @@ public class ContentAccessManager {
     private ContentAccessCertificate createNewScaCertificate(Consumer consumer, Owner owner)
         throws IOException, GeneralSecurityException {
         log.info("Generating new SCA certificate for consumer: \"{}\"", consumer.getUuid());
-        Validity oneYearValidity = Validity.oneYear();
-        CertificateSerial serial = createSerial(oneYearValidity);
+        OffsetDateTime start = OffsetDateTime.now().minusHours(1L);
+        OffsetDateTime end = start.plusYears(1L);
+
+        CertificateSerial serial = createSerial(end);
 
         KeyPair keyPair = this.pki.getConsumerKeyPair(consumer);
         byte[] pemEncodedKeyPair = this.pki.getPemEncoded(keyPair.getPrivate());
+        org.candlepin.model.dto.Product container = createSCAProdContainer(owner, consumer);
+
+        List<Environment> environments = this.environmentCurator.getConsumerEnvironments(consumer);
+        ContentPrefix contentPrefix = ScaContentPrefix.from(owner, this.standalone, environments);
+        PromotedContent promotedContent = new PromotedContent(contentPrefix).withAll(environments);
+
+        Map<org.candlepin.model.Content, Boolean> ownerContent = this.ownerContentCurator
+            .getActiveContentByOwner(owner.getId());
+
+        byte[] payloadBytes = createContentAccessDataPayload(consumer, ownerContent, promotedContent);
 
         ContentAccessCertificate existing = new ContentAccessCertificate();
         existing.setSerial(serial);
         existing.setKeyAsBytes(pemEncodedKeyPair);
         existing.setConsumer(consumer);
 
-        existing.setCert(createX509Cert(consumer, owner, serial, keyPair, oneYearValidity));
-        existing.setContent(this.createPayloadAndSignature(owner, consumer));
+        existing.setCert(createX509Cert(consumer.getUuid(), owner, serial, keyPair, container,
+            SCA_ENTITLEMENT_TYPE, start, end));
+        existing.setContent(this.createPayloadAndSignature(payloadBytes));
         ContentAccessCertificate savedCert = this.contentAccessCertificateCurator.create(existing);
         consumer.setContentAccessCert(savedCert);
         this.consumerCurator.merge(consumer);
@@ -305,23 +342,54 @@ public class ContentAccessManager {
         boolean isX509CertExpired = expiration.before(now);
 
         if (isX509CertExpired) {
-            Validity oneYearValidity = Validity.oneYear();
+            OffsetDateTime start = OffsetDateTime.now().minusHours(1L);
+            OffsetDateTime end = start.plusYears(1L);
+
             KeyPair keyPair = this.pki.getConsumerKeyPair(consumer);
             this.serialCurator.revokeById(existing.getSerial().getId());
-            CertificateSerial serial = createSerial(oneYearValidity);
+            CertificateSerial serial = createSerial(end);
+            org.candlepin.model.dto.Product container = createSCAProdContainer(owner, consumer);
+
             existing.setSerial(serial);
-            existing.setCert(createX509Cert(consumer, owner, serial, keyPair, oneYearValidity));
+            existing.setCert(createX509Cert(consumer.getUuid(), owner, serial, keyPair, container,
+                SCA_ENTITLEMENT_TYPE, start, end));
             this.contentAccessCertificateCurator.saveOrUpdate(existing);
         }
 
         Date contentUpdate = owner.getLastContentUpdate();
         boolean shouldUpdateContent = !contentUpdate.before(existing.getUpdated());
         if (shouldUpdateContent || isX509CertExpired) {
-            existing.setContent(this.createPayloadAndSignature(owner, consumer));
+            List<Environment> environments = this.environmentCurator.getConsumerEnvironments(consumer);
+            ContentPrefix contentPrefix = ScaContentPrefix.from(owner, this.standalone, environments);
+            PromotedContent promotedContent = new PromotedContent(contentPrefix).withAll(environments);
+
+            Map<org.candlepin.model.Content, Boolean> ownerContent = this.ownerContentCurator
+                .getActiveContentByOwner(owner.getId());
+
+            byte[] payloadBytes = createContentAccessDataPayload(consumer, ownerContent, promotedContent);
+            existing.setContent(this.createPayloadAndSignature(payloadBytes));
             this.contentAccessCertificateCurator.saveOrUpdate(existing);
         }
 
         return existing;
+    }
+
+    private org.candlepin.model.dto.Product createSCAProdContainer(Owner owner, Consumer consumer) {
+        org.candlepin.model.dto.Product container = new org.candlepin.model.dto.Product();
+        List<org.candlepin.model.dto.Content> dtoContents = new ArrayList<>();
+        List<Environment> environments = this.environmentCurator.getConsumerEnvironments(consumer);
+        ContentPrefix prefix = ScaContainerContentPrefix.from(owner, this.standalone, environments);
+        for (Environment environment : environments) {
+            dtoContents.add(getContent(prefix, environment.getId()));
+        }
+
+        if (dtoContents.isEmpty()) {
+            dtoContents.add(getContent(prefix, null));
+        }
+
+        container.setContent(dtoContents);
+
+        return container;
     }
 
     private ContentAccessCertificate wrap(ContentAccessCertificate cert) {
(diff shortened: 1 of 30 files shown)


src/main/java/org/candlepin/service/ProductServiceAdapter.java
```diff
@@ -50,6 +51,17 @@ public interface ProductServiceAdapter {
...
+    /**
+     * Retrieves the products matching the provided top level SKU IDs. Only the found products will be
+     * returned. When no matching products are found, an empty list will be returned.
+     *
+     * @param skuIds
+     *  the top level SKU IDs to fetch products for
+     *
+     * @return the products that match the provided SKU IDs
+     */
+    List<ProductInfo> getProductsBySkuIds(Collection<String> skuIds);
```
Ceiu, line 63: We may want to call this something more generic like "get children by product IDs" or some such. That we're passing in a "SKU" and getting engineering products back out is kind of an implementation detail based on how products are sold today -- it's not necessarily true going forward. And in the case of some future world where n-tier products must be supported, the term "SKU" loses meaning entirely.

src/main/java/org/candlepin/sync/Exporter.java
```diff
@@ -176,14 +178,70 @@ public File getFullExport(Consumer consumer, String cdnLabel, String webUrl,
...
+     * @param serials
+     *  the collection of serials used to filter the {@link AnonymousContentAccessCertificate} by
+     *
+     * @throws ExportCreationException
+     *  if unable to create the compressed entitlement exports
+     *
+     * @return
+     *  the compressed entitlement export
+     */
+    public File createEntitlementExport(AnonymousCloudConsumer consumer, Collection<Long> serials)
```
Ceiu, line 205: Are we ever going to be exporting an anonymous consumer? Is this something we *want* to be doing? What does the import logic look like for such an entitlement?

src/main/resources/db/changelog/202310040000000-create-anonymous-certificates.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="202310040000000-1" author="jalbrech">
```
Ceiu, line 9: Few things here:

Let's add checks here to make sure the table doesn't already exist before adding it. We can also stack the index creation into the table creation, since the index can't exist if the table didn't exist yet.

Also, if we're going to be adding a new cert table (pls no), that's another thing that we need to clean up in the cert cleanup job.

Finally, I noticed this is modifying the schema of stuff that's net-new to the feature as a whole. Since this is being built in a feature branch we can safely manipulate the existing changeset when it makes sense to do so, rather than updating the schema in a separate changeset as we do for stuff that hits a shippable branch. That way we can avoid the pain inherit to having several changesets that have to be merged and ordered together later when the feature branch is merged in.

src/main/resources/db/changelog/changelog-update.xml
```diff
@@ -1,192 +1,194 @@
...
-    <include file="db/changelog/20221117125815-prime_system_locks.xml"/>
-    <include file="db/changelog/20221003160605-remove_direct_ak_product_reference.xml"/>
-    <include file="db/changelog/20221003160623-remove_direct_env_content_reference.xml"/>
-    <include file="db/changelog/20221212140215-remove_prodcont_fk_delete_cascade.xml"/>
-    <include file="db/changelog/20230207161633-clear_entity_versions-1.xml"/>
-    <include file="db/changelog/20230220150433-drop-act-key-env-tables.xml"/>
-    <include file="db/changelog/20230322103117-reduce-consumer-name.xml"/>
-    <include file="db/changelog/20230811183400-add-indexes-certificate-cleanup.xml"/>
-    <include file="db/changelog/20230725171517-add-anonymous-claimed_column.xml"/>
+    <property name="project.name" value="candlepin" />
```
Ceiu, line 3: Oh boy... what linter caught this file?

Ceiu (changes requested):

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -467,8 +499,21 @@ else if (type.isType(ConsumerTypeEnum.HYPERVISOR)) {
         return entitlementVersion != null && entitlementVersion.startsWith("3.");
     }
 
-    private String createDN(Consumer consumer, Owner owner) {
-        return "CN=" + consumer.getUuid() + ", O=" + Rdn.escapeValue(owner.getKey());
+    private String createDN(String commonName, String organizationName) {
+        StringBuilder builder = new StringBuilder();
+        if (commonName != null && !commonName.isEmpty()) {
+            builder.append("CN=" + commonName);
```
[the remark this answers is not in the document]
Ceiu: So I dug into this a bit, and it looks like the concatenation is indeed being done before the append call. Using the test class:

```
public class TestClass {

    public static void main(String[] args) {
        StringBuilder builder = new StringBuilder("test - ");

        if (args.length > 1) {
            builder.append("arg count: " + args.length);
        }
        else {
            builder.append("no args :(");
        }

        System.out.println(builder.toString());
    }

}
```

When we compile that down and then look at the resultant bytecode, the important bits on line 7 look as follows:

```
      18: arraylength
      19: invokedynamic #14,  0             // InvokeDynamic #0:makeConcatWithConstants:(I)Ljava/lang/String;
      24: invokevirtual #18                 // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
```

The compiler is using some [very magical dynamically created function](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/invoke/StringConcatFactory.html#makeConcatWithConstants(java.lang.invoke.MethodHandles.Lookup,java.lang.String,java.lang.invoke.MethodType,java.lang.String,java.lang.Object...)) to generate an optimized string concatenation function for handling that specific concatenation.

When we turn this into a chain of appends, however, we get this:
```
      17: ldc           #14                 // String arg count:
      19: invokevirtual #16                 // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      22: aload_0
      23: arraylength
      24: invokevirtual #20                 // Method java/lang/StringBuilder.append:(I)Ljava/lang/StringBuilder;
```

So, it looks like the compiler *won't* reuse an existing/active StringBuilder to handle concatenation within the arguments going into the string builder.

As to whether or not that's faster or slower... I don't know. Last I looked into this, the compiler was straight up replacing concatenation operations with inline StringBuilders itself, so the comparison was a bit more clear cut. I have no idea what this new generated concat factory method is doing under the hood, but based on the fact that it's still string concatenation at the end of the day, I have to assume that it's not going to be any faster than just appending to our target string builder. Another test for another day, I suppose.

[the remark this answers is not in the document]
Januson: It used to be compiled to StringBuilder append chains in Java 8. Since Java 9 compiler handles concatenation with invokeDynamic call to [StringConcatFactory](https://docs.oracle.com/javase/9/docs/api/java/lang/invoke/StringConcatFactory.html).

src/main/java/org/candlepin/sync/Exporter.java
JoshAlbrecht23: Probably not! Should we just return a 400 HTTP response for when someone makes a request to GET `/consumers/{consumer_uuid}/certificates` with the header accepting application/zip?

src/main/resources/db/changelog/changelog-update.xml
JoshAlbrecht23: I think this was a VS Code formatting change haha I usually turn off the formatter, but it looks like it was on at some point

src/main/java/org/candlepin/service/ProductServiceAdapter.java
JoshAlbrecht23: Yeah, that makes sense! Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -804,4 +850,234 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
...
+     */
+    public AnonymousContentAccessCertificate getCertificate(AnonymousCloudConsumer consumer) {
+        if (consumer == null) {
+            throw new IllegalArgumentException("anonymous cloud consumer is null");
+        }
+
+        String rollBackMsg = "Rolling back anonymous content access cert (re)generation transaction";
+
+        try {
+            AnonymousContentAccessCertificate cert = this.anonCloudConsumerCurator
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated!

src/main/java/org/candlepin/controller/ContentAccessManager.java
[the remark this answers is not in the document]
JoshAlbrecht23: This was interesting! Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -360,6 +379,18 @@ public static Validity oneYear() {
             return new Validity(start, end);
         }
 
+        /**
+         * Creates a validity duration starting one hour in the past and ends after two days.
+         */
+        public static Validity twoDays() {
```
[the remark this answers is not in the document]
JoshAlbrecht23: > It may even reduce the need for explicit abstractions like this.

I agree! I removed the Validity object entirely and used start and end OffsetDateTime objects instead

src/main/resources/db/changelog/202310040000000-create-anonymous-certificates.xml
JoshAlbrecht23: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -467,8 +499,21 @@ else if (type.isType(ConsumerTypeEnum.HYPERVISOR)) {
...
+        if (commonName != null && !commonName.isEmpty()) {
+            builder.append("CN=" + commonName);
+        }
+
+        if (organizationName != null && !organizationName.isBlank()) {
+            if (builder.length() != 0) {
+                builder.append(", ");
+            }
+
+            builder.append("O=" + Rdn.escapeValue(organizationName));
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -279,19 +301,24 @@ public ContentAccessCertificate getCertificate(Consumer consumer) {
     private ContentAccessCertificate createNewScaCertificate(Consumer consumer, Owner owner)
         throws IOException, GeneralSecurityException {
         log.info("Generating new SCA certificate for consumer: \"{}\"", consumer.getUuid());
-        Validity oneYearValidity = Validity.oneYear();
-        CertificateSerial serial = createSerial(oneYearValidity);
+        OffsetDateTime start = OffsetDateTime.now().minus(1, ChronoUnit.HOURS);
```
Januson, line 304: java.time classes have a set of convenience methods for time updates such as `minusHours` or `plusYears`.

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -804,4 +820,209 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
...
+
+        CertificateSerial serial = createSerial(end);
+        KeyPair keyPair = this.pki.generateKeyPair();
+        byte[] pemEncodedKeyPair = this.pki.getPemEncoded(keyPair.getPrivate());
+
+        List<ContentInfo> contentInfo = getContentInfo(products);
+        org.candlepin.model.dto.Product container = new org.candlepin.model.dto.Product();
+        List<Content> dtoContents = new ArrayList<>();
+        dtoContents.addAll((convertContentInfoToContentDto(contentInfo)));
+        container.setContent(dtoContents);
```
Januson, line 896: You could create dtoContents list in a single step via constructor, but why do you need a new one when `convertContentInfoToContentDto` already return a new list?

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -467,14 +466,29 @@ else if (type.isType(ConsumerTypeEnum.HYPERVISOR)) {
...
-        return "CN=" + consumer.getUuid() + ", O=" + Rdn.escapeValue(owner.getKey());
+    private String createDN(String commonName, String organizationName) {
+        StringBuilder builder = new StringBuilder();
+        if (commonName != null && !commonName.isEmpty()) {
+            builder.append("CN=")
+                .append(commonName);
+        }
+
+        if (organizationName != null && !organizationName.isBlank()) {
+            if (builder.length() != 0) {
```
Januson, line 477: Could be `!builder.isEmpty()`.

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -804,4 +820,209 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
...
+            wrappedCert.setKey(cert.getKey());
+            wrappedCert.setSerial(cert.getSerial());
+
+            return wrappedCert;
+        }
+        catch (Exception e) {
+            String msg = "Unexpected exception occurred while fetching certificate for anonymous" +
+                "cloud consumer: {}";
+            log.error(msg, consumer, e);
+        }
```
Januson, line 869: We should limit the scope of try blocks to places that actually throw exceptions. Wrapping whole method body can get confusing as to why is the scope so wide.

Also something I noticed. The flow currently does `if null -> log generic error -> return null` on each step in the flow. Wouldn't exception make more sense? The code is not recovering from this anyway.

It would also allow resources to return errors to indicate what went wrong.

src/main/java/org/candlepin/controller/ContentAccessManager.java
JoshAlbrecht23: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
JoshAlbrecht23: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
JoshAlbrecht23: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -804,4 +850,234 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
...
+     *  the content to convert
+     *
+     * @return the converted {@link org.candlepin.model.Content} objects
+     */
+    private List<org.candlepin.model.Content> convertContentInfoToContent(
+        Collection<ContentInfo> contentInfo) {
+        List<org.candlepin.model.Content> convertedContent = new ArrayList<>();
+        contentInfo.stream()
+            .filter(Objects::nonNull)
+            .forEach(content -> {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -804,4 +850,234 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
...
+     * @param contentInfo
+     *  the content to convert
+     *
+     * @return the converted {@link ContentInfo} objects
+     */
+    private List<Content> convertContentInfoToContentDto(Collection<ContentInfo> contentInfo) {
+        List<Content> convertedContent = new ArrayList<>();
+        contentInfo.stream()
+            .filter(Objects::nonNull)
+            .forEach(content -> {
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -804,4 +850,234 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
...
+        List<ContentInfo> contentInfo = new ArrayList<>();
+        prodInfo.stream()
+            .filter(Objects::nonNull)
+            .forEach(prod -> {
+                prod.getProductContent().stream()
+                    .filter(Objects::nonNull)
+                    .forEach(prodContent -> contentInfo.add(prodContent.getContent()));
+            });
+
+        return contentInfo;
```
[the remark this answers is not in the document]
JoshAlbrecht23: Yeah, this is much better! Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
JoshAlbrecht23: This is a good point. We really should be using Exceptions instead of logging and then returning null. Updated to throw exceptions instead of returning null

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -804,4 +820,192 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
...
+
+    private AnonymousContentAccessCertificate createAnonContentAccessCertificate(
+        AnonymousCloudConsumer consumer) throws Exception {
+        log.info("Generating content access certificate for anonymous cloud consumer: {}",
+            consumer.getUuid());
+
+        List<ProductInfo> products = prodAdapter.getChildrenByProductIds(consumer.getProductIds());
+        if (products == null || products.isEmpty()) {
+            String msg = "Unable to retrieve products for anonymous cloud consumer: " + consumer.getUuid();
+            throw new Exception(msg);
```
Januson, line 874: Please, never do this. Throwing `Exception` directly breaks any usefulness the Java's exceptions have.

- It forces all callers to declare `throws Exception` which hides all information about what the method actually throws.
- If the the exception needs to be handled it forces the caller into pokemon exception handling. Causing it to potentially catch unwanted exceptions (such as NPE).

Also it should probably be unchecked exception. It could be checked exception, but it won't be handled. Just propagated to the endpoint and translated to http error, so it would only pollute method definitions along the way. Either way it should be our exception to signify certificate generation failed.

JoshAlbrecht23: Yeah, those are great points! Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -804,4 +820,195 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
...
+        Set<String> entitledProductIds = Set.of("content-access");
+
+        PromotedContent promotedContent = new PromotedContent(new AnonymousContentPrefix());
+        org.candlepin.model.dto.Product productModel = v3extensionUtil
+            .mapProduct(engProduct, skuProduct, promotedContent, null, emptyPool, entitledProductIds);
+        byte[] payloadBytes = v3extensionUtil
+            .createEntitlementDataPayload(List.of(productModel), consumer.getUuid(), emptyPool, null);
+
+        return createPayloadAndSignature(payloadBytes);
+    }
```
JoshAlbrecht23, line 934: @nikosmoum @Ceiu Do we even need to populate the content for the `AnonymousContentAccessCertificate` if the content paths will be read from the x509 certificate via the 1.3.6.1.4.1.2312.9.7 object ID?

Ceiu: I'd need to double check to be certain, but I'm pretty sure the client uses the entitlement payload to build the .repo files, and the CDN is what checks the content paths. So we'll still need both IIRC.

JoshAlbrecht23: Ah yeah! I believe you are correct!

src/main/java/org/candlepin/sync/Exporter.java
nikosmoum: We won't be exporting an anon consumer. If they hit this endpoint with `application/zip` mime type, we should be returning a 400

src/main/resources/db/changelog/changelog-update.xml
```diff
@@ -189,4 +189,5 @@
     <include file="db/changelog/20230725171517-add-anonymous-claimed_column.xml"/>
     <include file="db/changelog/20230803000000-create-anonymous-consumers.xml" />
     <include file="db/changelog/20230921144200-add_reg_auth_method_to_cp_consumers.xml"/>
-</databaseChangeLog>
+    <include file="db/changelog/202310040000000-create-anonymous-certificates.xml" />
+</databaseChangeLog>
```
Ceiu, line 193: new line went missing at EOF

Ceiu (approved):

src/main/resources/db/changelog/changelog-update.xml
JoshAlbrecht23: updated

nikosmoum (changes requested):

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2067,21 +2069,43 @@ public void deleteConsumer(@Verify(Consumer.class) String uuid) {
...
      */
-    public List<CertificateDTO> getEntitlementCertificates(@Verify(Consumer.class) String consumerUuid,
-        String serials) {
-        log.debug("Getting client certificates for consumer: {}", consumerUuid);
+    public List<CertificateDTO> getEntitlementCertificates
+    (@Verify({AnonymousCloudConsumer.class, Consumer.class}) String consumerUuid, String serials) {
+        Principal principal = ResteasyContext.getContextData(Principal.class);
+        if (principal instanceof AnonymousCloudConsumerPrincipal) {
+            log.debug("Getting client certificates for anonymous consumer: {}", consumerUuid);
+            AnonymousCloudConsumerPrincipal p = (AnonymousCloudConsumerPrincipal) principal;
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2102,6 +2126,43 @@ public List<CertificateDTO> getEntitlementCertificates(@Verify(Consumer.class) S
...
+     * @return a {@link AnonymousContentAccessCertificate}, or null if a certificate was not able to be
+     *  retrieved, created, or filtered by the serial IDs
+     */
+    private CertificateDTO getCertForAnonCloudConsumer(AnonymousCloudConsumer consumer, String serials) {
+        Certificate caCert;
+        try {
+            caCert = this.contentAccessManager.getCertificate(consumer);
+        }
+        catch (Exception e) {
+            String msg = "Unable to retrieve or create anonymous content access certificate for consumer";
```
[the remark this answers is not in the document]
JoshAlbrecht23: Updated

src/main/java/org/candlepin/model/AnonymousContentAccessCertificateCurator.java
```diff
@@ -0,0 +1,51 @@
...
+     *  the anonymous cloud consumer to retrieve an {@link AnonymousContentAccessCertificate} for
+     *
+     * @return an anonymous content access certificate or null if unable to retrieve the certificate
+     */
+    public AnonymousContentAccessCertificate getByAnonymousCloudConsumer(AnonymousCloudConsumer consumer) {
+        if (consumer == null) {
+            return null;
+        }
+
+        Criteria criteria = this.createSecureCriteria()
```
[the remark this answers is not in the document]
JoshAlbrecht23: This method wasn't being used after the code to export the anonymous certificates was removed. So, I removed the method!

src/main/resources/db/changelog/202310040000000-create-anonymous-certificates.xml
```diff
@@ -0,0 +1,46 @@
...
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="cp_anonymous_certificates"/>
+            </not>
+        </preConditions>
+
+        <createTable tableName="cp_anonymous_certificates">
```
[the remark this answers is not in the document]
JoshAlbrecht23: Created a Jira to update the CertificateCleanupJob: https://issues.redhat.com/browse/CANDLEPIN-697

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -221,7 +239,10 @@ public ContentAccessManager(
         ConsumerTypeCurator consumerTypeCurator,
         EnvironmentCurator environmentCurator,
         ContentAccessCertificateCurator contentAccessCertCurator,
-        EventSink eventSink) {
+        EventSink eventSink,
+        AnonymousCloudConsumerCurator anonCloudConsumerCurator,
+        AnonymousContentAccessCertificateCurator anonContentAccessCertCurator,
+        ProductServiceAdapter prodAdapter) {
```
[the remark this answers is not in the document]
JoshAlbrecht23: This makes sense. I wasn't aware that certificate generation changes were getting migrated from Graviton to Candlepin. Refactoring the anonymous certificate code to match the new pattern once the pattern is introduced into Candlepin makes sense!

src/main/java/org/candlepin/resource/ConsumerResource.java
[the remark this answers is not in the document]
nikosmoum: The way this was translated (`i18n.tr(msg)`) won't work. This method (`tr`) needs to get the raw string itself  `i18n.tr("Unable to..")` instead of a variable, because that's what the gettext cli looks for when extracting translatable strings.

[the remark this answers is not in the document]
JoshAlbrecht23: Oh interesting. Updated

nikosmoum (approved):

PR #4289, opened 2023-10-06
[M] 2242299: Fix date-time parsing for If-Modified-Since header

Januson:
- Fixed locale settings on date-parser for If-Modified-Since header


nikosmoum (approved):
just needs unit tests

PR #4290, opened 2023-10-06
[4.2] 2242455: Fix date-time parsing for If-Modified-Since header

Januson:
- Fixed locale settings on date-parser for If-Modified-Since header


nikosmoum (approved):
just needs unit tests

PR #4297, opened 2023-10-11
[Entity Layering] CANDLEPIN-670: Update entitlement certificate generation to include product namespace roduct namespace

Januson:


src/main/java/org/candlepin/service/impl/DefaultEntitlementCertServiceAdapter.java
```diff
@@ -332,11 +332,15 @@ public Set<X509ExtensionWrapper> prepareV1Extensions(Set<Product> products, Pool
...
 
-    public Set<X509ExtensionWrapper> prepareV3Extensions() {
+    public Set<X509ExtensionWrapper> prepareV3Extensions(Pool pool) {
         Set<X509ExtensionWrapper> result = v3extensionUtil.getExtensions();
-        X509ExtensionWrapper typeExtension = new X509ExtensionWrapper(OIDUtil.REDHAT_OID + "." +
-            OIDUtil.TOPLEVEL_NAMESPACES.get(OIDUtil.ENTITLEMENT_TYPE_KEY), false, "Basic");
+        String entTypeOID = OIDUtil.getOid(OIDUtil.Namespace.ENTITLEMENT_TYPE);
+        String productOID = OIDUtil.getOid(OIDUtil.Namespace.PRODUCT);
+        X509ExtensionWrapper typeExtension = new X509ExtensionWrapper(entTypeOID, false, "Basic");
+        X509ExtensionWrapper productExtension = new X509ExtensionWrapper(
```
Ceiu, line 340: Rather than just referencing the product's ID, this is supposed to be the namespace of the product. Currently this is not available as the work to add it has not yet been finished.

src/main/java/org/candlepin/util/OIDUtil.java
```diff
@@ -28,15 +29,20 @@ public final class OIDUtil {
...
-    public static final String ORDER_NAMESPACE_KEY = "Order";
-    public static final String CHANNEL_FAMILY_NAMESPACE_KEY = "Channel Family";
-    public static final String ROLE_ENT_NAMESPACE_KEY = "Role Entitlement";
-    public static final String PRODUCT_CERT_NAMESPACE_KEY = "Product";
-    public static final String ENTITLEMENT_VERSION_KEY = "Entitlement Version";
-    public static final String ENTITLEMENT_DATA_KEY = "Entitlement Data";
-    public static final String ENTITLEMENT_TYPE_KEY = "Entitlement Type";
+    private static final String REDHAT_OID = "1.3.6.1.4.1.2312.9";
+
+    public enum Namespace {
```
Ceiu, line 34: This is interesting. On the surface I like this, because it gets rid of the magic strings and makes the code far safer everywhere down the line.

However, once we have an enum, we're at the point where we may not even *need* the map to do the OID conversion. It could be a case where we have a generic enum interface for like `OIDComponent` which defines a `toOidString()` method that all of our enums could implement. Then `OIDUtil` can get paired down to a single method along the lines of `build(OIDComponent...)` which would end up being basically just a join of all the components in the order specified.

Even better is if it built to an OID itself and that could be passed to our extensions. But this is getting into the crypto layer refactor which may or may not even be necessary in the near term.

Ceiu (changes requested):
Some good changes in here, but unfortunately this task is largely blocked until the schema changes and core logic are in place so that the entitlement cert generation has something to populate the new extension field.

We could work on just the OID refactoring as an isolated task, but we'd not want to close out the task in Jira until the new extension is being populated.

There's also something I want to explore next week in that instead of an extension, it may make sense to add it as part of the DN of the cert instead. We'll want to talk with Henning and the other folks that are more familiar with this to see what is the best fit in terms in that regard.

src/main/java/org/candlepin/util/OIDUtil.java
```diff
@@ -99,14 +105,15 @@ private OIDUtil() {
...
-        TOPLEVEL_NAMESPACES.put(ENTITLEMENT_TYPE_KEY, "8");
+        TOPLEVEL_NAMESPACES.put(Namespace.PRODUCT_CERTIFICATE, "1");
+        TOPLEVEL_NAMESPACES.put(Namespace.CHANNEL_FAMILY, "2");
+        TOPLEVEL_NAMESPACES.put(Namespace.ROLE_ENTITLEMENT, "3");
+        TOPLEVEL_NAMESPACES.put(Namespace.ORDER, "4");
+        TOPLEVEL_NAMESPACES.put(Namespace.SYSTEM, "5");
+        TOPLEVEL_NAMESPACES.put(Namespace.ENTITLEMENT_VERSION, "6");
+        TOPLEVEL_NAMESPACES.put(Namespace.ENTITLEMENT_DATA, "7");
+        TOPLEVEL_NAMESPACES.put(Namespace.ENTITLEMENT_TYPE, "8");
+        TOPLEVEL_NAMESPACES.put(Namespace.PRODUCT, "9");
```
Ceiu, line 116: This should probably be named something closer to "entitlement namespace" or "product namespace" instead of "product"

src/main/java/org/candlepin/util/OIDUtil.java
Januson: I agree. There's quite a lot we could do to improve OID mapping. Separate card is a good idea. I didn't want to push this refactoring too far as it would drown the actual change that's pretty small.

src/main/java/org/candlepin/util/OIDUtil.java
Januson: Done.

Januson:
I am closing this until it gets unblocked.

PR #4325, opened 2023-10-24
CANDLEPIN-665: Create compose file for running candlepin:dev-latest container image

wottop:


wottop:
This will need to wait until the dev image is updated to include the hostedtest directory.

dev-container/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
+version: '3.8'
+
+services:
+  postgres:
+    image: postgres:latest
```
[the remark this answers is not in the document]
nikosmoum: That image needs logging in / acceptance of T&Cs, so lets not do that after all.

dev-container/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+      JPA_CONFIG_HIBERNATE_CONNECTION_USERNAME: candlepin
+      JPA_CONFIG_HIBERNATE_CONNECTION_PASSWORD: candlepin
+      CANDLEPIN_AUTH_CLOUD_ENABLE: "false"
+      CANDLEPIN_AUTH_TRUSTED_ENABLE: "true"
+      CANDLEPIN_AUTH_OAUTH_ENABLE: "true"
+      CANDLEPIN_AUTH_OAUTH_CONSUMER_RSPEC_SECRET: rspec-oauth-secret
+      CANDLEPIN_DB_DATABASE_MANAGE_ON_STARTUP: Manage
+      CANDLEPIN_REFRESH_ORPHAN_ENTITY_GRACE_PERIOD: 0
+      CANDLEPIN_STANDALONE: "false"
+      MODULE_CONFIG_HOSTED_CONFIGURATION_MODULE: org.candlepin.testext.hostedtest.HostedTestModule
```
joshmalbrecht, line 36: When we build the development image we are also populating the candlepin.conf file with [default values](https://github.com/candlepin/candlepin/blob/main/containers/release.Containerfile#L98-L109) and including it in the image. So, we could remove the duplicates that are already defined in the candlepin.conf file and only define the environment variables that differ or are not include in the default configurations.

Environment variables to include in the docker compose file:
- CANDLEPIN_AUTH_CLOUD_ENABLE
- CANDLEPIN_AUTH_TRUSTED_ENABLE
- CANDLEPIN_STANDALONE
- MODULE_CONFIG_HOSTED_CONFIGURATION_MODULE

joshmalbrecht (changes requested):
Noticed that we could remove some of the environment variables that are already defined in the candlepin.conf file in the development image.

dev-container/docker-compose.yml
wottop: This one is still needed:
 JPA_CONFIG_HIBERNATE_CONNECTION_URL: jdbc:postgresql://postgres/candlepin

joshmalbrecht (approved):

PR #4334, opened 2023-11-02
Enable sonar base scan for all feature branches

bakajstep:
With new feature branches we were not doing scan on push on them. So sonar was not working. 
Added rule to be able scan all feature branches on push.

### DIFF
--- a/.github/workflows/rerun_workflow_on_push.yml
+++ b/.github/workflows/rerun_workflow_on_push.yml
@@ -6,7 +6,7 @@ on:
     branches:
       - main
       - candlepin-*-HOTFIX
-      - feature/autoreg_v2
+      - feature/*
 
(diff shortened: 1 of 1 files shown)


.github/workflows/rerun_workflow_on_push.yml
```diff
@@ -6,7 +6,7 @@ on:
     branches:
       - main
       - candlepin-*-HOTFIX
-      - feature/autoreg_v2
+      - feature/*
```
nikosmoum, line 9: you would also need to change the `Rerun workflows` step below, because it references `feature/autoreg_v2` specifically

bakajstep: done

PR #4337, opened 2023-11-03
[F] CANDLEPIN-707: Handle rejection for 1P offering systems

nikosmoum:
- First party (1P) offerings are not supported for the autoregv2 MVP implementation. Handle them gracefully with a meaningful error for the clients.

### DIFF
--- a/src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
+++ b/src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
@@ -17,6 +17,7 @@ package org.candlepin.hostedtest;
 import org.candlepin.service.CloudProvider;
 import org.candlepin.service.CloudRegistrationAdapter;
 import org.candlepin.service.exception.CloudRegistrationAuthorizationException;
+import org.candlepin.service.exception.CloudRegistrationNotSupportedForOfferingException;
 import org.candlepin.service.exception.CouldNotAcquireCloudAccountLockException;
 import org.candlepin.service.exception.CouldNotEntitleOrganizationException;
 import org.candlepin.service.exception.MalformedCloudRegistrationException;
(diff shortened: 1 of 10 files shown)


nikosmoum:
Note for reviewers: the low code coverage is only due to no tests for the new exception class' unused constructors, so I think we can safely ignore it and merge

src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
```diff
@@ -129,6 +130,10 @@ public CloudAuthenticationResult resolveCloudRegistrationDataV2(CloudRegistratio
             throw new MalformedCloudRegistrationException(
                 "missing offer ID in registration information metadata");
         }
+        else if (offerId.equals("1POffering")) {
+            throw new CloudRegistrationNotSupportedForOfferingException(
+                "cloud registration v2 is not supported for 1P offerings");
+        }
```
JoshAlbrecht23, line 136: The magic string scares me a bit. I wonder if we should add a offer type field to the body of the `POST cloud/offers` hosted test endpoint and store the type in the `HostedTestDataStore`.  Then we can check the type when we retrieve the product ID based on the offer ID and then throw the `CloudRegistrationNotSupportedForOfferingException` when the type is 1P.

JoshAlbrecht23 (changes requested):
Looks good, I had a question regarding the `1POffering` string.

src/main/java/org/candlepin/hostedtest/HostedTestCloudRegistrationAdapter.java
nikosmoum: I'm undecided if that is needed, or just overcomplicates things. The `HostedTestDataStore` and `HostedTestResource` in general are test code, and the only reason we structure data (like products, subscriptions, etc.) is because those are also reflected on internal candlepin code. While in this case, Candlepin won't be dealing with an offering/product/subscription/pool '1P' type. This piece of data is entirely transparent to candlepin (adapters throw an exception, candlepin throws it further on).

It is not like we'll ever need to add more test scenarios with other 1P offerings

nikosmoum: Updated. See if this is the kind of thing you meant

JoshAlbrecht23 (approved):

PR #4347, opened 2023-11-08
[M] Enable sonar scan for feature branches

bakajstep:


.github/workflows/sonar_branch_analysis.yml
```diff
@@ -6,7 +6,7 @@ on:
     branches:
       - main
       - candlepin-*-HOTFIX
-      - feature/autoreg_v2
+      - feature/*
```
nikosmoum, line 9: Currently this workflow isn't running on the feature/autoreg_v2 branch, because we need to add this line in that branch too

nikosmoum (approved):
Looks good, but we also need this change in the feature/autoreg_v2 branch too

bakajstep:
> Looks good, but we also need this change in the feature/autoreg_v2 branch too

Yes, but i was thinking that we can wait for the rebase. Or should i make these changes rn?

nikosmoum:
Better to make them now, to ensure sonar is run after pushes. And during the rebase there shouldn't be a conflict in the gh workflow file, if it is identical with main anyway

nikosmoum (approved):

nikosmoum:
Also, that probably means we need to do this change also in `feature/entity_layering` and `feature/hms_environment_changes`

PR #4349, opened 2023-11-08
CANDLEPIN-611: Change cert generation to Bouncycastle

wottop:
- Removed uses of JSS in all cases including dependency
- Removed code to extract .war file for JSS .jar contingencies
- Replaced with Bouncycastle implementations of existing interfaces
- Updated unit and spec tests to use new implementations
- Changed test data creation of OID's to account for BC contraints


.github/containers/Dockerfile
```diff
@@ -22,8 +22,7 @@ RUN ./gradlew war -Phostedtest=true && \
     cp $(find ./build/libs -name 'candlepin*.war' | head -n 1) /app/build
 
 WORKDIR /app/build
-RUN jar xf $(find . -name 'candlepin*.war' | head -n 1); \
-    sed -i 's/jss4.jar/jss.jar/g' ./META-INF/context.xml
+RUN jar xf $(find . -name 'candlepin*.war' | head -n 1);
```
nikosmoum, line 25: The only reason for extracting the war file was to update the context.xml file, but if we don't have that file any more you don't need to extract at all (same for the release container image too)

wottop: It doesn't run otherwise.

Januson: Yes, but now that we don't need to make jss adjustments we can start deploying war directly.

src/main/java/org/candlepin/pki/impl/BouncyCastlePKIUtility.java
```diff
@@ -0,0 +1,650 @@
...
+    private byte[] getKeyEncoding(PrivateKey privateKey) throws KeyException {
+        byte[] unwrapped = privateKey.getEncoded();
+
+        if (unwrapped != null) {
+            return unwrapped;
+        }
+
+        try {
+            String algorithm = "AES";
+            String transformation = "AES/CBC/PKCS5Padding";
```
nikosmoum, line 541: We can replace this with "AES/GCM/NoPadding" to resolve the sonarcloud vulnerability alerts.
However, it needs to be tested with subscription-manager (registration, entitlement cert generation, etc.) to make sure this is compatible.

bin/deployment/deploy
```diff
@@ -384,13 +384,6 @@ fi
 # deploy the webapp
 deploy
 
-# Starting with jss 5, the jar is called 'jss.jar' instead of 'jss4.jar'
-if [ -f "/usr/lib64/jss/jss.jar" ]; then
-    $SUDO unzip -q $DEPLOY -d $DEPLOY_DIR
-    $SUDO sed -i 's/jss4.jar/jss.jar/g' $DEPLOY_DIR/META-INF/context.xml
```
nikosmoum, line 390: btw, you also need to update the spec file (candlepin.spec.tmpl) because it's doing the same kind of context.xml manipulation when building the RPM. Looking at the brew error from the jenkins release job from yesterday, it failed because it couldn't find context.xml

src/main/java/org/candlepin/pki/impl/BouncyCastlePKIUtility.java
wottop: This is the same transformation JSS was using. How did we get past it before? Is it because sonarcloud is only monitoring changes?

config/checkstyle/checkstyle.xml
```diff
@@ -177,11 +177,12 @@
         </module>
 
         <!-- This module should be suppressed in our test classes since we use BouncyCastle there to compare
-        our results against what BouncyCastle calculates -->
+        our results against what BouncyCastle calculates
```
Ceiu, line 180: Can we just remove this block entirely now?

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Contents.java
```diff
@@ -78,7 +78,8 @@ public static ContentDTO copy(ContentDTO source) {
      *  a ContentDTO instance with the required fields populated with semi-random values
      */
     public static ContentDTO random() {
-        String cid = StringUtil.random(18, StringUtil.CHARSET_NUMERIC);
+        // replace is needed because this id is used in a OID. BC does not allow leading zeros in a segment
+        String cid = StringUtil.random(18, StringUtil.CHARSET_NUMERIC).replace('0', '1');
```
Ceiu, line 82: Given how this works, rather than do a replacement at the end, change the charset to a custom one that doesn't have a zero in it:
```
String cid = StringUtil.random(18, "123456789");
```

Alternatively, the cid could be prefixed with a 1 to avoid the need to do either:
```
String cid = StringUtil.random("1", 17, StringUtil.CHARSET_NUMERIC);
```

src/main/java/org/candlepin/pki/impl/BouncyCastlePKIUtility.java
nikosmoum: is subman good with this change?

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
```diff
@@ -43,7 +43,8 @@ public static ProductDTO random() {
      *  a product DTO with a randomly generated SKU ID and name
      */
     public static ProductDTO randomEng() {
-        String id = StringUtil.random(8, StringUtil.CHARSET_NUMERIC);
+        // replace is needed because this id is used in a OID. BC does not allow leading zeros in a segment
+        String id = StringUtil.random(8, StringUtil.CHARSET_NUMERIC).replace('0', '1');
```
Ceiu, line 47: Same here -- don't universally replace when you can just remove the bad option from the character set entirely.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/util/CertificateUtil.java
```diff
@@ -148,34 +150,40 @@ private static byte[] fromBase64(byte[] data) {
...
-            if (derOctetValue == null) {
+            byte[] derValue = cert.getExtensionValue(extensionId);
+            if (derValue == null) {
                 return null;
             }
-
-            DerValue value = new DerValue(derOctetValue);
-            byte[] octetString = value.getOctetString();
-            return new DerValue(octetString);
+            ASN1Primitive value = ASN1Primitive.fromByteArray(derValue);
```
Ceiu, line 174: Keep the line break between validation and logic.

spec-tests/src/test/java/org/candlepin/spec/content/ConditionalContentSpecTest.java
```diff
@@ -725,4 +725,7 @@ private String getCertFromEnt(JsonNode ent) {
         return certs.get(0).get("cert").asText();
     }
 
+    private String randomContentId(int length) {
+        return StringUtil.random(5, StringUtil.CHARSET_NUMERIC).replace('0', '1');
```
Ceiu, line 729: Two issues here:
1) same as above where we're erroneously replacing after generating rather than just generating as intended
2) the length argument is ignored entirely

src/main/java/org/candlepin/pki/impl/BouncyCastlePKIUtility.java
```diff
@@ -0,0 +1,650 @@
...
+     * @param publicKey
+     *  the public key to use to build the insecure keypair
+     *
+     * @param privateKey
+     *  the private key to use to build the insecure keypair
+     *
+     * @return
+     *  a keypair which allows extracting the encoded form of both the public and private key
+     */
+    private KeyPair buildInsecureKeyPair(PublicKey publicKey, PrivateKey privateKey)
```
Ceiu, line 509: Do we need the concept of "insecure" keypairs with BouncyCastle? This path is likely obsolete with BC.

The wrapping and unwrapping logic came about because JSS's key implementations disallow getting the raw key bytes at runtime, so we needed to do this to get them. If BC doesn't have the same restriction, we need not bother with the slow wrap/unwrap cycle to get it.

Ceiu (changes requested):

Ceiu:
There should also be something in the commit message describing the minutia of the change.

dependencies.gradle
```diff
@@ -31,6 +31,8 @@ libraries["artemisServer"] = "org.apache.activemq:artemis-server:2.31.2"
 libraries["artemisStomp"] = "org.apache.activemq:artemis-stomp-protocol:2.31.2"
 libraries["assertj"] = 'org.assertj:assertj-core:3.24.2'
 libraries["awaitility"] = 'org.awaitility:awaitility:4.2.0'
+libraries["bouncycastle-prov"] = 'org.bouncycastle:bcprov-jdk18on:1.71'
+libraries["bouncycastle-pkix"] = 'org.bouncycastle:bcpkix-jdk18on:1.74'
```
nikosmoum, line 35: Version 1.77 is out for both of these. Lets update them

wottop:
Successfully smoke tested against RHEL8 and RHEL9 in FIPS mode. Certificate generation is as expected.

nikosmoum (changes requested):
The CONTRIBUTING.md also mentions jss; lets remove those references

.github/workflows/checkstyle.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     steps:
       - name: Install dependencies
         shell: bash
-        run: dnf --setopt install_weak_deps=False install -y gettext git-core jss
+        run: dnf --setopt install_weak_deps=False install -y gettext git-core
```
bakajstep, line 21: With the removal of JSS we no longer need fedora container in our CIs  and if i remebered correctly you can also delete instalation of docker, because it is installed on the ubuntu runner by default.

bakajstep (changes requested):

.github/workflows/checkstyle.yml
wottop: So added

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Contents.java
```diff
@@ -78,7 +78,8 @@ public static ContentDTO copy(ContentDTO source) {
      *  a ContentDTO instance with the required fields populated with semi-random values
      */
     public static ContentDTO random() {
-        String cid = StringUtil.random(18, StringUtil.CHARSET_NUMERIC);
+        // prefix is needed because this id is used in a OID. BC does not allow leading zeros in a segment
+        String cid = StringUtil.random("1", 17, StringUtil.CHARSET_NUMERIC);
```
Ceiu, line 82: This works, and is probably acceptable, but you could have just as easily just removed 0s from the character set:

`String cid = StringUtil.random(18, "123456789");`

src/main/java/org/candlepin/pki/impl/BouncyCastlePKIUtility.java
```diff
@@ -0,0 +1,528 @@
...
+    /**
+     * Fetches the encoded form of the specified private key
+     *
+     * @param privateKey
+     *  the private key from which to fetch the encoded form
+     *
+     * @return
+     *  the encoded form of the given private key
+     */
+    private byte[] getKeyEncoding(PrivateKey privateKey) throws KeyException {
```
Ceiu, line 490: This method is still using the old wrap->unwrap flow to get the key data as bytes that JSS used to impose on us. Everything below 495 is likely unused entirely under BouncyCastle.

Ceiu (changes requested):
Still have some remnants of the insecure keypair logic to remove

Januson (approved):

PR #4353, opened 2023-11-13
[M] Build Github actions spec test containers using docker compose build

JoshAlbrecht23:
- Removed the Github actions build_image job
- Added build section to spec Github actions docker-compose files

Instead of using a Github actions job to build an image and push to the Github container registry for non-forked PRs and then pulling the image when we are running `docker compose up`, we can use the `--build` option when running `docker compose up` and build the image and then run the container in the same command.

Also, for forked PRs we won't have to upload and download the image tar file, we can just run `docker compose up` with the `--build` option.

Build of a forked PR to verify that the changes are compatible with forked PRs: https://github.com/candlepin/candlepin/actions/runs/6856289293/job/18643172392

### DIFF
--- a/.github/workflows/spec_tests.yml
+++ b/.github/workflows/spec_tests.yml
@@ -1,5 +1,5 @@
 ---
-name: Build Candlepin image and run spec tests
+name: Run spec tests
 
 on:
   pull_request:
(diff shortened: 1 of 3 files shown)


nikosmoum (approved):
This is much better than the manual building, so, let's merge it. However, eventually it would be nice to move to podman and podman-based tools (e.g. buildah) to build and deploy container images throughout our project (meaning GH Actions, and dev/prod images)

.github/workflows/spec_tests.yml
```diff
@@ -11,77 +11,6 @@ concurrency:
   cancel-in-progress: true
```
nikosmoum, line 11: While we're here, can we rename the name of this workflow from `Build Candlepin image and run spec tests` to something more concise? Like 'Run spec tests'? The fact that we're building/running container images is just an implementation detail

JoshAlbrecht23: Good catch! I agree on the name change. Updated to 'Run spec tests'

JoshAlbrecht23:
> This is much better than the manual building, so, let's merge it. However, eventually it would be nice to move to podman and podman-based tools (e.g. buildah) to build and deploy container images throughout our project (meaning GH Actions, and dev/prod images)

Yeah, I agree! However, there are issues with running podman inside Docker and that is why I am using Docker here. Unfortunately we have to run a Fedora container inside of a Ubuntu VM where Docker is running in Ubuntu because we need the JSS package to compile Candlepin.

PR #4359, opened 2023-11-15
[F] Update GH action to include feature brnaches

bakajstep:
- Add feature branches into rerun job
- Add feature branches into sonar base scan


.github/workflows/rerun_workflow_on_push.yml
```diff
@@ -27,13 +28,12 @@ jobs:
         shell: bash
         run: |
           base_branch="${GITHUB_REF#refs/heads/}"
-          if [ $base_branch = main ]; then
+          if [[ $base_branch == "main" ]] || [[ $base_branch == "feature/"* ]]; then
```
nikosmoum, line 31: I think `"feature/"*` should be `"feature/*"` here, right?

nikosmoum:
Also, the commit message has a typo: `brnaches` instead of `branches`

.github/workflows/rerun_workflow_on_push.yml
bakajstep: i dont think. [Example](https://stackoverflow.com/a/19897118/19839167)

bakajstep:
> Also, the commit message has a typo: `brnaches` instead of `branches`

done

PR #4364, opened 2023-11-15
[M] Added the manifest generator test extension

Ceiu:
- Added a new test extension for generating manifests without going through the normal creation APIs, as this bypasses certain limitations the APIs have which interfere with test data initialization
- Updated the entire import test suite to use the new manifest generator instead of going through the API
- Updated or moved several tests to better target the function being tested
- Added new checks to CandlepinMode for detecting the presence of various test extensions
- Moved HostedTest to the new testext package
- Fixed a bug in the import process which caused an export metadata record to be persisted if the manifest failed at the second validation step but before beginning the actual import
- Added a new flag, -M, to the deploy script for including the new manifest generator test extension in the build
- Updated the gradle war task to exclude subpackages in the testext package unless they're explicitly included in the "test_extensions" property


Ceiu:
I'm, again, ignoring the Sonarcloud results, because the vast *vast* majority of them are nonsense that lean more on linting/stylistic choices than actual issues, at best; and several of them are straight up incorrect.

Honestly, that entire check is more noise than value. I hope we're not spending actual money on it.

src/main/java/org/candlepin/testext/manifestgen/EntityMapper.java
```diff
@@ -0,0 +1,353 @@
...
+            .setProduct(this.mergeProductData(sdto.getProduct()))
+            .setQuantity(sdto.getQuantity())
+            .setStartDate(Util.toDate(sdto.getStartDate()))
+            .setEndDate(Util.toDate(sdto.getEndDate()))
+            .setContractNumber(sdto.getContractNumber())
+            .setAccountNumber(sdto.getAccountNumber())
+            .setOrderNumber(sdto.getOrderNumber());
+
+            // Apparently subscription DTOs don't have attributes!?
+            // .setAttributes(this.convertAttributes(sdto.getAttributes()))
```
JoshAlbrecht23, line 254: Just want to double check and see if you intentionally wanted to leave this commented line

JoshAlbrecht23 (changes requested):
Looks good! I just noticed a comment that might need updating and had a question for a commented out line

spec-tests/src/main/java/org/candlepin/spec/bootstrap/assertions/OnlyWithManifestGen.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.jupiter.api.condition.EnabledIf;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation marks a test or test suite to run only in hosted mode.
+ */
```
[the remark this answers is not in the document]
Ceiu: This should be rewritten to look more like the `OnlyWithCapability` annotation (as should `OnlyInHosted` and `OnlyInStandalone`) since `EnabledIf`/`DisabledIf` are implemented so poorly and unreliably.

For now I've just removed it since none of the tests use it and instead rely on manual use of `assumeTrue` for conditional execution.

src/main/java/org/candlepin/testext/manifestgen/EntityMapper.java
Ceiu: Yeah, I'm fine leaving that there for now. I feel like it's just a matter of time before someone needs to set specific attributes on the manifest pool and then we'll need to update the API accordingly to enable that line.

nikosmoum: Ugh, the SubscriptionDTO is defined in the openapi spec file because it is actually used in production APIs. Adding fields strictly required for testing will pollute our API. Can we add to this comment, that if anyone wants to add attributes, please create a new SubscriptionDTO only for the testext layer, or at least extend the existing SubscriptionDTO, instead of adding to it directly?

nikosmoum (approved):
Looks good to me. I just have a couple of queries

src/main/java/org/candlepin/testext/manifestgen/ManifestGeneratorResource.java
```diff
@@ -0,0 +1,333 @@
...
+            log.error("Unable to deserialize attached file:", e);
+            throw new IseException("Unable to deserialize attached file", e);
+        }
+    }
+
+    @POST
+    @Path("export")
+    @Produces("application/zip")
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    public Response generateManifestFromFile(
```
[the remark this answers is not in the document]
Ceiu: Same way the upstream subscription injector is: it supersedes the need for a complicated client-side script (that I, unfortunately, already wrote in the previous PR). I could just update the script in the same way the injector was updated to simply hit its respective endpoint.

src/main/java/org/candlepin/testext/manifestgen/EntityMapper.java
Ceiu: Done.

JoshAlbrecht23 (approved):

PR #4370, opened 2023-11-17
[M] CANDLEPIN-318: Added new utility for abstract universal paging logic

Ceiu:
- Added a new utility, PagingUtil, which provides methods to apply paging to streams and collections
- Moved AbstractHibernateObject.DEFAULT_SORT_FIELD to PageRequest to better align with the other defaults and to further separate paging logic from the database model
- Updated Page and PageRequest's void mutators to return a reference to the object for fluent-style invocation chaining

### DIFF
--- a/src/main/java/org/candlepin/paging/FieldComparatorFactory.java
+++ b/src/main/java/org/candlepin/paging/FieldComparatorFactory.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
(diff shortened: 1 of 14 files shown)


Ceiu:
This implementation accidentally skipped the design phase, as this is something I implemented in the entity layering work out of necessity; I had forgotten we even had such a task.

That said, if there are any implementation or design weirdness that seems off, please call it out.

JoshAlbrecht23 (approved):
Looks good!

src/main/java/org/candlepin/paging/FieldComparatorFactory.java
```diff
@@ -0,0 +1,65 @@
...
+ * @param <T>
+ *  the class type for which this factory's mapping logic applies
+ */
+@FunctionalInterface
+public interface FieldComparatorFactory<T> {
+
+    /**
+     * Fetches a comparator which orders elements of this comparator factory's type by using the
+     * value returned by accessor best matching the given field name. The comparator this method
+     * returns should always order elements in ascending order. If the type supported by this
```
nikosmoum, line 34: Why ascending when the default is descending?

Ceiu: To keep the comparator logic decoupled from our API convention. The PagingUtil will reverse the comparator as necessary based on whatever the request needs.

PR #4376, opened 2023-11-17
Tighten up job wait status checking for spec tests

wottop:


nikosmoum:
Closing this in favor of https://github.com/candlepin/candlepin/pull/4382 which seems to have fixed the root cause

PR #4381, opened 2023-11-22
CANDLEPIN-699: Sonar issue with logging user data

wottop:


nikosmoum (changes requested):
I think we should avoid sprinkling the code with these annotations. You can directly mark the issues as either 'false positive' or 'won't fix' with a comment in sonarcloud UI itself.

PR #4382, opened 2023-11-23
Fix job spec test timeouts

Januson:
Isolated spec tests requiring stopping of job scheduler from the rest of spec tests.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
@@ -14,8 +14,8 @@
  */
 package org.candlepin.spec.jobs;
 
-import static java.lang.Thread.sleep;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


spec-tests/src/test/java/org/candlepin/spec/jobs/JobStatusSpecTest.java
```diff
@@ -418,4 +313,90 @@ public void shouldNotAllowConsumerToCancelOthersJob() throws Exception {
...
+
+    @Nested
+    @Isolated
+    @Execution(ExecutionMode.SAME_THREAD)
+    class WithScheduler {
+
+        @AfterEach
+        void tearDown() {
+            jobsClient.setSchedulerStatus(true);
+        }
```
nikosmoum, line 325: :+1:

nikosmoum:
Note that the `shouldCreateOwnerWhenNoOwnerExistsForCloudAccountDuringV2Auth` test failure is flaky, but unrelated to the root cause that this PR is trying to fix. That test will be fixed in another PR

PR #4388, opened 2023-11-27
[M] Expunge scripts from the olden days

nikosmoum:


nikosmoum:
@Ceiu Since you're the longest in the team, can you take a look and see if any of these still have value?

Ceiu:
I don't think these are used anymore, nor have they been used for several years. Most of them I haven't even bothered to read before today.

As an aside, we can also remove the `refresh_data_extractor.sh` script, as that has also been obsoleted by better tooling.

nikosmoum:
> As an aside, we can also remove the `refresh_data_extractor.sh` script, as that has also been obsoleted by better tooling.

done

Ceiu (approved):

PR #4400, opened 2023-11-30
[M] Fix ConcurrentModificationException in import spectests

nikosmoum:
- The import spec tests would sometimes fail with a ConcurrentModificationException when trying to setup the available test extensions map.


Januson (approved):

nikosmoum:
> If you have time, the check should be updated to basically be a copy/paste of the logic for OnlyWithCapability, and then we can use it like the other environment-based test annotations.

Lets get this one merged now to unblock other PRs, and I can put up a PR with this change next week

PR #4403, opened 2023-12-01
[4.3] CANDLEPIN-727: Revert cpdb changes & unwanted satellite bug fixes

nikosmoum:
This PR reverts:
- The cpdb changes, because they break the satellite installer.
  They will be integrated in a future satellite release.
- Satellite bug fixes 2231503 and 2213261, because they are
  unwanted for the upcoming z-stream. They will be picked up
  in a future satellite release.


nikosmoum:
> Does cpdb work correctly with the liquibase changes being performed by CP, and with the refactor of the root changelogs?

cpdb's behaviour didn't really change, only its interface. No matter if you're running `cpdb --update-schema` with the changes, or `cpdb --update` or `cpdb --create --schema-only` before the changes, the final command being run is the same:
```
/usr/share/candlepin/liquibase.sh --driver=org.postgresql.Driver
--classpath=/var/lib/tomcat/webapps/candlepin/WEB-INF/lib/postgresql-42.6.0.jar:build/classes:/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/
--changelog-file=db/changelog/changelog-update.xml
--url="jdbc:postgresql:candlepin" --username=$DBUSERNAME
--headless=true --log-level=severe update -Dcommunity=False
```
So reverting this cpdb change has no bearing on if the liquibase or the changelog file refactors will work or not.
However just to be I'll test the upgrade

nikosmoum:
@Ceiu 
Smoke Test:
- Checked out candlepin-4.3.1-1 (which satellite is currently on) and deployed with the deploy script (deploy -ag).
- Switched to this branch (nikos/4.3/revert_sat_breakages).
- Built the war (./gradlew war) and deployed it manually under /var/lib/tomcat/webapps/.
- In candlepin.log I can see the `2023-12-04 17:05:36,545 [thread=main] [=, org=, csid=] INFO  org.candlepin.guice.CandlepinContextListener - Liquibase startup management set to NONE` as expected, so no liquibase updates made during startup. NONE is the default value in ConfigProperties, which is what satellite expects.
- Then I run cpdb manually: `cpdb --update`, which successfully updated the schema, 
by updating the checksums for hundreds of changesets, and also run the single schema update:
```
...
Upgrading checksum for Changeset db/changelog/20230220150433-drop-act-key-env-tables.xml::20230220150433-2::wpoteat from 8:5492b59e9db540f8db8cf019b22a4f63 to 9:c057fdccd02e947b9654995b04ee3f7d.
Running Changeset: db/changelog/20230322103117-reduce-consumer-name.xml::20230322103117-1::wpoteat

UPDATE SUMMARY
Run:                          1
Previously run:             606
Filtered out:                 0
-------------------------------
Total change sets:          607

Liquibase: Update has been successful. Rows affected: 607
Liquibase command 'update' was executed successfully.
```
- Switched to main branch, and manually build and deployed the war again.
- Checked that the liquibase update on tomcat startup was a no-op again.
- Finally run `cpdb --update-schema` which successfully run 9 changesets.

PR #4407, opened 2023-12-05
[Entity Layering] CANDLEPIN-676: Removed OrphanCleanupJob

Ceiu:
- Removed the obsoleted OrphanCleanupJob and all related configuration and references


wottop:
candlepin.refresh.orphan_entity_grace_period is still in the CONTRIBUTING.md file.
Do we still need the orphaned_date column?

Ceiu:
Fixed the references and config from the containers and whatnot.

The column will still be retained because the table and all of that junk will live on for a bit. In a few releases after this gets merged, those will eventually be cleaned up and it'll be removed at that point.

Ceiu:
Looks like a bunch of markdown line breaks got removed, but in looking through the rendered output, I'm not sure they were doing anything significant anyway. We aren't consistently breaking on a specific line width, nor are we applying such a limit globally. Going to leave them removed.

nikosmoum (approved):

PR #4414, opened 2023-12-06
CANDLEPIN-728: Enable tracing for calls between Candlepin async job and RHSM-API

bakajstep:
- Enable tracing for CloudAccountOrgSetupJob


src/main/java/org/candlepin/async/tasks/CloudAccountOrgSetupJob.java
```diff
@@ -71,6 +73,7 @@ public CloudAccountOrgSetupJob(CloudRegistrationAdapter cloudAdapter, OwnerCurat
      *     if an exception occurs during job execution
      */
     @Override
+    @WithSpan
```
nikosmoum, line 76: does the '@WithSpan' annotation generate a traceID by itself? Or does it only propagate an existing tradeID if one exists?
I thought its just the former, in which case there will be no traceID to propagate to adapter and then later on to RHSM-API

bakajstep: it should, because every span has ID(Span and trace). [Traces](https://opentelemetry.io/docs/concepts/signals/traces/). But i will verify that after i will create test enviroment for that. But then depends if they want it as seperate trace or include in the trace where they are calling it

nikosmoum (approved):

PR #4417, opened 2023-12-07
Fix checkstyle dependency resolution

Januson:


buildSrc/src/main/groovy/checkstyle-conventions.gradle
```diff
@@ -12,6 +12,12 @@ dependencies {
     checkstyle libraries["checkstyleSevntu"]
 }
 
+configurations.checkstyle {
```
nikosmoum, line 15: just a comment needed to explain what/why this is doing before ready for merging

PR #4419, opened 2023-12-08
[Templating] CANDLEPIN-695: Update environment APIs to support typing

wottop:


Ceiu (commented):
We also need to add something a bit more descriptive in the git commit describing the what/how/why of the change, and document the net-new functionality -- especially at the curator level which is used like a library.

PR #4422, opened 2023-12-11
[Entity Layering] CANDLEPIN-674: Added data migration script for products and content

Ceiu:
- Added the data migration task for migrating product and content data from the versioned-entity schema to the layered-entity schema

### DIFF
--- a/src/main/resources/db/changelog/20231003113535-entity_layering_schema.xml
+++ b/src/main/resources/db/changelog/20231003113535-entity_layering_schema.xml
@@ -354,7 +354,357 @@
         <dropForeignKeyConstraint baseTableName="cp_pool" constraintName="cp_pool_fk1"/>
     </changeSet>
 
-    <!-- TODO: Add migration logic here -->
+   <changeSet id="20231003113535-11" author="crog" dbms="postgresql, mysql, mariadb">
+        <sql>
+            -- Migrate global products and content
+            DROP TABLE IF EXISTS tmp_active_global_products;
+            CREATE TEMPORARY TABLE tmp_active_global_products (
+                uuid VARCHAR(32) NOT NULL,
+                pid VARCHAR(32) NOT NULL,
+                new_uuid VARCHAR(32) NOT NULL,
+                best BOOLEAN NOT NULL DEFAULT false,
+                PRIMARY KEY (uuid));
+
+            DROP TABLE IF EXISTS tmp_active_global_contents;
+            CREATE TEMPORARY TABLE tmp_active_global_contents (
+                uuid VARCHAR(32) NOT NULL,
+                cid VARCHAR(32) NOT NULL,
+                new_uuid VARCHAR(32) NOT NULL,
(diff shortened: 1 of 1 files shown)


src/main/resources/db/changelog/20231003113535-entity_layering_schema.xml
```diff
@@ -354,7 +354,357 @@
...
+                            SELECT product_uuid AS uuid FROM cp_pool) active ON active.uuid = prod.uuid
+                        WHERE prod.locked != 0
+                    UNION
+                    SELECT pcmap.child_uuid, pg.depth + 1
+                        FROM parent_child_map pcmap
+                        JOIN product_graph pg ON pg.uuid = pcmap.parent_uuid
+                        -- shouldn't be needed, but here just in case someone decided to be cheeky and define a
+                        -- cyclical product, or some other crazy thing that's entirely unsupported by CP's logic
+                        -- and queries
+                        WHERE pg.depth &lt; 9)
```
nikosmoum, line 399: is level 9 really needed here? isn't our current hierarchy a max depth of 2 (excluding contents, marketing product is 0, derived product is 1, provided product 2)?

src/main/resources/db/changelog/20231003113535-entity_layering_schema.xml
```diff
@@ -354,7 +354,357 @@
...
+                SELECT prod.uuid, prod.product_id, md5(concat('--global--', prod.product_id)) AS new_uuid, prod.uuid = best.uuid AS best_uuid
+                    FROM (SELECT DISTINCT uuid FROM product_graph) active
+                    JOIN cp2_products prod ON prod.uuid = active.uuid
+                    JOIN (SELECT prod2.product_id, MAX(prod2.uuid) AS uuid
```
nikosmoum, line 403: is this MAX depending on the fact that our uuids are incremental (so a newer version always has a higher uuid)?

Ceiu: Very good observation! This was why I picked it as the tie breaker, but that behavior isn't required for it. The inner query is our primary selector, using the most recent updated date for a given product ID (and owner ID + product ID in the custom block), only falling back to the `MAX(prod2.uuid)` bit in the unlikely case we have multiple entities with the exact same last update date.

In terms of how the query behaves, it's not required to pick the "newest" UUID, just that we pick a consistent UUID.

Ceiu: As an added bit of history here: I think the tie breaker selection mattered more when I was manually doing that "best" lookup more than once. When I moved the flag to the active_custom and active_global tables, the consistency tie breaker probably became less valuable, but it doesn't hurt to retain it.

nikosmoum (approved):

nikosmoum:
@Ceiu do you need to retest the queries after the last change, or should I go ahead and merge?

Ceiu:
@nikosmoum It should be good to merge.

PR #4428, opened 2023-12-13
CANDLEPIN-695: Update environment APIs to support typing

wottop:


spec-tests/src/main/java/org/candlepin/spec/bootstrap/client/api/OwnerClient.java
```diff
@@ -266,12 +266,14 @@ public AsyncJobStatusDTO importManifestAsync(String ownerKey, File input) throws
         return super.importManifestAsync(ownerKey, List.of(), input);
     }
 
+    // this is intended to be type agnostic
```
Ceiu, line 269: javadoc block instead of an inline comment

spec-tests/src/test/java/org/candlepin/spec/environments/EnvironmentSpecTest.java
```diff
@@ -144,6 +144,44 @@ public void shouldFindEnvironmentByName() {
...
+    public void shouldFindTypedEnvironments() {
+        EnvironmentDTO env1 = ownerClient.owners().createEnv(owner.getKey(), Environments.random()
+            .type(null));
+        EnvironmentDTO env2 = ownerClient.owners().createEnv(owner.getKey(), Environments.random()
+            .type("test_type"));
+        EnvironmentDTO env3 = ownerClient.owners().createEnv(owner.getKey(), Environments.random()
+            .type("another_type"));
+
+        List<EnvironmentDTO> found1 = ownerClient.environments().getEnvironments(null, false);
+        assertThat(found1).contains(env1);
```
Ceiu, line 157: `.contains` only verifies that the element(s) is contained; it doesn't do verification that other elements *aren't* present.

This should probably be something like `.containsOnly` along with a `.isNotNull()`

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -71,19 +69,88 @@ public List<Consumer> getEnvironmentConsumers(String environmentId) {
             .getResultList();
     }
 
-    public CandlepinQuery<Environment> listForOwner(Owner o) {
-        DetachedCriteria criteria = this.createSecureDetachedCriteria()
-            .add(Restrictions.eq("owner", o));
+    public List<Environment> listForOwnerByType(Owner o, List<String> type) {
```
Ceiu, line 72: Few things across these methods:
- missing documentation
- missing input validation (nulls and empty values should be safe or fail fast and explicitly)
- avoid single-character variable names in these cases

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -71,19 +69,88 @@ public List<Consumer> getEnvironmentConsumers(String environmentId) {
...
+        String jpql = "SELECT e FROM Environment e WHERE e.owner.id = :owner_id and e.type in :type";
+        return this.getEntityManager()
+            .createQuery(jpql, Environment.class)
+            .setParameter("owner_id", o.getId())
+            .setParameter("type", type.stream().map(x -> x.toLowerCase()).toList())
+            .getResultList();
+    }
+
+    public List<Environment> listForOwnerIgnoreType(Owner o) {
```
Ceiu, line 81: nit: naming convention for this type of query is generally `listAllBy[distinguishing criteria]`

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -71,19 +69,88 @@ public List<Consumer> getEnvironmentConsumers(String environmentId) {
...
+        String jpql = "SELECT e FROM Environment e WHERE e.owner.id = :owner_id";
+        return this.getEntityManager()
+            .createQuery(jpql, Environment.class)
+            .setParameter("owner_id", o.getId())
+            .getResultList();
+    }
+
+    public List<Environment> listForOwnerNullType(Owner o) {
```
Ceiu, line 89: nit: `listUntyped[Environments]ByOwner`

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -71,19 +69,88 @@ public List<Consumer> getEnvironmentConsumers(String environmentId) {
...
+        String jpql = "SELECT e FROM Environment e WHERE e.owner.id = :owner_id and e.type is null";
+        return this.getEntityManager()
+            .createQuery(jpql, Environment.class)
+            .setParameter("owner_id", o.getId())
+            .getResultList();
+    }
 
-        return this.cpQueryFactory.buildQuery(this.currentSession(), criteria);
+    public List<Environment> listForOwnerByNameAndType(Owner o, String envName, List<String> type) {
```
Ceiu, line 97: nit: For the methods where the overall behavior isn't changing but the criteria is, we can just use one method name and overload it to avoid the Java naming stereotype in some of these cases.

Ceiu (changes requested):
Need some additional testing on the unit test side to verify the bad-input test cases (null, empty, invalid, missing types, names, etc.)

Also, if we're approaching the point where we need six variants for the queries, maybe it's time we look at something like the filter object ala JobStatusResult or Consumer filters. Though it looks like the owner and name are both optional parameters across everything, so those could just be individual overloads to only have one or two actual query implementations if we don't want to go that far.

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -255,8 +260,21 @@ private void deleteConsumers(Environment environment, List<Consumer> consumers)
...
+        else if (type == null || type.size() == 0) {
+            envs = this.envCurator.listAllNullType();
+        }
+        else {
+            envs = this.envCurator.listAllByType(type);
+        }
+        Stream<EnvironmentDTO> stream = envs.stream()
+            .map(this.translator.getStreamMapper(Environment.class, EnvironmentDTO.class));
+        return this.pagingUtilFactory.forClass(EnvironmentDTO.class)
+            .applyPaging(stream, envs.size());
```
nikosmoum, line 277: Up until now, only satellites used environments, and with single-digit amount of environments per org, and single-digit amount of orgs in a satellite, this API would never return more than double-digit amount of environments (realistically, no more than 25 in extreme cases), so this would be fine.

However, with this feature, environments are now in Hosted, which has around ~133k active owners (millions total), and it is rising every month. Assuming even only a fifth of these active owners use this feature eventually, and an average of 2 environments per owner, this API will fetch ~50k environment instances into memory every time its called and paginate after the fact, which is not good.

Instead of the PagingUtilFactory we need to do paging at the DB query level, similarly to `JobResource.listJobStatuses` and `ConsumerResource.searchConsumers` (and limit the amount of results returned if no paging was requested by throwing an error after a reasonable amount of results would be returned).

The owners/owner_key/environments can use the pageutil because it won't even return enough environments to be an issue

nikosmoum: Also, looking at Katello code, they don't use this endpoint, and looking at the HMS content-sources list of endpoints they want access to, they're not going to use this endpoint either.

Which makes it a good candidate for removal, or alternatively leaving it for debugging purposes.
@Ceiu @wottop what do you think?

nikosmoum (changes requested):

src/main/java/org/candlepin/resource/EnvironmentResource.java
Ceiu: Given how environments are used, the "get me all environments globally" query is one of those ones that didn't make sense to me, but I didn't look into it at all. If no one is using it, I'm down for just removing it.

If we opt to keep it, Nikos' assessment of the potential data load is fair and I would agree that we should apply more thorough paging.

nikosmoum: Personally I don't see how this endpoint would be helpful for debugging, so lets remove it.

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -71,19 +70,61 @@ public List<Consumer> getEnvironmentConsumers(String environmentId) {
...
+     * @param owner
+     *  The owner of the environments
+     * @param envName
+     *  An exact name for the environments
+     * @param type
+     *  A list of types to filter the environments
+     * @return
+     *  A List of environments based on the criteria
+     */
+    public List<Environment> listByType(@NotNull Owner owner, String envName, List<String> type) {
```
Ceiu, line 84: The `@NotNull` annotation doesn't inherently do any validation -- it needs to be checked with a validator. Both of which are pretty heavy for doing input validation here (not to mention the silliness that is annotation-based almost-meta programming), so we're better off with a traditional null check and IAE/fast exit.

src/test/java/org/candlepin/model/EnvironmentCuratorTest.java
```diff
@@ -54,25 +57,62 @@ public void delete() {
     }
```
Ceiu, line 58: Should also add tests for verifying null/empty inputs (can be done nicely with `@ParameterizedTest` and `@NullAndEmptySource`) and invalid inputs (such as env names and/or types that don't exist).

Ceiu (changes requested):

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -98,6 +99,7 @@ public class EnvironmentResource implements EnvironmentApi {
     private final ContentAccessCertificateCurator contentAccessCertificateCurator;
     private final EntitlementEnvironmentFilter entitlementEnvironmentFilter;
     private final EntitlementCertificateGenerator entCertGenerator;
+    private final PagingUtilFactory pagingUtilFactory;
```
nikosmoum, line 102: No need for this now that the endpoint is removed entirely

nikosmoum (approved):
Approved from me. I'll let @Ceiu re-review/merge

Ceiu (approved):

PR #4429, opened 2023-12-14
Close an unclosed stream

Januson:


Ceiu (changes requested):
Looks good, but can we add a commit message detailing the change a bit? Was this causing an issue outside of resource utilization?

Januson:
Added the message, but there's not much to add. I noticed an unclosed stream so I closed it.

Ceiu (approved):

PR #4433, opened 2023-12-14
CANDLEPIN-742: Replaced CloudProvider enum for String

joshmalbrecht:


Ceiu (changes requested):
Looks good. Let's just add a commit message to go along with the change.

Januson (commented):
Changes look good. Though I wonder if it would be worth going for the value object instead of a string. It would make our code less stringly typed and it would give us a nice place for its validation logic.

joshmalbrecht:
> Changes look good. Though I wonder if it would be worth going for the value object instead of a string. It would make our code less stringly typed and it would give us a nice place for its validation logic.

Interesting. Yeah, we would encapsulate the validation like the character limit, null, and empty check in the value object class instead of the entity class.

joshmalbrecht:
> Changes look good. Though I wonder if it would be worth going for the value object instead of a string. It would make our code less stringly typed and it would give us a nice place for its validation logic.

We had a discussion on this and Shayne gave his preference of using a String. However, this pattern is good to keep in mind for the future, as it can be useful!

https://redhat-internal.slack.com/archives/C04H33UF5U6/p1702667437384879

Ceiu (approved):

nikosmoum (approved):
Regarding the 15 char limit: RHSM api which owns the data has a limit of 4 right now, so 15 from our side is generous enough to accommodate a future increase.

Merging this since we don't plan to do anything more with this value other than receive it, store it, send it back when needed. If we ever need anything more in the future it could be changed to a value object.

PR #4437, opened 2023-12-15
[M] CANDLEPIN-736: Updated entitlement certificate namespace logic

Ceiu:
- Updated the logic for entitlement certificate generation to only include the top-level product namespace OID if the namespace is a non-null, non-blank value
- Added additional spec tests to verify the presence or non-presence of the entitlement namespace OID in the certificates
- Updated the unit tests for entitlement namespace extension generation

### DIFF
--- a/spec-tests/src/main/java/org/candlepin/spec/bootstrap/assertions/JobStatusAssert.java
+++ b/spec-tests/src/main/java/org/candlepin/spec/bootstrap/assertions/JobStatusAssert.java
@@ -15,6 +15,7 @@
 package org.candlepin.spec.bootstrap.assertions;
 
 import org.candlepin.dto.api.client.v1.AsyncJobStatusDTO;
+import org.candlepin.spec.bootstrap.client.ApiClient;
 
 import org.assertj.core.api.AbstractAssert;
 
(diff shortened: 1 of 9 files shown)


spec-tests/src/main/java/org/candlepin/spec/bootstrap/assertions/JobStatusAssert.java
```diff
@@ -36,6 +37,68 @@ public static JobStatusAssert assertThatJob(AsyncJobStatusDTO actual) {
...
+     * The object under test will be mapped to the final AsyncJobStatusDTO upon successful
+     * completion.
+     *
+     * @param client
+     *  the client to use to verify the job has completed
+     *
+     * @return
+     *  a new JobStatusAssert for the completed job status
+     */
+    public JobStatusAssert terminatesWithinDuration(ApiClient client, long duration) {
```
joshmalbrecht, line 55: Missing the duration param in the javadocs

spec-tests/src/test/java/org/candlepin/spec/entitlements/EntitlementCertificateV3SpecTest.java
```diff
@@ -475,6 +477,59 @@ public void shouldEncodeManyContentUrls() {
...
+        PoolDTO pool = pools.stream()
+            .filter(elem -> productId.equals(elem.getProductId()))
+            .findAny()
+            .get();
+
+        List<EntitlementDTO> entitlements = consumerApi.bindPoolSync(system.getUuid(), pool.getId(), 1);
+
+        assertThat(entitlements)
+            .isNotNull()
+            .hasSize(1);
```
joshmalbrecht, line 527: We can use `singleElement()` here

joshmalbrecht (changes requested):
Looks good! Just noticed some small things

spec-tests/src/main/java/org/candlepin/spec/bootstrap/assertions/JobStatusAssert.java
Ceiu: Fixed

spec-tests/src/test/java/org/candlepin/spec/entitlements/EntitlementCertificateV3SpecTest.java
Ceiu: Yeah, but unless I'm actually mapping the collection to the element for further validation, I tend to prefer just specifying the size.

joshmalbrecht (approved):

PR #4439, opened 2023-12-18
cloud.enable set to true when hosted flag is used on deploy

wottop:
As it was set, one of the spec tests fails.


nikosmoum (changes requested):
Can you also remove the `bin/scripts/generate_custom_yaml.py` file, which used to set the `candlepin.auth_cloud_enable`? it doesn't seem to be used any more

nikosmoum (approved):

PR #4445, opened 2023-12-20
[M] CANDLEPIN-743: Fix autoregv2 reauthentication & other flow updates

nikosmoum:
- Fix issue with checking if the relevant pools exist in Candlepin during registration with an anonymous token, which should not be checked when the owner does not even exist upstream of Candlepin. Also re-factored this method to read better.
- Make sure the CloudAccountOrgSetupJob creates anonymous owners with SCA as the current and only possible content access mode.
- Make sure the CloudAccountOrgSetupJob creates non-anonymous owners with the default content access mode and content access mode list.
- Remove CloudAccountOrgSetupJob from the list of scheduled jobs since it isn't one.

### DIFF
--- a/src/main/java/org/candlepin/resource/CloudRegistrationResource.java
+++ b/src/main/java/org/candlepin/resource/CloudRegistrationResource.java
@@ -165,14 +165,9 @@ public class CloudRegistrationResource implements CloudRegistrationApi {
         CloudAuthenticationResult authResult) {
         validateCloudAuthenticationResult(authResult);
 
-        String cloudInstanceId = authResult.getCloudInstanceId();
-        AnonymousCloudConsumer existingAnonConsumer = anonymousCloudConsumerCurator
-            .getByCloudInstanceId(cloudInstanceId);
-
-        String ownerKey = authResult.getOwnerKey();
-
         // verify that the owner exists upstream and is entitled
-        boolean createAnonConsumer = false;
+        String ownerKey = authResult.getOwnerKey();
+        boolean isOwnerReadyForRegistration = false;
         if (ownerKey == null || ownerKey.isBlank() || !authResult.isEntitled()) {
             CloudAccountOrgSetupJobConfig jobConfig = CloudAccountOrgSetupJob.createJobConfig()
                 .setCloudAccountId(authResult.getCloudAccountId())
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -193,60 +188,62 @@ private CloudAuthenticationResultDTO processAdapterAuthResult(Principal principa
...
-        CloudAuthenticationResultDTO cloudAuthResultDTO = new CloudAuthenticationResultDTO()
+            cloudAuthResultDTO = new CloudAuthenticationResultDTO()
             .ownerKey(ownerKey)
             .token(token)
-            .tokenType(tokenType.toString());
-
-        if (anonymousConsumerUuid != null) {
-            cloudAuthResultDTO.anonymousConsumerUuid(anonymousConsumerUuid);
+            .tokenType(tokenType.toString())
+            .anonymousConsumerUuid(anonymousConsumerUuid);
```
joshmalbrecht, line 246: These lines should be indented

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -193,60 +188,62 @@ private CloudAuthenticationResultDTO processAdapterAuthResult(Principal principa
...
-            createdAnonConsumer = anonymousCloudConsumerCurator.create(createdAnonConsumer);
-            anonymousConsumerUuid = createdAnonConsumer.getUuid();
-
-            log.info("Anonymous consumer created for instance {} using cloud account {} {}",
-                authResult.getCloudInstanceId(), authResult.getCloudProvider(),
-                authResult.getCloudAccountId());
+            cloudAuthResultDTO = new CloudAuthenticationResultDTO()
+            .ownerKey(ownerKey)
+            .token(token)
+            .tokenType(tokenType.toString());
```
joshmalbrecht, line 210: Need indention on these lines

src/test/java/org/candlepin/async/tasks/CloudAccountOrgSetupJobTest.java
```diff
@@ -143,6 +143,39 @@ void ensureJobSuccessWithNewOrganization()
...
     }
 
+    @Test
+    void ensureJobSuccessWithNewNonAnonymousOrganization()
+        throws JobExecutionException, CouldNotAcquireCloudAccountLockException {
+        when(cloudReg.setupCloudAccountOrg(anyString(), anyString(), any(), anyString()))
+            .thenReturn(new CloudAccountData("owner_key", false));
+
+        CloudAccountOrgSetupJob regJob = new CloudAccountOrgSetupJob(cloudReg, ownerCurator);
+
```
joshmalbrecht, line 153: Extra new line

src/test/java/org/candlepin/async/tasks/CloudAccountOrgSetupJobTest.java
```diff
@@ -143,6 +143,39 @@ void ensureJobSuccessWithNewOrganization()
...
+        AsyncJobStatus status = mock(AsyncJobStatus.class);
+        JobExecutionContext context = spy(new JobExecutionContext(status));
+        when(status.getJobArguments()).thenReturn(jobConfig.getJobArguments());
+
+        ArgumentCaptor<Object> captor = ArgumentCaptor.forClass(Object.class);
+
+        regJob.execute(context);
+
+        verify(context, times(1)).setJobResult(captor.capture());
```
joshmalbrecht, line 171: You can use just `verify(context)` here instead of adding the `times(1)` parameter

joshmalbrecht (changes requested):
Looks good!! I just noticed some small changes

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
nikosmoum: Done

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
nikosmoum: Done

src/test/java/org/candlepin/async/tasks/CloudAccountOrgSetupJobTest.java
nikosmoum: Done

src/test/java/org/candlepin/async/tasks/CloudAccountOrgSetupJobTest.java
nikosmoum: Done

joshmalbrecht (approved):

PR #4455, opened 2024-01-02
Migrate to Gradle Version Catalog

Januson:
- Added Gradle version catalog file gradle/libs.versions.toml
- Replaced dependency definitions with generated catalog accessors
- Removed dependencies.gradle


Ceiu (approved):
This looks better to my eye, but I'm very much not an expert when it comes to Gradle.

joshmalbrecht (approved):

PR #4462, opened 2024-01-02
[M] CANDLEPIN-733: Created anonymous certificate content cache

joshmalbrecht:
- Created ConsumerCertificateContentCache to cache cert content
- Added the cache to the ContentAccessManager and attempt to use the cached content when creating anonymous content access certificates
- Removed unused CloudRegistrationAdapter and PoolCurator from the ConsumerResource class

I have some concerns about needing to add an upper bound entry limit to the ConsumerCertificateContentCache to prevent the cache from consuming potentially too much memory.  I created the following jira for determining if we should limit the max number of entries:
https://issues.redhat.com/browse/CANDLEPIN-759

### DIFF
--- a/src/main/java/org/candlepin/cache/AnonymousCertContentCache.java
+++ b/src/main/java/org/candlepin/cache/AnonymousCertContentCache.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.cache;
+
+import org.candlepin.config.ConfigProperties;
+import org.candlepin.config.Configuration;
+import org.candlepin.config.ConfigurationException;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+
+import java.time.Duration;
+import java.util.Collection;
+import java.util.Objects;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * A thread safe cache for consumer certificate content. Entries are evicted based on a time-to-live eviction
+ * policy.
+ */
+@Singleton
+public class AnonymousCertContentCache {
+    private Cache<MultiKey, AnonymousCertContent> cache;
+
+    @Inject
+    public AnonymousCertContentCache(Configuration config) throws ConfigurationException {
+        Objects.requireNonNull(config);
+
+        long expirationDuration = config.getLong(ConfigProperties.CACHE_ANON_CERT_CONTENT_TTL);
+        if (expirationDuration <= 0) {
+            String msg = ConfigProperties.CACHE_ANON_CERT_CONTENT_TTL + " value must be larger than 0";
+            throw new ConfigurationException(msg);
+        }
+
+        cache = Caffeine.newBuilder()
+            .expireAfterWrite(Duration.ofMillis(expirationDuration))
+            .build();
+    }
+
+    /**
+     * Retrieves cached {@link AnonymousCertContent} for the provided top level SKU IDs.
+     *
+     * @param skuIds
+     *  the top level SKU IDs to retrieve consumer certificate content for
+     *
+     * @throws IllegalArgumentException
+     *  if the provided SKU IDs are null
+     *
+     * @return
+     *  the cached consumer certificate content for the provided top level SKU IDs
+     */
+    public AnonymousCertContent get(Collection<String> skuIds) {
+        if (skuIds == null) {
+            throw new IllegalArgumentException("sku ID is null");
+        }
+
+        if (skuIds.isEmpty()) {
+            return null;
+        }
+
+        return cache.getIfPresent(new MultiKey(skuIds));
+    }
+
+    /**
+     * Inserts {@link AnonymousCertContent} into the cache for the provided top level SKU IDs. An existing
+     * entry in the cache for the same SKU IDs will be replaced.
+     *
+     * @param skuIds
+     *  the top level SKU IDs to associate the consumer certificate content to in the cache
+     *
+     * @param content
+     *  the consumer certificate content to insert into the cache
+     *
+     * @throws IllegalArgumentException
+     *  if the provided SKU IDs are null or the provided consumer certificate content is null
+     */
+    public void put(Collection<String> skuIds, AnonymousCertContent content) {
+        if (skuIds == null) {
+            throw new IllegalArgumentException("sku ID is null");
+        }
+
+        if (content == null) {
+            throw new IllegalArgumentException("content is null");
+        }
+
+        cache.put(new MultiKey(skuIds), content);
+    }
+
+    /**
+     * Removes the cached {@link AnonymousCertContent} associated to the provided top level SKU IDs.
+     *
+     * @param skuIds
+     *  the top level SKU IDs to remove cached consumer certificate content for
+     *
+     * @throws IllegalArgumentException
+     *  if the provided SKU IDs are null
+     */
+    public void remove(Collection<String> skuIds) {
+        if (skuIds == null) {
+            throw new IllegalArgumentException("sku IDs is null");
+        }
+
+        cache.invalidate(new MultiKey(skuIds));
+    }
+
+    /**
+     * Clears all entries in the cache
+     */
+    public void removeAll() {
+        cache.invalidateAll();
+    }
+
+}
--- a/src/main/java/org/candlepin/cache/MultiKey.java
+++ b/src/main/java/org/candlepin/cache/MultiKey.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.cache;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Objects;
+
+/*
+ * A key composed of multiple elements. Duplicate elements are removed to form a distinct set including null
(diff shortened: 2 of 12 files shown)


src/main/java/org/candlepin/cache/ConsumerCertificateContentCache.java
```diff
@@ -0,0 +1,134 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * A thread safe cache for consumer certificate content. Entries are evicted based on a time-to-live eviction
+ * policy.
+ */
+@Singleton
+public class ConsumerCertificateContentCache {
+    private Cache<Set<String>, ConsumerCertContent> cache;
```
Januson, line 39: We should consider more specialized cache implementation like Caffeine. It [offers better performance](https://github.com/ben-manes/caffeine/wiki/Benchmarks) and is based on lessons learned from Guava's implementation.

joshmalbrecht: Oh wow! The performance difference is pretty incredible when comparing to Guava.  I would be on board with using Caffeine!

src/test/java/org/candlepin/cache/ConsumerCertificateContentCacheTest.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.Random;
+
+@ExtendWith(MockitoExtension.class)
+public class ConsumerCertificateContentCacheTest {
+
+    // In milliseconds
+    private static final long DEFAULT_CACHE_TTL = 2000L;
+
+    @Mock
+    private Configuration mockConfiguration;
```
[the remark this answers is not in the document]
joshmalbrecht: Updated to use DevConfig

src/main/java/org/candlepin/cache/ConsumerCertificateContentCache.java
joshmalbrecht: Updated to use Caffeine

src/main/java/org/candlepin/cache/ConsumerCertContent.java
```diff
@@ -0,0 +1,60 @@
...
+
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("ConsumerCertContent [content=%s]", content.toString());
+    }
+
+}
```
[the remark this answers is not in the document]
Januson: Can you please add this as a comment? It will be helpful when someone revisits the code.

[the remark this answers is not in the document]
joshmalbrecht: Good idea. Added a comment

src/main/java/org/candlepin/cache/ConsumerCertificateContentCache.java
```diff
@@ -0,0 +1,134 @@
...
```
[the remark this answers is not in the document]
nikosmoum: more specifically, HashSet makes no guarantees on the order of items. Could there be a problem if someone is searching in the cache using a set/collection that has RH001,RH002, while the cache has a set with items RH002,RH002?

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -15,6 +15,8 @@
 package org.candlepin.controller;
 
 import org.candlepin.audit.EventSink;
+import org.candlepin.cache.ConsumerCertContent;
+import org.candlepin.cache.ConsumerCertificateContentCache;
```
nikosmoum, line 19: For the same reason as the config property name: the 'consumer' part of the name is redundant. can we make it more clear these are for anonymous certificate content? something like `AnonymousCertContent` and `AnonymousCertContentCache`?

nikosmoum (changes requested):

src/main/java/org/candlepin/controller/ContentAccessManager.java
joshmalbrecht: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -870,49 +875,84 @@ public AnonymousContentAccessCertificate getCertificate(AnonymousCloudConsumer c
...
-
-        PromotedContent promotedContent = new PromotedContent(new AnonymousContentPrefix());
-        byte[] payloadBytes = createContentAccessDataPayload(null, activeContent, promotedContent);
+        certificateContent.forEach(content -> activeContent.put(content, true));
 
         AnonymousContentAccessCertificate caCert = new AnonymousContentAccessCertificate();
         caCert.setSerial(serial);
         caCert.setKeyAsBytes(pemEncodedKeyPair);
-        caCert.setCert(x509Cert + createPayloadAndSignature(payloadBytes));
+        caCert.setCert(x509Cert + createPayloadAndSignature(contentAccessPayload));
```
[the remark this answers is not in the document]
joshmalbrecht: Updated to cache the payload and signature string

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -870,49 +875,84 @@ public AnonymousContentAccessCertificate getCertificate(AnonymousCloudConsumer c
...
+            List<ProductInfo> products = prodAdapter.getChildrenByProductIds(consumer.getProductIds());
+            if (products == null || products.isEmpty()) {
+                String msg = "Unable to retrieve products for anonymous cloud consumer: " +
+                    consumer.getUuid();
+                throw new RuntimeException(msg);
+            }
+
+            payload = createContentAccessDataPayloadFromProdInfo(products);
+            List<ContentInfo> contentInfo = getContentInfo(products);
+            content = convertContentInfoToContent(contentInfo);
```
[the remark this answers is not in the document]
joshmalbrecht: Updated to cache the contentDTO

src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -178,6 +178,7 @@ private ConfigProperties() {
     // Cache
     public static final String CACHE_JMX_STATS = "cache.jmx.statistics";
     public static final String CACHE_CONFIG_FILE_URI = JPA_CONFIG_PREFIX + "hibernate.javax.cache.uri";
+    public static final String CACHE_SKU_CONTENT_TTL = "candlepin.cache.sku_content_ttl";
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

src/main/java/org/candlepin/cache/ConsumerCertificateContentCache.java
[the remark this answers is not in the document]
Ceiu: That's assuming the equality method of the underlying set(s) in question are properly implemented or don't reject others outright. If these sets are coming from Hibernate entities, for instance, they won't have valid equals or hashcode implementations.

This is definitely doable, but we'll need to be very careful about what kinds of sets are going into this cache object as a key. Also, I'd be interested to know how the lookup and storing of the cached value works in the case where the set is changed after it is cached. Mutable keys are generally bad for consistency and can lead to very strange bugs if that behavior isn't defined or handled.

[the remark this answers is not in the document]
joshmalbrecht: In the `AnonymousCertContentCache.put(Collection<String> skuIds, AnonymousCertContent content)` we are taking in a collection of Strings, but then we are copying them to a HashSet on the following line. So, the key will always be a HashSet and shouldn't mutate, but we could make sure by using `Collections.unmodifiableSet()` when we insert into the cache!

https://github.com/candlepin/candlepin/blob/jalbrech/m/CANDLEPIN-733/src/main/java/org/candlepin/cache/AnonymousCertContentCache.java#L103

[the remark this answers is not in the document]
Ceiu: If that's the case, would it make sense to convert it to something immutable and more universally stable? As an example, we could process it as a stream that filters nulls and hashes the inputs as it goes and use that for our key.

(Also, if it's not clear, I've not read any of the code for this PR -- I'm only following the conversations as they hit my inbox.)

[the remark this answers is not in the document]
joshmalbrecht: Yeah, I think we could definitely do that to generate a key based on the provided SKU Ids!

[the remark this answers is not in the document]
joshmalbrecht: In a new commit, I added the changes to filter the provided collections in a stream while removing nulls and duplicates, and then creating a hash value from the strings to be used as a key. Let me know what you think!

src/main/java/org/candlepin/cache/AnonymousCertContentCache.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Clears all entries in the cache
+     */
+    public void removeAll() {
+        cache.invalidateAll();
+    }
+
+    private int getKey(Collection<String> skuIds) {
```
Januson, line 128: If we are going to with custom keys we should create a class for it to make it explicit. E.G. MultiKey which could define constructors for supported inputs and explicitly define a hashCode.
 
Btw. You could get the same result with new HashSet<>(skuIds).hashCode();

src/main/java/org/candlepin/cache/MultiKey.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package org.candlepin.cache;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Objects;
+
+/*
+ * A key composed of multiple distinct elements.
+ */
```
Januson, line 23: Please add some javadoc to document the assumptions we made about order, deduplication, nulls, etc.

src/main/java/org/candlepin/cache/MultiKey.java
```diff
@@ -0,0 +1,64 @@
...
+public class MultiKey {
+
+    private HashSet<?> elements;
```
Januson, line 26: You can just use string. If we ever need more, we can find a way to generalize.

src/main/java/org/candlepin/cache/AnonymousCertContentCache.java
nikosmoum: This MultiKey class is just a wrapper for a HashSet. What are we getting out of it? Wouldn't an unmodifiableSet be more straightforward as previously suggested?

src/main/java/org/candlepin/cache/MultiKey.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public MultiKey(Collection<String> elements) {
+        Objects.requireNonNull(elements);
+
+        this.elements = new HashSet<>(elements);
+    }
+
+    public HashSet<?> getElements() {
```
Januson, line 34: We shouldn't expose the elements.

src/main/java/org/candlepin/cache/AnonymousCertContentCache.java
Januson: We get an explicit place where the rules for our cache key are defined. We could make the wrapped set unmodifiable, but unless we expose the reference we don't have to.

src/main/java/org/candlepin/cache/MultiKey.java
joshmalbrecht: Added

src/main/java/org/candlepin/cache/MultiKey.java
joshmalbrecht: Updated

src/main/java/org/candlepin/cache/MultiKey.java
joshmalbrecht: Removed

Januson (approved):

src/main/java/org/candlepin/cache/AnonymousCertContentCache.java
nikosmoum: alright

nikosmoum (approved):

PR #4467, opened 2024-01-04
CANDLEPIN-720: Implement Anonymous Owner Claim endpoint

Januson:
- Added a new claim endpoint
- Added a new ConsumerMigrationJob


src/main/java/org/candlepin/async/tasks/ConsumerMigration.java
```diff
@@ -0,0 +1,100 @@
...
+     * Migrates all consumers of the given anonymous org to the target org.
+     *
+     * @param originOwnerKey key of the source anonymous org
+     * @param destinationOwnerKey key of the destination org
+     */
+    public void migrate(String originOwnerKey, String destinationOwnerKey) {
+        Owner originOwner = this.ownerCurator.getByKey(originOwnerKey);
+        Owner destinationOwner = this.ownerCurator.getByKey(destinationOwnerKey);
+
+        List<String> consumerIds = this.ownerCurator.getConsumerIds(originOwner).list();
```
nikosmoum, line 71: This curator method can be changed to return a Collection directly instead of a CandlepinQuery. The only other instance where it is used(OwnerManager), isn't used for paging.

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1648,28 +1613,37 @@ public CandlepinQuery<ConsumerDTOArrayElement> getHypervisors(
...
+        if (Boolean.TRUE.equals(destinationOwner.getAnonymous())) {
+            throw new IllegalArgumentException("Claimant owner cannot be anonymous!");
         }
 
-        return owner;
+        JobConfig jobConfig = ConsumerMigrationJob.createConfig()
+            .setOriginOwner(anonymousOwnerKey)
+            .setDestinationOwner(claimantOwner.getClaimantOwnerKey());
+
+        return queueJob(jobConfig);
```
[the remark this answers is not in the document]
nikosmoum: Also, that means we need a `@Transactional` annotation on the endpoint method

src/main/java/org/candlepin/async/tasks/ConsumerMigration.java
```diff
@@ -0,0 +1,100 @@
...
+        for (List<String> consumerIdsBlock : Iterables.partition(consumerIds, this.batchSize)) {
+            try {
+                this.ownerCurator.transactional(args -> {
+                    migrateBatch(consumerIdsBlock, destinationOwner);
+                    return null;
+                }).execute();
+            }
+            catch (Exception e) {
+                log.warn("Failed to migrate a batch of consumers.", e);
+                // Retry later
```
nikosmoum, line 81: I think if at least one transaction fails, we want to fail the job so that it is retried later

src/main/java/org/candlepin/async/tasks/ConsumerMigration.java
```diff
@@ -0,0 +1,100 @@
...
+    private void migrateBatch(List<String> consumerIds, Owner destination) {
+        this.consumerCurator.lockAndLoad(consumerIds);
+
+        List<String> idCerts = this.consumerCurator.getIdentityCertIds(consumerIds);
+        List<String> caCerts = this.consumerCurator.getContentAccessCertIds(consumerIds);
+        this.consumerCurator.unlinkIdCertificates(idCerts);
+        this.consumerCurator.unlinkCaCertificates(caCerts);
+
+        this.identityCertificateCurator.deleteByIds(idCerts);
+        this.contentAccessCertificateCurator.bulkDelete(caCerts);
```
nikosmoum, line 95: I think we can do it in a similar way to the CertificateCLeanupJob here:
use the existing `contentAccessCertificateCurator.deleteByIds` method, then `serialCurator.revokeByIds`, without a need for the new `bulkDelete` method.
And same thing for the identity certs

nikosmoum (changes requested):
I know this is still in progress, but just adding a few comments early

src/main/java/org/candlepin/async/tasks/ConsumerMigration.java
nikosmoum: clarification after discussion: we should still try to work through all the batches, but if any of them fail, we should mark it with a flag, so we fail the job at the very end

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
```diff
@@ -792,4 +795,41 @@ public void shouldAllowUserWithOwnerPoolsPermissionCanSeeSystemPurposeOfTheOwner
         assertEquals(owner.getKey(), attributes.getOwner().getKey());
     }
 
+    @Test
+    @OnlyInHosted
+    public void shouldMigrateAllConsumersOfAnonymousOwnerToTargetOwner() {
+        String accountId = StringUtil.random("cloud-account-id-");
+        String instanceId = StringUtil.random("cloud-instance-id-");
+        String offerId = StringUtil.random("cloud-offer-");
+        OwnerDTO anonOwner = admin.owners().createOwner(Owners.random().anonymous(true));
```
nikosmoum, line 804: instead of `admin.owners()` you can use the existing `owners` field here, like the line right below this one

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
```diff
@@ -792,4 +795,41 @@ public void shouldAllowUserWithOwnerPoolsPermissionCanSeeSystemPurposeOfTheOwner
...
+        CloudAuthenticationResultDTO result = ApiClients.noAuth().cloudAuthorization()
+            .cloudAuthorizeV2(accountId, instanceId, offerId, "test-type", "");
+
+        ApiClient anonClient = ApiClients.bearerToken(result.getToken());
+        anonClient.consumers().createConsumer(Consumers.random(anonOwner));
+        anonClient.consumers().createConsumer(Consumers.random(anonOwner));
+        anonClient.consumers().createConsumer(Consumers.random(anonOwner));
+
+        AsyncJobStatusDTO job = owners.claim(anonOwner.getKey(),
+            new ClaimantOwner().claimantOwnerKey(destOwner.getKey()));
```
nikosmoum, line 823: once the remaining work mentioned in other comments is done, we should assert after this call that the anonOwner has claimed==true, and claimant_owner==destOwner.key

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
```diff
@@ -792,4 +795,41 @@ public void shouldAllowUserWithOwnerPoolsPermissionCanSeeSystemPurposeOfTheOwner
...
+    public void shouldMigrateAllConsumersOfAnonymousOwnerToTargetOwner() {
+        String accountId = StringUtil.random("cloud-account-id-");
+        String instanceId = StringUtil.random("cloud-instance-id-");
+        String offerId = StringUtil.random("cloud-offer-");
+        OwnerDTO anonOwner = admin.owners().createOwner(Owners.random().anonymous(true));
+        OwnerDTO destOwner = owners.createOwner(Owners.random());
+        admin.hosted().createOwner(anonOwner);
+        ProductDTO prod = admin.ownerProducts().createProduct(anonOwner.getKey(), Products.random());
+        admin.hosted().createProduct(prod);
+        admin.owners().createPool(anonOwner.getKey(), Pools.random(prod));
```
nikosmoum, line 809: same here: use the existing `owners` field, or remove it from the setup method and use `admin.owners()` everywhere

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
```diff
@@ -792,4 +795,41 @@ public void shouldAllowUserWithOwnerPoolsPermissionCanSeeSystemPurposeOfTheOwner
...
+            new ClaimantOwner().claimantOwnerKey(destOwner.getKey()));
+
+        job = admin.jobs().waitForJob(job);
+        assertThatJob(job).isFinished();
+
+        List<ConsumerDTOArrayElement> anonConsumers = owners.listOwnerConsumers(anonOwner.getKey());
+        assertThat(anonConsumers).isEmpty();
+
+        List<ConsumerDTOArrayElement> migratedConsumers = owners.listOwnerConsumers(destOwner.getKey());
+        assertThat(migratedConsumers).hasSize(3);
```
nikosmoum, line 832: we could also assert on one of those consumers, to check that they get new identity cert (by comparing with their old identity cert) when we call `ConsumerResource.getConsumer()`, and do the same for their SCA cert

src/main/java/org/candlepin/controller/ConsumerMigration.java
```diff
@@ -0,0 +1,115 @@
...
+                }).execute();
+            }
+            catch (Exception e) {
+                log.warn("Failed to migrate a batch of consumers.", e);
+                failed = true;
+            }
+        }
+
+        if (failed) {
+            throw new ConsumerMigrationFailedException("Failed to migrate consumers");
```
nikosmoum, line 92: This will need to get translated into a `JobExecutionException` at the job level (or, if this method is merged into the job class, you would throw a `JobExecutionException` directly), with terminal=false, so the intent is clear for such failures

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
```diff
@@ -792,4 +797,75 @@ public void shouldAllowUserWithOwnerPoolsPermissionCanSeeSystemPurposeOfTheOwner
...
+        // Anon owner should be claimed
+        OwnerDTO updatedOwner = owners.getOwner(anonOwner.getKey());
+        assertThat(updatedOwner)
+            .returns(true, OwnerDTO::getClaimed)
+            .returns(destOwner.getKey(), OwnerDTO::getClaimantOwner);
+
+        // Consumers should have regenerated identity certificates
+        List<ConsumerDTO> updatedConsumers = fetchConsumers(consumers, admin);
+        List<CertificateDTO> newIdCerts = updatedConsumers.stream().map(ConsumerDTO::getIdCert).toList();
+        assertThat(newIdCerts).doesNotContainAnyElementsOf(oldIdCerts);
```
nikosmoum, line 851: this kind of assertion will wrongly pass if newIdCerts was empty in the first place (for whatever reason). can we make sure both lists actually contain 3 certs?

nikosmoum: and same thing for sca certs

src/main/java/org/candlepin/resource/OwnerResource.java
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/async/tasks/ConsumerMigration.java
Januson: Done. I handled the id gathering a bit differently.

src/main/java/org/candlepin/async/tasks/ConsumerMigration.java
Januson: Done.

src/main/java/org/candlepin/async/tasks/ConsumerMigration.java
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
Januson: Done.

src/main/java/org/candlepin/async/tasks/ConsumerMigration.java
nikosmoum: I still see the same code here: getConsumerIds returns a CandlepinQuery and then it unpacks it with `.list()`

src/main/java/org/candlepin/async/tasks/ConsumerMigrationJob.java
```diff
@@ -0,0 +1,136 @@
...
+        this.consumerMigration = Objects.requireNonNull(consumerMigration);
+    }
+
+    @Override
+    public void execute(JobExecutionContext context) throws JobExecutionException {
+        String originOwnerKey = context.getJobArguments().getAsString(ORIGIN_OWNER);
+        String destinationOwnerKey = context.getJobArguments().getAsString(DESTINATION_OWNER);
+        log.info("Starting consumer migration from: {} to: {}", originOwnerKey, destinationOwnerKey);
+
+        this.consumerMigration.migrate(originOwnerKey, destinationOwnerKey);
```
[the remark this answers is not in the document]
Januson: I split it because we shouldn't be coupling business logic with our job infrastructure. It is also easier for testing.

src/main/java/org/candlepin/controller/ConsumerMigration.java
```diff
@@ -0,0 +1,118 @@
...
+            }
+        }
+
+        if (failed) {
+            throw new ConsumerMigrationFailedException(i18n.tr("Failed to migrate consumers"));
+        }
+    }
+
+    private void migrateBatch(List<String> consumerIds, Owner destination) {
+        this.consumerCurator.lockAndLoad(consumerIds);
```
nikosmoum, line 100: Something I missed here is that this method locks the consumers **and loads the entire entities** into memory. which is something I was hoping to avoid by doing all the id-based queries below. I can't currently find in our curators an implementation for locking by ids, and only fetching the ids (instead of the entire entities), but I think its worth investigating if we can do it (it should be possible). E.g. by adding a method like `lockAndLoadIds`. If it is not possible to do this, it may not necessarily be bad if the batch size is kept small enough

src/main/java/org/candlepin/controller/ConsumerMigration.java
nikosmoum: hmmm I also see the same code for this: the job `execute()` method isn't translating this exception into a JobExecutionException

nikosmoum (changes requested):

src/main/resources/db/changelog/202401081559-add-claimant-owner-column.xml
```diff
@@ -0,0 +1,30 @@
...
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <addForeignKeyConstraint baseTableName="cp_owner"
+            baseColumnNames="claimant_owner"
+            referencedTableName="cp_owner"
+            referencedColumnNames="account"
+            constraintName="fk_cp_owner_claimant_owner"
+            onDelete="CASCADE" />
```
joshmalbrecht, line 27: Are the columns in this FK constraint backwards? If we delete the anonymous owner that has been claimed and has the `claimant_owner` value populated, won't this delete the claimant owner as well? That would be problematic because the claimant owner is the owner the consumers should be registered to. If the claimant owner gets deleted, we should cascade and delete the anonymous owners that were claimed by that owner.

@nikosmoum just want a double check that I'm right on this

joshmalbrecht (changes requested):
Looks good! Just one thing I noticed with the FK constraint

src/main/resources/db/changelog/202401081559-add-claimant-owner-column.xml
Januson: This is correct. When referenced row gets deleted, base will get deleted as well, but I double checked to be sure.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1786,4 +1791,61 @@ public int unlinkCaCertificates(Collection<String> certIds) {
...
+            .createQuery(hql, String.class)
+            .setLockOptions(new LockOptions(LockMode.PESSIMISTIC_WRITE));
+
+        List<String> lockedIds = new ArrayList<>(idSet.size());
+        for (List<Serializable> idBlock : this.partition(idSet)) {
+            lockedIds.addAll(query.setParameter("ids", idBlock).getResultList());
+        }
+
+        return lockedIds;
+    }
```
Januson, line 1849: This seems to be working fine. Tested by running 20 jobs with a same set of consumers and locking for 10 seconds. Attempts to lock it with `byMultipleIds()` and entity graphs were unsuccessful. I couldn't figure out how to limit the amount of fetched data.

src/main/resources/db/changelog/202401081559-add-claimant-owner-column.xml
nikosmoum: Yes, the current code is correct. If you delete the destination owner, both get deleted. If you delete the anonymous owner, the destination owner isn't getting deleted

src/main/java/org/candlepin/model/ConsumerCurator.java
nikosmoum: It looks good to me, but I'd also ask @Ceiu who is more familiar with hibernate

src/main/resources/db/changelog/202401081559-add-claimant-owner-column.xml
joshmalbrecht: Thanks for double checking that guys!

src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: Not getting locks via `byMultipleIds` is very disconcerting to hear, since we have other locking calls and critical parts of the locking model that rely on it. But I'm also not entirely surprised to hear that it didn't let you modify the graph.

That aside, the only other ways of locking I know of via Hibernate is either this or a direct SQL query with `FOR UPDATE` (which the above code should be triggering).

nikosmoum: You can use these config properties to show the SQL generated: https://www.candlepinproject.org/docs/candlepin/logging.html#logging-sql-queries-and-parameters to see if the `FOR UPDATE` is there

Januson: Yes. It is triggering FOR UPDATE. This is the generated sql.

```sql
select consumer0_.id as col_0_0_ from cp_consumer consumer0_ where consumer0_.id in (? , ?) for update
```

src/main/java/org/candlepin/controller/ConsumerMigration.java
nikosmoum: @Januson Once this one is done, I can merge the PR

Januson: Done.

PR #4469, opened 2024-01-05
[M] 2256950: Fix CertificateCleanupJob fk violation (CANDLEPIN-741)

nikosmoum:
- An old db changelog (20170301083925) accidentally marked all serials for cp_certificate certs as revoked, even though those are not supposed to be revocable. This change reverts the bad data by unrevoking those serials.
- The CertificateCleanupJob now will only delete serials that are revoked and expired, if and only if they are not referenced in any certificate table, in order to avoid fk violations due to bad data such as the ones previously mentioned.
- Remove unused API endpoint GET /products/product_uuid/certificate.
- Stop exporting product certificates in manifests. This has never been used (never implemented in the Adapters even).


src/main/java/org/candlepin/model/CertificateSerialCurator.java
```diff
@@ -52,17 +55,53 @@ private Date getExpiryRestriction() {
...
+            .getResultList();
+
+        String deleteSQL = "DELETE FROM cp_cert_serial WHERE id IN (:serials_to_delete)";
+        nativeSpaces = List.of(CertificateSerial.class.getName());
+        int deleted = 0;
+        for (List<Long> serialsToDeleteBlock : this.partition(serials)) {
+            deleted += entityManager.createNativeQuery(deleteSQL)
+                .setHint(QueryHints.NATIVE_SPACES, nativeSpaces)
+                .setParameter("serials_to_delete", serialsToDeleteBlock)
+                .executeUpdate();
```
nikosmoum, line 102: @Ceiu Is this what you had in mind in trying to avoid any kind of fk violation when the job runs? I am a bit conflicted with the potential performance hit here from subqeries and splitting into 2 queries, if it is really worth doing all this for avoiding any fk violation from rogue bad data.

Ceiu: There's a lot to consider here:

- When it comes to certificates, the model is extremely silly. Having everything broken up the way it is comes down to a Hibernate-ism where the ownership of the joins is on what I would argue is the wrong side of the relationship. For instance certificates don't have entitlements, entitlements have certificates. As a direct result of this, doing any kind of cleanup or sanitation gets clunky and tedious. Worse, because the *cert* tables differ but all link back to a single table for the serials, we get into this situation where we have some subset of the certs which aren't revocable in terms of logic, but *can* be flagged as revoked in the data. Because of this desync possibility, IMO our logic needs to be solid enough that valid data in the database does not cause issues.
- It may or may not be more efficient to left join on the other tables instead of a correlated subquery. Given how simple those subqueries are, and how it basically looks like a join anyway, I'm betting the planner for PostgreSQL and MariaDB will end up doing nearly the same thing anyway. Worth looking into if performance is a concern with this.
- In terms of performance, how frequently is this going to be run? If this is part of the cleanup job that runs once/day at most, accuracy is more important to me than raw performance so long as that performance hit doesn't have knock-on effects elsewhere, and doesn't turn it into bubble-sort levels of bad. If it runs in under 60s as a periodic background task and isn't causing resource contention issues elsewhere, I'm not too concerned about it.
- In terms of certificates and how the model works logically, do we have any cases where we want to keep a cert around when we're deleting its serial? Putting aside the issue of non-revocable certs for a moment, it really feels like our model is backward here (again) -- we manage serials and certs are optional as far as our logic is concerned, so why is cert the parent and serial the child in this case? It prevents us from doing things like delete-cascades or any other method that would be less hands-on in terms of juggling both the cert and the serial. As it is now, we treat them both as related first-class objects rather than the parent-child relationship it actually is. Granted, this is starting to lean on a larger refactor a bit, but if we're starting to even consider shortcuts in terms of how we manage this stuff because it's either too slow or too tedious to do it right, maybe it's time we face this problem head-on.

Ceiu (approved):
This addresses the minimum requirement of preventing serials still referenced by an active cert from attempting to be deleted. Going to leave it for a moment for others to review or for some contemplation on how we want to deal with the cert problem in general.

src/main/java/org/candlepin/model/CertificateSerialCurator.java
nikosmoum: I used correlated subqueries because I initially was trying to do this in a single DELETE query, and Mariadb/Postgres have different syntaxes on how to do deletes with joins. But since I turned it into a SELECT I can use ANSI SQL LEFT joins now.

Performance wise: I run the correlated-subquery version in Prod and it took 2.5 minutes (that is, just for the SELECT, not including the subsequent DELETE with a large IN clause), while the LEFT JOIN version of the query takes 2-3 seconds. So we should be good here with this.

nikosmoum: Regarding the model situation (not something to deal with in this PR):

Logically, we never want to keep a cert after we deleted the serial.
I suspect the current model is the way it is, because the serial needs to be kept around for CRL purposes. 

I was looking into if we could remove the serial entirely from the subscription certificate entity (this is what the product certificates do, they don't actually have a serial), but it's probably not a good idea, because while subscription certificates are not supposed to be revocable, they still likely need the expiration data (which is part of the serial entity). I think satellite wants it (either they're using it, or they wanted to use it to warn users their manifests are about to expire).

So distilling the requirements down:
- We never want to keep a cert after we deleted the serial.
- We want some certs to be revocable, and some not.
- We want all certs to be expirable.
- For revocable certs, we need to track a state change from unrevoked to revoked.

If I would do this from scratch, this would be two separate entities and both of them contain all the fields of the current 2 entities (cert+serial). One of them would be mapped to cp_revocable_certificate, and the other to a cp_nonrevocable_certificate.

Entities that do not care about revocation (SubscriptionCertificate, ProductCertificate) would use the later, while all the others the former.

For the revocable ones, we wouldn't actually delete them as we do today. We would do a soft-delete, by simply changing the revoked field from false to true (the con is that this would use more space, by keeping all these revoked certs).

The CertificateCleanupJob would then just delete from cp_nonrevocable_certificate when they expire, and from the cp_revocable_certificate when they're both expired and revoked.

PR #4470, opened 2024-01-05
[4.3] Fix CertificateCleanupJob fk violation & add back indexes (CANDLEPIN-741)

nikosmoum:
This PR adds back the previously reverted commits for improving the speed of CertificateCleanupJob (9fef94d5f625ae24f530cd0caadae1dc5da3dbec and 7d67a8b1dff719ff12475db63af03c4911a42cc1),
while also fixes the foreign key violation against the cp_certificate table.


nikosmoum:
For any reviewers: please disregard the bugzilla check failure. The 2 commits that reference 2231503 are failing because when they were initially commited, the main branch was still in version 4.3 (while now main is 4.4).

Ceiu (approved):

PR #4473, opened 2024-01-08
CANDLEPIN-739: Catch scenario with null event data

wottop:
- Will bypass email sending if event data is null or a subscription id is not in the data

### DIFF
--- a/src/main/java/org/candlepin/audit/ActivationListener.java
+++ b/src/main/java/org/candlepin/audit/ActivationListener.java
@@ -16,6 +16,7 @@ package org.candlepin.audit;
 
 import org.candlepin.service.SubscriptionServiceAdapter;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/audit/ActivationListener.java
```diff
@@ -45,10 +46,15 @@ public ActivationListener(SubscriptionServiceAdapter subService,
...
             event.getTarget().equals(Event.Target.POOL)) {
+            JsonNode subscriptionId = null;
             try {
-                String subscriptionId = mapper.readTree(event.getEventData())
-                    .get("subscriptionId").asText();
-                subscriptionService.sendActivationEmail(subscriptionId);
+                if (event.getEventData() != null) {
+                    subscriptionId = mapper.readTree(event.getEventData())
+                        .get("subscriptionId");
+                }
```
Ceiu, line 54: Add some whitespace between these two blocks so it doesn't look like an if/else-if at a glance.

Ceiu (changes requested):
Looks good; I just have a very minor whitespace request.

Ceiu (approved):

PR #4480, opened 2024-01-11
CANDLEPIN-754: Create ClaimedOwnerConsumerDetectionJob

Januson:


nikosmoum (changes requested):
After some discussion with @joshmalbrecht, it turned out that the detection job is not really needed at all. Any late consumers trying to register after the main ConsumerMigrationJob has run for that anonymous owner, can just be redirected by checking the owner.claimant_owner field:

We can add a check to ConsumerResource.setupOwner that checks if the owner that is set on the anonymous cloud consumer has claimed=true and has a claimant_owner value and if it does, we register them directly to the claimant_owner (instead of the anonymous owner).

PR #4484, opened 2024-01-12
[M] CANDLEPIN-770: Added max size for AnonymousCertContentCache

joshmalbrecht:
- Added configuration for adding max number of entries for AnonymousCertContentCache
- Added default value for new max size configuration


src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -179,6 +179,8 @@ private ConfigProperties() {
     public static final String CACHE_JMX_STATS = "cache.jmx.statistics";
     public static final String CACHE_CONFIG_FILE_URI = JPA_CONFIG_PREFIX + "hibernate.javax.cache.uri";
     public static final String CACHE_ANON_CERT_CONTENT_TTL = "candlepin.cache.anon_cert_content_ttl";
+    public static final String CACHE_ANON_CERT_CONTENT_MAX_ENTRIES =
+        "candlepin.cache.anon_cert_content_max_entries";
```
nikosmoum, line 183: since we already have 2 of those, we can put them under an even more common namespace, and reduce the underscores. E.g.:
```
candlepin.cache.anonymous.cert.content.ttl
candlepin.cache.anonymous.cert.content.max_entries
```

nikosmoum (changes requested):
Just a minor suggestion and this should be good

src/main/java/org/candlepin/config/ConfigProperties.java
joshmalbrecht: That makes sense, I like that a lot more. Updated

nikosmoum (approved):

PR #4488, opened 2024-01-16
CANDLEPIN-754: Register late anonymous consumers to the claimant org

Januson:


spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
```diff
@@ -587,8 +587,54 @@ public void shouldCreateConsumerWhenUsingAnonymousToken() throws JsonProcessingE
...
+        admin.hosted().createOwner(anonOwner);
+        admin.hosted().associateOwnerToCloudAccount(accountId, anonOwner.getKey());
+        admin.hosted().createSubscription(Subscriptions.random(anonOwner, productDTO));
+        admin.owners().createOwner(anonOwner);
+        admin.ownerProducts().createProduct(anonOwner.getKey(), productDTO);
+        admin.owners().createPool(anonOwner.getKey(), Pools.random(productDTO));
+
+        CloudAuthenticationResultDTO result = ApiClients.noAuth().cloudAuthorization()
+            .cloudAuthorizeV2(accountId, instanceId, offerId, "test-type", "");
+        ApiClient anonClient = ApiClients.bearerToken(result.getToken());
```
nikosmoum, line 615: Because the org was fully ready before you called the cloudAuthorizeV2, the token you're getting here isn't anonymous, its the standard one. Its good we are testing this, but we also want to test that clients that had an anonymous token can also register after the claiming is done.
You can add another `cloudAuthorizeV2` call right before the `admin.owners().createPool()` call, which should return an anonymous token, and test registering with that too after the claiming

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
```diff
@@ -102,18 +103,27 @@ void runFilter(ContainerRequestContext requestContext) {
...
-            String ownerKey = checkCloudAccountOrganization(anonymousCloudConsumer);
-            if (!ownerCurator.existsByKey(ownerKey)) {
-                throw new TooManyRequestsException(ORG_NOT_CREATED_IN_CANDLEPIN);
-            }
-            if (!poolCurator.hasPoolsForProducts(ownerKey, anonymousCloudConsumer.getProductIds())) {
-                throw new TooManyRequestsException(ORG_DOES_NOT_HAVE_POOLS);
+            validateAnonymousCloudConsumer(anonymPrincipal);
+        }
+        if (principal instanceof CloudConsumerPrincipal) {
+            if (!securityHole.anonConsumer()) {
```
nikosmoum, line 109: The name 'anonConsumer' in SecurityHole.java is not quite accurate now, since it is used to determine if either a standard autoreg token or an anonymous autoreg token user is authorized. Can we rename it into something like `autoregToken`?

nikosmoum (changes requested):
Only a few minor things needed

nikosmoum:
Also, test coverage

src/main/java/org/candlepin/auth/CloudRegistrationAuth.java
```diff
@@ -149,33 +145,20 @@ public Principal getPrincipal(HttpRequest httpRequest) {
...
      * @param ownerKey
-     *     the key of an organization in which the principal will have authorization to register clients
-     *
-     * @return a minimal UserPrincipal representing the cloud registration token
+     *  the key of an organization in which the principal will have authorization to register clients
+     * @return
+     *  a minimal {@link Principal} representing the cloud registration token
      */
-    private UserPrincipal createCloudUserPrincipal(String subject, String ownerKey) {
+    private CloudConsumerPrincipal createCloudUserPrincipal(String ownerKey) {
```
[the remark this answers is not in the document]
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
Januson: Done.

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
Januson: Done.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
nikosmoum: Now it is creating an anonClient, but it still needs to use it to register a consumer after the claiming job is finished

Januson: Done.

PR #4504, opened 2024-01-19
CANDLEPIN-774: HostedTestCloudRegistrationAdapter should decode metadata as base64

wottop:


src/main/java/org/candlepin/testext/hostedtest/HostedTestCloudRegistrationAdapter.java
```diff
@@ -32,6 +32,8 @@
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
+import org.apache.commons.codec.binary.Base64;
```
Januson, line 35: Since Java 8 the java.util.Base64 should be preferred over the one from apache.commons.

Januson (changes requested):
Changes look good. Just a small comment regarding the used Base64 implementation.

Januson (approved):

PR #4505, opened 2024-01-22
CANDLEPIN-718: Turn BC provider into injectable dependency

Januson:


Ceiu (changes requested):
nack on this

While the change is conceptually good, the reality is that the security provider implementations are not all created equal. 

Some history here:
One of the bugs we've run into prior surrounding the transition to FIPS support is that the JSS functionality did not provide introspection into keys, but the instances returned by the default Java provider did. Our previous implementations using the various `.getInstance` calls did not specify a provider, so we got whatever the JVM happened to return, which just so happened to allow key introspection... because it wasn't a JSS provider.

Once the FIPS flag was flipped and the Java security provider no longer implemented/supported the things we needed, things broke as the behavior we got was entirely different, without any change to the code.

Because the security provider is generic and unknown to the specific PKIUtility instance, the same issue could resurface as soon as that provider changes at all.

Given that this already is an implementation of a generic interface providing services with a specific library, it should also be explicit in using *only* that library, without any indirection or risk of mismatch. At the absolute least, this should be requesting a security provider specifically named or typed BouncyCastle.

Januson:
@Ceiu Updated according to the discussion.

src/main/java/org/candlepin/pki/impl/BouncyCastlePKIUtility.java
```diff
@@ -79,45 +78,50 @@
...
 
     @Inject
-    public BouncyCastlePKIUtility(CertificateReader reader, SubjectKeyIdentifierWriter subjectKeyWriter,
-        Configuration config, KeyPairDataCurator keypairDataCurator) {
+    public BouncyCastlePKIUtility(Provider<BouncyCastleProvider> securityProvider, CertificateReader reader,
+        SubjectKeyIdentifierWriter subjectKeyWriter, Configuration config,
+        KeyPairDataCurator keypairDataCurator) {
         super(reader, subjectKeyWriter, config);
         this.keypairDataCurator = keypairDataCurator;
+        this.securityProvider = securityProvider;
```
Ceiu, line 124: Can we use the opportunity to add the null checks on our dependencies here? I wonder if we have some old tests which currently throw nulls at this.

Ceiu (changes requested):
Much better with the type specification. I've just one last request given we're hitting this ancient area and then we're good to merge.

src/main/java/org/candlepin/pki/impl/BouncyCastlePKIUtility.java
Januson: Done.

Ceiu (approved):

PR #4507, opened 2024-01-22
CANDLEPIN-765: Remove Rules upload and delete

wottop:


src/main/java/org/candlepin/resource/RulesResource.java
```diff
@@ -54,35 +52,6 @@ public RulesResource(RulesCurator rulesCurator,
...
-        }
-        catch (Throwable t) {
-            log.error("Exception in rules upload", t);
-            throw new BadRequestException(
-                i18n.tr("Error decoding the rules. The text should be base 64 encoded"));
-        }
-
-        Rules oldRules = rulesCurator.getRules();
-        rulesCurator.update(rules);
-        sink.emitRulesModified(oldRules, rules);
```
Januson, line 78: I think we should remove supporting bits that are now left unused as well. Such as things from EventSink and RulesCurator.

Januson (changes requested):

src/main/java/org/candlepin/resource/RulesResource.java
wottop: Both are still used by the RulesImporter in the Manifest import

Januson (approved):

PR #4510, opened 2024-01-24
CANDLEPIN-718: Refactor certificate extensions

Januson:
- Added a set of enums for handling OIDs
- Unified cert extensions with an interface


joshmalbrecht:
I think this task also includes splitting up the ContentAccessManager class as it currently generates SCA certificates and anonymous consumer certificates.

"Most importantly, any class that generates certificates (ContentAccessManager, 
DefaultIdentityCertServiceAdapter, etc.) should be re-using the X509CertificateBuilder class, and not doing X509-y things by itself. That includes splitting out ContentAccessManager into 2 classes: one for SCA certificates, and one for anonymous entitlement certificates (introduced as part of autoreg_v2), because currently both are generated from that same class to avoid code duplication."

Januson:
@joshmalbrecht Yes. I am working on that and will post it in one of the next PRs. Creation of these specialized certificate generators requires some supporting refactoring. This is one of them. I unified the extensions behind an interface similar to what graviton is doing and refactored OIDs into a hierarchy of enums.

joshmalbrecht:
> @joshmalbrecht Yes. I am working on that and will post it in one of the next PRs. Creation of these specialized certificate generators requires some supporting refactoring. This is one of them. I unified the extensions behind an interface similar to what graviton is doing and refactored OIDs into a hierarchy of enums.

Okay perfect! Just wanted to make sure that part wasn't missed.

joshmalbrecht (approved):

PR #4511, opened 2024-01-24
[TEST] Introducing OTEL with automatical instrumation to Candlepin

bakajstep:
With the OTEL we can see detailed information about what is happening in the Candlepin. 
**How to start it up:**
1) Be in the root of the project
2) Build war of the Candlepin
```shell
./gradlew war -Ptest_extensions=hostedtest,manifestgen
``` 
3) Optional step. Build image of the candlepin.
```shell
sudo docker compose -f dev-container/otel.docker-compose.yml build --no-cache
``` 
4 Run the enviroment
```shell
sudo docker compose -f dev-container/otel.docker-compose.yml up 
``` 
then we can open Jeager, where we can list the individual traces. 
```
http://localhost:16686/
``` 
Initially we will only see the traces associated with the database migration and then when we call an endpoint we will see the trace associated with that endpoint. To test this, we can run some spec-test and see how a given endpoint works and what queries it performs on the database and how long the different parts took.

### DIFF
--- a/containers/release.Containerfile
+++ b/containers/release.Containerfile
@@ -122,3 +122,32 @@ USER tomcat
 EXPOSE 8080 8443 5432 3306
 
 ENTRYPOINT ["/opt/tomcat/bin/catalina.sh", "run"]
+
+################################# OTEL Image ################################
+
+FROM production as otel
+
+LABEL org.opencontainers.image.title="Candlepin Image with OTEL agent" \
+    org.opencontainers.image.description="Candlepin is an open source subscription & entitlement engine \
+    which is designed to manage software subscriptions from both vendor's & customer's perspectives. \
+    This is a development image not intended for production use."
+
+USER root
+
+# Setup development certificate and key
+WORKDIR /etc/candlepin/certs
+COPY --from=builder /app/certs /etc/candlepin/certs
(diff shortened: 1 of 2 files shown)


Ceiu:
Can you provide an example, screenshot, or something showing what this will do that we don't get from existing profiling tools like yourkit's or jprofiler's jdbc introspection?

bakajstep:
> Can you provide an example, screenshot, or something showing what this will do that we don't get from existing profiling tools like yourkit's or jprofiler's jdbc introspection?

I never work with these two profiling tool what you mentioned and i did not know that we are using them. But i was doing this for adapters, because they need to trace some calls and I thought it was a good insight into how our endpoints work and what part of them lasts.

Some examples

You can see what is liquibase doing.
db_migration

You can filter single enpoint and see all the calls
filtering_endpoints

List of the calls and information about them
calls_on_endpoint

Then you can inspect call and look what is happening during that call
detail_of_call

You can see what call is being done on the db
detail_of_query_in_call

You can get more information, like variables value as attributes in span in traces, but for that you would need to use manual instrumation, because for security reason they do not include them in automatical instrumentation.

This is just idea if we do not like it or we have something similuar tool just close it.

PR #4514, opened 2024-01-25
CANDLEPIN-778: Add API endpoint for Anonymous Consumer deletion

wottop:


src/main/java/org/candlepin/model/AnonymousCloudConsumerCurator.java
```diff
@@ -99,6 +99,26 @@ public AnonymousCloudConsumer getByCloudInstanceId(String instanceId) {
...
+     *
+     * @return the anonymous consumer based on the cloud instance ID, or null if no matching anonymous
+     * cloud consumer is found
+     */
+    public List<AnonymousCloudConsumer> getByCloudAccountId(String accountId) {
+        if (accountId == null || accountId.isBlank()) {
+            return new ArrayList<>();
+        }
+
+        Criteria criteria = this.createSecureCriteria()
```
nikosmoum, line 116: Please do not use the 'secure' criteria, they're deprecated and we're trying to get rid of them. Same for Hibernate Criteria API (use the `javax.persistence.criteria.*` imports instead of the `org.hibernate.criterion.*` ones) . We should use either JPQL queries, or JPA criteria API.

api/candlepin-api-spec.yaml
```diff
@@ -614,6 +614,35 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /cloud/consumers/anonymous:
+    delete:
+      description: Removes Anonymous Consumers
```
nikosmoum, line 619: Can we add a parenthesis here that this is for testing purposes only

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -288,4 +298,23 @@ private void validateCloudAuthenticationResult(CloudAuthenticationResult result)
             throw new NotAuthorizedException(errmsg);
         }
     }
+
+
+    @Override
+    @Transactional
+    public void deleteAnonymousConsumersByAccountId(String cloudAccountId) {
```
nikosmoum, line 305: Can we add a javadoc here explaining this endpoint is only for testing purposes?

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -288,4 +298,23 @@ private void validateCloudAuthenticationResult(CloudAuthenticationResult result)
...
+        Principal principal = this.principalProvider.get();
+        if (!principal.hasFullAccess()) {
+            throw new NotAuthorizedException(
+                i18n.tr("This endpoint requires access that the user does not have"));
+        }
```
nikosmoum, line 310: This is redundant; if an endpoint does not have a `@SecurityHole` annotation on it, and there is also no `@Verify` annotation on any of the arguments, then the default authorization filter is used (`SuperAdminAuthorizationFilter`) which is already doing this check

nikosmoum (changes requested):

nikosmoum (approved):

PR #4519, opened 2024-01-29
CANDLEPIN-779: Add API endpoint to cancel CloudAccountOrgSetupJobs

bakajstep:


src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -160,6 +163,52 @@ else if (version == 2) {
...
+        AsyncJobStatusQueryArguments asyncJobStatusQueryArguments = new AsyncJobStatusQueryArguments();
+        asyncJobStatusQueryArguments.setJobKeys(CloudAccountOrgSetupJob.JOB_KEY)
+            .setJobStates(AsyncJobStatus.JobState.CREATED, AsyncJobStatus.JobState.QUEUED,
+                AsyncJobStatus.JobState.RUNNING, AsyncJobStatus.JobState.SCHEDULED,
+                AsyncJobStatus.JobState.WAITING, AsyncJobStatus.JobState.FAILED_WITH_RETRY);
+        List<AsyncJobStatus> jobs = this.jobManager.findJobs(asyncJobStatusQueryArguments);
+
+        jobs = jobs.stream()
+            .filter(job -> job.getJobArguments().getAsString("cloud_account_id").equals(cloudAccountId))
+            .collect(Collectors.toList());
```
nikosmoum, line 182: Depending on when this is run, the query could fetch a very large amount of job statuses into memory.
I think using this one is preferred, because it will only fetch the jobs for that cloud account id:
```
Map<String,String> jobArgs = Map.of("cloud_account_id", cloudAccountId);
AsyncJobStatusCurator.fetchJobIdsByArguments(CloudAccountOrgSetupJob.JOB_KEY, jobArgs);
```
and then you can filter based on the job non-terminal job states

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -160,6 +163,52 @@ else if (version == 2) {
...
+
+        jobs.forEach(job -> {
+            jobManager.cancelJob(job.getId());
```
nikosmoum, line 185: There is a good chance that the cancelJob method can throw an `IllegalStateException` with the message "job is already in a terminal state", because between the time the query returns that a job is in QUEUED state, and the time you try to cancel it, the state could have changed to be terminal already. Can we catch those here, check for that exact message, and just continue?

nikosmoum: Also, a unit test with this use case would be good

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
```diff
@@ -796,4 +801,59 @@ private static Stream<Arguments> tokenVariation() throws ApiException {
...
+
+            assertThat(jobs).hasSize(1);
+            String beforeJobId = jobs.get(0).getId();
+            assertThatJob(jobs.get(0)).isCreated();
+
+            adminClient.cloudAuthorization().cancelCloudAccountJobs(accountId);
+
+            jobs = adminClient.jobs().listJobStatuses(null,
+                Set.of("CloudAccountOrgSetupJob"), Set.of("CANCELED"), null, Set.of("Anonymous"), null, null,
+                timeBeforeJobStarts, null, null, null, null, null);
```
nikosmoum, line 850: `cancelCloudAccountJobs` should now be returning a list of cancelled job ids, which you can use to pass into `listJobStatuses`, instead of all these other variables now: 
`listJobStatuses(Set.copyOf(returnList), null, null, null...)`

nikosmoum (changes requested):
a few things to change and we're good

api/candlepin-api-spec.yaml
```diff
@@ -643,6 +643,41 @@ paths:
...
+        - in: path
+          name: cloud_account_id
+          description: The ID of the cloud account
+          required: true
+          schema:
+            type: string
+      security: [ ]
+      responses:
+        204:
+          description: Jobs associated with cloud account ID were canceled
```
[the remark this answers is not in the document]
bakajstep: done

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -160,6 +163,52 @@ else if (version == 2) {
         }
     }
 
+    @Override
+    public List<String> cancelCloudAccountJobs(String cloudAccountId) {
```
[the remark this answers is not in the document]
bakajstep: done

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
bakajstep: done

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
bakajstep: done

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
bakajstep: done

PR #4525, opened 2024-01-29
CANDLEPIN-755: Update org_migrator for database schema changes

wottop:
I tested it by 
1. exported a test data org
2. deleted that org
3. imported that org again
4. exported that org again
5. compared output of step 1 and 4

This will remain in draft until we can vet it against stage. We don't yet know how long it will be until that is possible.

### DIFF
--- a/bin/scripts/org_migrator/org_migrator.py
+++ b/bin/scripts/org_migrator/org_migrator.py
@@ -14,19 +14,16 @@ import zipfile
 
 import cp_connectors as cp
 
-# Tables as of CP v4.2 (2022-09-29)
-#   - cp2_content
-#   - cp2_content_modified_products
-#   - cp2_owner_content
-#   - cp2_owner_products
-#   - cp2_pool_source_sub
-#   - cp2_product_attributes
-#   - cp2_product_branding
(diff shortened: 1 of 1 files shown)


bin/scripts/org_migrator/org_migrator.py
```diff
@@ -207,13 +205,13 @@ def fetch_product_uuids(db, org_id):
     # logic layer, so it's not critical we only select distinct UUIDs with these queries. Further,
     # our second query must be partitioned and repeatedly executed, eliminating the value of any
     # DB-level deduplication feature anyway.
-    product_uuid_query = 'SELECT op.product_uuid FROM cp2_owner_products op WHERE op.owner_id = %s ' + \
+    product_uuid_query = 'SELECT op.uuid FROM cp_products op WHERE op.namespace = %s ' + \
```
Ceiu, line 208: This won't catch global products unless they're used by a pool, which is supposed to be a catch-all for improperly mapped entities.

Ceiu (changes requested):

Ceiu:
Superseded by #4642

PR #4542, opened 2024-02-05
[M] CANDLEPIN-272: Removed cp query in ImportRecordCurator and RoleCurator 

joshmalbrecht:
- Removed the use of CandlepinQuery in ImportRecordCurator
- Removed the use of CandlepinQuery in RoleCurator

### DIFF
--- a/src/main/java/org/candlepin/resource/OwnerResource.java
+++ b/src/main/java/org/candlepin/resource/OwnerResource.java
@@ -82,6 +82,7 @@ import org.candlepin.model.Environment;
 import org.candlepin.model.EnvironmentCurator;
 import org.candlepin.model.ExporterMetadata;
 import org.candlepin.model.ExporterMetadataCurator;
+import org.candlepin.model.ImportRecord;
 import org.candlepin.model.ImportRecordCurator;
 import org.candlepin.model.InvalidOrderKeyException;
 import org.candlepin.model.Owner;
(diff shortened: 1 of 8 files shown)


src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1547,8 +1547,11 @@ public Iterable<ImportRecordDTO> getImports(
         @Verify(Owner.class) String ownerKey) {
         Owner owner = findOwnerByKey(ownerKey);
 
-        return this.translator.translateQuery(this.importRecordCurator.findRecords(owner),
-            ImportRecordDTO.class);
+        List<ImportRecordDTO> records = new ArrayList<>();
+        this.importRecordCurator.findRecords(owner).stream()
+            .forEach(record -> records.add(this.translator.translate(record, ImportRecordDTO.class)));
```
Ceiu, line 1552: This will do the translator look up for every element in the record. When we're processing streams, use the stream mapper instead:

```
.findRecords(owner)
.stream()
.map(this.translator.getStreamMapper(ImportRecord.class, ImportRecordDTO.class))
```

And then you can just return the mapped stream directly instead of having to recollect it as a list.

Ceiu (changes requested):

src/main/java/org/candlepin/resource/OwnerResource.java
joshmalbrecht: Good catch! Updated to use the StreamMapper

Ceiu: A couple other things that come to mind after having looked at the other CPQ removal PR:

- The title of this commit implies a wider reach than it actually has. We should probably clean that up a bit for the sanity of future maintainers going through the commit history
- CPQ endpoints have implicit paging functionality which will be lost by this change unless the PagingUtil or a custom paging implementation is added.

nikosmoum: Based on the investigation done when creating this epic, we don't want to do paging for every single endpoint that returns CPQ today. We identified an explicit list of endpoints that either stakeholders (katello/rhsm-api/etc.) already use with paging, or can potentially return a massive amount of results, and we'll only apply paging to those. For this specific PR, the `GET /owners/<orgid>/imports` endpoint is the one that's affected, which isn't in the list of endpoints that we want/need to page

joshmalbrecht: Updated the title of the commit!

Ceiu: Based on what @nikosmoum said, we need not worry about the loss of paging here; as much as it hurts me to lose.

PR #4543, opened 2024-02-05
CANDLEPIN-131: Removed CP Query from CDN, GuestId, and Entitlement 

wottop:
- CdnResource
- CdnCurator
- GuestIdResource
- GuestIdCurator
- EntitlementCurator

### DIFF
--- a/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -23,7 +23,6 @@ import com.google.inject.persist.Transactional;
 import org.hibernate.Criteria;
 import org.hibernate.Hibernate;
 import org.hibernate.ReplicationMode;
-import org.hibernate.criterion.DetachedCriteria;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Restrictions;
 import org.slf4j.Logger;
@@ -67,17 +66,15 @@ public class EntitlementCurator extends AbstractHibernateCurator<Entitlement> {
     private static final Logger log = LoggerFactory.getLogger(EntitlementCurator.class);
 
     private final ConsumerTypeCurator consumerTypeCurator;
-    private final CandlepinQueryFactory cpQueryFactory;
 
     /**
      * default ctor
      */
     @Inject
-    public EntitlementCurator(ConsumerTypeCurator consumerTypeCurator, CandlepinQueryFactory cpQueryFactory) {
+    public EntitlementCurator(ConsumerTypeCurator consumerTypeCurator) {
         super(Entitlement.class);
 
         this.consumerTypeCurator = Objects.requireNonNull(consumerTypeCurator);
-        this.cpQueryFactory = Objects.requireNonNull(cpQueryFactory);
     }
 
     // TODO: handles addition of new entitlements only atm!
@@ -512,11 +509,13 @@ public class EntitlementCurator extends AbstractHibernateCurator<Entitlement> {
         return this.entityManager.get().createQuery(query).getSingleResult().intValue();
     }
 
-    public CandlepinQuery<Entitlement> listByOwner(Owner owner) {
-        DetachedCriteria criteria = DetachedCriteria.forClass(Entitlement.class)
-            .add(Restrictions.eq("owner", owner));
+    public List<Entitlement> listByOwner(Owner owner) {
+        String jpql = "SELECT e FROM Entitlement e WHERE e.owner = :owner";
 
-        return this.cpQueryFactory.buildQuery(this.currentSession(), criteria);
+        return this.getEntityManager()
+            .createQuery(jpql, Entitlement.class)
+            .setParameter("owner", owner)
+            .getResultList();
     }
 
     /**
@@ -609,7 +608,7 @@ public class EntitlementCurator extends AbstractHibernateCurator<Entitlement> {
      * @param activeOn The date we want to see entitlements which are active on.
      * @return List of entitlements.
      */
-    public CandlepinQuery<Entitlement> listByConsumerAndDate(Consumer consumer, Date activeOn) {
+    public List<Entitlement> listByConsumerAndDate(Consumer consumer, Date activeOn) {
 
         /*
          * Essentially the opposite of the above query which searches for entitlement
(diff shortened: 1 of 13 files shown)


src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -1060,23 +1066,30 @@ private List<Entitlement> findByStackIds(Consumer consumer, List<String> stackId
...
-            .createAlias("ent_pool.attributes", "attrs")
-            .add(Restrictions.eq("attrs.indices", attributeName))
-            .add(Restrictions.eq("attrs.elements", value));
+        String jpql = "SELECT e FROM Entitlement e " +
+            "JOIN Pool p on e.pool = p.id " +
+            "JOIN PoolAttribute pa on p.id = pa.poolId " +
+            "WHERE pa.name = :attributeName " +
+            "AND pa.value = :value ";
+        if (consumer != null) {
+            jpql += "AND e.consumer = :consumer";
```
Ceiu, line 1078: This may be a good place for a JPA criteria query instead of string concatenation.

src/test/java/org/candlepin/controller/EntitlerTest.java
```diff
@@ -487,10 +486,10 @@ public void testRevokesLapsedUnmappedGuestEntitlementsOnAutoHeal() throws Except
 
         p1.setEntitlements(Set.of(e1));
 
-        CandlepinQuery cqmock = mock(CandlepinQuery.class);
-        when(cqmock.iterator()).thenReturn(Collections.singletonList(e1).iterator());
+        List queryList = mock(List.class);
```
Ceiu, line 489: We don't need to (nor want to) mock a list. Just return a proper list straight away and avoid all the pain that comes with mocking it out.

Note that this applies to all the areas in which this was done.

Ceiu (changes requested):
Couple minor issues to address, but also the commit message format should follow the [commit guidelines](https://github.com/candlepin/candlepin/blob/main/CONTRIBUTING.md#coding-guidelines). The PR title can be edited separately if the commit header is unable to properly encapsulate it.

src/main/java/org/candlepin/resource/CdnResource.java
```diff
@@ -44,18 +46,26 @@ public class CdnResource implements CdnApi {
...
 
     @Override
-    public CandlepinQuery<CdnDTO> getContentDeliveryNetworks() {
-        return this.translator.translateQuery(curator.listAll(), CdnDTO.class);
+    public Stream<CdnDTO> getContentDeliveryNetworks() {
+        Collection<Cdn> cdns;
+        cdns = curator.listAllCdns();
+        Stream<CdnDTO> stream = cdns.stream()
+            .map(this.translator.getStreamMapper(Cdn.class, CdnDTO.class));
+        return this.pagingUtilFactory.forClass(CdnDTO.class)
```
Ceiu, line 67: Based on [this comment](https://github.com/candlepin/candlepin/pull/4542#discussion_r1479617294) from @nikosmoum, we only need to add paging to endpoints which are known to use paging today. If this isn't one of them, we should pull out the paging util usage.

src/test/java/org/candlepin/resource/GuestIdResourceTest.java
```diff
@@ -147,15 +151,13 @@ public void setUp() {
...
-
-        CandlepinQuery<GuestIdDTOArrayElement> result = resource.getGuestIds(consumer.getUuid());
+        List<GuestId> queryList = spy(new LinkedList<>());
+        List<GuestIdDTOArrayElement> dtoQueryList = new LinkedList<>();
+        when(guestIdCurator.listByConsumer(eq(consumer))).thenReturn(queryList);
 
-        verify(query, times(1)).transform(any(ElementTransformer.class));
-        assertEquals(result, dtoQuery);
+        Stream<GuestIdDTOArrayElement> result = resource.getGuestIds(consumer.getUuid());
+        verify(queryList, times(1)).stream();
```
Ceiu, line 159: We do not care if the `.stream` method was called. This check locks us into a specific implementation for processing the list, rather than just verifying the output is correct.

Ceiu (changes requested):
The commit title is still too general (we've only removed it from a handful of classes in this commit).

Something closer to what was done [here](https://github.com/candlepin/candlepin/pull/4542/commits) would be clearer. Keep in mind that we use the commits to generate changelogs, so the clearer and more consistent our commit messages are, the better our changelogs are for our stakeholders.

src/main/java/org/candlepin/resource/CdnResource.java
nikosmoum: We don't need to have paging on this endpoint, so yes, lets remove the paging util

src/main/java/org/candlepin/resource/GuestIdResource.java
```diff
@@ -141,10 +145,13 @@ public void deleteGuest(@Verify(Consumer.class) String consumerUuid, String gues
...
-    public CandlepinQuery<GuestIdDTOArrayElement> getGuestIds(@Verify(Consumer.class) String consumerUuid) {
+    public Stream<GuestIdDTOArrayElement> getGuestIds(@Verify(Consumer.class) String consumerUuid) {
         Consumer consumer = consumerCurator.findByUuid(consumerUuid);
-        return translator.translateQuery(guestIdCurator.listByConsumer(consumer),
-            GuestIdDTOArrayElement.class);
+        List<GuestId> guestIds = guestIdCurator.listByConsumer(consumer);
+        Stream<GuestIdDTOArrayElement> stream = guestIds == null ? Stream.empty() : guestIds.stream()
+            .map(this.translator.getStreamMapper(GuestId.class, GuestIdDTOArrayElement.class));
+        return this.pagingUtilFactory.forClass(GuestIdDTOArrayElement.class)
+            .applyPaging(stream, guestIds.size());
```
Ceiu, line 154: Something from Sonarqube that actually checks out: guestIds is implied to be nullable here, but we use it on line 154 as if it's not.

Since we should treat other classes as black boxes and we don't have any contract established around the nullability of the output of our curator methods, we should gate the whole thing around whether or not it's null:

```
List<GuestId> guestIds = guestIdCurator.listByConsumer(consumer);

if (guestIds != null) {
  Stream<GuestIdDTOArrayElement> stream = guestIds.stream()
    .map(this.translator.getStreamMapper(...);

  return this.pagingUtilFactory.forClass(...)
    .applyPaging(...);
}

return Stream.empty();
```

src/main/java/org/candlepin/resource/CdnResource.java
```diff
@@ -44,18 +46,25 @@ public class CdnResource implements CdnApi {
     private CdnCurator curator;
     private CdnManager cdnManager;
     private ModelTranslator translator;
+    private PagingUtilFactory pagingUtilFactory;
```
Ceiu, line 49: Another win from Sonarqube (wow!). This is unused after the removal of paging from the CDN accessor.

src/main/java/org/candlepin/model/CdnCurator.java
```diff
@@ -51,6 +53,13 @@ public Cdn getByLabel(String label) {
         }
     }
 
+    public List<Cdn> listAllCdns() {
```
Ceiu, line 56: Is an explicit query needed over `.listAll`?

Ceiu (changes requested):

Ceiu (approved):

PR #4550, opened 2024-02-08
[M] CANDLEPIN-214: Removed cp query from ContentOverride and Pool curators

joshmalbrecht:
- removed CandlepinQuery from ContentOverrideCurator
- removed CandlepinQuery from PoolCurator
- removed some unused variables in ConsumerContentOverrideResourceTest


src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1426,46 +1433,56 @@ public Map<String, List<Pool>> mapPoolsBySubscriptionIds(Collection<String> subs
         return output;
     }
 
-    @SuppressWarnings("unchecked")
-    public CandlepinQuery<Pool> getPoolsBySubscriptionId(String subId) {
+    public List<Pool> getPoolsBySubscriptionId(String subId) {
```
Ceiu, line 1436: This method can call into `getPoolsBySubscriptionIds` with a list of the single element provided so we only have one query to maintain, both literally and figuratively.

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -106,17 +105,23 @@ public CandlepinQuery<Pool> listByOwner(Owner owner) {
...
     @Transactional
-    public CandlepinQuery<Pool> listByOwner(Owner owner, Date activeOn) {
-        DetachedCriteria criteria = DetachedCriteria.forClass(Pool.class)
-            .add(Restrictions.eq("owner", owner))
-            .add(Restrictions.eq("activeSubscription", Boolean.TRUE));
+    public List<Pool> listByOwner(Owner owner, Date activeOn) {
+        String jpql = "SELECT p FROM Pool p " +
+            "WHERE p.owner.id = :owner_id AND p.activeSubscription = true ";
+
+        if (activeOn != null) {
```
Ceiu, line 112: When we're doing stuff like dynamically building JPQL strings, we should evaluate if it'd be better to use JPA criteria instead.

src/main/java/org/candlepin/model/ContentOverrideCurator.java
```diff
@@ -42,12 +45,14 @@ public ContentOverrideCurator(Class<T> entityType, String parentAttrName) {
...
-            .add(Restrictions.eq(parentAttr, parent));
+    public List<T> getList(Parent parent) {
+        CriteriaBuilder cb = this.getEntityManager().getCriteriaBuilder();
+        CriteriaQuery<T> query = cb.createQuery(this.entityType());
+        Root<T> root = query.from(this.entityType());
+        query.select(root);
+        query.where(cb.equal(root.get(parentAttr), parent));
 
-        return this.cpQueryFactory.<T>buildQuery(this.currentSession(), criteria);
+        return this.listByCriteria(query);
```
Ceiu, line 55: Going through `listByCriteria` adds overhead we didn't have before... kind of (CPQ magic not withstanding). Since we're just a list (aptly matching our poor method name; side note: we need not maintain existing method names when they're not good), we need not be worried about the paging bits. If this method is used primarily for a paged endpoint, we should be updating this method to be more explicit about that fact rather than relying on magic.

Hopefully a simple JPA query works here.

Ceiu (changes requested):

src/main/java/org/candlepin/model/PoolCurator.java
joshmalbrecht: Good catch! Yeah, that's way better. Updated

src/main/java/org/candlepin/model/PoolCurator.java
joshmalbrecht: Good point, this works better using JPA criteria. Updated

src/main/java/org/candlepin/model/ContentOverrideCurator.java
joshmalbrecht: I'm confused on this. We definitely don't need use the `AbstractHibernateCurator.listByCriteria` method, but that method is only creating the query and then returning the result list:

```
return this.getEntityManager()
    .createQuery(query)
    .getResultList();
```

It is not doing anything with paging and I don't believe that the results from this method need to be paged. I updated the method to just create the query and then return the result list.

Ceiu: My mistake. I confused it with the `listByCriteria` that also does paging. You can disregard this comment.

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1426,46 +1440,36 @@ public Map<String, List<Pool>> mapPoolsBySubscriptionIds(Collection<String> subs
...
+    public List<Pool> getPoolsBySubscriptionIds(Collection<String> subIds) {
+        if (subIds == null || subIds.isEmpty()) {
+            return new ArrayList<>();
+        }
 
-            List<String> ids = session.createCriteria(SourceSubscription.class)
-                .add(CPRestrictions.in("subscriptionId", subIds))
-                .setProjection(Projections.distinct(Projections.property("pool.id")))
-                .list();
+        String jpql = "SELECT p FROM SourceSubscription ss " +
```
Ceiu, line 1456: Missing the DISTINCT keyword on our selection

Ceiu (changes requested):
Other changes look good, but I caught one more thing on the second pass.

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -106,17 +109,26 @@ public CandlepinQuery<Pool> listByOwner(Owner owner) {
...
 
         if (activeOn != null) {
-            criteria.add(Restrictions.le("startDate", activeOn))
-                .add(Restrictions.ge("endDate", activeOn));
+            predicates.add(cb.lessThanOrEqualTo(root.get(Pool_.startDate), activeOn));
+            predicates.add(cb.greaterThanOrEqualTo(root.get(Pool_.endDate), activeOn));
         }
 
-        return this.cpQueryFactory.buildQuery(this.currentSession(), criteria);
+        query.where(predicates.toArray(new Predicate[predicates.size()]));
```
Ceiu, line 127: It's so crazy that the JPA standard doesn't just accept a list straight away and instead forces this double allocation and copy nonsense on us.

The Predicate interface doesn't look to be too complex, and we could probably build a handy version of that which is mutable or at least has a builder so we could do this without incurring the pain of that second array allocation and copy.

joshmalbrecht: I know right! Having to move over the Predicates into a new array is silly haha

I like the idea of having our own version or some sort of builder that can be used throughout the code base to avoid this issue! I can get a Jira logged for this

joshmalbrecht: https://issues.redhat.com/browse/CANDLEPIN-803

Ceiu (approved):

PR #4551, opened 2024-02-08
CANDLEPIN-793: Removed CandlepinQuery from AHC.listAll and listAllByIds

wottop:
- Updated AbstractHibernateCurator.listAll and .listAllByIds to return a List
  instead of the deprecated CandlepinQuery
- Updated implementation of AbstractHibernateCurator.listAll and .listAllByIds
  to use JPA criteria instead of the deprecated Hibernate criteria and
  CandlepinQuery
- Updated affected callers with logic utilizing Java collection functionality
- Removed CandepinQueryInterceptorTest now because it relied on the
  listAll method in its CandlepinQuery form. It was already slated for removal
  in this transition.


src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -255,17 +255,17 @@ public E create(E entity, boolean flush) {
     /**
      * @return all entities for a particular type.
      */
-    public CandlepinQuery<E> listAll() {
+    public List<E> listAll() {
         DetachedCriteria criteria = this.createSecureDetachedCriteria();
 
-        return this.cpQueryFactory.buildQuery(this.currentSession(), criteria);
+        return this.cpQueryFactory.<E>buildQuery(this.currentSession(), criteria).list();
```
Ceiu, line 261: Let's just replace this with a CPQ-less implementation now.

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -255,17 +255,17 @@ public E create(E entity, boolean flush) {
...
     }
 
-    public CandlepinQuery<E> listAllByIds(Collection<? extends Serializable> ids) {
+    public List<E> listAllByIds(Collection<? extends Serializable> ids) {
         DetachedCriteria criteria = this.createSecureDetachedCriteria()
             .add(CPRestrictions.in("id", ids));
 
-        return this.cpQueryFactory.<E>buildQuery(this.currentSession(), criteria);
+        return this.cpQueryFactory.<E>buildQuery(this.currentSession(), criteria).list();
```
Ceiu, line 268: Same here; no reason to keep the CPQ implementation around anymore.

Ceiu (changes requested):

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -255,17 +255,44 @@ public E create(E entity, boolean flush) {
...
+        Root<E> root = criteriaQuery.from(this.entityType);
+        criteriaQuery.select(root);
+
+        List<Predicate> predicates = new ArrayList<>();
+        Predicate securityPredicate = this.getSecurityPredicate(this.entityType, criteriaBuilder, root);
+
+        if (securityPredicate != null) {
+            predicates.add(securityPredicate);
+            criteriaQuery.where(predicates.toArray(new Predicate[predicates.size()]));
+        }
```
joshmalbrecht, line 272: You should be able to just add the security predicate to the `criteria.where` and reduce the following to:

```
        Predicate securityPredicate = this.getSecurityPredicate(this.entityType, criteriaBuilder, root);
        }

```

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -255,17 +255,44 @@ public E create(E entity, boolean flush) {
...
-            .add(CPRestrictions.in("id", ids));
+    public List<E> listAllByIds(Collection<? extends Serializable> ids) {
+        EntityManager em = this.getEntityManager();
+        CriteriaBuilder criteriaBuilder = em.getCriteriaBuilder();
+        CriteriaQuery<E> criteriaQuery = criteriaBuilder.createQuery(this.entityType);
+        Root<E> root = criteriaQuery.from(this.entityType);
+        criteriaQuery.select(root);
+
+        CriteriaBuilder.In<String> inClause = criteriaBuilder.in(root.get("id"));
+        ids.stream().forEach(id -> inClause.value(id.toString()));
```
joshmalbrecht, line 284: I think you can do something like `Predicate idsInPredicate = inClause.in(ids);` instead of looping over all the ids and then you can add `idsInPredicate` to the `predicates` list. There is probably a better name than idsInPredicate, I was just giving an example on this!

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -100,11 +106,21 @@ public int createDeletedConsumers(Collection<String> consumerIds) {
     }
 
     @SuppressWarnings("unchecked")
-    public CandlepinQuery<DeletedConsumer> findByDate(Date date) {
```
joshmalbrecht, line 103: Why was this method updated, it doesn't look like it was affected by the `listAll` and `listAllByIds` changes?

src/main/java/org/candlepin/model/OwnerCurator.java
```diff
@@ -154,11 +154,13 @@ public Owner getByKey(String key) {
     }
 
     @Transactional
-    public CandlepinQuery<Owner> getByKeys(Collection<String> keys) {
-        DetachedCriteria criteria = this.createSecureDetachedCriteria()
-            .add(CPRestrictions.in("key", keys));
+    public List<Owner> getByKeys(Collection<String> keys) {
```
joshmalbrecht, line 157: Same comment as `DeletedConsumerCurator.findByDate`

joshmalbrecht (changes requested):
Had a couple questions and found a couple lines that could be improved

src/main/java/org/candlepin/model/OwnerCurator.java
wottop: See the comment for DeletedConsumerResource

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -776,12 +776,15 @@ protected void populateEntity(Pool entity, PoolDTO dto) {
...
     @Wrapped(element = "owners")
-    public CandlepinQuery<OwnerDTO> listOwners(String keyFilter) {
-        CandlepinQuery<Owner> query = keyFilter != null ?
+    public Iterable<OwnerDTO> listOwners(String keyFilter) {
+        List<Owner> query = keyFilter != null ?
             this.ownerCurator.getByKeys(Arrays.asList(keyFilter)) :
             this.ownerCurator.listAll();
 
-        return this.translator.translateQuery(query, OwnerDTO.class);
+        return query.stream().map(this.translator.getStreamMapper(
```
Ceiu, line 784: Let's drop the `.map` call down to its own line so the inner `getStreamMapper` call doesn't need to be broken up across three lines.

Additionally, we can return a Stream rather than an Iterable to avoid needing to convert the stream back into a list.

```
return query.stream()
  .map(this.translator.getStreamMapper(Owner.class, OwnerDTO.class));
```

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
joshmalbrecht: It looks like this was updated for the changes in DeletedConsumerResource

wottop: This method is used in the resource where it uses listAll and this method as a result of a condition. They must return the same object.

src/main/java/org/candlepin/resource/ProductResource.java
```diff
@@ -99,7 +99,11 @@ private Product fetchProduct(String productUuid) {
 
     @Override
     public Iterable<ProductDTO> getProducts() {
-        return this.translator.translateQuery(this.productCurator.listAll(), ProductDTO.class);
+        List<Product> products = this.productCurator.listAll();
+        return products.stream().map(this.translator.getStreamMapper(
```
Ceiu, line 103: Same deal here. We can avoid the collection step by just returning a stream. If we are not able to change that for OpenAPI reasons (I don't see why we wouldn't be able to, the resulting response type is still "list"), then we can at least make it clearer by stacking the operations at the same indentation level:

```
return this.productCurator.listAll()
  .stream()
  .map(this.translator.getStreamMapper(Product.class, ProductDTO.class))
  .toList();
```

Note this occurs in a number of places throughout this PR.

src/main/java/org/candlepin/sync/Exporter.java
```diff
@@ -600,33 +600,28 @@ private void exportDistributorVersions(File baseDir) throws IOException {
...
-                    try {
-                        File file = new File(cdnDir.getCanonicalPath(), cdn.getLabel() + ".json");
-                        writer = new FileWriter(file);
-                        cdnExporter.export(mapper, writer, cdn);
-                    }
-                    finally {
-                        if (writer != null) {
-                            writer.close();
-                        }
+        if (iterator.hasNext()) {
```
Ceiu, line 605: Since we have to hit this method, let's take the time to clean it up. We can make use of try-with-resources to avoid the finally blocks, and it's no longer necessary to explicitly close the outer iterator since we aren't worried about the cursor hanging open. In fact, an iterator probably isn't even the most effective way to do this.

```
List<Cdn> cdns = this.cdnCurator.listAll();

if (cdns != null && !cdns.isEmpty()) {
  // create cdn directory

  for (CDN cdn : cdns) {
    File cdnFile = ...
    try (FileWriter writer = ...) {


    }
  }
}
```

src/test/java/org/candlepin/async/tasks/EntitlerJobTest.java
```diff
@@ -231,9 +230,8 @@ private void stubEntitlerErrorResult(String poolId) throws EntitlementRefusedExc
 
     private void stubPoolCuratorPoolsById() {
         final Pool pool = createPool("hello");
-        final CandlepinQuery<Pool> cqmock = mock(CandlepinQuery.class);
-        when(cqmock.iterator()).thenReturn(Collections.singletonList(pool).iterator());
-        when(poolCurator.listAllByIds(anySet())).thenReturn(cqmock);
+        final List<Pool> pools = Collections.singletonList(pool);
```
Ceiu, line 233: Use `List.of(...)` instead of `Collections.singletonList(...)`

Note that this occurs multiple times throughout this PR.

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -255,17 +255,38 @@ public E create(E entity, boolean flush) {
...
-            .add(CPRestrictions.in("id", ids));
+    public List<E> listAllByIds(Collection<? extends Serializable> ids) {
+        EntityManager em = this.getEntityManager();
+        CriteriaBuilder criteriaBuilder = em.getCriteriaBuilder();
+        CriteriaQuery<E> criteriaQuery = criteriaBuilder.createQuery(this.entityType);
+        Root<E> root = criteriaQuery.from(this.entityType);
+        criteriaQuery.select(root);
+
+        List<Predicate> predicates = new ArrayList<>();
+        predicates.add(root.get("id").in(ids));
```
Ceiu, line 281: Note that this will run headlong into the IN-operator limit issues. The old code has the somewhat unreliable/buggy `CPRestrictions` variant, but that shouldn't be used anymore either. Instead we should be partitioning the input collection(s) as we do elsewhere. This can be done with `ParameterExpression`s and the same looping logic we use with our JPQL queries.

An example/tutorial of this can be found here: https://www.baeldung.com/jpa-query-parameters

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -255,17 +255,38 @@ public E create(E entity, boolean flush) {
     /**
      * @return all entities for a particular type.
      */
-    public CandlepinQuery<E> listAll() {
-        DetachedCriteria criteria = this.createSecureDetachedCriteria();
+    public List<E> listAll() {
+        EntityManager em = this.getEntityManager();
+        CriteriaBuilder criteriaBuilder = em.getCriteriaBuilder();
```
Ceiu, line 260: :+1: 

Happy to see the JPA variant of this and not the old Hibernate criteria version.

Ceiu (changes requested):
For the git commit, we can be a bit more explicit and clear with what's happening in this commit:

```
CANDLEPIN-793: Removed CandlepinQuery from AHC.listAll and listAllByIds

- Updated AbstractHibernateCurator.listAll and .listAllByIds to return a List
  instead of the deprecated CandlepinQuery
- Updated implementation of AbstractHibernateCurator to use JPA criteria instead
  of the deprecated Hibernate criteria and CandlepinQuery
- Updated affected callers with logic utilizing Java collection functionality
```

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
Ceiu: To add to this, it's not strictly necessary to maintain a list of predicates, since we have a known predicate with the IN clause, and we *may* have one with the "security" predicate. So we can avoid the list with a single predicate as follows:

```
Predicate predicate = ...

Predicate securityPredicate = ...
  predicate = criteriaBuilder.and(predicate, securityPredicate);
}

criteriaQuery.where(predicate);
```

Ceiu (changes requested):

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -255,17 +258,47 @@ public E create(E entity, boolean flush) {
...
+
+        Predicate securityPredicate = this.getSecurityPredicate(this.entityType, criteriaBuilder, root);
+        if (securityPredicate != null) {
+            predicate = criteriaBuilder.and(predicate, securityPredicate);
+        }
+        criteriaQuery.where(predicate);
+        TypedQuery<E> query = em.createQuery(criteriaQuery);
+
+        // to ensure uniqueness of the members of the input collection
+        Set<Serializable> dedupe = new HashSet<>();
```
Ceiu, line 294: This addresses the immediate concern with output duplication, but now we're doing this on every invocation, which isn't necessary. We could also reuse the `ids` variable here since we're not interested in any methods on the set class itself, nor do we need to worry about keeping the original in scope:

```
if (!(ids instanceof Set)) {
  ids = new HashSet<>(ids);
}
```

This ensures we only do the reallocation and deduplication in instances where we don't already have a set, and discards the now-useless original input set from the scope.

Ceiu (approved):

joshmalbrecht (approved):

PR #4553, opened 2024-02-12
CANDLEPIN-718: Extract key pair generator

Januson:


src/main/java/org/candlepin/pki/impl/BouncyCastleKeyPairGenerator.java
```diff
@@ -0,0 +1,254 @@
...
+     *
+     * @param bytes
+     *  the data to deserialize
+     *
+     * @param keyClass
+     *  the key class to which the data should be deserialized
+     *
+     * @throws IOException
+     *  if the key data is malformed, invalid, or otherwise cannot be deserialized to the specified
+     *  class
```
nikosmoum, line 189: The javadoc should be updated to correspond with the code (IOException for null data, KeyPairCreationException for the rest)

nikosmoum (commented):
Only a minor javadoc updated needed, but otherwise this is great

src/main/java/org/candlepin/pki/impl/BouncyCastleKeyPairGenerator.java
Januson: Done.

PR #4568, opened 2024-02-20
CANDLEPIN-140: Remove CandlepinQuery from ConsumerResource and ConsumerCurator

bakajstep:
- Remove the CandlepinQuery classes from ConsumerResource and ConsumerCurator and their test classes.

### DIFF
--- a/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -27,13 +27,10 @@ import com.google.inject.persist.Transactional;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.hibernate.Criteria;
-import org.hibernate.FetchMode;
 import org.hibernate.Hibernate;
 import org.hibernate.LockMode;
 import org.hibernate.LockOptions;
 import org.hibernate.ReplicationMode;
-import org.hibernate.criterion.Criterion;
-import org.hibernate.criterion.DetachedCriteria;
 import org.hibernate.criterion.Disjunction;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
@@ -61,6 +58,7 @@ import java.util.TreeSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.inject.Inject;
 import javax.inject.Provider;
@@ -507,48 +505,6 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
         return query.list();
     }
 
-    @SuppressWarnings("unchecked")
-    @Transactional
-    public CandlepinQuery<Consumer> listByOwner(Owner owner) {
-        DetachedCriteria criteria = this.createSecureDetachedCriteria()
-            .add(Restrictions.eq("ownerId", owner.getId()));
-
-        return this.cpQueryFactory.<Consumer>buildQuery(this.currentSession(), criteria);
-    }
-
-    /**
-     * Search for Consumers with fields matching those provided.
-     *
-     * @param userName
-     *     the username to match, or null to ignore
-     * @param types
-     *     the types to match, or null/empty to ignore
-     * @param owner
-     *     Optional owner to filter on, pass null to skip.
-     * @return a list of matching Consumers
-     */
-    @SuppressWarnings("unchecked")
-    @Transactional
-    public CandlepinQuery<Consumer> listByUsernameAndType(String userName, List<ConsumerType> types,
-        Owner owner) {
-
-        DetachedCriteria criteria = this.createSecureDetachedCriteria();
-
-        if (userName != null) {
-            criteria.add(Restrictions.eq("username", userName));
-        }
-
-        if (types != null && !types.isEmpty()) {
-            criteria.add(Restrictions.in("type", types));
-        }
-
-        if (owner != null) {
-            criteria.add(Restrictions.eq("owner", owner));
-        }
-
-        return this.cpQueryFactory.<Consumer>buildQuery(this.currentSession(), criteria);
-    }
-
     /**
      * @param updatedConsumer
      *     updated Consumer values.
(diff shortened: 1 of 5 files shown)


src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1202,43 +1156,38 @@ public List<String> getInactiveConsumerIds(Instant lastCheckedInRetention, Insta
...
+            SELECT c FROM Consumer c
+            JOIN c.hypervisorId h
+            WHERE c.ownerId = :ownerId
+                AND LOWER(h.hypervisorId) IN :lowerCaseHypervisorIds
+            ORDER BY h.hypervisorId ASC""";
 
-        return disjunction;
+        return getEntityManager().createQuery(jpql, Consumer.class)
+            .setParameter("ownerId", ownerId)
+            .setParameter("lowerCaseHypervisorIds", lowerCaseHypervisorIds)
```
Ceiu, line 1177: This could run into the limit of parameters on the IN operator. The input needs to be properly partitioned.

src/main/java/org/candlepin/model/ConsumerContentOverrideCurator.java
```diff
@@ -31,4 +33,12 @@ public ConsumerContentOverrideCurator() {
     protected ConsumerContentOverride createOverride() {
         return new ConsumerContentOverride();
     }
+
+    public Collection<ConsumerContentOverride> listConsumerContentOverrides(Consumer consumer) {
```
Ceiu, line 37: Where will this be used? This doesn't have a corresponding usage from a resource method.

Ceiu (changes requested):
It looks like the changes for ConsumerResource have not yet been completed and still need to be added.

Also, let's follow our [commit guidelines](https://github.com/candlepin/candlepin/blob/main/CONTRIBUTING.md#git-commits) and use past-tense verbiage in the commit header and body, and wrap the lines earlier so they don't break in the middle of words.

src/main/java/org/candlepin/model/ConsumerContentOverrideCurator.java
bakajstep: sorry I forgot to remove it I thought we are calling the database in the resource layer so I wanted to extract it into the curator, but then I noticed the changed getList() and how it work.

src/main/java/org/candlepin/model/ConsumerCurator.java
bakajstep: done

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1202,43 +1156,45 @@ public List<String> getInactiveConsumerIds(Instant lastCheckedInRetention, Insta
...
-            .setLockMode(LockModeType.PESSIMISTIC_WRITE);
-    }
+        String jpql = """
+            SELECT c FROM Consumer c
+            JOIN c.hypervisorId h
+            WHERE c.ownerId = :ownerId
+                AND LOWER(h.hypervisorId) IN :lowerCaseHypervisorIds
+            ORDER BY h.hypervisorId ASC""";
+
+        List<Consumer> consumers = new ArrayList<>();
```
Ceiu, line 1175: This addresses the partitioning, but now we're at risk of three other problems:
- The lock mode on the query has gone missing, so any locking behavior existing callers may be expecting is gone.
- The input may or may not have duplicate entries, and if those duplicates exist in different partitions, we're guaranteed to get duplicate outputs since we lose the implied uniqueness from the schema.
- If the locking is restored, we're at risk of deadlock depending on the order of our input since we don't know which partition an input will end up in; we could end up locking consumers A, B, and C in various orders. Two or more threads doing this with overlapping partitioned inputs will eventually deadlock.

Ceiu (changes requested):

src/main/java/org/candlepin/model/ConsumerCurator.java
bakajstep: done

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1202,43 +1157,49 @@ public List<String> getInactiveConsumerIds(Instant lastCheckedInRetention, Insta
...
+            return new ArrayList<>();
         }
 
-        DetachedCriteria criteria = DetachedCriteria.forClass(Consumer.class)
-            .createAlias("hypervisorId", "hvsr")
-            .add(Restrictions.eq("ownerId", ownerId))
-            .add(this.getHypervisorIdRestriction(hypervisorIds))
-            .addOrder(org.hibernate.criterion.Order.asc("hvsr.hypervisorId"))
-            .setFetchMode("type", FetchMode.SELECT);
+        List<String> lowerCaseHypervisorIds = hypervisorIds.stream()
```
Ceiu, line 1166: As someone who enjoys streams, I like this solution; but I don't like that it means we're generating a new list just to throw it out again as we start partitioning.

Let me think on this for a bit to figure out a clean way to build a stream that we can throw at a partitioner without needing to recollect it first.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1202,43 +1157,49 @@ public List<String> getInactiveConsumerIds(Instant lastCheckedInRetention, Insta
      *     Org namespace to search
      * @return Consumer that matches the given
      */
-    @SuppressWarnings("unchecked")
     @Transactional
-    public CandlepinQuery<Consumer> getHypervisorsBulk(Iterable<String> hypervisorIds, String ownerId) {
+    public List<Consumer> getHypervisorsBulk(Collection<String> hypervisorIds, String ownerId) {
         if (hypervisorIds == null || !hypervisorIds.iterator().hasNext()) {
```
Ceiu, line 1162: With the type change, we can update this validation to use `.isEmpty` rather than pulling an iterator.

src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: Looks like `Stream.iterator` is also lazy, so we can use the magic of functional interfaces to pass that along the stream without first recollecting it:

```
Stream<String> hypervisorIds.stream()
  .distinct()
  ...

for (List<String> block : this.partition(lowerCaseHypervisorIds::iterator)) {
```

bakajstep: done

src/main/java/org/candlepin/model/ConsumerCurator.java
bakajstep: done

Ceiu (approved):

PR #4570, opened 2024-02-20
CANDLEPIN-796: Create ueber certificate generator

Januson:
- Add x509 certificate builder
- Add Ueber certificate generator


src/main/java/org/candlepin/pki/certs/UeberCertificateGenerator.java
```diff
@@ -286,11 +294,9 @@ private Content createUeberContent(UniqueIdGenerator idGenerator, Owner owner, P
          * See https://bugzilla.redhat.com/show_bug.cgi?id=1242310
          */
         private Date lateIn2049() {
-            Calendar cal = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
             // December 1, 2049 at 13:00 GMT
```
joshmalbrecht, line 297: Nit: This comment should be UTC and not GMT

joshmalbrecht (approved):
Looks good to me! I'll add a tag to get an additional set of eyes on this

src/main/java/org/candlepin/pki/certs/X509CertificateBuilder.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    private void addSubjectAltName(X509v3CertificateBuilder builder,
+        DistinguishedName distinguishedName, DistinguishedName subjectAltName) {
+
+        if (subjectAltName == null) {
+            return;
+        }
+
+        GeneralName subject = new GeneralName(GeneralName.directoryName, distinguishedName.value());
```
nikosmoum, line 222: Can we keep the implementation comment from BouncyCastlePKIUtility regarding subject alt names here? it provides details/rationale for why we're doing this, and referencing standards/RFC documents. I think this is the kind of knowledge we need to preserve

src/main/java/org/candlepin/pki/certs/UeberCertificateGenerator.java
```diff
@@ -109,22 +116,21 @@ public UeberCertificate generate(String ownerKey, Principal principal) {
...
             // There can only be one ueber certificate per owner, so delete the existing and regenerate it.
             this.ueberCertCurator.deleteForOwner(owner);
-            return  this.generateUeberCert(owner, principal.getUsername());
+            return  this.generateUeberCert(owner, username);
         }
         catch (Exception e) {
-            log.error("Problem generating ueber cert for owner: {}", ownerKey, e);
-            throw new BadRequestException(i18n.tr("Problem generating ueber cert for owner {0}", ownerKey),
-                e);
+            throw new BadRequestException(i18n.tr(
```
nikosmoum, line 122: Can we add back the log.error before throwing this? It is useful to have the details/stacktrace of why exactly the cert generation failed in the logs. If we only throw an HTTP-level exception then that would be lost

nikosmoum (changes requested):
2 minor requests and this is ready

nikosmoum (approved):

PR #4575, opened 2024-02-21
CANDLEPIN-203: Removed CP Query from Environment

wottop:
- removed use of CandlepinQuery from EnvironmentResourceTest
- the use of CandlepinQuery in EnvironmentResource was handled
  previously

### DIFF
--- a/src/test/java/org/candlepin/resource/EnvironmentResourceTest.java
+++ b/src/test/java/org/candlepin/resource/EnvironmentResourceTest.java
@@ -20,7 +20,6 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyList;
 import static org.mockito.ArgumentMatchers.anyString;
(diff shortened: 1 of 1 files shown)


src/test/java/org/candlepin/resource/EnvironmentResourceTest.java
```diff
@@ -172,7 +169,7 @@ void nothingToDelete() {
     @Test
     void canDeleteEmptyEnvironment() {
         when(this.envCurator.get(anyString())).thenReturn(this.environment1);
-        List<Consumer> mockedQuery = mockedQueryOf().list();
+        List<Consumer> mockedQuery = List.of();
```
joshmalbrecht, line 172: The `mockedQuery` variable name is a little misleading since it is just a list now that you removed the `mockedQueryOf()`. You could move this to line 174 and do `.thenReturn(List.of())` or rename the variable. This also applies to the other test changes you made.

joshmalbrecht (commented):
Just noticed a small change regarding the pre-existing variable names

PR #4580, opened 2024-02-22
[M] CANDLEPIN-200: Removed CandlepinQuery from ModelTranslator

bakajstep:


Ceiu (changes requested):
Removal looks fine. Let's just update the commit to change the focus away from a subclass and onto the main interface -- the method is removed from the ModelTranslator interface; SimpleModelTranslator is being updated as a consequence of that.

If you still want to call out the SMT change, that's fair, but it's a detail for the commit body.

bakajstep:
> Removal looks fine. Let's just update the commit to change the focus away from a subclass and onto the main interface -- the method is removed from the ModelTranslator interface; SimpleModelTranslator is being updated as a consequence of that.
> 
> If you still want to call out the SMT change, that's fair, but it's a detail for the commit body.

done

Ceiu (approved):

PR #4588, opened 2024-02-28
CANDLEPIN-123: Updated EntitlementCurator to not use Hibernate Criteria

wottop:
- Also updated tests for code cleanup

### DIFF
--- a/src/main/java/org/candlepin/model/EntitlementCurator.java
+++ b/src/main/java/org/candlepin/model/EntitlementCurator.java
@@ -20,11 +20,7 @@ import org.candlepin.paging.PageRequest;
 import com.google.common.collect.Iterables;
 import com.google.inject.persist.Transactional;
 
-import org.hibernate.Criteria;
 import org.hibernate.Hibernate;
-import org.hibernate.ReplicationMode;
-import org.hibernate.criterion.Order;
-import org.hibernate.criterion.Restrictions;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -44,6 +40,7 @@ import java.util.Set;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.persistence.EntityManager;
+import javax.persistence.NoResultException;
 import javax.persistence.Query;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/model/EntitlementCurator.java
```diff
@@ -1046,21 +1048,28 @@ private List<Entitlement> findByStackIds(Consumer consumer, List<String> stackId
...
 
         if (consumer != null) {
-            criteria.add(Restrictions.eq("consumer", consumer));
+            predicates.add(criteriaBuilder.equal(entRoot.get(Entitlement_.CONSUMER), consumer));
         }
-
-        return (List<Entitlement>) criteria.list();
+        Predicate[] predicateArray = new Predicate[predicates.size()];
+        criteriaQuery.where(predicates.toArray(predicateArray));
+        return em.createQuery(criteriaQuery).getResultList();
```
Januson, line 1072: It might be better to convert this to HQL as we don't need to do anything dynamic here.

Same with the `findUpstreamEntitlementForStack`.

wottop: In one of the CandlepinQuery PR's, the review requested that I use a JPA query instead of HQL when I had a condition for the consumer in the where clause.

https://github.com/candlepin/candlepin/pull/4543

Either is fine by me, but we should at least have a standard.

Januson: If it's a simple conditional part of WHERE such as `consumer != null` it can be expressed in query. I don't mind using criteria, but I think we should prefer HQL where possible.

Something like the following where the last AND replaces the condition.
```java
    String hql = "SELECT e FROM Entitlement e " +
                 "JOIN e.pool p " +
                 "JOIN p.product pr " +
                 "JOIN pr.attributes a " +
                 "LEFT JOIN p.sourceStack s " +
                 "WHERE a.key = :key AND a.value IN :stackIds " +
                 "AND p.sourceEntitlement IS NULL " +
                 "AND s.id IS NULL " +
                 "AND (e.consumer = :consumer OR :consumer IS NULL)";
```

Ceiu: Before we do something like this, I would like to see what the query planner for Postgres and MariaDB/MySQL do with it in the cases where we pass in a null and a non-null. I would hope they're smart enough to short-circuit on that last conditional, but I would like to see verification before we go about throwing our work to the DB in the name of avoiding JPA criteria.

This is one of those cases where standardization is going to be hard. There are conventions and habits that are functional but dangerous for maintenance purposes (string concat), some that are safe but harder to read (JPA criteria), and then we have things like this which are easy to read and avoid concat, but are entirely contingent on how our backend works when it comes to efficiency.

The one thing I want to avoid with any standardization we do is stifle creativity or picking the "correct" or "better" answer because the standard is too strict.

wottop: As it turns out, consumer_id cannot be null in the entitlement table. The dynamic part is whether we want to filter on consumer or not.

It means the condition of conusmer != null for the query string concat *and* the parameter assignment. Based on that, I'm inclined to keep the CriteriaBuilder code.

Januson (approved):

PR #4592, opened 2024-02-29
[F] CANDLEPIN-246: Defined default and max page size for pagination

joshmalbrecht:
- Created default page size configuration
- Created default max page size configuration
- Updated existing paginated endpoints to use the new pagination configurations
- Added configuration validation classes to validate configurations on startup

### DIFF
--- a/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -25,9 +25,13 @@ import org.candlepin.async.tasks.InactiveConsumerCleanerJob;
 import org.candlepin.async.tasks.JobCleaner;
 import org.candlepin.async.tasks.ManifestCleanerJob;
 import org.candlepin.async.tasks.UnmappedGuestEntitlementCleanerJob;
+import org.candlepin.config.validation.ConfigurationValidator;
+import org.candlepin.config.validation.IntegerConfigurationValidator;
 import org.candlepin.guice.CandlepinContextListener;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 
@@ -181,6 +185,10 @@ public class ConfigProperties {
     public static final String CACHE_ANON_CERT_CONTENT_TTL = "candlepin.cache.anonymous.cert.content.ttl";
     public static final String CACHE_ANON_CERT_CONTENT_MAX_ENTRIES =
         "candlepin.cache.anonymous.cert.content.max_entries";
+
+    // Paging
+    public static final String PAGING_SIZE = "candlepin.paging.default_page_size";
+    public static final String MAX_PAGING_SIZE = "candlepin.paging.max_page_size";
 
     public static final String SYNC_WORK_DIR = "candlepin.sync.work_dir";
 
@@ -413,6 +421,9 @@ public class ConfigProperties {
             this.put(CACHE_ANON_CERT_CONTENT_TTL, "120000"); // milliseconds
             this.put(CACHE_ANON_CERT_CONTENT_MAX_ENTRIES, "10000");
 
+            this.put(PAGING_SIZE, "10");
+            this.put(MAX_PAGING_SIZE, "1000");
+
             this.put(SUSPEND_MODE_ENABLED, "true");
 
             this.put(IDENTITY_CERT_YEAR_ADDENDUM, "5");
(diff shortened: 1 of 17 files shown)


src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -182,6 +184,10 @@ private ConfigProperties() {
     public static final String CACHE_ANON_CERT_CONTENT_MAX_ENTRIES =
         "candlepin.cache.anonymous.cert.content.max_entries";
 
+    // Paging
+    public static final String PAGING_SIZE = "candlepin.paging.size";
+    public static final String MAX_PAGING_SIZE = "candlepin.paging.max_size";
```
nikosmoum, line 189: can we make these a bit more clear? e.g.:
`candlepin.paging.default_page_size` and `candlepin.paging.max_page_size`?
That way the base namespace `candlepin.paging.` could be used in the future for other properties (e.g. `candlepin.paging.default_order` or `candlepin.paging.default_sort_field`

src/main/java/org/candlepin/config/IntegerConfigurationValidator.java
```diff
@@ -0,0 +1,217 @@
...
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.config;
```
nikosmoum, line 15: It may be worth moving all the 'config validation' classes in an `org.candlepin.config.validation` subpackage

src/main/java/org/candlepin/config/IntegerConfigurationValidator.java
```diff
@@ -0,0 +1,217 @@
...
+        if (value == null) {
+            return;
+        }
+
+        if (!lessThanKeys.isEmpty()) {
+            for (String lessThanKey : lessThanKeys) {
+                Integer otherValue = getInt(lessThanKey, config);
+                if (otherValue != null && value > otherValue) {
+                    this.addFailure("Configuration \"" + getKey() +
+                        "\" must be smaller than \"" + lessThanKey + "\"");
```
nikosmoum, line 164: This (and the greaterThan method as well) can be slightly improved if it was `Configuration X must be smaller than the value of configuration Y`.

src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -413,6 +419,9 @@ public static String jobConfig(String jobKey, String cfgName) {
             this.put(CACHE_ANON_CERT_CONTENT_TTL, "120000"); // milliseconds
             this.put(CACHE_ANON_CERT_CONTENT_MAX_ENTRIES, "10000");
 
+            this.put(PAGING_SIZE, "10");
+            this.put(MAX_PAGING_SIZE, "10000");
```
nikosmoum, line 423: I think we should reduce this. The `PageRequest.DEFAULT_PER_PAGE` was 10. And out of the 3 endpoints that specified their own max values, two of them had the max at 1000, and only one at 10000.
1000 seems like a reasonable max page value

nikosmoum (changes requested):
I have some suggestions, but it looks great overall

src/main/java/org/candlepin/config/ConfigProperties.java
joshmalbrecht: That makes sense! I like `candlepin.paging.default_page_size` and `candlepin.paging.max_page_size` a lot more. Updated

src/main/java/org/candlepin/config/IntegerConfigurationValidator.java
joshmalbrecht: Updated

src/main/java/org/candlepin/config/IntegerConfigurationValidator.java
joshmalbrecht: Good catch, I think that makes the message more clear. Updated in both methods

nikosmoum (approved):

PR #4594, opened 2024-03-01
CANDLEPIN-798: Create identity certificate generator

Januson:
- Extracted identity certificate creation into a certificate generator
- Removed IdentityCertServiceAdapter
- Added cascade delete for identity certificate


Ceiu:
Let's follow the [commit message guidelines](https://github.com/candlepin/candlepin/blob/main/CONTRIBUTING.md#git-commits) for tense so our changelog is consistent.

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -213,7 +213,7 @@ public static final class Facts {
      * field is a DTO for transmission to the client carrying PEM in plain text, and is
      * not stored in the database.
      */
-    @OneToOne
+    @OneToOne(cascade = CascadeType.REMOVE)
```
nikosmoum, line 216: Can we add a unit/integration test (probably in either ConsumerTest or ConsumerCuratorTest ) to test this?

nikosmoum (changes requested):

src/main/java/org/candlepin/model/Consumer.java
Januson: Done.

PR #4596, opened 2024-03-01
CANDLEPIN-799: Create anonymous content access and content access certificate generator

Januson:
- Extracted anonymous content access certificate creation to a certificate generator
- Extracted content access certificate creation to a certificate generator


src/test/java/org/candlepin/pki/certs/ContentAccessCertificateGeneratorTest.java
```diff
@@ -0,0 +1,269 @@
...
+        content.setVendor("vendor");
+        content.setContentUrl("/content/dist/rhel/$releasever/$basearch/os");
+        content.setGpgUrl("gpgUrl");
+        content.setArches("x86_64");
+        content.setMetadataExpiration(3200L);
+        content.setRequiredTags("TAG1,TAG2");
+
+        return content;
+    }
+
```
joshmalbrecht, line 248: Extra new line

src/test/java/org/candlepin/pki/certs/AnonymousCertificateGeneratorTest.java
```diff
@@ -0,0 +1,196 @@
...
+    void setUp() throws CertificateException, IOException {
+        Configuration config = TestConfig.defaults();
+        ObjectMapper objectMapper = ObjectMapperFactory.getObjectMapper();
+        X509V3ExtensionUtil extensionUtil = spy(new X509V3ExtensionUtil(
+            config, this.entitlementCurator, objectMapper, new Huffman()));
+        BouncyCastleSecurityProvider securityProvider = new BouncyCastleSecurityProvider();
+        BouncyCastleKeyPairGenerator keyPairGenerator = new BouncyCastleKeyPairGenerator(
+            securityProvider, mock(KeyPairDataCurator.class));
+        CertificateReaderForTesting certificateReader = new CertificateReaderForTesting();
+        this.contentCache = mock(AnonymousCertContentCache.class);
```
joshmalbrecht, line 96: We can use Mock annotation for AnonymousCertContentCache

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
```diff
@@ -0,0 +1,382 @@
...
+                converted.setRequiredTags(content.getRequiredTags());
+                converted.setArches(content.getArches());
+                converted.setContentUrl(content.getContentUrl());
+                converted.setReleaseVersion(content.getReleaseVersion());
+
+                return converted;
+            })
+            .toList();
+    }
+
```
joshmalbrecht, line 351: Extra new line

joshmalbrecht (changes requested):
I just noticed some small changes, but other than that it looks good!

src/main/java/org/candlepin/controller/ContentAccessMode.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /** simple content access (SCA) mode; clients have access to all org content by default */
+    ORG_ENVIRONMENT;
+
+    /**
+     * Fetches the default content access mode list database value
+     *
+     * @return the default content access mode list database value as a string
+     */
+    public static String defaultContentAccessModeList() {
```
Ceiu, line 34: I'm not much a fan of this method living here. Its output is not a list of enums, and the default list isn't a function of this type. This is a CA utility method living on a type, which feels kinda weird when we already have the CA manager to host such things.

To some degree, the decision to hard code the default mode and list was probably a mistake. These should (have) be(en) configurable.

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
```diff
@@ -0,0 +1,382 @@
...
+        this.certificateBuilder = Objects.requireNonNull(certificateBuilder);
+    }
+
+    /**
+     * Creates a new certificate. Certificate content and payload is cached and reused.
+     *
+     * @param consumer An anonymous consumer for which to create a certificate.
+     * @return Anonymous content access certificate
+     */
+    public AnonymousContentAccessCertificate createCertificate(AnonymousCloudConsumer consumer) {
```
Ceiu, line 119: The naming conventions chosen for the public generation op are different from generator to generator. I'm not sure if we have the ability to standardize these with an interface without getting too crazy with Java memes and stereotypes, but we can at the very least choose to normalize on something like "generate" or "createCertificate" across the lot of them.

Ceiu (changes requested):

nikosmoum (changes requested):

src/main/java/org/candlepin/pki/certs/ContentAccessCertificateGenerator.java
```diff
@@ -0,0 +1,327 @@
...
+import java.util.Set;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+
+/**
+ * Handles creation of the content access certificates.
+ */
+public class ContentAccessCertificateGenerator {
```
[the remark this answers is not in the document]
Januson: Do we want to rename the certificate a well?

src/test/java/org/candlepin/pki/certs/ContentAccessCertificateGeneratorTest.java
Januson: Done.

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
Januson: Done.

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -850,14 +463,14 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
      *  if unable to create x509 certificate due to security issue
      *
      * @return
-     *  the retieved or generated certificate
+     *  the retrieved or generated certificate
      */
     @Transactional
     public AnonymousContentAccessCertificate getCertificate(AnonymousCloudConsumer consumer)
```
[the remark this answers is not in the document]
Januson: Done.

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
Januson: I don't think there's a way to unify them behind an interface as they all do something different, but I at least unified the method name to `generate`.

src/main/java/org/candlepin/controller/ContentAccessMode.java
Januson: Do you want to move it back? I only wanted to move the enum as the nesting when using it from outside was annoying. The method looked like it might be a good enough fit there as well, but I can move it back.

src/main/java/org/candlepin/pki/certs/ContentAccessCertificateGenerator.java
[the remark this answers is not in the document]
nikosmoum: If its easy enough, do it (I know its referenced in a lot of places), otherwise we don't have to

nikosmoum (approved):
Looks good to me. I'll let others do the re-review/merge

joshmalbrecht (approved):

src/main/java/org/candlepin/controller/ContentAccessMode.java
Ceiu: @nikosmoum @joshmalbrecht Thoughts on this?

Ceiu (approved):

Ceiu:
My main feedback has been addressed. Only thing pending is what we want to do about CA util type stuff.

src/main/java/org/candlepin/controller/ContentAccessMode.java
nikosmoum: The enum class should be about the building block (the mode - org_env or entitlement), while the 'mode list' is at a higher level, using the building block, but not necesserily belonging to it. So I'd say this method fits the COntentAccessManager more

PR #4601, opened 2024-03-05
CANDLEPIN-804: Created kube file for candlepin-dev image

wottop:
- Converted from docker-compose.yml via Kompose tool
- Uses the most recent Candlepin dev image and the newest Postres image
- The Candlepin container start does not have actual pause based on the Postgres container startup as 'podman play kube' does not have the Sidecar feature from Kubernetes yet
- The Postgres container starts quickly and does not adversely affect the Candlepin container startup
- A single command 'podman play kube candlepin-deployment.yaml' will bring everything up for use

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -238,13 +238,19 @@ ENTRYPOINT ["/opt/tomcat/bin/catalina.sh", "run"]
 
 The Candlepin development image is designed to run right after pulling the image using default Candlepin and Tomcat configurations as well as a default certificate and key for TLS communication and encryption. Since this certificate and key is packaged in a publicly available container image, the use of the Candlepin development image should **not** be used in a production environment to avoid security risks.
 
-The following is an example on how to run the Candlepin development container using the docker compose file.
+The following is an example on how to run the Candlepin development container using a docker compose file or a kubernetes file via podman.
 
-The compose file is in the dev-container directory that will pull that development image and the most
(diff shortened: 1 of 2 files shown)


joshmalbrecht (changes requested):
I'm having some trouble getting this to work. My Candlepin container is getting the following error that indicates the Candlepin container cannot connect to the database container.

```
ERROR org.candlepin.guice.CandlepinContextListener - Liquibase exception: org.postgresql.util.PSQLException: Connection to postgres:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
```

Looking at the output of `podman ps`, all of the containers have `0.0.0.0:5432->5432/tcp, 0.0.0.0:8443->8443/tcp` in the `PORTS` column, which doesn't seem correct and may be part of the issue. Are you also running into this issue?

joshmalbrecht:
> I'm having some trouble getting this to work. My Candlepin container is getting the following error that indicates the Candlepin container cannot connect to the database container.
> 
> ```
> ERROR org.candlepin.guice.CandlepinContextListener - Liquibase exception: org.postgresql.util.PSQLException: Connection to postgres:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
> ```
> 
> Looking at the output of `podman ps`, all of the containers have `0.0.0.0:5432->5432/tcp, 0.0.0.0:8443->8443/tcp` in the `PORTS` column, which doesn't seem correct and may be part of the issue. Are you also running into this issue?

I was pulling in an old version of the Candlepin dev image and I needed to change my Podman network backend from cni to netavark!

joshmalbrecht (approved):

PR #4602, opened 2024-03-05
CANDLEPIN-238: Updated PoolCurator to not use Hibernate Criteria

bakajstep:
- Updated PoolCurator to not use Hibernate Criteria in queries.
- Added missing test for PoolCurator methods.
- Removed unused code.

### DIFF
--- a/src/main/java/org/candlepin/model/PoolCurator.java
+++ b/src/main/java/org/candlepin/model/PoolCurator.java
@@ -16,7 +16,6 @@ package org.candlepin.model;
 
 import org.candlepin.model.Pool.PoolType;
 import org.candlepin.model.activationkeys.ActivationKey;
-import org.candlepin.model.activationkeys.ActivationKeyPool;
 import org.candlepin.paging.Page;
 import org.candlepin.paging.PageRequest;
 import org.candlepin.util.Util;
@@ -25,22 +24,14 @@ import com.google.common.collect.Iterables;
 import com.google.inject.persist.Transactional;
 
 import org.hibernate.Criteria;
-import org.hibernate.FetchMode;
-import org.hibernate.Filter;
-import org.hibernate.Hibernate;
-import org.hibernate.ReplicationMode;
 import org.hibernate.criterion.Criterion;
 import org.hibernate.criterion.DetachedCriteria;
 import org.hibernate.criterion.Disjunction;
-import org.hibernate.criterion.Junction;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Property;
 import org.hibernate.criterion.Restrictions;
-import org.hibernate.criterion.SimpleExpression;
 import org.hibernate.criterion.Subqueries;
-import org.hibernate.internal.FilterImpl;
-import org.hibernate.query.Query;
 import org.hibernate.sql.JoinType;
 import org.hibernate.type.StringType;
 import org.slf4j.Logger;
@@ -56,7 +47,6 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
@@ -64,13 +54,14 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import javax.persistence.PersistenceUnitUtil;
+import javax.persistence.Query;
 import javax.persistence.TypedQuery;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Join;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
-
-
 
 /**
  * PoolCurator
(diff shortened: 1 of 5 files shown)


src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1335,21 +1255,24 @@ public Set<String> getDerivedPoolIdsForPools(Iterable<String> poolIds) {
...
+            """;
 
         if (poolIds != null && !poolIds.isEmpty()) {
             for (List<String> block : this.partition(poolIds)) {
-                eids.addAll(this.currentSession().createCriteria(Pool.class, "p")
-                    .createAlias("p.entitlements", "e")
-                    .add(Restrictions.in("p.id", block))
-                    .setProjection(Projections.distinct(Projections.property("e.id")))
-                    .list());
+                entitlementIds.addAll(this.currentSession().createQuery(jpql, String.class)
```
Ceiu, line 1269: This is creating a new instance of the query for every iteration, which we should avoid. Additionally, this is going through the Hibernate session rather than the JPA entity manager to create the query. We should update this to use the latter.

Note this occurs in several places throughout this PR.

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -942,97 +902,49 @@ public List<Pool> getBySubscriptionIds(String ownerId, Collection<String> subIds
...
-                .add(Restrictions.or(srcEntIsNull, srcEntEquals));
-
-            subIdMapCriteria.add(subscriptionJunction);
+        List<Predicate> predicates = new ArrayList<>();
+        predicates.add(cb.equal(pool.get(Pool_.OWNER).get(Owner_.ID), ownerId));
+        predicates.add(cb.ge(pool.get(Pool_.QUANTITY), 0L));
+        predicates.add(cb.gt(pool.get(Pool_.CONSUMED), pool.get(Pool_.QUANTITY)));
+
+        List<Predicate> orPredicates = new ArrayList<>();
+        for (Map.Entry<String, Entitlement> entry : subIdMap.entrySet()) {
```
Ceiu, line 917: With a large enough map, we could run into the query parameter limit. We should gate against that in some capacity, either with a check at the beginning or partitioning around the map.

Ceiu (changes requested):
First pass isn't too bad, but we have a lot of instances where we're still using Hibernate-specific APIs for query creation, and instances where we're rebuilding/parsing the entire query on every iteration during query partitioning.

src/main/java/org/candlepin/model/PoolCurator.java
bakajstep: done

src/main/java/org/candlepin/model/PoolCurator.java
bakajstep: done and we decide to split this task to 2 card to have smaller PRs for better review.

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -942,97 +901,44 @@ public List<Pool> getBySubscriptionIds(String ownerId, Collection<String> subIds
...
-    @SuppressWarnings("unchecked")
-    public List<EntitlementCertificate> retrieveEntCertsOfPoolsWithSourceEntitlement(String entId) {
-        return currentSession().createCriteria(EntitlementCertificate.class)
-            .createCriteria("entitlement")
-            .createCriteria("pool")
-            .createCriteria("sourceEntitlement")
-            .add(Restrictions.idEq(entId))
-            .list();
-    }
+        for (Map<String, Entitlement> block: this.partitionMap(subIdMap)) {
```
Ceiu, line 919: Space between the colon and block variable

Ceiu: Additionally, we need to partition the map differently. Since we have three params, two of which are blocks, our partition size needs to be the lesser of the in-block size or the parameter limit minus one, divided by two.

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -942,97 +901,44 @@ public List<Pool> getBySubscriptionIds(String ownerId, Collection<String> subIds
...
-        for (ActivationKeyPool akp : activationKeyPools) {
-            activationKeys.add(akp.getKey());
-        }
+        List<ActivationKeyPool> activationKeyPools = getEntityManager()
+            .createQuery(jpql, ActivationKeyPool.class)
+            .setParameter("pool", pool)
+            .getResultList();
 
-        return activationKeys;
+        return activationKeyPools.stream()
```
Ceiu, line 939: For this, we can go right into the stream and combine it with the previous line:

```
this.getEntityManager()
  .createQuery(jpql, ActivationKeyPool.class)
  .setParameter("pool", pool)
  .getResultStream() // return a stream instead of list
  .map(ActivationKeyPool::getKey)
  .toList();
```

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -1550,24 +1493,34 @@ public boolean exists(Pool pool) {
...
-            "(select sum(quantity) from Entitlement ent where ent.pool.id = p.id),0) " +
-            "where p.owner = :owner";
+        String jpql = """
+            UPDATE Pool p
+            SET p.consumed = coalesce(
+                (SELECT sum(quantity) FROM Entitlement ent
+                 WHERE ent.pool.id = p.id), 0)
+            WHERE p.owner = :owner""";
+
+        javax.persistence.Query query = getEntityManager().createQuery(jpql);
```
Ceiu, line 1503: With the removal of the Hibernate Query import, we can use this Query class directly.

Note this occurs in a few places

Ceiu (changes requested):

src/main/java/org/candlepin/model/PoolCurator.java
bakajstep: done

src/main/java/org/candlepin/model/PoolCurator.java
bakajstep: done

src/main/java/org/candlepin/model/PoolCurator.java
bakajstep: You can not use getResultStream() on TypedQuery/Query class, but i just picked the ActivationKey in hql

Ceiu: Oh is that method not available in the version of the JPA interface we have available? Unfortunate. I guess when we move to the versions where it's got the Jakarta namespace we can start using it.

But good call on this change. It'll pull less data from the DB to throw out, so it'll likely be more performant anyway.

src/main/java/org/candlepin/model/PoolCurator.java
```diff
@@ -942,97 +900,40 @@ public List<Pool> getBySubscriptionIds(String ownerId, Collection<String> subIds
...
-    @SuppressWarnings("unchecked")
-    public List<EntitlementCertificate> retrieveEntCertsOfPoolsWithSourceEntitlement(String entId) {
-        return currentSession().createCriteria(EntitlementCertificate.class)
-            .createCriteria("entitlement")
-            .createCriteria("pool")
-            .createCriteria("sourceEntitlement")
-            .add(Restrictions.idEq(entId))
-            .list();
-    }
+        for (Map<String, Entitlement> block : this.partitionMap(subIdMap, (getInBlockSize() - 1) / 2)) {
```
Ceiu, line 918: Close, but we literally need to check the smaller of our limits, as we don't know how each DB is configured or what hard limits they have:

```
int blockSize = Math.min(this.getQueryParameterLimit(), (this.getInBlockSize() - 1) / 2);
```

bakajstep: done

PR #4605, opened 2024-03-07
[M] CANDLEPIN-814: Addressed some refresh performance and deadlock issues

Ceiu:
- Updated entity creation during refresh to no longer flush after each JPA persist operation
- Reworked the refresh node processor to perform graph traversals in a way that result in semi-ordered entity processing, where the order is determined by graph depth, entity class, and entity ID in that order. This is done to minimize the possibility of hitting deadlock at the database level when updating indexes or foreign key references when two refreshes occur simultaneously with overlapping data sets.
- Refactored the ConstraintViolationRetryWrapper to the more generic ExpectedExceptionRetryWrapper, which now accepts one or more expection classes and will retry if any of them occur, rather than being hard-coded against a specific exception
- Refresh will now perform an internal retry if it hits a database deadlock in addition to a database constraint violation, rather than appearing to fail the whole job and retrying at the job level


Ceiu:
I've taken a look at the Sonarcloud issues and none are worth the level of effort to address them or how ugly the code becomes to satisfy Sonar.

joshmalbrecht (approved):
Looks good! Good call on the algorithm you are using in `NodeProcessor`. I'll add a tag to get additional eyes on this.

Ceiu:
I noticed I forgot to add a test to the RefreshWorker suite for verifying it'll do an internal retry on DB deadlock, so I added that.

nikosmoum (approved):

PR #4606, opened 2024-03-08
CANDLEPIN-269: Added paging to GET /content endpoint

wottop:
- Created new methods in ContentCurator listAll and getContentCount to take QueryArguments
- Updated ConsumerResource method getContents to take paging parameters and use them to call the new curator methods
- Updated candlepin-api-spec.yaml to include the paging parameters

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/ContentResourceSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/ContentResourceSpecTest.java
@@ -14,6 +14,8 @@
  */
 package org.candlepin.spec;
 
+import static java.lang.Thread.sleep;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import org.candlepin.dto.api.client.v1.AttributeDTO;
@@ -42,13 +44,18 @@ import org.candlepin.spec.bootstrap.data.util.StringUtil;
 import com.fasterxml.jackson.databind.JsonNode;
 
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+import org.junit.jupiter.api.parallel.Isolated;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.IntStream;
 
 @SpecTest
 public class ContentResourceSpecTest {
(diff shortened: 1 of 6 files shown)


src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -688,4 +699,84 @@ public Map<Content, Boolean> getActiveContentByOwner(String ownerId) {
...
+
+        List<Predicate> predicates = new ArrayList<>();
+        Predicate securityPredicate = this.getSecurityPredicate(Content.class, criteriaBuilder, root);
+        if (securityPredicate != null) {
+            predicates.add(securityPredicate);
+        }
+
+        if (!predicates.isEmpty()) {
+            criteriaQuery.where(predicates.toArray(new Predicate[predicates.size()]));
+        }
```
joshmalbrecht, line 729: Since the predicates list will be one element at max we can just check if securityPredicate is not null and call `criteriaQuery.where()` if it is not null. This avoids creating a list and then moving the element from the list to an array.

```
Predicate securityPredicate = this.getSecurityPredicate(DeletedConsumer.class, criteriaBuilder,
    root);
}
```

This can also be applied in the `getContentCount` method.

src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -688,4 +699,84 @@ public Map<Content, Boolean> getActiveContentByOwner(String ownerId) {
...
+
+    /**
+     * Fetches the count of content available.
+     *
+     * @param queryArguments
+     *     a ContentQueryArguments instance containing the various arguments
+     *
+     * @return the number of contents available
+     */
+    public long getContentCount(ContentQueryArguments queryArguments) {
```
joshmalbrecht, line 761: Need to check if `queryArguments` is null to avoid NPE

src/test/java/org/candlepin/model/ContentCuratorTest.java
```diff
@@ -832,4 +834,50 @@ public void testGetChildrenContentOfProductsByUuidsHandlesNullAndEmptyCollection
         assertNotNull(output);
         assertTrue(output.isEmpty());
     }
+
+    @Test
+    public void testListAllPaged() throws InterruptedException {
+        Content content1 = this.createContent();
```
joshmalbrecht, line 840: `content1` is an unused variable

joshmalbrecht (changes requested):
Found some small changes

src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -688,4 +699,84 @@ public Map<Content, Boolean> getActiveContentByOwner(String ownerId) {
...
+    public List<Content> listAll(ContentQueryArguments queryArgs) {
+        CriteriaBuilder criteriaBuilder = this.getEntityManager().getCriteriaBuilder();
+        CriteriaQuery<Content> criteriaQuery = criteriaBuilder.createQuery(Content.class);
+
+        Root<Content> root = criteriaQuery.from(Content.class);
+        criteriaQuery.select(root)
+            .distinct(true);
+
+        List<Predicate> predicates = new ArrayList<>();
+        Predicate securityPredicate = this.getSecurityPredicate(Content.class, criteriaBuilder, root);
```
nikosmoum, line 722: Same here: we have no need for security predicates

src/main/java/org/candlepin/resource/ContentResource.java
```diff
@@ -33,18 +40,53 @@ public class ContentResource implements ContentApi {
...
+            Page<Stream<ContentDTO>> pageResponse = new Page<>();
+            pageResponse.setPageRequest(pageRequest);
+
+            if (pageRequest.isPaging()) {
+                queryArgs.setOffset((pageRequest.getPage() - 1) * pageRequest.getPerPage())
+                    .setLimit(pageRequest.getPerPage());
+            }
+
+            if (pageRequest.getSortBy() != null) {
+                boolean reverse = pageRequest.getOrder() == PageRequest.Order.DESCENDING;
```
nikosmoum, line 70: Instead of explicitly setting the order as descending, lets use the default variable (in case we ever change the default): `PageRequest.DEFAULT_ORDER`

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/ContentResourceSpecTest.java
```diff
@@ -223,6 +230,44 @@ public void shouldFilterContentWithMismatchedArchitectureFromTheProduct() throws
...
+            // Most likely, there will already be data left from other test runs that will show up
+            IntStream.range(0, 5).forEach(entry -> {
+                adminClient.ownerContent().createContent(ownerKey, Contents.random());
+                // for timestamp separation
+                try {
+                    sleep(1000);
+                }
+                catch (InterruptedException ie) {
+                    throw new RuntimeException("Unable to sleep as expected");
+                }
```
joshmalbrecht, line 254: Can we remove the catch an throw and just keep the `throws InterruptedException` in the method signature?

Also, are the sleeps needed between each content creation or is a sleep needed before calling `adminClient.content().getContents`? If the sleep is needed before calling `adminClient.content().getContents` we could move the sleep after creating all  the content.

joshmalbrecht (changes requested):
Had a couple questions, but other than that it looks good to me!

spec-tests/src/test/java/org/candlepin/spec/ContentResourceSpecTest.java
wottop: 1. The exception is in the stream operation and the method signature won't see it. I went ahead and dropped the 'throws'
2. I need a delay in between creations to ensure that the created contents each have a unique timestamp in MySql [in seconds]

nikosmoum (approved):

spec-tests/src/test/java/org/candlepin/spec/ContentResourceSpecTest.java
joshmalbrecht: +1 Thanks for the explanation! That makes sense

PR #4608, opened 2024-03-11
CANDLEPIN-234: Added paging to GET /deleted_consumers endpoint

wottop:
- Created new method in DeletedConsumerCurator listAll with DeletedConsumerQueryArguments
- Updated DeletedConsumerResource method listByDate to take paging parameters and call the new curator method
- Updated candlepin-api-spec.yaml to include the paging parameters

### DIFF
--- a/src/main/java/org/candlepin/model/DeletedConsumerCurator.java
+++ b/src/main/java/org/candlepin/model/DeletedConsumerCurator.java
@@ -20,15 +20,17 @@ import org.candlepin.guice.PrincipalProvider;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
 
+import java.time.OffsetDateTime;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
-import javax.persistence.EntityManager;
+import javax.persistence.TypedQuery;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Order;
 import javax.persistence.criteria.Root;
 
 
@@ -39,6 +41,22 @@ import javax.persistence.criteria.Root;
 public class DeletedConsumerCurator extends AbstractHibernateCurator<DeletedConsumer> {
 
     private PrincipalProvider principalProvider;
+
+    /**
+     * Container object for providing various arguments to the deleted consumer lookup method(s).
+     */
+    public static class DeletedConsumerQueryArguments extends QueryArguments<DeletedConsumerQueryArguments> {
+        private OffsetDateTime date;
+
+        public DeletedConsumerCurator.DeletedConsumerQueryArguments setDate(OffsetDateTime date) {
+            this.date = date;
+            return this;
+        }
+
+        public OffsetDateTime getDate() {
+            return this.date;
+        }
+    }
 
     @Inject
     public DeletedConsumerCurator(PrincipalProvider principalProvider) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -103,19 +123,93 @@ public int createDeletedConsumers(Collection<String> consumerIds) {
...
-        criteriaQuery.where(criteriaBuilder.greaterThanOrEqualTo(
-            consumer.get(DeletedConsumer_.CREATED), date));
-        criteriaQuery.orderBy(criteriaBuilder.desc(consumer.get(DeletedConsumer_.CREATED)));
+        Root<DeletedConsumer> root = criteriaQuery.from(DeletedConsumer.class);
+        criteriaQuery.select(root)
+            .distinct(true);
+        if (queryArgs.getDate() != null) {
+            criteriaQuery.where(criteriaBuilder.greaterThanOrEqualTo(
+                root.get(DeletedConsumer_.CREATED),
+                new Date(queryArgs.getDate().toInstant().toEpochMilli())));
```
joshmalbrecht, line 146: An alternative here would be `Date.from(queryArgs.getDate().toInstant())`

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -103,19 +123,93 @@ public int createDeletedConsumers(Collection<String> consumerIds) {
...
+        List<Predicate> predicates = new ArrayList<>();
+        Predicate securityPredicate = this.getSecurityPredicate(DeletedConsumer.class, criteriaBuilder,
+            root);
+        if (securityPredicate != null) {
+            predicates.add(securityPredicate);
+        }
+
+        if (!predicates.isEmpty()) {
+            criteriaQuery.where(predicates.toArray(new Predicate[predicates.size()]));
+        }
```
joshmalbrecht, line 158: Since the predicates list will be one element at max we can just check if securityPredicate is not null and call `criteriaQuery` if it is not null. This avoids creating a list and then moving the element from the list to an array.

```
Predicate securityPredicate = this.getSecurityPredicate(DeletedConsumer.class, criteriaBuilder,
    root);
}
```

This can also be applied in the `getDeletedConsumerCount` method.

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -103,19 +123,93 @@ public int createDeletedConsumers(Collection<String> consumerIds) {
...
+
+    /**
+     * Fetches the count of deleted consumers available.
+     *
+     * @param queryArgs
+     *     a DeletedConsumerQueryArguments instance containing the various arguments
+     *
+     * @return the number of deleted consumers available
+     */
+    public long getDeletedConsumerCount(DeletedConsumerQueryArguments queryArgs) {
```
joshmalbrecht, line 189: We should check for a null `queryArgs` to avoid NPE

joshmalbrecht (changes requested):
Found a couple changes that can be made

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -103,19 +123,93 @@ public int createDeletedConsumers(Collection<String> consumerIds) {
...
+        Root<DeletedConsumer> root = criteriaQuery.from(DeletedConsumer.class);
+        criteriaQuery.select(criteriaBuilder.countDistinct(root));
+        if (queryArgs.getDate() != null) {
+            criteriaQuery.where(criteriaBuilder.greaterThanOrEqualTo(
+                root.get(DeletedConsumer_.CREATED),
+                new Date(queryArgs.getDate().toInstant().toEpochMilli())));
+        }
+
+        List<Predicate> predicates = new ArrayList<>();
+        Predicate securityPredicate = this.getSecurityPredicate(DeletedConsumer.class, criteriaBuilder, root);
```
nikosmoum, line 202: Same here: no need for security predicate

src/main/java/org/candlepin/resource/DeletedConsumerResource.java
```diff
@@ -33,21 +40,56 @@
...
+            Page<Stream<DeletedConsumerDTO>> pageResponse = new Page<>();
+            pageResponse.setPageRequest(pageRequest);
+
+            if (pageRequest.isPaging()) {
+                queryArgs.setOffset((pageRequest.getPage() - 1) * pageRequest.getPerPage())
+                    .setLimit(pageRequest.getPerPage());
+            }
+
+            if (pageRequest.getSortBy() != null) {
+                boolean reverse = pageRequest.getOrder() == PageRequest.Order.DESCENDING;
```
nikosmoum, line 73: Instead of explicitly setting the order as descending, lets use the default variable (in case we ever change the default): PageRequest.DEFAULT_ORDER

nikosmoum (changes requested):

PR #4616, opened 2024-03-14
[M] CANDLEPIN-828: Added owner key in OWNER_CONTENT_ACCESS_MODE event

joshmalbrecht:
- Replaced the ownerId field with ownerKey field for all Owned objects
- Corrected serialization issue for the Event.eventData field where the field was being serialized twice
- Removed ActivationListenerObjectMapper as it is no longer being used
- Removed EventFactoryObjectMapper as it is no longer being used

The event that was being published to the broker before the changes:

```xml
   <message id="100" priority="4" expiration="0" timestamp="1710439677286" type="text">
      <properties>
         <property name="_AMQ_ROUTING_TYPE" value="1" type="byte"/>
         <property name="EVENT_TARGET" value="OWNER_CONTENT_ACCESS_MODE" type="simple-string"/>
         <property name="EVENT_TYPE" value="MODIFIED" type="simple-string"/>
      </properties>
      <body>
         <![CDATA[{"id":"d4713ca9-0324-42bf-a508-a0cc82ad5f42","type":"MODIFIED","target":"OWNER_CONTENT_ACCESS_MODE","targetName":"Test Owner 995FC2D8","timestamp":1710439677268,"entityId":"ff8081818e3e24e2018e3e263bab0002","ownerId":"ff8081818e3e24e2018e3e263bab0002","consumerUuid":null,"referenceId":null,"referenceType":null,"eventData":"{\"contentAccessMode\":\"org_environment\"}","messageText":null,"principal":{"type":"user","name":"admin"}}]]>
      </body>
   </message>
```

The message that is published to the broker after the changes:

```xml
   <message id="131" priority="4" expiration="0" timestamp="1710438704468" type="text">
      <properties>
         <property name="_AMQ_ROUTING_TYPE" value="1" type="byte"/>
         <property name="EVENT_TARGET" value="OWNER_CONTENT_ACCESS_MODE" type="simple-string"/>
         <property name="EVENT_TYPE" value="MODIFIED" type="simple-string"/>
      </properties>
      <body>
         <![CDATA[{"id":"ea9c3055-1fac-4074-8d4b-981b3ea79049","type":"MODIFIED","target":"OWNER_CONTENT_ACCESS_MODE","targetName":"Test Owner E3EF1037","timestamp":1710438704461,"entityId":"ff8081818e3e14f2018e3e1763bd0002","ownerKey":"test_owner-E3EF1037","consumerUuid":null,"referenceId":null,"referenceType":null,"eventData":{"contentAccessMode":"org_environment"},"messageText":null,"principal":{"type":"user","name":"admin"}}]]>
      </body>
   </message>
```

After these changes are made, I will create a PR for candlepin-4.4.1-HOTFIX and cherry pick this commit to that PR.


src/main/java/org/candlepin/audit/EventFactory.java
```diff
@@ -206,19 +197,13 @@ public Event complianceCreated(Consumer consumer, ComplianceStatus compliance) {
...
-        catch (JsonProcessingException e) {
-            log.error("Error while building JSON for compliance.created event.", e);
-            throw new IseException("Error while building JSON for compliance.created event.", e);
-        }
+
+        // Instead of an internal db id, compliance.created events now use
+        // UUID for the 'consumerId' and 'entityId' fields, since Katello
+        // is concerned only with the consumer UUID field.
+        return new Event(Event.Type.CREATED, Event.Target.COMPLIANCE,
+            consumer.getName(), principalProvider.get(), consumer.getOwnerId(), consumer.getUuid(),
```
nikosmoum, line 205: we should change `consumer.getOwnerId` to `consumer.getOwnerKey` here

src/main/java/org/candlepin/audit/EventFactory.java
```diff
@@ -243,20 +228,12 @@ public Event complianceCreated(Consumer consumer, SystemPurposeComplianceStatus
...
-        catch (JsonProcessingException e) {
-            log.error("Error while building JSON for system purpose compliance.created event.", e);
-            throw new IseException("Error while building JSON for system purpose compliance.created event.",
-                e);
-        }
+        // Instead of an internal db id, compliance.created events now use
+        // UUID for the 'consumerId' and 'entityId' fields, since Katello
+        // is concerned only with the consumer UUID field.
+        return new Event(Event.Type.CREATED, Target.SYSTEM_PURPOSE_COMPLIANCE,
+            consumer.getName(), principalProvider.get(), consumer.getOwnerId(), consumer.getUuid(),
```
nikosmoum, line 235: same here: `consumer.getOwnerKey`

nikosmoum (changes requested):
Just one thing to change; otherwise it looks good

src/main/java/org/candlepin/audit/EventBuilder.java
```diff
@@ -83,14 +80,10 @@ public EventBuilder setEventData(Eventful entity) {
...
 
-            // TODO: FIXME:
-            // This is entirely useless. Our owner APIs expect the owner key, not the ID, so
-            // providing the ID does very little for event consumers.
             if (entity instanceof Owned) {
-                String ownerId = ((Owned) entity).getOwnerId();
-
-                if (ownerId != null) {
-                    event.setOwnerId(ownerId);
+                String ownerKey = ((Owned) entity).getOwnerKey();
```
Januson, line 84: You can use pattern matching instanceOf to avoid manual casts.

src/main/java/org/candlepin/audit/EventFactory.java
joshmalbrecht: Good catch!! Updated

src/main/java/org/candlepin/audit/EventFactory.java
joshmalbrecht: Good catch!! Updated

src/main/java/org/candlepin/audit/EventBuilder.java
joshmalbrecht: Good catch! Updated

src/main/java/org/candlepin/audit/Event.java
```diff
@@ -104,15 +105,16 @@ public enum ReferenceType {
...
+    private Map<String, Object> eventData;
 
     private String messageText;
 
     public Event() {
     }
 
-    public Event(Type type, Target target, String targetName, Principal principal, String ownerId,
-        String consumerUuid, String entityId, String eventData, String referenceId,
+    public Event(Type type, Target target, String targetName,
```
Ceiu, line 115: From Sonarcloud: this constructor is absurd. Since we're in this area, let's take the time to remove it entirely in favor of chainable mutators.

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -494,6 +494,14 @@ public String getOwnerId() {
         return ownerId;
     }
 
+    /**
+     * @return the owner key of this Consumer.
+     */
+    @Override
+    public String getOwnerKey() {
+        return this.owner == null ? null : this.owner.getOwnerKey();
```
Ceiu, line 502: Don't access the owner directly here, lest we run into the Hibernate caching issue where our local state and the proxy state differ and our results are undefined.

Call `getOwner` and store the result, then we can use this same ternary operator on the output:

```
Owner owner = this.getOwner();
return owner != null ? owner.getKey() : null;
```

src/main/java/org/candlepin/dto/manifest/v1/SubscriptionDTO.java
```diff
@@ -125,6 +125,15 @@ public String getOwnerId() {
         return this.getOwner() != null ? this.getOwner().getId() : null;
     }
 
+    /**
+     * @return the key to the owner associated with this subscription
+     */
+    @Override
```
Ceiu, line 131: For methods we override, we can save some documentation pain/overhead by just inheriting the docs:

```
/**
 * {@inheritDoc}
 */
```

Unless our implementation does something weird or special, we shouldn't bother rewriting the docs and having to keep multiple places in sync.

(Note: this occurs in several places)

src/main/java/org/candlepin/model/Entitlement.java
```diff
@@ -166,6 +166,14 @@ public String getOwnerId() {
         return (owner == null) ? null : owner.getId();
     }
 
+    /**
+     * @return the owner key of this Entitlement.
+     */
+    @Override
+    public String getOwnerKey() {
+        return this.owner == null ? null : this.owner.getOwnerKey();
```
Ceiu, line 174: Same deal here; make sure we don't access Hibernate-managed fields directly in the utility methods of our entities.

(note this occurs in more entity classes, but I'm not going to comment on all of them)

Ceiu (changes requested):
Few things to update surrounding maintenance and avoiding Hibernate pain. Sonarcloud also has a handful of issues that are reasonable that we could choose to address.

I'm not entirely sure if we should get rid of the private log references in classes where we aren't logging anything, since we'll probably end up adding it back in again later; but it's not strictly *wrong*, either.

src/main/java/org/candlepin/audit/Event.java
Ceiu: Note that this will require updating the mutators to return the Event object, and we may still need a constructor that takes some immutable fields (if those exist). Ideally we can drop to just a default and maybe one non-default constructor if our hand is forced.

src/main/java/org/candlepin/model/Consumer.java
joshmalbrecht: Good catch!! I definitely overlooked that. Updated throughout

src/main/java/org/candlepin/dto/manifest/v1/SubscriptionDTO.java
joshmalbrecht: Updated throughout

src/main/java/org/candlepin/model/Entitlement.java
joshmalbrecht: Updated throughout

src/main/java/org/candlepin/audit/Event.java
joshmalbrecht: I had to keep the default constructor with no parameters for Jackson serialization. However, I added one non-default constructor with parameters for the non-null required fields

src/main/java/org/candlepin/model/GuestId.java
```diff
@@ -174,6 +174,14 @@ public String getOwnerId() {
         return null;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getOwnerKey() {
+        return this.consumer == null ? null : this.consumer.getOwnerKey();
```
Ceiu, line 182: Missed a direct field access here

joshmalbrecht: Oops, my bad! Updated

Ceiu (approved):

PR #4630, opened 2024-03-19
[M] CANDLEPIN-819: Added max page size option for request paging

Ceiu:
- Added a new option, candlepin.paging.max_page_size, for controlling the maximum size of a given page request; defaults to 3000
- Updated the PageRequestFilter to throw a BadRequestException if a page size is specified larger than the configured max page size

### DIFF
--- a/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -293,6 +293,10 @@ public class ConfigProperties {
     // How long (in seconds) to wait for job threads to finish during a graceful Tomcat shutdown
     public static final String ASYNC_JOBS_THREAD_SHUTDOWN_TIMEOUT = "candlepin.async.thread.shutdown.timeout";
 
+    // The maximum allowable size for pages; defaults to 3000
+    public static final String PAGING_MAX_PAGE_SIZE = "candlepin.paging.max_page_size";
+
+
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -293,6 +293,10 @@ private ConfigProperties() {
     // How long (in seconds) to wait for job threads to finish during a graceful Tomcat shutdown
     public static final String ASYNC_JOBS_THREAD_SHUTDOWN_TIMEOUT = "candlepin.async.thread.shutdown.timeout";
 
+    // The maximum allowable size for pages; defaults to 3000
+    public static final String PAGING_MAX_PAGE_SIZE = "candlepin.paging.max_page_size";
```
nikosmoum, line 297: There's already a config property like this, the suggestion was to update its default value

nikosmoum (commented):
this change should go in the feature/candlepinquery_removal feature branch

src/main/java/org/candlepin/config/ConfigProperties.java
Ceiu: Oh good, I thought it was kinda odd I didn't see them when I searched for them. I must have typo'd "paging" while looking for the existing ones.

Fixed.

Ceiu:
@nikosmoum Going to close this PR and reopen it with a new branch name against the target branch.

PR #4631, opened 2024-03-20
[F] CANDLEPIN-819: Added enforcement of the max page size configuration

Ceiu:
- Updated the PageRequestFilter to throw a BadRequestException if a page size is specified larger than the configured max page size
- Increased the default max page size from 1000 to 3000
- Renamed the PAGING_SIZE and MAX_PAGING_SIZE constants to PAGING_DEFAULT_PAGE_SIZE and PAGING_MAX_PAGE_SIZE respectively


wottop:
lgtm

PR #4635, opened 2024-03-21
[M] CANDLEPIN-827: Removed IllegalStateException when owner does not exist

joshmalbrecht:
- Updated the PermissionFactory to return an empty list of permissions when there is a cache miss to retrieve an owner and the owner does not yet exist in the database instead of throwing an IllegalStateException.

### DIFF
--- a/src/main/java/org/candlepin/auth/permissions/PermissionFactory.java
+++ b/src/main/java/org/candlepin/auth/permissions/PermissionFactory.java
@@ -80,9 +80,8 @@ public class PermissionFactory {
         }
 
         public Owner resolve(OwnerInfo oinfo) {
-            if (oinfo == null || oinfo instanceof Owner) {
-                // nothing to resolve, it's already as good as it's going to get
-                return (Owner) oinfo;
+            if (oinfo == null) {
+                return null;
             }
 
             String ownerKey = oinfo.getKey();
@@ -94,7 +93,7 @@ public class PermissionFactory {
             if (cached == null) {
                 cached = this.ownerCurator.getByKey(ownerKey);
                 if (cached == null) {
-                    throw new IllegalStateException("No such owner: " + ownerKey);
+                    return null;
                 }
 
                 // Impl note:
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/auth/permissions/PermissionFactory.java
```diff
@@ -136,26 +136,59 @@ public PermissionFactory(OwnerCurator ownerCurator) {
     @SuppressWarnings("checkstyle:indentation")
     private Map<String, PermissionBuilder> initBuilders() {
         Map<String, PermissionBuilder> builders = new HashMap<>();
+        builders.put(PermissionType.OWNER.name(), (user, bp, res) -> {
+            Owner owner = res.resolve(bp.getOwner());
+            if (owner == null) {
```
Ceiu, line 141: nit: For cases like this where we're doing an either/or thing, we can flip the operator and use a ternary op to compress the code down to only a few lines and save a few nanoseconds of execution:

```
return owner != null ?
  List.of(new OwnerPermission(...)) :
  List.of();
```

src/test/java/org/candlepin/auth/permissions/PermissionFactoryTest.java
```diff
@@ -0,0 +1,409 @@
...
+
+        List<Permission> actual = new PermissionFactory(ownerCurator)
+            .createPermissions(userInfo);
+
+        assertThat(actual)
+            .isNotNull()
+            .isEmpty();
+    }
+
+    public class TestOwner implements OwnerInfo {
```
Ceiu, line 246: What do we get out of these classes that we're not getting from the concrete implementations or mocks? I don't see any logic in any of the methods.

Ceiu (commented):
Good to see some tests for the permission factory

Couple questions on the implementation though

src/main/java/org/candlepin/auth/permissions/PermissionFactory.java
joshmalbrecht: Good call! Updated throughout

src/test/java/org/candlepin/auth/permissions/PermissionFactoryTest.java
joshmalbrecht: The problem with using the concrete implementations is that the `PermissionFactory.OwnerResolver` will always return that instance of `Owner` and we won't be able to test the actual case of `PermissionFactory.OwnerResolver.resolve` returning null.

https://github.com/candlepin/candlepin/blob/main/src/main/java/org/candlepin/auth/permissions/PermissionFactory.java#L83-L86

Ceiu: I'm not following. Can you point to one of the tests where things are broken when an existing concrete implementation is used? I glanced through them again and it's not popping out to me.

joshmalbrecht: If you look at `PermissionFactoryTest.testCreatePermissionsWithNoExistingOwner` and used the concrete implementations, so populated a `User` object with a `Role` that has a `PermissionBlueprint` object that then has an `Owner` object on it. Then called `permissionFactory.createPermissions` using that `User`, the PermissionFactory.OwnerResolver.resolve would return that `Owner` object that is on the `PermissionBlueprint` instead of checking the cache or the DB.

joshmalbrecht: So we wouldn't be actually testing the case of the Owner not existing in the cache or DB

joshmalbrecht: Updated!

Ceiu (approved):

PR #4636, opened 2024-03-22
[M] CANDLEPIN-829: Updated consumer activation key registration testing

Ceiu:
- Updated the concurrent activation key registration testing to no longer need the external awaitility library, nor need multiple runs to trigger deadlock when it's present

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceActivationKeySpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceActivationKeySpecTest.java
@@ -15,23 +15,22 @@
 package org.candlepin.spec.consumers;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.awaitility.Awaitility.await;
 import static org.candlepin.spec.bootstrap.assertions.StatusCodeAssertions.assertBadRequest;
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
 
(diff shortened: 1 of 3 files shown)


spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceActivationKeySpecTest.java
```diff
@@ -103,144 +101,95 @@ public void shouldAllowConsumerToRegisterWithActivationKeys() {
...
-            while (thread.isAlive()) {
-                thread.join();
-            }
+        // We can use a set seed if this test becomes inconsistent in the future, but so far the order just
+        // needs to be non-sequential and every iteration attempted has triggered the deadlock as expected
+        // long seed = -1753978213090389925L;
+        long seed = rnd.nextLong();
+        rnd.setSeed(seed);
+
+        System.out.printf("Shuffling pools using seed %d\n", seed);
```
joshmalbrecht, line 133: This might have been left from debugging

joshmalbrecht (changes requested):
Looks good! I just noticed a print statement that might have been left from debugging

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceActivationKeySpecTest.java
Ceiu: Kind of, but I left it intentionally in the event we wanted to grab the seed for a given run. I suppose we can always put it back when need be.

joshmalbrecht: Okay! I'm fine with leaving it in

joshmalbrecht (approved):

PR #4642, opened 2024-03-26
[M] CANDLEPIN-755: Updated org_migrator to reflect layering schema

Ceiu:
- Updated the org migrator utility to reflect schema changes made for the entity layering model changes


Ceiu:
Manual testing can be performed by using the hosted test infrastructure to import some prod-like data, perform a refresh, and then export using the updated org migrator.

From there, redeploy with a clean database, then import and verify all of the data is as expected. This can be done with both mysql/mariadb and postgresql

joshmalbrecht (approved):

PR #4645, opened 2024-03-29
[M] CANDLEPIN-832: Revoke all entitlements when owner changes to SCA mode

joshmalbrecht:
- Created RevokeEntitlementsJob to revoke all entitlements for an owner and remove all activation key pools for an owner's activation keys.
- Created a RevokeEntitlementJob for an owner that changes from entitlement mode to SCA mode.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
@@ -24,6 +24,7 @@ import static org.candlepin.spec.bootstrap.assertions.StatusCodeAssertions.asser
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
+import org.candlepin.dto.api.client.v1.ActivationKeyDTO;
 import org.candlepin.dto.api.client.v1.AsyncJobStatusDTO;
 import org.candlepin.dto.api.client.v1.AttributeDTO;
 import org.candlepin.dto.api.client.v1.CertificateDTO;
@@ -49,6 +50,7 @@ import org.candlepin.spec.bootstrap.client.ApiClients;
 import org.candlepin.spec.bootstrap.client.SpecTest;
 import org.candlepin.spec.bootstrap.client.api.OwnerClient;
 import org.candlepin.spec.bootstrap.client.api.Paging;
+import org.candlepin.spec.bootstrap.data.builder.ActivationKeys;
 import org.candlepin.spec.bootstrap.data.builder.ConsumerTypes;
 import org.candlepin.spec.bootstrap.data.builder.Consumers;
 import org.candlepin.spec.bootstrap.data.builder.Owners;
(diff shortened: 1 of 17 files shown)


src/main/java/org/candlepin/async/tasks/RevokeEntitlementsJob.java
```diff
@@ -0,0 +1,169 @@
...
+        if (owner == null) {
+            log.info("Unable to find owner for owner key \"{}\"", ownerKey);
+            context.setJobResult("Unable to find owner for owner key \"%s\"", ownerKey);
+            return;
+        }
+
+        int batchSize = getBatchSize();
+
+        int removedPools = activationKeyCurator.removeActivationKeyPools(ownerKey);
+        log.debug("Removed {} activation key pools", removedPools);
```
joshmalbrecht, line 90: I'm wondering if this really should be its own job so that each job has one concern and can possibly be used for other situations other than an owner changing from entitlement mode to SCA mode.

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -247,6 +259,19 @@ public Owner updateOwnerContentAccess(Owner owner, String updatedList, String up
                 this.contentAccessCertificateCurator.deleteForOwner(owner);
             }
 
+            // Revoke entitlements and remove activation key pools if moving from entitlement mode to SCA mode
+            if (this.isTransition(currentMode, updatedMode, ContentAccessMode.ENTITLEMENT,
+                ContentAccessMode.ORG_ENVIRONMENT)) {
+                try {
+                    RevokeEntitlementsJobConfig jobConfig = new RevokeEntitlementsJobConfig();
+                    jobConfig.setJobArgument(RevokeEntitlementsJob.OWNER_KEY, owner.getKey());
+                    jobManager.queueJob(jobConfig);
```
Ceiu, line 268: I'm not sure how to feel about the job coming into existence here but we have no means of tracking its state. I don't know that we have a more reliable way to do it, and even at the owner update endpoint, we can't return the job status without omitting the owner output, so this may be our lot anyway.

Are we okay with the job blinking into and out of existence transparently like this? @nikosmoum

Ceiu (changes requested):
Couple minor change requests, but a few questions to mull over here.

src/main/java/org/candlepin/async/tasks/RevokeEntitlementsJob.java
```diff
@@ -0,0 +1,169 @@
...
+        this.ownerCurator = Objects.requireNonNull(ownerCurator);
+        this.poolService = Objects.requireNonNull(poolService);
+        this.activationKeyCurator = Objects.requireNonNull(activationKeyCurator);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void execute(JobExecutionContext context) throws JobExecutionException {
```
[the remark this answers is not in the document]
joshmalbrecht: Yeah, I was wanting to a transaction for each individual batch. I added  a comment to help make that more clear!

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -247,6 +259,19 @@ public Owner updateOwnerContentAccess(Owner owner, String updatedList, String up
                 this.contentAccessCertificateCurator.deleteForOwner(owner);
             }
 
+            // Revoke entitlements and remove activation key pools if moving from entitlement mode to SCA mode
+            if (this.isTransition(currentMode, updatedMode, ContentAccessMode.ENTITLEMENT,
```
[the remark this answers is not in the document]
joshmalbrecht: That makes sense! I updated to use the `isTransitioningFrom` method

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1848,4 +1846,27 @@ public Collection<String> lockAndLoadIds(Collection<? extends Serializable> ids)
...
+     * @return all non-manifest type consumer UUIDs for the provided owner. This method will not return null.
+     */
+    public List<String> getSystemConsumerUuidsByOwner(String ownerKey) {
+        List<String> consumerUuids  = new ArrayList<>();
+        if (ownerKey == null || ownerKey.isBlank()) {
+            return consumerUuids;
+        }
+
+        String jpql = "SELECT c.uuid FROM Consumer c " +
+            "JOIN ConsumerType type ON type.id=c.typeId " +
```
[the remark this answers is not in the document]
joshmalbrecht: Good catch! Updated

src/main/java/org/candlepin/async/tasks/RevokeEntitlementsJob.java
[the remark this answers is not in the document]
Ceiu: I'm not seeing the comment anywhere. Is it waiting for the next push?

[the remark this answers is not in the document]
joshmalbrecht: Omg haha I didn't save the file in my IDE. Thanks for catching that!

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
```diff
@@ -901,6 +903,71 @@ public void shouldMigrateAllConsumersOfAnonymousOwnerToTargetOwner() {
...
+            .isNotNull()
+            .singleElement()
+            .returns(pool2.getId(), PoolDTO::getId);
+
+        // Update owner to SCA mode
+        owner.contentAccessMode(Owners.SCA_ACCESS_MODE);
+        owner = admin.owners().updateOwner(owner.getKey(), owner);
+
+        // Wait for the entitlement revoke job to finish
+        Thread.sleep(5000);
```
nikosmoum, line 952: instead of sleeping, you can use the listJobStatuses API with arguments being the owner key and the job key, which should return that single job status, then you can wait until the job is done, and assert on its end state. We have other spec tests (JobStatusSpecTest/CloudRegistrationSpecTest) that do this kind of thing

src/main/java/org/candlepin/async/tasks/RevokeEntitlementsJob.java
```diff
@@ -0,0 +1,173 @@
...
+         *
+         * @param ownerKey
+         *  the owner key to set
+         *
+         * @throws IllegalArgumentException
+         *  if the owner key is null or blank
+         *
+         * @return a reference to this RevokeEntitlementsJobConfig
+         */
+        public RevokeEntitlementsJobConfig setOwnerKey(String ownerKey) {
```
nikosmoum, line 133: Lets keep consistent with other owner-scoped jobs(e.g. UndoImportsJob, RefreshPoolsJob), and use the `setOwner(Owner owner)` signature here instead

src/main/java/org/candlepin/async/tasks/RevokeEntitlementsJob.java
```diff
@@ -0,0 +1,173 @@
...
+            if (ownerKey == null || ownerKey.isBlank()) {
+                throw new IllegalArgumentException("owner key is null or blank");
+            }
+
+            this.setJobArgument(OWNER_KEY, ownerKey);
```
nikosmoum, line 138: in addition to adding the owner key as an argument here, we want to call `setContextOwner(owner)` on the JobConfig here (see other owner-scoped jobs for examples). that will allow us to to query for job statuses using the owner key

src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -287,7 +288,8 @@ private ConfigProperties() {
         JobCleaner.JOB_KEY,
         ManifestCleanerJob.JOB_KEY,
         UnmappedGuestEntitlementCleanerJob.JOB_KEY,
-        InactiveConsumerCleanerJob.JOB_KEY
+        InactiveConsumerCleanerJob.JOB_KEY,
+        RevokeEntitlementsJob.JOB_KEY
```
nikosmoum, line 292: This job shouldn't be on this list, since its scoped to an owner, while the list should only contain scheduled/global jobs instead

nikosmoum (changes requested):

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
Ceiu: On that note, there are some newer methods on the job assertion for this which we should be sure to use. Some examples of the newer logic is in the import spec tests, but as an example:

```
  AsyncJobStatusDTO importJob1 = this.importAsync(owner, manifest);
  assertThatJob(importJob1)
      .isFinished()
      .contains("SUCCESS");
```

src/main/java/org/candlepin/config/ConfigProperties.java
joshmalbrecht: Removed

src/main/java/org/candlepin/async/tasks/RevokeEntitlementsJob.java
joshmalbrecht: Updated

src/main/java/org/candlepin/async/tasks/RevokeEntitlementsJob.java
joshmalbrecht: Updated

spec-tests/src/test/java/org/candlepin/spec/owners/OwnerResourceSpecTest.java
joshmalbrecht: Updated

nikosmoum: We could also assert on the result of the job here. The result string should contain the amount of entitlements revoked and pools removed from activation keys

joshmalbrecht: Updated

PR #4647, opened 2024-03-29
[F] CANDLEPIN-839: Updated schema to support environment content overrides

Ceiu:
- Updated the database schema to support environment content overrides
- Added indexes to the consumer_id and the new environment_id
  columns to improve content override lookup performance
- Added a constraint to the content override table to help prevent
  creating overrides in invalid states
- Removed legacy constructors for content overrides and the various
  subclasses
- Removed the allowance of null values on the content override
  subclasses which add a "parent" entity reference
- Removed old XML annotations on the content override entities

### DIFF
--- a/src/main/java/org/candlepin/model/EnvironmentContentOverride.java
+++ b/src/main/java/org/candlepin/model/EnvironmentContentOverride.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2009 - 2024 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.model;
+
+import javax.persistence.DiscriminatorValue;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.validation.constraints.NotNull;
+
+
+
+/**
+ * A content override applied to an environment
+ */
+@Entity
+@DiscriminatorValue("environment")
+public class EnvironmentContentOverride extends ContentOverride<EnvironmentContentOverride, Environment> {
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(nullable = false)
+    @NotNull
+    private Environment environment;
+
+    /**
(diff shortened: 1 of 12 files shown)


Ceiu:
Whoops, forgot about adding tests for the new entity. I'll go back and do that on Monday.

src/main/resources/db/changelog/20240329143555-add_environment_content_override_schema.xml
```diff
@@ -0,0 +1,91 @@
...
+            Scrub the data to ensure we don't have malformed content overrides with both an AK and
+            a consumer set. We'll follow the design's outlined priority and clear the consumer on
+            any row that has a non-null key ID.
+
+            This way we need not worry (as much) about the constraint we'll be adding in the next
+            task.
+        -->
+
+        <sql dbms="mariadb, postgresql">
+            UPDATE cp_content_override SET consumer_id = NULL WHERE key_id IS NOT NULL
```
Ceiu, line 41: Looking at this during self-review and I'm beginning to wonder if it'd be better/safer to clear the consumer instead. Both options kinda suck here for the same reason:

- If we clear the activation key, existing consumers that are currently getting the affected override will continue getting it, but new consumers registering with the activation key will no longer inherit the override.
- If we clear the consumer as we do here, existing consumers that may have registered with the activation key to get the override will no longer get it, but new consumers registering with it will.

In either case we end up in a weird state desync that will be confusing. But failing to do either will cause the addition of the check constraint to fail.

joshmalbrecht (changes requested):
Looks good! I just noticed a wrong year on one of the Copyright headers.

src/main/java/org/candlepin/model/EnvironmentContentOverride.java
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright (c) 2009 - 2023 Red Hat, Inc.
```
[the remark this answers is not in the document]
Ceiu: Fixed

joshmalbrecht (approved):

src/main/resources/db/changelog/20240329143555-add_environment_content_override_schema.xml
nikosmoum: are there actually any such data in prod?

nikosmoum: I just run this query and it seems there aren't (in hosted at least)

Ceiu: In prod, no; but we don't know what the many DBs of Sat installs look like, though. And if there are none, we're in a great place. The concern is that for some reason we never required the parent object on our content override subtypes, nor did we have a constraint preventing two parents, so it's very possible something got busted.

nikosmoum (approved):
Looks good. The malformed data update shouldn't affect anyone (I'm not even sure how that kind of data would be formed?).
The missing test coverage seems very minor, so we can ignore it

src/main/resources/db/changelog/20240329143555-add_environment_content_override_schema.xml
nikosmoum: I can't see how this kind of malformed data would appear through our APIs

PR #4656, opened 2024-04-04
[M] CANDLEPIN-845: Added retry logic for database connection failures

joshmalbrecht:
- Moved database connection and migration logic from
      CandlepinContextListener to a new DatabaseConnectionManager class and
      a new MigrationManager class.
    - Added database connection retry logic for when Candlepin is starting
      and fails to connect to the database.
    - Added max retry attempts and retry interval configurations to be used
      for the database connection retry logic.
    - Removed documentation in CONTRIBUTING.md regarding needing to
      restart the candlepin container when using podman play kube command.

### DIFF
--- a/src/main/java/org/candlepin/config/ConfigProperties.java
+++ b/src/main/java/org/candlepin/config/ConfigProperties.java
@@ -28,7 +28,7 @@ import org.candlepin.async.tasks.RevokeEntitlementsJob;
 import org.candlepin.async.tasks.UnmappedGuestEntitlementCleanerJob;
 import org.candlepin.config.validation.ConfigurationValidator;
 import org.candlepin.config.validation.IntegerConfigurationValidator;
-import org.candlepin.guice.CandlepinContextListener;
+import org.candlepin.database.MigrationManagementLevel;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -179,6 +179,10 @@ public class ConfigProperties {
     public static final String DB_URL = JPA_CONFIG_PREFIX + "hibernate.connection.url";
     public static final String DB_USERNAME = JPA_CONFIG_PREFIX + "hibernate.connection.username";
     public static final String DB_PASSWORD = JPA_CONFIG_PREFIX + "hibernate.connection.password";
+
+    // Database
+    public static final String DB_MAX_CONNECTION_ATTEMPTS = "candlepin.db.max_connection_attempts";
+    public static final String DB_CONNECTION_RETRY_INTERVAL = "candlepin.db.retry_interval";
 
     // Cache
     public static final String CACHE_JMX_STATS = "cache.jmx.statistics";
@@ -350,6 +354,10 @@ public class ConfigProperties {
         {
             this.put(CANDLEPIN_URL, "https://localhost");
 
+            // Database connection
+            this.put(DB_CONNECTION_RETRY_INTERVAL, "5"); // seconds
+            this.put(DB_MAX_CONNECTION_ATTEMPTS, "3");
+
             this.put(JWT_ISSUER, "Candlepin");
             this.put(JWT_TOKEN_TTL, "600"); // seconds
             this.put(ANON_JWT_TOKEN_TTL, "172800"); // seconds
@@ -415,7 +423,7 @@ public class ConfigProperties {
             // submit one when registering.
             this.put(HIDDEN_RESOURCES, "environments");
             this.put(HIDDEN_CAPABILITIES, "");
-            this.put(DB_MANAGE_ON_START, CandlepinContextListener.DBManagementLevel.NONE.getName());
+            this.put(DB_MANAGE_ON_START, MigrationManagementLevel.NONE.getName());
             this.put(SSL_VERIFY, "false");
 
             this.put(FAIL_ON_UNKNOWN_IMPORT_PROPERTIES, "false");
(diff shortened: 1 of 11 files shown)


Ceiu:
I want to take a closer look at this, but as the CCL approaches something like 30-40% of the logic being the DB handling stuff, we should really start looking at pulling that out into its own class to encapsulate the logic a little better, and stop using CCL as a dumping grounds for misc initialization code.

As a slight aside to something from a PR I missed earlier, and an artifact of this type of issue, the configuration validation happens as a separate initialization step rather than as a part of the logic that loads the configuration. Is there a reason for that I'm not seeing?

joshmalbrecht:
> I want to take a closer look at this, but as the CCL approaches something like 30-40% of the logic being the DB handling stuff, we should really start looking at pulling that out into its own class to encapsulate the logic a little better, and stop using CCL as a dumping grounds for misc initialization code.
> 
> As a slight aside to something from a PR I missed earlier, and an artifact of this type of issue, the configuration validation happens as a separate initialization step rather than as a part of the logic that loads the configuration. Is there a reason for that I'm not seeing?

Yeah, we really should have the DB connection logic in like a `DatabaseConnectionManager` class like you said.

I introduced the org/candlepin/config/validation package as way to validate  configurations at startup instead of whenever the application happens to hit the code path that uses the configuration. That way the application isn't in a weird state where a portion of functionality isn't working because of a typo or something in the config and you have to notice the issues in the logs.

Ceiu:
> I introduced the org/candlepin/config/validation package as way to validate configurations at startup instead of whenever the application happens to hit the code path that uses the configuration. That way the application isn't in a weird state where a portion of functionality isn't working because of a typo or something in the config and you have to notice the issues in the logs.

Right, and that goal is a good one; but what I mean is we load the configuration as one step, do some other stuff, then validate it as a separate, disconnected step as part of the responsibilities of the CCL. Given the nature of it, it should be part of the logical load operation so that the CCL is only calling out to one sub operation for it.

joshmalbrecht:
> Right, and that goal is a good one; but what I mean is we load the configuration as one step, do some other stuff, then validate it as a separate, disconnected step as part of the responsibilities of the CCL. Given the nature of it, it should be part of the logical load operation so that the CCL is only calling out to one sub operation for it.

Ooooh, my bad, I misunderstood what you were saying! Yeah, you are right the validation should be part of the load operation! It makes no sense to load the configurations then do some other stuff then validate.

joshmalbrecht:
Moving this to draft as I move the Database Management into its own class

src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -179,6 +179,10 @@ private ConfigProperties() {
     public static final String DB_URL = JPA_CONFIG_PREFIX + "hibernate.connection.url";
     public static final String DB_USERNAME = JPA_CONFIG_PREFIX + "hibernate.connection.username";
     public static final String DB_PASSWORD = JPA_CONFIG_PREFIX + "hibernate.connection.password";
+    public static final String DB_MAX_CONNECTION_ATTEMPTS = JPA_CONFIG_PREFIX +
+        "hibernate.connection.max_connection_attempts";
+    public static final String DB_CONNECTION_RETRY_INTERVAL = JPA_CONFIG_PREFIX +
+        "hibernate.connection.retry_interval";
```
nikosmoum, line 185: I think we can use a more generic namespace for these properties, such as `candlepin.db.*`, since there's nothing specific about jpa or hibernate to them. They're only used by candlepin logic (not passed to jpa/jdbc/hibernate) to retry the connection to the database, regardless of the underlying tech (so that their names would still make sense even if we changed from jpa/hibernate to something else)

src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -350,6 +354,10 @@ public static String jobConfig(String jobKey, String cfgName) {
         {
             this.put(CANDLEPIN_URL, "https://localhost");
 
+            // Datebase connection
```
nikosmoum, line 357: typo

src/main/java/org/candlepin/database/DatabaseConnectionManager.java
```diff
@@ -0,0 +1,144 @@
...
+
+        while (attemptNumber <= maxAttempts) {
+            log.info("Attempt {} out of {} to connect to the database.", attemptNumber, maxAttempts);
+            try {
+                return DriverManager.getConnection(
+                    config.getString(ConfigProperties.DB_URL),
+                    config.getString(ConfigProperties.DB_USERNAME),
+                    config.getString(ConfigProperties.DB_PASSWORD));
+            }
+            catch (SQLException e) {
```
nikosmoum, line 107: Is this always going to be a SQLException when we can't connect to the DB? Or is there any circumstance it can be something else, like IOException (with e.g. some kind of network issues)?

nikosmoum (changes requested):
Looks good, just a few minor requests/questions

src/main/java/org/candlepin/config/ConfigProperties.java
joshmalbrecht: I agree, and that makes sense! Updated to use `candlepin.db.*`

src/main/java/org/candlepin/config/ConfigProperties.java
joshmalbrecht: Oops haha updated

src/main/java/org/candlepin/database/DatabaseConnectionManager.java
joshmalbrecht: I think it will always be SQLException, but we can just catch Exception here to be sure! Updated

PR #4660, opened 2024-04-08
[4.3] CANDLEPIN-848 - 2213261: Ensure pool lock order for activation keys

nikosmoum:
- This will stop a deadlock from occuring when multiple consumers are force registered concurrently
- An underlying issue with how many can be run concurrently is still subject to lock wait timeouts

### DIFF
--- a/src/main/java/org/candlepin/bind/PoolOpProcessor.java
+++ b/src/main/java/org/candlepin/bind/PoolOpProcessor.java
@@ -12,7 +12,6 @@
  * granted to use or replicate Red Hat trademarks that are incorporated
  * in this software or its documentation.
  */
-
 package org.candlepin.bind;
 
 import org.candlepin.audit.EventSink;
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/bind/PoolOpProcessor.java
```diff
@@ -90,7 +90,9 @@ private void setPoolQuantity(Map<Pool, Long> poolQuantities) {
             return;
         }
 
-        poolCurator.lock(poolQuantities.keySet());
+        poolCurator.lock(poolQuantities.keySet().stream()
+            .filter(x -> !x.isLocked())
```
Ceiu, line 94: This was a bug in the original PR; the locked filter was erroneously applied and causes pools to be skipped in the Sat case.

Ceiu (changes requested):

src/main/java/org/candlepin/bind/PoolOpProcessor.java
nikosmoum: good catch! updated

Ceiu (approved):

PR #4683, opened 2024-04-19
[M] CANDLEPIN-844: Restored an annotation on the reregistration test

Ceiu:
- Restored an annotation that was removed from the consumer resource test to verify concurrent reregistration which prevented it from running against MySQL backends


Ceiu:
I'm not terribly thrilled about doing this, but unless we're going to rewrite the whole locking model, this is the quick-n-dirty way to get us back to a point we were apparently okay with.

PR #4695, opened 2024-05-02
[M] CANDLEPIN-853: Fixed namespace column type for layered entities

Ceiu:
- Updated the namespace column definition to be the same width as the owner key, which is the primary source for the namespace values


joshmalbrecht (changes requested):
I think we also need to update `cp_contents.namespace` column as well and then this should be good to go!

Ceiu:
Ahh, yes we should. I forgot I applied the layering to more than just products.

joshmalbrecht (approved):

PR #4706, opened 2024-05-10
[M] Remove 'Gradle Enterprise' build scans

nikosmoum:
- I have not found a use for this feature so far. And trying to use it has been cumbersome. Having to add an email address for the scan to be sent at, but never receiving such an email. It only seems like a way for gradle to collect build information and email addresses from people, without providing any value.
- Newer versions of gradle seem to have no need for defining the plugin. you can do a build scan by simply passing the --scan option, if we ever want to get any gradle insights.


joshmalbrecht (approved):
Yeah, it seems like all the information that is provided by the buildscan can be retrieved by looking at the Github actions logs

PR #4708, opened 2024-05-14
[M] Added fixes for various database migration scripts

Ceiu:
- Updated the entity layering migration to avoid unnecessarily copying all content to every org's namespace
- Added some valid checksums on some other scripts which were modified across branches


Ceiu:
Creating this as a draft so we can try to catch any other issues that may come out of testing with production-like data. Will update the PR again when we're ready to merge.

joshmalbrecht (approved):

PR #4720, opened 2024-05-22
CANDLEPIN-823: Removed hibernate criteria from select curator classes

joshmalbrecht:
- Removed hibernate criteria from DeletedConsumerCurator, EntitlementCertificateCurator, ExporterMetadataCurator, and OwnerCurator.

Note that I did not modify `OwnerCurator.getByKeySecure` method because it is used by `StoreFactory`

### DIFF
--- a/src/main/java/org/candlepin/model/DeletedConsumerCurator.java
+++ b/src/main/java/org/candlepin/model/DeletedConsumerCurator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ * Copyright (c) 2009 - 2024 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -17,9 +17,6 @@ package org.candlepin.model;
 import org.candlepin.auth.Principal;
 import org.candlepin.guice.PrincipalProvider;
 
-import org.hibernate.criterion.Projections;
-import org.hibernate.criterion.Restrictions;
-
 import java.time.OffsetDateTime;
 import java.util.Collection;
 import java.util.Date;
@@ -27,6 +24,7 @@ import java.util.List;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import javax.persistence.NoResultException;
 import javax.persistence.TypedQuery;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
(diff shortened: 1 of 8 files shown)


src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -94,9 +104,17 @@ public int countByConsumer(Consumer c) {
...
 
     public int countByConsumerUuid(String uuid) {
-        return ((Long) currentSession().createCriteria(DeletedConsumer.class)
-            .add(Restrictions.eq("consumerUuid", uuid))
-            .setProjection(Projections.rowCount()).uniqueResult()).intValue();
+        if (uuid == null || uuid.isBlank()) {
+            return 0;
+        }
+
+        String query = "SELECT dc FROM DeletedConsumer dc WHERE dc.consumerUuid = :uuid";
```
Ceiu, line 111: We can just do a count and not pull all the data back here:

`SELECT COUNT(dc) FROM ...`

Then the `.getResultList()` below becomes a `.getSingleResult().intValue()`

Ceiu (changes requested):
The `.getByKeySecure` should be updated to use security predicates instead of secure criteria instead. Bonus points if all of the various `.getByKey[...]` methods are moved to be near each other in the class as well.

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
joshmalbrecht: Good catch! Updated

joshmalbrecht:
> The `.getByKeySecure` should be updated to use security predicates instead of secure criteria instead. Bonus points if all of the various `.getByKey[...]` methods are moved to be near each other in the class as well.

For the `OwnerCurator.getByKeySecure` method, I left that method untouched since it is used in `StoreFactory`, and CANDLEPIN-823 says to not touch those methods as they will be addressed in epic [CANDLEPIN-49](https://issues.redhat.com/browse/CANDLEPIN-49).

I did reorganize those `.getByKey` methods in `OwnerCurator`!

Ceiu:
CP-49 can't be relied on in this case, and anything left using secure criteria (or criteria in any capacity for that matter) prevents upgrading to Hibernate 6.

As a side note, CP-49 is superseded by the migration to Kessel, but may still be important for the on-prem solution. However, I'd rather not delay any upgrades or moves to Hibernate, Quarkus, etc. on it.

joshmalbrecht:
> CP-49 can't be relied on in this case, and anything left using secure criteria (or criteria in any capacity for that matter) prevents upgrading to Hibernate 6. I'd rather we fully clear out each curator as we hit them so we don't end up missing stuff or needing to revisit later if we can avoid it. Especially in this case since we have the "secure predicate" shims which can be used in their place.
> 
> As a side note, CP-49 is superseded by the migration to Kessel, but may still be important for the on-prem solution. However, I'd rather not delay any upgrades or moves to Hibernate, Quarkus, etc. on it.

That makes sense! I updated the `.getByKeySecure` method.

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -94,9 +104,17 @@ public int countByConsumer(Consumer c) {
...
-            .add(Restrictions.eq("consumerUuid", uuid))
-            .setProjection(Projections.rowCount()).uniqueResult()).intValue();
+        if (uuid == null || uuid.isBlank()) {
+            return 0;
+        }
+
+        String query = "SELECT COUNT(dc) FROM DeletedConsumer dc WHERE dc.consumerUuid = :uuid";
+
+        return this.getEntityManager()
+            .createQuery(query, Number.class)
```
Ceiu, line 114: Can you explain the usage of `Number.class` here instead of a more specific class?

joshmalbrecht: Oops, this should be `Long.class` because the value returned from the query is a Long. Both work, but since we know the value is a Long, it is better to use the subclass.

Ceiu (approved):

PR #4724, opened 2024-05-30
[M] CANDLEPIN-856: Support environment deletion without deleting consumers

bakajstep:
- If an environment was deleted, any consumers in the environment that did not belong to any other environment were also deleted. For console.redhat.com content templates, we did not want this behavior. With deleteConsumers set to false, we left the consumers environment-less.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/environments/EnvironmentSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/environments/EnvironmentSpecTest.java
@@ -56,6 +56,9 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.NullSource;
+import org.junit.jupiter.params.provider.ValueSource;
 
 import java.util.Arrays;
 import java.util.Collection;
@@ -106,7 +109,7 @@ public class EnvironmentSpecTest {
     public void shouldAllowOwnerAdminToDeleteEnvironments() {
         EnvironmentDTO env = ownerClient.owners().createEnv(owner.getKey(), Environments.random());
 
-        ownerClient.environments().deleteEnvironment(env.getId());
+        ownerClient.environments().deleteEnvironment(env.getId(), true);
 
         List<EnvironmentDTO> environments = ownerClient.owners().listEnvironments(owner);
         assertThat(environments)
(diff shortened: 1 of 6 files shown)


src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -191,22 +192,24 @@ public EnvironmentDTO getEnvironment(@Verify(Environment.class) String envId) {
...
-            .filter(consumer -> consumer.getEnvironmentIds().size() == 1)
-            .toList();
-        deleteConsumers(environment, consumersToDelete);
 
-        List<Consumer> consumersToKeep = consumers.stream()
-            .filter(consumer -> consumer.getEnvironmentIds().size() > 1)
-            .toList();
-        removeConsumersFromEnvironment(environment, consumersToKeep);
+        if (Boolean.TRUE.equals(deleteConsumers)) {
+            Map<Boolean, List<Consumer>> partitionedConsumers = consumers.stream()
```
bakajstep, line 204: With this approach we should save time because we won't have to go through the list twice but only once.

spec-tests/src/test/java/org/candlepin/spec/environments/EnvironmentSpecTest.java
```diff
@@ -662,14 +662,32 @@ public void shouldDeleteConsumerTogetherWithHisLastEnvironment() {
...
+        ownerClient.environments().deleteEnvironment(env1.getId(), true);
 
         assertGone(() -> ownerClient.consumers().getConsumer(consumer1.getUuid()));
 
         ConsumerDTO consumer = ownerClient.consumers().getConsumer(consumer2.getUuid());
         assertThat(consumer.getEnvironments()).hasSize(1);
     }
 
+    @Test
+    public void shouldNotDeleteConsumerWithDeletingHisLastEnvironment() {
```
Ceiu, line 674: nit (grammar): "shouldNotDeleteConsumerWhenDeletingItsLastEnvironment"

Ceiu (changes requested):
Looks pretty good, just need to ensure we don't NPE since our boolean flag is an object and not a primitive.

Some minor grammar improvements can also be made since there are other things to address as well.

src/test/java/org/candlepin/resource/EnvironmentResourceTest.java
```diff
@@ -224,12 +226,32 @@ void onlyConsumersWithTheirLastEnvRemovedShouldBeCleanedUp() {
...
 
-        this.environmentResource.deleteEnvironment(ENV_ID_1);
+        this.environmentResource.deleteEnvironment(ENV_ID_1, true);
 
         verify(this.consumerCurator).delete(consumer1);
         verify(this.contentAccessManager).removeContentAccessCert(consumer2);
     }
 
+    @Test
+    void shouldConsumersBeRemovedFromEnvironmentButNotDeletedWhenDeleteConsumersIsFalse() {
```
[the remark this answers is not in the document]
bakajstep: done

spec-tests/src/test/java/org/candlepin/spec/environments/EnvironmentSpecTest.java
bakajstep: done

Ceiu (approved):

PR #4731, opened 2024-06-05
[M] CANDLEPIN-822: Removed hibernate criteria from select curator classes

bakajstep:
- Updated ContentOverrideCurator, RoleCurator, UserCurator and CdnCurator to stop using Hibernate APIs, such as Hibernate criteria and Hibernate implementations of JPA interfaces

### DIFF
--- a/src/main/java/org/candlepin/model/ContentOverrideCurator.java
+++ b/src/main/java/org/candlepin/model/ContentOverrideCurator.java
@@ -14,15 +14,17 @@
  */
 package org.candlepin.model;
 
-import org.hibernate.criterion.Restrictions;
+import com.google.inject.persist.Transactional;
 
 import java.util.List;
 
+import javax.persistence.EntityManager;
+import javax.persistence.NoResultException;
 import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaDelete;
 import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
-
-
 
 /**
  * ContentOverrideCurator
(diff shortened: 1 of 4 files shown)


src/main/java/org/candlepin/model/ContentOverrideCurator.java
```diff
@@ -56,35 +56,49 @@ public List<T> getList(Parent parent) {
     }
 
     public void removeByName(Parent parent, String contentLabel, String name) {
-        List<T> overrides = currentSession()
-            .createCriteria(this.entityType())
-            .add(Restrictions.eq(parentAttr, parent))
-            .add(Restrictions.eq("contentLabel", contentLabel))
-            .add(Restrictions.eq("name", name).ignoreCase())
-            .list();
+        CriteriaBuilder cb = getEntityManager().getCriteriaBuilder();
```
Ceiu, line 59: While this works, we could instead just do a direct delete either with JPQL or a CriteriaDelete, rather than fetching everything and then deleting them one-by-one.

This occurs in a few places.

src/main/java/org/candlepin/model/ContentOverrideCurator.java
```diff
@@ -93,13 +107,24 @@ public void removeByParent(Parent parent) {
...
+            Root<T> root = cq.from(this.entityType());
+
+            Predicate parentPredicate = cb.equal(root.get(parentAttr), parent);
+            Predicate contentLabelPredicate = cb.equal(root.get("contentLabel"), contentLabel);
+            Predicate namePredicate = cb.equal(cb.lower(root.get("name")), name.toLowerCase());
+
+            cq.where(cb.and(parentPredicate, contentLabelPredicate, namePredicate));
+
+            try {
+                return getEntityManager().createQuery(cq)
```
Ceiu, line 121: On this line and in most of the updates in this class, we're doubly fetching the entity manager rather than getting it once and then reusing it.

Ceiu (changes requested):

src/main/java/org/candlepin/model/ContentOverrideCurator.java
bakajstep: done

src/main/java/org/candlepin/model/ContentOverrideCurator.java
bakajstep: done

src/main/java/org/candlepin/model/UserCurator.java
```diff
@@ -32,18 +34,39 @@ public UserCurator() {
...
+        EntityManager em = getEntityManager();
+        CriteriaBuilder cb = em.getCriteriaBuilder();
+        CriteriaQuery<User> cq = cb.createQuery(User.class);
+        Root<User> root = cq.from(User.class);
+
+        Predicate usernamePredicate = cb.equal(root.get("username"), login);
+        Predicate securityPredicate = getSecurityPredicate(User.class, cb, root);
+
+        if (securityPredicate != null) {
+            cq.where(cb.and(usernamePredicate, securityPredicate));
```
Ceiu, line 46: FYI: It's not necessary to make the conjunction for a situation like this. The `.where` method is variadic and multiple predicates will be applied as a conjunction, so the following code is functionally identical to this line:

```
cq.where(usernamePredicate, securityPredicate);
```

Ceiu (approved):

src/main/java/org/candlepin/model/UserCurator.java
bakajstep: good to know. but isn't it better if the and is there? if someone goes through the code after a while, they might not remember that there is an implicit and.

Ceiu: Depends on how you read it and whether or not the reader understands that as well. It's functionally identical and probably does exactly what you did under the hood anyway, so if you'd rather be explicit in such cases, that's fine, too.

PR #4733, opened 2024-06-11
[4.4.1] CANDLEPIN-862: Use default content enablement in anonymous certs

nikosmoum:
- Content in anonymous entitlement certificates should not be always overriden to enabled. Instead the default value from the content definition in the IT Product Service should be used, like for other (SCA/Entitlement) certs.


joshmalbrecht (approved):

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -960,32 +961,31 @@ private AnonymousContentAccessCertificate createAnonContentAccessCertificate(
...
-    private List<ContentInfo> getContentInfo(Collection<ProductInfo> prodInfo) {
+    private Map<ContentInfo, Boolean> getContentInfo(Collection<ProductInfo> prodInfo) {
         return prodInfo.stream()
             .filter(Objects::nonNull)
             .map(ProductInfo::getProductContent)
             .flatMap(Collection::stream)
             .filter(Objects::nonNull)
-            .map(ProductContentInfo::getContent)
-            .toList();
+            .collect(Collectors.toMap(ProductContentInfo::getContent, ProductContentInfo::isEnabled));
```
Ceiu, line 988: I think this could lead to some inconsistency in cases where the input collection has multiple products with the same content but having different enabled states -- the order may not be consistent from run to run.

We may need to fallback to the `enabled ||= pcinfo.isEnabled` behavior rather than just taking it straight up.

Ceiu (changes requested):
Bulk of this looks sufficient, but I'm concerned about the map collector

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -960,32 +961,48 @@ private AnonymousContentAccessCertificate createAnonContentAccessCertificate(
...
-        return prodInfo.stream()
+    private Map<ContentInfo, Boolean> getContentInfo(Collection<ProductInfo> prodInfo) {
+        Collection<ProductContentInfo> productContentInfos = prodInfo.stream()
             .filter(Objects::nonNull)
             .map(ProductInfo::getProductContent)
             .flatMap(Collection::stream)
             .filter(Objects::nonNull)
-            .map(ProductContentInfo::getContent)
-            .toList();
+            .collect(Collectors.toList());
```
Ceiu, line 989: We can still use a `.collect(Collectors.toMap(...))` here, just need to use the version that takes a merge function:

```
.collect(Collectors.toMap(ProductContentInfo::getContent, ProductContentInfo::isEnabled, (e, u) -> { e || u }))
```

This also avoids doubly iterating through the collection.

The other concern here is whether or not ContentInfo will even collide as a key in the way we want/expect for resolution. Since it's just an interface, we're hoping that each instance properly implements hashCode and equals.

We may be in a situation where our list always has duplicates in the case of multiple products with the same content. Though in the case of specifically SCA mode, we're mostly isolated by known properties of the upstream data (no content is shared).

Maybe this is fine for now, but I think in terms of our actual data model and long-term maintenance, this is a ticking time bomb.

Ceiu (changes requested):

src/main/java/org/candlepin/controller/ContentAccessManager.java
nikosmoum: Nice, I didn't know about the merge function. Updated.

With regards to how collisions are handled, the adapter where this data is coming from today is using our ProductContentData/ContentData concrete implementations of the *Info interfaces, which do have proper hashcode/equals

Ceiu: Yikes, that's no good. Those do field-by-field comparison rather than just ID comparison like we'd want here.

Looking at the code a bit today, it seems we do all this conversion just to build a final product instance. To that end, we can approach this in two ways: one is to kinda shim that together as done here, or we can do some more foundational changes to actually clean up the build process so it's not so abstracted that we lose the goal through the weeds of abstraction calls.
In the interest of getting this through for 4.4.1 and 4.4.2, we can do a bit more of the former, but still kinda lean into what we should do for a more thorough fix in main:

In the `createAnonPayloadAndSignature` method, we can rip out most of its guts and build a map of content IDs to `ProductContent` instances. Using ProductContent gives us a nice wrapper that holds the content and the enabled state already, so we can then use a stable ID as the key to avoid any potential issues there. Streams make this very easy, but we'll have to preload some conversions to make the final code not look like spaghetti:

```
    private String createAnonPayloadAndSignature(Collection<ProductInfo> prodInfo) throws IOException {
        Function<ProductContentInfo, ProductContent> converter = (pcinfo) -> {
            ...
        };

        Function<ProductContent, String> cidFetcher = (pcinfo) -> pcinfo.getContent().getId();

        Map<String, ProductContent> activeContent = prodInfo.stream()
            .filter(Objects::nonNull)
            .flatMap(ProductInfo::getProductContent)
            .filter(Objects::nonNull)
            .map(converter)
            .collect(Collectors.toMap(cidFetcher, Function.identity(), (v1, v2) -> {
                return new ProductContent()
                    .setContent(v2.getContent()) // last in wins
                    .setEnabled(v1.isEnabled() || v2.isEnabled());
            }));

        PromotedContent promotedContent = new PromotedContent(new AnonymousContentPrefix());

        byte[] data = this.createContentAccessDataPayload(null, activeContent, promotedContent);
        return this.createPayloadAndSignature(data);
    }
```

From there, now we just need to update `createContentAccessDataPayload` to accept it, which is actually less work than what we're doing now:

```
    private byte[] createContentAccessDataPayload(Consumer consumer, 
        Map<String, ProductContent> activeContent, PromotedContent promotedContent) {

        Product engProduct = new Product()
            .setId("content_access_eng")
            .setName("Content Access Eng")
            .setProductContent(activeContent.values());

        Product skuProduct = new Product()
            ...
```

We'll also need to update the two other places this is called to also provide the same map:

```
        Function<Map.Entry<Content, Boolean>, ProductContent> pcbuilder = (entry) -> new ProductContent()
          .setContent(entry.getKey())
          .setEnabled(entry.getValue());
          
        Function<ProductContent, String> cidFetcher = (pcinfo) -> pcinfo.getContent().getId();

        Map<String, ProductContent> ownerContent = this.ownerContentCurator
            .getActiveContentByOwner(owner.getId())
            .keySet()
            .stream()
            .map(pcbuilder)
            .collect(Collectors.toMap(cidFetcher, Function.identity(), (v1, v2) -> {
                return new ProductContent()
                    .setContent(v2.getContent()) // last in wins
                    .setEnabled(v1.isEnabled() || v2.isEnabled());
            }));

        byte[] payloadBytes = createContentAccessDataPayload(consumer, ownerContent, promotedContent);  
```

Realistically, that `getActiveContentByOwner` function should be rewritten to kick out ProductContent instances and use a proper CTE to avoid recursive queries locally (since we're already using a native query anyway), but that's a bit of a heavy lift for this branch and 4.4.2.

nikosmoum: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -229,6 +231,28 @@ public static String getListDefaultDatabaseValue() {
 
     private final boolean standalone;
 
+    private final Function<ProductContentInfo, ProductContent> productContentInfoToProductContent =
```
Ceiu, line 234: if we're going declare this as an instance variable, we can just declare it as a function instead:

```
private ProductContent translateProductContent(ProductContentInfo pcinfo) {
  ...
}
```

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -229,6 +231,28 @@ public static String getListDefaultDatabaseValue() {
...
+        (pcinfo) -> {
+            ContentInfo cinfo = pcinfo.getContent();
+
+            org.candlepin.model.Content converted = new org.candlepin.model.Content();
```
Ceiu, line 238: nit: we can use the fluent nature of the mutators on Content to avoid repeating the `converted.` all the way down.

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -906,8 +957,8 @@ public AnonymousContentAccessCertificate getCertificate(AnonymousCloudConsumer c
             }
 
             payload = createAnonPayloadAndSignature(products);
-            List<ContentInfo> contentInfo = getContentInfo(products);
-            content = convertContentInfoToContentDto(contentInfo);
+            Map<ContentInfo, Boolean> contentInfo = getContentInfo(products);
```
Ceiu, line 960: Were we worried at all about the collision issue here as well?

Ceiu (changes requested):

src/main/java/org/candlepin/controller/ContentAccessManager.java
nikosmoum: I think we should be, if we're worried about the other case. But here it will have to be translated all the way into the model.dto.Content. I'll update it

nikosmoum: Looking at this again, we don't really need the enabled/disabled flag here. In fact, from the `dto.Content` objects we're building in order to get into the x509 part of the cert, we only ever use the `path` field from them.
However, we should still translate them into a Map of Content ID to ProductContent, to be consistent with the list of content we have in the payload. So in short, I'm changing this to:
- Translate ProductInfo into Map<String, ProductContent>
- Use that map to build the payload / content list part of the cert.
- Translate the ProductContent from that same map into dto.Content objects that only have the `path` field populated, and throw that into the X509 part of the cert.

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -959,93 +987,82 @@ private AnonymousContentAccessCertificate createAnonContentAccessCertificate(
...
+            .setType(cinfo.getType())
+            .setLabel(cinfo.getLabel())
+            .setVendor(cinfo.getVendor())
+            .setGpgUrl(cinfo.getGpgUrl())
+            .setMetadataExpiration(cinfo.getMetadataExpiration())
+            .setRequiredTags(cinfo.getRequiredTags())
+            .setArches(cinfo.getArches())
+            .setContentUrl(cinfo.getContentUrl());
+
+        ProductContent productContent = new ProductContent();
```
Ceiu, line 1044: nit: same deal here. All of this can be condensed down to an inline build:

```
return new ProductContent()
  .setContent(converted)
  .setEnabled(pcinfo.isEnabled());
```

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -959,93 +987,82 @@ private AnonymousContentAccessCertificate createAnonContentAccessCertificate(
...
-                return converted;
+    public List<Content> convertProductContentToContentDto(Collection<ProductContent> productContentSet) {
+        return productContentSet.stream()
+            .map(productContent -> {
+                Content contentContainer = new Content();
+                contentContainer.setPath(productContent.getContent().getContentUrl());
+                return contentContainer;
             })
-            .toList();
+            .collect(Collectors.toList());
```
Ceiu, line 1066: nit: `.toList()` does this without needing to explicitly import/use `Collectors`

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -524,15 +553,14 @@ private Set<X509ByteExtensionWrapper> prepareV3ByteExtensions(org.candlepin.mode
...
+        Map<String, ProductContent> activeContent, PromotedContent promotedContent)
         throws IOException {
         log.info("Generating SCA payload for consumer \"{}\"...", consumer != null ? consumer.getUuid() :
             "unknown");
         Product engProduct = new Product()
             .setId("content_access")
-            .setName(" Content Access");
-
-        ownerContent.forEach(engProduct::addContent);
+            .setName(" Content Access")
```
Ceiu, line 562: non-review question: Is the space in front of "Content Access" intentional here? What was that for?

Ceiu (approved):
A question and two very minor nitpicky things. But it's ready to go if you don't feel like changing them.

src/main/java/org/candlepin/controller/ContentAccessManager.java
nikosmoum: Looking at the log, it's been like that since SCA was first implemented. I thought maybe this was done as a 'UX' enhancement to make sure it looks when someone prints the cert using `rct cat-cert`, or maybe the traditional entitlement certs also had a space appended before product names, but neither seem to be the case. So it looks accidental.
However, I'm not sure if changing it now is a good idea, because there may be users that are parsing this (for whatever reason), which would break.

This made me think though, do we want to set a different product id/name for the anonymous cert? Something like 'Anonymous Content Access', or 'Cloud Offering Content Access' (there is maybe a better name I can't think of right now). It may make it easier for testing/debugging purposes, if there's such an indication.

nikosmoum: cc: @joshmalbrecht

src/main/java/org/candlepin/controller/ContentAccessManager.java
nikosmoum: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
nikosmoum: Updated

src/main/java/org/candlepin/controller/ContentAccessManager.java
joshmalbrecht: I like that idea of giving some indication that this is an anonymous Certificate for the reasons you listed! I was taking a look at Authorizer and I don't think that would cause any issues in their parsing of the certificate, but I'm not super familiar with that code.

PR #4734, opened 2024-06-11
[M] CANDLEPIN-235: Removed hibernate criteria from select curator classes

bakajstep:
- Updated GuestIdCurator, ConsumerTypeCurator, RulesCurator, ContentCurator, ActivationKeyCurator and AnonymousCloudConsumerCurator to stop using Hibernate APIs, such as Hibernate criteria and Hibernate implementations of JPA interfaces

### DIFF
--- a/src/main/java/org/candlepin/model/ConsumerTypeCurator.java
+++ b/src/main/java/org/candlepin/model/ConsumerTypeCurator.java
@@ -16,8 +16,6 @@ package org.candlepin.model;
 
 import org.candlepin.model.ConsumerType.ConsumerTypeEnum;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/candlepin/model/ConsumerTypeCurator.java
```diff
@@ -133,10 +131,12 @@ public ConsumerType getByLabel(String label, boolean createIfAbsent) {
      * @param labels
      * @return all types matching the specified labels;
      */
-    @SuppressWarnings("unchecked")
```
Ceiu, line 136: :+1:

Ceiu (approved):

PR #4738, opened 2024-06-13
[M] CANDLEPIN-865: Added explicit casts in entity layering migration

Ceiu:
- Updated the entity layering script to include explicit casts in the common table expressions for recursively fetching products to avoid an issue with PostgreSQL's type inference system not having a varchar length in some cases


Ceiu:
This change somewhat scares me. It seems innocuous in general and it fixes the issue in the reproducer, but I don't fully understand *why* Postgre decided to throw out the field length on the product UUID field, but *not* the owner ID field, despite them being used and queried in the exact same way in the exact same type of recursive CTE.

This is probably safe, but being unable to explain why it's happening leaves me feeling very uneasy.

joshmalbrecht (approved):

PR #4743, opened 2024-06-18
[M] CANDLEPIN-864: Gracefully handled DB driver class conf not existing

bakajstep:
- Gracefully handled missing DB driver configuration to prevent NPE for Candlepin users.

### DIFF
--- a/src/test/java/org/candlepin/config/validation/StringConfigurationValidatorTest.java
+++ b/src/test/java/org/candlepin/config/validation/StringConfigurationValidatorTest.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2009 - 2024 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.config.validation;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.doReturn;
+
+import org.candlepin.config.Configuration;
+import org.candlepin.test.TestUtil;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.NullAndEmptySource;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+@ExtendWith(MockitoExtension.class)
+public class StringConfigurationValidatorTest {
+
+    @Mock
+    protected Configuration mockConfig;
+
+    @ParameterizedTest
+    @NullAndEmptySource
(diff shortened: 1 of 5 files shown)


src/main/java/org/candlepin/exceptions/MissingConfigurationException.java
```diff
@@ -0,0 +1,23 @@
...
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.exceptions;
+
+public class MissingConfigurationException extends RuntimeException {
```
joshmalbrecht, line 18: Another approach to this might be to leverage the `org.candlepin.config.validation` package for this. Currently we validate some configurations at the following step in the startup process, and generate a message for each fatal configuration issue that prevents the startup from happening.

https://github.com/candlepin/candlepin/blob/ea897a3863960bb4e99698a112a9b068e46007a7/src/main/java/org/candlepin/guice/CandlepinContextListener.java#L317-L319

If we were to throw this MissingConfigurationException instead of using the validation classes, then it's possible that the user has multiple configuration problems that are displayed from `ConfigurationValidatorUtil.validateConfigurations` and they address those issues, and then run into this  `MissingConfigurationException` message after fixing the other configurations. If we added this configuration as one that is validated with the other essential configurations, then the user would see this missing configuration with the other possible configurations issues and can be addressed at once.

To do this, you will have to add a `StringConfigurationValidator` class that implements `ConfigurationValidator` that validates that a value exists for the `ConfigProperties.DB_DRIVER_CLASS)` key. Then you need to add the validation for the key in the `ConfigProperties.CONFIGURATION_VALIDATORS` list.

https://github.com/candlepin/candlepin/blob/ea897a3863960bb4e99698a112a9b068e46007a7/src/main/java/org/candlepin/config/ConfigProperties.java#L572-L587

joshmalbrecht (changes requested):

src/main/java/org/candlepin/exceptions/MissingConfigurationException.java
joshmalbrecht: It's probably a good idea to have both the validation using the `org.candlepin.config.validation` package and the null check in `DatabaseConnectionManager.loadDatabaseDriver` so we don't hit the NPE in the case that someone removes the configuration validation from `ConfigProperties.CONFIGURATION_VALIDATORS`.

joshmalbrecht (changes requested):

src/main/java/org/candlepin/exceptions/MissingConfigurationException.java
bakajstep: and do we want to still throw `MissingConfigurationException` in the second validation?

joshmalbrecht: I would just throw a `RuntimeException` instead of creating the `MissingConfigurationException` class

joshmalbrecht (changes requested):
Looks good! I just noticed an extra new line to remove

src/main/java/org/candlepin/exceptions/MissingConfigurationException.java
bakajstep: done, i had initially `RuntimeException` but sonar will give you warning do not throw generic exceptions. But `RuntimeException` is alse fine by me.

src/test/java/org/candlepin/config/validation/StringConfigurationValidatorTest.java
```diff
@@ -0,0 +1,113 @@
...
+        assertThrows(IllegalArgumentException.class, () -> validator.allowedValues(null));
+    }
+
+    @Test
+    public void testAllowedValuesWithEmptyValues() {
+        String key = TestUtil.randomString();
+        StringConfigurationValidator validator = new StringConfigurationValidator(key);
+        assertThrows(IllegalArgumentException.class, () -> validator.allowedValues(Collections.emptySet()));
+    }
+
```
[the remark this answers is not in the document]
bakajstep: done

joshmalbrecht (approved):

PR #4752, opened 2024-06-26
[M] CANDLEPIN-868: Anon Registration results in Ents with all repos Enabled

bakajstep:
- Content in anonymous entitlement certificates should not be always overriden to enabled. Instead the default value from the content definition in the IT Product Service should be used, like for other (SCA/Entitlement) certs.
- Refactor OwnerContentCurator.getActiveContentByOwner() to use CTE and to avoid recursive queries.

### DIFF
--- a/src/main/java/org/candlepin/model/ContentCurator.java
+++ b/src/main/java/org/candlepin/model/ContentCurator.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
@@ -635,7 +636,7 @@ public class ContentCurator extends AbstractHibernateCurator<Content> {
     }
 
     /**
-     * Returns a mapping of content and content enabled flag.
+     * Returns a list of active product content for a given owner
      *
      * Note - If same content (say C1) is being enabled/disabled by two or more products, with at least
      * one of the product enabling the content C1 then, content (C1) with enabled (true)
@@ -645,63 +646,63 @@ public class ContentCurator extends AbstractHibernateCurator<Content> {
      *  Id of an owner
      *
      * @return
-     *  Returns a mapping of content and content enabled flag.
+     *  Returns list of active product content
      */
-    public Map<Content, Boolean> getActiveContentByOwner(String ownerId) {
+    public List<ProductContent> getActiveContentByOwner(String ownerId) {
         EntityManager entityManager = this.getEntityManager();
         Date now = new Date();
 
-        // Store all of the product UUIDs gathered between the various queries we need to do here
-        Set<String> productUuids = new HashSet<>();
+        // Use a CTE to get all derived and provided products
+        String sql = """
+                WITH RECURSIVE parent_child_map(parent_uuid, child_uuid) AS (
+                    SELECT CAST(uuid AS CHAR(32)), CAST(derived_product_uuid AS CHAR(32))
+                    FROM cp_products
+                    WHERE derived_product_uuid IS NOT NULL
+                    UNION
+                    SELECT CAST(product_uuid AS CHAR(32)), CAST(provided_product_uuid AS CHAR(32))
+                    FROM cp_product_provided_products
+                ),
+                product_graph(uuid, depth) AS (
+                    SELECT CAST(pool.product_uuid AS CHAR(32)), 0
+                    FROM cp_pool pool
+                    WHERE pool.owner_id = :owner_id
+                    AND pool.startDate <= :start_date AND pool.endDate >= :end_date
+                    UNION
+                    SELECT CAST(pcmap.child_uuid AS CHAR(32)), pgraph.depth + 1
+                    FROM parent_child_map pcmap
+                    JOIN product_graph pgraph ON pgraph.uuid = pcmap.parent_uuid
+                    WHERE pgraph.depth < 10
+                )
+                SELECT content.uuid, pc.enabled
+                FROM product_graph pgraph
+                JOIN cp_product_contents pc ON pc.product_uuid = pgraph.uuid
+                JOIN cp_contents content ON content.uuid = pc.content_uuid
+            """;
 
-        // Get all of the products from active pools for the specified owner
-        String jpql = "SELECT pool.product.uuid FROM Pool pool " +
-            "WHERE pool.owner.id = :owner_id AND pool.startDate <= :start_date AND pool.endDate >= :end_date";
-
-        String sql = "SELECT prod.derived_product_uuid FROM cp_products prod " +
-            "  WHERE prod.uuid IN (:product_uuids) AND prod.derived_product_uuid IS NOT NULL " +
-            "UNION " +
-            "SELECT prov.provided_product_uuid FROM cp_product_provided_products prov " +
-            "  WHERE prov.product_uuid IN (:product_uuids)";
-
-        productUuids.addAll(entityManager.createQuery(jpql, String.class)
+        Map<String, Boolean> contentEnablementMap =
+            (Map<String, Boolean>) entityManager.createNativeQuery(sql)
             .setParameter("owner_id", ownerId)
             .setParameter("start_date", now)
             .setParameter("end_date", now)
-            .getResultList());
+            .getResultStream()
+            .collect(Collectors.toMap(
+                row -> (String) ((Object[]) row)[0],  // UUID of the content
+                row -> (Boolean) ((Object[]) row)[1], // Enabled state
+                (enabled1, enabled2) -> enabled1 || enabled2 // Collision resolution: logical OR
+            ));
 
-        Query prodQuery = entityManager.createNativeQuery(sql);
-        int blockSize = Math.min(this.getInBlockSize(), this.getQueryParameterLimit() / 2);
+        List<ProductContent> activeContent = new ArrayList<>();
 
-        // Get all of the derived products and provided products for the products we've found, recursively
-        for (Set<String> lastBlock = productUuids; !lastBlock.isEmpty();) {
-            Set<String> children = new HashSet<>();
+        String contentLookupJPQL =
+            "SELECT content FROM Content content WHERE content.uuid IN (:content_uuids)";
+        TypedQuery<Content> contentQuery = entityManager.createQuery(contentLookupJPQL, Content.class);
 
-            for (List<String> block : this.partition(lastBlock, blockSize)) {
-                children.addAll(prodQuery.setParameter("product_uuids", block)
-                    .getResultList());
-            }
-
-            productUuids.addAll(children);
-            lastBlock = children;
-        }
-
-        // Use the product UUIDs to select all related enabled content
-        jpql = "SELECT DISTINCT pc.content, pc.enabled FROM ProductContent pc " +
-            "WHERE pc.product.uuid IN (:product_uuids)";
-
-        Query contentQuery = entityManager.createQuery(jpql);
-        HashMap<Content, Boolean> activeContent = new HashMap<>();
-
-        for (List<String> block : this.partition(productUuids)) {
-            contentQuery.setParameter("product_uuids", block)
-                .getResultList()
-                .forEach(col -> {
-                    Content content = (Content) ((Object[]) col)[0];
-                    Boolean enabled = (Boolean) ((Object[]) col)[1];
-                    activeContent.merge(content, enabled, (v1, v2) ->
-                        v1 != null && v1.booleanValue() ? v1 : v2);
-                });
+        // This step is necessary to handle large sets of UUIDs and avoid query parameter limits
+        for (List<String> block : this.partition(contentEnablementMap.keySet())) {
+            contentQuery.setParameter("content_uuids", block)
+                .getResultStream()
+                .map(content -> new ProductContent(content, contentEnablementMap.get(content.getUuid())))
+                .forEach(activeContent::add);
         }
 
         return activeContent;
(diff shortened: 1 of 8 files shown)


src/main/java/org/candlepin/model/ProductContent.java
```diff
@@ -44,6 +48,32 @@
 @Immutable
 @Cache(usage = CacheConcurrencyStrategy.READ_ONLY)
 @Table(name = ProductContent.DB_TABLE)
+@SqlResultSetMapping(
```
bakajstep, line 51: unlike the previous solution we use native query and it doesn't support mapping to object if there are other attributes in the result than from the given class so I had to use this functionality

Ceiu: nack on this

This is a major failing on the part of the JPA spec in that to properly use this mapping, we have to define it with annotations, in a place the annotation scanner picks it up, which is likely in a separate location from the query that requires it. This is just extra maintenance pain for something that should absolutely not be defined with annotations, or at the very least, should be something that can be manually constructed with POJOs.

Additionally, due to the inconsistent behavior with the default mapping vs this mapping with regards to field strictness, it's not clear what would happen in the future if a column in the defined list were to go missing or get renamed. Also, will Content instances still lazy-load their collection fields when fetched using this mapping?

As a distant third, we're also using the splat operator in the native query itself, which isn't great for our production usage since we can no longer rely upon the order or presence of any given field.

It's an interesting solution overall, but JPA's interface for implementing this behavior is unacceptably terrible, and creates an unnecessary maintenance burden and code bloat.

Ceiu (changes requested):

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
```diff
@@ -317,16 +313,15 @@ private Set<X509Extension> prepareV3ByteExtensions(org.candlepin.model.dto.Produ
...
         String consumerUuid = consumer != null ? consumer.getUuid() : null;
         log.info("Generating SCA payload for consumer \"{}\"...", consumerUuid);
 
         Product engProduct = new Product()
             .setId("content_access")
-            .setName(" Content Access");
-
-        ownerContent.forEach(engProduct::addContent);
+            .setName(" Content Access")
+            .setProductContent(activeContent.values());
```
nikosmoum, line 324: Can we update `content_access` and ` Content Access` here into something more accurate, like `anonymous_cloud_offering_access`/`Anonymous Cloud Offering Access`?

And same thing for the `skuProduct` object below this?

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
```diff
@@ -317,16 +313,15 @@ private Set<X509Extension> prepareV3ByteExtensions(org.candlepin.model.dto.Produ
     }
 
     private byte[] createContentAccessDataPayload(Consumer consumer,
-        Map<org.candlepin.model.Content, Boolean> ownerContent, PromotedContent promotedContent) {
+        Map<String, ProductContent> activeContent, PromotedContent promotedContent) {
 
         String consumerUuid = consumer != null ? consumer.getUuid() : null;
         log.info("Generating SCA payload for consumer \"{}\"...", consumerUuid);
```
nikosmoum, line 319: Similarly, lets change `SCA` to `ACOA` (anonymous cloud offering access) here? the current log string is leftover from when this code was shared between anonymous and SCA certs, and leaving it like this is innacurate and would confuse during debugging.

nikosmoum (changes requested):

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
nikosmoum: Also, let me know if you can think of a more accurate name. @bakajstep @Ceiu @joshmalbrecht Once the cert implementation is out, its hard to change because people start parsing it. 

I thought only adding 'Anonymous Access' isn't specific enough, because there's been talk of eventually introucing anonymous/automatic registration for non-cloud systems. Perhaps `Anonymous Cloud Access` is enough (without the Offering part)?

joshmalbrecht: I like `Anonymous Cloud Content Access` the most

bakajstep: We talked about it at scrum and agreed on `Anonymous Cloud Content Access`

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
nikosmoum: Then I suppose this should also be `ACCA` (Anonymous Cloud Content Access`) instead of `SCA`

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
bakajstep: done

src/main/java/org/candlepin/model/ProductContent.java
bakajstep: I was looking for recommended approach and found that. I also find it strange that you can actually define a mapping above the table definition and for unrelated tables and then use it anywhere.

Best practife for this example is to use something like this: `SELECT content.uuid as uuid` so if you rename attribute you dont need to change mapping and if you delete attribute it is ok it can handle that you would only have unused code in mapping. 

Yes, it will still be lazy-loaded.

But I agree with you that it's terrible. Definitely the mapping from the second command seems better or can you think of a better way?

src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -645,68 +645,76 @@ public Set<Content> getChildrenContentOfProductsByUuids(Collection<String> produ
...
-                        v1 != null && v1.booleanValue() ? v1 : v2);
-                });
+        for (Object[] result : results) {
+            activeContent.add(productContentMapping(result));
         }
 
         return activeContent;
     }
 
+    protected ProductContent productContentMapping(Object[] result) {
```
Ceiu, line 699: This has two problems:
1) the proper functionality of this method is very tightly coupled to the correctness of the array passed in -- which itself is entirely unchecked throughout the method -- and that is really only viable in the context of the query in question.

2) while we're building most of the content correctly, this doesn't include the required or "modified" product IDs. Additionally, because it's a concrete Content instance and not a Hibernate proxy, it will return an empty list instead of fetching them; which could have consequences should this method be used in the future for other functionality (e.g. the proposed "active content" endpoint).

This logic would be better served as a closure within the method executing the query so that it could be applied as a stream, and logically separated from the core functionality of the method, without being decoupled and at risk of being used elsewhere.

For addressing the second point, it might be best or easiest to simply fetch the content IDs, and then do a second lookup to get Hibernate instances of them. Yes, it's less efficient overall, but it saves us the potential pain of these various edge cases where manually building the object doesn't work completely for one reason or another.

If there are alternate means of obtaining Hibernate proxies which can be used to fully hydrate objects as needed, I'm also open to exploring those options.

Ceiu (changes requested):

src/main/java/org/candlepin/pki/certs/AnonymousCertificateGenerator.java
```diff
@@ -317,20 +313,19 @@ private Set<X509Extension> prepareV3ByteExtensions(org.candlepin.model.dto.Produ
...
-        log.info("Generating SCA payload for consumer \"{}\"...", consumerUuid);
+        log.info("Generating ACCA payload for consumer \"{}\"...", consumerUuid);
 
         Product engProduct = new Product()
-            .setId("content_access")
-            .setName(" Content Access");
-
-        ownerContent.forEach(engProduct::addContent);
+            .setId("anonymous_cloud_content_access")
+            .setName(" Anonymous Cloud Content Access")
```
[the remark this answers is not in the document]
bakajstep: done

src/main/java/org/candlepin/model/ContentCurator.java
bakajstep: so I took the route of another query.
Done

src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -645,63 +645,50 @@ public Set<Content> getChildrenContentOfProductsByUuids(Collection<String> produ
...
-            contentQuery.setParameter("product_uuids", block)
-                .getResultList()
-                .forEach(col -> {
-                    Content content = (Content) ((Object[]) col)[0];
-                    Boolean enabled = (Boolean) ((Object[]) col)[1];
-                    activeContent.merge(content, enabled, (v1, v2) ->
-                        v1 != null && v1.booleanValue() ? v1 : v2);
-                });
+        for (Object[] result : results) {
+            Content content = this.getByUuid((String) result[0]);
```
Ceiu, line 690: This will N+1 us to death. When I mentioned "another query" I meant doing the secondary lookup in bulk. How this will end up looking is something like this:

```
List<ProductContent> output = new ArrayList<>();
Map<String, Boolean> contentEnablementMap = entityManager.createNativeQuery(sql)
  .setParameter(...)
  .getResultStream()
  .collect(Collectors.toMap(row -> (String) row[0], row -> (boolean) row[1], (ele1, ele2) -> ele1 || ele2));

String contentLookupJPQL = "SELECT content FROM Content content " +
  "WHERE content.uuid IN (:content_uuids)";

TypedQuery<Content> contentQuery = entityManager.createQuery(contentLookupJPQL, Content.class);

for (List<String> block : this.partition(contentEnablementMap.keySet())) {
  contentQuery.setParameter("content_uuids", block)
    .getResultStream()
    .map(content -> new ProductContent(content, contentEnablementMap.get(content.getUuid())))
    .forEach(output::add);
}

return output;
```

Note that there are a few decisions to make in there (map construction, collision resolution, how the output should be collected, etc.); some of which should be properly documented so there aren't any surprises for callers.

Ceiu (changes requested):

src/main/java/org/candlepin/model/ContentCurator.java
bakajstep: done

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -345,16 +358,15 @@ private Set<X509Extension> prepareV3ByteExtensions(org.candlepin.model.dto.Produ
...
 
         String consumerUuid = consumer != null ? consumer.getUuid() : null;
         log.info("Generating SCA payload for consumer \"{}\"...", consumerUuid);
 
         Product engProduct = new Product()
             .setId("content_access")
-            .setName(" Content Access");
-
-        ownerContent.forEach(engProduct::addContent);
+            .setName(" Content Access")
```
bakajstep, line 368: Is this sapece also accident? @nikosmoum

nikosmoum: Yes, but we cannot change this now because SCA certs have been like this for years, and people out there may be parsing them and expecting this space to be there. The anonymous certs aren't 'released' anywhere yet, that's why we can change them

bakajstep: ack, thanks

src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -645,63 +646,65 @@ public Set<Content> getChildrenContentOfProductsByUuids(Collection<String> produ
...
-            contentQuery.setParameter("product_uuids", block)
-                .getResultList()
-                .forEach(col -> {
-                    Content content = (Content) ((Object[]) col)[0];
-                    Boolean enabled = (Boolean) ((Object[]) col)[1];
-                    activeContent.merge(content, enabled, (v1, v2) ->
-                        v1 != null && v1.booleanValue() ? v1 : v2);
-                });
+            .getResultList()
+            .stream()
```
Ceiu, line 688: You should be able to call `.getResultStream()` here instead of `.getResultList().stream()`.

src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -645,63 +646,65 @@ public Set<Content> getChildrenContentOfProductsByUuids(Collection<String> produ
...
+
+        String contentLookupJPQL =
+            "SELECT content FROM Content content WHERE content.uuid IN (:content_uuids)";
+        TypedQuery<Content> contentQuery = entityManager.createQuery(contentLookupJPQL, Content.class);
+
+        // This step is necessary to handle large sets of UUIDs and avoid query parameter limits
+        for (List<String> block : this.partition(contentEnablementMap.keySet())) {
+            contentQuery.setParameter("content_uuids", block);
+            List<Content> contents = contentQuery.getResultList();
+            contents.stream()
```
Ceiu, line 705: Same with this. The whole operation should be chainable here from contentQuery down:

```
contentQuery.setParam(...)
  .getResultStream()
  .map(...)
  ...
```

Ceiu (changes requested):

src/main/java/org/candlepin/model/ContentCurator.java
bakajstep: i was looking for that, but for that you need JPA api 2.2 and we are using 2.1 so we need to use this option, or maybe I missed something.

src/main/java/org/candlepin/model/ContentCurator.java
bakajstep: Answer above and done

src/main/java/org/candlepin/model/ContentCurator.java
Ceiu: We can bump the API version to 2.2 without running into the package name change:

`hibernate = { module = "javax.persistence:javax.persistence-api", version = "2.2" }`

bakajstep: done

Ceiu (approved):

PR #4759, opened 2024-07-02
[M] CANDLEPIN-871: Added CS8 container to CI

bakajstep:


containers/cs8.Containerfile
```diff
@@ -0,0 +1,89 @@
...
+COPY ${WAR_FILE} ./candlepin.war
+
+# Prepare development certs
+RUN mkdir -p /app/certs
+WORKDIR /app/certs
+COPY ./bin/deployment/gen_certs.sh .
+RUN ./gen_certs.sh --cert_out ./candlepin-ca.crt --key_out ./candlepin-ca.key --hostname candlepin; \
+    rm gen_certs.sh;
+
+
```
joshmalbrecht, line 34: Extra new line

joshmalbrecht (changes requested):
I just noticed an extra new line in the new containerfile.

I'm also wondering if this containerfile should live in `.github/containers` as it seems that this containerfile is to only be used for spec tests that use the Postgres database.

containers/cs8.Containerfile
bakajstep: done

bakajstep:
> I just noticed an extra new line in the new containerfile.
> 
> I'm also wondering if this containerfile should live in `.github/containers` as it seems that this containerfile is to only be used for spec tests that use the Postgres database.

Thats a good question. I was thinking about that too. We can have containers on one place in `containers` folder or have one in `containers` and other in `.github/containers`, but then we would have `Containerfiles` for CI on two diffrent places. Maybe @nikosmoum what dou you think about that?

nikosmoum:
I think moving the cs8 containerfile to `.github/containers` make sense, since the main `/containers` directory is for containers that end up in shippable images (to quay). So it makes sens to keep them in separate places

bakajstep:
Moved to `./.github/containers/`

joshmalbrecht (approved):

PR #4762, opened 2024-07-03
[M] Updated deprecated gradle step in CI

bakajstep:
Gradle actions will be deprecated and triggering gradle actions using arguments will be removed in v4 recommended approach for projects that include gradlew wrapper to use that.
[Deprecating gradle
](https://github.com/gradle/actions/blob/main/docs/deprecation-upgrade-guide.md#the-action-gradlegradle-build-action-has-been-replaced-by-gradleactionssetup-gradle) [Recommended approach](https://github.com/gradle/actions/blob/main/docs/deprecation-upgrade-guide.md#using-the-action-to-execute-gradle-via-the-arguments-parameter-is-deprecated)

### DIFF
--- a/.github/workflows/checkstyle.yml
+++ b/.github/workflows/checkstyle.yml
@@ -38,7 +38,8 @@ jobs:
           distribution: ${{ env.JAVA_DISTRIBUTION }}
           java-version: ${{ env.JAVA_VERSION }}
 
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v3
+
       - name: Run checkstyle
-        uses: gradle/gradle-build-action@v3
-        with:
-          arguments: checkstyle
+        run: ./gradlew checkstyle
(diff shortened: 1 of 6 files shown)


nikosmoum (commented):
Based on the link you added (the example here specifically: https://github.com/gradle/actions/blob/main/docs/deprecation-upgrade-guide.md#project-uses-gradle-wrapper), you should replace the `gradle/gradle-build-action@v3` that is everywhere, with an initial `uses: gradle/actions/setup-gradle@v3` and then just run `./gradlew` commands, but this PR never sets the `uses: gradle/actions/setup-gradle@v3`. Why is that?

bakajstep:
> Based on the link you added (the example here specifically: https://github.com/gradle/actions/blob/main/docs/deprecation-upgrade-guide.md#project-uses-gradle-wrapper), you should replace the `gradle/gradle-build-action@v3` that is everywhere, with an initial `uses: gradle/actions/setup-gradle@v3` and then just run `./gradlew` commands, but this PR never sets the `uses: gradle/actions/setup-gradle@v3`. Why is that?

Yes, you are right. I forgot to add that step.

nikosmoum (approved):

PR #4769, opened 2024-07-09
[M] CANDLEPIN-885: Upgraded to openapi-generator 7.7.0

bakajstep:
- Removed the `withXml` property from the `generateApiDocs` and `generateOpenApiJson` tasks.
- `withXml` option was deleted in version 7.7.0, because the value was ignored.

https://github.com/OpenAPITools/openapi-generator/pull/18568


bakajstep:
Also checked mustache files and there are no chagers excepts comments and new line on end of the file

nikosmoum (approved):

PR #4771, opened 2024-07-10
[M] CANDLEPIN-888: Removed encoding of slashes in environment name

joshmalbrecht:
- Updated the SCACertificateGenerator to not encode slashes in the environment name when creating the content paths.

### DIFF
--- a/src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
+++ b/src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
@@ -289,17 +289,28 @@ public class SCACertificateGenerator {
     }
 
     private Content createContent(Owner owner, Environment environment) {
-        Content dContent = new Content();
-
         String path = "";
         if (owner != null) {
             path += "/" + Util.encodeUrl(owner.getKey());
         }
         if (environment != null) {
-            path += "/" + Util.encodeUrl(environment.getName());
+            StringBuilder envPrefix = new StringBuilder();
+            for (String chunk : environment.getName().split("/")) {
+                if (chunk.isEmpty()) {
+                    continue;
+                }
+
+                envPrefix.append("/");
+                envPrefix.append(Util.encodeUrl(chunk));
+            }
+
+            path += envPrefix.toString();
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -296,7 +296,12 @@ private Content createContent(Owner owner, Environment environment) {
             path += "/" + Util.encodeUrl(owner.getKey());
         }
         if (environment != null) {
-            path += "/" + Util.encodeUrl(environment.getName());
+            String environmentName = environment.getName();
+            if (environmentName != null) {
+                environmentName = environmentName.replaceAll("[^a-zA-Z0-9._-]", "");
```
Ceiu, line 301: Are we certain removing the character is the right choice here? Seems like the intent is to have the slash recognized as a path segment separator.

joshmalbrecht: Good call! Verified that the slash in the environment name should not be removed and recognized as a path segment separator.

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -296,7 +296,17 @@ private Content createContent(Owner owner, Environment environment) {
...
-            path += "/" + Util.encodeUrl(environment.getName());
+            StringBuilder envPrefix = new StringBuilder();
+            for (String chunk : environment.getName().split("/")) {
+                if (chunk.isEmpty()) {
+                    continue;
+                }
+
+                envPrefix.append("/");
+                envPrefix.append(Util.encodeUrl(chunk));
+            }
```
joshmalbrecht, line 307: This is the logic that was removed from the following:
https://github.com/candlepin/candlepin/pull/4333/files#diff-6328ae639f7b8abb7bb741f14992daffb9f98d6a73d7fd60f3b6d0aac71e343fL89-L97

Ceiu:
Closing in favor of a forthcoming PR

PR #4773, opened 2024-07-12
[M] CANDLEPIN-64:Updated AbstractHibernateCurator not use Hibernate Criteria

bakajstep:
- Updated AbstractHibernateCurator to stop using Hibernate APIs,such as Hibernate criteria and Hibernate implementations of JPA interfaces
- Updated follow-up classes
- Deleted unused SessionWrapper class


src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -1080,24 +706,10 @@ public void refresh(E... entities) {
     }
 
     public E evict(E entity) {
-        this.currentSession().evict(entity);
+        this.getEntityManager().detach(entity);
         return entity;
     }
```
joshmalbrecht, line 711: This method is now the same as `AbstractHibernateCurator.detach(Entity)`. I think there are some differences between these two methods. I believe that the `AbstractHibernateCurator.detach(Entity)` method should disregard modifications to an entity by removing it from the persistence context and the  `AbstractHibernateCurator.evict(Entity)` method should remove the entity from the Hibernate first level cache.

So, I think this method should still be:
```
  public E evict(E entity) {
      this.currentSession().evict(entity);
      return entity;
  }
```

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -996,24 +615,32 @@ public Collection<E> updateAll(Collection<E> entities, boolean flush, boolean ev
...
+                        }
                     }
 
                     if (flush) {
                         em.flush();
 
                         if (evict) {
                             for (E entity : block) {
-                                session.evict(entity);
+                                em.detach(entity);
```
joshmalbrecht, line 643: I think that this should still be `session.evict` for the same reason that I left on the comment for AbstractHibernateCurator.evict. @Ceiu  am I understanding the difference between `Session.evict` and `EntityManager.detach` correctly?

joshmalbrecht (changes requested):
I had some small questions and a couple concerns.

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -130,17 +111,28 @@ public String getDatabaseDialect() {
...
+        Predicate securityPredicate = this.getSecurityPredicate(clazz, cb, root);
+        Predicate idPredicate = cb.equal(root.get("id"), id);
+
+        if (securityPredicate != null) {
+            query.select(root).where(cb.and(securityPredicate, idPredicate));
+        }
+        else {
+            query.select(root).where(idPredicate);
+        }
+        try {
```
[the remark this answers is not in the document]
bakajstep: done

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1116,11 +1157,22 @@ SELECT COUNT(c) FROM Consumer c
...
+
+        if (securityPredicate != null) {
+            query.select(cb.count(consumerRoot.get("id"))).where(cb.and(securityPredicate, uuidPredicate));
+        }
+        else {
+            query.select(cb.count(consumerRoot.get("id"))).where(uuidPredicate);
+        }
 
+        Long result = em.createQuery(query).getSingleResult();
         return result != 0;
```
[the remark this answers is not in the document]
bakajstep: done

src/main/java/org/candlepin/model/ManifestFileRecordCurator.java
```diff
@@ -51,17 +52,33 @@ public boolean deleteById(String id) {
...
+    private Blob createBlob(File fileToStore) throws IOException {
+        FileInputStream inputStream = new FileInputStream(fileToStore);
+        byte[] fileBytes = new byte[(int) fileToStore.length()];
+        inputStream.read(fileBytes);
+        inputStream.close();
+
+        try {
+            return new SerialBlob(fileBytes);
+        }
+        catch (Exception e) {
```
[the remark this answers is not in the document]
bakajstep: done

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
nikosmoum: Inspecting the decompiled code that implements EntityManager.detach() (implemented by Hibernate's SessionImpl class) it looks like the detach method is just calling evict underneath. They're doing the same thing

src/test/java/org/candlepin/model/CuratorPaginationTest.java
```diff
@@ -57,129 +51,6 @@ public void init() throws Exception {
             o.setKey(String.valueOf(i));
             ownerCurator.create(o);
         }
-
-        session = (Session) this.getEntityManager().getDelegate();
-    }
-
-    @Test
-    public void testPaging() {
```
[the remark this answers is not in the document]
bakajstep: these were tests for the old paging methods that I deleted and the new methods have their own

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -1156,14 +769,11 @@ private String getConcurrentModificationMessage() {
...
         Map<Serializable, E> entityMap = new TreeMap<>();
 
-        SessionImpl session = (SessionImpl) this.currentSession();
-        ClassMetadata metadata = session.getSessionFactory().getClassMetadata(this.entityType);
-
         // Step through and toss all the entities into our TreeMap, which orders its entries using
         // the natural order of the key. This will ensure that we have our entity collection sorted
         // by entity ID, which should help avoid deadlock by having a deterministic locking order.
         for (E entity : entities) {
-            entityMap.put(metadata.getIdentifier(entity, session), entity);
```
nikosmoum, line 1166: This seems to also have been written this way because we use the 'uuid' field as the primary key for Content and Product entities. However, if E ends up being either Content/Product, now after the Entity Layering change, these will end up being unique per the 'id' field, so we should still be getting a consistent locking order here. 

@Ceiu  is my assumption correct here?^

Ceiu: Nope. With layering+namespaces, the constraint with IDs is on namespace+ID, so this runs into the problem of potentially pulling more than one thing.

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
joshmalbrecht: Okay perfect! Good idea on decompiling the code.

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
nikosmoum: I can't seem to find a way to get this metadata through EntityManager, so this may be a case where we'll have to use the Hibernate session still

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -996,24 +616,32 @@ public Collection<E> updateAll(Collection<E> entities, boolean flush, boolean ev
         return entities;
     }
 
+    private boolean isEntityNew(E entity) {
+        return entity.getId() == null;
+    }
```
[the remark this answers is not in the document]
Ceiu: The general logic here is exactly what Hibernate+JPA does as well: check if the entity already has a primary key defined, if so: it is an existing entity, otherwise it is a new entity. We've leveraged it a couple times to work around weird edge cases with JPA throughout CP.

However, to Nikos' point, we don't always use the ID field as the primary key (despite it being what is defined by the `Persisted` interface), so this method won't work for all entities and yet it's extracted and named as if it's generic enough to work.

There are a few ways around this. AHC could define it as a public or protected method and let subclasses override it as necessary. Alternatively, we could use the clunky JPA interface to get it generically as we were before:

```
Object id = this.getEntityManager()
  .getEntityManagerFactory()
  .getPersistenceUnitUtil()
  .getIdentifier(entity);
```

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
Ceiu: In pure JPA, it's behind the `PersistenceUnitUtil`, which itself is fetchable through the `EntityManagerFactory`.

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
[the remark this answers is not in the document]
bakajstep: done

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
bakajstep: done

PR #4776, opened 2024-07-16
[M] Add a product to the 'Donald Duck' test organization

m-horky:
The test data are used by the subscription-manager integration tests. Since entitlement mode will be disabled, the tests are being migrated to SCA.

---

Specifically, I'm trying to fix the `testRegisterWithKey` that fails in https://github.com/candlepin/subscription-manager-cockpit/pull/70 (see the source code [failing line](https://github.com/candlepin/subscription-manager-cockpit/blob/5529ef7a04287facef6d229468944a875f077c4b/test/check-subscriptions#L317-L318)).

Will this fix our issue, will it show up in the list of installed products? I don't know how to test it directly, I don't have a control over the Candlepin that runs the Cockpit tests.

### DIFF
--- a/bin/deployment/test_data.json
+++ b/bin/deployment/test_data.json
@@ -650,7 +650,29 @@
             "name": "donaldduck",
             "displayName": "Donald Duck",
             "contentAccessModeList": "entitlement,org_environment",
-            "contentAccessMode": "org_environment"
+            "contentAccessMode": "org_environment",
+            "products": [
+                {
+                    "name": "Duck OS Premium Architecture Bits",
+                    "id": "6050",
+                    "version": "6.1",
+                    "arch": "ppc64",
+                    "provided_products": [
+                        "37060"
+                    ],
+                    "attributes": {
+                        "sockets": 2,
+                        "vcpu": 4,
+                        "warning_period": 30
+                    },
+                    "content": [
+                        [
+                            6000,
+                            false
+                        ]
+                    ]
+                }
+            ]
         }
     ],
     "users": [


Ceiu:
There's a lot to unpack here due to the legacy that surrounds the test data.

"Installed products" are a special thing having to do with the product certs that are present on the consuming system. While the test data importer *does* fetch product certs for "engineering" products, it does so from an admin context, not from a consumer context. So whether or not a given product counts as "installed" will depend on where the certs are written as specified by the `--cert_dir` CLI option, the operations performed by subman/cockpit, and/or the surrounding test infrastructure.

What this will do is create a product definition in Candlepin, but that's it. The definition does not appear to be flagged as a "marketing" product (or SKU), so the test data importer will not attempt to create a corresponding subscription for it within the org. But since it is flagged as an engineering product by virtue of not being a marketing product, its product cert will be fetched during the cert fetching step. If that's all y'all need, then this is change looks to be sufficient.

As for why the test in question is failing, I'm at a loss. The product doesn't have a subscription above on line 309, but does on 318 after unregistering? Or at least, the row exists in the table suggesting it's registered. I'd need to see how that table is being populated to have a better idea of what's going on there.

---

As a side note, since your tests are so tightly coupled to the test data, I would *highly* recommend making a copy of the test_data.json (i.e. subman_test_data.json), and then specifying it directly when importing data via `./bin/deployment/test_data_importer.py subman_test_data.json`. That way y'all are entirely decoupled from whatever we may do with the default file, and you won't need to wait on us to accept a PR and ship a build for the changes to land in your tests.

PR #4789, opened 2024-07-30
[M] CANDLEPIN-878: Updated Vagrant el8 to use AlmaLinux instead of CentOS

joshmalbrecht:
- Replaced Centos Stream 8 for AlmaLinux 8 for the el8 Vagrant box. This is due to the cs8 end of life changing the repository URLs and causing issues standing up the el8 Vagrant VM.

To test the changes, run `vagrant up el8`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -78,8 +78,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   end
 
   config.vm.define("el8", primary: true) do |vm_config|
-    vm_config.vm.box = "centos.cloud/centos8s"
-    vm_config.vm.box_url = "https://cloud.centos.org/centos/8-stream/x86_64/images/CentOS-Stream-Vagrant-8-latest.x86_64.vagrant-libvirt.box"
+    vm_config.vm.box = "almalinux/8"
     vm_config.vm.host_name = "candlepin-el8.example.com"
 
     # Vagrant allows to create a forwarded port mapping which allows access to a specific port
--- a/ansible/roles/candlepin/tasks/main.yml
+++ b/ansible/roles/candlepin/tasks/main.yml
@@ -9,7 +9,7 @@
   ansible.builtin.include_tasks:
     file: el8/el8_tasks.yml
   when:
-    - ansible_distribution in ["CentOS", "RedHat"]
+    - ansible_distribution in ["CentOS", "AlmaLinux", "RedHat"]
     - ansible_distribution_major_version == "8"
   tags:
     - always


Vagrantfile
```diff
@@ -77,22 +78,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
-    #
-    # Uncomment these lines to forward the Candlepin standard dev ports to this guest machine.
-    # vm_config.vm.networking "forwarded_port", protocol: "tcp", guest: 8080, host: 8080
-    # vm_config.vm.networking "forwarded_port", protocol: "tcp", guest: 8443, host: 8443
-    vm_config.vm.provision "shell", inline: "dnf update -y dnf ca-certificates"
-    configure_ansible_provisioning(vm_config)
-  end
-
-  config.vm.define("el9", autostart: false) do |vm_config|
+  config.vm.define("el9", primary: true) do |vm_config|
```
bakajstep, line 81: nitpick: if only one vm is specified in vagrantfil there is no need to set the primary machine

bakajstep, line 81: nitpick: if only one vm is specified in vagrantfil there is no need to set the primary machine

bakajstep (commented):
I don't know if it's a good idea to make it part of this task, but wouldn't it be a good idea to add `git config --global --add safe.directory /vagrant` to el9 so that we don't have to set it up every time we start?

Otherwise lgtm.

Vagrantfile
Ceiu: We'll be adding more boxes back in, so it's good to be explicit here

Ceiu:
If we're going to be adding a replacement el8 box, maybe it doesn't make sense to blast all these files and then recreate them again. In discussions, I was under the assumption that we would just be updating the vagrant file, but if that's not the case we should probably rethink it.

joshmalbrecht:
> If we're going to be adding a replacement el8 box, maybe it doesn't make sense to blast all these files and then recreate them again. In discussions, I was under the assumption that we would just be updating the vagrant file, but if that's not the case we should probably rethink it.

Yeah, I think it makes more sense to just update the el8 box in this PR! I'll update this PR

Vagrantfile
joshmalbrecht: Updated this PR to change el8 to use an AlmaLinux 8 box instead of cs8.

joshmalbrecht:
Moving this PR back to 'ready for review' as I have changed the PR to simply update the el8 Vagrant VM definition to use an AlmaLinux 8 box.

joshmalbrecht:
> I don't know if it's a good idea to make it part of this task, but wouldn't it be a good idea to add git config --global --add safe.directory /vagrant to el9 so that we don't have to set it up every time we start?

@bakajstep I think it would make sense to open a separate PR for that since it is fairly different than these changes.

Vagrantfile
```diff
@@ -78,8 +78,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   end
 
   config.vm.define("el8", primary: true) do |vm_config|
-    vm_config.vm.box = "centos.cloud/centos8s"
-    vm_config.vm.box_url = "https://cloud.centos.org/centos/8-stream/x86_64/images/CentOS-Stream-Vagrant-8-latest.x86_64.vagrant-libvirt.box"
+    vm_config.vm.box = "almalinux/8"
+    vm_config.vm.box_url = "https://app.vagrantup.com/almalinux/boxes/8/versions/8.10.20240604/providers/libvirt/amd64/vagrant.box"
```
nikosmoum, line 82: Let's remove the box_url variable here, because it points to a specific version. If we just live it to the box name (almalinux/8), we'll always get the latest.

The only reason we set the box_url for centos images was because the Centos images in vagrant weren't actually updated frequently by the centos devs, so we had to point to `cloud.centos.org`.

joshmalbrecht: Updated

ansible/roles/candlepin/tasks/main.yml
```diff
@@ -9,7 +9,7 @@
   ansible.builtin.include_tasks:
     file: el8/el8_tasks.yml
   when:
-    - ansible_distribution in ["CentOS", "RedHat"]
+    - ansible_distribution in ["CentOS", "AlmaLinux", "RedHat"]
```
bakajstep, line 12: Do we need to still have here CentOS?

nikosmoum: no, lets remove it

joshmalbrecht: I figured might as well leave it in there because we should still support using CentOS. We just had to switch to AlmaLinux because the CentOS 8 box has the broken repo urls at the moment, but if they ever released a CentOS 8 box with the URLs updated, then people could use that box still.

PR #4792, opened 2024-07-30
[M] CANDLEPIN-901: Added missing validCheckSum to the entity layering changeset

Ceiu:
- Added missing validCheckSum tags to the entity layering changeset to
  prevent Liquibase migration errors that could occur if an environment
  deployed Candlepin 4.4.2 to 4.4.9, and then attempted to deploy 4.4.10
  or later


Ceiu:
Note that this is only a speculative fix for the hanging we're seeing some users hit. It *probably* won't fix them, but it should fix anyone else who happened to fall into the triggering condition.

Also of note/concern: sometimes Liquibase just doesn't fail this changeset with a checksum diff error. I could not determine why, nor could I even find any specific behavior or state that could be correlated.

bakajstep (approved):

nikosmoum (approved):

PR #4793, opened 2024-07-30
[M] CANDLEPIN-901: Rolled back to previous version of Apache Commons lang

Ceiu:
- Rolled back to version 3.14 of commons-lang from Apache Commons to work around an issue in Liquibase which could cause Candlepin to hang on startup on systems which do not have high entropy for the purposes of generating random data via /dev/random


Ceiu:
For reviewers:
This can be checked by running `./gradlew dependencies` and ensuring all instances of `commons-lang3` resolve to 3.14, and 3.15 does not appear at all.

bakajstep:
> For reviewers: This can be checked by running `./gradlew dependencies` and ensuring all instances of `commons-lang3` resolve to 3.14, and 3.15 does not appear at all.

I just checked and it's version 3.14 everywhere except checkstyle where it has this dependency in version 3.8.1

nikosmoum (approved):

nikosmoum:
For posterity, the issue has been reported against Apache Commons Lang3 here: https://issues.apache.org/jira/browse/LANG-1748

PR #4794, opened 2024-07-31
[M] CANDLEPIN-893: Added API for listing consumers in a specific environment

bakajstep:
- Enhanced 'GET /consumers' API to support 'environmentId' parameter.

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -734,6 +734,11 @@ paths:
             items:
               type: string
               format: key:value
+        - name: environment_id
+          in: query
+          description: Environment identifier
+          schema:
+            type: string
         - $ref: "#/components/parameters/paging_page"
         - $ref: "#/components/parameters/paging_per_page"
         - $ref: "#/components/parameters/paging_order"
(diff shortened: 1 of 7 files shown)


nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceSpecTest.java
```diff
@@ -921,12 +925,101 @@ public void shouldProperlyEscapeValuesToAvoidSqlInjection() {
...
+            .doesNotContain(consumer2.getId());
+    }
+
+    @Test
+    public void shouldOnlyFindConsumersForGivenEnvironmentIdAndOwner() {
+        OwnerDTO anotherOwner = adminClient.owners().createOwner(Owners.random());
+        EnvironmentDTO environment1 = adminClient.owners().createEnvironment(
+            owner.getKey(), Environments.random());
+        EnvironmentDTO environment2 = adminClient.owners().createEnvironment(
+            anotherOwner.getKey(), Environments.random());
```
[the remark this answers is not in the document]
bakajstep: no, we would hit: `duplicate key value violates unique constraint`

api/candlepin-api-spec.yaml
```diff
@@ -734,6 +734,11 @@ paths:
             items:
               type: string
               format: key:value
+        - name: environmentId
```
nikosmoum, line 737: I think we use snake case instead of camel case for API parameters, so lets change this to environment_id

bakajstep: done

PR #4795, opened 2024-07-31
[M] Added owner and group options to synced folder configuration

bakajstep:
- Added Git safe directory configuration using git config command

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -58,7 +58,7 @@ end
 
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  config.vm.synced_folder ".", "/vagrant", type: "sshfs"
+  config.vm.synced_folder ".", "/vagrant", type: "sshfs", :owner => 'vagrant', :group => 'vagrant'
   config.vm.host_name = "candlepin.example.com"
   config.ssh.forward_agent = true
 


Ceiu:
What is this doing?

bakajstep:
> What is this doing?

The git config --global --add safe.directory /vagrant command is needed to add /vagrant to the Git safe directory list. This is especially useful in situations where you are using Git in an environment where directories are not considered safe by default, such as when working in shared folders in virtualized environments like Vagrant.

When you first start vm until you add the folder as safe, it doesn't let you start candlepin. This will do it for us on the first start-up.

You can test this:
In main:
```
vagrant destroy el8(el9)
vagrant up el8(el9)
vagrant ssh el8(el9)
```
then when you will try to start candlepin you will need to run this command: `git config --global --add safe.directory /vagrant`
otherwise it doesn't let you start candlepin

If you do these steps in this branch it will let you start candlepin because the folder will already be marked as safe

Ceiu:
I'm curious what about your environment or flow differs from mine, because I've never had to deal with safe directories before

bakajstep:
> I'm curious what about your environment or flow differs from mine, because I've never had to deal with safe directories
> 
> Edit: depending on what that difference is, this may be better off as a custom dev task. I'm not sure I want to be flagging directories arbitrarily without fully understanding exactly what's happening.

and have you tried vagrant destroy? because git has this feature since version 2.35.2. It was introduce like 1 and half years ago. I remeber it was one of my first task on jenkins pipeline.

nikosmoum:
I've done destroy/up/ssh multiple times lately, and I've never run into any issues with safe directories. As an example my el9 vm has git-2.43.0-1.el9.x86_64.

Can you explain the exact steps and what is the error you're seeing?

Ceiu:
Same. I've been rebuilding boxes quite frequently as recently as yesterday and I've never run into this issue.

bakajstep:
step when i have 100% reproducer:
```
 $ vagrant destroy 
 $ vagrant up
 $ vagrant ssh
 $ cd /vagrant
 $ ./bin/deployment/deploy -ga
```
 Output:
```
 (08:37:40) vagrant@candlepin-el9:/vagrant  $ ./bin/deployment/deploy -ga
fatal: detected dubious ownership in repository at '/vagrant'
To add an exception for this directory, call:

        git config --global --add safe.directory /vagrant
./bin/deployment/deploy: line 250: /bin/deployment/bash_functions: No such file or directory
```

bakajstep:
Nikos and I just found out that this only occurs with me because the files in the shared folder have an owner and a group as a user from the host. Because I have two accounts (personal, work) where I created the work one later so it had 1001 and no user with this id existed in the vm and there was an error with the ownership. with this change the files should always belong to the user vagrant and this problem should not occur anymore.

Ceiu (approved):

PR #4798, opened 2024-08-01
[M] CANDLEPIN-898: Create EventAdapter interface

joshmalbrecht:
- Created the EventAdapter interface.
- Updated CandlepinContextListener class to call the EventAdapter startup and shutdown method during the application lifecycle.

### DIFF
--- a/src/main/java/org/candlepin/guice/CandlepinContextListener.java
+++ b/src/main/java/org/candlepin/guice/CandlepinContextListener.java
@@ -29,6 +29,7 @@ import org.candlepin.logging.LoggingConfigurator;
 import org.candlepin.messaging.CPMContextListener;
 import org.candlepin.resteasy.MethodLocator;
 import org.candlepin.resteasy.ResourceLocatorMap;
+import org.candlepin.service.EventAdapter;
 
 import com.google.inject.AbstractModule;
 import com.google.inject.Injector;
@@ -108,6 +109,7 @@ public class CandlepinContextListener extends GuiceResteasyBootstrapServletConte
 
     private ActiveMQContextListener activeMQContextListener;
     private JobManager jobManager;
+    private EventAdapter eventAdapter;
     private LoggerContextListener loggerListener;
 
     // a bit of application-initialization code. Not sure if this is the
@@ -206,6 +208,9 @@ public class CandlepinContextListener extends GuiceResteasyBootstrapServletConte
 
         loggerListener = injector.getInstance(LoggerContextListener.class);
 
+        eventAdapter = injector.getInstance(EventAdapter.class);
+        eventAdapter.initialize();
+
         this.injector = injector;
     }
 
@@ -249,6 +254,8 @@ public class CandlepinContextListener extends GuiceResteasyBootstrapServletConte
                 log.info("Failed to de-registering driver {}", driver, e);
             }
         }
+
+        eventAdapter.shutdown();
 
         if (config.getBoolean(ACTIVEMQ_ENABLED)) {
             activeMQContextListener.contextDestroyed(injector);
--- a/src/main/java/org/candlepin/service/EventAdapter.java
+++ b/src/main/java/org/candlepin/service/EventAdapter.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2009 - 2024 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service;
+
+import org.candlepin.service.model.AdapterEvent;
+
+/**
+ * An interface for publishing Candlepin {@link AdapterEvent}s.
+ */
+public interface EventAdapter {
+
+    /**
+     * An application lifecycle method called during the initialization sequence.
+     */
+    void initialize();
+
+    /**
+     * An application lifecycle method called during the shutdown sequence.
+     */
+    void shutdown();
+
+    /**
+     * Publishes the provided {@link AdapterEvent} using the event's type and body. The provided event type on
+     * the {@link AdapterEvent} provides a mechanism for conditionally routing the events. If the provided
+     * event is null, event body is null, or the event type is null, the publish operation should not be
+     * performed.
+     *
+     * @param event
+     *  an event to be published
+     */
+    void publish(AdapterEvent event);
+}
--- a/src/main/java/org/candlepin/service/impl/DefaultEventAdapter.java
+++ b/src/main/java/org/candlepin/service/impl/DefaultEventAdapter.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2009 - 2024 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service.impl;
+
+import org.candlepin.service.EventAdapter;
+import org.candlepin.service.model.AdapterEvent;
+
+public class DefaultEventAdapter implements EventAdapter {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void initialize() {
+        // Intentionally left blank
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void shutdown() {
+        // Intentionally left blank
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void publish(AdapterEvent event) {
+        // Intentionally left blank
+    }
+
+}
--- a/src/main/java/org/candlepin/service/model/CloudCheckInEvent.java
+++ b/src/main/java/org/candlepin/service/model/CloudCheckInEvent.java
@@ -1,0 +1,230 @@
+/*
+ * Copyright (c) 2009 - 2024 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service.model;
+
+import org.candlepin.model.Consumer;
+import org.candlepin.model.ConsumerCloudData;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+import java.util.Date;
+import java.util.List;
+
+/**
+ * An event where a cloud consumer has updated their last check-in time stamp.
+ */
+public class CloudCheckInEvent implements AdapterEvent {
+
+    private static final String SYSTEM_UUID_KEY = "systemUuid";
+    private static final String CHECK_IN_KEY = "checkIn";
+    private static final String PROVIDER_KEY = "cloudProviderId";
+    private static final String ACCOUNT_KEY = "cloudAccountId";
+    private static final String CLOUD_OFFERINGS_KEY = "cloudOfferingIds";
+
+    private String consumerUuid;
+    private Date checkIn;
+    private String cloudProviderId;
+    private String cloudAccountId;
+    private List<String> cloudOfferingIds;
+
+    private String eventBody;
+
+    /**
+     * Creates a new CloudCheckInEvent instance using the provided {@link ConsumerCloudData}.
+     *
+     * @param cloudData
+     *  the consumer cloud data used to populate the event
+     *
+     * @param mapper
+     *  the object mapper used to serialize the event body
+     *
+     * @throws IllegalArgumentException
+     *  if the provided consumer cloud data or object mapper is null or if the consumer cloud data has no
+     *  cloud provider shortname, cloud account ID, consumer, or has a consumer that does not have a UUID, or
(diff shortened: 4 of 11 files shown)


src/main/java/org/candlepin/guice/CandlepinContextListener.java
```diff
@@ -250,6 +254,9 @@ private void destroySubsystems() throws Exception {
             }
         }
 
+        EventAdapter eventAdapter = injector.getInstance(EventAdapter.class);
```
Ceiu, line 257: Are we certain the scope and lifecycle of this needs to be managed by us at this level? I would expect that the provider of such a thing would be handling that transparently.

joshmalbrecht: I'm not sure I understand what you mean.  Unless I am missing something, CandlepinContextListener seems to be the only place to perform logic during application startup and shutdown. I don't believe (unless I was missing something) Guice provides a way for hook into the lifecycle of classes in comparison to Spring with the `PostConstruct` and `PreDestroy` annotations.

joshmalbrecht: So, the purpose of these lifecycle events is to establish and close the `javax.jms.Connection` which is essentially a TCP connection with SSL when using the `ActiveMQSslConnectionFactory` class to establish the connection.  This is based on the `javax.jms.Connection` documentation:

https://docs.oracle.com/javaee%2F7%2Fapi%2F%2F/javax/jms/Connection.html
> It encapsulates an open connection with a JMS provider. It typically represents an open TCP/IP socket between a client and the service provider software.

So, failure to close the TCP connection on our application shutdown could result in our client leaking inactive connection and consuming from the ActiveMQ maximumConnections count.

Changes made:
- I updated CandlepinContextListener to only initialize and shutdown one instance of `EventAdapter`.
- I updated `EventAdapter.startup` to `EventAdapter.initialize`.

src/main/java/org/candlepin/service/model/Event.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package org.candlepin.service.model;
+
+/**
+ * The Event interface represents a Candlepin event.
+ */
+public interface Event extends ServiceAdapterModel {
+
+    /**
+     * @return the event serialized in a format such as JSON.
```
Ceiu, line 23: Docblock needs more details here to define what is expected from a proper implementation. Is null something it can return? Is there a way to determine the output format without parsing the output directly? Are there any lifecycle concerns an implementer or user would need to be aware of? etc.

Also, as a grammar thing, the existing line isn't a complete sentence (and annotated lines generally won't be) so it shouldn't end with a period.

src/main/java/org/candlepin/service/model/Event.java
```diff
@@ -0,0 +1,32 @@
...
+     */
+    String getBody();
+
+    /**
+     * @return the event's type
```
Ceiu, line 28: Same here. When we're defining an interface, we should be very specific as to what is happening with regards to the input, output, and expected lifecycle or state management of the implementations.

src/main/java/org/candlepin/service/impl/DefaultEventAdapter.java
```diff
@@ -0,0 +1,37 @@
...
+ * in this software or its documentation.
+ */
+package org.candlepin.service.impl;
+
+import org.candlepin.service.EventAdapter;
+import org.candlepin.service.model.Event;
+
+public class DefaultEventAdapter implements EventAdapter {
+
+    @Override
```
Ceiu, line 22: Also add inherited docblocks here for anyone using an IDE which pulls javadoc:

```
/**
 * {@inheritDoc}
 */
```

src/main/java/org/candlepin/service/EventAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    void initialize();
+
+    /**
+     * An application lifecycle method called during the shutdown sequence.
+     */
+    void shutdown();
+
+    /**
+     * Publishes the provided event.
```
Ceiu, line 35: What should implementers do if event is null? We need to describe more of the intended behavior for all cases, not just the general case.

Ceiu (changes requested):
Github isn't letting me comment on the line I'd like, so my apologies for not attaching this comment on the code itself:

Can we add some TODO/FIXME comments on the definition of the CandlepinContextListenerTest? That thing's a mess -- we're mocking in weird ways, but not enough that we avoid having to backup and restore drivers in a weird way? It's not following either paradigm and it's really awful to read.

src/main/java/org/candlepin/service/impl/DefaultEventAdapter.java
joshmalbrecht: Updated

src/main/java/org/candlepin/service/model/Event.java
joshmalbrecht: Updated to hopefully make the interface more clear

src/main/java/org/candlepin/service/EventAdapter.java
joshmalbrecht: Updated to hopefully make the interface more clear

src/main/java/org/candlepin/service/model/Event.java
joshmalbrecht: Yeah, you are right. The docs for these interfaces are unclear. For this method, I thought it might make more sense to return an `EventBody` object that encapsulates the serialized text along with the serialization type. That way we can enforce the null and empty string constraints of the event's body in that class instead of hoping the implementers of the Event interface follow the instructions on the JavaDoc.

joshmalbrecht:
> Can we add some TODO/FIXME comments on the definition of the CandlepinContextListenerTest? That thing's a mess -- we're mocking in weird ways, but not enough that we avoid having to backup and restore drivers in a weird way? It's not following either paradigm and it's really awful to read.

I agree! I added a TODO comment in the `CandlepinContextListenerTest` class.

src/main/java/org/candlepin/service/model/EventBody.java
```diff
@@ -0,0 +1,114 @@
...
+        if (body == null) {
+            throw new IllegalArgumentException("event body is null");
+        }
+
+        if (body.isBlank()) {
+            throw new IllegalArgumentException("event body is blank");
+        }
+
+        this.body = body;
+        this.serializationType = Objects.requireNonNull(serializationType);
```
Ceiu, line 53: Is there a reason that this method uses IAEs for body, but NPEs for serialization type?

src/main/java/org/candlepin/service/model/EventBody.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        this.serializationType = type;
+        return this;
+    }
+
+    /**
+     * @return the type of serialization used to create the body text
+     */
+    public SerializationType getSerializationType() {
```
Ceiu, line 110: The serialization type looks to be a function of the underlying event -- either the specific event itself, or the eventing platform/interface. It doesn't look like something that will be changing arbitrarily to need this kind of behavior.

What was the intent behind separating the event body into the (string, type enum) pair for every event over something like having both be available on the event itself or declaring the format as part of the event API?

src/main/java/org/candlepin/service/model/EventBody.java
joshmalbrecht: Good point, we should be consistent here. Updated the null check on serializationType to return an IAE instead of NPE and updated the javadoc.

src/main/java/org/candlepin/service/model/EventBody.java
joshmalbrecht: The intention was being able to enforce that the serialized event string will not be null or blank and the serialization type will not be null because of the validation done in the EventBody class.  We could have the serialized body string and the serialization type be returned by the event interface and have java docs explaining that the values should not be null/blank. 

I don't feel too strongly either way, but that was my thought process on breaking the EventBody into another class.

joshmalbrecht: After finally revisiting this, I think it makes more sense to just have the `getBody()` and `getSerializationType()` methods on the `Event` interface instead of having an `EventBody` object.  Updated

src/main/java/org/candlepin/service/model/Event.java
```diff
@@ -0,0 +1,48 @@
...
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.service.model;
+
+/**
+ * The Event interface represents a Candlepin event.
```
nikosmoum, line 18: Something that's somewhat bothering me is that we have duplicate class names in 2 packages now (Event/EventAdapter exist in both org.candlepin.audit and org.candlepin.service), and we have nothing that explains their difference. I wouldn't go as far as changing the class names, but at the very least adding something to the class-level javadocs that explains what these specific events are for, but I can't come up with something good enough.

The best I can think of here for example is: `The Event interface represents a Candlepin event for consumption by external components`. 'external' as opposed to Artemis events being used for candlepin-internal purposes such as the job system, and audit log, but the problem is that even Artemis events are sometimes used by external components like Katello, so its not really accurate.

If you can think of something better, you can update it, but if not ignore my comment

src/main/java/org/candlepin/service/model/EventType.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.candlepin.service.model;
+
+/**
+ * An enum that defines the types of Candlepin {@link Event}s.
+ */
+public enum EventType {
+
+    /**
+     * An event where a cloud system has checked-in.
```
nikosmoum, line 23: It would be good to describe what exactly is a check-in in this context (that the Consumer.lastCheckin timestamp was updated).

src/main/java/org/candlepin/service/model/Event.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public interface Event extends ServiceAdapterModel {
```
nikosmoum, line 20: Shouldn't we also be creating the implementation of the CloudCheckInEvent class for this interface as part of this task? (In the next task ConsumerManager.updateLastCheckIn would be creating an instance of it, so it could also be created as part of that one?)

nikosmoum (changes requested):

src/main/java/org/candlepin/service/model/Event.java
joshmalbrecht: Oh good catch! I thought that the implementation was to be done in the ConsumerManager task (CANDLEPIN-899), but it is not listed on that task. It does make sense to add the implementation to this PR. I will get that added

src/main/java/org/candlepin/service/model/Event.java
joshmalbrecht: I agree that is confusing to have both an Event and EventAdapter in two different packages. I updated the class name to `AdapterEvent` to help make the distinction more clear, and I updated the class-level java doc. I couldn't think of a better description than the one you came up with, and so I updated it to that!

src/main/java/org/candlepin/service/model/EventType.java
joshmalbrecht: I agree! Updated

src/main/java/org/candlepin/service/model/Event.java
joshmalbrecht: Added the `CloudCheckInEvent` class

src/main/java/org/candlepin/service/model/CloudCheckInEvent.java
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * An event where a cloud consumer has updated their last check-in time stamp.
+ */
+public class CloudCheckInEvent implements AdapterEvent {
+
+    private static final String SYSTEM_UUID_KEY = "systemUuid";
+    private static final String CHECK_IN_KEY = "checkIn";
+    private static final String PROVIDER_KEY = "cloudProviderId";
+    private static final String ACCOUNT_KEY = "system-uuid";
```
nikosmoum, line 39: instead of `system-uuid` this should be `cloudAccountId`

src/test/java/org/candlepin/service/model/CloudCheckInEventTest.java
```diff
@@ -0,0 +1,382 @@
...
+
+import java.util.Date;
+import java.util.List;
+
+public class CloudCheckInEventTest {
+
+    private static final String SYSTEM_UUID_KEY = "systemUuid";
+    private static final String CHECK_IN_KEY = "checkIn";
+    private static final String PROVIDER_KEY = "cloudProviderId";
+    private static final String ACCOUNT_KEY = "system-uuid";
```
nikosmoum, line 44: same here

src/main/java/org/candlepin/service/model/CloudCheckInEvent.java
```diff
@@ -0,0 +1,219 @@
...
+            .append(this.getCheckIn())
+            .append(this.getCloudProviderId())
+            .append(this.getCloudAccountId())
+            .append(this.getCloudOfferingIds());
+
+        return builder.toHashCode();
+    }
+
+    private String getSerializedBody() throws JsonProcessingException {
+        ObjectMapper mapper = new ObjectMapper();
```
nikosmoum, line 203: ObjectMapper instance creation is expensive, which is why we've limited the amount of instances we create by centralizing it to ObjectMapperFactory.java, and then marking each as a Singleton and Named in CandlepinModule.java. I think you can just re-use one of the existing instances by injecting it in this class (HypervisorUpdateJobObjectMapper looks the best, since it is just a `new Objectmapper()`  with the only configuration applied to it being on the deserialization side, which we don't care about in this case).

nikosmoum (changes requested):

src/main/java/org/candlepin/service/model/CloudCheckInEvent.java
joshmalbrecht: Updated

src/test/java/org/candlepin/service/model/CloudCheckInEventTest.java
joshmalbrecht: Updated

src/main/java/org/candlepin/service/model/CloudCheckInEvent.java
joshmalbrecht: That's a good point, although I don't think we can directly inject the mapper into the `CloudCheckInEvent` class since there can be multiple instances of the class and I think you would have to use the Guice Injector to create an instance of the class that way. 

I did add the ObjectMapper as a mandatory argument to the constructor, so the class that is creating the CloudCheckInEvents can inject the ObjectMapper into that class as it is probably a manager or resource class, and then pass the ObjectMapper to the CloudCheckInEvent.

src/main/java/org/candlepin/service/model/CloudCheckInEvent.java
```diff
@@ -0,0 +1,230 @@
...
+     *  if unable to serialize the body of the event
+     */
+    public CloudCheckInEvent(ConsumerCloudData cloudData, ObjectMapper mapper)
+        throws JsonProcessingException {
+        if (cloudData == null) {
+            throw new IllegalArgumentException("consumer cloud data is null");
+        }
+
+        if (mapper == null) {
+            throw new IllegalArgumentException("object mappter is null");
```
nikosmoum, line 74: typo: mappter

src/test/java/org/candlepin/service/model/CloudCheckInEventTest.java
```diff
@@ -0,0 +1,384 @@
...
+
+    @Test
+    public void testConstructorWithNullConsumerCloudData() {
+        assertThrows(IllegalArgumentException.class, () -> {
+            new CloudCheckInEvent(null, mapper);
+        });
+    }
+
+    @Test
+    public void testConstuctorWithNullConsumer() {
```
nikosmoum, line 57: typo here, and in other methods in this class: `Constuctor` instead of `Constructor`

nikosmoum (changes requested):

src/main/java/org/candlepin/service/model/CloudCheckInEvent.java
joshmalbrecht: updated

src/test/java/org/candlepin/service/model/CloudCheckInEventTest.java
joshmalbrecht: updated throughout

PR #4807, opened 2024-08-07
[M] CANDLEPIN-903: Increased el8 and el9 Vagrant VM filesystem size

joshmalbrecht:
-  Increased the el8 and el9 Vagrant VM filesystem size from 10 GB to 30 GB.

## Testing

Run the following command.

`vagrant up el9`

## Result

```
(14:51:33) vagrant@candlepin-el9:~/devel/candlepin (jalbrech/m/CANDLEPIN-903) $ lsblk
NAME   MAJ:MIN RM SIZE RO TYPE MOUNTPOINTS
vda    253:0    0  20G  0 disk 
└─vda1 253:1    0  20G  0 part /

```

Output of running `df -H` inside of the Vagrant VM

```
(14:51:44) vagrant@candlepin-el9:~/devel/candlepin (jalbrech/m/CANDLEPIN-903) $ df -H
Filesystem                                                 Size  Used Avail Use% Mounted on
devtmpfs                                                   4.2M     0  4.2M   0% /dev
tmpfs                                                      2.1G   17k  2.1G   1% /dev/shm
tmpfs                                                      822M   12M  810M   2% /run
/dev/vda1                                                   22G  6.4G   14G  32% /
tmpfs                                                      411M     0  411M   0% /run/user/1000
:/home/<Removed>/candlepin  255G  219G   26G  90% /vagrant
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -75,11 +75,16 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     provider.memory = 4096
     provider.graphics_type = "spice"
     provider.video_type = "qxl"
+    provider.machine_virtual_size = 30
   end
 
   config.vm.define("el8", primary: true) do |vm_config|
     vm_config.vm.box = "almalinux/8"
     vm_config.vm.host_name = "candlepin-el8.example.com"
+
+    # Increase box disk size and resize partitions accordingly
+    vm_config.vm.disk :disk, size: "30GB", primary: true
+    vm_config.vm.provision "shell", inline: "echo '- +' | sfdisk --no-reread -N 4 /dev/vda && partprobe && xfs_growfs /dev/vda4"
 
     # Vagrant allows to create a forwarded port mapping which allows access to a specific port
     # within the guest machine from a port on the host machine.
@@ -88,6 +93,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     # vm_config.vm.networking "forwarded_port", protocol: "tcp", guest: 8080, host: 8080
     # vm_config.vm.networking "forwarded_port", protocol: "tcp", guest: 8443, host: 8443
     vm_config.vm.provision "shell", inline: "dnf update -y dnf ca-certificates"
+
     configure_ansible_provisioning(vm_config)
   end
 
@@ -95,7 +101,14 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     vm_config.vm.box = "centos.cloud/centos9s"
     vm_config.vm.box_url = "https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-Vagrant-9-latest.x86_64.vagrant-libvirt.box"
     vm_config.vm.host_name = "candlepin-el9.example.com"
+
+    # Increase box disk size and resize partitions accordingly
+    vm_config.vm.disk :disk, size: "30GB", primary: true
+    vm_config.vm.provision "shell", inline: "echo '- +' | sfdisk --no-reread -N 1 /dev/vda && partprobe && resize2fs /dev/vda1"
+
+    # Update DNF CA certs
     vm_config.vm.provision "shell", inline: "dnf update -y dnf ca-certificates"
+
     configure_ansible_provisioning(vm_config)
   end
 


Vagrantfile
```diff
@@ -97,6 +98,12 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     vm_config.vm.host_name = "candlepin-el9.example.com"
     vm_config.vm.provision "shell", inline: "dnf update -y dnf ca-certificates"
     configure_ansible_provisioning(vm_config)
+
+    # Increase the partition size
+    vm_config.vm.provision "shell", inline: "growpart /dev/vda 1"
```
Ceiu, line 103: We can accomplish this without extra dependencies with the following:

```    
vm_config.vm.disk :disk, size: "30GB", primary: true
vm_config.vm.provision "shell", inline: "echo '- +' | sfdisk --no-reread -N 1 /dev/vda && partprobe && resize2fs /dev/vda1"
```

This should also be the first operation we perform in the box. Configure -> correct -> provision

Ceiu (changes requested):
This works, but I'm not fond of the order of operations, nor adding another dependency to do the work. I've provided an example using tools that should be native to both images we're currently using; but I'm open to other means of accomplishing it with pre-packaged tools.

Also, as an aside, we should start transitioning toward the Vagrant-native way of specifying the box size. It doesn't yet work on all versions (my F37 box still needs the libvirt declaration), but having both present doesn't break anything as far as I can tell. This is also present in the example blurb below; though note I used 30GB rather than 20GB. The difference there isn't massive, and 20GB is probably fine so whatever. Follow your heart there.

Vagrantfile
Ceiu: As an example of what my config looked like for this test:

```
  config.vm.define("el9", autostart: false) do |vm_config|
    vm_config.vm.box = "centos.cloud/centos9s"
    vm_config.vm.box_url = "https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-Vagrant-9-latest.x86_64.vagrant-libvirt.box"
    vm_config.vm.host_name = "candlepin-el9.example.com"

    # Increase base disk size and resize partitions accordingly
    vm_config.vm.disk :disk, size: "30GB", primary: true
    vm_config.vm.provision "shell", inline: "echo '- +' | sfdisk --no-reread -N 1 /dev/vda && partprobe && resize2fs /dev/vda1"

    # Update DNF ca certs
    vm_config.vm.provision "shell", inline: "dnf update -y dnf ca-certificates"

    configure_ansible_provisioning(vm_config)
  end
```

joshmalbrecht: I reverted my changes and used your config that you provided, but when I run `df -H` the partition is still 12 GB inside of the Vagrant el9 VM.

```
(13:02:00) vagrant@candlepin-el9:~/devel/candlepin (jalbrech/m/CANDLEPIN-903) $ df -H
Filesystem                                                 Size  Used Avail Use% Mounted on
devtmpfs                                                   4.2M     0  4.2M   0% /dev
tmpfs                                                      2.1G   17k  2.1G   1% /dev/shm
tmpfs                                                      822M   12M  810M   2% /run
/dev/vda1                                                   12G  6.4G  4.6G  58% /
```

joshmalbrecht: I think the issue is that the box "virtual_size" metadata definition is overriding the vagrant native disk size definition (`vm_config.vm.disk :disk, size: "30GB", primary: true`). If you download the `CentOS-Stream-Vagrant-9-latest.x86_64.vagrant-libvirt.box` box and look at the metadata.json file it has `"virtual_size": 11`.

If I keep all of your config the same, but add the libvirt config `provider.machine_virtual_size = 20`, then the VM partition and filesystem is 20 GB as expected.

Ceiu: Yeah, that's what I was mentioning above by needing both still. The disk size configuration is still new, so may not work universally/everywhere yet.

Also make sure the sizes line up so we don't get different results depending on which is used.

joshmalbrecht: My bad, I misread that portion of your comment! Updated and I used 30 GB for both configurations.

Ceiu (approved):

Ceiu:
We should probably also apply this to the el8 box as well

joshmalbrecht:
> We should probably also apply this to the el8 box as well (and then update the PR+commit title/description as well)

Good idea! I had to use `xfs_growfs` for el8 because the filesystem is XFS.

Ceiu (approved):
Was it necessary to specify the specific device with partprobe on Alma8?

joshmalbrecht:
> Was it necessary to specify the specific device with partprobe on Alma8?

That is my bad, I didn't mean to leave that in there. Updated!

PR #4818, opened 2024-08-21
[F] Added spec tests for new content API query parameters

joshmalbrecht:
- Added new spec tests for new content API query parameters.
- Updated existing spec tests.


spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -1157,4 +957,1008 @@ public void shouldPermitSafeParallelModification() throws Exception {
...
+            // Dummy org we use for creating a bunch of future pools for our global products. We'll also
+            // create per-org subscriptions for these products later. Also note that these will never be
+            // fully resolved.
+            OwnerDTO globalOwner = this.createOwner("global");
+
+            List<ProductDTO> globalProducts = new ArrayList<>();
+            for (int i = 1; i <= 3; ++i) {
+                ContentDTO content = Contents.random();
+
+                ProductContentDTO productContent = new ProductContentDTO();
```
Ceiu, line 1117: You can use the fluent-style mutators to chain these:
```
ProductContentDTO productContent = new ProductContentDTO()
  .content(content)
  .enabled(true);
```

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -1157,4 +957,1008 @@ public void shouldPermitSafeParallelModification() throws Exception {
...
+
+        private List<String> getUnexpectedIds(List<String> expectedPids) {
+            List<String> unexpectedPid = this.productMap.values()
+                .stream()
+                .map(ProductInfo::dto)
+                .map(ProductDTO::getId)
+                .filter(Predicate.not(expectedPids::contains))
+                .toList();
+
+            List<String> unexpectedCids = new ArrayList<>();
```
Ceiu, line 1185: Same here. You can either chain directly off the first stream (which should probably look closer to the previous example I provided to avoid the n*m lookup on contains), or you could call into the getContentIdsFromPids after resolving the unexpected products.

Any time you find yourself modifying or building a collection outside the scope of a functional block of calls, there's a chance something may not be quite right.

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -1157,4 +957,1009 @@ public void shouldPermitSafeParallelModification() throws Exception {
...
+        public void shouldErrorWithInvalidCustomInclusion() {
+            OwnerDTO owner = this.owners.get(1);
+
+            assertBadRequest(() -> this.adminClient.ownerContent()
+                .getContentsByOwner(owner.getKey(), null, null, INCLUSION_INCLUDE, "invalid_type"));
+        }
+
+        @Test
+        public void shouldAllowQueryingWithMultipleFilters() {
+            Random rand = new Random(14023);
```
Ceiu, line 1550: As a note: you'll want to check the data selected with each of these seeds to make sure it's somewhat representative of the shape we're looking for. The test should still work regardless, but it's nice when it gives us distributed data.

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -1157,4 +957,1009 @@ public void shouldPermitSafeParallelModification() throws Exception {
...
+            List<SubscriptionDTO> subscriptions = new ArrayList<>();
+
+            // Dummy org we use for creating a bunch of future pools for our global products. We'll also
+            // create per-org subscriptions for these products later. Also note that these will never be
+            // fully resolved.
+            OwnerDTO globalOwner = this.createOwner("global");
+
+            List<ProductDTO> globalProducts = new ArrayList<>();
+            for (int i = 1; i <= 3; ++i) {
+                ContentDTO content = Contents.random();
```
Ceiu, line 1115: We should create at least two content instances per product so we can verify/demonstrate that the query isn't stopping on the first match it finds.

This applies to all of the places where content is created. You can follow the paradigm established in the [Owner]ResourceTest unit tests where an 'a' and 'b' content is created for each product. Additionally, for debugging purposes, we should be nice and give them readable names for when things go wrong.

Ceiu (changes requested):

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
joshmalbrecht: Updated

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -1157,4 +957,1008 @@ public void shouldPermitSafeParallelModification() throws Exception {
...
+            return unexpectedCids;
+        }
+
+        private Predicate<ProductInfo> buildOwnerProductPredicate(OwnerDTO owner) {
+            return pinfo -> pinfo.owner() == null || owner.getKey().equals(pinfo.owner().getKey());
+        }
+
+        private List<String> getContentIdsFromPids(List<String> pids) {
+            List<String> cids = new ArrayList<>();
+            pids.forEach(pid -> {
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
joshmalbrecht: Updated

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
joshmalbrecht: Updated

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
joshmalbrecht: Updated

spec-tests/src/test/java/org/candlepin/spec/ContentResourceSpecTest.java
```diff
@@ -254,7 +257,20 @@ public void shouldListContentPagedAndSorted() {
                 }
```
Ceiu, line 257: We can actually remove this test entirely since we have better tests that cover this behavior.

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -1157,4 +958,1038 @@ public void shouldPermitSafeParallelModification() throws Exception {
...
+
+            OwnerDTO owner = this.owners.get(1);
+
+            List<String> pids = this.productMap.values()
+                .stream()
+                .filter(elem -> rand.nextBoolean())
+                .map(ProductInfo::id)
+                .toList();
+
+            List<String> expectedCids = this.productMap.values()
```
Ceiu, line 1277: This test is using the expected output as the input list, which lowers the value of the test. The first list (pids) is supposed to be a collection of input content IDs we're filtering on. The second is the expected output which is filtered down by taking that input list and removing anything that isn't from the org.

This should be updated such that we're pulling content IDs for the first list and then building our expected list from that.

Additionally, you should probably respin/check the seed to ensure that it's picking a reasonable chunk of data from the set. If it's biased toward removing too much or too little, the test trends toward edge cases.

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -1157,4 +958,1038 @@ public void shouldPermitSafeParallelModification() throws Exception {
...
+            OwnerDTO owner = this.owners.get(1);
+
+            ProductDTO expected = this.productMap.values()
+                .stream()
+                .filter(this.buildOwnerProductPredicate(owner))
+                .findAny()
+                .map(ProductInfo::dto)
+                .get();
+
+            List<String> expectedPids = List.of(expected.getId());
```
Ceiu, line 1310: Same here. We're still being very product centric for content-specific tests.

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
```diff
@@ -1157,4 +958,1038 @@ public void shouldPermitSafeParallelModification() throws Exception {
...
+                .filter(elem -> rand.nextBoolean())
+                .flatMap(ProductInfo::content)
+                .filter(elem -> rand.nextBoolean())
+                .map(ContentDTO::getLabel)
+                .toList();
+
+            // Add all the product IDs from the select labels
+            pids.addAll(productMap.values()
+                .stream()
+                .filter(buildContentLabelPredicate(labels))
```
Ceiu, line 1608: The presence of this predicate suggests something has gone off the rails here.

Ceiu (changes requested):
The bulk of these tests are still very product centric instead of filtering down to content. We're basically doubling or tripling the amount of work any of the per-test setup needs by doing it this way, while also narrowing the test results to entire product sets, rather than individual content.

Additionally, no tests have been added for the ContentResourceSpecTest endpoint.

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
joshmalbrecht: I see what you mean. I updated this seed to get more than one content ID. I was having trouble finding a value to get more than two.

[o2-content-1a, o2-content-1b]

spec-tests/src/test/java/org/candlepin/spec/ContentResourceSpecTest.java
joshmalbrecht: Removed

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
joshmalbrecht: Updated the test and I also updated the seed. I was able to find a seed to get more distributed data:

[g-content-2a, g-content-2b, g-content-3a, g-content-3b, g-content-1a, g-content-1b, o2-content-1b, o2-content-1a]

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
joshmalbrecht: updated

spec-tests/src/test/java/org/candlepin/spec/content/OwnerContentSpecTest.java
joshmalbrecht: updated

Ceiu:
In the interest of time, I'm cherry-picking this and I'll make the remaining changes in the feature branch before it goes up for review.

PR #4826, opened 2024-08-28
[M] CANDLEPIN-912: Updated product endpoints to support improved querying

Ceiu:
- Added the "name", "active", and "custom" filters to the query parameters of the GET /owners/{key}/products endpoint
- Added the "owner", "name", "active", and "custom" filters to the query parameters of the GET /products endpoint
- Removed the now-redundant "omit_globals" query parameter from both GET /owners/{key}/products and GET /products endpoints
- Reworked the GET /products and GET /owner/{key}/products endpoints to support filtering on product IDs, product names, and the "active" and "custom" states
- Added a new query builder object for building dynamic product queries

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -6076,23 +6076,93 @@ paths:
         - name: product
           in: query
           description: |
-            The ID of a product to fetch. If defined, the list of products returned by this method
-            will only include those matching the given ID. May be specified multiple times to filter
-            on multiple product IDs.
+            The ID of a product to fetch. If defined, the list of products returned by this endpoint will
+            only include those matching the given ID. May be specified multiple times to filter on multiple
+            product IDs. If multiple IDs are provided, any products matching any of the provided IDs will
+            be returned.
           required: false
           schema:
             type: array
             items:
               type: string
-        - name: omit_global
+        - name: name
           in: query
           description: |
-            whether or not to limit the lookup to only products defined within the organization's
-            namespace, excluding any globally defined products
+            The names of products to fetch. If defined, the list of products returned by this endpoint
+            will only include those matching the given names (case-insensitive). May be specified multiple
+            times to filter on multiple names. If multiple names are provided, any products matching any
+            of the provided names will be returned.
           required: false
           schema:
-            type: boolean
-            default: false
+            type: array
+            items:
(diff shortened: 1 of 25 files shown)


api/candlepin-api-spec.yaml
```diff
@@ -6076,23 +6076,93 @@ paths:
...
+            be passively included, excluded or omitted from the output, or exclusively selected
+            as the only products to return.
+          required: false
+          schema:
+            type: string
+            enum:
+              - include
+              - exclude
+              - exclusive
+            default: exclusive
```
joshmalbrecht, line 6149: This should be "include"

Ceiu: Fixed

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
```diff
@@ -781,17 +622,895 @@ private ProductDTO createProductWithProvidedAndDerivedProduct(String ownerKey) {
...
+        @Test
+        public void shouldErrorWithInvalidActiveInclusion() {
+            OwnerDTO owner = this.owners.get(1);
+
+            assertBadRequest(() -> this.adminClient.ownerProducts()
+                .getProductsByOwner(owner.getKey(), null, null, "invalid_type", INCLUSION_INCLUDE));
+        }
+
+        @Test
+        public void shouldAllowFilteringWithCustomIncluded() {
```
bakajstep, line 1079: This one

bakajstep (commented):
It look good to me just found 3 test which are the same so i just have question if it is ok or we want to update it?

joshmalbrecht (approved):

spec-tests/src/test/java/org/candlepin/spec/OwnerProductResourceSpecTest.java
Ceiu: These being effectively duplicates is fine. When it comes to writing tests and verifying behavior, you don't want to combine separate validations as you end up losing that in-code contract should test implementations change in the future.

To some extent, this is along the lines of self-documenting code; but we're leaning more on proper test isolation and verifying a specification.

bakajstep (approved):

PR #4834, opened 2024-09-04
[M] CANDLEPIN-918: Fixed bug preventing removal of consumer environments

Ceiu:
- Fixed a bug that prevented the removal of all environments from a consumer via the consumer update endpoint (PUT /consumers/{uuid})

### DIFF
--- a/src/main/java/org/candlepin/util/Util.java
+++ b/src/main/java/org/candlepin/util/Util.java
@@ -43,7 +43,6 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalAccessor;
-import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/util/Util.java
```diff
@@ -472,6 +472,61 @@ public static <T> boolean collectionsAreEqual(Collection<T> c1, Collection<T> c2
         return true;
     }
 
+    /**
+     * Compares to lists for equality without using the List.equals(...) method. This is primarily intended
```
joshmalbrecht, line 476: Typo: `Compares to lists` to `Compares two lists`

joshmalbrecht (approved):

src/main/java/org/candlepin/util/Util.java
Ceiu: Fixed

joshmalbrecht (approved):

PR #4836, opened 2024-09-04
[M] CANDLEPIN-914: Optimized the update content overrides endpoints

joshmalbrecht:
- Refactored and optimized the endpoints PUT /consumers/{uuid}/content_overrides, PUT /activation_keys/{id}/content_overrides, and PUT /environments/{id}/content_overrides such that they pull all overrides in bulk instead of each individually for update.

### DIFF
--- a/src/main/java/org/candlepin/model/ContentOverrideCurator.java
+++ b/src/main/java/org/candlepin/model/ContentOverrideCurator.java
@@ -16,7 +16,11 @@ package org.candlepin.model;
 
 import com.google.inject.persist.Transactional;
 
+import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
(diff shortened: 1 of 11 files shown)


bakajstep:
I don't know if it's intentional, but the commit message is missing the name of the jira card

joshmalbrecht:
> I don't know if it's intentional, but the commit message is missing the name of the jira card

Good catch! That was not intentional and I accidentally did not include it

Ceiu:
The way the commits are broken up aren't very valuable in terms repo history. The first hits both the backend and one endpoint, and the second is the other two endpoints and testing. In my mind, those should either be split up to be entirely backend+testing/frontend+testing, or just squash the whole thing down into one commit.

joshmalbrecht:
> The way the commits are broken up aren't very valuable in terms repo history. The first hits both the backend and one endpoint, and the second is the other two endpoints and testing. In my mind, those should either be split up to be entirely backend+testing/frontend+testing, or just squash the whole thing down into one commit.
> 
> One thing we should strive to maintain is the ability to jump to any commit in the history and have fully representative CI output and functionality. If at any point the commits entering main don't adhere to that, they should either be squashed down with everything necessary to achieve it, or refactored/rebased until they do.

That makes sense! I squashed to one commit.

nikosmoum (approved):

PR #4846, opened 2024-09-13
[M] CANDLEPIN-924: Added reconnection attempts to ingress session

joshmalbrecht:
- Added reconnect attempts to the server locator for the ingress session manager so that ClientSessionFactory will reconnect to another broker after a failover.

### DIFF
--- a/src/main/java/org/candlepin/async/impl/ActiveMQSessionFactory.java
+++ b/src/main/java/org/candlepin/async/impl/ActiveMQSessionFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ * Copyright (c) 2009 - 2024 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -115,7 +115,6 @@ public class ActiveMQSessionFactory {
         }
     }
 
-
     private Configuration config;
 
     private SessionManager ingressSessionManager;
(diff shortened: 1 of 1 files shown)


src/main/java/org/candlepin/async/impl/ActiveMQSessionFactory.java
```diff
@@ -163,10 +168,13 @@ protected synchronized SessionManager getIngressSessionManager() throws Exceptio
...
      * Fetches the egress session manager, creating a new instance as necessary.
      *
+     * @throws Exception
+     *  if unable to create a {@link ServerLocator} with the configured broker URL
+     *
      * @return
      *  a SessionManager for sessions configured for sending messages
      */
-    protected synchronized SessionManager getEgressSessionManager() throws Exception {
+    protected SessionManager getEgressSessionManager() throws Exception {
```
Ceiu, line 177: Wait, why did we remove the synchronized modifier from this?

joshmalbrecht: I thought the method wasn't used by multiple threads, but it is. Added the synchronized modifier back!

Ceiu: For future reference, even if your initial assessment were accurate, we still wouldn't want to remove the synchronized modifier from this method as its design is dependent on it for (future) multi-threaded behavior; Especially given that one method was to remain thread-safe. The way that method is implemented, if multiple threads call into it at roughly the same time, we may end up with multiple session managers.

joshmalbrecht: That makes sense!

PR #4851, opened 2024-09-17
[M] CANDLEPIN-925: Updated namespace column size

joshmalbrecht:
- Added change sets in 20231003113535-entity_layering_schema.xml so that the cp_products and cp_contents namespace column size can be updated to equal the size of cp_owner.account column size.
- Removed db/changelog/20240502145033-fix_entity_namespace_type.xml.

To validate that the changes resolved the problematic database, I did the following steps:

1. Created a Vagrant VM.
2. Imported the problematic database snapshot/datadump into Postgres which includes the data and schema.
3. Deployed Candlepin using `./bin/deployment/deploy -aH`.
4. Validated Candlepin was properly deployed and changesets were run.


Ceiu:
While this is the correct course of action, the various states of the changesets surrounding these columns needs to be considered:

- neither changeset run
- entity layering changeset run, namespace length changeset not run
- entity layering changeset run, namespace length changeset run

I imagine this will involve, at minimum, validChangeSet tags in the two changesets updated in this PR; but we'll want to make sure all three of those cases get through the migration properly.

joshmalbrecht:
> While this is the correct course of action, the various states of the changesets surrounding these columns needs to be considered:
> 
>     * neither changeset run
> 
>     * entity layering changeset run, namespace length changeset not run
> 
>     * entity layering changeset run, namespace length changeset run
> 
> 
> I imagine this will involve, at minimum, validChangeSet tags in the two changesets updated in this PR; but we'll want to make sure all three of those cases get through the migration properly.

Ah that makes sense. In that case, we could just add two new changesets for both the `cp_products` and `cp_contents` table that updates the column size before the migration changeset is run (20231003113535-11). That should cover all 3 cases.

Ceiu:
If that works without making Liquibase complain about a new changeset showing up in the file, then we can probably get away with removing the dedicated changeset file [1] for this fix entirely.

[1] db/changelog/20240502145033-fix_entity_namespace_type.xml

joshmalbrecht:
> If that works without making Liquibase complain about a new changeset showing up in the file, then we can probably get away with removing the dedicated changeset file [1] for this fix entirely.
> 
> [1] db/changelog/20240502145033-fix_entity_namespace_type.xml

Oh good catch! I didn't see that there was another changeset file with those changes already in it.

Ceiu:
Looks like our three potential cases are safe with this change, including the broken one.

PR #4852, opened 2024-09-17
[M] CANDLEPIN-923: Optimized content DTO translation

Ceiu:
- Updated the GET /content and GET /owners/.../content endpoints to translate internal model objects to DTOs in bulk to avoid performance issues associated with fetching lazily populated content fields

### DIFF
--- a/src/main/java/org/candlepin/model/ContentCurator.java
+++ b/src/main/java/org/candlepin/model/ContentCurator.java
@@ -717,4 +717,45 @@ public class ContentCurator extends AbstractHibernateCurator<Content> {
         return output;
     }
 
+    /**
+     * Fetches the required product IDs for the contents specified by the collection of content UUIDs.
+     * The required product IDs will be returned as a mapping of content UUID to required product ID
+     * collection. If a given content does not have any required product IDs, or the content UUID itself
+     * does not map to an existing content, it will not be represented in the output map. If none of the
+     * given content have any required product IDs, or none of the contents could be found, this method
+     * returns an empty map.
+     *
+     * @param contentUuids
+     *  a collection of UUIDs of the content for which to fetch required (modified) product IDs
+     *
+     * @return
+     *  the required product IDs for the given contents, represented as a map of content UUIDs to required
+     *  product IDs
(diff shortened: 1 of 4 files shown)


Ceiu:
Verifying this PR can (and should) be done in three ways:
- traditional static analysis
- measuring the runtime of the two cited endpoints with a reasonable amount of data
- enabling SQL query logging and verifying fetching content from the two endpoints does not trigger a query against `cp_content_required_products` for each entity returned

SQL query logging can be enabled by setting the following in `candlepin.conf` and restarting Tomcat:
```
log4j.logger.org.hibernate.SQL=DEBUG
```

This won't show parameters, but we aren't interested in those here. We only need to verify that we have a single (or couple) hits against the required products table, rather than the n+1 issue of one per entity.

Ceiu:
As much as it pains me to say this: Sonarcloud is actually right for once. But the more correct approach here is a design and multi-task epic in and of itself, which we simply do not have the time for in this PR and will be addressed in such a future task.

src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -717,4 +717,45 @@ WITH RECURSIVE pool_products (product_uuid) AS (
         return output;
     }
 
+    /**
+     * Fetches the required product IDs for the contents specified by the collection of content UUIDs.
+     * The required product IDs will be returned as a mapping of consumer UUID to required product ID
```
joshmalbrecht, line 722: Update `mapping of consumer UUID` to `mapping of content UUID`

joshmalbrecht (changes requested):
Looks good! I just noticed a small change in one of the Java docs

src/main/java/org/candlepin/model/ContentCurator.java
Ceiu: You wouldn't believe the number of times I've made this mistake working with content. Fixed.

Ceiu:
Also added missing tests for the content curator addition

joshmalbrecht (approved):

PR #4856, opened 2024-09-20
[M] CANDLEPIN-928: Updated updateLastCheckIn to refresh updated consumer

joshmalbrecht:
- Updated ConsumerManager.updateLastCheckIn to merge the updated
  entity so that the new check-in date is populated. An IAE exception
  would be thrown while creating a CloudCheckInEvent if the consumer
  was not refreshed with the new last check-in date and the consumer
  had a null last check-in date before the update.
- Updated CloudCheckInEvent constructor validation checks to throw
  IllegalStateException instead of IllegalArgumentException for invalid
  consumer state.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceSpecTest.java
@@ -1020,6 +1020,28 @@ public class ConsumerResourceSpecTest {
             .doesNotContain(consumer1.getId(), consumer2.getId());
     }
 
+    @Test
+    public void shouldProduceCheckInForCloudConsumer() {
+        ApiClient adminClient = ApiClients.admin();
+
+        OwnerDTO owner = adminClient.owners().createOwner(Owners.random());
+        ConsumerDTO consumer = adminClient.consumers().createConsumer(Consumers.randomAWS(owner));
+
+        OffsetDateTime before = consumer.getLastCheckin();
+
+        // Produce a check-in event
+        ApiClients.ssl(consumer)
+            .consumers()
+            .getEntitlementCertificateSerials(consumer.getUuid());
+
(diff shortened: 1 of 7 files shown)


spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceSpecTest.java
```diff
@@ -1020,6 +1020,28 @@ public void shouldOnlyFindConsumersForGivenEnvironmentIdAndOwner() {
...
+    public void shouldProduceCheckInForCloudConsumer() {
+        ApiClient adminClient = ApiClients.admin();
+
+        OwnerDTO owner = adminClient.owners().createOwner(Owners.random());
+        ConsumerDTO consumer = adminClient.consumers().createConsumer(Consumers.randomAWS(owner));
+
+        OffsetDateTime before = consumer.getLastCheckin();
+
+        // Produce a check-in event
+        ApiClients.trustedConsumer(consumer.getUuid())
```
nikosmoum, line 1033: We should avoid using trustedConsumer auth in spec tests (unless we're directly trying to test that authentication method or something relating to it). Lets use ApiClients.ssl() instead

joshmalbrecht: Good catch! Updated

src/main/java/org/candlepin/controller/ConsumerManager.java
```diff
@@ -58,6 +58,7 @@ public void updateLastCheckIn(Consumer consumer) {
         }
 
         consumerCurator.updateLastCheckin(consumer);
+        consumerCurator.refresh(consumer);
```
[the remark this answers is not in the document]
joshmalbrecht: That makes sense! Updated both the ConsumerManager and the CloudCheckInEvent

nikosmoum (approved):

Ceiu (approved):

PR #4860, opened 2024-09-25
[M] Updated logic for offering id in ConsumerCloudData

bakajstep:
- Refactored the ConsumerCloudData class to include a `setCloudOfferingIds` method, enabling direct replacement of cloud offering IDs instead of only appending them via `addCloudOfferingIds`. This allows for flexible management of cloud offering IDs, with validation ensuring that the maximum length is not exceeded.

### DIFF
--- a/src/main/java/org/candlepin/model/ConsumerCloudData.java
+++ b/src/main/java/org/candlepin/model/ConsumerCloudData.java
@@ -44,8 +44,6 @@ public class ConsumerCloudData extends AbstractHibernateObject<ConsumerCloudData
 
     /** Max length for a value in the ID field */
     public static final int ID_MAX_LENGTH = 32;
-    /** Max length for a value in the UUID field */
-    public static final int UUID_MAX_LENGTH = 36;
     /** Max length for a value in the cloud account ID field */
     public static final int CLOUD_ACCOUNT_ID_MAX_LENGTH = 255;
     /** Max length for a value in the cloud instance ID field */
@@ -177,7 +175,21 @@ public class ConsumerCloudData extends AbstractHibernateObject<ConsumerCloudData
     }
 
     /**
-     * @return the list of cloud offering IDs for this cloud consumer
+     * Retrieves the list of cloud offering IDs associated with this cloud consumer.
+     *
+     * <p>This method parses the internal comma-separated string of cloud offering IDs
+     * and returns them as a list of individual IDs. It handles redundant whitespace,
+     * repeated commas, and ensures that the returned list does not contain empty strings.</p>
+     *
+     * <p>If there are no cloud offering IDs set (i.e., the internal string is {@code null} or blank),
+     * this method returns an empty list.</p>
+     *
+     * <p><strong>Note:</strong> The returned list is a fixed-size list backed by an array.
+     * Modifications to the list (such as adding or removing elements) are not supported and
+     * will result in an {@code UnsupportedOperationException}. However, you can modify the elements
+     * within the list if needed.</p>
+     *
+     * @return a list of cloud offering IDs; never {@code null}, possibly empty
      */
     public List<String> getCloudOfferingIds() {
         if (this.cloudOfferingIds == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/model/ConsumerCloudData.java
```diff
@@ -222,6 +222,34 @@ public ConsumerCloudData setCloudOfferingIds(Collection<String> cloudOfferingIds
...
+    }
+
+    /**
+     * @param cloudOfferingIds
+     *  the cloud offering IDs to be set
+     *
+     * @return a reference to this ConsumerCloudData instance
+     */
+    public ConsumerCloudData setCloudOfferingIds(Collection<String> cloudOfferingIds) {
+        String joinedCloudOfferingIds = String.join(",", cloudOfferingIds);
```
Ceiu, line 243: Need null checks

src/main/java/org/candlepin/model/ConsumerCloudData.java
```diff
@@ -222,6 +222,34 @@ public ConsumerCloudData setCloudOfferingIds(Collection<String> cloudOfferingIds
...
     }
 
+    /**
+     * @param cloudOfferingIds
+     *  the cloud offering IDs to be set
+     *
+     * @return a reference to this ConsumerCloudData instance
+     */
+    public ConsumerCloudData setCloudOfferingIds(String... cloudOfferingIds) {
+        this.setCloudOfferingIds(Arrays.asList(cloudOfferingIds));
```
Ceiu, line 232: null check here

src/test/java/org/candlepin/model/ConsumerCloudDataTest.java
```diff
@@ -93,6 +93,26 @@ public void testSetCloudInstanceId() {
         assertEquals("cloudInstanceId exceeds the max length", exception.getMessage());
     }
 
+    @Test
+    public void testAddCloudOfferingIds() {
```
Ceiu, line 97: Let's split this up into distinct tests for the expected cases (single, multiple for both the array and collection input), and then our null, empty, and malformed cases.

Ceiu (changes requested):

Ceiu:
We should also be commenting these methods to be much clearer on how they function. What happens when a null collection is provided? How about a null element within that collection? The documentation needs to be very clear on what is expected to happen such that unshared expectations don't cause problems in the future.

Additionally, public methods in an abstraction or application boundary need to be air tight. They shouldn't be triggering exceptions that they aren't explicitly throwing themselves as a function of violating those aforementioned expectations.

src/main/java/org/candlepin/model/ConsumerCloudData.java
bakajstep: done

src/main/java/org/candlepin/model/ConsumerCloudData.java
bakajstep: done

bakajstep:
> We should also be commenting these methods to be much clearer on how they function. What happens when a null collection is provided? How about a null element within that collection? The documentation needs to be very clear on what is expected to happen such that unshared expectations don't cause problems in the future.
> 
> Additionally, public methods in an abstraction or application boundary need to be air tight. They shouldn't be triggering exceptions that they aren't explicitly throwing themselves as a function of violating those aforementioned expectations.

I hope they are better documented. I just don't know if I shouldn't change it everywhere and rewrite the tests to be separate?

src/main/java/org/candlepin/model/ConsumerCloudData.java
```diff
@@ -215,13 +265,80 @@ public ConsumerCloudData setCloudOfferingIds(Collection<String> cloudOfferingIds
...
+     */
+    public ConsumerCloudData setCloudOfferingIds(Collection<String> cloudOfferingIds) {
+        if (cloudOfferingIds == null) {
+            throw new IllegalArgumentException("cloudOfferingIds is null");
+        }
+        if (cloudOfferingIds.isEmpty()) {
+            this.cloudOfferingIds = null;
+            return this;
+        }
+        for (String cloudOfferingId : cloudOfferingIds) {
```
Ceiu, line 325: Let's add line breaks between each of these blocks so it's clear each is separate from the others.

Ceiu (changes requested):
One last whitespace change and this is good.

Ceiu (approved):

PR #4862, opened 2024-09-25
[M] Use guestIdLower column instead of LOWER SQL function

nikosmoum:
- Queries for GuestId entities used to use an optimization where a specific guestIdLower field/column was used instead of using the LOWER SQL function. That optimization got accidentally lost during the refactor and this change brings it back.


Ceiu:
Yeeesh... we're still doing this? We should be using functional indexes rather than doubling the number of columns storing this value. Especially since it looks like we're indexing both columns anyway -- this is literally the least efficient way of doing it.

Something to note for future designs, I suppose.

PR #4863, opened 2024-09-25
[M] Added candlepin.hidden_resources configuration to the dev image

joshmalbrecht:
- Set the candlepin.hidden_resources configuration as an empty string in the development container Candlepin configurations.

# Testing

1. Create the war file using `./gradlew war -Ptest_extensions=hostedtest,manifestgen`
2. Build the dev container image using the following command:

```
sudo docker build --target development -f ./containers/release.Containerfile -t testing:test . \
	--build-arg WAR_FILE=./build/libs/candlepin-4.4.17.war \
	--no-cache
```

3. Modify the `./dev-container/docker-compose.yml` file to use the newly created container image
4. Run `docker-compose up -d` to create the Candlepin container and the db container
5. Validate that the environment resource exists by running `curl -k https://localhost:8443/candlepin/`

### DIFF
--- a/containers/release.Containerfile
+++ b/containers/release.Containerfile
@@ -103,7 +103,8 @@ RUN echo "jpa.config.hibernate.dialect=org.hibernate.dialect.PostgreSQL92Dialect
     echo "candlepin.auth.oauth.enable=true" >> /etc/candlepin/candlepin.conf; \
     echo "candlepin.auth.oauth.consumer.rspec.secret=rspec-oauth-secret" >> /etc/candlepin/candlepin.conf; \
     echo "candlepin.db.database_manage_on_startup=Manage" >> /etc/candlepin/candlepin.conf; \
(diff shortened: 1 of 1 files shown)


src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -418,9 +418,7 @@ public static String jobConfig(String jobKey, String cfgName) {
             this.put(ACTIVATION_KEY_AUTHENTICATION, "true");
 
             this.put(AUTH_OVER_HTTP, "false");
-            // By default, environments should be hidden so clients do not need to
-            // submit one when registering.
```
Ceiu, line 422: This comment worries me. Have we vetted what modern subman does in this case? While environments are supported in hosted, they aren't *required*; so we need to make sure that flow isn't busted before we can change this. Probably.

Ceiu (approved):

PR #4869, opened 2024-10-04
[M] CANDLEPIN-933: Updated AHC.lockAndLoad to not return null elements in collections

Ceiu:
- Updated AbstractHibernateCurator.lockAndLoad to no longer include null elements in its output if one or more input IDs could not be found
- Updated the output type on AHC.lockAndLoad to be a List, rather than the more generic Collection
- Updated the documentation on the two list-outputting lockAndLoad implementations to be clear and consistent regarding the inclusion of nulls when a provided ID was not found
- Replaced use of Hamcrest.assertThat in the AbstractHibernateCurator test suite with AssertJ.assertThat

### DIFF
--- a/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
+++ b/src/main/java/org/candlepin/model/AbstractHibernateCurator.java
@@ -44,11 +44,11 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.SortedSet;
 import java.util.TreeMap;
-import java.util.TreeSet;
+import java.util.stream.StreamSupport;
 
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
@@ -915,8 +915,10 @@ public abstract class AbstractHibernateCurator<E extends Persisted> {
     }
 
     /**
-     * Loads the entities represented by the given IDs with a pessimistic write lock. If no
-     * entities were found with the given IDs, this method returns an empty collection.
+     * Loads the entities represented by the given IDs with a pessimistic write lock. If no entities
+     * could be found matching the given IDs, this method returns an empty collection. Note that this
+     * method makes no attempt to ensure that an entity is loaded for every ID provided. It is possible
+     * for the output collection to be smaller than the provided set of IDs.
      * <p></p>
      * <strong>Note:</strong> There is no guarantee that the entities returned by this method will
      * be loaded from the database at the time it is called. Due to various caching mechanisms, it's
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -915,8 +915,10 @@ protected E lockAndLoad(Class<E> entityClass, Serializable id) {
     }
 
     /**
-     * Loads the entities represented by the given IDs with a pessimistic write lock. If no
-     * entities were found with the given IDs, this method returns an empty collection.
+     * Loads the entities represented by the given IDs with a pessimistic write lock. If no entities
+     * could be found matching the given IDs, this method returns an empty collection. Note that this
+     * method makes no attempt to ensure that an entity is loaded for every ID provided. It is possible
```
nikosmoum, line 920: Can you explain what you mean by `makes no attempt to ensure that an entity is loaded for every ID provided` ? Do you mean it in the same way as the 'Note' below? If so, isn't that redundant?

src/main/java/org/candlepin/model/AbstractHibernateCurator.java
```diff
@@ -958,22 +960,28 @@ public Collection<E> lockAndLoad(Iterable<? extends Serializable> ids) {
...
+            .sorted()
+            .distinct()
+            .toList();
 
         // Fetch the entities from the DB...
-        if (idSet.size() > 0) {
+        if (ordered.size() > 0) {
             return this.currentSession()
                 .byMultipleIds(entityClass)
+                .enableOrderedReturn(false)
```
nikosmoum, line 982: are we sure this only affects the return order, and doesn't affect the actual lock order?

Ceiu: Yes. While this is technically undocumented behavior, the method name here implies that it's just a poor documentation thing and not unintended or undefined behavior. In either case, the provided list is iterated and entities within L1 and L2 cache are resolved. Any remaining entities are then loaded from the DB at the end. 

In both "standard" implementations of the MultiIdLoader [1][2], the difference in processing is that the ordered variant keeps an extra bit of data for the destination index into the output array for the entities that have to be loaded from the DB.

The concerning part here is it's not at all clear when the locks themselves are applied. In both cases it looks like unless there is some query caching and ordering done in the background, it's possible that the intended lock order is potentially violated if there are any cache hits. That is, it looks like all the cache hits will be ordered, but then the DB locks will be ordered as a separate blob [3][4]. Depending on how interlaced they are, that could pull things out of order in a way that invites deadlock. Just one more reason to not do hard DB locks and instead try to rely on things like session isolation so we aren't fighting stacks of dependency shenanigans like this.

[1] https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiIdEntityLoaderStandard.java
[2] https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiIdEntityLoaderArrayParam.java
[3] https://github.com/hibernate/hibernate-orm/blob/2e2d2948b0fd5042c9e8a60be8a2c8e5e984767a/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiIdEntityLoaderStandard.java#L156
[4] https://github.com/hibernate/hibernate-orm/blob/2e2d2948b0fd5042c9e8a60be8a2c8e5e984767a/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiIdEntityLoaderStandard.java#L381

PR #4878, opened 2024-10-17
[M] CANDLEPIN-940: Fixed bug in Pool.isManaged when managed flag is null

Ceiu:
- Fixed a bug in the Pool entity that can occur when checking the managed state if the manage flag had been changed to null in the backing database


nikosmoum (commented):
what would cause this value to be null? the constructor is always setting the value; is this the result of a changeset update?

Ceiu:
Not directly, no. Given the history of this field, it's likely that it was just never defined for a given pool and the database has *always* had a null; but it wasn't until the new wrapper function for it added in the rename patch that it became a problem.

The original column definition did not define it as non-nullable, nor did it provide a default, so that is likely our origin.

PR #4881, opened 2024-10-21
[M] CANDLEPIN-938: Redefined transactional boundaries for http requests

Ceiu:
- Added transactional tags for any request that performs a database operation at the start of the request handler
- Removed transactional tags from most internal single-query lookup at the curator or controller level
- Removed use of the transactional utility from SCA cert generation as this was only in place to coersce a flush to trigger Hibernate hooks and ID generation
- Removed some redundant operations for persisting entities


Ceiu:
This is going to need a lot of eyes on it to validate, and I have concerns that this is going to cause as many problems as it solves.

Specifically, with the translation of entities to DTOs occurring inside the transaction boundaries in virtually every case, there's a risk that anything reliant on Hibernate hooks to trigger may not before the translation occurs, so date fields may end up being incorrect, and ID fields may not be populated.

Initial testing shows we're actually pretty safe here by accident, but I don't know if that is due to it actually working properly or if our testing is incomplete in this area. At least one area was changed to fix the type of issue I'm worried about wrt SCA certs and the content access blob, so I suspect others are there, but they just didn't trip any existing test.

nikosmoum:
So the list of hooks we have are:

src/main/java/org/candlepin/model/AbstractHibernateObject.java:    @PrePersist
src/main/java/org/candlepin/model/AbstractHibernateObject.java:    @PreUpdate
src/main/java/org/candlepin/model/Consumer.java:    @PrePersist
src/main/java/org/candlepin/model/activationkeys/ActivationKeyPool.java:    @PrePersist
src/main/java/org/candlepin/model/activationkeys/ActivationKeyPool.java:    @PreUpdate
src/main/java/org/candlepin/model/RevocableCertificate.java:    @PreRemove

The AbstractHibernateObject ones are for all entities' created/updated dates. While we probably have some spec tests that indirectly check for those, can we add maybe one or two spec tests for key entities like Consumer and Product (see test example below)? The consumer test could also test the ensureUUID PrePersist as well.

The ActivationKeyPool, we may not care a lot with SCA-only coming up, but we wouldn't want to break the few customers that will stay in Entitlement mode for a while now, so maybe one test is needed there.

The RevocableCertificate @PreRemove is setting the serial to revoked=true whenever we delete/revoke a certificate. Maybe adding 1 spec test for each such cert that checks the CrlResource to confirm revocation would be good enough?

Example consumer created/updated test (I've run that against your branch and it seems to pass, so we're probably good):

```
    @Test
    public void shouldAutoPopulateCreatedUpdatedDatesOnConsumerRegister() {
        OwnerClient ownerClient = adminClient.owners();
        OwnerDTO someOwner = ownerClient.createOwner(Owners.random());
        ApiClient someUser = ApiClients.basic(UserUtil.createUser(adminClient, someOwner));

        ConsumerDTO consumer = someUser.consumers().createConsumer(
            Consumers.random(someOwner));
        assertThat(consumer)
            .returns(true, x -> x.getCreated() != null)
            .returns(true, x -> x.getUpdated() != null);
        // reload to be sure it was persisted
        consumer = someUser.consumers().getConsumer(consumer.getUuid());
        assertThat(consumer)
            .returns(true, x -> x.getCreated() != null)
            .returns(true, x -> x.getUpdated() != null);
    }
```

nikosmoum:
Also, there are already tests that check the timestamps are getting updated after SCA cert regeneration, so we should be fine about that one

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2359,6 +2383,8 @@ public Response getContentAccessBody(@Verify(Consumer.class) String consumerUuid
                 .build();
         }
 
+        log.info("GENERATING NEW CERTIFICATE!");
```
joshmalbrecht, line 2386: Double checking to see if this log statement and the one on line 2404 were accidentally left in from testing

src/main/java/org/candlepin/model/ContentCurator.java
```diff
@@ -511,8 +509,7 @@ public Collection<Content> resolveContentsByNamespace(String namespace) {
 
     // Needs an override due to the use of UUID as db identifier.
     @Override
-    @Transactional
-    public void delete(Content entity) {
+        public void delete(Content entity) {
```
joshmalbrecht, line 512: indention

joshmalbrecht (changes requested):
Had a question and noticed a small formatting issue

src/main/java/org/candlepin/model/ContentCurator.java
Ceiu: Yup, this needs to be reverted for this PR. It may get removed in the caching PR, but I'll let that PR deal with this change.

src/main/java/org/candlepin/resource/ConsumerResource.java
Ceiu: It was, good catch.

Ceiu:
@nikosmoum True enough, the problem is that to date we've only really used the timestamps kind of in passing as a convenient way of checking if a thing has changed, not really explicitly testing that it is being set. While this did work in our favor and catch this particular issue in the SCA cert generation case, I fear there are more such cases lurking.

The test blueprint you've provided should work for catching these, but should also include the object's primary ID as well as that is also auto-generated through Hibernate annotation magic.

---

As an aside, I want to be clear on explaining the issue and the timeline where we run into issues.

We have a timeline that looks something like this for our general request flow:
```
?---|-----|-|
    1     2 3
```

?. Some database stuff occurs for the purposes of processing data and logic required by our Tomcat/RESTeasy filters. These generally occur in their own transactions whenever a request begins because we aren't using the request context to start a transaction

1. transaction begins via `@Transactional` annotation on our target request endpoint handler (e.g. `ConsumerResource.searchConsumers`). 

2. We've fetched the results of our operation and translated them to DTOs, and return them as output to be processed by Jackson via more filters and magic handlers.

3. The transaction ends by way of the `@Transactional` annotation on our method falling out of scope. Ending the transaction also triggers an implicit flush and invokes all the Hibernate hooks listed above on any affected entity. This is where at least the created and updated fields are populated.

In terms of what the database sees, this is fine. No data violation has occurred, and everything ran in the correct order. However, because we translate at step 2 *before* the persist/update hooks have had a chance to run, we risk translating the object prematurely and not including those timestamps in the output.

This can be sidestepped by manually flushing before the translation step on any write operation (and in many cases, that's what's already happening -- we've been very liberal in our usage of `.flush()`) -- but flushes aren't free and we still have an implicit one at the end of the transaction, whether we like it or not. I would need to measure the performance impact of a no-op flush to see if it's worth just doing as a kind of SOP, or if that would be a major hit to overall throughput.

As a side note to a side note, this is effectively what the current implementation of the SCA cert generation stuff does, it just completely misuses transactions to accomplish it. It avoided starting a transaction so it could use the transactional util to create the cert and persist it (triggering the flush and hooks) before passing it along to the weird unwrap-rewrap logic we do for the pseudo-DTO. That was a *very* bad way of going about it when a simple flush will get the job done, and is why that was changed in this PR.

nikosmoum:
Yes I understand the flow. What I'm unsure of is if nested @Transactional annotations will also do a flush (even though they won't end the transaction). Is that why @Transactional was left on most AHC save/create/merge/update operations?

Ceiu:
They shouldn't. Hibernate/Guice (possibly JTA?) doesn't support nested transactions, so I believe it's just a no-op whenever a transactional block is re-entered, or an exception is thrown in some configurations.

src/main/java/org/candlepin/resource/ConsumerResource.java
Ceiu: Fixed

src/main/java/org/candlepin/model/ContentCurator.java
Ceiu: Fixed

Ceiu:
I've added some of the tests, but there are a few still lacking. Just getting this out early so we can get another pass on it.

Ceiu:
We should, but involving the various filters and whatnot is outside the scope of this PR.

Ceiu:
Okay, I think this is ready for final review. Hopefully.

nikosmoum (approved):

PR #4882, opened 2024-10-21
[M] CANDLEPIN-938: Fixed column name format in dirty_product precondition

Ceiu:
- Fixed the format of the column name in the precondition for the liquibase task for adding the dirty_product column from camel case to snake case

### DIFF
--- a/src/main/resources/db/changelog/20231211101743-add-pool-dirty-product-column.xml
+++ b/src/main/resources/db/changelog/20231211101743-add-pool-dirty-product-column.xml
@@ -15,7 +15,7 @@
     <changeSet id="20231211101743-1" author="crog">
         <preConditions onFail="MARK_RAN">
             <not>
-                <columnExists tableName="cp_pool" columnName="dirtyProduct"/>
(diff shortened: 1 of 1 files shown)


Ceiu:
This shouldn't have any impact on any users who have run this task already, as the docs and Liquibase's own output state that the preconditions do not impact the actual checksum of the task.

This can be verified with the following command from the CP root directory:
`./bin/deployment/liquibase.sh calculate-checksum --driver=org.postgresql.Driver --classpath=/var/lib/tomcat/webapps/candlepin/WEB-INF/lib/postgresql-42.7.4.jar --url="jdbc:postgresql:candlepin" --username=candlepin --changelog-file=src/main/resources/db/changelog/20231211101743-add-pool-dirty-product-column.xml --changeset-identifier 'src/main/resources/db/changelog/20231211101743-add-pool-dirty-product-column.xml::20231211101743-1::crog'`

The output should be `9:a5a29afc66914e124a687b9d07aa92d3` in both main and this branch.

bakajstep (approved):

joshmalbrecht (approved):

PR #4886, opened 2024-10-22
[M] CANDLEPIN-938: Restored L2 caching to product and content entities

Ceiu:
- Restored the cache annotations to the product and content entities,
  and their respective collections
- Removed use of the deprecated `@fetch` annotation on the product
  entity
- Enabled the setting to automatically evict entity collections when
  the entity itself is evicted from cache in persistence.xml
- Updated the Branding entity to be properly immutable, following
  the changes to ProductContent in an earlier commit
- Removed some extraneous or duplicate operations from
  AbstractHibernateCurator
- Fixed a number of potential avenues for null pointer exceptions
  in AbstractHibernateCurator and derived curators


nikosmoum (approved):

Ceiu:
Rebased against dependent changes

PR #4890, opened 2024-10-23
[M] CANDLEPIN-936: Create bulk set consumer environments endpoint

joshmalbrecht:
- Created PUT /owners/{owner_key}/consumers/environments endpoint for setting the environments of multiple consumers in one operation.

### DIFF
--- a/src/main/java/org/candlepin/controller/ConsumerManager.java
+++ b/src/main/java/org/candlepin/controller/ConsumerManager.java
@@ -18,14 +18,28 @@ package org.candlepin.controller;
 import org.candlepin.model.Consumer;
 import org.candlepin.model.ConsumerCloudData;
 import org.candlepin.model.ConsumerCurator;
+import org.candlepin.model.ContentAccessCertificateCurator;
+import org.candlepin.model.Environment;
+import org.candlepin.model.EnvironmentCurator;
+import org.candlepin.model.Owner;
 import org.candlepin.service.EventAdapter;
 import org.candlepin.service.model.CloudCheckInEvent;
+import org.candlepin.util.NonNullLinkedHashSet;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.persist.Transactional;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
 import java.util.Date;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 import javax.inject.Inject;
 
@@ -34,19 +48,25 @@ import javax.inject.Inject;
  * to the Consumer object.
  */
 public class ConsumerManager {
+    private static final Logger log = LoggerFactory.getLogger(ConsumerManager.class);
 
     private final ConsumerCurator consumerCurator;
+    private final ContentAccessCertificateCurator caCertificateCurator;
+    private final EnvironmentCurator envCurator;
     private final EventAdapter eventAdapter;
     private final ObjectMapper objectMapper;
 
     @Inject
     public ConsumerManager(ConsumerCurator consumerCurator,
+        ContentAccessCertificateCurator caCertificateCurator,
+        EnvironmentCurator envCurator,
         EventAdapter eventAdapter,
         ObjectMapper objectMapper) {
         this.consumerCurator = Objects.requireNonNull(consumerCurator);
+        this.caCertificateCurator = Objects.requireNonNull(caCertificateCurator);
+        this.envCurator = Objects.requireNonNull(envCurator);
         this.eventAdapter = Objects.requireNonNull(eventAdapter);
         this.objectMapper = Objects.requireNonNull(objectMapper);
-
     }
 
     /**
@@ -70,4 +90,92 @@ public class ConsumerManager {
             eventAdapter.publish(cloudCheckInEvent);
         }
     }
+
+    /**
+     * Sets the {@link Environment}s of the {@link Consumer}s that correspond to the provided consumer UUIDs.
+     * Consumers will be unchanged if they already exist in the provided environments and are in the right
+     * priority. Other consumers will be removed from their existing environments and set to the provided
+     * environments. The ordering of the provided environment IDs dictates the priority. The first environment
+     * ID in the list being the top priority and the last environment ID in the list being the least priority.
+     * All updated consumers will have their content access certificate revoked.
+     *
+     * @param owner
+     *  the owner of the consumers
+     *
+     * @param consumerUuids
+     *  the UUIDs of the consumers that should have the environments set
+     *
+     * @param targetEnvIds
+     *  the IDs of the {@link Environment}s that should be set for the consumers
+     *
+     * @throws IllegalArgumentException
+     *  if the owner is null
+     *
+     * @return
+     *  the UUIDs of the {@link Consumer}s that were updated and had their environments set
+     */
+    public Set<String> setConsumersEnvironments(Owner owner, Collection<String> consumerUuids,
+        NonNullLinkedHashSet<String> targetEnvIds) {
+
+        if (owner == null) {
+            throw new IllegalArgumentException("owner is null");
+        }
+
+        if (consumerUuids == null || consumerUuids.isEmpty() || targetEnvIds == null) {
+            return new HashSet<>();
+        }
+
+        log.info("setting {} consumers to {} environments",
+            consumerUuids.size(), targetEnvIds.size());
+
+        // If there are no target environments, then all the consumers should be removed from all of their
+        // current environments.
+        if (targetEnvIds.isEmpty()) {
+            int updated = envCurator.setConsumersEnvironments(consumerUuids, targetEnvIds);
+            log.info("{} consumers removed from all environments", updated);
+
+            return new HashSet<>(consumerUuids);
+        }
+
+        Set<String> consumersToUpdate = new HashSet<>();
+        Map<String, List<String>> consumerUuidToEnvs = envCurator.findEnvironmentsOf(consumerUuids);
+        for (String consumerUuid : consumerUuids) {
+            List<String> currentEnvs = consumerUuidToEnvs.get(consumerUuid);
+            if (currentEnvs == null || !hasSameElementsInOrder(targetEnvIds, currentEnvs)) {
+                consumersToUpdate.add(consumerUuid);
+            }
+        }
+
+        log.info("{} of the provided consumers are not in the target environments currently",
+            consumersToUpdate.size());
+
+        if (consumersToUpdate.isEmpty()) {
+            return new HashSet<>();
+        }
+
+        int updated = envCurator.setConsumersEnvironments(consumerUuids, targetEnvIds);
+        log.info("{} consumers have been set to the target environments", updated);
+
+        int deletedCerts = caCertificateCurator.deleteForConsumers(consumerUuids);
+        log.info("{} content access certs unlinked and removed", deletedCerts);
+
+        return consumersToUpdate;
+    }
+
+    private boolean hasSameElementsInOrder(NonNullLinkedHashSet<String> list1, Collection<String> list2) {
+        if (list1.size() != list2.size()) {
+            return false;
+        }
+
+        Iterator<String> list1Iterator = list1.iterator();
+        Iterator<String> list2Iterator = list2.iterator();
+        while (list1Iterator.hasNext()) {
+            if (!list1Iterator.next().equals(list2Iterator.next())) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
 }
--- a/src/main/java/org/candlepin/controller/ContentAccessManager.java
+++ b/src/main/java/org/candlepin/controller/ContentAccessManager.java
@@ -27,6 +27,7 @@ import org.candlepin.model.OwnerCurator;
 import org.candlepin.model.SCACertificate;
 import org.candlepin.util.Util;
 
+import com.google.inject.Provider;
 import com.google.inject.persist.Transactional;
 
 import org.apache.commons.lang3.ArrayUtils;
@@ -51,7 +52,7 @@ public class ContentAccessManager {
     private final OwnerCurator ownerCurator;
     private final ContentAccessCertificateCurator contentAccessCertificateCurator;
     private final ConsumerCurator consumerCurator;
-    private final EventSink eventSink;
+    private final Provider<EventSink> eventSink;
     private final JobManager jobManager;
     private final I18n i18n;
 
@@ -60,7 +61,7 @@ public class ContentAccessManager {
         ContentAccessCertificateCurator contentAccessCertificateCurator,
         OwnerCurator ownerCurator,
         ConsumerCurator consumerCurator,
-        EventSink eventSink,
+        Provider<EventSink> eventSink,
         JobManager jobManager,
         I18n i18n) {
 
@@ -272,7 +273,7 @@ public class ContentAccessManager {
 
             // Update sync times & report
             this.syncOwnerLastContentUpdate(owner);
-            this.eventSink.emitOwnerContentAccessModeChanged(owner);
+            this.eventSink.get().emitOwnerContentAccessModeChanged(owner);
 
             log.info("Content access mode changed from {} to {} for owner {}", currentMode,
                 updatedMode, owner.getKey());
@@ -373,5 +374,4 @@ public class ContentAccessManager {
         owner.syncLastContentUpdate();
         return this.ownerCurator.merge(owner);
     }
-
 }
--- a/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -1208,6 +1208,44 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
         return existingUuids;
     }
 
+    /**
+     * Determines if any of the provided {@link Consumer} UUIDs do not exist or do not belong to the provided
+     * {@link Owner}.
+     *
+     * @param ownerKey
+     *  the key of the owner that should own the consumers
+     *
+     * @param consumerUuids
+     *  the consumer UUIDs to check
+     *
+     * @return all of the provided consumer UUIDs that do not exist or do not belong to the provided owner
+     */
+    public Set<String> getNonExistentConsumerUuids(String ownerKey, Collection<String> consumerUuids) {
+        if (ownerKey == null || ownerKey.isBlank() || consumerUuids == null || consumerUuids.isEmpty()) {
+            return new HashSet<>();
+        }
+
+        String jpql = "SELECT c.uuid " +
+            "FROM Consumer c " +
+            "JOIN Owner o on o.id = c.ownerId " +
+            "WHERE c.uuid IN (:uuids) and o.key = :ownerKey";
+
+        Query query = getEntityManager()
+            .createQuery(jpql, String.class)
+            .setParameter("ownerKey", ownerKey);
+
+        Set<String> uniqueConsumerUuids = new HashSet<>(consumerUuids);
+        for (List<String> block : partition(uniqueConsumerUuids)) {
+            List<String> result = query
+                .setParameter("uuids", block)
+                .getResultList();
+
+            uniqueConsumerUuids.removeAll(result);
+        }
+
+        return uniqueConsumerUuids;
+    }
+
     public Consumer verifyAndLookupConsumer(String consumerUuid) {
         Consumer consumer = this.findByUuid(consumerUuid);
 
--- a/src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
+++ b/src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
@@ -218,4 +218,40 @@ public class ContentAccessCertificateCurator extends AbstractHibernateCurator<SC
         return serials;
     }
 
+    /**
+     * Deletes the content access certificates for the provided consumers and also revokes the certificate
+     * serials.
+     *
+     * @param consumerUuids
+     *  the UUIDs of {@link Consumer}s to delete content access certificates for
+     *
+     * @return the number of content access certificates that were deleted
+     */
+    public int deleteForConsumers(Collection<String> consumerUuids) {
+        if (consumerUuids == null || consumerUuids.isEmpty()) {
+            return 0;
+        }
+
+        String hql = """
+            SELECT cac.id, s.id
+            FROM Consumer c
+                JOIN c.contentAccessCert cac
+                JOIN cac.serial s
+            WHERE c.uuid IN (:uuids)
+            """;
+
+        Query query = this.getEntityManager()
+            .createQuery(hql);
+
+        List<Object[]> rows = new ArrayList<>();
+        for (List<String> block : this.partition(consumerUuids)) {
+            List<Object[]> result = query
+                .setParameter("uuids", block)
+                .getResultList();
+
+            rows.addAll(result);
+        }
+
+        return deleteCerts(rows);
+    }
 }
--- a/src/main/java/org/candlepin/model/EnvironmentCurator.java
+++ b/src/main/java/org/candlepin/model/EnvironmentCurator.java
@@ -14,22 +14,26 @@
  */
 package org.candlepin.model;
 
+import org.candlepin.util.NonNullLinkedHashSet;
+
 import org.hibernate.query.NativeQuery;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import javax.inject.Singleton;
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
 import javax.persistence.Query;
 import javax.persistence.TypedQuery;
-
 
 @Singleton
 public class EnvironmentCurator extends AbstractHibernateCurator<Environment> {
@@ -216,7 +220,7 @@ public class EnvironmentCurator extends AbstractHibernateCurator<Environment> {
             .getResultList();
     }
 
-    public Map<String, List<String>> findEnvironmentsOf(List<String> consumerIds) {
+    public Map<String, List<String>> findEnvironmentsOf(Collection<String> consumerIds) {
         if (consumerIds.isEmpty()) {
             throw new IllegalArgumentException("Consumers must not be empty!");
         }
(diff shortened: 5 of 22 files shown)


src/main/java/org/candlepin/controller/ConsumerManager.java
```diff
@@ -69,4 +92,99 @@ public void updateLastCheckIn(Consumer consumer) {
...
+            .getNonExistentConsumerUuids(consumerUuids, owner.getKey());
+        if (!unknownConsumerUuids.isEmpty()) {
+            throw new BadRequestException(i18n.tr("Unkown consumer UUIDs: {0}",
+                unknownConsumerUuids));
+        }
+
+        Set<String> unknownEnvironmentIds = envCurator
+            .getNonExistentEnvironmentIds(targetEnvIds, owner);
+        if (!unknownEnvironmentIds.isEmpty()) {
+            throw new BadRequestException(i18n.tr("Unknown environment IDs: {0}",
```
Ceiu, line 150: The service/controller level should not be throwing http- or user-facing exceptions. These are things that should be validated ahead of time and then handed over in a known good state.

src/main/java/org/candlepin/controller/ConsumerManager.java
```diff
@@ -69,4 +92,99 @@ public void updateLastCheckIn(Consumer consumer) {
...
+        if (consumersToUpdate.isEmpty()) {
+            return new HashSet<>();
+        }
+
+        int added = envCurator.setConsumersEnvironments(consumersToUpdate, targetEnvIds);
+        log.info("{} consumers have been set to the target environments", added);
+
+        // Delete all of the content access certificates for consumers that had an environment change
+        List<String> ids = caCertCurator.getIdsForConsumers(consumersToUpdate);
+        int unlinked = consumerCurator.unlinkCaCertificates(ids);
```
Ceiu, line 182: Is it necessary to delete the whole cert, or just the accessible content json? Also, this should be performed by the ContentAccessManager (or perhaps the SCACertGenerator/Manager?), so we don't have to ask that question nor deal with the specifics of triggering it here.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1212,6 +1212,46 @@ WHERE c.uuid IN (:uuids)
...
+            "FROM Consumer c " +
+            "JOIN Owner o on o.id = c.ownerId " +
+            "WHERE c.uuid IN (:uuids) and o.key = :ownerKey";
+
+        Query query = getEntityManager()
+            .createQuery(jpql, String.class)
+            .setParameter("ownerKey", ownerKey);
+
+        Set<String> uniqueConsumerUuids = new HashSet<>(consumerUuids);
+        Set<String> existingUuids = new HashSet<>();
```
Ceiu, line 1242: This is a lot of extra ID/data juggling for something we can get from the database straight away. We should avoid doubly converting our set. Or, at the very least, if we're going to stick with this paradigm, rather than add our results to a new set, we should instead remove matching results from it and then return it straight away.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -1212,6 +1212,46 @@ WHERE c.uuid IN (:uuids)
...
+     *
+     * @param consumerUuids
+     *  the consumer UUIDs to check
+     *
+     * @param ownerKey
+     *  the key of the owner that should own the consumers
+     *
+     * @return all of the provided consumer UUIDs that do not exist or do not belong to the provided owner
+     */
+    public Set<String> getNonExistentConsumerUuids(Collection<String> consumerUuids, String ownerKey) {
```
Ceiu, line 1227: nit: owner key should be first in the parameter order, as it has higher impact on the output.

Ceiu (changes requested):

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -308,4 +311,121 @@ public int removeEnvironmentContentReferences(Owner owner, Collection<String> co
...
+     *  the UUIDs for all the consumers that should be removed from their environments
+     *
+     * @return the number of consumers removed from all environments
+     */
+    private int removeConsumersFromAllEnvironments(Collection<String> consumerUuids) {
+        if (consumerUuids == null || consumerUuids.isEmpty()) {
+            return 0;
+        }
+
+        String statement = "DELETE FROM cp_consumer_environments " +
```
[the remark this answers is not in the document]
joshmalbrecht: Yeah, ideally we can use JQPL or JPA criteria to remove the environment relationship to the consumers and not have to deal with any of the caching issues. However, I'm struggling to find a way to do so

src/main/java/org/candlepin/controller/ConsumerManager.java
joshmalbrecht: I agree with moving this operation to the ContentAccessManager, however an interesting problem occurs when moving the operation to the ContentAccessManager and injecting the ContentAccessManager into the ConsumerManager. 

The problem is that during startup, `ConsumerCheckInFilter` is being created which injects `ConsumerManager` and this class now injects `ContentAccessManager` and this class injects `EventSinkImpl` which has a `CandlepinRequestScoped` annotation. This throws the error:

```
1) [Guice/ErrorInCustomProvider]: OutOfScopeException: Cannot access Key[type=EventSinkImpl, annotation=[none]] outside of a scoping block
  at EventSinkImpl.class(EventSinkImpl.java:55)
  while locating EventSinkImpl
  at ContentAccessManager.<init>(ContentAccessManager.java:67)
      \_ for 4th parameter
  at ConsumerManager.<init>(ConsumerManager.java:60)
      \_ for 2nd parameter
  at ConsumerCheckInFilter.<init>(ConsumerCheckInFilter.java:50)
      \_ for 2nd parameter
  while locating ConsumerCheckInFilter
```

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -308,4 +311,121 @@ public int removeEnvironmentContentReferences(Owner owner, Collection<String> co
...
+        String jpql = "SELECT DISTINCT env.id FROM Environment env " +
+            "WHERE env.id IN (:envIds) AND env.ownerId = :ownerId";
+
+        Set<String> distinctEnvIds = new HashSet<>(envIds);
+
+        Query query = this.getEntityManager()
+            .createQuery(jpql, String.class);
+
+        Set<String> actualEnvIds = new HashSet<>();
+        for (List<String> block : partition(distinctEnvIds)) {
```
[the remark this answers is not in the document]
joshmalbrecht: ahh!! Good catch. Updated

src/main/java/org/candlepin/model/EnvironmentCurator.java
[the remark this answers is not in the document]
joshmalbrecht: I couldn't find a way to handle this with JPA criteria or JPQL unfortunately. I removed the native queries to avoid the cache evictions and opted to load the Consumers in memory and update the `environmentIds` collection on the `Consumer`s objects.

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -308,4 +311,121 @@ public int removeEnvironmentContentReferences(Owner owner, Collection<String> co
...
+        removeConsumersFromAllEnvironments(consumerUuids);
+
+        String stmt = "INSERT INTO cp_consumer_environments (cp_consumer_id, environment_id, priority) " +
+            "SELECT id, :envId, :priority FROM cp_consumer WHERE uuid = :uuid";
+
+        for (String consumerUuid : consumerUuids) {
+            int priority = 0;
+            for (String envId : envIds) {
+                this.getEntityManager()
+                    .createNativeQuery(stmt)
```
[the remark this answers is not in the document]
joshmalbrecht: Removed this native query

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -308,4 +311,121 @@ public int removeEnvironmentContentReferences(Owner owner, Collection<String> co
...
+                    .setParameter("uuid", consumerUuid)
+                    .setParameter("envId", envId)
+                    .setParameter("priority", priority)
+                    .executeUpdate();
+
+                priority++;
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -308,4 +311,121 @@ public int removeEnvironmentContentReferences(Owner owner, Collection<String> co
...
+     *  if the provided environment IDs contains a duplicate or null value
+     *
+     * @return the number of consumers updated
+     */
+    public int setConsumersEnvironments(Collection<String> consumerUuids, List<String> envIds) {
+        if (consumerUuids == null || consumerUuids.isEmpty() || envIds == null || envIds.isEmpty()) {
+            return 0;
+        }
+
+        if (Util.containsDuplicateOrNull(envIds)) {
```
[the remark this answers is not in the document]
joshmalbrecht: I really like using our own custom type for this! I introduced `NonNullLinkedHashSet`. I also removed this method in favor of modifying the environmentIds on the Consumer object.

src/main/java/org/candlepin/controller/ConsumerManager.java
joshmalbrecht: Ah yeah! Moved this validation to the Resource layer

src/main/java/org/candlepin/model/ConsumerCurator.java
joshmalbrecht: Updated

src/main/java/org/candlepin/model/ConsumerCurator.java
joshmalbrecht: Updated

src/main/java/org/candlepin/util/NonNullLinkedHashSet.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Compares the elements and determines equality with the provided collection. All elements must be the
+     * same and in the same order for this method to return true.
+     *
+     * @param other
+     *  the collection to check equality for
+     *
+     * @return true if and only if the elements are all equal and in the same order
+     */
+    public boolean equalsCollection(Collection<E> other) {
```
Ceiu, line 78: With this being an object now, is this method needed? Would using the natural .equals method suffice?

joshmalbrecht: Yeah, we should just override the .equals method here. Updated

src/main/java/org/candlepin/util/NonNullLinkedHashSet.java
```diff
@@ -0,0 +1,108 @@
...
+        return true;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
```
joshmalbrecht, line 105: Apparently we must override both `equals` and `hashcode` based on our checkStyle rules

```
[ERROR] /vagrant/src/main/java/org/candlepin/util/NonNullLinkedHashSet.java:73:5: Definition of 'equals()' without corresponding definition of 'hashCode()'. [EqualsHashCode]
```

src/main/java/org/candlepin/util/NonNullLinkedHashSet.java
Ceiu: You probably don't even need to override it (or hashcode). The parent class should still handle null values in its equality check, and since we don't allow them, the target collection having nulls should fail it.

joshmalbrecht: The problem with using the `AbstractSet` equals method is that it will always return false for non Set instances. This is a problem because `EnvironmentCurator.findEnvironmentsOf` returns a List of environment IDs and we compare this List with a provided NonNullLinkedHashSet in `ConsumerManager.setConsumersEnvironments`. 

We could update `ConsumerManager.setConsumersEnvironments` to return a NonNullLinkedHashSet or Set instead of a List, but then we would have to update `EnvironmentResource.prepareEnvironmentUpdates` and probably `EnvironmentUpdates` as well.

AbstractSet.equals
```
   public boolean equals(Object o) {
      if (o == this) {
         return true;
      } else if (!(o instanceof Set)) {
         return false;
      } else {
         Collection<?> c = (Collection)o;
         if (c.size() != this.size()) {
            return false;
         } else {
            try {
               return this.containsAll(c);
            } catch (NullPointerException | ClassCastException var4) {
               return false;
            }
         }
      }
   }
```

joshmalbrecht: Removed the equals and hashcode override and added a private method in ConsumerManager that checks the elements of a `NonNullLinkedHashSet` with the elements in a collection.

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
```diff
@@ -218,4 +218,28 @@ public List<CertSerial> listCertSerials(Collection<String> consumerIds) {
...
+     *
+     * @return the content access certificate IDs for all of the provided consumers
+     */
+    public List<String> getIdsForConsumers(Collection<String> consumerUuids) {
+        if (consumerUuids == null || consumerUuids.isEmpty()) {
+            return new ArrayList<>();
+        }
+
+        String jpql = "SELECT c.contentAccessCert.id FROM Consumer c WHERE c.uuid IN (:uuids)";
+        Query query = entityManager.get().createQuery(jpql, String.class);
```
Ceiu, line 235: Let's use the `getEntityManager` method instead of hitting the provider directly.

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1903,4 +1908,64 @@ private Owner createOwnerFromDTO(OwnerDTO ownerDTO) {
...
+    // PUT /owners/{owner_key}/consumers/environments
+
+    @Override
+    @Transactional
+    public void setConsumersToEnvironments(@Verify(Owner.class) String ownerKey,
+        SetConsumerEnvironmentsDTO request) {
+
+        List<String> consumerUuids = request.getConsumerUuids();
+        if (consumerUuids == null || consumerUuids.isEmpty()) {
+            throw new BadRequestException(i18n.tr("Consumer UUIDs is null or empty"));
```
Ceiu, line 1921: Any user-facing messages should be less terse than our usual messaging we leave for ourselves. I would expect this to be something closer to "No consumer UUIDs provided".

src/main/java/org/candlepin/resource/OwnerResource.java
```diff
@@ -1903,4 +1908,64 @@ private Owner createOwnerFromDTO(OwnerDTO ownerDTO) {
...
+
+        Owner owner = findOwnerByKey(ownerKey);
+        if (!ContentAccessMode.ORG_ENVIRONMENT.toDatabaseValue().equals(owner.getContentAccessMode())) {
+            throw new BadRequestException(i18n.tr("Owner is not in SCA content access mode"));
+        }
+
+        Set<String> unknownConsumerUuids = consumerCurator
+            .getNonExistentConsumerUuids(owner.getKey(), consumerUuids);
+        if (!unknownConsumerUuids.isEmpty()) {
+            throw new BadRequestException(i18n.tr("Unkown consumer UUIDs: {0}",
```
Ceiu, line 1949: Typo in the error message here

src/main/java/org/candlepin/util/Util.java
```diff
@@ -702,4 +702,33 @@ public static boolean isFalse(Boolean value) {
...
+     *
+     * @return true if there is a duplicate in the provided collection or a null value, false otherwise
+     */
+    public static <T> boolean containsDuplicateOrNull(Collection<T> collection) {
+        if (collection == null || collection.isEmpty()) {
+            return false;
+        }
+
+        Set<T> objects = new HashSet<>();
+        for (T obj : collection) {
```
Ceiu, line 723: This may be a fun experiment in perf testing to check if it's faster to do it this way under the assumption that our input *does* contain duplicates or null, or if -- since we're using a set already -- it's faster to dump all of the items into the collection and then check if the set contains null or has fewer elements than our input:

```
Set<T> objects = new HashSet<>(collection);
return objects.contains(null) || objects.size() < collection.size();
```

My guess would be that the iterative method is faster in cases where the input indeed contains dupes or nulls early on, but slower in cases where the input is functionally a set of non-nulls or has many elements with a null or dupe near the tail of the collection's iterator.

Also, if the input collection is a set already, there's no reason to add everything to a new set (iteratively or otherwise). A simple `.contains(null)` check would be sufficient. So our potential solution becomes:

```
Set<T> objects = collection instanceof Set set ? set : new HashSet<>(collection);
return objects.contains(null) || objects.size() < collection.size();
```

---

Something else to consider with this method is whether or not this behavior needs to be generalized or should be handled as part of the conversion. We're doing a lot of rework iterating over the elements just to pass them to another *different* set later, which already imposes these restrictions. That being the case, we could have that conversion do the work of both checks here:

```
NonNullLinkedHashSet<String> envSet = new NonNullLinkedHashSet<>();

try {
  envSet.addAll(environmentIds);

  if (envSet.size() < environmentIds.size()) {
    throw new BadRequestException(this.i18n.tr("Request contains duplicate environment IDs"));
  }
}
catch (IllegalArgumentException e) {
  throw new BadRequestException(this.i18n.tr("Environment IDs cannot be null or empty"));
}
```

Ceiu (changes requested):

src/main/java/org/candlepin/controller/ConsumerManager.java
Ceiu: Anything injecting a scoped or stateful object needs to be requested on demand via providers within the appropriate context and should not be injected directly.

If we're running into circular dependencies or are otherwise misusing our DI layer, that's something we should address sooner than later. Possibly as a prereq to this PR.

src/main/java/org/candlepin/model/EnvironmentCurator.java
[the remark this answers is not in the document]
Ceiu: I need to go through and look at this again to see what our options are. One thing we were hitting with SCA certs is that the consumer lookup and whatever else Hibernate is doing was really costing us a lot in terms of performance. It got bad enough that dropping the second JPA call and going native only -- even with the cache evictions -- ended up being an order of magnitude faster.

Knowing what we know now wrt query spaces, query optimization around the IN operator, and Hibernate's L2 cache shenanigans, we should revisit this and perhaps profile our various approaches.

This also applies to the "return the objects that don't exist" queries in this PR. Something about them doesn't feel quite right to me, but I haven't really dedicated much time into thinking out a better approach.

src/main/java/org/candlepin/util/NonNullLinkedHashSet.java
Ceiu: That's really obnoxious, but I suppose somewhat makes sense. I'd have liked it if it were smart enough to do logical equality, but such is life I suppose.

src/main/java/org/candlepin/controller/ConsumerManager.java
joshmalbrecht: Moved this operation to the `ContentAccessManager` and used the `Provider` for retrieving the `EventSinkImpl` in `ContentAccessManager`.

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
joshmalbrecht: Updated

src/main/java/org/candlepin/resource/OwnerResource.java
joshmalbrecht: Updated

src/main/java/org/candlepin/resource/OwnerResource.java
joshmalbrecht: Oops haha updated

src/main/java/org/candlepin/util/Util.java
joshmalbrecht: That is a great point, I really like adding the environment IDs into the `NonNullLinkedHashSet` and determining if there were duplicates by comparing the size with the provided list! I updated `OwnerResource` to use this approach and removed the `Util.containsDuplicateOrNull` method

src/main/java/org/candlepin/model/EnvironmentCurator.java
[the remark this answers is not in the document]
joshmalbrecht: Yeah, we can probably improve the performance here by using native queries.

With the current approach, it takes 1.641 seconds to update the environments of 1000 consumers locally. However, this doesn't account for the production scale of the DB and I'm not sure we can test this without promoting the change to the stage environment and load testing the new endpoint. 

I'm not sure what our acceptable performance metrics are for the code review stage (before release and before hitting stage for loadtesting).

src/main/java/org/candlepin/controller/ContentAccessManager.java
```diff
@@ -374,4 +377,28 @@ public Owner syncOwnerLastContentUpdate(Owner owner) {
...
+    /**
+     * Unlinks all content access certificates from the provided {@link Consumer}s and deletes the
+     * content access certificates.
+     *
+     * @param consumerUuids
+     *  the UUIDs for the consumers that should have their content access certificates deleted
+     *
+     * @return the number of deleted content access certificates
+     */
+    public int deleteContentAccessCertificates(Collection<String> consumerUuids) {
```
joshmalbrecht, line 389: We also need to revoke the serials for these certificates

joshmalbrecht: Updated

src/main/java/org/candlepin/model/EnvironmentCurator.java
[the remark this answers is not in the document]
joshmalbrecht: I updated the setting of environments to use native queries and reduced the response time to around 1000 ms for 1000 consumers locally.

Ceiu (commented):
Either this has shrunk since last I looked or there's not much to it overall. Anyway, added some comments around the main query. Curious what everyone's thoughts are surrounding it.

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -307,4 +310,130 @@ public int removeEnvironmentContentReferences(Owner owner, Collection<String> co
...
+        if (envIds == null || envIds.isEmpty()) {
+            return removedFromEnvs;
+        }
+
+        String stmt = "INSERT INTO cp_consumer_environments (cp_consumer_id, environment_id, priority) " +
+            "SELECT id, :envId, :priority FROM cp_consumer WHERE uuid = :uuid";
+
+        for (String consumerUuid : consumerUuidsSet) {
+            int priority = 0;
+            for (String envId : envIds) {
```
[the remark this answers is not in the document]
joshmalbrecht: Nice, I like this a lot!! Getting rid of all the update executions in the loops is a huge improvement.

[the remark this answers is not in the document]
joshmalbrecht: Updated this in a new commit. As discussed in a Slack thread, this had great performance improvements.

Before changes with doubly nested loop:
Migration of 1000 consumers to 1 environment: 1088 ms
Migration of 1000 consumers to 9 environments: 2465 ms

After changes:
Migration of 1000 consumers to 1 environment: 608 ms
Migration of 1000 consumers to 9 environments: 885 ms

src/main/java/org/candlepin/model/EnvironmentCurator.java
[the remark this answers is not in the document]
joshmalbrecht: Including some information around the problematic deletion for cp_consumer_environments in prod. Based on the ANALYZE on the deletion statement, we are using indexes for both the subquery and the deletion statement which is good.

```
MariaDB [candlepin]> ANALYZE SELECT cp_consumer_id FROM cp_consumer_environments
    -> WHERE cp_consumer_id IN (SELECT id FROM cp_consumer where owner_id = '<Redacted Prod Owner ID>');
+------+-------------+--------------------------+------+----------------------------------------------------+-------------------+---------+--------------------------+------+--------+----------+------------+--------------------------+
| id   | select_type | table                    | type | possible_keys                                      | key               | key_len | ref                      | rows | r_rows | filtered | r_filtered | Extra                    |
+------+-------------+--------------------------+------+----------------------------------------------------+-------------------+---------+--------------------------+------+--------+----------+------------+--------------------------+
|    1 | PRIMARY     | cp_consumer              | ref  | PRIMARY,fk_consumer_owner,cp_consumer_owner_fk_idx | fk_consumer_owner | 98      | const                    | 8    | 8.00   |   100.00 |     100.00 | Using where; Using index |
|    1 | PRIMARY     | cp_consumer_environments | ref  | PRIMARY,cp_consumer_environments_idx               | PRIMARY           | 98      | candlepin.cp_consumer.id | 1    | 0.12   |   100.00 |     100.00 | Using index              |
+------+-------------+--------------------------+------+----------------------------------------------------+-------------------+---------+--------------------------+------+--------+----------+------------+--------------------------+
```

Ceiu (changes requested):

src/main/java/org/candlepin/model/EnvironmentCurator.java
```diff
@@ -307,4 +311,159 @@ public int removeEnvironmentContentReferences(Owner owner, Collection<String> co
...
+            "FROM consumer_ids cids " +
+            "CROSS JOIN env_priorities ep";
+
+        String cteSQL = null;
+        for (List<String> block : partition(consumerUuidsSet)) {
+            // Hibernate does not allow parameter expansion for positional parameters, so we must replace
+            // the positional parameter IN statement based on the partition size. Only create or re-create the
+            // query if we have not already created the query or if we are on the last partition and the
+            // number of remaining consumers is smaller than the block size.
+            if (cteSQL == null || block.size() != getInBlockSize()) {
```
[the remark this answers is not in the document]
joshmalbrecht: Ah, I see what you are saying. I added logic for calculating the consumer UUID partition block size.

For the limitation of the number of environments we can pass in, we are limiting the number with the `BULK_SET_CONSUMER_ENV_MAX_ENV_LIMIT` configuration which is currently set to 10. Also, the `BULK_SET_CONSUMER_ENV_MAX_CONSUMER_LIMIT` limits the number of consumers for this endpoint and the default max value is set to 1000.

I added a check that checks if the number of environment ID query parameters exceeds or equals the minimum of the IN block query param limit and the query param limit. This shouldn't happen in practice, but it is technically possible if someone sets the `BULK_SET_CONSUMER_ENV_MAX_ENV_LIMIT`  really high and passes in a large amount of environment IDs, so it is a good idea to check this case.

Ceiu (approved):

Ceiu:
Just need the squash and final merge/squash artifact check and this looks like it'll be good for now.

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
```diff
@@ -218,4 +218,35 @@ public List<CertSerial> listCertSerials(Collection<String> consumerIds) {
...
+            FROM Consumer c
+                JOIN c.contentAccessCert cac
+                JOIN cac.serial s
+            WHERE c.uuid IN (:uuids)
+            """;
+
+        Query query = this.getEntityManager()
+            .createQuery(hql);
+
+        List<Object[]> rows = query
```
Ceiu, line 246: Missing partitioning on consumerUuids

Ceiu (changes requested):
Either I missed the missing partition on the deletion in the previous passes, or it went missing in the squash. Hopefully this is the last necessary change.

src/main/java/org/candlepin/model/ContentAccessCertificateCurator.java
joshmalbrecht: Good catch! Added partitioning on the consumerUuids

Ceiu (approved):

PR #4893, opened 2024-10-25
[M] CANDLEPIN-938: Reverted auto-detach when removing child products and content

Ceiu:
- Updated the product and content removal functions to no longer automatically remove references from parent products or subscriptions. Endpoints and internal functions will now, instead, fail when attempting to remove/delete a child entity that is still referenced by one or more parent entities.
- Restored a missing check that ensures products and contents being removed or updated are existing, managed entities
- Updated some tests on update and removal to verify the received errors more explicitly to avoid allowing a test pass when receiving an error from a different fail state

### DIFF
--- a/src/main/resources/db/changelog/20241024154545-revert_product_content_delete_cascades.xml
+++ b/src/main/resources/db/changelog/20241024154545-revert_product_content_delete_cascades.xml
@@ -1,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <!--
+        Removes the delete cascades and delete nulls added as QoL to allow deleting children without having
+        to manually unlink them from parents first. While this worked fine for what it did, it directly
+        interferes with Hibernate's ability to keep L2 cache in sync, and Hibernate does not provide tools
+        to manually resync the cache without also violating transaction isolation.
+    -->
+
+    <changeSet id="20241024154545-1" author="crog">
+        <preConditions onFail="MARK_RAN">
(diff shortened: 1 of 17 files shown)


src/main/resources/db/changelog/20241024154545-revert_product_content_delete_cascades.xml
```diff
@@ -0,0 +1,62 @@
...
+        interferes with Hibernate's ability to keep L2 cache in sync, and Hibernate does not provide tools
+        to manually resync the cache without also violating transaction isolation.
+    -->
+
+    <changeSet id="20241024154545-1" author="crog">
+        <preConditions onFail="MARK_RAN">
+            <foreignKeyConstraintExists foreignKeyTableName="cp_products" foreignKeyName="cp_products_fk1" />
+        </preConditions>
+
+        <dropForeignKeyConstraint baseTableName="cp_products" constraintName="cp_products_fk1"/>
```
nikosmoum, line 21: Maybe I'm misunderstanding how this check is supposed to work, but do we really need this?
We know there's a `cp_products_fk1` FK, so if we're expecting the drop+add to work. If this FK doesn't exist, it means the schema is in a bad state, and we should error out instead, instead of marking it as RAN (which means the cp_products_fk1 will just continue to not exist (because the changeset it skipped), so we further let the schema stay in a bad state.

I think it would make more sense that the preCondition+dropFK be in one changeset, and the addFK in a separate changeset, following it, to make sure the FK always gets created no matter what.

(same comment/question for the other 2 changesets)

Ceiu: I'm not sure I follow. The first paragraph seems to suggest that we want to error on any desync (which I don't think is a good idea given the known state of things), but then the second paragraph describes a method for being more lenient with detected state and (silently) self-healing.

They're at odds with each other and it's not clear to me for which you're advocating.

nikosmoum: First paragraph is more me thinking out loud; but I'm advocating for what I mentioned in the 2nd paragraph (self-healing)

Ceiu: Alright. In that case, fixed.

PR #4900, opened 2024-10-30
[M] CANDLEPIN-944: Droped FK for cp2_owner_content/product on cp_owner

bakajstep:
- When the satellite tried to update from version 6.15 to 6.16, we decided not to use cp2_ownerproduct/content anymore and to utilize new tables for that purpose. However, we chose to keep the original tables in case we needed to revert to the previous version. We didn’t delete the foreign keys (FK) on those tables, which prevented the deletion of custom owners due to constraint violations.

### DIFF
--- a/src/main/resources/db/changelog/20241029000000-drop-fk-cp-owner.xml
+++ b/src/main/resources/db/changelog/20241029000000-drop-fk-cp-owner.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <changeSet id="20241029000000-1" author="sbakaj">
+        <preConditions onFail="MARK_RAN">
+            <foreignKeyConstraintExists foreignKeyTableName="cp2_owner_products"
+                foreignKeyName="cp2_owner_products_fk1" />
+        </preConditions>
+        <dropForeignKeyConstraint baseTableName="cp2_owner_products" constraintName="cp2_owner_products_fk1" />
+    </changeSet>
+
+    <changeSet id="20241029000000-2" author="sbakaj">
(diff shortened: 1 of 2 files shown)


src/main/resources/db/changelog/202410290000000-drop-fk-cp-owner.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <changeSet id="202410290000000-1" author="sbakaj">
+        <dropForeignKeyConstraint baseTableName="cp2_owner_products" constraintName="cp2_owner_products_fk1"/>
```
Ceiu, line 10: Add checks to make sure these FKs still exist before we do the drop

src/main/resources/db/changelog/202410290000000-drop-fk-cp-owner.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <changeSet id="202410290000000-1" author="sbakaj">
```
Ceiu, line 9: There should only be 6 digits after the date part of the timestamp.

Ceiu (changes requested):

src/main/resources/db/changelog/202410290000000-drop-fk-cp-owner.xml
bakajstep: done

src/main/resources/db/changelog/202410290000000-drop-fk-cp-owner.xml
bakajstep: done

Ceiu (approved):

PR #4908, opened 2024-11-05
[M] Fixed name of the JIRA TOKEN in env variables

bakajstep:


nikosmoum (changes requested):
as mentioned offline this isn't needed; this action fails on PRs from forks (like weblate PRs) because secrets aren't accessible there

PR #4910, opened 2024-11-06
[M] 2018469: Disabled subscription attachment by Activation Key in SCA

bakajstep:
- Manual subscription attachment via activation keys in SCA mode was initially allowed to ease the transition from Entitlement, but it was no longer needed. Auto-attachment was disabled at the Candlepin level, and Satellite 6.9+ removed subscription attachment from the UI.

### DIFF
--- a/src/test/java/org/candlepin/resource/util/ConsumerBindUtilTest.java
+++ b/src/test/java/org/candlepin/resource/util/ConsumerBindUtilTest.java
@@ -375,33 +375,7 @@ public class ConsumerBindUtilTest {
 
         consumerBindUtil.handleActivationKeys(consumer, Arrays.asList(key1), false);
 
-        verify(this.entitler, times(1)).bindByPoolQuantity(eq(consumer), eq(pool1.getId()), eq(1));
-    }
-
-    @Test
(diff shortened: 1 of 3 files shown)


src/test/java/org/candlepin/resource/util/ConsumerBindUtilTest.java
```diff
@@ -395,13 +395,10 @@ public void handleActivationKeysWithoutAutoAttachUsingSCAWhenBindFails() throws
 
         ConsumerBindUtil consumerBindUtil = this.buildConsumerBindUtil();
 
-        doThrow(new ForbiddenException("exception")).when(this.entitler)
-            .bindByPoolQuantity(eq(consumer), eq(pool1.getId()), eq(5));
-
```
nikosmoum, line 400: since now we'll never get into the `bindByPoolQuantity` method in SCA mode, this exception would never get thrown, so this test is no longer needed, so we can remove it

nikosmoum (changes requested):
minor change and this is good to merge

src/test/java/org/candlepin/resource/util/ConsumerBindUtilTest.java
bakajstep: done

PR #4911, opened 2024-11-07
[M] 2017774: Disallowed Manual Subscription Attachment in SCA Mode

bakajstep:
- In SCA mode, manual attachment of subscriptions was no longer needed. Auto-attachment had already been disabled at the Candlepin and Satellite (6.9+) levels. Only manifest consumers ('candlepin' or 'satellite' types) were allowed to attach entitlements to ensure manifest generation continued to function properly.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
@@ -15,6 +15,7 @@
 package org.candlepin.spec;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.candlepin.spec.bootstrap.assertions.JobStatusAssert.assertThatJob;
 import static org.candlepin.spec.bootstrap.assertions.StatusCodeAssertions.assertForbidden;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
@@ -169,7 +170,61 @@ class ExportSpecTest {
     @Test
     public void shouldImportContentAccessCertsForAConsumerBelongingToOwnerInSCAMode() throws Exception {
         ApiClient adminClient = ApiClients.admin();
-        OwnerDTO owner = adminClient.owners().createOwner(Owners.randomSca());
+        OwnerDTO owner = adminClient.owners().createOwner(Owners.random());
+        String ownerKey = owner.getKey();
+
+        ProductDTO modifiedProd = adminClient.ownerProducts().createProduct(ownerKey, Products.random());
+        ProductDTO prod = adminClient.ownerProducts().createProduct(ownerKey, Products.random());
+        ContentDTO cont = adminClient.ownerContent().createContent(ownerKey, Contents.random()
+            .modifiedProductIds(Set.of(modifiedProd.getId())));
+        adminClient.ownerProducts().addContentToProduct(ownerKey, prod.getId(), cont.getId(), true);
+        PoolDTO pool = adminClient.owners().createPool(ownerKey, Pools.random(prod));
+        ConsumerDTO consumer = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ApiClient consumerClient = ApiClients.ssl(consumer);
+        consumerClient.consumers().bindPool(consumer.getUuid(), pool.getId(), 1);
+
+        // Switching to SCA to verify that only the SCA certificate is present and not the entitlement
+        // certificate. We need to wait here for a while to allow the EntitlementRevokingJob to complete.
+        owner.setContentAccessMode("org_environment");
+        adminClient.owners().updateOwner(ownerKey, owner);
+
+        AsyncJobStatusDTO entJob = adminClient.jobs()
+            .listMatchingJobStatusForOrg(ownerKey, null, null)
+            .stream()
+            .filter(job -> job.getKey().equals("EntitlementRevokingJob"))
+            .findFirst()
+            .get();
+
+        assertThatJob(entJob)
+            .isNotNull()
+            .terminates(adminClient)
+            .isFinished();
+
+
+        File manifest = createCertExport(consumerClient, consumer.getUuid());
+        ZipFile export = ExportUtil.getExportArchive(manifest);
+
+        // Check if content access certs are present in exported zip file.
+        List<ZipEntry> caCerts = export.stream()
+            .filter(entry -> entry.getName().startsWith(CONTENT_ACCESS_CERTS_PATH))
+            .filter(entry -> entry.getName().lastIndexOf('/') == CONTENT_ACCESS_CERTS_PATH.length() - 1)
+            .collect(Collectors.toList());
+
+        assertThat(caCerts).singleElement();
+
+        // Should not contain entitlement certificate in SCA mode
+        List<ZipEntry> entitlementCerts = export.stream()
+            .filter(entry -> entry.getName().startsWith(ENTITILEMENT_CERTIFICATES_PATH))
+            .filter(entry -> entry.getName().lastIndexOf('/') == ENTITILEMENT_CERTIFICATES_PATH.length() - 1)
+            .collect(Collectors.toList());
+
+        assertThat(entitlementCerts).isEmpty();
+    }
+
+    @Test
+    public void shouldImportEntitlementCertsForAConsumerBelongingToOwnerInEntitlementMode() throws Exception {
+        ApiClient adminClient = ApiClients.admin();
+        OwnerDTO owner = adminClient.owners().createOwner(Owners.random());
         String ownerKey = owner.getKey();
 
         ProductDTO modifiedProd = adminClient.ownerProducts().createProduct(ownerKey, Products.random());
@@ -185,18 +240,18 @@ class ExportSpecTest {
         File manifest = createCertExport(consumerClient, consumer.getUuid());
         ZipFile export = ExportUtil.getExportArchive(manifest);
 
-        // Check if content access certs are present in exported zip file.
+        // Should not contain content access certs in exported zip file in Entitlement mode.
         List<ZipEntry> caCerts = export.stream()
-            .filter(entry -> entry.getName().startsWith(CONTENT_ACCESS_CERTS_PATH))
-            .filter(entry -> entry.getName().lastIndexOf('/') == CONTENT_ACCESS_CERTS_PATH.length() - 1)
+            .filter(e -> e.getName().startsWith(CONTENT_ACCESS_CERTS_PATH))
+            .filter(e -> e.getName().lastIndexOf('/') == CONTENT_ACCESS_CERTS_PATH.length() - 1)
             .collect(Collectors.toList());
 
-        assertThat(caCerts).singleElement();
+        assertThat(caCerts).isEmpty();
 
-        // Check if entitlement certs are present in exported zip file.
+        // Should contain entitlement certificate in Entitlement mode
         List<ZipEntry> entitlementCerts = export.stream()
-            .filter(entry -> entry.getName().startsWith(ENTITILEMENT_CERTIFICATES_PATH))
-            .filter(entry -> entry.getName().lastIndexOf('/') == ENTITILEMENT_CERTIFICATES_PATH.length() - 1)
+            .filter(e -> e.getName().startsWith(ENTITILEMENT_CERTIFICATES_PATH))
+            .filter(e -> e.getName().lastIndexOf('/') == ENTITILEMENT_CERTIFICATES_PATH.length() - 1)
             .collect(Collectors.toList());
 
         assertThat(entitlementCerts).singleElement();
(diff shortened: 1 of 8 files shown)


spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -193,13 +193,13 @@ public void shouldImportContentAccessCertsForAConsumerBelongingToOwnerInSCAMode(
...
 
-        // Check if entitlement certs are present in exported zip file.
+        // Should not contain entitlement certificate in SCA mode
         List<ZipEntry> entitlementCerts = export.stream()
             .filter(entry -> entry.getName().startsWith(ENTITILEMENT_CERTIFICATES_PATH))
             .filter(entry -> entry.getName().lastIndexOf('/') == ENTITILEMENT_CERTIFICATES_PATH.length() - 1)
             .collect(Collectors.toList());
 
-        assertThat(entitlementCerts).singleElement();
+        assertThat(entitlementCerts).isEmpty();
```
Ceiu, line 202: Double check with the Sat/Katello folks on this one. I'm pretty sure the manifests still need one or more of the certs for resyncing later. I forget which one off hand, so it'd be good to vet this with them to ensure we aren't about to break manifest import and resync.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2550,6 +2548,15 @@ public Response bind(
         ConsumerType ctype = this.consumerTypeCurator.getConsumerType(consumer);
         log.debug("Consumer (post verify): {}", consumer);
 
+        // Not allow to attach ent to consumer in SCA mode(Only manifest consumers are allowed)
+        boolean scaEnabled = consumer.getOwner().isUsingSimpleContentAccess();
+        if (scaEnabled && !ctype.isManifest()) {
+            ExceptionMessage message = new ExceptionMessage(
+                i18n.tr("Ignoring request to attach entitlements. It is disabled for org: {0} because of " +
```
Ceiu, line 2555: Let's clean up this error message, and try to avoid splitting the string across multiple lines, as that makes searching for it harder.

Something to the effect of: "Bind operations are disabled in organizations using simple content access". Maybe "ignored" is better in this context since it's not a failure?

May be valuable to reach out to PMs for recommendations on how this should be expressed to end users.

Ceiu (changes requested):

src/main/java/org/candlepin/resource/ConsumerResource.java
nikosmoum: To add to this: by now, most components that used to call this API (Satellite, RHSM-Web, RHSM-API, subscription-manager in RHEL8+) have been patched to simply not call it when in SCA mode. The only thing we need to test is old subscription-manager clients (RHEL7/RHEL6). What is expected is, the operation should:
- not result in an unhandled stacktrace error
- should result in a descriptive message to the user
I don't know how subman in RHEL7/6 works, so I'm not sure if sending back a 200 status with an exception message will work. Perhaps it could be a 4XX error code that would work

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
nikosmoum: It is confusing, but what this test is calling isn't the 'manifest export' endpoint (`ConsumerResource.exportDataAsync`). It is the 'export certificates of a non-manifest consumer in a zip format' endpoint (`ConsumerResource.exportCertificates`), which isn't related to Satellite. Instead this is called from RHSM-Web/customer portal, where a 'disconnected' Hosted customer can go to the portal, create a normal (non-manifest consumer), add entitlements to it, and export the entitlement certs as a zip file. The reason the test is now getting an empty list of entitlements is because this consumer is non-manifest.

However, this endpoint should not longer be called if an org has been switched to SCA mode, because the user will be redirected to consoleDot, and never reach the customer portal page that would call this.
So I think what we need here is two versions of this test:
- The current one, where the org is in SCA mode, and no entitlement certs are exported
- Another where the org is in Entitlement mode, and entitlement certs are exported

bakajstep: done, test added

src/main/java/org/candlepin/resource/ConsumerResource.java
bakajstep: i did not throw error because this is written in bugzilla: `(We do not want to throw an error in order to make sure older, unpatched clients do not break)`

nikosmoum: I think I wrote that BZ, and I agree its not very clear. The goal is for unpatched clients (RHEL7 and 6) to not have unhandled errors that break them. As long as the client is able to handle this by, for example, showing user-friendly error, then we are fine

nikosmoum: have you tested with a RHEL7 or 6 client? how does it look from that perspective?

bakajstep: Updated the error message, opting to use "Attach" instead of "Bind" because the customer is using subman attach. The response now returns a 400 status code instead of 200. Verified on RHEL 6 to ensure this change does not break the subman client. The client will display the updated error message we provided and still conclude with a zero exit code.

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -182,6 +183,12 @@ public void shouldImportContentAccessCertsForAConsumerBelongingToOwnerInSCAMode(
         ApiClient consumerClient = ApiClients.ssl(consumer);
         consumerClient.consumers().bindPool(consumer.getUuid(), pool.getId(), 1);
 
+        // Switching to SCA to verify that only the SCA certificate is present and not the entitlement
+        // certificate.We need to wait here for a while to allow the EntitlementRevokingJob to complete.
+        owner.setContentAccessMode("org_environment");
+        adminClient.owners().updateOwner(ownerKey, owner);
+        sleep(1000);
```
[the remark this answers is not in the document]
Ceiu: There are custom assertions we wrote specifically for this. You can use the following to wait for a job and verify its final state:

```
assertThatJob(job)
    .isNotNull()
    .terminates(this.adminClient)
    .isFinished()
    <whatever other checks you'd like here>
```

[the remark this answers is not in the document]
bakajstep: done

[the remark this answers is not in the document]
bakajstep: done

nikosmoum (approved):
Looks good to me. I'll let @Ceiu re-review/merge

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -182,6 +183,24 @@ public void shouldImportContentAccessCertsForAConsumerBelongingToOwnerInSCAMode(
         ApiClient consumerClient = ApiClients.ssl(consumer);
         consumerClient.consumers().bindPool(consumer.getUuid(), pool.getId(), 1);
 
+        // Switching to SCA to verify that only the SCA certificate is present and not the entitlement
+        // certificate. We need to wait here for a while to allow the EntitlementRevokingJob to complete.
+        owner.setContentAccessMode("org_environment");
+        adminClient.owners().updateOwner(ownerKey, owner);
+
+        AsyncJobStatusDTO entJob = adminClient.jobs()
```
Ceiu, line 191: double indents on all of this

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
```diff
@@ -193,12 +212,48 @@ public void shouldImportContentAccessCertsForAConsumerBelongingToOwnerInSCAMode(
...
+        // Should not contain content access certs in exported zip file in Entitlement mode.
+        List<ZipEntry> caCerts = export.stream()
+                .filter(e -> e.getName().startsWith(CONTENT_ACCESS_CERTS_PATH))
+                .filter(e -> e.getName().lastIndexOf('/') == CONTENT_ACCESS_CERTS_PATH.length() - 1)
+                .collect(Collectors.toList());
+
+        assertThat(caCerts).isEmpty();
+
+        // Should contain entitlement certificate in Entitlement mode
+        List<ZipEntry> entitlementCerts = export.stream()
```
Ceiu, line 252: Double indents here as well. Likely throughout all of these changes

src/test/java/org/candlepin/resource/ConsumerResourceTest.java
```diff
@@ -451,6 +451,27 @@ protected Owner createOwner(String key) {
         return owner;
     }
 
+    protected Owner createOwnerEnt() {
+        return createOwnerEnt(null);
+    }
+
+    protected Owner createOwnerEnt(String key) {
```
Ceiu, line 458: Better naming and/or documentation would go a long way here to avoiding confusion. It's not clear until you read the method's logic if we're creating an entitlement for an owner, an owner with an entitlement, or an owner in entitlement mode.

e.g. `createEntitlementModeOwner`, `createNonSCAOwner`, `createLegacyCAOwner`

src/test/java/org/candlepin/test/DatabaseTestFixture.java
```diff
@@ -650,6 +651,20 @@ protected Owner createOwner(String key, String name) {
         return owner;
     }
 
+    protected Owner createOwnerEnt(String key) {
```
Ceiu, line 654: nack

If this is going to be a generalized test util method, it needs to be named better or have some docs associated with it. "owner ent" is ambiguous in terms of what it will actually do; especially within the context of some of the other testutil methods which indeed do things like create an owner with an entitlement.

src/test/java/org/candlepin/test/DatabaseTestFixture.java
```diff
@@ -650,6 +651,20 @@ protected Owner createOwner(String key, String name) {
...
+        return this.createOwnerEnt(key, key);
+    }
+
+    protected Owner createOwnerEnt(String key, String name) {
+        Owner owner = TestUtil.createOwner(key, name);
+        owner.setId(null);
```
Ceiu, line 660: you can use method chaining here to save some repetition:

```
Owner owner = TestUtil.createOwner(key, name)
  .setId(null)
  .setContentAccessMode(...);
```

src/test/java/org/candlepin/test/DatabaseTestFixture.java
```diff
@@ -650,6 +651,20 @@ protected Owner createOwner(String key, String name) {
...
+        owner.setContentAccessMode(ContentAccessMode.ENTITLEMENT.toDatabaseValue());
+
+        this.ownerCurator.create(owner);
```
Ceiu, line 663: Return the value from create or merge in utility methods like this.

Ceiu (changes requested):

src/test/java/org/candlepin/resource/ConsumerResourceTest.java
bakajstep: done

src/test/java/org/candlepin/test/DatabaseTestFixture.java
bakajstep: done

src/test/java/org/candlepin/test/DatabaseTestFixture.java
bakajstep: done

src/test/java/org/candlepin/test/DatabaseTestFixture.java
bakajstep: done

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
bakajstep: done

spec-tests/src/test/java/org/candlepin/spec/ExportSpecTest.java
bakajstep: done

src/test/java/org/candlepin/resource/ConsumerResourceEntitlementRulesTest.java
```diff
@@ -52,7 +51,7 @@ public class ConsumerResourceEntitlementRulesTest extends DatabaseTestFixture {
     public void setUp() {
         consumerResource = injector.getInstance(ConsumerResource.class);
         standardSystemType = consumerTypeCurator.create(new ConsumerType("system"));
-        this.owner = this.createOwner("test-owner");
+        this.owner = this.createNonSCAOwner("test-owner");
```
Ceiu, line 54: :+1:

Ceiu (approved):

PR #4913, opened 2024-11-07
[M] Update supported hotfix branch list in i18n job

nikosmoum:


bakajstep (approved):

bakajstep:
and it might not be a bad idea to write this somewhere so when we do a new hotfix or depricate another hotfix to edit the translation

PR #4920, opened 2024-11-11
[M] CANDLEPIN-938: Registered ObjectMapper as an eager singleton

Ceiu:
- Updated the module configuration for ObjectMapper to ensure it is created as an eager singleton, rather than creating it on every injection

### DIFF
--- a/src/main/java/org/candlepin/guice/CandlepinModule.java
+++ b/src/main/java/org/candlepin/guice/CandlepinModule.java
@@ -277,7 +277,7 @@ public class CandlepinModule extends AbstractModule {
         bind(JsRunner.class).toProvider(JsRunnerProvider.class);
         bind(SyncUtils.class).asEagerSingleton();
         bind(ObjectMapperFactory.class).asEagerSingleton();
-        bind(ObjectMapper.class).toProvider(ObjectMapperFactory.class);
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/guice/CandlepinModule.java
```diff
@@ -526,7 +526,7 @@ private ObjectMapper configureExportObjectMapper() {
 
     @Provides
     @Singleton
-    @Named("ExportObjectMapper")
+    @Named("RulesObjectMapper")
```
nikosmoum, line 529: I am unsure (its been a while since I looked this up) but I don't think this is enough to get this working. I think you also need to set `@Named("RulesObjectMapper")` on the RulesObjectMapper constructor argument for it to know to use this provider method. The actual name in the Named annotation doesn't matter (could be `abc`) as long as both the provider method and the method argument we're injecting both have the annotation

Ceiu: Nah, it works fine due to the typing being specific (RulesObjectMapper vs vanilla ObjectMapper). I kept this in place with the corrected name for the sake of consistency, and because I didn't feel like risking anything breaking if I didn't find the named tag on an inject method somewhere.

PR #4934, opened 2024-11-14
POC: Added singleton for resource layer

bakajstep:
In this PR, we've changed the scope of the ActivationKeyResource class to `@Singleton`. This modification introduced issues related to dependency injection of beans with shorter lifecycles into the singleton class. Specifically, directly injecting non-singleton scoped beans into a singleton can lead to scope mismatches and threading problems.

### Why Provider<T> Is Necessary

- **Scope Mismatch Problem:** A singleton bean is created once and shared across the entire application. If it directly injects a bean that is scoped differently (e.g., request-scoped), it will hold onto a reference that is not meant to be shared globally. This can cause unexpected behavior, memory leaks, and concurrency issues because the singleton may access stale or invalid instances of the dependency.

- **Thread Safety Concerns**: Singleton beans are accessed by multiple threads simultaneously. If they contain references to mutable, non-thread-safe dependencies, this can lead to race conditions and data inconsistencies.

### How Provider<T> Resolves the Issue

- **Lazy Retrieval of Dependencies:** The **Provider<T>** interface allows us to obtain instances of dependencies on-demand by calling the `get()` method. Instead of injecting the dependency directly, we inject a provider that supplies the dependency when needed.

- **Scope Compliance:** By using `Provider<T>`, each call to **get()** retrieves an instance of the dependency that is appropriate for the current scope. This ensures that even within a singleton class, we can work with dependencies that have shorter lifecycles without violating their scope constraints.

- **Avoiding Shared Mutable State:** Since the singleton does not hold onto the actual instance of the dependency, but rather obtains it when needed, we prevent the singleton from inadvertently sharing mutable state across threads.

### DIFF
--- a/src/main/java/org/candlepin/resource/ActivationKeyResource.java
+++ b/src/main/java/org/candlepin/resource/ActivationKeyResource.java
@@ -57,41 +57,58 @@ import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import javax.inject.Inject;
+import javax.inject.Provider;
+import javax.inject.Singleton;
 
 
-
+@Singleton
 public class ActivationKeyResource implements ActivationKeyApi {
     private static final Logger log = LoggerFactory.getLogger(ActivationKeyResource.class);
     private static final Pattern AK_CHAR_FILTER = Pattern.compile("^[a-zA-Z0-9_-]+$");
 
+    // Singleton dependencies
     private final ActivationKeyCurator activationKeyCurator;
     private final ActivationKeyContentOverrideCurator contentOverrideCurator;
     private final ProductCurator productCurator;
-    private final PoolService poolService;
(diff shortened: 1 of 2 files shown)


bakajstep:
This is just example how we can achive resource layer as singletons. Also want to look at JProfiler to compare it with main how this is performing. Will leave results later.

Ceiu:
If this is the fundamental problem for all of the resources, then I'm alright with this approach. However, I want it to not be a sledgehammer we use to fix all of our DI problems. A number of objects here should not need to be injected on demand like this and should be singletons themselves.

So in the actual approach, I'd expect this to cascade to a number of other objects and a more intentional approach taken to determine which of our objects are stateful and which are not, and configure/inject/provide them as appropriate.

PR #4939, opened 2024-11-19
[M] CANDLEPIN-954: Relaxed constraint on null values in ConsumerCloudData

Ceiu:
- ConsumerCloudData.addCloudOfferingIds and .setCloudOfferingIds now
  accept null collections and collections with null or empty values,
  but will silently ignore them for the purposes of collecting and
  retaining the offering IDs
- Removed use of Apache Commons StringUtil from the Util class
- Removed ConsumerCloudData.setId, .equals, and .hashCode
- Fixed a bug in the Request object which would pass through null
  values to the underlying regex replace operation

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
@@ -45,6 +45,8 @@ import org.candlepin.spec.bootstrap.client.ApiClients;
 import org.candlepin.spec.bootstrap.client.SpecTest;
 import org.candlepin.spec.bootstrap.client.api.CloudRegistrationClient;
 import org.candlepin.spec.bootstrap.client.api.OwnerClient;
+import org.candlepin.spec.bootstrap.client.request.Request;
+import org.candlepin.spec.bootstrap.client.request.Response;
 import org.candlepin.spec.bootstrap.data.builder.Consumers;
 import org.candlepin.spec.bootstrap.data.builder.Contents;
 import org.candlepin.spec.bootstrap.data.builder.Owners;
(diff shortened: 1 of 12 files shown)


spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
```diff
@@ -464,6 +466,54 @@ public void shouldNotImplementedWithRegOnlyOfferTypesAndNoOwnerForV2Auth(String
...
+        adminClient.owners().createPool(owner.getKey(), Pools.random(prod));
+
+        adminClient.hosted().createOwner(owner);
+        adminClient.hosted().createProduct(prod);
+        adminClient.hosted().createSubscription(Subscriptions.random(owner, prod));
+        adminClient.hosted().associateOwnerToCloudAccount(accountId, owner.getKey());
+        adminClient.hosted().associateProductIdsToCloudOffer(offerId, List.of(prod.getId()));
+
+        CloudAuthenticationResultDTO cloudAuthDTO = ApiClients.noAuth().cloudAuthorization()
+            .cloudAuthorizeV2(accountId, instanceId, offerId, "test-type", "");
```
nikosmoum, line 488: As I mentioned elsewhere, this issue is unrelated to autoregistration. It is related to registration of a cloud VM (which will send cloud-related facts), regardless of the type of auth (you could move this test to `ConsumerResourceSpecTest` and use username/password or AK; as long as the consumer has those aws fact keys with null values, you'll be able to reproduce)

src/main/java/org/candlepin/model/ConsumerCloudData.java
```diff
@@ -232,46 +234,38 @@ public ConsumerCloudData addCloudOfferingIds(String... cloudOfferingIds) {
...
      * @throws IllegalArgumentException
-     *         if {@code cloudOfferingIds} is {@code null}, contains {@code null} elements,
-     *         or if the combined cloud offering IDs exceed the maximum allowed length of 255 characters
+     *  if {@code cloudOfferingIds} is {@code null}, or if the combined cloud offering IDs exceed the
+     *  maximum allowed length of 255 characters
      */
     public ConsumerCloudData addCloudOfferingIds(Collection<String> cloudOfferingIds) {
         if (cloudOfferingIds == null) {
-            throw new IllegalArgumentException("cloudOfferingIds is null or empty");
+            throw new IllegalArgumentException("cloudOfferingIds is null");
```
nikosmoum, line 248: All 3 fields (cloudAccountId, cloudInstanceId, cloudOfferingId) are nullable in the DB, which is according to what we want. So we shouldn't be throwing IAE if they're null. The only field that is non-nullable (except the auto-generated id and the Consumer reference) is the `cloudProviderShortName`.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
Ceiu: Ugh...

So do we care about the cloud registration details when it's not a cloud registration request? Seems like what you're describing is a second bug where during a non-cloud registration, we're still treating it like one and processing facts very specific to cloud registration.

Additionally, because we opted to have very highly specific facts for each cloud provider, we probably need tests for every set of cloud facts. Worse, I fear that combinations of them are also "valid".

src/main/java/org/candlepin/model/ConsumerCloudData.java
Ceiu: Due to a ton of various abstractions before we get here, that still shouldn't happen, and what you're referencing in this line isn't going to be triggered.

I also find it odd that we don't accept null but we accept empty here, but the issue was triggered by null elements in the collection.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
nikosmoum: I've explained in more detail in the relevant slack thread, but the cloud facts _are not used/related_ to autoregistration. They're used for reporting reasons by other upstream systems (e.g. Swatch), for any cloud system that registers, regardless if they used autoreg or other means of registration

src/main/java/org/candlepin/model/ConsumerCloudData.java
nikosmoum: Right I agree this shouldn't happen as part of this particular bug, but I thought it is a good idea to codify it on the entity itself that nulls are allowed. And maybe null should be set even in case the input is empty (what is the value of persisting an empty value? nothing afaik)

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
Ceiu: Please link the slack thread in the Jira so it can be tracked.

That aside, we need to be a lot more clear on our terminology. Facts are generally not operative parameters for us, and calling a system a "cloud system" heavily implies cloud registration -- especially given everything else around this.

Now, rewinding a bit, if these are only used for reporting and are non-operative, why are we putting so much extra, multi-class processing behind these? Why are they stored separately? What happens if the facts update later? What if the values desync from an actual cloud registration (where appropriate)? What if more are added later?

There are so many questions given what you're implying about their lack of importance relative to how much code and state exists surrounding them.

nikosmoum: - They're stored separately because it is impossible to do any performant querying against the facts table.
- The current multi-class processing (especially the excessive validation) is a miss on the initial implementation/review.
- These facts never change, and this was taken into account when designing this (no need for updates to these values). A cloud VM will always belong to the same cloud provider, cloud account, have the same cloud instance id, and the same cloud offering ids. You can't 'move' cloud vms from AWS to Azure for example, and you can't move it from one AWS account to another. The cloud instance id can't change because it is the unique id assigned for it. The cloud offering ids won't change because they are linked to a specific cloud image type, if that changed, that would mean it is an entirely different VM based on an entirely different cloud image type.

Ceiu: For the first point: true enough, I suppose. But while our storage issues are something we aim to address in the future, any convention or specification we defined now is still something we have to carry forward. And if we get to a place where we're back to querying against facts directly instead of storing facts we care about specially, we're back to this weird point of some facts being special and having special logic and behavior, and some not.

For the last point: it doesn't matter what the client conventions are (today) when *our* limitations are very much not that. Facts can change arbitrarily, and it's little more than a key/value pair from our perspective. Assuming the client will never change nor misbehave is a grave error in our design.

Glancing through the code it looks like we pull the facts once during registration (and in a weird place relative to all the other input processing), and then never again. So while the facts could change, we won't ever process them again as cloud facts; which itself is really weird.

nikosmoum: It wasn't really client conventions, but that cloud provider metadata services wouldn't be updating this data for a specific cloud VM logically that was used to decide not to have these being updatable. But I guess adding logic to make them updateable in candlepin as a guard against subman bugs is a good idea (maybe as a separate task than this bug fix?)

nikosmoum: The scope of this PR is to fix the immediate blocker (registration failing for all cloud systems against Stage CP), when cloud facts which have a key but a null value are sent

src/main/java/org/candlepin/model/ConsumerCloudData.java
Ceiu: My worry here is that this is spidering into redesigning the entity and having to change/update the initial parsing layer, as well as writing a bunch of tests that were overlooked.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
Ceiu: I've added tests for all sets of cloud facts for both cloud registration and standard registration.

src/main/java/org/candlepin/model/ConsumerCloudData.java
Ceiu: Fixed, I guess. Tests updated accordingly.

src/test/java/org/candlepin/model/ConsumerCloudDataTest.java
```diff
@@ -12,63 +12,59 @@
...
-        consumerCloudData.setCloudProviderShortName(validName);
-        assertEquals(validName, consumerCloudData.getCloudProviderShortName());
+    public void testSetCloudProviderShortNameWithLongValue() {
+        ConsumerCloudData ccdata = new ConsumerCloudData();
+
+        String value = "a".repeat(ConsumerCloudData.CLOUD_PROVIDER_MAX_LENGTH + 1);
 
         Exception exception = assertThrows(IllegalArgumentException.class, () -> {
-            consumerCloudData.setCloudProviderShortName(null);
+            ccdata.setCloudProviderShortName("1234567890123456");
```
joshmalbrecht, line 65: Use the "value" variable here instead of the "1234567890123456" string.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceSpecTest.java
```diff
@@ -107,6 +110,64 @@ public void shouldPopulateGeneratedFieldsWhenCreatingConsumers() {
...
+        return Stream.of(
+            Arguments.of(awsFacts),
+            Arguments.of(azureFacts),
+            Arguments.of(gcpFacts));
+    }
+
+    @ParameterizedTest
+    @MethodSource("cloudFactsSource")
+    public void shouldAllowRegistrationWithNullDataInCloudFacts(List<String> facts) throws Exception {
+        ApiClient adminClient = ApiClients.admin();
```
joshmalbrecht, line 130: We can remove this line or use this adminClient instead of the adminClient that is created in the "setup" method. This also includes `shouldAllowRegistrationWithEmptyDataInCloudFacts`.

joshmalbrecht (changes requested):
Looks good! I just noticed a couple of small changes

src/test/java/org/candlepin/model/ConsumerCloudDataTest.java
Ceiu: Argh! I even re-wrote the whole test to avoid this very mistake. Fixed.

spec-tests/src/test/java/org/candlepin/spec/consumers/ConsumerResourceSpecTest.java
Ceiu: Fixed

src/test/java/org/candlepin/model/ConsumerCloudDataTest.java
joshmalbrecht: I figured that you probably refactored the test haha

joshmalbrecht (approved):

mbartusiak:
Hi,

is this fix going to be released for the version 4.4 of candlepin?

Ceiu:
Yup, see #4951

nikosmoum:
@mbartusiak Hi, can I ask where/how you encountered this issue? are you an upstream foreman user?

mbartusiak:
@nikosmoum 
Hi, yes I am, I tried to migrate to Foreman 3.11.4 but I had this exception thrown whilst trying to register a new client:

`2024-11-20 17:37:06,859 [thread=https-openssl-nio-127.0.0.1-23443-exec-7] [req=c24a5bc6-460a-4c32-b707-c1975a3d569a, org=, csid=d7ae88fa-dcae-420e-b590-a4f13e4296fa] ERROR org.candlepin.exceptions.mappers.CandlepinExceptionMapper - Runtime Error cloudOfferingIds contains null element at org.candlepin.model.ConsumerCloudData.setCloudOfferingIds:331
java.lang.IllegalArgumentException: cloudOfferingIds contains null element
	at org.candlepin.model.ConsumerCloudData.setCloudOfferingIds(ConsumerCloudData.java:331)
	at org.candlepin.resource.ConsumerResource.createConsumerCloudData(ConsumerResource.java:1050)
	at org.candlepin.resource.ConsumerResource.createConsumer(ConsumerResource.java:1013)`

I found it surprising because I had migrated a second instance just a few days ago without encountering this exception. Upon investigation, I noticed that my other instance was using Candlepin 4.4.16, while the 'broken' instance was using 4.4.19. Based on this, I suspect that [this commit](https://github.com/candlepin/candlepin/commit/05cd538c2c5c6f76e1c093827a41542468c2273c) might be the culprit.

nikosmoum:
Thanks, I'm working on a candlepin-4.4.20-1 release with this fix

PR #4946, opened 2024-11-21
[4.5] Revert "CANDLEPIN-938: Reverted explicit content bulk translation"

Ceiu:
- Reverts commit 285644b525460623645831487905a50eac461c69 that reverted the patch for explicit bulk content translation


Ceiu:
How many times can one person make the same branch targeting error?

PR #4957, opened 2024-11-28
[4.4.1] Cherry-picked autoreg v2

bakajstep:


nikosmoum (commented):
Looks good, the only thing is that the change that was required should be made as part of the commit that needed it (the 1st commit looks like). When doing cherry-picks and rebases we should strive that each individual commit can stand on its own (it can be compiled/deployed and all tests pass)

bakajstep:
> Looks good, the only thing is that the change that was required should be made as part of the commit that needed it (the 1st commit looks like). When doing cherry-picks and rebases we should strive that each individual commit can stand on its own (it can be compiled/deployed and all tests pass)

Ack, done

nikosmoum:
Closing the PR since we don't intend to do this backport just yet. If we have to, we'll re-open this PR.

PR #4961, opened 2024-12-03
[M] Added blank address for local pg_hba.conf entry

joshmalbrecht:
- Added a blank address for the el8 and el9 ansible tasks to prevent an issue with the postgresql_pg_hba module and allow the Vagrant VM to be created.

The following is a workaround to an issue with postgresql.postgresql_pg_hba component in Version 3.9.0 that prevents the vagrant VM from being created.

https://github.com/ansible-collections/community.postgresql/issues/777


Ceiu (approved):
Unfortunate that this is necessary, but it gets the job done.

PR #4964, opened 2024-12-11
[4.4] 2332067: Disallowed Manual Subscription Attachment in SCA Mode

bakajstep:
- In SCA mode, manual attachment of subscriptions was no longer needed. Auto-attachment had already been disabled at the Candlepin and Satellite (6.9+) levels. Only manifest consumers ('candlepin' or 'satellite' types) were allowed to attach entitlements to ensure manifest generation continued to function properly.


nikosmoum (commented):
the bugzilla check is failing because you're using the wrong bugzilla ID in the commit. you need to clone https://bugzilla.redhat.com/show_bug.cgi?id=2017774, and update the clone's Target Release to 4.4. Then update this commit message with the clone's BZ id

bakajstep:
> the bugzilla check is failing because you're using the wrong bugzilla ID in the commit. you need to clone https://bugzilla.redhat.com/show_bug.cgi?id=2017774, and update the clone's Target Release to 4.4. Then update this commit message with the clone's BZ id

done

nikosmoum (approved):

PR #4969, opened 2024-12-15
Update ko.po

simmon-nplob:
update for simmon


nikosmoum (approved):

simmon-nplob:
Dear @nikosmoum - I thought it would be better to make some brief changes, so I asked for it. Thanks, Have a nice day! from Seoul, Korea

PR #4971, opened 2024-12-18
[M] CANDLEPIN-983: Cached tomcat package in GH action

bakajstep:


.github/containers/cs9.Containerfile
```diff
@@ -0,0 +1,80 @@
...
+COPY ${WAR_FILE} ./candlepin.war
+
+# Prepare development certs
+RUN mkdir -p /app/certs
+WORKDIR /app/certs
+COPY ./bin/deployment/gen_certs.sh .
+RUN ./gen_certs.sh --cert_out ./candlepin-ca.crt --key_out ./candlepin-ca.key --hostname candlepin; \
+    rm gen_certs.sh;
+
+FROM quay.io/centos/centos:stream9
```
joshmalbrecht, line 30: @nikosmoum I forget, but were we intentional about wanting to test `ubi9-minimal:latest` base image in our spec tests? Introducing this cs9.Containerfile and using it in our spec tests GH actions means we no longer test against `ubi9-minimal:latest`.

joshmalbrecht (commented):
Just had a question about the `cs9.Containerfile`

.github/containers/cs9.Containerfile
joshmalbrecht: I remember the conversation now, and it was around wanting to test both rhel8 and rhel9 equivalent in our spec tests. So, this change is fine!

PR #4972, opened 2024-12-18
CANDLEPIN-930: Created the POST /hypervisors/{owner}/guests endpoint

joshmalbrecht:
- Created the POST /hypervisors/{owner}/guests endpoint for retrieving the most recent hypervisor host consumer information for a list of provided guest consumer UUIDs.

# Notes on the changes

## Endpoint definition

Unfortunately we could not use the GET verb for this endpoint definition because OpenApi 3.0 does not allow for a request body for GET requests. So, the generated client will not send a body for this endpoint if it uses the GET verb.

Quote from the following link:

> GET, DELETE and HEAD are no longer allowed to have request body because it does not have defined semantics as per [RFC 7231](https://tools.ietf.org/html/rfc7231#section-4.3).

https://swagger.io/docs/specification/v3_0/describing-request-body/describing-request-body/

## Performance

To get an idea of performance, I created a spec test to create 1000 guest consumers that mapped to 1 host consumer each, and then made a request to the endpoint with all 1000 consumer UUIDs. I got the following total request response times:
- 635 ms
- 591 ms

The performance spec test:
```Java
    @Test
    public void perf() {
        ApiClient admin = ApiClients.admin();
        OwnerDTO owner = admin.owners().createOwner(Owners.random());

        List<String> consumerUuids = new ArrayList<>();

        for (int i = 0; i < 1000; i++) {
            ConsumerDTO host = admin.consumers().createConsumer(Consumers.random(owner)
                .name(StringUtil.random("host-" + i + "-")));

            String guestVirtUuid = StringUtil.random("host-" + i + "-guest-virt-uuid-");
            ConsumerDTO guest = createGuest(admin, owner, guestVirtUuid);

            consumerUuids.add(guest.getUuid());
            linkHostToGuests(admin, host, guestVirtUuid);
        }

        List<HypervisorConsumerWithGuestDTO> actual = admin.hypervisors()
            .getHypervisorsAndGuests(owner.getKey(), consumerUuids);

        assertThat(actual)
            .isNotNull()
            .hasSize(1000);
    }
```

### DIFF
--- a/api/candlepin-api-spec.yaml
+++ b/api/candlepin-api-spec.yaml
@@ -4059,6 +4059,57 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /hypervisors/{owner}/guests:
+    # Note that we had to use the POST verb instead of the GET verb here because we need to provide a list of
+    # consumer UUIDs in the body of the request, but OpenAPI 3.0 does not allow a body of a GET request.
+    post:
+      tags:
+        - hypervisors
+      description: |
+        Retrieves the latest hypervisor host information for the provided list of guest consumer UUIDs. If a
+        guest is mapped to multiple hypervisor host consumers, then the most recent host will be returned.
+        This endpoint accounts for case sensitivity and endianness differences in the guest virtual UUID that 
+        is returned by different hypervisors.
+      operationId: getHypervisorsAndGuests
+      parameters:
+        - name: owner
+          in: path
+          description: Owner key
+          required: true
+          schema:
+            type: string
+      requestBody:
+        description: |
+          Contains the list of guest Consumer UUIDs to retrieve host information for. Any non-guest
+          Consumer UUIDs that are provided will be ignored.
+        required: true
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                type: string
+      security: []
+      x-java-response:
+        type: java.util.stream.Stream
+        isContainer: true
+      responses:
+        200:
+          description: A successful operation
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/HypervisorConsumerWithGuestDTO'
+        400:
+          description: |
+            Invalid owner key or the number of provided guest consumer UUIDs exceeds the configured limit.
+        404:
+          description: Owner with provided key was not found.
+        default:
+          $ref: '#/components/responses/default'
+
   /hypervisors/{owner}/heartbeat:
     put:
       tags:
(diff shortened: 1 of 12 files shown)


Ceiu:
Are we sure we can't force OpenAPI to do what we want here? Their assessment of the RFC is incorrect.

> A payload within a GET request message has no defined semantics; sending a payload body on a GET request might cause some existing implementations to reject the request.

It *might* cause some *existing* server-side implementations to reject the request. That's not a reason to reject it universally. Moreover, RFC 9110 more clearly states, emphasis mine:

> Although request message framing is independent of the method used, content received in a GET request has no generally defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack ([Section 11.2](https://www.rfc-editor.org/rfc/rfc9112#section-11.2) of [[HTTP/1.1](https://www.rfc-editor.org/rfc/rfc9110#HTTP11)]). A client SHOULD NOT generate content in a GET request **unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported.**

That is, if the server and client both agree that it's supported, then it's okay. OpenAPI needs to get out of the way here.

joshmalbrecht:
@Ceiu Yeah, I agree with you and I would rather use GET for this operation.  I went down a bit of a rabbit hole because this interpretation is interesting and the forceful limitation in a client generator based on this interpretation is even more interesting. I found the following Github discussion on this topic and many others were expressing the same concerns we have.

**Github discussion on this topic:**
- https://github.com/swagger-api/swagger-ui/issues/2136

In that thread, there is a link to a blog post that further explains OpenAPI's stance on their interpretations:
- https://evertpot.com/get-request-bodies/

## Can we force the OpenApi generated client to send the body on a GET request?

Looking at the generated client, the problematic check is in the `buildRequest` method in the `ApiClient` class [0] where the method uses the `HttpMethod.permitsRequestBody` method from the `okhttp3.internal.http` package [1] to determine if the body is allowed on the client request. The `HttpMethod.permitsRequestBody` will return false if the method verb is GET or HEAD.

```
  @JvmStatic // Despite being 'internal', this method is called by popular 3rd party SDKs.
  fun permitsRequestBody(method: String): Boolean = !(method == "GET" || method == "HEAD")
```

So, unfortunately we won't be able to force the generated client to send the request body.

Links:
- [0] https://github.com/OpenAPITools/openapi-generator/blob/43f59bab8182ca446e00c1aa51b5ac496fd583de/samples/client/echo_api/java/okhttp-gson/src/main/java/org/openapitools/client/ApiClient.java#L1234-L1235
- [1] https://github.com/square/okhttp/blob/fed29a20217e1834164d89b2d9782d74569b0ccb/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/HttpMethod.kt#L37

Ceiu:
Out of curiosity, what does the request to RHSM API look like? Is it also a POST?

joshmalbrecht:
Due to design updates, I am closing this PR

PR #4992, opened 2025-01-14
[M] CANDLEPIN-980: Updated RHEL dependency from pki-servlet-engine to tomcat

bakajstep:
- Re-introduced the 'tomcat' package on all supported RHELs(RHEL8 and RHEL9) and deprecated 'pki-servlet-engine'.

### DIFF
--- a/candlepin.spec.tmpl
+++ b/candlepin.spec.tmpl
@@ -45,11 +45,7 @@ BuildRequires: selinux-policy-doc
 Requires: java-17 >= 1:17.0.0
 Requires: wget
 # TODO: remove the version restriction on Tomcat once CP supports the Java EE namespace changes
-%if 0%{?fedora} || 0%{?rhel} >= 9
 Requires: tomcat < 1:10.0.0
-%else
-Requires: pki-servlet-engine
-%endif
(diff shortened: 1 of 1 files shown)


joshmalbrecht (commented):
Did we validate the following: 
- "The generated candlepin RPM can be installed/upgraded on a Satellite instance and verify that pki-servlet-engine is not a dependency, but tomcat is."

bakajstep:
> Did we validate the following:
> 
> * "The generated candlepin RPM can be installed/upgraded on a Satellite instance and verify that pki-servlet-engine is not a dependency, but tomcat is."

For the installed/upgraded part, the last time I worked on this task, it was enough to simply check if it was installable on RHEL. Since Satellite is running on RHEL, I believe that should suffice. What do you think, @nikosmoum?
As for the second part of the "done" criteria, that’s why I shared the console output of the installation. During the installation, you can see the `Installing dependencies` section, which lists some `tomcat` packages, but there is no mention of `pki-servlet-engine`.

nikosmoum (approved):
Approving, since @bakajstep tested the rpm installation/upgrade on a satellite setup, and it works

PR #5003, opened 2025-01-17
CANDLEPIN-989: Updated AdapterEvent to sealed interface

joshmalbrecht:
- Updated the AdapterEvent class from interface to sealed interface and removed the getEventType method.
- Removed the EventType enum.
- Removed the getEventType method from CloudCheckInEvent.

Tagging @nikosmoum as this deviates from the initial design.


src/main/java/org/candlepin/service/model/AdapterEvent.java
```diff
@@ -18,7 +18,7 @@
  * The AdapterEvent interface represents a Candlepin event that should be published using the
  * {@link EventAdapter} and consumed by external components.
  */
-public interface AdapterEvent extends ServiceAdapterModel {
+public sealed interface AdapterEvent permits CloudCheckInEvent {
```
Ceiu, line 21: I don't remember if we discussed this, but we don't strictly need this to be sealed. Pattern matching and junk still works with or without a sealed interface, they'll just need a default or top-level branch to handle the unknown ones.

If we did discuss this and you're making a preference-based decision, then that's fine. I think for our purposes, either is acceptable.

Ceiu (approved):

Ceiu:
Since he was pinged directly, I'll give Nikos a chance to review and merge this.

nikosmoum (approved):
Looks good

PR #5007, opened 2025-01-29
[F] CHAINSAW-356: Added last content update field to environments

bakajstep:
- Updated environments to record the last content update timestamp whenever content was promoted or demoted to an environment, or the environment's prefix was changed. Used this timestamp to determine if the SCA certificate needed regeneration: if the last content update is newer than the certificate’s timestamp, the certificate needs to be regenerated.

### DIFF
--- a/src/main/java/org/candlepin/model/Environment.java
+++ b/src/main/java/org/candlepin/model/Environment.java
@@ -15,8 +15,10 @@
 package org.candlepin.model;
 
 import org.candlepin.util.SetView;
+import org.candlepin.util.Util;
 
 import java.io.Serializable;
+import java.util.Date;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -90,12 +92,15 @@ public class Environment extends AbstractHibernateObject<Environment> implements
     @Size(max = CONTENT_PREFIX_MAX_LENGTH)
     private String contentPrefix;
 
+    @Column(name = "last_content_update", nullable = false)
+    private Date lastContentUpdate;
+
     @OneToMany(mappedBy = "environment", targetEntity = EnvironmentContent.class,
         cascade = CascadeType.ALL)
     private Set<EnvironmentContent> environmentContent = new HashSet<>();
 
     public Environment() {
-        // Intentionally left empty
+        this.syncLastContentUpdate();
     }
 
     public Environment(String id, String name, Owner owner) {
@@ -105,6 +110,8 @@ public class Environment extends AbstractHibernateObject<Environment> implements
         if (owner != null) {
             this.setOwner(owner);
         }
+
+        this.syncLastContentUpdate();
     }
 
     /**
@@ -236,8 +243,62 @@ public class Environment extends AbstractHibernateObject<Environment> implements
         return contentPrefix;
     }
 
+    /**
+     * Sets the content prefix for this environment. Because changing the content prefix
+     * requires the SCA content payloads to be re-generated with the correct paths, this method
+     * also updates the environment's {@code lastContentUpdate} timestamp by calling
+     * {@link #syncLastContentUpdate()}.
+     *
+     * @param contentPrefix the new content prefix for this environment
+     * @return this {@code Environment} instance
+     * @throws IllegalArgumentException if {@code contentPrefix} exceeds {@value #CONTENT_PREFIX_MAX_LENGTH}
+     * characters
+     */
     public Environment setContentPrefix(String contentPrefix) {
+        if (contentPrefix != null && contentPrefix.length() > CONTENT_PREFIX_MAX_LENGTH) {
+            throw new IllegalArgumentException("contentPrefix is too long");
+        }
+
         this.contentPrefix = contentPrefix;
+        this.syncLastContentUpdate();
+        return this;
+    }
+
+    /**
+     * Returns the date of the last content update for this environment.
+     * If {@code lastContentUpdate} is not set, this method falls back to the environment's creation date,
+     * and if that's also not available, the current date is returned.
+     *
+     * @return the date of the last content update, never {@code null}
+     */
+    public Date getLastContentUpdate() {
+        return Util.firstOf(this.lastContentUpdate, this.getCreated(), new Date());
+    }
+
+    /**
+     * Sets the date of the last content update for this environment.
+     *
+     * @param update the new last content update date; must not be {@code null}
+     * @return this {@code Environment} instance
+     * @throws IllegalArgumentException if {@code update} is {@code null}
+     */
+    public Environment setLastContentUpdate(Date update) {
+        if (update == null) {
+            throw new IllegalArgumentException("update is null");
+        }
+
+        this.lastContentUpdate = update;
+        return this;
+    }
+
+    /**
+     * Updates the {@code lastContentUpdate} field to the current date and time,
+     * then returns this {@code Environment} instance.
+     *
+     * @return the updated {@code Environment} instance
+     */
+    public Environment syncLastContentUpdate() {
+        this.setLastContentUpdate(new Date());
         return this;
     }
 
--- a/src/main/java/org/candlepin/resource/EnvironmentResource.java
+++ b/src/main/java/org/candlepin/resource/EnvironmentResource.java
@@ -399,6 +399,8 @@ public class EnvironmentResource implements EnvironmentApi {
         Set<String> contentIds;
         try {
             contentIds = this.batchCreate(contentToPromote, environment);
+            // TODO: Remove the line below once certificate generation based on the `lastContentUpdate` field
+            // in environments is implemented.
             this.contentAccessManager.syncOwnerLastContentUpdate(environment.getOwner());
         }
         catch (PersistenceException pe) {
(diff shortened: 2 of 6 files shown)


src/main/java/org/candlepin/model/Environment.java
```diff
@@ -241,6 +248,25 @@ public Environment setContentPrefix(String contentPrefix) {
...
+
+    public Environment setLastContentUpdate(Instant update) {
+        if (update == null) {
+            throw new IllegalArgumentException("update is null");
+        }
+        this.lastContentUpdate = update;
+        return this;
+    }
+
+    public Environment syncLastContentUpdate() {
```
Ceiu, line 265: Missing javadoc for all the new methods

src/test/java/org/candlepin/resource/EnvironmentResourceTest.java
```diff
@@ -665,6 +667,61 @@ void shouldCreateConsumerInMultipleEnvironment() {
         assertEquals("env2", dto.getEnvironments().get(2).getId());
     }
 
+    @Test
+    void shouldUpdateLastContentFieldOnContentPromote() {
```
Ceiu, line 671: Missing scope modifiers

Ceiu (changes requested):
The commit messaging also needs a bit of work here:
- title is missing the word "update" of "last content update"
- body describing the change goes into detail about work that does not (yet) exist; specifically, anything involving SCA content payload generation

src/main/java/org/candlepin/model/Environment.java
```diff
@@ -241,6 +248,25 @@ public Environment setContentPrefix(String contentPrefix) {
         return this;
     }
 
+    public Instant getLastContentUpdate() {
+        // If we don't have a content update date yet, return this environments's creation date. Never
+        // actually return null.
+        return Util.firstOf(this.lastContentUpdate, this.getCreated().toInstant(), Instant.now());
```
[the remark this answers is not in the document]
bakajstep: ack, done

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -261,6 +261,10 @@ public EnvironmentDTO updateEnvironment(@Verify(Environment.class) String envId,
 
         if (contentPrefix != null) {
             environment.setContentPrefix(!contentPrefix.isBlank() ? contentPrefix : null);
+
+            // Since the content prefix is changing, we need to reroll the SCA content payloads for this
+            // env so they get the right paths
+            environment.syncLastContentUpdate();
```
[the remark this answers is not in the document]
bakajstep: done

src/main/java/org/candlepin/model/Environment.java
bakajstep: done

src/test/java/org/candlepin/resource/EnvironmentResourceTest.java
bakajstep: done

src/main/java/org/candlepin/model/Environment.java
```diff
@@ -241,6 +248,25 @@ public Environment setContentPrefix(String contentPrefix) {
...
+    }
+
+    public Environment setLastContentUpdate(Instant update) {
+        if (update == null) {
+            throw new IllegalArgumentException("update is null");
+        }
```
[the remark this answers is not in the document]
bakajstep: done

src/main/resources/db/changelog/202501281629000-add-last-content-update-field.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="202501281629000-1" author="sbakaj">
```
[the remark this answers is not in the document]
bakajstep: done, updated also file name

src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -440,6 +442,8 @@ public AsyncJobStatusDTO demoteContent(@Verify(Environment.class) String envId,
 
         try {
             this.envContentCurator.bulkDelete(demotedContent.values());
+            // TODO: Remove the line below once certificate generation based on the `lastContentUpdate` field
+            // in environments is implemented.
```
Ceiu, line 446: :+1:

Ceiu (changes requested):
Latest revision looks pretty good. Just need to add tests for the new functionality in Environment.

bakajstep:
> Latest revision looks pretty good. Just need to add tests for the new functionality in Environment.

done, also updated test in EnvironmentResourceTest so they do not fail on same timestamp

joshmalbrecht (approved):

Ceiu (approved):

src/test/java/org/candlepin/model/EnvironmentTest.java
```diff
@@ -42,4 +47,52 @@ public void testGetOwnerKeyWithNoOwner() {
         assertNull(environment.getOwnerKey());
     }
 
+    @Test
+    public void testSetContentPrefix() {
```
Ceiu, line 51: We could split this up into two distinct, named tests:
- testSetContentPrefix(): tests the default behavior of changing the context prefix
- testSetContentPrefixChangesLastContentUpdate(): verifies that changing the content prefix also triggers a change to the last content update

This gives us more clarity in terms of what we're testing and captures the full requirements in two declaritive tests.

src/test/java/org/candlepin/model/EnvironmentTest.java
```diff
@@ -42,4 +47,52 @@ public void testGetOwnerKeyWithNoOwner() {
...
+
+        environment.setContentPrefix("newPrefix");
+        Date after = environment.getLastContentUpdate();
+
+        assertEquals("newPrefix", environment.getContentPrefix());
+        assertTrue(after.after(before));
+    }
+
+    @Test
+    public void testSetLastContentUpdate() {
```
Ceiu, line 66: This should be renamed to something indicative of verifying the UUT rejects a null input

src/test/java/org/candlepin/model/EnvironmentTest.java
```diff
@@ -42,4 +47,52 @@ public void testGetOwnerKeyWithNoOwner() {
...
+        Date date = minusTwoSeconds();
+
+        assertThrows(IllegalArgumentException.class, () -> environment.setLastContentUpdate(null));
+
+        environment.setLastContentUpdate(date);
+        assertEquals(date, environment.getLastContentUpdate());
+    }
+
+    @Test
+    public void testGetLastContentUpdate() {
```
Ceiu, line 77: We should add more testing around this method to verify the various conditions we've defined in the comments and design.

This does get a bit spicy with the constructors calling syncLastContentUpdate. I'm somewhat torn on if that should be a thing since it's a tad extraneous and could lead to incorrect results if Hibernate gets its order of operations wrong. But if it gets in the way of us verifying output, maybe it should be cut. I'll leave that decision to you.

Ceiu (changes requested):

src/test/java/org/candlepin/model/EnvironmentTest.java
bakajstep: done

src/test/java/org/candlepin/model/EnvironmentTest.java
Ceiu: Based on discussions in slack, this comment can be disregarded.

For those not part of the discussion: if we don't set the lastContentUpdate field during construction (either directly setting it or calling sync/setLastContentUpdate), it leaves the entity in an invalid state with a null LCU which will get rejected by the backing DB schema. So while testing does lose some granularity, I would rather not introduce the potential for invalid or undefined states for the sake of easing our unit tests.

src/test/java/org/candlepin/model/EnvironmentTest.java
bakajstep: done, also split the test for more clarity

src/test/java/org/candlepin/model/EnvironmentTest.java
Ceiu: Sorry, I should have also specified that we should be testing all of the input types we consider to be valid for the context prefix. A parameterized test here would be acceptable, just so long as we capture that we permit things like nulls, or empty strings and normal prefixes.

We do also have a max length on the content prefix which we should be validating against as well.

Basically, our unit tests should reflect our specification, as they are the first line of defense when it comes to ensuring we don't violate our own convention or introduce regression. It also shows intent for future devs who need to know whether or not we expected or intended certain inputs to work.

Ceiu (changes requested):

src/test/java/org/candlepin/model/EnvironmentTest.java
bakajstep: it should be done, but there is no validation for max lenght i added that validation in setter.

bakajstep: but none of the setter have lenght validation shoould i fix that part of this PR?

PR #5018, opened 2025-01-31
[M] CANDLEPIN-995: Updated the entity layering migration's tiebreakers

Ceiu:
- Updated the way the entity layering DB migration calculates its "best" rows in cases where collisions are expected to use window functions instead of nested aggregate queries
- Fixed a couple errors in how product attributes are queried in cases with multiple sources of a given attribute
- Fixed issues with the way custom product and content queries were flagging entities as "best", causing some entities to be omitted from the migration, or fail on partial migration

### DIFF
--- a/src/main/resources/db/changelog/20231003113535-entity_layering_schema.xml
+++ b/src/main/resources/db/changelog/20231003113535-entity_layering_schema.xml
@@ -373,8 +373,10 @@
         <validCheckSum>9:16c11c8edc80574ab030cd5fbcec5a07</validCheckSum>
         <!-- 4.4.10 -->
         <validCheckSum>9:56c16786668c16c345e8ce64ae890851</validCheckSum>
-        <!-- 4.4.11+ -->
+        <!-- 4.4.11 -->
         <validCheckSum>9:d11ca759108c7919ce364d946c825460</validCheckSum>
+        <!-- 4.5.2+ -->
+        <validCheckSum>9:5b03d74e0f43abd4e82e22303916f92f</validCheckSum>
 
         <sql>
             -- Migrate global products and content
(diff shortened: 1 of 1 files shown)


Ceiu:
Reviewers:
Please give this one a very thorough review, as there's a lot that can sneak through and cause problems for customers. Beyond the obvious "it works":

- aliases line up between the expression declarations in the CTEs and their queries; some older PostgreSQL and MariaDB environments get very upset when these don't perfectly line up
- window functions make sense; this is only the 4th or 5th time I've had to use one, so there's a possibility I'm still doing it wrong
- intent is clear for future maintainers that aren't me
- any other weird issues we've had in trying to thread the needle with native queries and cross-platform support that are escaping me at the moment

For testing, there are CP database dumps in the Sat JIRA attached to the CP task for this PR which can be used to verify this; but by all means, _try_ to break this in any way you can with goofy or malformed data. The more weird edge cases we can knock out now, the better.

Ceiu:
Oh, lastly, do not merge this without at least three approvals.

joshmalbrecht (approved):

src/main/resources/db/changelog/20231003113535-entity_layering_schema.xml
```diff
@@ -373,8 +373,10 @@
         <validCheckSum>9:16c11c8edc80574ab030cd5fbcec5a07</validCheckSum>
         <!-- 4.4.10 -->
         <validCheckSum>9:56c16786668c16c345e8ce64ae890851</validCheckSum>
-        <!-- 4.4.11+ -->
+        <!-- 4.4.11 -->
         <validCheckSum>9:d11ca759108c7919ce364d946c825460</validCheckSum>
+        <!-- 4.5.2+ -->
```
nikosmoum, line 378: once we backport this to candlepin-4.4-HOTFIX, I assume this should say 4.4.21+ right?

nikosmoum (approved):
Looks good. Lets get this backported

PR #5024, opened 2025-02-05
[F] CHAINSAW-357: Refactored SCACertficateGenerator

joshmalbrecht:
- Refactored the SCACertificateGenerator to allow individual component fetching.
- Created ContentAccessPayload which stores the content access payload for a SCA owner.


src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -439,6 +445,7 @@ public static String jobConfig(String jobKey, String cfgName) {
 
             this.put(IDENTITY_CERT_YEAR_ADDENDUM, "5");
             this.put(IDENTITY_CERT_EXPIRY_THRESHOLD, "90");
+            this.put(SCA_X509_CERT_EXPIRY_THRESHOLD, "10");
```
Ceiu, line 448: I wonder if this should be closer to the 90 days we use for the identity cert. Maybe bump it to 30 for the default and we can adjust accordingly later.

src/test/java/org/candlepin/model/ContentAccessPayloadTest.java
```diff
@@ -0,0 +1,203 @@
...
+        ContentAccessPayload payload = new ContentAccessPayload()
+            .setTimestamp(expected);
+
+        assertThat(payload.getTimestamp())
+            .isNotNull()
+            .isEqualTo(expected);
+    }
+
+    @Test
+    public void testToString() {
```
Ceiu, line 166: I'm iffy about adding tests for toString since we don't use this as anything other than debug output. Maybe as a quick verification that it can't return null or otherwise break, but I'm not sure we need to (nor want to) verify the format. Thoughts?

Ceiu (changes requested):
Some pretty minor things on the first pass. I'll take a closer look at testing on the next pass.

src/main/java/org/candlepin/model/ContentAccessPayload.java
```diff
@@ -0,0 +1,232 @@
...
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+/**
+ * A ContentAccessPayload entity represents an encoded and compressed collection of content that is typically
+ * included in a {@link SCACertificate}.
+ */
+@Entity
+@Table(name = ContentAccessPayload.DB_TABLE)
+public class ContentAccessPayload {
```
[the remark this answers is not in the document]
joshmalbrecht: Updated. I also had to add the `updated` and `created` columns to the `cp_content_access_payload` table to extend the AbstractHibernateObject and also have the `ContentAccessPayloadCurator` extend the `AbstractHibernateCurator`.

src/main/java/org/candlepin/model/ContentAccessPayloadCurator.java
```diff
@@ -0,0 +1,138 @@
...
+import javax.inject.Singleton;
+import javax.persistence.EntityExistsException;
+import javax.persistence.EntityManager;
+import javax.persistence.NoResultException;
+
+/**
+ * Curator responsible for datastore operation for the {@link ContentAccessPayload} entity.
+ */
+@Singleton
+public class ContentAccessPayloadCurator {
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -58,216 +64,326 @@
...
-        SCACertificate result = new SCACertificate();
-        result.setCert(cert.getCert() + cert.getContent());
-        result.setCreated(cert.getCreated());
-        result.setUpdated(cert.getUpdated());
-        result.setId(cert.getId());
-        result.setConsumer(cert.getConsumer());
-        result.setKey(cert.getKey());
-        result.setSerial(cert.getSerial());
-        return result;
+    private SCACertificate combineCertAndPayload(SCACertificate certificate, ContentAccessPayload payload) {
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

src/test/java/org/candlepin/model/ContentAccessPayloadTest.java
joshmalbrecht: I agree. We should at least verify that the we are generating a string value and don't error out. It does become annoying to maintain a test that verifies the exact string value as time goes on and fields are added/removed.

src/main/java/org/candlepin/model/ContentAccessPayloadCurator.java
```diff
@@ -0,0 +1,69 @@
...
+     *  the ID of the owner which owns the content access payload
+     *
+     * @param payloadKey
+     *  the payload of key of the content access payload
+     *
+     * @return the content access payload that has the provided payload key and belongs to the provided owner,
+     *  or null if the payload cannot be found
+     */
+    public ContentAccessPayload getContentAccessPayload(String ownerId, String payloadKey) {
+        if (ownerId == null || ownerId.isBlank()) {
```
Ceiu, line 43: As a note: these validations aren't bad, but in this particular case it isn't necessary. If either field is null or blank, the resultant query will look like `... WHERE owner_id = NULL AND payload_key = ''`; both of which will result in a valid but effectively dead query that won't return anything. In either case we end up on the null path.

This isn't a request to remove them, just an explanation

src/test/java/org/candlepin/pki/certs/SCACertificateGeneratorTest.java
```diff
@@ -163,18 +169,646 @@ private String urlEncode(String text) {
...
+        doReturn(false).when(v3CapabilityCheck)
+            .isCertV3Capable(consumer);
+
+        ContentAccessPayload actual = generator.getContentPayload(consumer);
+
+        assertNull(actual);
+    }
+
+    @Test
+    public void testGetSCAContentPayloadWithDifferentArchAndSupportedArches() throws Exception {
```
Ceiu, line 213: This is good, but we have three other states I'd like to see explicitly tested: consumer w/no arches specified at all, consumer with only arch, and consumer with only supported arches (parameterized with one, two, etc.).

There are also the cases of what happens when supported arches has duplicates, or overlaps with the base arch fact.

Ceiu (changes requested):

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -58,216 +64,322 @@
...
+            combined.setId(cert.getId());
+            combined.setKey(cert.getKey());
+            combined.setSerial(cert.getSerial());
+
+            return combined;
+        }
+        catch (Exception e) {
+            // Is this necessary? Why are we eating the whole exception? We're likely very broken at
+            // this point
+            log.error("Unexpected exception occurred while fetching SCA certificate for consumer: {}",
```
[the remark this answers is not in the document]
joshmalbrecht: Did we want to return a 429 response back to the client in case we hit a constraint violation? and if so, what should we set for the retry-after header?

src/test/java/org/candlepin/pki/certs/SCACertificateGeneratorTest.java
joshmalbrecht: Good call! I updated this test to be a parameterized test that takes a matrix of arches and supported arches and verifies that the correct content is added to the SCA content payload.

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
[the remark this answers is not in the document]
joshmalbrecht: Updated to throw a `ConcurrentCertificateCreationException` and I bubbled the exception to the resource request handlers where I caught the exception and comments to handle the exception in other Jiras.

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -58,216 +64,342 @@
...
+                container = new ContentAccessPayload()
+                    .setOwner(owner)
+                    .setPayloadKey(payloadKey)
+                    .setTimestamp(timestamp)
+                    .setPayload(payload);
+
+                try {
+                    contentAccessPayloadCurator.create(container);
+                }
+                catch (EntityExistsException e) {
```
Ceiu, line 357: Did this work? I was almost certain we'd need to catch an SQLException and then parse the message for "constraint violation" and our constraint name.

Ceiu: Also, my fear is that we won't actually get the exception until we commit the transaction -- which means another custom mapper or something. :(

src/main/java/org/candlepin/pki/certs/ContentAccessPayloadKeyBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+     * Sets the environments used to generate the content access payload key. If a null value is provided then
+     * no environments will be used.
+     *
+     * @param environments
+     *  the environments used to generate the content access payload key
+     *
+     * @return this instance of ContentAccessPayloadKeyBuilder
+     */
+    public ContentAccessPayloadKeyBuilder setEnvironments(List<Environment> environments) {
+        this.environments = environments == null ? new ArrayList<>() : environments;
```
Ceiu, line 70: If we're going to store an empty list, and we don't expect to be modifying it internally, let's use List.of instead so we aren't actually allocating a new list ever.

Ceiu (changes requested):
Getting there; just one question on that exception. Specifically, how are we testing it? We'll probably want a spec test to try to request the payload for multiple consumers at the same time to trigger and verify it doesn't blow up on us.

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
joshmalbrecht: I should have functionally tested this, but I didn't and I jumped the gun and assumed the `EntityExistsException` is what I needed to catch, but I think you are right! The SQLException is probably being thrown when we  commit the transaction which is unfortunate. I'll do some testing and adjust this and get some spec tests written!

Ceiu: As we discussed online, we won't be able to have spec tests to verify this behavior quite yet, but we'll want them once we get around to updating the specific endpoints which will need to be aware of this behavior.

src/main/java/org/candlepin/pki/certs/ContentAccessPayloadKeyBuilder.java
joshmalbrecht: This is good to know! Updated

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -58,216 +64,342 @@
...
+
+            consumer.setContentAccessCert(scaCertificate);
+        }
 
-    private SCACertificate updateCertificate(Consumer consumer,
-        Owner owner, SCACertificate existing) {
-        Date now = new Date();
-        Date expiration = existing.getSerial().getExpiration();
-        boolean isX509CertExpired = expiration.before(now);
+        int threshold = configuration.getInt(ConfigProperties.SCA_X509_CERT_EXPIRY_THRESHOLD);
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

src/main/java/org/candlepin/pki/certs/ContentAccessPayloadKeyBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+ * Builder class responsible for generating a key for a {@link ContentAccessPayload}.
+ */
+public class ContentAccessPayloadKeyBuilder {
+
+    private static final String HASH_ALGORITHM = "SHA-256";
+    private static final int SCA_CONTENT_ACCESS_PAYLOAD_KEY_VERSION = 1;
+    private static final HexFormat HEX_FORMATTER = HexFormat.of();
+
+    private Consumer consumer;
+    private List<Environment> environments = new ArrayList<>();
```
Ceiu, line 41: Missed a conversion to `List.of()`

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -58,216 +65,333 @@
...
     private final EntitlementPayloadGenerator payloadGenerator;
-    private final ConsumerCurator consumerCurator;
     private final EnvironmentCurator environmentCurator;
     private final PemEncoder pemEncoder;
     private final KeyPairGenerator keyPairGenerator;
     private final Signer signer;
     private final Provider<X509CertificateBuilder> certificateBuilder;
+    private final Configuration configuration;
+
+    private int x509CertExpirationThreshold;
```
Ceiu, line 98: nit: this should be final

Ceiu (approved):
Only one super minor thing if you feel like fixing it here. Otherwise, this seems like it catches everything.

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
joshmalbrecht: Updated

Ceiu:
@nikosmoum would you like to give this a once over to make sure I didn't miss anything? If not, feel free to merge this in.

src/main/java/org/candlepin/config/ConfigProperties.java
```diff
@@ -439,6 +445,7 @@ public static String jobConfig(String jobKey, String cfgName) {
 
             this.put(IDENTITY_CERT_YEAR_ADDENDUM, "5");
             this.put(IDENTITY_CERT_EXPIRY_THRESHOLD, "90");
+            this.put(SCA_X509_CERT_EXPIRY_THRESHOLD, "90");
```
nikosmoum, line 448: I think 90 days is too much for the SCA cert. The ID cert has 90 because it is way more critical to make sure to regenerate it before it gets expired, because if it does gets expired then subman has no way to communicate with candlepin any more, and the user would be forced to unregister/re-register the system.
SCA certs expiring is not that big of a problem, since they would get a new cert the next time they check in anyway. I would say something like 10 days is enough

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -58,216 +65,333 @@
...
+            org.candlepin.model.dto.Product container = createProductContainer(owner, environments);
+
+            X509Certificate x509Cert = createX509Cert(consumer.getUuid(), owner, serial, keypair,
+                container, start, end);
 
-            existing.setSerial(serial);
-            existing.setCert(this.pemEncoder.encodeAsString(x509Cert));
-            this.contentAccessCertificateCurator.saveOrUpdate(existing);
+            scaCertificate.setSerial(serial);
+            scaCertificate.setCert(pemEncoder.encodeAsString(x509Cert));
```
nikosmoum, line 318: we did a `curator.create()` earlier; don't we need to also do a merge after setting this data?

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2441,6 +2457,11 @@ public Object exportCertificates(
             throw new IseException(
                 i18n.tr("Unable to create entitlement certificate archive"), e);
         }
+        catch (ConcurrentCertificateCreationException e) {
+            // TODO: Handle exception
```
nikosmoum, line 2461: Is this also going to be handled in CHAINSAW-359?

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -58,216 +65,333 @@
...
+        SCACertificate scaCertificate = consumer.getContentAccessCert();
+        if (scaCertificate == null) {
+            log.info("Generating new SCA certificate key for consumer: \"{}\"", consumer.getUuid());
 
-        SCACertificate existing = new SCACertificate();
-        existing.setSerial(serial);
-        existing.setConsumer(consumer);
-        existing.setKeyAsBytes(this.pemEncoder.encodeAsBytes(keyPair.getPrivate()));
-        existing.setCert(this.pemEncoder.encodeAsString(x509Cert));
-        existing.setContent(this.createPayloadAndSignature(payloadBytes));
```
nikosmoum, line 213: Since the payload now lives in a separate entity/table, we should be removing the SCACertificate.content field, and the respective db column

src/main/java/org/candlepin/pki/certs/ConcurrentCertificateCreationException.java
```diff
@@ -0,0 +1,31 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.pki.certs;
+
+/**
+ *  An exception that occurs when a concurrent request causes a database constraint violation when persisting
+ *  a certificate.
+ */
+public class ConcurrentCertificateCreationException extends Exception {
```
nikosmoum, line 21: Perhaps `ConcurrentContentPayloadCreationException` would be more accurate for this?

src/main/resources/db/changelog/20250130000000-add-cp-content-access-payload-table.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
```
nikosmoum, line 7: not that important, but we should be setting dbchangelog-4.19.xsd (anything 4.x at minimum)

nikosmoum (changes requested):
I'll take another look, but here's a few comments for now

src/main/java/org/candlepin/pki/certs/ContentAccessPayloadKeyBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+            sorted.addAll(sarches);
+        }
+
+        return String.join(",", sorted);
+    }
+
+    private String getNormalizedEnvironments(List<Environment> environments) {
+        return environments.stream()
+            .map(Environment::getId)
+            .collect(Collectors.joining(","));
```
nikosmoum, line 120: Instead of relying on the callers to give us an ordered environment list, can we instead make sure this method orders them in a consistent manner? It could be in the same way the SQL in the environmentcurator does it (ascending by env id)

src/test/java/org/candlepin/pki/certs/ContentAccessPayloadKeyBuilderTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.candlepin.model.Consumer;
+import org.candlepin.model.Environment;
+import org.candlepin.model.Owner;
+import org.candlepin.test.TestUtil;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+public class ContentAccessPayloadKeyBuilderTest {
```
nikosmoum, line 29: Can we add a couple more tests here? Specifically, tests that check that
- the generated key is the same when the same list of environments but in different order is given
- the generated key is the same when the same list of architectures but in different order in the 'supported_architectures' fact

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -58,216 +65,333 @@
...
-                .collect(Collectors.toMap(cidFetcher, Function.identity(),
-                    (v1, v2) -> new ProductContent(v2.getContent(), v1.isEnabled() || v2.isEnabled())));
+        String payloadKey = new ContentAccessPayloadKeyBuilder(consumer)
+            .setEnvironments(environments)
+            .build();
 
-            byte[] payloadBytes = createContentAccessDataPayload(consumer, ownerContent, promotedContent);
-            existing.setContent(this.createPayloadAndSignature(payloadBytes));
-            this.contentAccessCertificateCurator.saveOrUpdate(existing);
+        ContentAccessPayload container = contentAccessPayloadCurator
```
nikosmoum, line 331: Can we rename container to payload? container within the context of SCA cert generation has been typically the name we use for the mock product that contains the full list of content

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
Ceiu: Nope. So long as the entity is managed and a new instance hasn't been spun up or anything, we need not do another merge. The concern would be if we ever shut off the auto commit stuff, then we probably would need it.

src/main/java/org/candlepin/pki/certs/ContentAccessPayloadKeyBuilder.java
Ceiu: We can't at the moment. Due to the fact that priority is buried deep in the consumer object and on the table itself, we don't have any way of sharing that once we pass it around. Or, rather, we *do* have a way, but this is it. We would need a custom wrapper object to enforce it.

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
```diff
@@ -58,216 +65,333 @@
...
-            this.createCertificate(consumer, owner);
+        if (!v3CapabilityCheck.isCertV3Capable(consumer)) {
+            log.debug("Consumer is not v3 cert capable");
+            return null;
+        }
 
-        // Due to the shenanigans we do to wrap and juggle the data in this cert, it's critical we invoke
-        // Hibernate's persist and update hooks by flushing the creation/update operations above so the
-        // dates are correct in the output cert object.
-        this.contentAccessCertificateCurator.flush();
```
nikosmoum, line 175: I think we need to preserve this flush, since when we're later doing this: `combined.setUpdated(payloadUpdated.after(certUpdated) ? payloadUpdated : certUpdated);` we could get in a situation were we generated a brand new X509 cert that has a null `certUpdated` date, so we end up setting the date to `payloadUpdated` here, which could be quite old. So we would end up sending back a very old date to the client, which it would send back as an old if-modified-since date, resulting in churn for the getContentAccessBody endpoint

Ceiu: We don't. We're already getting flushes on each create op, and even those aren't strictly necessary for all paths.

src/main/java/org/candlepin/config/ConfigProperties.java
Ceiu: Sure

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
joshmalbrecht: Updated

src/main/resources/db/changelog/20250130000000-add-cp-content-access-payload-table.xml
joshmalbrecht: Updated

src/main/java/org/candlepin/pki/certs/ConcurrentCertificateCreationException.java
joshmalbrecht: Yeah, I like that name much more and I think it is more accurate as well! Updated

src/main/java/org/candlepin/pki/certs/SCACertificateGenerator.java
joshmalbrecht: Removed the SCACertificate.content field and added a change set to drop the content column from the cp_cont_access_cert table.

src/main/java/org/candlepin/resource/ConsumerResource.java
joshmalbrecht: I didn't have the Jira at the time because it wasn't created yet, but it will be updated in CHAINSAW-376 and I updated the comment to add the Jira number

src/main/java/org/candlepin/config/ConfigProperties.java
joshmalbrecht: Updated to 10 days

src/test/java/org/candlepin/pki/certs/ContentAccessPayloadKeyBuilderTest.java
joshmalbrecht: I added more tests around the generated key that also negatively tests that the key is updated with different architectures, supported architectures, and environments.

I didn't add the test that asserts the same key with the same environments in a different order based on the following other comment. It looks like we have to use the environment priority ordering that is returned from the curator.

https://github.com/candlepin/candlepin/pull/5024#discussion_r1956207437

joshmalbrecht: Ah, yeah! I will update this and get some tests to assert this logic

joshmalbrecht: I added `distinct()` to `ContentAccessPayloadKeyBuilder.getNormalizedEnvironments` and added a unit test to assert these changes

nikosmoum (approved):

PR #5032, opened 2025-02-17
[F] CHAINSAW-375: Refactored exceptions

joshmalbrecht:
- Refactored the ServiceUnavailableException and the TooManyRequestsException to have the standard exception constructors.
- Added the ability to define a 'Retry-After' header for the ServiceUnavailableException.
- Added exception messages for TooManyRequestsExceptions in SecurityHoleAuthorizationFilter.

`ServiceUnavailableException` and `TooManyRequestsException` extend a `CandlepinException` and a `CandlepinException` extends a RuntimeException. This means that the headers for both the `ServiceUnavailableException` and `TooManyRequestsException` are added by the `RuntimeExceptionMapper`.

https://github.com/candlepin/candlepin/blob/main/src/main/java/org/candlepin/exceptions/mappers/RuntimeExceptionMapper.java#L67-L69

### DIFF
--- a/src/main/java/org/candlepin/exceptions/ServiceUnavailableException.java
+++ b/src/main/java/org/candlepin/exceptions/ServiceUnavailableException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ * Copyright (c) 2009 - 2025 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -14,19 +14,73 @@
  */
 package org.candlepin.exceptions;
 
+import java.util.Map;
+
 import javax.ws.rs.core.Response.Status;
 
 
 /**
- * ServiceUnavailableException
- * Represents a Service Unavailable (HTTP 503) error.
+ * ServiceUnavailableException represents an exception that generates a 503 HTTP response status code.
  */
 public class ServiceUnavailableException extends CandlepinException {
+    private static final String RETRY_AFTER_HEADER_KEY = "Retry-After";
+
+    private Integer retryAfter;
+
+    /**
+     * Creates a new service unavailable exception with the provided message.
+     *
+     * @param message
+     *  the message used to create a new service unavailable exception
+     */
     public ServiceUnavailableException(String message) {
         super(Status.SERVICE_UNAVAILABLE, message);
     }
 
+    /**
+     * Creates a service unavailable exception with the provided message and provided throwable cause.
+     *
+     * @param message
+     *  the message used to create a new service unavailable exception
+     *
+     * @param cause
+     * the throwable cause used to create a new service unavailable exception
+     */
     public ServiceUnavailableException(String message, Throwable cause) {
         super(Status.SERVICE_UNAVAILABLE, message, cause);
     }
+
+    /**
+     * Sets the retry after time value used to populate the 'Retry-After' header. If a null value is provided,
+     * the 'Retry-After' header will not be populated in the response. Subsequent calls will overwrite the
+     * previously set value.
+     *
+     * @param retryAfter
+     *  the value to set for the 'Retry-After' header
+     *
+     * @return this too many requests exception instance
+     */
+    public ServiceUnavailableException setRetryAfterTime(Integer retryAfter) {
+        this.retryAfter = retryAfter;
+        return this;
+    }
+
+    /**
+     * @return the value that will be set for the 'Retry-After' header. A null value indicates that the header
+     *  will not be set
+     */
+    public Integer getRetryAfterTime() {
+        return this.retryAfter;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Map<String, String> headers() {
+        return this.retryAfter == null ?
+            Map.of() :
+            Map.of(RETRY_AFTER_HEADER_KEY, String.valueOf(this.retryAfter));
+    }
+
 }
--- a/src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
+++ b/src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ * Copyright (c) 2009 - 2025 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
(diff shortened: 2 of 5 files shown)


joshmalbrecht:
It looks like SonarCube is upset at the duplicate lines between `ServiceUnavailableException` and `TooManyRequestsException`. I think we can ignore this failure as it is failing on the getter, setter, and headers method which will be identical for handling the retry-after header value.

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
```diff
@@ -132,10 +134,14 @@ private String checkCloudAccountOrganization(AnonymousCloudConsumer anonymousClo
                 anonymousCloudConsumer.getCloudOfferingId());
         }
         catch (OrgForCloudAccountNotCreatedYetException e) {
-            throw new TooManyRequestsException(ORG_DOES_NOT_EXIST_AT_ALL);
+            String msg = "Upstream anonymous owner not yet created by adapters";
+            throw new TooManyRequestsException(i18nProvider.get().tr(msg))
```
nikosmoum, line 138: You cannot use `tr()` with a variable; it only works with a hardcoded string

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
```diff
@@ -132,10 +134,14 @@ private String checkCloudAccountOrganization(AnonymousCloudConsumer anonymousClo
...
+                .setRetryAfterTime(ORG_DOES_NOT_EXIST_AT_ALL);
         }
         catch (OrgForCloudAccountNotEntitledYetException e) {
-            throw new TooManyRequestsException(ORG_IS_NOT_ENTITLED);
+            String msg = "Upstream anonymous owner not yet entitled by adapters";
+            throw new TooManyRequestsException(i18nProvider.get().tr(msg))
```
nikosmoum, line 143: Same, you need to pass in the hardcoded string or else it won't get translated. The way it works is gettext scans the actual code for instances of tr(), and finds the strings passed to it as keys to translate. if a variable is passed it won't work

nikosmoum (changes requested):

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
```diff
@@ -116,10 +116,12 @@ private void validateAnonymousCloudConsumer(AnonymousCloudConsumerPrincipal prin
         AnonymousCloudConsumer anonymousCloudConsumer = principal.getAnonymousCloudConsumer();
         String ownerKey = checkCloudAccountOrganization(anonymousCloudConsumer);
         if (!ownerCurator.existsByKey(ownerKey)) {
-            throw new TooManyRequestsException(ORG_NOT_CREATED_IN_CANDLEPIN);
+            throw new TooManyRequestsException(i18nProvider.get().tr("Anonymous owner does not yet exist"))
+                .setRetryAfterTime(ORG_NOT_CREATED_IN_CANDLEPIN);
```
nikosmoum, line 120: This specific case can happen with non-anonymous orgs as well (in the case a proper org exists, but hasn't been created in Candlepin yet).

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
```diff
@@ -116,10 +116,12 @@ private void validateAnonymousCloudConsumer(AnonymousCloudConsumerPrincipal prin
...
         }
         if (!poolCurator.hasPoolsForProducts(ownerKey, anonymousCloudConsumer.getProductIds())) {
-            throw new TooManyRequestsException(ORG_DOES_NOT_HAVE_POOLS);
+            throw new TooManyRequestsException(i18nProvider.get().tr("Anonymous owner is not yet entitled"))
```
nikosmoum, line 123: Same here, this can happen with non-anonymous owners

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
```diff
@@ -132,10 +134,14 @@ private String checkCloudAccountOrganization(AnonymousCloudConsumer anonymousClo
                 anonymousCloudConsumer.getCloudOfferingId());
         }
         catch (OrgForCloudAccountNotCreatedYetException e) {
-            throw new TooManyRequestsException(ORG_DOES_NOT_EXIST_AT_ALL);
+            String msg = "Upstream anonymous owner not yet created by adapters";
```
nikosmoum, line 137: I don't think we should be mentioning the word `adapters` to API responses. Something better may be `Anonymous owner not yet created upstream of candlepin`

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
```diff
@@ -132,10 +134,14 @@ private String checkCloudAccountOrganization(AnonymousCloudConsumer anonymousClo
...
+            throw new TooManyRequestsException(i18nProvider.get().tr(msg))
+                .setRetryAfterTime(ORG_DOES_NOT_EXIST_AT_ALL);
         }
         catch (OrgForCloudAccountNotEntitledYetException e) {
-            throw new TooManyRequestsException(ORG_IS_NOT_ENTITLED);
+            String msg = "Upstream anonymous owner not yet entitled by adapters";
```
nikosmoum, line 142: Similar to the other one, this is probably better: `Anonymous owner not yet entitled upstream of candlepin`

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
joshmalbrecht: Updated

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
joshmalbrecht: Ah yeah, I forget that is a thing. Updated

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
joshmalbrecht: Removed "Anonymous" from the message

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
joshmalbrecht: Yeah, I agree! That message is much better. Updated

src/main/java/org/candlepin/resteasy/filter/SecurityHoleAuthorizationFilter.java
Ceiu: As a note for when/if this comes up in the future: if we need to have a string that will eventually be translated as a variable (as was done in the original code here), you can have a string primed for translation via `i18n.marktr(<string>)`. That will allow that specific string to be used with `.tr(...)` later without needing to use a string literal. For example:

```
I18n i18n = this.i18nProvider.get();
String msg = i18n.marktr("Some string we want translated");

...

throw new CandlepinException(i18n.tr(msg));
```

joshmalbrecht: That is good to know!

Ceiu:
If we're interested in appeasing SonarCloud, we could turn this into a hierarchy of exceptions like RetryAfterException -> TooManyRequestsException, ServiceUnavailableException; but I feel like that's a misuse of subclassing as the retry behavior is closer to a trait than a hierarchical property.

joshmalbrecht:
> Do we have any spec tests which verify the retry-after header is being set? I have concerns that it won't be present if the wrong mapper is picking it up.

Yes, we have spec tests in `CloudRegistrationSpecTest` that assert that the header exists and is correct. These tests are for the Auto Registration V2 workflow and a specific example is `CloudRegistrationSpecTest.shouldCreateConsumerWhenUsingAnonymousToken`.

src/main/java/org/candlepin/exceptions/ServiceUnavailableException.java
```diff
@@ -14,19 +14,79 @@
...
+    public Integer getRetryAfterTime() {
+        return this.retryAfter;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Map<String, String> headers() {
+        HashMap<String, String> headers = new HashMap<>();
```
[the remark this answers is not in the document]
joshmalbrecht: Updated in both exception classes

Ceiu (approved):

PR #5036, opened 2025-02-19
[F] CHAINSAW-361: Removed obsoleted methods from ContentAccessManager

joshmalbrecht:
- Removed the 'hasCertChangesSince' and 'syncOwnerLastContentUpdate' methods from the ContentAccessManager class.

### DIFF
--- a/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -205,6 +205,9 @@ public class ConsumerResource implements ConsumerApi {
 
     private static final Logger log = LoggerFactory.getLogger(ConsumerResource.class);
 
+    private static final DateTimeFormatter SINCE_DATE_FORMATER = DateTimeFormatter
+        .ofPattern("EEE, dd MMM yyyy HH:mm:ss z", Locale.US);
+
     private final ConsumerCurator consumerCurator;
     private final ConsumerTypeCurator consumerTypeCurator;
     private final SubscriptionServiceAdapter subAdapter;
@@ -2372,27 +2375,13 @@ public class ConsumerResource implements ConsumerApi {
 
     @Override
     @Transactional
-    public Response getContentAccessBody(@Verify(Consumer.class) String consumerUuid, String sinceDate) {
+    public Response getContentAccessBody(@Verify(Consumer.class) String consumerUuid, String since) {
         log.debug("Getting content access certificate for consumer: {}", consumerUuid);
         Consumer consumer = consumerCurator.verifyAndLookupConsumer(consumerUuid);
-        OffsetDateTime since = null;
 
         Owner owner = ownerCurator.findOwnerById(consumer.getOwnerId());
         if (!owner.isUsingSimpleContentAccess()) {
             throw new BadRequestException(i18n.tr("Content access mode does not allow this request."));
-        }
-
-        if (sinceDate != null) {
-            DateTimeFormatter formatter = DateTimeFormatter
-                .ofPattern("EEE, dd MMM yyyy HH:mm:ss z", Locale.US);
-            since = Util.parseOffsetDateTime(formatter, sinceDate);
-        }
-
-        Date date = since != null ? Util.toDate(since) : new Date(0);
-        if (!this.contentAccessManager.hasCertChangedSince(consumer, date)) {
-            return Response.status(Response.Status.NOT_MODIFIED)
-                .entity("Not modified since date supplied.")
-                .build();
         }
 
         SCACertificate cac = null;
@@ -2405,6 +2394,23 @@ public class ConsumerResource implements ConsumerApi {
 
         if (cac == null) {
             throw new BadRequestException(i18n.tr("Cannot retrieve content access certificate"));
+        }
+
+        // The date provided by the client does not preserve milliseconds, so we need to round down
+        // the dates preserved on the server by removing milliseconds for a proper date comparison
+        Date scaCertificateLastUpdated = cac.getUpdated() != null ?
+            Util.roundDownToSeconds(cac.getUpdated()) :
+            new Date();
+
+        OffsetDateTime sinceOffsetDateTime = since != null ?
+            Util.parseOffsetDateTime(SINCE_DATE_FORMATER, since) :
+            null;
+
+        Date sinceDate = sinceOffsetDateTime != null ? Util.toDate(sinceOffsetDateTime) : new Date(0);
+        if (sinceDate.after(scaCertificateLastUpdated) || sinceDate.equals(scaCertificateLastUpdated)) {
+            return Response.status(Response.Status.NOT_MODIFIED)
+                .entity("Not modified since date supplied.")
+                .build();
         }
 
         String cert = cac.getCert();
--- a/src/test/java/org/candlepin/controller/ContentAccessManagerTest.java
+++ b/src/test/java/org/candlepin/controller/ContentAccessManagerTest.java
@@ -15,12 +15,11 @@
 package org.candlepin.controller;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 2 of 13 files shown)


src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2407,6 +2396,18 @@ public Response getContentAccessBody(@Verify(Consumer.class) String consumerUuid
             throw new BadRequestException(i18n.tr("Cannot retrieve content access certificate"));
         }
 
+        OffsetDateTime sinceOffsetDateTime = since != null ?
+            Util.parseOffsetDateTime(SINCE_DATE_FORMATER, since) :
+            null;
+
+        Date sinceDate = sinceOffsetDateTime != null ? Util.toDate(sinceOffsetDateTime) : new Date(0);
+        Date scaCertificateLastUpdated = cac.getUpdated() != null ? cac.getUpdated() : new Date();
```
joshmalbrecht, line 2404: Note that this will change when we retrieve the individual components as part of CHAINSAW-359 instead of using the `scaCertificateGenerator.generate` method.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2407,6 +2396,18 @@ public Response getContentAccessBody(@Verify(Consumer.class) String consumerUuid
...
+        if (sinceDate.after(scaCertificateLastUpdated)) {
+            return Response.status(Response.Status.NOT_MODIFIED)
+                .entity("Not modified since date supplied.")
+                .build();
```
nikosmoum, line 2408: Why was this moved until after the .generate() call? Wasn't the whole point to avoid going into the generate() method if the cert does not need change?

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -797,7 +794,8 @@ public void shouldReturnANotModifiedReturnCodeWhenTheDataHasNotBeenUpdatedSinceD
 
         ZonedDateTime inputZonedDateTime = ZonedDateTime
             .parse(lastUpdate, DateTimeFormatter.ofPattern(CONTENT_ACCESS_OUTPUT_DATE_FORMAT));
-        OffsetDateTime since = inputZonedDateTime.toOffsetDateTime();
+        OffsetDateTime since = inputZonedDateTime.toOffsetDateTime()
+            .plusHours(1);
```
nikosmoum, line 798: We shouldn't have to do spec test changes for any of these changes, since it all should be internal refactoring. I just run this test in its original state, against the prod changes in this PR, and it failed, which means our API output changed

src/main/java/org/candlepin/resource/ConsumerResource.java
joshmalbrecht: This was part of the changes in the "ConsumerResource Optimizations" section and under the "Accessible Content" subsection in the design document.  I left the the `.generate()` method untouched here as CHAINSAW-359 will update this method to use the retrieve the individual components instead of the entire assembled certificate.

> The current implementation first calls a utility method to check if the target timestamp – that is, the timestamp provided by the user, or the Unix epoch if the user did not provide a timestamp – is "newer" than those used to determine if a regeneration is necessary. This is performed by the now-obsolete ContentAccessManager.hasCertChangedSince method
> 
> While this seems like a reasonable optimization to avoid expensive generation at first glance, this actually doubles the amount of checks against the timestamp performed by this operation, and doesn't actually avoid any work that would be done in either case. We can, instead, fetch the SCA certificate, or its constituent components, and perform the same check at that time to determine which output the user receives.
> 
> Again, on the surface this seems inefficient as we'll trigger a regeneration of either component if either are outdated or missing; but in the general case we'll be doing this anyway – either the target timestamp is in the past and our current version of the SCA certificate is valid and fetching it is a simple lookup, or the SCA certificate is outdated relative to the target timestamp and would need to be regenerated.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
joshmalbrecht: Yeah, you are right! Taking a look at this spec test, the `ifModifiedSince` value that is provided to the second `getContentAccessBody` is the last update value of the content access body. Since no content or environment changes occurred, the `sinceDate` ends up being the same date as the `scaCertificateLastUpdated` in `ConsumerResource.getContentAccessBody`. We were only returning the 304 response when `sinceDate.after(scaCertificateLastUpdated)`, but I believe that we should also return 304 when `sinceDate.equals(scaCertificateLastUpdated))`. Updated with these changes.

Also, I noticed the following comment in the removed `ContentAccessManager.hasCertChangedSince` method and added the use of `roundDownToSeconds` in case it is important. Based on the comment it looks like it is.

https://github.com/candlepin/candlepin/blob/main/src/main/java/org/candlepin/controller/ContentAccessManager.java#L147-L149

Ceiu (changes requested):

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2407,6 +2396,23 @@ public Response getContentAccessBody(@Verify(Consumer.class) String consumerUuid
             throw new BadRequestException(i18n.tr("Cannot retrieve content access certificate"));
         }
 
+        // The date provided by the client does not preserve milliseconds, so we need to round down
+        // the dates preserved on the server by removing milliseconds for a proper date comparison
+        Date scaCertificateLastUpdated = cac.getUpdated() != null ?
+            Util.roundDownToSeconds(cac.getUpdated()) :
```
Ceiu, line 2402: Does this work as a stand-in for the format used by HTTP headers (and the formatter removed above)? IIRC, the formats are slightly different from typical time-related ISO/RFC formats.

src/test/java/org/candlepin/controller/ContentAccessManagerTest.java
```diff
@@ -512,6 +511,7 @@ public void testUpdateOwnerContentAccessModeChangedFromEntModeToSCA() throws Job
         assertEquals(owner.getContentAccessModeList(), contentAccessModeList);
         assertEquals(owner.getContentAccessMode(), contentAccessMode);
         verify(jobManager).queueJob(any(RevokeEntitlementsJobConfig.class));
+        verify(owner).syncLastContentUpdate();
```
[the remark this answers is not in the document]
joshmalbrecht: I used a spy here because the unit test can run so fast that the Dates can be the same because Dates only have millisecond precision. I don't run into this on my machine, but the test can fail in the Github runner. We could add a sleep in the test to ensure the Date difference, but I wanted to avoid that if possible.

Failed GH actions job where I was comparing the dates directly:
https://github.com/candlepin/candlepin/actions/runs/13421268490/job/37497939746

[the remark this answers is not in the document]
Ceiu: Fair enough, but we're not interested in checking that a specific method or function is called; we want to know if certain aspects of the org's state has changed. If we make changes in the future to call `.setLastContentUpdate` instead of `syncLastContentUpdate`, this test breaks even if the behavior doesn't.

I'd much rather see an appropriate sleep here as long as we're tied to the old Date objects. If we ever get around to refactoring all of our dates to use stuff from `java.time`, we'll be able to drop many/most/all? of those.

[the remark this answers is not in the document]
joshmalbrecht: Yeah, I think this problem goes away with the use of `java.time` because they have nanosecond precision.

Removed the use of spies, added some small sleeps, and compared the last content update dates

nikosmoum (approved):
Looks good to me. I'll let @Ceiu re-review/merge

Ceiu (approved):

PR #5037, opened 2025-02-20
[M] CHAINSAW-382: Updated dev container image to output Candlepin logs

joshmalbrecht:
- Added a ConsoleAppender to the logback.xml for the dev container image so that the Candlepin logs are output to stdout.
- Added a ConsoleAppender to the logback.xml for the dev container image to log the org.candlepin.audit.LoggingListener.AuditLog logs to stdout.
- Removed the host port for the Postgres container in dev-container/docker-compose.yml and dev-container/candlepin-deployment.yml so that the OS will choose the host port number in the port mapping.
- Updated some of the container documentation in CONTRIBUTING.md.

## Testing

To test and verify the changes we can build a new development image and run the container using `podman-compose`, `docker-compose`, or `podman play kube`.

### 1. Build the WAR file

Run the following command to build the WAR file that will be used in our new image. The command includes the extensions used for testing.

`./gradlew war -Ptest_extensions=hostedtest,manifestgen`
 
### 2. Build the container image

Run the following command from the root directory of the Candlepin project to build your image. Note that you will need to insert your image name and tag, and adjust the WAR file version if the version you built is not 4.6.0.

```
podman build -f ./containers/release.Containerfile -t <insert-name:insert-tag> . \
	--target development \
	--build-arg WAR_FILE=./build/libs/candlepin-4.6.0.war \
	--no-cache
```

### 3. Update the compose files to use your new image

Depending on which compose you will run, you will need to update the `image` value to be your newly built image in `dev-container/candlepin-deployment.yaml` and or `dev-container/docker-compose.yml`.

### 4. Start the containers

From the `dev-container` directory, run either `podman-compose up -d`, `docker-compose up -d`, or `podman play kube candlepin-deployment.yaml` to run your containers.

### 5. Run spec tests

Once the Candlepin container is running and started, you can run the spec tests to generate some log output.

`./gradlew spec`

### 6. View the logs

Now we should see the Candlepin logs by using `podman logs -f <container ID>`.

### DIFF
--- a/containers/logback.xml
+++ b/containers/logback.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <contextName>candlepin</contextName>
+
+    <turboFilter class="org.candlepin.logging.LoggerAndMDCFilter">
+        <key>logLevel</key>
+        <topLogger>org.candlepin</topLogger>
+        <OnMatch>ACCEPT</OnMatch>
+    </turboFilter>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{ISO8601} [thread=%thread] [%X{requestType}=%X{requestUuid}, %replace(job_key=%X{jobKey}, ){'job_key=, ', ''}org=%X{org}, csid=%X{csid}] %-5p %c - %m%n</pattern>
+        </encoder>
+    </appender>
+
+    <appender name="ASYNC" class="ch.qos.logback.classic.AsyncAppender">
+        <appender-ref ref="STDOUT"/>
+    </appender>
+
+    <appender name="AUDITLOG" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{ISO8601} %m</pattern>
+        </encoder>
+    </appender>
+
+    <appender name="ASYNC-AUDITLOG" class="ch.qos.logback.classic.AsyncAppender">
+        <appender-ref ref="AUDITLOG"/>
+    </appender>
+
+    <logger name="org.candlepin" level="INFO"/>
+    <logger name="org.candlepin.audit.LoggingListener.AuditLog" level="INFO" additivity="false">
+        <appender-ref ref="ASYNC-AUDITLOG"/>
+    </logger>
+
+    <!-- Silence the hibernate deprecation warnings -->
+    <logger name="org.hibernate.orm.deprecation" level="DEBUG" additivity="false"/>
+
+    <root level="WARN">
+        <appender-ref ref="ASYNC"/>
+    </root>
+</configuration>
--- a/containers/release.Containerfile
+++ b/containers/release.Containerfile
@@ -15,9 +15,11 @@ RUN wget https://archive.apache.org/dist/tomcat/tomcat-9/v9.0.76/bin/apache-tomc
     mv apache-tomcat-9.0.76/* /opt/tomcat/
 
 # Prepare Candlepin
-RUN mkdir -p /app/build
+RUN mkdir -p /app/build/candlepin
 WORKDIR /app/build
 COPY ${WAR_FILE} ./candlepin.war
+WORKDIR /app/build/candlepin
+RUN jar xvf /app/build/candlepin.war
 
 # Prepare development certs
 RUN mkdir -p /app/certs
@@ -89,7 +91,7 @@ FROM production as development
 LABEL org.opencontainers.image.title="Candlepin Development Image" \
     org.opencontainers.image.description="Candlepin is an open source subscription & entitlement engine \
 which is designed to manage software subscriptions from both vendor's & customer's perspectives. \
-    This is a development image not intended for production use."
+This is a development image and not intended for production use."
 
 USER root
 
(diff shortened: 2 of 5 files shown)


containers/logback.xml
```diff
@@ -0,0 +1,77 @@
...
+    <logger name="org.candlepin.audit.LoggingListener.AuditLog" level="INFO" additivity="false">
+        <appender-ref ref="AUDITLOG" />
+    </logger>
+
+    <!-- Silence the hibernate deprecation warnings -->
+    <logger name="org.hibernate.orm.deprecation" level="DEBUG" additivity="false"/>
+
+    <root level="WARN">
+        <appender-ref ref="CandlepinAppender" />
+        <appender-ref ref="ErrorAppender" />
```
joshmalbrecht, line 74: Do we want to still use the CandlepinAppender, ErrorAppender, and AUDITLOG appenders in the container if we are outputting to stdout?

containers/release.Containerfile
```diff
@@ -15,9 +15,12 @@ RUN wget https://archive.apache.org/dist/tomcat/tomcat-9/v9.0.76/bin/apache-tomc
     mv apache-tomcat-9.0.76/* /opt/tomcat/
 
 # Prepare Candlepin
-RUN mkdir -p /app/build
+RUN mkdir -p /app/build/candlepin
 WORKDIR /app/build
 COPY ${WAR_FILE} ./candlepin.war
+WORKDIR /app/build/candlepin
+RUN jar xvf /app/build/candlepin.war
+COPY ./containers/logback.xml /app/build/candlepin/WEB-INF/classes/logback.xml
```
nikosmoum, line 23: Based on https://logback.qos.ch/manual/appenders.html#ConsoleAppender the console is very slow, so we should only do this in the dev version of the image, not in prod. 

@ehelms do you agree? You'll be copying your own version of logback.xml anyway right?

containers/logback.xml
nikosmoum: If we're making the ConsoleAppender a dev-only option, then we should remove the rest from here. While the prod version would only have the log file appenders

nikosmoum (changes requested):

containers/release.Containerfile
joshmalbrecht: So, I was exploring some alternatives to try and avoid the performance degradation with using the Logback's ConsoleAppender, and it looks like we can keep the logback configurations as they currently are and create a symbolic link from the `/var/log/candlepin/candlepin.log` file to `/proc/1/fd/1`. This will output the data from `/var/log/candlepin/candlepin.log` to STDOUT without having to use the `ConsoleAppender`. I added these changes in a new commit to see what you think.

nikosmoum: If this way is faster, then sure let's do it. However we don't know what 'slow' means for ConsoleAppender, and we don't know how fast/slow the symlink solution is compared to it. How do we even measure it?

nikosmoum: Apparently you could have ConsoleAppender be asynchronous, like this: https://stackoverflow.com/questions/29523507/will-disabling-logback-console-appender-increase-performance-if-i-log-to-file-an but again, I don't know what the perf comparison would look like.

joshmalbrecht: Let me see if there are some ways of measuring the performance of these approaches and so we can get some concrete numbers to compare

joshmalbrecht: I did some profiling to get some numbers on the performance impact of using the `ConsoleAppender` and the `ConsoleAppender` with the `AsyncAppender` .  I also profiled the current container image that does not use the `ConsoleAppender` to get a baseline. The following are the invocation duration for `org.slf4j.Logger.info` for the same endpoint (GET /consumers).

| No Changes (main branch) | ConsoleAppender | ConsoleAppender and AsyncAppender |
| -------- | ------- | ------- |
| 51 microseconds | 414 microseconds | 256 microseconds |

**Percentage increase from the baseline:**

| ConsoleAppender | ConsoleAppender and AsyncAppender |
| -------- | ------- |
| 711.8% | 402% |

So, it does look like there is a relative significant increase in invocation time for both logback.xml configurations, but this is in the scale of microseconds.

As for testing the symlink approach performance, I have no idea on how to test the performance of this change. I did build an image with the symlink changes and ran the spec tests against the container and compared the time it took with the time it takes for main branch to run, and they were close to the same. If anyone has any other ideas on how to benchmark the symlink changes, I can give it a try! 

I would think that the symlink approach would be faster than both the `ConsoleAppender` and the `ConsoleAppender` and `AsyncAppender` approach, but I don't know how to measure and verify it.

joshmalbrecht: Sounds good to me! Do we want to apply this to just the dev container or both the dev container and the production image?

joshmalbrecht: Updated to add the AsyncAppender+ConsoleAppender to just the dev container

containers/logback.xml
joshmalbrecht: Updated and removed the unused appenders

nikosmoum (changes requested):
Looks good generally, we just need to add the auditlog to stdout

containers/logback.xml
```diff
@@ -0,0 +1,29 @@
...
+        <appender-ref ref="STDOUT" />
+    </appender>
+
+    <logger name="org.candlepin" level="INFO"/>
+
+    <!-- Silence the hibernate deprecation warnings -->
+    <logger name="org.hibernate.orm.deprecation" level="DEBUG" additivity="false"/>
+
+    <root level="WARN">
+        <appender-ref ref="ASYNC" />
```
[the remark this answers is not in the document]
joshmalbrecht: Updated. With the changes, here is an example of a log from the AUDITLOG appender that is logged out to stdout:

```
2025-03-26 14:38:02,906 principalType=user principal=admin target=SYSTEM_PURPOSE_COMPLIANCE entityId=bdfa1bfc-32c6-4201-9e8e-5ee9dddb926c type=CREATED owner=test_owner-FD6A883E eventData={nonCompliantUsage=null, compliantAddOns={}, nonCompliantRole=null, reasons=[], nonCompliantServiceType=null, compliantSLA={}, nonCompliantAddOns=[], compliantRole={}, nonCompliantSLA=null, compliantUsage={}, status=not specified, compliantServiceType={}}
```

nikosmoum (approved):

PR #5045, opened 2025-02-26
[F] CHAINSAW-359: Updated ConsumerResource.getContentAccessBody

joshmalbrecht:
- Updated ConsumerResource.getContentAccessBody to fetch the individual components from the SCACertificateGenerator instead of retrieving the entire assembled certificate.
- Updated ConsumerResource.getContentAccessBody to return a 429 response if there is a constraint violation when persisting the SCA certificate's content access payload.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
@@ -26,6 +26,7 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import org.candlepin.dto.api.client.v1.AsyncJobStatusDTO;
 import org.candlepin.dto.api.client.v1.CertificateDTO;
@@ -41,7 +42,9 @@ import org.candlepin.dto.api.client.v1.PoolDTO;
 import org.candlepin.dto.api.client.v1.ProductDTO;
 import org.candlepin.dto.api.client.v1.ReleaseVerDTO;
 import org.candlepin.dto.api.client.v1.SystemPurposeComplianceStatusDTO;
+import org.candlepin.invoker.client.ApiException;
 import org.candlepin.resource.client.v1.ConsumerApi;
+import org.candlepin.spec.bootstrap.assertions.CandlepinStatusAssert;
 import org.candlepin.spec.bootstrap.client.ApiClient;
 import org.candlepin.spec.bootstrap.client.ApiClients;
 import org.candlepin.spec.bootstrap.client.SpecTest;
@@ -52,6 +55,7 @@ import org.candlepin.spec.bootstrap.data.builder.ConsumerTypes;
 import org.candlepin.spec.bootstrap.data.builder.Consumers;
 import org.candlepin.spec.bootstrap.data.builder.Contents;
 import org.candlepin.spec.bootstrap.data.builder.Environments;
+import org.candlepin.spec.bootstrap.data.builder.Facts;
 import org.candlepin.spec.bootstrap.data.builder.Owners;
 import org.candlepin.spec.bootstrap.data.builder.Pools;
 import org.candlepin.spec.bootstrap.data.builder.Products;
@@ -72,12 +76,20 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 @SpecTest
 @SuppressWarnings("indentation")
 public class ContentAccessSpecTest {
     private static final String CONTENT_ACCESS_OUTPUT_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ssZ";
     private static final String CONTENT_ACCESS_INPUT_DATE_FORMAT = "EEE, dd MMM yyyy HH:mm:ss z";
+    private static final int CONTENT_PAYLOAD_CREATION_EXCEPTION_RETRY_AFTER_TIME = 2;
+    private static final String RETRY_AFTER_HEADER_KEY = "retry-after";
 
     @Test
     public void shouldFilterContentWithMismatchedArchitecture() {
@@ -803,6 +815,68 @@ public class ContentAccessSpecTest {
 
         assertNotModified(() -> adminClient.consumers()
             .getContentAccessBodyJson(consumer.getUuid(), formatted));
+    }
+
+    @Test
+    public void shouldReturnServiceUnavailableWithConcurrentPayloadCreation() throws Exception {
+        ApiClient adminClient = ApiClients.admin();
+        OwnerDTO owner = adminClient.owners().createOwner(Owners.randomSca());
+        String ownerKey = owner.getKey();
+
+        ProductDTO prod = adminClient.ownerProducts().createProduct(ownerKey, Products.random());
+        ContentDTO content = adminClient.ownerContent().createContent(ownerKey, Contents.random());
+        adminClient.ownerProducts()
+            .addContentToProduct(ownerKey, prod.getId(), content.getId(), true);
+        adminClient.owners().createPool(ownerKey, Pools.random(prod));
+
+        ConsumerDTO consumer1 = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ConsumerDTO consumer2 = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ConsumerDTO consumer3 = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ConsumerDTO consumer4 = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ConsumerDTO consumer5 = adminClient.consumers().createConsumer(Consumers.random(owner));
+
+        List<ConsumerDTO> consumers = List.of(consumer1, consumer2, consumer3, consumer4, consumer5);
+
+        int maxNumberOfAttempts = 5;
+        ApiException exception = assertThrows(ApiException.class, () -> {
+            for (int i = 0; i < maxNumberOfAttempts; i++) {
+                // Update the arch of all the consumers so that the content access payload key will be
+                // different than the key used to store the content access payload previously. This causes the
+                // regeneration of the content access payload which is what we need to cause the 503 response.
+                String newArch = StringUtil.random("arch-");
+                for (ConsumerDTO consumer : consumers) {
+                    consumer.putFactsItem(Facts.Arch.key(), newArch);
+                    adminClient.consumers().updateConsumer(consumer.getUuid(), consumer);
+                }
+
+                List<Callable<String>> tasks = List.of(
+                    () -> adminClient.consumers().getContentAccessBody(consumer1.getUuid(), null),
+                    () -> adminClient.consumers().getContentAccessBody(consumer2.getUuid(), null),
+                    () -> adminClient.consumers().getContentAccessBody(consumer3.getUuid(), null),
+                    () -> adminClient.consumers().getContentAccessBody(consumer4.getUuid(), null),
+                    () -> adminClient.consumers().getContentAccessBody(consumer5.getUuid(), null));
+
+                ExecutorService execService = Executors.newFixedThreadPool(consumers.size());
+                List<Future<String>> futures = execService.invokeAll(tasks);
+                execService.shutdown();
+                execService.awaitTermination(10L, TimeUnit.SECONDS);
+
+                for (Future<String> future : futures) {
+                    try {
+                        future.get();
+                    }
+                    catch (ExecutionException e) {
+                        throw e.getCause();
+                    }
+                }
+            }
+        });
+
+        new CandlepinStatusAssert(exception)
+            .isTooMany()
+            .hasHeaderWithValue(RETRY_AFTER_HEADER_KEY,
+                CONTENT_PAYLOAD_CREATION_EXCEPTION_RETRY_AFTER_TIME);
+
     }
 
     @Test
--- a/src/main/java/org/candlepin/resource/ConsumerResource.java
+++ b/src/main/java/org/candlepin/resource/ConsumerResource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ * Copyright (c) 2009 - 2025 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -76,6 +76,7 @@ import org.candlepin.exceptions.ForbiddenException;
 import org.candlepin.exceptions.GoneException;
 import org.candlepin.exceptions.IseException;
 import org.candlepin.exceptions.NotFoundException;
+import org.candlepin.exceptions.TooManyRequestsException;
 import org.candlepin.guice.PrincipalProvider;
 import org.candlepin.model.AnonymousCloudConsumer;
 import org.candlepin.model.AnonymousCloudConsumerCurator;
@@ -96,6 +97,7 @@ import org.candlepin.model.ConsumerInstalledProduct;
 import org.candlepin.model.ConsumerType;
 import org.candlepin.model.ConsumerType.ConsumerTypeEnum;
 import org.candlepin.model.ConsumerTypeCurator;
+import org.candlepin.model.ContentAccessPayload;
 import org.candlepin.model.ContentOverride;
 import org.candlepin.model.DeletedConsumer;
 import org.candlepin.model.DeletedConsumerCurator;
@@ -169,7 +171,9 @@ import org.xnap.commons.i18n.I18n;
 
 import java.io.File;
 import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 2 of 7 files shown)


spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -805,6 +812,85 @@ public void shouldReturnANotModifiedReturnCodeWhenTheDataHasNotBeenUpdatedSinceD
...
+
+        ConsumerDTO consumer1 = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ConsumerDTO consumer2 = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ConsumerDTO consumer3 = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ConsumerDTO consumer4 = adminClient.consumers().createConsumer(Consumers.random(owner));
+        ConsumerDTO consumer5 = adminClient.consumers().createConsumer(Consumers.random(owner));
+
+        List<ConsumerDTO> consumers = List.of(consumer1, consumer2, consumer3, consumer4, consumer5);
+
+        int maxNumberOfAttempts = 5;
```
joshmalbrecht, line 835: This test usually creates the `PersistenceException` after 1 or 2 attempts for me locally, so I bumped the number of attempts up to 5 so that the test should consistently reproduce the exception.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2384,45 +2386,49 @@ public Response getContentAccessBody(@Verify(Consumer.class) String consumerUuid
...
-        SCACertificate cac = null;
+        SCACertificate x509Cert = scaCertificateGenerator.getX509Certificate(consumer);
+        ContentAccessPayload contentPayload = null;
         try {
-            cac = this.scaCertificateGenerator.generate(consumer);
+            contentPayload = scaCertificateGenerator.getContentPayload(consumer);
         }
         catch (ConcurrentContentPayloadCreationException e) {
-            // TODO: Handle as part of CHAINSAW-359
+            throw new ServiceUnavailableException(i18n.tr("Unable to create content access payload"), e);
```
nikosmoum, line 2395: I thought the client is responding well for 429 status? Shouldn't we be throwing that instead of a 503?
Also, shouldn't we be setting the retry-after value?

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2384,45 +2388,50 @@ public Response getContentAccessBody(@Verify(Consumer.class) String consumerUuid
...
+            x509CertUpdated :
+            contentPayloadUpdated;
+
         OffsetDateTime sinceOffsetDateTime = since != null ?
             Util.parseOffsetDateTime(SINCE_DATE_FORMATER, since) :
             null;
 
         Date sinceDate = sinceOffsetDateTime != null ? Util.toDate(sinceOffsetDateTime) : new Date(0);
-        if (sinceDate.after(scaCertificateLastUpdated) || sinceDate.equals(scaCertificateLastUpdated)) {
+        if (sinceDate.after(lastUpdated) || sinceDate.equals(lastUpdated)) {
```
Ceiu, line 2424: This block should be optimized a bit. If we want "after or equal", rather than doing the check twice, we can express that as "not before". Also, we don't need to create a new date just to do the check if we didn't specify a date. The logic here intends to express: "if the user provided a timestamp, and that timestamp is after the last update of both components, return NOT_MODIFIED"

To that end, we could even compress the "lastUpdated" into this as well and combine everything into one block:

```
try {
    // Fetch the SCA certificate components
    SCACertificate scaCert = this.scaCertificateGenerator.getSCAX509Certificate(consumer);
    ContentAccessPayload scaPayload = this.scaCertificateGenerator.getSCAContentPayload(consumer);

    if (scaContainer == null || scaPayload == null) {
        throw new BadRequestException(i18n.tr("Cannot retrieve content access payload for consumer: {}",
            consumer));
    }

    // Check if either component has updated since the target date
    if (since != null) {
        OffsetDateTime sinceDate = Util.parseOffsetDateTime(SINCE_DATE_FORMATER, since);
        OffsetDateTime certUpdated = Util.toDateTime(Util.firstOf(scaCert.getUpdated(), new Date()));
        OffsetDateTime payloadUpdated = Util.toDateTime(scaPayload.getTimestamp());

        if (!since.isBefore(certUpdated) && !since.isBefore(payloadUpdated)) {
            return Response.status(Response.Status.NOT_MODIFIED)
                .entity("Not modified since date supplied.")
                .build();
        }
    }

    // No target date specified, or the components are newer than the target date
    List<String> pieces = List.of(scaCert.getCert(), scaPayload.getPayload());

    ContentAccessListing result = new ContentAccessListing()
        .setContentListing(scaCert.getSerial().getId(), pieces)
        .setLastUpdate(lastUpdated);

    return Response.ok(result, MediaType.APPLICATION_JSON)
        .build();
}
catch (ConcurrentContentPayloadCreationException exp) {
    throw new ServiceUnavailableException(i18n.tr("Unable to create content access payload"), exp);
}
```

Ceiu (changes requested):

src/main/java/org/candlepin/resource/ConsumerResource.java
Ceiu: As an alternative, if you don't like reading the date checks as "not before", we can reduce the number of negations with an or:

```
if (!(since.isBefore(certUpdated) || since.isBefore(payloadUpdated))) {
  ...
```

src/main/java/org/candlepin/resource/ConsumerResource.java
joshmalbrecht: I forgot to add the retry-after header value. Updated to set the retry-after header value to 2 seconds. The design doc doesn't specify what the value should be, so if anyone want's a different time, I can change it.

For the status code, the design doc says ` This can be either a 429 or a 503`, and I went with a 503 because it is really a server error and not a client error.  However, I don't have a strong opinion on this. With the investigation of CHAINSAW-354 we know the client does not retry with a 429 response code.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -805,6 +812,82 @@ public void shouldReturnANotModifiedReturnCodeWhenTheDataHasNotBeenUpdatedSinceD
...
+                () -> getContentAccessBodyTask(adminClient, consumer1),
+                () -> getContentAccessBodyTask(adminClient, consumer2),
+                () -> getContentAccessBodyTask(adminClient, consumer3),
+                () -> getContentAccessBodyTask(adminClient, consumer4),
+                () -> getContentAccessBodyTask(adminClient, consumer5));
+
+            List<Future<Boolean>> futures = execService.invokeAll(tasks);
+
+            // Wait for all the tasks to finish
+            boolean allTasksDone = false;
```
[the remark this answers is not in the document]
joshmalbrecht: I was able to replace the while loop that was waiting for the tasks to finish with the `execService.shutdown()` and `execService.awaitTermination`. 

However, in your example, if the `ApiException` occurs in one of the threads, the exception is not thrown and so the  `assertThrows` will never catch the exception. You can loop over the list of futures and call `future.get` and if the `ApiException` occurred in that thread, an `ExecutionException` will be thrown. But we can't assert against this exception. Unless I am missing something, I'm not seeing a better way of asserting the `ApiException` in each thread.

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2384,45 +2386,49 @@ public Response getContentAccessBody(@Verify(Consumer.class) String consumerUuid
...
 
         // The date provided by the client does not preserve milliseconds, so we need to round down
         // the dates preserved on the server by removing milliseconds for a proper date comparison
-        Date scaCertificateLastUpdated = cac.getUpdated() != null ?
-            Util.roundDownToSeconds(cac.getUpdated()) :
+        Date x509CertUpdated = x509Cert.getUpdated() != null ?
+            Util.roundDownToSeconds(x509Cert.getUpdated()) :
             new Date();
 
+        Date contentPayloadUpdated = contentPayload.getTimestamp() != null ?
```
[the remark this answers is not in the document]
joshmalbrecht: Updated.

I removed the `ConsumerResourceTest.testGetContentAccessBodyWithSinceDateEqualToLastUpdatedDate` test because with the removal of the `Util.roundDownToSeconds` the x509 update date and payload timestamp cannot equal the since target value because the target value is second precision and not millisecond precision.

src/main/java/org/candlepin/resource/ConsumerResource.java
joshmalbrecht: Updated.

I made a couple of changes to account for the `ContentAccessListing.setLastUpdate` needing the most recent date between the X509 update date and the content payload timestamp date.

Ceiu: Ahh, the perils of pseudo-code.

Alright, since we still need the combined last update date in either case, then we can revert to the original block of date juggling and save the ODT conversion for the bit inside the "since" block. This also means we can skip one of the before checks:

```
Date certUpdated = Util.firstOf(scaCert.getUpdated(), new Date());
Date payloadUpdated = scaPayload.getTimestamp();
Date lastUpdated = certUpdated.after(payloadUpdated) ? certUpdated : payloadUpdated;

if (since != null) {
  OffsetDateTime sinceDate = Util.parseOffsetDateTime(SINCE_DATE_FORMATER, since);

  if (!sinceDate.isBefore(Util.toDateTime(lastUpdated)) {
    <response stuff here>
  }
}

...

ContentAccessListing result = new ContentAccessListing()
  ...
  .setLastUpdate(lastUpdated);
```

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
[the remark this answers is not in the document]
Ceiu: Sorry, this was probably made unclear by my line saying to use the existing code from 838-855.

The existing code calls out to a method which has a try/catch around that. If you update the task bit to instead just invoke the call directly, does it propagate the exception as I'm expecting?

```
List<Callable<Boolean>> tasks = List.of(
  () -> adminClient.consumers().getContentAccessBody(consumer1.getUuid(), null),
  () -> adminClient.consumers().getContentAccessBody(consumer2.getUuid(), null),
  ...);
```

If that line is getting too long or unwieldy, the `ConsumerClient` could be captured and reused a bunch:

```
ConsumerClient cclient = this.adminClient.consumers();

List<Callable<Boolean>> tasks = List.of(
  () -> cclient.getContentAccessBody(consumer1.getUuid(), null),
  () -> cclient.getContentAccessBody(consumer2.getUuid(), null),
  ...
```

Or if you really want to go crazy with the closures:

```
java.util.function.Consumer<ConsumerDTO> task = (consumer) -> 
  adminClient.consumers().getContentAccessBody(consumer.getUuid(), null);

List<Callable<Boolean>> tasks = List.of(
  () -> task.accept(consumer1),
  () -> task.accept(consumer2),
  ...
```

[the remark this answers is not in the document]
Ceiu: Looking at the spec for `Future`, it implies the exception may be captured until you call `.get(...)` later on it. In which case, you'll need to expect *that* exception and then unwrap it while doing the confirmation, or unwrap the exception straight away and then rethrow it.

src/main/java/org/candlepin/resource/ConsumerResource.java
joshmalbrecht: Updated

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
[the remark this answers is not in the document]
joshmalbrecht: Yeah, I was doing the call directly and the error does not propagate even though the ApiException occurs in the thread. 

```
List<Callable<Boolean>> tasks = List.of(
  () -> adminClient.consumers().getContentAccessBody(consumer1.getUuid(), null),
  () -> adminClient.consumers().getContentAccessBody(consumer2.getUuid(), null),
  ...);
```

Like you said, unwrapping the `ExecutionException` when calling `Future.get()` and rethrowing the `ApiException` works nicely!

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -797,14 +810,77 @@ public void shouldReturnANotModifiedReturnCodeWhenTheDataHasNotBeenUpdatedSinceD
...
+                List<Future<String>> futures = execService.invokeAll(tasks);
+                execService.shutdown();
+                execService.awaitTermination(10L, TimeUnit.SECONDS);
+
+                for (Future<String> future : futures) {
+                    try {
+                        future.get();
+                    }
+                    catch (ExecutionException e) {
+                        throw (ApiException) e.getCause();
```
Ceiu, line 871: This is a tad risky if anything else goes wrong in here -- any other exception will result in a `ClassCastException` when we try to rethrow. Brain is fairly toasted today, so the only thing that comes to mind is a simple `instanceof` check and rethrowing in both cases to ensure we fail the right way no matter what:

```
Throwable cause = e.getCause();
if (cause instanceof ApiException) {
  throw (ApiException) cause;
}

throw cause;
```

If there's a cleaner way to do this check and cast safely, feel free to use that instead.

Ceiu (changes requested):
One last thing from me, and then I'll leave this to @nikosmoum.

src/main/java/org/candlepin/resource/ConsumerResource.java
nikosmoum: to be clear, the reason I mention 429 was beause I thought that even though the client wasn't retrying, it was at least handling it and printing out a user-friendly error (based on the investigation in CHAINSAW-354), while we don't know if 503 is being handled at all and may result in a stacktrace and even a potentially broken state in the client. If we prefer 503 we should at least do the same investigation for that status code first and see how the client reacts (for each of the 3 affected endpoints)

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
joshmalbrecht: Updated

joshmalbrecht: Looking at this a little bit closer, we can just throw the `ExecutionException.getCause()` because the `assertThrows` will do the `instanceof` check and cast.

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -797,14 +810,77 @@ public void shouldReturnANotModifiedReturnCodeWhenTheDataHasNotBeenUpdatedSinceD
 
         ZonedDateTime inputZonedDateTime = ZonedDateTime
             .parse(lastUpdate, DateTimeFormatter.ofPattern(CONTENT_ACCESS_OUTPUT_DATE_FORMAT));
-        OffsetDateTime since = inputZonedDateTime.toOffsetDateTime();
+        OffsetDateTime since = inputZonedDateTime.toOffsetDateTime()
+            .plus(1L, ChronoUnit.MINUTES);
```
nikosmoum, line 814: Why was this changed? That implies that logic changed on the server that shouldn't have. If this is some kind of workaround for a limitation on what we're trying to test there should at least be an explanation for it in a comment

nikosmoum (changes requested):
I will look at it again next week, but there's one thing that stood out. Also the commit message should change to mention 429 instead of 503

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
joshmalbrecht: This test was failing because we removed the `Util.roundDownToSeconds` when comparing the client provided 'ifModifiedSince' date with the most recent x509 date or the content payload timestamp date. The reason we removed this is in the discussion link below.

In this particular test, the client provides the content payload update date as the 'ifModifiedSince' query parameter. However, this date loses it's milliseconds precision because the date format is `yyyy-MM-dd'T'HH:mm:ssZ`. So, when we compare the `ifModifiedSince` date with the content access payload date, the `ifModifiedSince` ends up being **before** the content access payload date. This results in the `Response.Status.NOT_MODIFIED` not being returned as a response and failing the test.

If we want to not modify the spec test  and have it pass, we need to add the `Util.roundDownToSeconds` back to shaving off the milliseconds from the x509 update date and content access payload timestamp before comparing to the `ifModifiedSince` in `ConsumerResource.getContentAccessBody`.

https://github.com/candlepin/candlepin/pull/5045#discussion_r1973852231

Ceiu: This is likely due to the removal of the unnecessary millisecond truncating on the component dates and something in the date comparison not going quite right. In the first review cycle I thought that was normal and maybe clients would have a 999ms window in which to make a request and not get an updated cert if they updated within that window; but thinking on it again, that's not right at all so something may have gone wrong here.

Basically, millisecond resolution on our cert or payload should be extraneous if our check is correct. It won't hurt to be there so it's a waste cycles to remove it. If either component has a date that is exactly equal to or after (e.g. not before) to the target date, then we return the updated container:

```
if (since != null) {
    OffsetDateTime sinceODT = Util.parseOffsetDateTime(SINCE_DATE_FORMATER, since);
    OffsetDateTime lastUpdatedODT = lastUpdated.toInstant().atOffset(ZoneOffset.UTC);

    if (lastUpdatedODT.isBefore(sinceODT)) {
        return Response.status(Response.Status.NOT_MODIFIED)
            .entity("Not modified since date supplied.")
    }
}
```

Currently the check elsewhere in code is similar but in the inverse order with a negation, having evolved from two independent checks; so maybe that's where things are going slightly sideways.

Ceiu: As a side note, something I find to be quite amusing here, following the consequences of the intent of the original logic that truncated milliseconds and explicitly allowed after or equals, is that as-written it would prevent repeat client requests within the first second of a cert's lifecycle from ever seeing the "not modified since" response.

In terms of optimizing the output and the whole reason you do this check in the first place, this is something I would argue to be a bug. If we're even *considering* reintroducing manipulation of the dates for the purposes of comparison, we should be *rounding up* the timestamps of the cert and payload, not truncating them.

joshmalbrecht: Wouldn't rounding up the timestamps of the cert and payload increase the span of time that the client would not receive the `Response.Status.NOT_MODIFIED`?

Ceiu: Argh... Github doesn't live update for me anymore and I thought I'd delete the comment before it started any side discussions, but was way too slow.

Anyway, yes it would in the case where a component was generated at an exact second, as we'd have to wait for the next second to roll around.

My original thought was a client that was hammering on the endpoint asking for the current second of the cert would get filtered out after the first request; but that's not accurate and given the timestamp is optional and client-provided here, a malicious client could just set the time in the past and keep requesting the cert anyway. Or just not provide the time.

PR #5053, opened 2025-03-03
[F] CHAINSAW-377: Updated ConsumerResource.getEntitlementCertificates

joshmalbrecht:
- Moved ConsumerResource.getEntitlementCertificates into ConsumerResource.exportCertificates.
- Updated ConsumerResource.getEntitlementCertificatesForConsumer to return a 429 response when there is a concurrent content payload creation exception.


src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -2224,56 +2224,6 @@ public void deleteConsumer(@Verify(Consumer.class) String uuid) {
...
-     * @param serials
-     *  Certificate serial
-     *
-     * @return
-     *  List of DTOs representing certificates
-     */
-    // TODO: FIXME: This method is only public due to it being called directly by tests. Refactor the
-    // tests to not be invoking individual methods like this and stick to the testing the public interface
-    // directly. Implementation details leaking into tests makes for very brittle and unreliable tests.
-    public List<CertificateDTO> getEntitlementCertificates(
```
nikosmoum, line 2243: I'm unsure about removing this as a public method. While the FIXME comment is technically correct, that is only because we had to apply a workaround. This method used to be a top-level API endpoint handler method.
When the caller of `GET /consumers/{consumer_uuid}/certificates` specified `application/json` as the content type, then `getEntitlementCertificates` would be called. But when for the same verb+url they specified `application/zip`, then the `exportCertificates` method would be called.

However, openapi-generator does not support different handler methods for different content types (as they should have been), so when we moved to it we had to do a workaround where only 1 method is defined in the spec (exportCertificates), and it does the content-type check itself (redirecting the call to `getEntitlementCertificates`  accordingly if need be).

Because of that, I think we should keep these as 2 separate methods, since they logically handle 2 different types of calls. What do you think @Ceiu @joshmalbrecht @bakajstep

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -876,7 +876,67 @@ public void shouldReturnServiceUnavailableWithConcurrentPayloadCreation() throws
             .isTooMany()
             .hasHeaderWithValue(RETRY_AFTER_HEADER_KEY,
                 CONTENT_PAYLOAD_CREATION_EXCEPTION_RETRY_AFTER_TIME);
+    }
+
+    @Test
+    public void shouldReturnServiceUnavailableWithConcurrentPayloadCreationWhenExportingCertificates() {
```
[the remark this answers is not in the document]
joshmalbrecht: ah good catch! Updated in both tests

spec-tests/src/test/java/org/candlepin/spec/content/ContentAccessSpecTest.java
```diff
@@ -876,7 +876,67 @@ public void shouldReturnServiceUnavailableWithConcurrentPayloadCreation() throws
...
+        ConsumerDTO consumer5 = adminClient.consumers().createConsumer(Consumers.random(owner));
+
+        List<ConsumerDTO> consumers = List.of(consumer1, consumer2, consumer3, consumer4, consumer5);
 
+        int maxNumberOfAttempts = 5;
+        ApiException exception = assertThrows(ApiException.class, () -> {
+            for (int i = 0; i < maxNumberOfAttempts; i++) {
+                // Update the arch of all the consumers so that the content access payload key will be
+                // different than the key used to store the content access payload previously. This causes the
+                // regeneration of the content access payload which is what we need to cause the 503 response.
```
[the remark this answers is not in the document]
joshmalbrecht: Updated in both tests

src/main/java/org/candlepin/resource/ConsumerResource.java
Ceiu: Nah, I don't really agree here. The end result is that we have one endpoint which handles two types of requests, and that's what we should be testing. By testing the implementation details that is the two branching paths and testing that instead*, we're not properly testing the unit (endpoint) itself. Besides, if we already have proper testing for each individual method, then it's just a matter of updating them to call the single entry point with the right headers set to get the targeted output.

* I don't believe that this choice was made because of the limitations imposed by OpenAPI, but rather to minimize the amount of code changes.

joshmalbrecht: Yeah, I wasn't really considering the OpenAPI limitations by moving `ConsumerResource.getEntitlementCertificates` into `ConsumerResource.exportCertificates`. I was wanting to just test only the endpoint's entry point method and not have two sets of tests when they are really testing the same unit.

nikosmoum (approved):

PR #5057, opened 2025-03-05
[4.5] SCA Component Generation Optimizations

joshmalbrecht:
Merging the feature/sca_component_generation_optimizations into candlepin-4.5-HOTFIX


Ceiu:
nack, we'll want to squash this down to a handful (or fewer) commits, and not include things like the various library updates we've had along the way.

PR #5058, opened 2025-03-05
[M] CHAINSAW-341: Improved SCA payload regeneration performance

joshmalbrecht:
CHAINSAW-360: Moved the fact constants from X509Util to Consumer.Facts

- Renamed the UNAME_MACHINE consumer fact to ARCHITECTURE.

CHAINSAW-356: Added last content update field to environments

- Updated environments to record the last content update timestamp whenever content was promoted or demoted to an environment, or the content's prefix was changed.

CHAINSAW-357: Refactored SCACertficateGenerator

- Refactored the SCACertificateGenerator to allow individual component fetching.
- Created ContentAccessPayload which stores the content access payload for a SCA owner.

CHAINSAW-375: Refactored exceptions

- Refactored the ServiceUnavailableException and the TooManyRequestsException to have the standard exception constructors.
- Added the ability to define a 'Retry-After' header for the ServiceUnavailableException.
- Added exception messages for TooManyRequestsExceptions in SecurityHoleAuthorizationFilter.

CHAINSAW-358: Updated getEntitlementCertificateSerials serial retrieval

- Updated ConsumerResource.getEntitlementCertificateSerials to retrieve only the X509 certificate from the SCACertificateGenerator when retrieving the content access certificate serial.

CHAINSAW-361: Removed obsoleted methods from ContentAccessManager

- Removed the 'hasCertChangesSince' and 'syncOwnerLastContentUpdate' methods from the ContentAccessManager class.

CHAINSAW-359: Updated ConsumerResource.getContentAccessBody

- Updated ConsumerResource.getContentAccessBody to fetch the individual components from the SCACertificateGenerator instead of retrieving the entire assembled certificate.
- Updated ConsumerResource.getContentAccessBody to return a 429 response if there is a constraint violation when persisting the SCA certificate's content access payload.

CHAINSAW-377: Updated ConsumerResource.getEntitlementCertificates

- Moved ConsumerResource.getEntitlementCertificates into ConsumerResource.exportCertificates.
- Updated ConsumerResource.getEntitlementCertificatesForConsumer to return a 429 response when there is a concurrent content payload creation exception.

CHAINSAW-376: Updated ConsumerResource.exportCertificates exception handling

- Updated ConsumerResource.exportCertificates to throw TooManyRequestsException (HTTP status code 429) when a conflict in SCA payload generation happens.
- Renamed two tests in ExportSpecTest to be more accurate.

## Note

I did have merge conflicts when squashing the commits, so double checking that nothing is missing is a good idea. The original feature branch can be found here for comparison:

https://github.com/candlepin/candlepin/tree/feature/sca_component_generation_optimizations


src/main/java/org/candlepin/resource/EnvironmentResource.java
```diff
@@ -399,7 +399,9 @@ public AsyncJobStatusDTO promoteContent(@Verify(Environment.class) String envId,
         Set<String> contentIds;
         try {
             contentIds = this.batchCreate(contentToPromote, environment);
-            this.contentAccessManager.syncOwnerLastContentUpdate(environment.getOwner());
+            // TODO: Remove the line below once certificate generation based on the `lastContentUpdate` field
```
Ceiu, line 402: Whoops... these slipped through the cracks. Neither should exist anymore: batchCreate is doing an env sync already here, and in demoteContent we have an env sync after the try/catch.

Ceiu (changes requested):
Some TODOs snuck through here; possibly as a merge artifact.

src/main/java/org/candlepin/resource/EnvironmentResource.java
joshmalbrecht: Good catch! It looks like we missed removing these in the feature branch:

https://github.com/candlepin/candlepin/blob/feature/sca_component_generation_optimizations/src/main/java/org/candlepin/resource/EnvironmentResource.java#L445-L447

Ceiu (approved):

src/main/java/org/candlepin/resource/EnvironmentResource.java
joshmalbrecht: So, this exposed an issue with the `ContentAccessSpecTest.shouldShowEnvironmentContentChangeInContentAccessCert` spec tests for MySQL because MySQL does not store millisecond precision. Sometimes, the environment's last content update datetime would be in the same second as the content access payloads timestamp, but with more milliseconds in the future. So, when we check if there have been content changes, the comparison between the environment's last content update time and the content access payloads timestamp ends up being equal because of the loss of the milliseconds on the env last content update datetime.

To fix this, I had to add two sleeps for a second to guarantee that the environment's last conet

Ceiu: Wouldn't that have been a problem even with that extra org.syncLastContentUpdate in place?

joshmalbrecht: Yeah, that's a good point. Maybe the `environment.getOwner().syncLastContentUpdate()` call was far enough in the call stack that it usually was being set in the next second in comparison to the content access payloads timestamp, but that is just my best guess. I wouldn't be surprised if this test was occasionally failing/flaky.

PR #5062, opened 2025-03-10
[M] CANDLEPIN-999: Fixed paging for list owner pools

joshmalbrecht:
- Updated the list owner pools endpoint to not filter the pools twice when paging is requested.
- Updated list owner pools to return a bad request response instead of an internal server error response when an invalid order key is provided in the request.
- Moved the paging of the pools out of the PoolManager class and into the OwnerResource.listOwnerPools and PoolResource.listPools.

### DIFF
--- a/src/main/java/org/candlepin/resource/PoolResource.java
+++ b/src/main/java/org/candlepin/resource/PoolResource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ * Copyright (c) 2009 - 2025 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -37,6 +37,7 @@ import org.candlepin.model.Owner;
 import org.candlepin.model.OwnerCurator;
 import org.candlepin.model.Pool;
 import org.candlepin.model.Pool.PoolType;
+import org.candlepin.model.PoolCurator;
 import org.candlepin.model.PoolQualifier;
 import org.candlepin.model.SubscriptionsCertificate;
 import org.candlepin.paging.Page;
@@ -73,6 +74,7 @@ public class PoolResource implements PoolsApi {
 
     private final ConsumerCurator consumerCurator;
     private final OwnerCurator ownerCurator;
+    private final PoolCurator poolCurator;
     private final I18n i18n;
     private final PoolManager poolManager;
     private final PoolService poolService;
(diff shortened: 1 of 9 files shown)


src/test/java/org/candlepin/resource/ConsumerResourceTest.java
```diff
@@ -876,7 +876,7 @@ public void validIdCertDoesNotRegenerate() {
         ComplianceStatus status = new ComplianceStatus();
         when(complianceRules.getStatus(any(Consumer.class), any(Date.class), anyBoolean()))
             .thenReturn(status);
-        consumer.setIdCert(createIdCert(TestUtil.createDate(2025, 6, 9)));
+        consumer.setIdCert(createIdCert(TestUtil.createDateOffset(1, 0, 0)));
```
joshmalbrecht, line 879: This test started failing today (03/11) because this hard coded date is exactly 90 days from now and we consider ID certs expired if the expiration date is 90 days in the future. So, this test began taking a different code path because the ID cert was considered expired. Updated to use a relative date.

Ceiu: Given the other branches got a dedicated PR for this, that's probably apt here as well; if for no other reason than git blame and history cleanliness.

src/main/java/org/candlepin/resource/PoolResource.java
```diff
@@ -171,10 +174,18 @@ public List<PoolDTO> listPools(
...
                 .setLimit(pageRequest.getPerPage());
+
+            boolean reverse = pageRequest.getOrder() == PageRequest.DEFAULT_ORDER;
+            qualifier.addOrder(pageRequest.getSortBy(), reverse);
         }
 
         Page<List<Pool>> pageResponse = poolManager.listAvailableEntitlementPools(qualifier);
+
         List<Pool> poolList = pageResponse.getPageData();
+        if (qualifier.getOffset() != null && qualifier.getLimit() != null) {
```
Ceiu, line 185: If we have the offset and limit from the qualifier, we could be having them applied to the query rather than manually parsing out the list ourselves. Then we can just pass through the list of pools as received from the curator and throw it in the page. Unless there's something I'm not seeing that requires we pull everything and do the offset/limit ourselves.

May also be apt to not return a Page object at all from the manager/curator level. That was always a bit strange.

Ceiu: As an example of this, the `AsyncJobStatusCurator` uses a `QueryArguments` object to build out queries and applies the offset+limit from it to the query. This can be seen in `.findJobs(...)` around line 307:

```
        TypedQuery<AsyncJobStatus> query = this.getEntityManager()
            .createQuery(criteriaQuery);

        if (queryArgs != null) {
            Integer offset = queryArgs.getOffset();
            if (offset != null && offset > 0) {
                query.setFirstResult(offset);
            }

            Integer limit = queryArgs.getLimit();
            if (limit != null && limit > 0) {
                query.setMaxResults(limit);
            }
        }

        return query.getResultList();
```

Ceiu (changes requested):
We're likely kicking at a hornet's nest here, but something about seeing `.subList` against the output of a collection of expensive objects worries me.

src/main/java/org/candlepin/resource/PoolResource.java
joshmalbrecht: I don't think we can do the paging on the query because the set of pools is being filtered again in `PoolManager.listAvailableEntitlementPools`.  If the curator returns 5 pools because the limit on the qualifier is 5 and then we filter the pools again in `PoolManager.listAvailableEntitlementPools`, then we could return less than the requested page size.

https://github.com/candlepin/candlepin/blob/jalbrech/m/CANDLEPIN-999/src/main/java/org/candlepin/controller/PoolManager.java#L1562

Ceiu: Uuuuugh... alright, disregard.

src/test/java/org/candlepin/resource/ConsumerResourceTest.java
joshmalbrecht: I rebased with main to pick up the changes that were merged into main that addressed this issue

Ceiu (approved):

PR #5071, opened 2025-03-12
[4.3] Increased upload action version due to deprecation

bakajstep:
Version 3 was deprecated at the beginning of the year.[1]

[1] https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/


Ceiu:
Typo in commit/pr title. Should be "deprecation"

PR #5072, opened 2025-03-13
[M] CANDLEPIN-996: Additional SAST scan fixes

bakajstep:
- Refactored the get() method to automatically close the InputStream acquired from loader.getResourceAsStream().
- Added a null check for when the "settings.properties" resource was not found.


bakajstep:
After this changes will get approved i will backport them in rest of the branches.

bakajstep:
> Are there going to be more changes coming from the output of this static analysis tool? If so, we should bulk them all up so we don't have a bunch of semi-ambiguous commits in the history.

currently there are no more changes coming, this is split into two commits because I fixed only the positives and the rest I marked as false positives but then I was told that we have to fix all the ones that we can fix so I had to fix the rest too, that's why it's in two commits. For the other branches I marked the PRs as drafts so I can add the changes to one commit

Ceiu:
> I was told that we have to fix all the ones that we can fix

Told by whom?

PR #5080, opened 2025-03-26
[M] CANDLEPIN-990: Updated cp_consumer_environments.priority to tinyint

joshmalbrecht:
- Updated the cp_consumer_environments.priority column from a varchar to a tinyint to prevent ordering issues when a consumer has more than 10 environments.

### DIFF
--- a/src/main/resources/db/changelog/20250325000000-update-env-priority-column-type.xml
+++ b/src/main/resources/db/changelog/20250325000000-update-env-priority-column-type.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <changeSet id="20250325000000-1" author="jalbrech">
+        <sql dbms="mysql, mariadb">
+            DELETE FROM cp_consumer_environments WHERE priority REGEXP '[^0-9]+';
+        </sql>
+
+        <sql dbms="postgresql">
+            DELETE FROM cp_consumer_environments WHERE priority ~ '[^0-9]+';
+        </sql>
+
+        <modifyDataType tableName="cp_consumer_environments"
+            columnName="priority"
+            newDataType="SMALLINT" />
+    </changeSet>
+
+</databaseChangeLog>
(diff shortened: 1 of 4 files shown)


bakajstep (approved):

src/main/resources/db/changelog/20250325000000-update-env-priority-column-type.xml
```diff
@@ -0,0 +1,23 @@
...
+    <changeSet id="20250325000000-1" author="jalbrech">
+        <sql dbms="mysql, mariadb">
+            DELETE FROM cp_consumer_environments WHERE priority REGEXP '[^0-9]+';
+        </sql>
+
+        <sql dbms="postgresql">
+            DELETE FROM cp_consumer_environments WHERE priority ~ '[^0-9]+';
+        </sql>
+
+        <modifyDataType tableName="cp_consumer_environments"
```
Ceiu, line 18: Does this work on both if there is already data in the table? MySQL automatically converts the data as we discovered, but PostgreSQL requires an explicit cast (`USING priority::integer` likely).

Also the data type of "TINYINT" may cause problems on psql unless something in Liquibase is aliasing it for us.

joshmalbrecht: Postgres will convert the data in the column like MySQL. To test this, I did the following steps:

1. Reverted the commit so that the database changes and service changes are not present.
2. Deploy using Postgres:  `./bin/deployment/deploy -gaH` .
3. Create and run a spec test that creates environments and consumers in those environments to populate data.
4. Remove the revert commit.
5. Deploy again with the database and service changes: `./bin/deployment/deploy -aH`
6. Verify a successful deploy and check the cp_consumer_environments table in Postgres.

It looks like Liquibase is aliasing "TINYINT" to the smallint data type in Postgres:

```
candlepin=> \d cp_consumer_environments;
                 Table "public.cp_consumer_environments"
     Column     |         Type          | Collation | Nullable | Default 
----------------+-----------------------+-----------+----------+---------
 cp_consumer_id | character varying(32) |           | not null | 
 environment_id | character varying(32) |           | not null | 
 priority       | smallint              |           | not null | 

```

Ceiu: That's good, though we should probably change it to SMALLINT and let the conversion happen for MySQL as that's the standard.

The concerning bit here is that while Liquibase provides (a table)[https://docs.liquibase.com/concepts/data-type-handling.html] for showing how their data type magic translates across databases, HyperSQL/HSQLDB is currently absent from the table. Fortunately it looks like we're fine here for now in either case, as HyperSQL supports both:
> Non Standard types such as AUTO_INCREMENT, TINYINT, DATETIME, TEXT, TINYLOB, MEDIUMLOB are also supported.

PR #5081, opened 2025-03-27
[M] CANDLEPIN-1001: Refactored ConsumerCloudData creation/update

bakajstep:
- Created builder for ConsumerCloudData with logic that ensure that only valid instances are created
- Created parser for extracting proper value from Consumer facts
- Added logic for updating ConsumerCloudData when Consumer update is performed with new cloud facts.


src/main/java/org/candlepin/resource/util/ConsumerCloudDataBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+     * @throws BadRequestException if more than one parser supports the provided facts.
+     */
+    private Optional<ConsumerCloudData> buildConsumerCloudData(Map<String, String> facts) {
+        List<CloudProviderFactParser> supportedParsers = parsers.stream()
+            .filter(parser -> parser.isSupported(facts))
+            .toList();
+
+        if (supportedParsers.size() > 1) {
+            throw new BadRequestException(this.i18n.tr("During consumer cloud creation more than one " +
+                "cloud provider facts found"));
```
joshmalbrecht, line 91: The translated text needs to be on the same line. You can use `i18n.marktr` on a separate line and use the returned string for the `BadRequestException`.

src/main/java/org/candlepin/resource/util/ConsumerCloudDataBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+import javax.inject.Inject;
+
+/**
+ * Builder for constructing {@link ConsumerCloudData} from consumer facts.
+ * <p>
+ * This class selects the appropriate {@link CloudProviderFactParser} based on the provided facts and
+ * builds a {@link ConsumerCloudData} object containing cloud-specific information.
+ * </p>
+ */
+public class ConsumerCloudDataBuilder {
```
joshmalbrecht, line 39: We can use a `Singleton` annotation on this class

src/main/java/org/candlepin/model/Consumer.java
```diff
@@ -1294,6 +1295,26 @@ public ConsumerCloudData getConsumerCloudData() {
         return this.consumerCloudData;
     }
 
+    public boolean checkForCloudIdentifierFacts(Map<String, String> incomingFacts) {
```
joshmalbrecht, line 1298: I think this method might make more sense as a private method in `ConsumerResource`. You could also add an additional parameter for the existing consumer facts so that you can use `Consumer.getFacts()` as the existingFacts parameter input. Also, I think we can name this method to indicate that it is checking for updates between the existing consumer facts and the provided consumer facts. Maybe something like `checkForCloudFactUpdates`?

`private boolean checkForCloudFactUpdates(Map<String, String> incomingFacts, Map<String, String> existingFacts)`

src/main/java/org/candlepin/resource/ConsumerResource.java
```diff
@@ -1654,6 +1627,18 @@ public void updateConsumer(@Verify(Consumer.class) String uuid, ConsumerDTO dto)
         guestMigration.buildMigrationManifest(dto, toUpdate);
 
         if (performConsumerUpdates(dto, toUpdate, guestMigration)) {
+            if (toUpdate.checkForCloudIdentifierFacts(dto.getFacts())) {
+                log.warn("Cloud identifier fact change detected");
```
joshmalbrecht, line 1631: Can we also add the Consumer UUID to this warning? It will be helpful for when we are looking in the logs

joshmalbrecht (changes requested):
I reviewed the non-test files today, and can take a look at the tests tomorrow!

src/test/java/org/candlepin/model/ConsumerCloudDataTest.java
```diff
@@ -466,15 +466,73 @@ public void testSetConsumer() {
...
+        assertEquals("acc1", original.getCloudAccountId());
+        assertEquals("inst1", original.getCloudInstanceId());
+        assertEquals(List.of("off1", "off2"), original.getCloudOfferingIds());
+    }
+
+    @Test
+    void testUpdateFromUpdatesOnlyChangedFields() {
+        ConsumerCloudData original = new ConsumerCloudData()
+            .setCloudProviderShortName("aws")
+            .setCloudAccountId("acc1")
```
joshmalbrecht, line 489: We should also verify that cloud account ID and offering IDs get updated as expected when they are updated. Currently there are no tests that validate that these fields get updated

src/test/java/org/candlepin/model/ConsumerTest.java
```diff
@@ -703,6 +706,77 @@ public void testSetConsumerCloudData() {
...
+
+    @ParameterizedTest(name = "Should return false for existing: {0} and incoming: {1}")
+    @MethodSource("falseCases")
+    void testCheckForCloudIdentifierFactsFalse(Map<String, String> existingFacts,
+        Map<String, String> incomingFacts) {
+        Consumer consumer = new Consumer();
+        consumer.setFacts(existingFacts);
+
+        assertFalse(consumer.checkForCloudIdentifierFacts(incomingFacts));
+    }
```
joshmalbrecht, line 727: Since the only difference between these two tests is the expected true/false result from `consumer.checkForCloudIdentifierFacts`, we can combine these two tests into one test, combine the method sources into one method source, and add an `expected` boolean value to the method source. This way, you can assert against the expected value provided by the method source.

joshmalbrecht (changes requested):

src/main/java/org/candlepin/model/Consumer.java
bakajstep: This was one of the comments in design document from Nikos that we allready have similiar method so I should use similiar approach. https://github.com/candlepin/candlepin/blob/main/src/main/java/org/candlepin/model/Consumer.java#L560

nikosmoum: I proposed adding it to Consumer.java because there was a precedent already, but I suppose we could move both of these methods (`checkForCloudProfileFacts` and `checkForCloudIdentifierFacts`) into ConsumerResource, since it does make sense. 
Also, I think the word 'identifier' needs to stay in the new method name since it is clearer which facts we are talking about compared with the other method

joshmalbrecht: Ah that is my bad! I didn't see that there was already a method like this in Consumer.java. Knowing this now, I'm fine with keeping the two methods in Consumer.java or moving them to ConsumerResource, but we should do the same thing for both methods!

I also agree with Nikos on keep 'identifier' in the new method name.

bakajstep: so how do we do it then? should I move it to ConsumerResource?

src/main/java/org/candlepin/resource/util/ConsumerCloudDataBuilder.java
bakajstep: Done

src/main/java/org/candlepin/resource/ConsumerResource.java
bakajstep: Good idea, done

src/test/java/org/candlepin/model/ConsumerCloudDataTest.java
bakajstep: done

src/test/java/org/candlepin/model/ConsumerTest.java
bakajstep: `    @ParameterizedTest(name = "For existing: {0} and incoming: {1} expected: {2}")
    @MethodSource("cases")
    void testCheckForCloudIdentifierFacts(Map<String, String> existingFacts, Map<String, String> incomingFacts, boolean expected) {
        Consumer consumer = new Consumer();
        consumer.setFacts(existingFacts);
        assertEquals(expected, consumer.checkForCloudIdentifierFacts(incomingFacts));
    }

    private static Stream<Arguments> cases() {
        return Stream.of(
            // Cases where the method should return true:
            // Case 1: When existing facts are null and incoming facts contain a cloud identifier.
            Arguments.of(null, Map.of(AWS_ACCOUNT_ID.getValue(), "123"), true),

            // Case 2: When the cloud identifier value differs (e.g., "123" vs "456").
            Arguments.of(Map.of(AWS_ACCOUNT_ID.getValue(), "123"),
                         Map.of(AWS_ACCOUNT_ID.getValue(), "456"), true),

            // Case 3: When there are multiple cloud keys and at least one value differs.
            Arguments.of(
                Map.of(AWS_ACCOUNT_ID.getValue(), "123", 
                       AWS_INSTANCE_ID.getValue(), "inst1",
                       AZURE_OFFER.getValue(), "offer1"),
                new HashMap<String, String>() {{
                    put(AWS_ACCOUNT_ID.getValue(), "123");
                    put(AWS_INSTANCE_ID.getValue(), "inst2");
                    put(AZURE_OFFER.getValue(), "offer1");
                }},
                true
            ),

            // Cases where the method should return false:
            // Case 4: When incoming facts are null.
            Arguments.of(Map.of(AWS_ACCOUNT_ID.getValue(), "123"), null, false),

            // Case 5: When incoming facts are identical to existing facts.
            Arguments.of(
                Map.of(AWS_ACCOUNT_ID.getValue(), "123", 
                       AWS_INSTANCE_ID.getValue(), "inst1"),
                Map.of(AWS_ACCOUNT_ID.getValue(), "123", 
                       AWS_INSTANCE_ID.getValue(), "inst1"),
                false
            ),

            // Case 6: When incoming facts do not contain any cloud identifier key.
            Arguments.of(Map.of("non_cloud_key", "value1"), Map.of("non_cloud_key", "value2"), false),

            // Case 7: When incoming contains an extra non-cloud key but cloud facts remain the same.
            Arguments.of(
                Map.of(AWS_ACCOUNT_ID.getValue(), "123", 
                       AWS_INSTANCE_ID.getValue(), "inst1",
                       AZURE_OFFER.getValue(), "offer1"),
                new HashMap<String, String>() {{
                    put(AWS_ACCOUNT_ID.getValue(), "123");
                    put(AWS_INSTANCE_ID.getValue(), "inst1");
                    put(AZURE_OFFER.getValue(), "offer1");
                    put("non_cloud_key", "some_value");
                }},
                false
            )
        );
    }
`

bakajstep: So which way we prefer?

src/main/java/org/candlepin/model/Consumer.java
joshmalbrecht: Let's keep this method in Consumer.java.

src/test/java/org/candlepin/model/ConsumerTest.java
joshmalbrecht: Now that I see the combined cases, I agree with you that separating the tests into two methods `testCheckForCloudIdentifierFactsTrue` and `testCheckForCloudIdentifierFactsFalse` is less confusing! Lets keep the tests as they are. Sorry for the confusion!

joshmalbrecht (approved):

PR #5103, opened 2025-04-18
[M] CANDLEPIN-1014: Fix Content URLs in SCA cert for Multi Env

nikosmoum:
- When 2 or more environment names contain slashes, where one environment's name is a sub path of the other, the SCA cert's Authorized Content URLs section will now only contain the path from the environment with the base path (regardless of priority of environments). Example: Given environments A and A/1 in OrgX, applied to a given consumer, only the path /OrgX/A should be present in the cert.

### DIFF
--- a/src/main/java/org/candlepin/pki/huffman/Huffman.java
+++ b/src/main/java/org/candlepin/pki/huffman/Huffman.java
@@ -135,30 +135,36 @@ public class Huffman {
         }
     }
 
-    private void makePathForURL(StringTokenizer st, PathNode parent, PathNode endMarker) {
-        if (st.hasMoreTokens()) {
-            String childVal = st.nextToken();
-            if (childVal.isEmpty()) {
+    private void makePathForURL(StringTokenizer tokenizer, PathNode parent, PathNode endMarker) {
+        if (tokenizer.hasMoreTokens()) {
+            String nextValue = tokenizer.nextToken();
+            if (nextValue.isEmpty()) {
                 return;
             }
             boolean isNew = true;
-            for (NodePair child : parent.getChildren()) {
-                if (child.getName().equals(childVal) &&
-                    !child.getConnection().equals(endMarker)) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/candlepin/pki/huffman/Huffman.java
```diff
@@ -137,28 +137,34 @@ public void printTrie(HuffNode hn, int tab) {
...
-            for (NodePair child : parent.getChildren()) {
-                if (child.getName().equals(childVal) &&
-                    !child.getConnection().equals(endMarker)) {
-                    makePathForURL(st, child.getConnection(), endMarker);
+            for (NodePair peer : parent.getChildren()) {
+                if (peer.getName().equals(nextValue)) {
                     isNew = false;
+                    if (!peer.getConnection().equals(endMarker)) {
+                        if (!st.hasMoreTokens()) {
+                            peer.getConnection().getChildren().removeAll(peer.getConnection().getChildren());
```
joshmalbrecht, line 150: We can use `peer.getConnection().getChildren().clear()` here

joshmalbrecht (changes requested):
Looks good! I just noticed two small changes

src/main/java/org/candlepin/pki/huffman/Huffman.java
```diff
@@ -137,28 +137,34 @@ public void printTrie(HuffNode hn, int tab) {
 
     private void makePathForURL(StringTokenizer st, PathNode parent, PathNode endMarker) {
```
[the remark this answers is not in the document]
nikosmoum: done

src/main/java/org/candlepin/pki/huffman/Huffman.java
nikosmoum: done

joshmalbrecht (approved):

PR #5134, opened 2025-06-05
[M] HOTFIX: Added a check to prevent sorting on nested object fields

Ceiu:
- Added an emergency patch to prevent attempting to sort on fields of nested children in GET /owner/{key}/pools


joshmalbrecht (approved):

nikosmoum (approved):
I can confirm by looking at splunk that instead of 400s, we now get 200s and I can see this warning message. I also looked for other similar queries that are potentially made but I don't see any.

PR #5159, opened 2025-06-25
[F] CHAINSAW-514: Added POST /rhsmapi/consumers/entitlement_counts endpoint

joshmalbrecht:
- Added the POST /rhsmapi/consumers/entitlement_counts endpoint for retrieving entitlement count information for a set of consumers.

### Notes on RhsmApiCompatCurator.getConsumerEntitlementCounts query

I had to add the consumer.uuid column to the GROUP BY clause so that we don't need an aggregate function for consumer.uuid in the SELECT clause because that was breaking the ORDER BY clause for HSQLDB.  HSQLDB requires that the columns being ordered must be in the SELECT clause. This is a change from the query in the design.

### Performance Notes

I did some performance testing locally, and the average response time was 189.4 ms for a request containing the UUIDs and IDs to 1000 consumers that are each entitled to 1 pool.  Only around 40-50 ms of that time is the database query, and most of the remaining amount of time is spent serializing the DTOs into JSON unfortunately.


src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import javax.persistence.EntityManager;
+import javax.persistence.TypedQuery;
+
+/**
+ * Curator for retrieving persisted data for operations needed by RHSM API.
+ */
+public class RhsmApiCompatCurator {
```
bakajstep, line 38: Shouldn't this be singleton?

api/candlepin-api-spec.yaml
```diff
@@ -8669,6 +8669,61 @@ paths:
...
+      tags:
+        - rhsmapi
+      operationId: listConsumerEntitlementCounts
+      parameters:
+        - $ref: "#/components/parameters/paging_page"
+        - $ref: "#/components/parameters/paging_per_page"
+      requestBody:
+        description: |
+          Request body containing all of the consumer IDs and UUIDs to retrieve entitlement counts for. The
+          consumerIds field is deprecated and will be removed in a future version. Use the consumerUuids field
```
bakajstep, line 8684: Do we know the version? Becasue i think it would be nice to specify the version when we deleted this parametr. Same for DTO bellow.

bakajstep (commented):
I wonder if it would be faster to remove the aggregation function max over prod.name and put prod.name into GROUP BY instead?

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
joshmalbrecht: Good catch, it should be a singleton. Updated

api/candlepin-api-spec.yaml
joshmalbrecht: Unfortunately we don't know a specific version that we can explicitly list here

bakajstep (approved):

Ceiu:
I need to review this more in depth, but I have concerns over the query changes. What was causing the need for an aggregate on prod.name, and I'm fairly certain the additions to the grouping is changing the query result. Adding the UUID is probably fine since it's 1:1 with the ID anyway, but adding name to the grouping will definitely cause more segregation to occur.

joshmalbrecht:
> I need to review this more in depth, but I have concerns over the query changes. What was causing the need for an aggregate on prod.name? Also, I'm fairly certain the additions to the grouping will end up changing the query results. Adding the UUID is probably fine since it's 1:1 with the ID anyway, but adding product name to the grouping will definitely cause more segregation to occur.

HSQLDB requires `product.name` be in the GROUP BY clause or an aggregate function if it is to exist in the SELECT clause.

`org.hsqldb.HsqlException: expression not in aggregate or GROUP BY columns: PRODUCT3_.NAME`

Ceiu:
> HSQLDB requires product.name be in the GROUP BY clause or an aggregate function if it is to exist in the SELECT clause.

I suppose HSQLDB continuing to be a pain in the ass when it comes to doing actual database work shouldn't come as a surprise anymore.

The thing I don't like here is that by appeasing it, we're adding more and more assumptions to the query to get correct results -- assumptions that aren't always correct or enforc(ed/able) at the schema level. For instance, we know that product ID and product name are 1:M today, so as long as the order in which the groups appear in the GROUP BY clause, it's *probably* fine; but if that assumption were to ever change again (e.g. we go back to entity versioning), then this query breaks.

We do this already with consumer ID and consumer UUID, but those are at least guaranteed 1:1 in the current schema by way of the UUID column also being unique. But within product IDs and names, that's not necessarily true; and it's definitely not true according to the schema. For starters, the name column isn't unique, nor is the ID column without also examining the namespace. 

So this will possibly work as written, but we'll want to add some pretty specific testing around the counts and grouping to validate it. Possibly even some comments or documentation somewhere to avoid losing the reasoning for the query being written this way to time.

As an aside, one thing we can do in the future is to perform the grouping part of the query as a separate inner query that we can join against and fill out with data that isn't part of the group. It'll probably have worse performance if the query planner doesn't recognize the pattern; but it'll better convey the intent of the query than grouping on seemingly arbitrary or undesired fields.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        String jpql = "SELECT new org.candlepin.model.ConsumerEntitlementCount(consumer.id, consumer.uuid, " +
+            "pool.contractNumber, pss.subscriptionId, prod.id, prod.name, SUM(ent.quantity)) " +
+            "FROM Entitlement ent " +
+            "JOIN ent.consumer consumer " +
+            "JOIN ent.pool pool " +
+            "JOIN pool.product prod " +
+            "LEFT JOIN pool.sourceSubscription pss " +
+            "WHERE consumer.id IN (:consumerIds) OR consumer.uuid IN (:consumerUuids) " +
```
Ceiu, line 83: I would expect this to be conditionally built into the query as a conjunction rather than trying to generalize a single query for both options.

> Where "&lt;criteria&gt;" is assembled from the input components as a logical conjunction

I've updated the doc to have a clearer example of this. I imagine if we stick with JPQL, it'll end up looking similar to the ContentQueryBuilder or ProductQueryBuilder in terms of how the criteria is built.

Fun fact: I have learned that github processes &lt; and &gt; as markup. I wonder if that means you can do <b>html</b> <ul><li>directly</li></ul> in the comment...


**Edit**: Uh oh...

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -0,0 +1,109 @@
...
+
+            current.addEntitlementCountsItem(entitlementCountsItem);
+        }
+
+        // Add the last item
+        if (current != null) {
+            consumerEntCountDTOs.add(current);
+        }
+
+        if (page != null && perPage != null) {
```
Ceiu: nack on this. If we're going to do post-processing paging, we can use the pagingutil to do it for us. It also alleviates doubling up the branching and return statements.

I *really* wish we had the gather stream API, because this entire endpoint would be a perfect fit for its use case. We can approximate it with some existing bits, but I don't know if it's more efficient than what you're doing here. I imagine double iterating over potentially huge lists is slower than the iterative building of each consumer ent count DTO.

For completion, something I tinkered with a bit (but, again, I would assume to be worse performing than what you have here) is:

```
Function<ConsumerEntitlementCountRecord, RhsmApiEntitlementCountDTO> entCountTranslator = rec -> {
    return new RhsmApiEntitlementCountDTO()
        .contractNumber(rec.contractNumber())
        ...
};

Function<List<ConsumerEntitlementCountRecord>, RhsmApiConsumerEntitlementCountDTO> consumerCountTranslator = (counts) -> {
    List<RhsmApiEntitlementCountDTO> entCounts = counts.stream()
        .map(entCountTranslator)
        .toList();

    ConsumerEntitlementCountRecord rec = counts.get(0);
    return new ConsumerEntitlementCountsDTO()
        .consumerId(...)
        .consumerUuid(...)
        .entitlementCounts(entCounts);
};

Map<String, List<ConsumerEntitlementCount>> consumerEntCounts = this.rhsmApiCompatCurator
    .getConsumerEntitlementCounts(consumerIds, consumerUuids)
    .stream()
    .collect(Collectors.groupingBy(ConsumerEntitlementCount::id));

Stream<RhsmApiConsumerEntitlementCountDTO> output = consumerEntCounts.values()
    .stream()
    .map(consumerCountTranslator);

return PagingUtil.applyPaging(output, consumerEntCounts.size());
```

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -0,0 +1,109 @@
...
+            // a new ConsumerEntitlementCountsDTO for this consumer
+            if (!current.getConsumerId().equals(consumerEntCount.id())) {
+                consumerEntCountDTOs.add(current);
+
+                current = new ConsumerEntitlementCountsDTO()
+                    .consumerId(consumerEntCount.id())
+                    .consumerUuid(consumerEntCount.uuid());
+            }
+
+            ConsumerEntitlementCountsDTOAllOfEntitlementCounts entitlementCountsItem =
```
Ceiu: Is this actually what the dto generator came up with for a name here? If so, gross. We're better off defining it ourselves with something that's sane and won't force line breaking early.

`RhsmApiConsumerEntitlementCountDTO` has `RhsmApiEntitlementCountDTO`

Ceiu (changes requested):

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
nikosmoum: Why not just page on the query directly? PagingUtil still is throwing away rows fetched by the query, which is a waste

Ceiu: The query would need a *lot* more work to make that happen. It's doable, probably, but I believe you'd end up in a similar place. The big problem here is that the output is: `[consumer => [ent counts]]`, whereas the query's grouping is fundamentally a list of: `consumer ent count`, where a given consumer will have multiple entries in the output.

Since we're paging on *consumers* and not the ent counts, bumping it to the query level means juggling the following:
- consumers have to be ordered (as they are currently)
- consumers without any entitlements have to be excluded
- consumers outside of the paging range have to be excluded
- entitlement counts have to be fetched with the grouping specified in the query

I think where you end up landing with such restrictions is something approaching the SELECT+GROUP BY join-workaround using an inner query to do your grouping, and the outer query to select your targets. If you apply the consumer restrictions to the outer-most query and the query planner is smart enough to see the pattern, it may even skip the aggregation for the excluded consumers.

I can take a look at this on Monday to see if I can come up with something that would perform well enough.

Ceiu: Another concern we'll have when it comes to paging is that a proper query will not have the page nor element counts, since we'll be leaving them entirely in the DB. Depending on scale, it may end up being slower to support it properly since we'll have to do as we do in other places where we have two queries: one to get the total element count for the given criteria, and a second to fetch the specific page of elements.

Given where entitlements are going, it may not be worth the additional effort at all.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
joshmalbrecht: Updated

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
joshmalbrecht: Updated to use pagingutil

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
joshmalbrecht: Yeah, which is a pretty wild name haha `RhsmApiEntitlementCountDTO` is much better. Updated

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -0,0 +1,117 @@
...
+        this.pagingUtilFactory = Objects.requireNonNull(pagingUtilFactory);
+    }
+
+    // POST /rhsmapi/consumers/entitlement_counts
+
+    @Override
+    @Transactional
+    public Stream<ConsumerEntitlementCountsDTO> listConsumerEntitlementCounts(
+        @Valid @NotNull RhsmApiConsumerEntitlementCountsQueryDTO rhsmApiConsumerEntitlementCountsQueryDTO,
+        Integer page, Integer perPage) {
```
joshmalbrecht: We need to check that the `perPage` value is [1, 1000]

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -0,0 +1,141 @@
...
+        this.pageLimit = this.config.getInt(ConfigProperties.RHSM_API_PAGE_LIMIT);
+    }
+
+    // POST /rhsmapi/consumers/entitlement_counts
+
+    @Override
+    @Transactional
+    public Stream<ConsumerEntitlementCountsDTO> listConsumerEntitlementCounts(
+        @Valid @NotNull RhsmApiConsumerEntitlementCountsQueryDTO rhsmApiConsumerEntitlementCountsQueryDTO,
+        Integer page, Integer perPage) {
```
Ceiu: It's not necessary to explicitly define these parameters in the spec and method. Since we have to do null checks anyway, it may be better here to pull it from the req. context and only need to do special max per-page validation.

It also lets us explicitly set the sorting parameters on the shared instance instead of overwriting it as we do below.

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -0,0 +1,141 @@
...
+
+        // Add the last item
+        if (current != null) {
+            consumerEntCountDTOs.add(current);
+        }
+
+        ResteasyContext.pushContext(PageRequest.class, new PageRequest()
+            .setPage(page)
+            .setPerPage(perPage)
+            .setSortBy("consumerId")
```
Ceiu: As mentioned above, ideally only the sort and order by adjustments are necessary, and those can be performed in the same place as the validation.

Ceiu (changes requested):

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
joshmalbrecht: Updated

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
joshmalbrecht: Updated

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -0,0 +1,137 @@
...
+
+        this.rhsmApiCompatCurator = Objects.requireNonNull(rhsmApiCompatCurator);
+        this.pagingUtilFactory = Objects.requireNonNull(pagingUtilFactory);
+        this.config = Objects.requireNonNull(config);
+        this.i18n = Objects.requireNonNull(i18n);
+
+        this.pageLimit = this.config.getInt(ConfigProperties.RHSM_API_PAGE_LIMIT);
+    }
+
+    // POST /rhsmapi/consumers/entitlement_counts
```
nikosmoum: leftover by accident?

api/candlepin-api-spec.yaml
```diff
@@ -8669,6 +8669,58 @@ paths:
         default:
           $ref: '#/components/responses/default'
 
+  /rhsmapi/consumers/entitlement_counts:
+    post:
+      description: |
+        A super-admin endpoint for retrieving the entitlement count information for a set of consumers based
+        on provided Consumer IDs and UUIDs.
```
nikosmoum: Can we add a disclaimer here to say something like 'This endpoint is intended for use only by RHSM-API, and is subject to change or removal without warning.'?  This should preferably be uniform, and added to all other `/rhsmapi/` endpoints as well.

nikosmoum (changes requested):

api/candlepin-api-spec.yaml
joshmalbrecht: Good idea, and I agree that this disclaimer should be on all the RHSM API endpoints! Updated

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
joshmalbrecht: This was intentional. It is just a comment for finding the endpoint's entry point a little easier without having to look at the OpenAPI spec file. There are a couple other resource methods that do this as well.

Ceiu (approved):

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
nikosmoum: its the first time I'm seeing it. I suppose its useful, but the way it is just floating above an empty line is confusing. I initially thought this was like a TODO note trying to say that 'We'll need to add this API here in the future'. Can we instead have it directly attached to the method and use the `/** */` instead of `//`?

joshmalbrecht: Good idea! I attached it to the method by prepending it to the existing Java doc from the super class.

PR #5165, opened 2025-07-02
[M] CANDLEPIN-1028: Added event generation to the inactive consumer cleanup job

Ceiu:
- Updated the InactiveConsumerCleanerJob to generate an event for each org in which inactive consumers were removed
- Optimized several database operations
- Added a TestEventSink test utility for testing and validating events that are generated by various units

### DIFF
--- a/src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
+++ b/src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
@@ -17,25 +17,30 @@ package org.candlepin.async.tasks;
 import org.candlepin.async.AsyncJob;
 import org.candlepin.async.JobExecutionContext;
 import org.candlepin.async.JobExecutionException;
+import org.candlepin.audit.Event;
+import org.candlepin.audit.EventFactory;
+import org.candlepin.audit.EventSink;
 import org.candlepin.config.ConfigProperties;
 import org.candlepin.config.Configuration;
 import org.candlepin.model.CertificateSerialCurator;
 import org.candlepin.model.ConsumerCurator;
-import org.candlepin.model.ContentAccessCertificateCurator;
 import org.candlepin.model.DeletedConsumer;
 import org.candlepin.model.DeletedConsumerCurator;
-import org.candlepin.model.IdentityCertificateCurator;
+import org.candlepin.model.InactiveConsumerRecord;
+import org.candlepin.util.Transactional;
 
 import com.google.common.collect.Iterables;
-import com.google.inject.persist.Transactional;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 import javax.inject.Inject;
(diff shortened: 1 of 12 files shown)


src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
```diff
@@ -64,56 +69,79 @@ public class InactiveConsumerCleanerJob implements AsyncJob {
...
+     *  Arguments for this method; abstracted out due to restrictions in the transactional framework. Must
+     *  consist of a single element which is a collection of InactiveConsumerRecord instances.
+     *
+     * @return
+     *  the number of consumers that have been deleted.
      */
-    @Transactional
-    public int deleteInactiveConsumers(Collection<String> inactiveConsumerIds) {
-        if (inactiveConsumerIds == null || inactiveConsumerIds.isEmpty()) {
+    public Integer deleteInactiveConsumers(Object... args) {
```
joshmalbrecht: Should we make this private?

Ceiu: Yeah, that can be private. Was probably only public to appease the Transactional annotation.

Fixed.

joshmalbrecht: Yeah, I bet you are right!

src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
```diff
@@ -64,56 +69,79 @@ public class InactiveConsumerCleanerJob implements AsyncJob {
...
+        List<Long> serialIdsToRevoke = this.consumerCurator.getConsumerCertSerialIds(consumerIds);
 
-        int deletedConsumers = consumerCurator
-            .deleteConsumers(inactiveConsumerIds);
+        this.deletedConsumerCurator.createDeletedConsumers(consumerIds);
+        int deletedConsumers = this.consumerCurator.deleteConsumers(consumerIds);
 
         // Delete the certificates and revoke their serials.
-        identityCertificateCurator.deleteByIds(idCertsToRemove);
-        contentAccessCertificateCurator.deleteByIds(scaCertsToRemove);
```
nikosmoum, line 115: I assume these queries were removed because the deletion of the consumers themselves would take care of them, but we haven't specified ON DELETE CASCADE for these foreign keys on the consumer table. Also, afaik, the `cascade = CascadeType.REMOVE` values in the hibernate annotations on Consumer.java only work when you delete managed entities, and not if you delete through SQL or JPQL (which we are here)

Ceiu: Ahh, yeah. I believe you're correct here. I meant to go back and verify this but got lost in other work and forgot.

I'll check this again today and put up a fix either in the form of a DB change adding the delete-cascade, or restoring these deletions -- highly likely to be the former.

joshmalbrecht: Great catch!

Ceiu: I've added a PR to restore the cert deletion, since unfortunately the relationship is "backward" for the purposes of DB-level delete cascades.

https://github.com/candlepin/candlepin/pull/5172

PR #5175, opened 2025-07-17
[F] CANDLEPIN-1030: Implemented consumer feed endpoint

bakajstep:


src/main/java/org/candlepin/util/ConsumerFeedFactExtractor.java
```diff
@@ -0,0 +1,104 @@
...
+
+    // List of allowed patterns (regular expressions)
+    private static final List<Pattern> ALLOWED_PATTERNS = List.of(
+        Pattern.compile("^net\\.interface\\..+?\\.mac_address$"),
+        Pattern.compile("^net\\.interface\\..+?\\.ipv4_address$"),
+        Pattern.compile("^net\\.interface\\..+?\\.ipv6_address.global$"),
+        Pattern.compile("^net\\.interface\\..+?\\.ipv6_address.link$"),
+        Pattern.compile("^net\\.interface\\..+?\\.ipv4_address_list$"),
+        Pattern.compile("^net\\.interface\\..+?\\.ipv6_address.global_list$"),
+        Pattern.compile("^net\\.interface\\..+?\\.pv6_address.link_list$"),
```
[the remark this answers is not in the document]
bakajstep: @Ceiu This is how it is in design doc is it typo or it is wanted?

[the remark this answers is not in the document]
bakajstep: resolved typo

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,401 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+     */
+    private Map<String, Set<ConsumerFeedInstalledProduct>> getInstalledProductsByConsumer(String ownerId,
+        String afterId, String afterUuid, OffsetDateTime afterCheckin, int page, int perPage) {
+        String cte = buildConsumersCte(afterId, afterUuid, afterCheckin);
+        String sql =
+            "SELECT ip.consumer_id, ip.product_id, ip.product_name, ip.product_version " +
+                "FROM cp_installed_products ip " +
+                "JOIN (" + cte + ") cids ON cids.id = ip.consumer_id";
+
+        Query q = entityManager.createNativeQuery(sql);
```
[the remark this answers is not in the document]
bakajstep: Not a issue. User input is not put directly into query.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,401 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+     */
+    private Map<String, Map<String, String>> getFactsByConsumer(String ownerId, String afterId,
+        String afterUuid, OffsetDateTime afterCheckin, int page, int perPage) {
+        String cte = buildConsumersCte(afterId, afterUuid, afterCheckin);
+        String sql =
+            "SELECT fact.cp_consumer_id AS consumer_id, fact.mapkey, fact.element " +
+                "FROM cp_consumer_facts fact " +
+                "JOIN (" + cte + ") cids ON cids.id = fact.cp_consumer_id";
+
+        Query q = entityManager.createNativeQuery(sql);
```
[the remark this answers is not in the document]
bakajstep: Not a issue. User input is not put directly into query.

src/test/java/org/candlepin/model/RhsmApiCompatCuratorTest.java
```diff
@@ -385,6 +394,404 @@ public void testGetConsumerEntitlementCountsWithExceedingInBlockSize() {
...
+    @Test
+    public void testGetConsumerFeedFilterByOwner() {
+        Owner ownerA = createOwner("A");
+        Owner ownerB = createOwner("B");
+        Consumer consumerA1 = createConsumer(ownerA);
+        Consumer consumerA2 = createConsumer(ownerA);
+        Consumer consumerB1 = createConsumer(ownerB);
+
+        List<ConsumerFeed> result = rhsmApiCompatCurator.getConsumerFeed(ownerA, null, null, null, 1,
+            10);
```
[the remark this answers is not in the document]
bakajstep: fixed

api/candlepin-api-spec.yaml
```diff
@@ -9125,6 +9196,59 @@ components:
                 items:
                   $ref: '#/components/schemas/EnvironmentDTO'
 
+    ConsumerFeedDTO:
```
joshmalbrecht, line 9199: Missing the `lastCheckin` field

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
```diff
@@ -110,6 +111,14 @@ public static ConsumerInstalledProductDTO toInstalled(ProductDTO product) {
             .productName(product.getName());
     }
 
+    public static ConsumerFeedInstalledProductDTO toConsumerFeedInstalled(
+        ConsumerInstalledProductDTO product) {
+        return new ConsumerFeedInstalledProductDTO()
+            .productId(product.getProductId())
+            .productName(product.getProductName())
+            .productVersion(product.getVersion());
+    }
```
joshmalbrecht: This doesn't belong in the `Product.java` class because the method does not input or output a ProductDTO. We should create a new class in this package that is probably `ConsumerFeedInstalledProductDTO` and put this method in there.

api/candlepin-api-spec.yaml
```diff
@@ -8722,6 +8722,77 @@ paths:
...
           $ref: '#/components/responses/default'
 
+  /rhsmapi/owners/{org_key}/consumer_feed:
+    get:
+      description: |
+        A super-admin-only endpoint to retrieve consumer feed information for a specific owner.
+        This endpoint is deprecated and maintained only for compatibility with RHSM API.
+      tags:
+        - rhsmapi
+      operationId: getConsumerFeed
```
joshmalbrecht: I think this should be plural since we are retrieving multiple `ConsumerFeed` objects. So, `getConsumerFeeds`.

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +566,534 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
...
+
+        List<ConsumerFeedDTO> consumerFeed = adminClient.rhsmApi()
+            .getConsumerFeed(ownerKey, null, null, null, null, null);
+
+        assertThat(consumerFeed)
+            .isNotNull()
+            .hasSize(3)
+            .extracting(ConsumerFeedDTO::getId)
+            .containsExactlyInAnyOrder(consumer1.getId(), consumer2.getId(), consumer3.getId());
+
```
joshmalbrecht: Extra new line

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -136,5 +146,52 @@ public Stream<ConsumerEntitlementCountsDTO> listConsumerEntitlementCounts(
...
+    @Override
+    @Transactional
+    public Stream<ConsumerFeedDTO> getConsumerFeed(String orgKey, String afterId, String afterUuid,
+        OffsetDateTime afterCheckin, Integer page, Integer perPage) {
 
+        if (orgKey == null || orgKey.trim().isEmpty()) {
+            // TODO: We need to agree what we want to return 404 or empty list depends on what we use in
+            // rest of the RHSM API replacement
+            return Stream.empty();
+        }
```
joshmalbrecht: The `orgKey` is mandatory based on the OpenAPI spec, so we shouldn't need to have this check here.

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -136,5 +146,52 @@ public Stream<ConsumerEntitlementCountsDTO> listConsumerEntitlementCounts(
...
+        PageRequest pageRequest = ResteasyContext.getContextData(PageRequest.class);
+        if (pageRequest != null) {
+            Page<Stream<ConsumerFeedDTO>> pageResponse = new Page<>();
+            pageResponse.setPageRequest(pageRequest);
+
+            if (pageRequest.isPaging()) {
+                offset = pageRequest.getPage();
+                limit = pageRequest.getPerPage();
+                if (offset < 1) {
+                    throw new BadRequestException(i18n.tr("Parameter page must be positive integer"));
```
joshmalbrecht: Update to `Parameter page must be a positive integer`

src/main/java/org/candlepin/model/ConsumerFeed.java
```diff
@@ -0,0 +1,178 @@
...
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.model;
+
+import java.util.Map;
+import java.util.Set;
+
+public class ConsumerFeed {
```
joshmalbrecht, line 21: Can we make this a record?

src/main/java/org/candlepin/dto/api/v1/ConsumerFeedTranslator.java
```diff
@@ -0,0 +1,91 @@
...
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.dto.api.v1;
+
```
joshmalbrecht, line 17: Extra new line

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,401 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+     * @param afterUuid     (optional) only consumers with a UUID lexicographically greater than this value
+     *                      will be included
+     * @param afterCheckin  (optional) only consumers with a last check-in timestamp after this value will
+     *                      be included
+     * @param page          the page number to retrieve (1-based)
+     * @param perPage       the number of consumers to return per page
+     * @return a list of {@link ConsumerFeed} objects matching the specified filters, with facts, installed
+     *         products, add-ons, and hypervisor association (if applicable) fully populated
+     */
+    public List<ConsumerFeed> getConsumerFeed(Owner owner, String afterId, String afterUuid,
```
joshmalbrecht: Let's update this to be plural since we are returning a list of `ConsumerFeed` objects.

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -136,5 +146,52 @@ public Stream<ConsumerEntitlementCountsDTO> listConsumerEntitlementCounts(
...
+                }
+                if (limit < 1 || limit > 1000) {
```
joshmalbrecht: Instead of hard coding 1000 here, you can use `this.pageLimit`. This also makes the page limit configurable as well.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,401 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+        Map<String, HypervisorGuestMapping> result = new HashMap<>();
+        for (Object[] row : rows) {
+            String hypervisorId = (String) row[0];
+            String hypervisorUuid = (String) row[1];
+            String hypervisorName = (String) row[2];
+            String guestId = (String) row[3];
+            HypervisorGuestMapping mapping = new HypervisorGuestMapping(
+                hypervisorId, hypervisorName, hypervisorUuid, guestId);
+
+            // Since we know there will be at most 1000 consumer records, we can preload all mappings.
```
joshmalbrecht: If we make the page limit configurable, I would remove the 1000 from this comment as the page limit could be different based on the configuration.

src/test/java/org/candlepin/util/ConsumerFeedFactExtractorTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+            Map<String, String> result = ConsumerFeedFactExtractor.extractRelevantFacts(facts);
+
+            assertEquals(1, result.size());
+            assertEquals("regexval", result.get(key));
+        }
+    }
+
+    @Test
+    void ignoresNonMatchingKey() {
```
joshmalbrecht: Update the scope from package private to public. Apply throughout the class.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,401 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+                "           hypervisor.uuid AS hypervisor_uuid, " +
+                "           hypervisor.name AS hypervisor_name, " +
+                "           guest.guest_id, " +
+                "           DENSE_RANK() OVER (PARTITION BY guest.guest_id " +
+                "               ORDER BY guest.updated DESC, guest.created DESC, guest.id ASC) AS rank " +
+                "      FROM cp_consumer hypervisor " +
+                "      JOIN cp_consumer_guests guest ON guest.consumer_id = hypervisor.id " +
+                "     WHERE hypervisor.owner_id = :owner_id " +
+                ") hgmap " +
+                "WHERE hgmap.rank = 1";
```
joshmalbrecht: I'm thinking we should just use the subquery approach instead of trying to maintain two separate queries that do the same thing.  From a maintenance perspective, I think it would simplify things. In an ideal world we wouldn't have to deal with HSQLDB not accepting window functions and we could reap the performance benefit of using this query.

I'm open to thoughts on this though!

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,401 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+        }
+        if (afterCheckin != null) {
+            q.setParameter("after_checkin", afterCheckin);
+        }
+
+        Number count = (Number) q.getSingleResult();
+        return count.intValue();
+    }
+
+    private String buildConsumersCte(String afterId, String afterUuid, OffsetDateTime afterCheckin) {
```
joshmalbrecht: This function is being used as subqueries and not a CTE, so maybe we should rename this from `buildConsumersCte` to `buildConsumersSubquery`

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,401 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+        }
+        if (afterCheckin != null) {
+            q.setParameter("after_checkin", afterCheckin);
+        }
+
+        int offset = (page - 1) * perPage;
+        q.setParameter("offset", offset);
+        q.setParameter("limit", perPage);
+    }
+
```
joshmalbrecht: Extra new line

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,401 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+            q.setParameter("after_uuid", afterUuid);
+        }
+        if (afterCheckin != null) {
+            q.setParameter("after_checkin", afterCheckin);
+        }
+
+        int offset = (page - 1) * perPage;
+        q.setParameter("offset", offset);
+        q.setParameter("limit", perPage);
+    }
```
joshmalbrecht: This is a bit subjective, but I'm not the biggest fan of this method because the name is pretty vague and yet only applies or works for queries that have specific parameter names. It's more verbose, but I'm a fan of just explicitly setting the parameters in each function instead of abstracting this out.

However, if want to keep this idea, I'm wondering if we should rename this to `setPagingParameters` and remove the `q.setParameter("owner_id", ownerId)`. Then the method would only be scoped to paging parameters and we could use this method in `getConsumersFiltered` as well.

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +566,534 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
...
+            Consumers.random(owner)
+                .id(StringUtil.random("C"))
+                .uuid(StringUtil.random("C"))
+                .lastCheckin(OffsetDateTime.now().minusDays(2)));
+        ConsumerDTO consumer4 = adminClient.consumers().createConsumer(
+            Consumers.random(owner)
+                .id(StringUtil.random("D"))
+                .uuid(StringUtil.random("D"))
+                .lastCheckin(OffsetDateTime.now().minusDays(1)));
+
```
joshmalbrecht, line 703: Extra new line

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +566,534 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
...
+            .containsExactly(addOns);
+    }
+
+    @Test
+    public void shouldGetConsumerFeedWithHypervisorsAndGuests() {
+        ApiClient admin = ApiClients.admin();
+        OwnerDTO owner = admin.owners().createOwner(Owners.random());
+
+        ConsumerDTO host1 = admin.consumers().createConsumer(Consumers.random(owner));
+        String host1Guest1VirtUuid = StringUtil.random("host1-guest1-virt-uuid-");
```
joshmalbrecht, line 786: We will need spec tests that also ensure that we return guests that have a `virt.uuid` that is the reverse endianness of the host consumer's GuestId.guestId value

api/candlepin-api-spec.yaml
bakajstep: Oh, i had it initially but then i looked at example output and there is no lastCheckin field so i deleted it, but now i put it back.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
bakajstep: Yes, that makes sense, but I looked at how it was done with `ConsumerInstalledProduct`, and it's also in this class, so I put it there too. Should I also create class for the `ConsumerInstalledProduct`?

api/candlepin-api-spec.yaml
bakajstep: Done, good catch

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: Done, good catch

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
bakajstep: done, good to know

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
bakajstep: done, good catch

src/main/java/org/candlepin/dto/api/v1/ConsumerFeedTranslator.java
bakajstep: done, good catch

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: done, good catch

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
bakajstep: done, good catch

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: done, good catch

src/test/java/org/candlepin/util/ConsumerFeedFactExtractorTest.java
bakajstep: done, good catch

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: done, good catch

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: done, good catch

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: done, good catch

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: To be honest, I also debated this a bit, and I don't really like this solution either, but I didn't want to repeat the same code. I'll modify it so that I can break it down into those methods.

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: good idea, done

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
Ceiu: Eh... it's probably fine. The hard part here is that we don't have a proper translation/DTO builder framework, nor do we have access to the ProductDTO.java (since it's generated) to put it there.

It's not optimal, but outside of doing a ton of extra work, it's likely the least clunky.

src/main/java/org/candlepin/model/ConsumerFeed.java
Ceiu: The constructor for a record version of this class would the stuff of nightmares.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
joshmalbrecht: @Ceiu what are your thoughts on this? I noticed in the design doc that the performance was slightly better for the window function query, but is the extra maintenance cost of having two queries based on database type worth this performance increase?

Ceiu: The perf delta between the two is fairly small, and are an order of magnitude lower than the perf hit from JSON serialization. To that end, we're better off maintaining only a single version of this query.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,421 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+     * @param afterCheckin (optional) only consumers with a last check-in timestamp after this value will
+     *                     be included
+     * @param page         the page number to retrieve
+     * @param perPage      the number of consumers to return per page
+     * @return a map where each key is a consumer ID and each value is a list of installed products
+     * for that consumer
+     */
+    private Map<String, Set<ConsumerFeedInstalledProduct>> getInstalledProductsByConsumer(String ownerId,
+        String afterId, String afterUuid, OffsetDateTime afterCheckin, int page, int perPage) {
+        String cte = buildConsumersSubquery(afterId, afterUuid, afterCheckin);
```
Ceiu: We should update the variable naming here to better indicate what the value is. Even if the subquery were a CTE, that's not important in this context.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,421 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+        String sql =
+            "SELECT spao.consumer_id, spao.add_on " +
+                "FROM cp_sp_add_on spao " +
+                "JOIN (" + cte + ") cids ON cids.id = spao.consumer_id";
+
+        Query q = entityManager.createNativeQuery(sql);
+        q.setParameter("owner_id", ownerId);
+        if (afterId != null) {
+            q.setParameter("after_id", afterId);
+        }
```
Ceiu: Missing whitespace between the various blocks here.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,421 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+        String cte = buildConsumersSubquery(afterId, afterUuid, afterCheckin);
+        String sql =
+            "SELECT ip.consumer_id, ip.product_id, ip.product_name, ip.product_version " +
+                "FROM cp_installed_products ip " +
+                "JOIN (" + cte + ") cids ON cids.id = ip.consumer_id";
+
+        Query q = entityManager.createNativeQuery(sql);
+        q.setParameter("owner_id", ownerId);
+        if (afterId != null) {
+            q.setParameter("after_id", afterId);
```
Ceiu: These are scary -- we're setting parameters blindly that do not have any way of guaranteeing their presence. 

An example of an alternate approach is in the Product/ContentQueryBuilder, which passes a map of arguments around instead of having disconnected logic setting values.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +116,421 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+        StringBuilder hql = new StringBuilder(
+            "SELECT new org.candlepin.model.ConsumerFeed(" +
+                "c.id, c.uuid, c.name, c.typeId, c.owner.key, c.lastCheckin, c.serviceLevel, c.role)" +
+            "FROM Consumer c " +
+            "WHERE c.owner.key = :owner");
+
+        if (afterId != null) {
+            hql.append(" AND c.id > :afterId");
+        }
+        if (afterUuid != null) {
```
Ceiu: Missing whitespace between the blocks. This applies to several areas throughout this PR.

src/main/java/org/candlepin/util/ConsumerFeedFactExtractor.java
```diff
@@ -0,0 +1,104 @@
...
+            String key = entry.getKey();
+
+            // Check for an exact allowed key
+            if (ALLOWED_FACTS.contains(key)) {
+                result.put(key, entry.getValue());
+                continue;
+            }
+
+            // Check all allowed regular expression patterns
+            for (Pattern pattern : ALLOWED_PATTERNS) {
```
Ceiu: This is a *very* inefficient way to go about this for a few reasons:
- regular expressions are expensive to evaluate
- there are many expressions in the list, many of which may be collapsed into one or two
- Using `.find` instead of `.matches` implies that we're doing partial, unanchored matches, when that is not the intent behind these and could lead to bugs should someone add additional regexs that don't follow the same rules in the future.

To start, the first step here is to collapse those list of expressions down. We basically have two types: network metrics, and cloud-host facts. The latter is a simple bit of or-ing:
```
Pattern.compile("^(ocm\\.|openshift\\.|azure_|aws_|gcp_")
```

The network interfaces are very similar, but the expression gets a little spicy near the end:
```
Pattern.compile("^net\\.interface\\..+?\\.(?:mac_address|ipv4_address(?:_list)?|ipv6_address\\.(?:global|link)(?:_list)?)$")
```

Re-reading through the design and the sourced requirement, I'm also seeing some inconsistencies there which I'm guessing are oversights in the spec. That being the case, I've updated the design a bit according to my interpretation. That update would change the above regex to look as follows:
```
Pattern.compile("^net\\.interface\\..+?\\.(?:mac_address|ipv[46]_address(?:\\.global|\\.link)?(?:_list)?)$")
```

This reduces the number of expressions checked per unmatched fact from 12 down to 2. And considering the expression is anchored to static text, they should be fairly quick to execute even with the branching and optionals near the end of the expression.

Lastly, changing the matching logic from using `Matcher.find` to `Matcher.matches` should also help with performance slightly. The anchor in the expression more or less does the same thing anyway, but this would guarantee the matcher doesn't step through the provided string should someone later drop them from the expressions.

Ceiu (changes requested):
There may be some design cues we can take from the ProductQueryBuilder and ContentQueryBuilder for assembling queries w/args across several methods like this without introducing potential bugs.

src/main/java/org/candlepin/util/ConsumerFeedFactExtractor.java
Ceiu: As a side note, I generally advise against copy/paste anyway, but some blithering idiot on GitHub's UI team decided the copy button for the code segment is not only a good idea, but should be inside the content region so it covers part of it. Be very careful that you don't miss the ending of the regex that may be hiding under it.

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +566,571 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
...
+                ConsumerFeedDTO::getHypervisorUuid,
+                ConsumerFeedDTO::getHypervisorName,
+                ConsumerFeedDTO::getGuestId
+            )
+            .containsExactly(
+                tuple(host1.getUuid(), host1.getName(), host1Guest1VirtUuid),
+                tuple(host1.getUuid(), host1.getName(), host1Guest2VirtUuid),
+                tuple(host2.getUuid(), host2.getName(), host2Guest1VirtUuid),
+                tuple(host2.getUuid(), host2.getName(), host2Guest2VirtUuid)
+            );
```
joshmalbrecht, line 829: Can we also validate that the guest consumer ID and UUID is correct?

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +566,571 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
...
+            .isNotNull()
+            .hasSize(6);
+
+        List<ConsumerFeedDTO> filteredResult = actual.stream()
+            .filter(x -> x.getHypervisorUuid() != null)
+            .toList();
+
+        assertThat(filteredResult)
+            .isNotNull()
+            .hasSize(4);
```
joshmalbrecht: We can remove this since we check for not null and size starting on line 809

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +566,571 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
...
+        assertThat(filteredFeeds)
+            .isNotNull()
+            .hasSize(2)
+            .extracting(
+                ConsumerFeedDTO::getHypervisorUuid,
+                ConsumerFeedDTO::getHypervisorName
+            )
+            .containsExactly(
+                tuple(host.getUuid(), host.getName()),
+                tuple(host.getUuid(), host.getName())
```
joshmalbrecht: Can we also validate the guest Consumer UUIDs and IDs here as well

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +566,571 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
...
+
+        assertThat(feed.getGuestId()).isEqualTo(host2Guest2VirtUuid);
+        assertThat(feed.getHypervisorUuid()).isEqualTo(host2.getUuid());
+        assertThat(feed.getHypervisorName()).isEqualTo(host2.getName());
+        assertThat(feed.getFacts())
+            .containsAllEntriesOf(
+                Map.of("cpu.cpu_socket(s)", "2",
+                    "uname.machine", "x86_64",
+                    "virt.is_guest", "true")
+            );
```
joshmalbrecht: You can chain all of these assertions together:

```java
        Map<String, String> expectedFacts = Map.of("cpu.cpu_socket(s)", "2",
            "uname.machine", "x86_64",
            "virt.is_guest", "true");

        assertThat(feed)
            .isNotNull()
            .returns(host2Guest2VirtUuid, ConsumerFeedDTO::getGuestId)
            .returns(host2.getUuid(), ConsumerFeedDTO::getHypervisorUuid)
            .returns(host2.getName(), ConsumerFeedDTO::getHypervisorName)
            .returns(expectedFacts, ConsumerFeedDTO::getFacts);
```

src/main/java/org/candlepin/dto/api/v1/ConsumerFeedTranslator.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        destination.setId(source.getId());
+        destination.setUuid(source.getUuid());
+        destination.setName(source.getName());
+        destination.setTypeId(source.getTypeId());
+        destination.setOwnerKey(source.getOwnerKey());
+
+        if (source.getLastCheckin() != null) {
+            destination.setLastCheckin(source.getLastCheckin().toInstant().atOffset(ZoneOffset.UTC));
```
joshmalbrecht: We can use the `Util.toDateTime(source.getLastCheckin())` method here and then we don't have to check for null as it is checked in `Util.toDateTime`.

src/main/java/org/candlepin/model/ConsumerFeed.java
joshmalbrecht: That's true haha and it looks like we partially populating these ConsumerFeed instances in  `RhsmApiCompatCurator.getConsumersFiltered` and then mutating the rest of the fields in `RhsmApiCompatCurator.getConsumers`, so we wouldn't be able to use an immutable record here.

src/test/java/org/candlepin/resource/RhsmApiCompatResourceTest.java
```diff
@@ -218,6 +220,84 @@ public void testListConsumerEntitlementCountsWithPaging() {
...
+            null, null, null);
+
+        assertThat(consumerFeed)
+            .isNotNull()
+            .isEmpty();
+    }
+
+    @Test
+    public void testGetConsumerFeedWithUnknownOwner() {
+        Stream<ConsumerFeedDTO> consumerFeed = rhsmApiCompatResource.getConsumerFeeds("unknowKey", null, null,
```
joshmalbrecht: Typo, updated `unknowKey` to `unknownKey`

joshmalbrecht (changes requested):
Pass through the tests and addressed comments.

spec-tests/src/main/java/org/candlepin/spec/bootstrap/data/builder/Products.java
bakajstep: done

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: done

src/main/java/org/candlepin/util/ConsumerFeedFactExtractor.java
bakajstep: The first regex doesn't work with `matches` because it only matched the beginning of the fact, and `matches` must match the entire string, so it would have to look something like this:
```
Pattern.compile("^(?:ocm\\.|openshift\\.|azure_|aws_|gcp_).*$")
```
But when I was fixing it, I realized that we could actually use `startWith` and forget about regex, which should be faster.
If it's better to have the regex there, I'll gladly rewrite it back.

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: done

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: done

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: good to know, done

src/main/java/org/candlepin/dto/api/v1/ConsumerFeedTranslator.java
bakajstep: done

src/test/java/org/candlepin/resource/RhsmApiCompatResourceTest.java
bakajstep: done

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: done

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: Came up with `filteredConsumersSubquery` is it ok?

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: removed most of this logic and I hope i added whitepsaces where needed

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: That's a pretty good idea, I didn't think of that. I added a generic BuiltSql record that we can use and wrote some documentation with an example.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: removed most of this logic and I hope i added whitepsaces where needed

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +567,575 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
...
+            Consumers.random(owner).lastCheckin(OffsetDateTime.now().minusDays(1)));
+
+        List<ConsumerFeedDTO> consumerFeed = adminClient.rhsmApi()
+            .getConsumerFeeds(ownerKey, null, null, consumer2.getLastCheckin().plusMinutes(1), null, null);
+
+        assertThat(consumerFeed)
+            .isNotNull()
+            .hasSize(1)
+            .extracting(ConsumerFeedDTO::getId)
+            .containsExactlyInAnyOrder(consumer3.getId());
```
joshmalbrecht: Since there is only one element, we can avoid extracting the ID into a list and verifying that the only element in the list is `consumer3.getId()` by using `singleElement()`:

```
        assertThat(consumerFeed)
            .isNotNull()
            .singleElement()
            .returns(consumer3.getId(), ConsumerFeedDTO::getId);
```

This can apply to other methods in this class as well.

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +567,575 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
             .isEqualTo(expected);
     }
 
+    @Test
```
joshmalbrecht, line 575: Can we also add some spec tests that tests the deduplication of the guest_id mappings based on the 'updated' field?

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
```diff
@@ -555,6 +567,575 @@ public void shouldNotIncludeEntitlementCountsForPoolsThatHaveNoSubscriptionId()
...
+            .returns(host2.getUuid(), ConsumerFeedDTO::getHypervisorUuid)
+            .returns(host2.getName(), ConsumerFeedDTO::getHypervisorName)
+            .returns(expectedFacts, ConsumerFeedDTO::getFacts)
+            .returns(addons, ConsumerFeedDTO::getSyspurposeAddons)
+            .returns(Set.of(ConsumerFeedInstalledProducts.toConsumerFeedInstalled(product)),
+                ConsumerFeedDTO::getInstalledProducts);
+    }
+
+    @Test
+    public void shouldReturnGuestsWithReverseEndianVirtUuid() {
```
joshmalbrecht: Can we also verify that the guestId value for the guest `ConsumerFeedDTO` objects are equal to the `virt.uuid` fact value and not the value that is in `cp_consumer_guests`?

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: Good idea change through all file

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: done

spec-tests/src/test/java/org/candlepin/spec/RhsmApiCompatResourceSpecTest.java
bakajstep: change the logic and done

src/main/java/org/candlepin/dto/api/v1/ConsumerFeedInstalledProductTranslator.java
```diff
@@ -0,0 +1,60 @@
...
+        return this.populate(null, source, destination);
+    }
+
+    @Override
+    public ConsumerFeedInstalledProductDTO populate(ModelTranslator translator,
+        ConsumerFeedInstalledProduct source, ConsumerFeedInstalledProductDTO destination) {
+        if (source == null) {
+            throw new IllegalArgumentException("source is null");
+        }
+        if (destination == null) {
```
Ceiu, line 50: missing whitespace between the blocks

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +113,376 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+     *                      will be included
+     * @param afterUuid     (optional) only consumers with a UUID lexicographically greater than this value
+     *                      will be included
+     * @param afterCheckin  (optional) only consumers with a last check-in timestamp after this value will
+     *                      be included
+     * @param page          the page number to retrieve (1-based)
+     * @param perPage       the number of consumers to return per page
+     * @return a list of {@link ConsumerFeed} objects matching the specified filters, with facts, installed
+     *         products, add-ons, and hypervisor association (if applicable) fully populated
+     */
```
Ceiu: For the docblock, let's follow the format outlined in the coding standards doc.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
Ceiu: The missing line break still seems to be the case in the latest commit. Candlepin's style for braces allows for the condition where two disconnected blocks can appear to be a connected block at a glance, as it the case here:

```
if (afterId != null) {
  ...
}
if (afterUuid != null) {
  ...
}

-vs-

if (afterId != null) {
  ...
}
else if (afterUuid != null) {
  ...
}
```

Adding an empty line between the two helps logically disconnect them for readers:

```
if (afterId != null) {
  ...
}

if (afterUuid != null) {
  ...
}

-vs-

if (afterId != null) {
  ...
}
else if (afterUuid != null) {
  ...
}
```

This is one of the reasons why the coding standard for new projects doesn't require a line break after closing brackets, so that such blocks look more connected and this situation comes up less. I'd still recommend an extra empty line for this format anyway, but it'll be more obvious:

```
if (afterId != null) {
  ...
}
if (afterUuid != null) {
  ...
}

-vs-

if (afterId != null) {
  ...
} else if (afterUuid != null) {
  ...
}
```

Generally speaking, a reasonable amount of whitespace is a good thing, and can make code and comments easier to digest in logical chunks. When everything is crammed together, it can be very difficult for readers to break it up into its constituent parts; even with syntax highlighting.

src/main/java/org/candlepin/util/ConsumerFeedFactExtractor.java
```diff
@@ -0,0 +1,95 @@
...
+
+public class ConsumerFeedFactExtractor {
+
+    private ConsumerFeedFactExtractor() {
+        // Intentionally left empty
+    }
+
+    // List of allowed fact keys (exact matches)
+    private static final Set<String> ALLOWED_FACTS = Set.of(
+        "cpu.cpu(s)",
```
Ceiu: nit: if we're going to have hardcoded lists of facts, we should sort them so adding new entries can be done in a way that reduces the chance for dupes, and it's easier to quickly identify if a given value is in it.

src/main/java/org/candlepin/util/ConsumerFeedFactExtractor.java
```diff
@@ -0,0 +1,95 @@
...
+
+package org.candlepin.util;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+public class ConsumerFeedFactExtractor {
```
Ceiu: nit: given that this will be used for exactly one specific operation, it probably shouldn't exist as a general util. I'm not sure there's a ton of value in abstracting it into its own class, but even still, it could exist as a subclass of the consumer feeds query builder.

src/main/java/org/candlepin/util/BuiltSql.java
```diff
@@ -0,0 +1,56 @@
...
+ * Query q = entityManager.createNativeQuery(built.sql());
+ * built.args().forEach(q::setParameter);
+ *
+ * List<Object[]> rows = q.getResultList();
+ * }</pre>
+ *
+ * @param sql  the fully assembled SQL/JPQL/HQL query string, including any parameter placeholders
+ * @param args a map of named parameter placeholders to their corresponding bound values
+ */
+public record BuiltSql(String sql, Map<String, Object> args) {}
```
Ceiu: nack: this record ends up being a very weird wrapper around a partial operation. Because the SQL component is immutable, but the args aren't, we end up with this partially stateful, partially stateless value class that doesn't fully capture the mid-steps you'd have when building a query from pieces.

The operative difference between what the consumer feeds query stuff is doing and the ProductQueryBuilder and ContentQueryBuilder is that the latter are stateful builders. The collections they share amongst the internal query builders captures the state as it goes and assembles the final pieces together. The current implementation for the consumer feed queries rebuilds the intermediary step every time around a fixed wrapper for each individual query. The amount of reuse the BuiltSql wrapper provides is minimal, because its best use case as a value class is to represent a single step or chunk of the query with the arguments that fragment needs; and it only ended up being used like that out of happenstance that the outer queries didn't need any arguments of their own from what I saw.

Given the complexity in executing multiple queries with a shared discriminator, it may be apt to encapsulate all of that logic in a similar, stateful builder object which can be obtained from the RhsmApiCompatCurator, much like the ProductQueryBuilder.

Ceiu (changes requested):
There's something wrong that I can't quite put my finger on with the way the queries are built. I think it has to do with the BuiltSql wrapper and the fact we're still passing around so much of the inputs instead of encapsulating it.

Definitely need some refinement there. My first thought is to follow the general approach used by the OwnerInfoCurator and/or ProductQueryBuilder classes; where each class is purpose-built for one specific class of lookup, but one is fully stateless (OwnerInfoCurator) and the other stateful (ProductQueryBuilder). Depending on your approach, one may be preferable to the other.

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
```diff
@@ -103,4 +113,376 @@ public List<ConsumerEntitlementCount> getConsumerEntitlementCounts(Collection<St
...
+            Map<String, String> facts = factsByConsumer.get(consumerFeed.getId());
+
+            // We must match the consumer's "virt.uuid" fact against all possible endianness variants,
+            // because guest IDs can be stored in partially reversed byte order in the database.
+            // By generating all possible UUID representations, we can correctly associate the consumer
+            // with its hypervisor if a mapping exists.
+            if (facts != null) {
+                String virtUuid = facts.get("virt.uuid");
+                if (virtUuid != null && !virtUuid.isEmpty()) {
+                    consumerFeed.setGuestId(virtUuid);
```
joshmalbrecht: I know you mentioned this to me today, but I wanted to add a comment here so we wouldn't forget. The guestId should only be set when there is a hypervisor for the guest system.

bakajstep: thnak you, it is updated and even test with hypervisor checkin

src/main/java/org/candlepin/dto/api/v1/ConsumerFeedInstalledProductTranslator.java
bakajstep: done

src/main/java/org/candlepin/model/RhsmApiCompatCurator.java
bakajstep: ack, noted

src/main/java/org/candlepin/util/ConsumerFeedFactExtractor.java
bakajstep: done

src/main/java/org/candlepin/util/ConsumerFeedFactExtractor.java
bakajstep: ok moved to the builder

src/main/java/org/candlepin/util/BuiltSql.java
bakajstep: ok, i removed that and i tryied do the builder i left it in separate commit if it will be completly wrong

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
```diff
@@ -0,0 +1,542 @@
...
+        params.forEach(q::setParameter);
+
+        if (page != null && perPage != null) {
+            q.setFirstResult(Math.max(0, (page - 1) * perPage));
+            q.setMaxResults(perPage);
+        }
+
+        return q;
+    }
+
```
joshmalbrecht: Extra newline

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
```diff
@@ -0,0 +1,542 @@
...
+        List<Object[]> rows = q.getResultList();
+
+        List<T> out = new ArrayList<>(rows.size());
+        for (Object[] r : rows) {
+            out.add(rowMapper.apply(r));
+        }
+
+        return out;
+    }
+
```
joshmalbrecht, line 228: Extra new line

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
```diff
@@ -0,0 +1,542 @@
...
+        Map<String, Set<String>> addons = fetchAddOns();
+        Map<String, Set<ConsumerFeedInstalledProduct>> ips = fetchInstalledProducts();
+        Map<String, HypervisorGuestMapping> hypMap = fetchHypervisorGuestMap();
+
+        for (ConsumerFeed feed : feeds) {
+            Map<String, String> consumerFacts = facts.get(feed.getId());
+            if (consumerFacts != null) {
+                String virtUuid = consumerFacts.get("virt.uuid");
+                if (virtUuid != null && !virtUuid.isBlank()) {
+                    for (String v : Util.getPossibleUuids(virtUuid)) {
```
joshmalbrecht: Can we rename the `v` variable to something like `possibleUuid` so that it is more clear that variable is?

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
bakajstep: done

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
bakajstep: done

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
bakajstep: done

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
```diff
@@ -136,5 +147,50 @@ public Stream<ConsumerEntitlementCountsDTO> listConsumerEntitlementCounts(
...
+                limit = pageRequest.getPerPage();
+                if (offset < 1) {
+                    throw new BadRequestException(i18n.tr("Parameter page must be a positive integer"));
+                }
+                if (limit < 1 || limit > this.pageLimit) {
+                    throw new BadRequestException(i18n.tr("Parameter per_page must be in range 1-1000"));
+                }
+            }
+
+            pageResponse.setMaxRecords(Math.toIntExact(consumerFeedQuery.getResultCount()));
```
joshmalbrecht, line 185: Extra new line

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
```diff
@@ -0,0 +1,540 @@
...
+     *
+     * @param rowMapper
+     *  maps a raw {@code Object[]} row into type {@code T}
+     *
+     * @param <T>
+     *  mapped result element type
+     *
+     * @return list of mapped results
+     */
+    private <T> List<T> runNative(String baseSql, Function<Object[], T> rowMapper) {
```
joshmalbrecht: Nit (non-blocker and subjective): You don't have to make this change if you don't want to, but I would update the name of this method because the name is a bit misleading in my opinion. It does run a native query, but the purpose is to run a native query that contains the consumer ID subquery.

**Possible name ideas:**
- executeWithConsumerIdSubquery()
- runWithConsumerIdSubquery()

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
```diff
@@ -0,0 +1,540 @@
...
+    }
+
+    /**
+     * Counts consumers matching the current filter set using the memoized subquery.
+     *
+     * @return total count of matching consumers
+     */
+    @Override
+    public long getResultCount() {
+        ensureConsumerIdSubquery();
```
joshmalbrecht: The problem here is that your fields like `afterId`, `afterUuid`, `afterCheckin`, `page`, and `perPage` are all mutable and `ensureConsumerIdSubquery` will return `consumerIdSubquerySql` if it is not null. So, you could call `ConsumerFeedQueryBuilder.getResultCount()` then update a field like `afterId` using the `setAfterId` method, and then call `ConsumerFeedQueryBuilder.getResultCount()` again, and you would get the same result as the first method invocation because `consumerIdSubquerySql` has been built from the first invocation.

I believe that this will be a problem for `ConsumerFeedQueryBuilder.getResultList` as well.

joshmalbrecht: The more I look at this, the more I think that we should encapsulate this Consumer ID subquery and args in a private inner class with getter methods for the SQL string and the args. We could create an instance of this `ConsumerIdSubquery` in `ConsumerFeedQueryBuilder.getResultCount()` and `ConsumerFeedQueryBuilder.getResultList`, but we could update the method signature for methods that depend on the subquery so that we can pass in our instance. For example: `private Map<String, Set<String>> fetchAddOns(ConsumerIdSubquery subquery)`.

Let me know what you think! There may be a better way to approach this as well.


Example (may need some tweaking): 
```java
    private class ConsumerIdSubquery {

        private String consumerIdSubquerySql;
        private Map<String, Object> consumerIdSubqueryArgs;

        public ConsumerIdSubquery(String afterId, String afterUuid, String afterCheckin, Integer page, Integer perPage) {
            StringBuilder sb = new StringBuilder(
                "SELECT id FROM cp_consumer WHERE owner_id = :owner_id");
            Map<String, Object> args = new HashMap<>();
            args.put("owner_id", ownerId);

            if (afterId != null) {
                sb.append(" AND id > :after_id");
                args.put("after_id", afterId);
            }

            if (afterUuid != null) {
                sb.append(" AND uuid > :after_uuid");
                args.put("after_uuid", afterUuid);
            }

            if (afterCheckin != null) {
                sb.append(" AND lastcheckin > :after_checkin");
                args.put("after_checkin", afterCheckin);
            }

            sb.append(" ORDER BY id ASC, lastcheckin ASC");

            if (page != null && perPage != null) {
                sb.append(" LIMIT :limit OFFSET :offset");
                args.put("limit", perPage);
                args.put("offset", Math.max(0, (page - 1) * perPage));
            }

            this.consumerIdSubquerySql = sb.toString();
            this.consumerIdSubqueryArgs = Collections.unmodifiableMap(args);
        }

        public String getQuery() {
            return consumerIdSubquerySql;
        }

        public Map<String, Object> getArgs() {
            return consumerIdSubqueryArgs;
        }

    }
```

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
```diff
@@ -0,0 +1,540 @@
...
+     */
+    private Map<String, Set<ConsumerFeedInstalledProduct>> fetchInstalledProducts() {
+        Map<String, Set<ConsumerFeedInstalledProduct>> map = new HashMap<>();
+
+        String sql =
+            "SELECT ip.consumer_id, ip.product_id, ip.product_name, ip.product_version " +
+            "FROM cp_installed_products ip " +
+            "JOIN (:CONSUMER_SUBQUERY) cids ON cids.id = ip.consumer_id";
+
+        runNative(sql, r -> r).forEach(r -> {
```
Ceiu: Few things here:
- the `r -> r` implies that nothing is being done there, calling into question if a mapping function is needed at all. In all of the uses, it doesn't appear to be, so it could/should be dropped
- `Functions.identity()` can be used in place of `r -> r` in a way that may be more performant, as it's a constant value. I'm unsure if the compiler is smart enough to identify this and automatically sub in the identity function, but last I've read/researched, it does not.
- Generally lambdas should not modify the a map outside of it. Given the map isn't modified multiple times, there's no real need to introduce the pattern in this case. This could either be unrolled into a regular for-each loop, or a stream collector to eliminate the possibility that this becomes a parallel map builder which is subject to race conditions during modification.
- nit: Single-character variable names are not great for readability, especially in this case where it's not immediately obvious what 'r' represents. "row" would be better here, IMO.

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
```diff
@@ -0,0 +1,540 @@
...
+
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+import javax.persistence.TypedQuery;
+
+/**
+ * Purpose-built, stateful query builder that represents the end-to-end
+ * construction of the “consumer feed” view.
+ */
+public class ConsumerFeedQueryBuilder extends QueryBuilder<ConsumerFeedQueryBuilder, ConsumerFeed> {
```
Ceiu, line 41: Let's start getting in the habit of following the new coding standard in places the legacy Candlepin standard hasn't defined, such as fully qualifying variable references with `this.` or `ClassName.` as appropriate. Especially in cases where there are many global or static references.

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
```diff
@@ -0,0 +1,540 @@
...
+     * Builds and memoizes the shared native subquery selecting consumer IDs
+     * based on configured filters. The subquery is referenced by
+     * dependent native queries via the {@code :CONSUMER_SUBQUERY} placeholder.
+     * <p>
+     * The SQL and its argument map are cached to avoid repeated rebuilding.
+     * </p>
+     */
+    private void ensureConsumerIdSubquery() {
+        if (this.consumerIdSubquerySql != null) {
+            return;
```
Ceiu: This block is dangerous. The parameters that make up the subquery are mutable, but the query itself is only ever built once. Call order becomes very important to getting a correct result, but that's not only very dangerous for maintenance, it's also not stated anywhere in the docs nor a comment.

Honestly, though, a better solution would be to not do this at all. Memoizing some state is fine, but the whole point of a class like this is to statefully collect input parameters, and then do something with that as one atomic-ish operation. This query should be built for every time an operation is executed, *or* the mutators need to start rejecting changes once the intermediary state is built and it's no longer valid to do so.

Additionally, building a string that needs to then be ripped apart and rebuilt isn't the most efficient way of approaching this problem. You lose a lot of the gains you were to get by pre-building the string.

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
```diff
@@ -0,0 +1,540 @@
...
+            .setParameter("owner_id", ownerId)
+            .getResultList();
+
+        Map<String, HypervisorGuestMapping> out = new HashMap<>();
+        for (Object[] row : rows) {
+            String hypUuid = (String) row[0];
+            String hypName = (String) row[1];
+            String guestId = (String) row[2];
+            HypervisorGuestMapping m = new HypervisorGuestMapping(hypName, hypUuid, guestId);
+            Util.getPossibleUuids(guestId).forEach(v -> out.put(v.toLowerCase(), m));
```
Ceiu: Single-character variables again. I legitimately didn't know what v or m are in this case initially.

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
Ceiu: While this solution does address the mutability problem *and* does keep the efficiency of avoiding constantly rebuilding the inner query, something still feels off. This still implies that the inner query is the only thing that will ever receive parameters. Or, at least, forces the owner of each instance to combine maps in a clunky way after the fact.

The map really should be the state collector, because all we really want is a query chunk as a string and the parameters that must go in the query as a whole.

src/main/java/org/candlepin/resource/RhsmApiCompatResource.java
bakajstep: done

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
bakajstep: removed

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
bakajstep: I hope I got it right with how I edited it

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
bakajstep: ack

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
bakajstep: ack hopefully i did not miss anything

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
bakajstep: removed and building it everytime

src/main/java/org/candlepin/model/ConsumerFeedQueryBuilder.java
bakajstep: ok, hopefully removed everything

PR #5183, opened 2025-07-22
[M] CANDLEPIN-1037: Fixed consumer env priority database column type

Ceiu:
- Updated the column type for the priority field of consumer environments to avoid issues with schema validation and potential integer overflow when storing new environments


Ceiu:
To test that this will unblock Sat/Foreman:

1. Verify the current state by deploying main
2. Update candlepin.conf to enable Hibernate's schema validation:
```
jpa.config.hibernate.hbm2ddl.auto=validate
```
3. Restart/redeploy Candlepin and verify that it no longer completes startup. tomcat/catalina.log and Tomcat's logging in journalctl should have some error involving concurrent modification exception.
4. Update tomcat's `logging.properties` to disallow async file access to avoid the problem with the previous exception. Replace all instances of "AsyncFileHandler" with "FileHandler". In vi[m] this can be accomplished with a simple find/replace command:
```
:%s/AsyncFileHandler/FileHandler/g
```
5. Restart/redeploy Candlepin again. It should still fail to deploy, but this time tomcat/localhost.log will have a few rather odd exceptions. In the second exception, it should fail with a `SchemaManagementException` indicating the size of the priority field does not match the expected size (smallint vs int)
6. Check out this branch and deploy. Candlepin should startup as normal.
7. The database column should be `INTEGER NOT NULL` on both PostgreSQL and MariaDB/MySQL

PR #5186, opened 2025-07-29
[M] CANDLEPIN-1039: Fixed dangling Liquibase connections

Ceiu:
- Updated the migration manager to use try-with-resources for its database connection to ensure it's closed when the migration operation is completed
- Renamed and moved the not-generic "DatabaseConnectionManager" to the Liquibase-specific "LiquibaseConnectionGenerator"


joshmalbrecht (approved):
Looks good!

PR #5196, opened 2025-08-13
[M] CANDLEPIN-1045: Disallow running sonar from a fork PR

joshmalbrecht:
- Removed the .github/workflows/sonar.yml file.
- Updated the unit_tests.yml workflow to include using sonarqube cloud if the branch from the PR is not from a forked repository.

I added the Sonarqube analysis step to the unit tests workflow because the Sonarqube step is dependent on the reports that are generated from the unit tests and now they are both run on the 'pull_request' event.

**Example of a PR from a forked repository:**
https://github.com/candlepin/candlepin/pull/5195

You can see that the 'Analyze' step is skipped and the sonarqube is not run.


.github/workflows/unit_tests.yml
```diff
@@ -45,19 +51,28 @@ jobs:
...
-
       - name: Archive reports
         uses: actions/upload-artifact@v4
         with:
           name: unit_test_reports
           path: |
             **/build/**/jacoco*.xml
             **/build/classes/
+
+      - if: github.event.pull_request.head.repo.fork == false
```
bakajstep: I suggested this solution at the beginning, but Nikos wants to have the sonar in a separate file.

And don't we want to run the sonar only if the unit tests pass? Because if that condition is missing here

.github/workflows/unit_tests.yml
```diff
@@ -45,19 +51,28 @@ jobs:
...
-      - name: Save PR number to file
-        run: echo ${{ github.event.number }} > PR_NUMBER.txt
-
-      - name: Archive PR number
-        uses: actions/upload-artifact@v4
-        with:
-          name: pr_number
-          path: PR_NUMBER.txt
-
       - name: Archive reports
```
bakajstep, line 48: note: If we would use this approach this step would not be needed, because it is for sonar and if we are running sonar in same runner the files are there.

.github/workflows/unit_tests.yml
joshmalbrecht: Oh interesting, I don't think it makes sense to put the sonarqube steps in its own workflow because it is directly dependent on the unit tests. I'll put this card back in 'To-do' until Nikos gets back next week to see what he thinks.

You are right on adding the condition if the unit tests step passes! I'll add that check.

.github/workflows/unit_tests.yml
joshmalbrecht: Good catch! Removed this step

.github/workflows/unit_tests.yml
joshmalbrecht: Added the `${{ success() }}` conditional so this step will only run if the previous steps were successful

joshmalbrecht: @nikosmoum are you okay with running the SonarQube task in the unit test workflow instead of creating a whole separate workflow for SonarQube?

I don't think we can even have SonarQube in its own separate workflow because the SonarQube task is dependent on the unit test output and you would have to use the `workflow_run` webhook event to ensure the workflow runs after the unit tests workflow, and the whole point of this task is to not use the `workflow_run` webhook event.

.github/workflows/unit_tests.yml
```diff
@@ -45,19 +51,20 @@ jobs:
...
-
-      - name: Archive reports
-        uses: actions/upload-artifact@v4
-        with:
-          name: unit_test_reports
-          path: |
-            **/build/**/jacoco*.xml
-            **/build/classes/
+      - if: ${{ success() }} && github.event.pull_request.head.repo.fork == false
+        name: Analyze
```
nikosmoum: Doesn't the `name` field need to be on the first line, and the `if` field under it (unsure if this is important or not, but at the very least, it would be more consistent/readable)?

.github/workflows/unit_tests.yml
```diff
@@ -45,19 +51,20 @@ jobs:
...
-        run: echo ${{ github.event.number }} > PR_NUMBER.txt
-
-      - name: Archive PR number
-        uses: actions/upload-artifact@v4
-        with:
-          name: pr_number
-          path: PR_NUMBER.txt
-
-      - name: Archive reports
-        uses: actions/upload-artifact@v4
```
nikosmoum, line 58: Does this still work without having to upload/download the artifact? Does the sonar find them automatically?

.github/workflows/unit_tests.yml
nikosmoum: It's fine to keep it in the same file as unit tests, but we should at least split it into its own job (a separate job under `jobs`, after the  `unit_tests` job)

nikosmoum: example here: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-jobs#example-requiring-successful-dependent-jobs
Note that you'll need to use upload-artifact and download-artifact actions to share the artifacts between the 2 jobs

.github/workflows/unit_tests.yml
bakajstep: yes, because it is running on one machine. This step would be needed only when we have it in separate job, becasue each job get new environment

.github/workflows/unit_tests.yml
joshmalbrecht: Updated

.github/workflows/unit_tests.yml
joshmalbrecht: That makes sense! Updated

.github/workflows/unit_tests.yml
```diff
@@ -45,19 +51,37 @@ jobs:
...
-          path: PR_NUMBER.txt
-
       - name: Archive reports
         uses: actions/upload-artifact@v4
         with:
           name: unit_test_reports
           path: |
             **/build/**/jacoco*.xml
             **/build/classes/
+  sonarqube:
```
nikosmoum, line 55: Does this job not also need the other steps that other jobs have, like 'Mask secrets', 'Check out repository', 'Set Environment Variables', 'Rebase', 'Set up Java', 'Setup Gradle' ?

joshmalbrecht: Yeah, that was my bad, I forgot to add those steps! Updated and now the job is working properly.

.github/workflows/unit_tests.yml
```diff
@@ -45,19 +45,70 @@ jobs:
...
+          java-version: ${{ env.JAVA_VERSION }}
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
+
+      - name: Download unit test report
+        uses: actions/download-artifact@v5
+        with:
+          name: unit_test_reports
+      - name: Analyze
```
nikosmoum, line 99: nitpick: can we put an empty line between this and the previous step to be more consistent with the rest of the steps?

joshmalbrecht: Updated

PR #5203, opened 2025-08-26
CANDLEPIN-1050: Fixed constraint violation in InactiveConsumerCleanerJob

bakajstep:
- Previously, when the job attempted to delete a consumer whose UUID already existed in `cp_deleted_consumers`, it failed with a unique constraint violation. The logic now uses an upsert to handle already-deleted consumers safely.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobSpecTest.java
@@ -15,6 +15,7 @@
 package org.candlepin.spec.jobs;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.candlepin.spec.bootstrap.assertions.JobStatusAssert.assertThatJob;
 import static org.candlepin.spec.bootstrap.assertions.StatusCodeAssertions.assertGone;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
@@ -42,11 +43,13 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.time.Instant;
+import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.UUID;
 
 @SpecTest
 public class InactiveConsumerCleanerJobSpecTest {
@@ -106,7 +109,8 @@ public class InactiveConsumerCleanerJobSpecTest {
 
         String jobId = jobsClient.scheduleJob(JOB_KEY).getId();
         AsyncJobStatusDTO status = jobsClient.waitForJob(jobId);
-        assertEquals("FINISHED", status.getState());
+        assertThatJob(status)
+            .isFinished();
 
         // Verify that the inactive consumer has been deleted.
         final String inactiveConsumer1Uuid = inactiveConsumer.getUuid();
@@ -134,6 +138,65 @@ public class InactiveConsumerCleanerJobSpecTest {
         compareConsumers(consumerWithEntitlement, consumerApi.getConsumer(consumerWithEntitlement.getUuid()));
         compareConsumers(activeConsumer, consumerApi.getConsumer(activeConsumer.getUuid()));
         compareConsumers(activeConsumer2, consumerApi.getConsumer(activeConsumer2.getUuid()));
+    }
+
+    @Test
+    public void shouldUpdateDeletedConsumerWhenUuidReused() {
+        final String uuid = UUID.randomUUID().toString();
+
+        Instant inactiveTime = Instant.now()
+            .minus(DEFAULT_LAST_CHECKED_IN_RETENTION_IN_DAYS + 10, ChronoUnit.DAYS);
+
+        ConsumerDTO c1 = Consumers.random(owner)
+            .uuid(uuid)
+            .lastCheckin(inactiveTime.atOffset(ZoneOffset.UTC));
+        c1 = consumerApi.createConsumer(c1, null, owner.getKey(), null, false);
+
+        // Delete Consumer1 (simulate previous deletion in the past)
+        consumerApi.deleteConsumer(c1.getUuid());
+
+        // Capture the first deletion record (must exist now)
+        List<DeletedConsumerDTO> afterFirstDeletion = deletedConsumerApi
+            .listByDate(c1.getCreated(), 1, 200, "desc", "created");
+        List<DeletedConsumerDTO> firstDeletionForUuid = afterFirstDeletion.stream()
+            .filter(dc -> uuid.equals(dc.getConsumerUuid()))
+            .toList();
+
+        assertThat(firstDeletionForUuid)
+            .hasSize(1);
+
+        OffsetDateTime firstDeletedCreated = firstDeletionForUuid.get(0).getCreated();
+
+        // create Consumer2 with the SAME UUID, make it inactive
+        ConsumerDTO c2 = Consumers.random(owner)
+            .uuid(uuid)
+            .lastCheckin(inactiveTime.atOffset(ZoneOffset.UTC));
+        c2 = consumerApi.createConsumer(c2, null, owner.getKey(), null, false);
+
+        String jobId = jobsClient.scheduleJob(JOB_KEY).getId();
+        AsyncJobStatusDTO status = jobsClient.waitForJob(jobId);
+        assertThatJob(status)
+            .isFinished();
+
+        // ensure only ONE row exists for this UUID and it reflects the 2nd deletion
+        List<DeletedConsumerDTO> afterSecondDeletion = deletedConsumerApi
+            .listByDate(firstDeletedCreated.minusDays(1), 1, 200, "desc", "created");
+        List<DeletedConsumerDTO> rowsForUuid = afterSecondDeletion.stream()
+            .filter(dc -> uuid.equals(dc.getConsumerUuid()))
+            .toList();
+
+        assertThat(rowsForUuid)
+            .isNotNull()
+            .singleElement()
+            .returns(c2.getName(), DeletedConsumerDTO::getConsumerName)
+            .returns(c2.getOwner().getId(), DeletedConsumerDTO::getOwnerId)
+            .returns(c2.getOwner().getKey(), DeletedConsumerDTO::getOwnerKey)
+            .returns(c2.getOwner().getDisplayName(), DeletedConsumerDTO::getOwnerDisplayName)
+            .extracting(DeletedConsumerDTO::getCreated)
+            .satisfies(created -> assertThat(created).isAfterOrEqualTo(firstDeletedCreated));
+
+        // And verify that the live consumer is now gone
+        assertGone(() -> consumerApi.getConsumer(uuid));
     }
 
     private void createProductAndPoolForConsumer(ConsumerDTO consumer) {
--- a/src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
+++ b/src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
@@ -26,7 +26,6 @@ import org.candlepin.model.CertificateSerialCurator;
 import org.candlepin.model.ConsumerCurator;
 import org.candlepin.model.ContentAccessCertificateCurator;
 import org.candlepin.model.DeletedConsumer;
-import org.candlepin.model.DeletedConsumerCurator;
 import org.candlepin.model.IdentityCertificateCurator;
 import org.candlepin.model.InactiveConsumerRecord;
 import org.candlepin.util.Transactional;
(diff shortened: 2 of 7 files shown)


src/test/java/org/candlepin/model/DeletedConsumerCuratorTest.java
```diff
@@ -218,4 +223,89 @@ public void testCreateDeletedConsumersWithExistingConsumers() {
...
+
+    private void assertDeletedConsumer(Consumer consumer, DeletedConsumer deletedConsumer) {
+        assertThat(deletedConsumer)
+            .isNotNull()
+            .returns(consumer.getOwner().getId(), DeletedConsumer::getOwnerId)
+            .returns(consumer.getUuid(), DeletedConsumer::getConsumerUuid)
+            .returns(consumer.getName(), DeletedConsumer::getConsumerName)
+            .returns(consumer.getOwner().getDisplayName(), DeletedConsumer::getOwnerDisplayName)
+            .returns(consumer.getOwner().getKey(), DeletedConsumer::getOwnerKey);
+    }
```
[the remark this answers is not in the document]
bakajstep: Principal and dates are not present on consumer entity

src/test/java/org/candlepin/model/DeletedConsumerCuratorTest.java
```diff
@@ -218,4 +223,89 @@ public void testCreateDeletedConsumersWithExistingConsumers() {
         assertEquals(consumer2.getOwner().getDisplayName(), deleteConsumer2.getOwnerDisplayName());
         assertEquals(consumer2.getOwner().getKey(), deleteConsumer2.getOwnerKey());
     }
+
+    @Test
+    public void testUpsertInsertsWhenNotExists() {
```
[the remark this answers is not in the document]
bakajstep: done

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -117,6 +121,98 @@ public int countByConsumerUuid(String uuid) {
...
+
+        // All uuids that needs to be processed
+        List<String> uuids = new java.util.ArrayList<>(rows.stream().map(r -> (String) r[1]).toList());
+
+        // Get existing uuids from the DeletedConsumer table
+        String deletedConsumerUuidsJpql =
+            "SELECT dc.consumerUuid " +
+            "FROM DeletedConsumer dc " +
+            "WHERE dc.consumerUuid IN (:uuids)";
+        Set<String> existingUuids = new HashSet<>(
```
Ceiu: The `Set` wrapper is unnecessary here, and it results in some extra memory util/alloc and iteration. If uniqueness is required, update the query to use DISTINCT such that it cannot return duplicates.

I'm guessing this is an optimization to deal with all the `.contains` that come later? At face value that makes sense, but it feels to me like there's a better approach somewhere.

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -117,6 +121,98 @@ public int countByConsumerUuid(String uuid) {
...
+            return 0;
+        }
+
+        Principal principal = this.principalProvider.get();
+        String principalName = (principal == null ? null : principal.getName());
+        OffsetDateTime now = OffsetDateTime.now();
+
+        // Get information about existing consumers to populate DeletedConsumer
+        String consumerJpql =
+            "SELECT c.id, c.uuid, c.ownerId, o.displayName, o.key, c.name " +
```
Ceiu: nit: don't use single-character aliases; try to use at least two or three characters to make it easier to quickly refactor bits of code using language-agnostic tooling (find/replace, multi-select).

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -117,6 +121,98 @@ public int countByConsumerUuid(String uuid) {
...
+        Set<String> existingUuids = new HashSet<>(
+            this.getEntityManager().createQuery(deletedConsumerUuidsJpql, String.class)
+                .setParameter("uuids", uuids)
+                .getResultList()
+        );
+
+        int updated = 0;
+        String updateJpql =
+            "UPDATE DeletedConsumer d " +
+            "   SET d.consumerName      = :consumerName, " +
```
Ceiu: nit: nack on this type of formatting. It very quickly falls out of alignment when things change or are rearranged (as has happened on 172)

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -117,6 +121,98 @@ public int countByConsumerUuid(String uuid) {
...
+            }
+
+            updateQ.setParameter("consumerName", (String) r[5])
+                .setParameter("ownerId", (String) r[2])
+                .setParameter("ownerDisplayName", (String) r[3])
+                .setParameter("ownerKey", (String) r[4])
+                .setParameter("principalName", principalName)
+                .setParameter("now", Util.toDate(now))
+                .setParameter("uuid", uuid);
+            updated += updateQ.executeUpdate();
```
Ceiu: This will run the query once per entry that exists, and since Hibernate tends to not batch updates in CP, this will end up running as a fully independent query for each duplicate.

Updates from another table aren't supported in JPQL, but I believe correlated subqueries and bulk deletions *are*. Given what we're doing here, a delete of the old record before inserting the new is fundamentally the same as updating an existing since we lose record of the old deletion anyway.

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -117,6 +121,98 @@ public int countByConsumerUuid(String uuid) {
...
+            inserted = this.getEntityManager()
+                .createQuery(insertJpql)
+                .setParameter("now", Util.toDate(now))
+                .setParameter("principalName", principalName)
+                .setParameter("uuids", uuids)
+                .executeUpdate();
+        }
+
+        this.flush();
+        this.clear();
```
Ceiu: These two lines should almost never appear on the production side of things. Something else is going very wrong if these are necessary to fix an issue you're seeing.

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -117,6 +121,98 @@ public int countByConsumerUuid(String uuid) {
             .intValue();
     }
 
+    @Transactional
+    public int upsertDeletedConsumers(Collection<String> consumerIds) {
```
Ceiu: nit: I'm not a huge fan of using the term "upsert" as a method verb. I understand why you chose it (it's fundamentally what the implementation does), but it feels like jargon at a level of abstraction where the details aren't supposed to be surfaced.

I think given the work being done here, this may make sense to merge into the actual deleteConsumers operation from an API perspective. As it is now, proper bulk deletion involves calling this method *before* calling the actual deleteConsumers bulk method. Failing to do one without the other will cause data integrity issues, which implies they're both part of the same logical operation. This method could still be retained as a private implementation detail of the delete operation, but probably shouldn't be surfaced as an independent op.

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -117,6 +121,98 @@ public int countByConsumerUuid(String uuid) {
             .intValue();
     }
 
+    @Transactional
```
Ceiu: nack: individual methods should not be marked as transactional. Transaction boundaries should be defined by the scope of an externally facing operation, such as a job or API request. 

I know we still have a number of these throughout the code base, but the goal is to eventually eliminate them all in favor of transaction boundaries as close to user operations as possible.

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -117,6 +121,98 @@ public int countByConsumerUuid(String uuid) {
...
+            updated += updateQ.executeUpdate();
+            uuids.remove(uuid);
+        }
+
+        int inserted = 0;
+        if (!uuids.isEmpty()) {
+            String insertJpql =
+                "INSERT INTO DeletedConsumer (id, created, updated, consumerUuid, ownerId, " +
+                "ownerDisplayName, ownerKey, principalName, consumerName) " +
+                "SELECT c.id, :now, :now, c.uuid, c.ownerId, o.displayName, o.key, :principalName, c.name " +
```
Ceiu: There's short-hand for now() as a constant/function that can be used so it doesn't need to be parameterized like this. The only reason we'd want to parameterize the timestamps is to ensure every entry uses the exact same timestamp, but in such a case you'd need to also update the parameter to use a fixed value rather than one which is generated repeatedly as is done on line 204.

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
bakajstep: ok, noted

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
bakajstep: ok, noted

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
bakajstep: ok, noted

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
bakajstep: ok, noted

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
bakajstep: ok, noted

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
bakajstep: ok, noted

src/main/java/org/candlepin/async/tasks/InactiveConsumerCleanerJob.java
```diff
@@ -141,6 +141,7 @@ private Integer deleteInactiveConsumers(Object... args) {
         List<String> scaCertsToRemove = consumerCurator.getContentAccessCertIds(consumerIds);
         List<Long> serialIdsToRevoke = this.consumerCurator.getConsumerCertSerialIds(consumerIds);
 
+        this.deletedConsumerCurator.deleteByConsumerIds(consumerIds);
         this.deletedConsumerCurator.createDeletedConsumers(consumerIds);
         int deletedConsumers = this.consumerCurator.deleteConsumers(consumerIds);
```
Ceiu, line 140: I think something got lost from the last review. These three lines should become a single line:
```
int deletedConsumers = this.consumerCurator.deleteConsumers(consumerIds);
```

The management of the deleted consumers entries is an implementation detail of the delete operation. Callers should not be expected to have to know how to manage that.

bakajstep: Good idea, I also move deleting of the `DeletedConsumer` to createDeletedConsumer method. Initialy that move seems wrong to me but it is how our creation should work so move it there. If you think it is wrong I will move it to delete method for consumer.

spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobSpecTest.java
```diff
@@ -136,6 +138,64 @@ public void shouldDeleteInactiveConsumers() {
...
+
+        // create Consumer2 with the SAME UUID, make it inactive
+        ConsumerDTO c2 = Consumers.random(owner)
+            .uuid(uuid)
+            .lastCheckin(inactiveTime.atOffset(ZoneOffset.UTC));
+        c2 = consumerApi.createConsumer(c2, null, owner.getKey(), null, false);
+
+        String jobId = jobsClient.scheduleJob(JOB_KEY).getId();
+        AsyncJobStatusDTO status = jobsClient.waitForJob(jobId);
+        assertEquals("FINISHED", status.getState());
```
joshmalbrecht: You can use `assertThatJob(status).isFinished();` here so that we are not hard coding the "FINISHED" status

joshmalbrecht (changes requested):
Just noticed a small change!

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
```diff
@@ -122,20 +122,58 @@ public int createDeletedConsumers(Collection<String> consumerIds) {
...
+    }
+
+    public int deleteByConsumerIds(Collection<String> consumerIds) {
+        if (consumerIds == null || consumerIds.isEmpty()) {
+            return 0;
+        }
+
+        int deleted = 0;
+
+        String hql =
```
Ceiu: nit: this is JPQL, not HQL. The difference is almost entirely semantics for our use case, but we're not using Hibernate-specific features in the query (as far as I know), only the JPQL subset.

Ceiu (approved):

spec-tests/src/test/java/org/candlepin/spec/jobs/InactiveConsumerCleanerJobSpecTest.java
bakajstep: done

src/main/java/org/candlepin/model/DeletedConsumerCurator.java
bakajstep: ack, done

PR #5227, opened 2025-09-16
[M] CANDLEPIN-1065: Added bulk migration event

joshmalbrecht:
- Added a new BULK_MIGRATION audit event for when multiple entities are moved to a target entity.
- Updated the Event class to include an "anonymousOwner" boolean field.
- Updated ConsumerMigrationJob to emit a new bulk consumer migration event for each batch of migrated consumers.
- Updated the Owned interface to include an "isOwnerAnonymous" method.
- Updated EventBuilder to set the event's anonymousOwner field for all Owned entities.
- Updated NestedOwnerDTO to include an "anonymous" boolean field.

### DIFF
--- a/src/main/java/org/candlepin/controller/ConsumerMigration.java
+++ b/src/main/java/org/candlepin/controller/ConsumerMigration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009 - 2024 Red Hat, Inc.
+ * Copyright (c) 2009 - 2025 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -15,6 +15,9 @@
 
 package org.candlepin.controller;
 
+import org.candlepin.audit.Event;
+import org.candlepin.audit.EventFactory;
+import org.candlepin.audit.EventSink;
 import org.candlepin.config.ConfigProperties;
 import org.candlepin.config.Configuration;
 import org.candlepin.model.CertSerial;
@@ -24,6 +27,7 @@ import org.candlepin.model.ContentAccessCertificateCurator;
 import org.candlepin.model.IdentityCertificateCurator;
 import org.candlepin.model.Owner;
 import org.candlepin.model.OwnerCurator;
+import org.candlepin.util.Transactional;
 
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Streams;
@@ -38,7 +42,7 @@ import java.util.Objects;
 import javax.inject.Inject;
 
 /**
- * Class responsible for migration of consumers from anonymous orgs.
+ * Class responsible for migration of consumers from anonymous owners.
  */
 public class ConsumerMigration {
 
@@ -49,6 +53,8 @@ public class ConsumerMigration {
     private final IdentityCertificateCurator identityCertificateCurator;
     private final ContentAccessCertificateCurator contentAccessCertificateCurator;
     private final CertificateSerialCurator serialCurator;
+    private final EventFactory eventFactory;
+    private final EventSink eventSink;
     private final I18n i18n;
     private final int batchSize;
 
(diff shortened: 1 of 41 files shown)


src/test/java/org/candlepin/model/ConsumerCuratorTest.java
```diff
@@ -2429,6 +2470,44 @@ public void testDeleteConsumerCreateCorrespondingRecordInDeleteConsumerTable() {
             .returns(consumer.getUuid(), DeletedConsumer::getConsumerUuid);
     }
 
+    @Test
+    public void testGetConsumerUuidsWithNullConsumerIds() {
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

src/main/java/org/candlepin/dto/manifest/v1/SubscriptionDTO.java
```diff
@@ -134,6 +134,15 @@ public String getOwnerKey() {
...
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @JsonIgnore
+    public boolean isOwnerAnonymous() {
+        return false;
+    }
```
joshmalbrecht, line 144: I believe that this is fine since the field is not being serialized

src/main/java/org/candlepin/model/Owned.java
```diff
@@ -15,11 +15,22 @@
...
 
+    /**
+     * @return the key of the {@link Owner} that owns this entity
+     */
     String getOwnerKey();
+
+    /**
+     * @return the anonymous state of the {@link Owner} that owns this entity
+     */
+    boolean isOwnerAnonymous();
```
nikosmoum, line 35: While I understand the original argument of adding this field at the root of the event for consistency, the way we have tied these interfaces to our model entities and events makes this change pollute all of them with this method. But I don't see a way around it, unless we do it in a non-standard way, which is also not ideal. So I guess lets leave it as is?

joshmalbrecht: Yeah, I thought the same thing when I was implementing this, but I'm not sure if there is a better way of handling this to be honest. :/

src/main/java/org/candlepin/controller/ConsumerMigration.java
```diff
@@ -49,38 +53,61 @@ public class ConsumerMigration {
...
+            .onRollback(status -> this.eventSink.rollback());
+
         boolean failed = false;
         for (List<String> consumerIdsBlock : Iterables.partition(consumerIds, this.batchSize)) {
             try {
-                this.ownerCurator.transactional(() -> this.migrateBatch(consumerIdsBlock, destinationOwner));
+                transaction.execute(() -> {
+                    this.migrateBatch(consumerIdsBlock, destinationOwner);
+
+                    List<String> consumerUuids = this.consumerCurator.getConsumerUuids(consumerIdsBlock);
```
nikosmoum: Is there a reason we need to fetch consumer IDs at the start of this method, and then do another query for each batch here to convert them to consumer uuids? could we not update the original query to fetch uuids instead (there is already such a method in OwnerCurator)?

nikosmoum (commented):
Looks good overall, I just have one question/suggestion

src/main/java/org/candlepin/controller/ConsumerMigration.java
joshmalbrecht: Good point, I think it makes sense to just update the initial query to retrieve the consumer uuids instead of ids as we avoid an unnecessary query per batch. We will have to update `identityCertificateCurator.listCertSerials` and `contentAccessCertificateCurator.listCertSerials` to use a collection of consumer UUIDs instead of IDs, but that isn't a big deal. We will also need to add a `lockAndLoadUuids` method in ConsumerCurator that creates a pessimistic write lock based on Consumer UUIDs.

joshmalbrecht: Updated `ConsumerMigration.migrate` to use consumer UUIDs instead of IDs

PR #5241, opened 2025-09-24
[M] CHAINSAW-474: Updated Refresher to use a single transaction

joshmalbrecht:
- Updated Refresher to use a single transaction to prevent some data being persisted while other data is rolled back to ensure data integrity.

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsSpecTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ * Copyright (c) 2009 - 2025 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -513,7 +513,6 @@ public class RefreshPoolsSpecTest {
             .isNotNull()
             .singleElement()
             .isEqualTo(content.getId());
-
 
         // Modify the content for this product/sub
         String updatedUrl = "http://www.updated-url.com";
@@ -1246,7 +1245,6 @@ public class RefreshPoolsSpecTest {
             .isNotEqualTo(serialBeforeUpdate.longValue());
     }
 
-
     @Test
     @OnlyInHosted
     public void shouldRegenerateEntsWhenProvidedProductIsRemoved() throws Exception {
@@ -1536,7 +1534,7 @@ public class RefreshPoolsSpecTest {
 
     @Test
     @OnlyInHosted
-    public void shouldRegenerateEntsWhenPooolEndDataChanges() throws Exception {
+    public void shouldRegenerateEntsWhenPoolEndDataChanges() throws Exception {
         ProductDTO prod = createProductWithContent();
         SubscriptionDTO sub = adminClient.hosted().createSubscription(Subscriptions.random(owner, prod));
 
@@ -1800,7 +1798,6 @@ public class RefreshPoolsSpecTest {
             .extracting(EntitlementDTO::getCertificates, as(collection(CertificateDTO.class)))
             .singleElement()
             .returns(owner2EntCertificate.getId(), CertificateDTO::getId);
-
 
         // Refresh the second org and verify it now detects the product change and regenerates its
         // entitlement(s)
--- a/src/main/java/org/candlepin/controller/Refresher.java
+++ b/src/main/java/org/candlepin/controller/Refresher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009 - 2023 Red Hat, Inc.
+ * Copyright (c) 2009 - 2025 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -14,6 +14,7 @@
  */
 package org.candlepin.controller;
 
+import org.candlepin.controller.util.ExpectedExceptionRetryWrapper;
 import org.candlepin.model.Owner;
 import org.candlepin.model.OwnerCurator;
 import org.candlepin.model.Pool;
@@ -25,6 +26,8 @@ import org.candlepin.service.exception.subscription.SubscriptionServiceException
 import org.candlepin.service.model.OwnerInfo;
 import org.candlepin.service.model.SubscriptionInfo;
 
+import org.hibernate.exception.ConstraintViolationException;
+import org.hibernate.exception.LockAcquisitionException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -40,8 +43,15 @@ import java.util.Set;
 
 
 
+
 public class Refresher {
     private static final Logger log = LoggerFactory.getLogger(Refresher.class);
+
+    /**
+     * The number of times to retry the refresh operation if it fails as a result of a constraint
+     * violation.
+     */
+    private static final int CONSTRAINT_VIOLATION_RETRIES = 4;
 
     private final PoolManager poolManager;
     private final SubscriptionServiceAdapter subAdapter;
@@ -80,6 +90,25 @@ public class Refresher {
         return this;
     }
 
+    /**
+     * @return true if the refresher should perform lazy entitlement certificate regeneration or false if the
+     *  refresher should regenerate entitlement certificates immediately.
+     */
+    public boolean getLazyCertificateRegeneration() {
+        return this.lazy;
+    }
+
+    /**
+     * Adds the {@link Owner} to list of owners that should be refreshed.
+     *
+     * @param owner
+     *  that should be refreshed
+     *
+     * @throws IllegalArgumentException
+     *  if the provided owner is null or has a null key
+     *
+     * @return a reference to this refresher
+     */
     public Refresher add(Owner owner) {
         if (owner == null || owner.getKey() == null) {
             throw new IllegalArgumentException("Owner is null or lacks identifying information");
@@ -100,6 +129,8 @@ public class Refresher {
      * for other orgs.
      *
      * @param product
+     *  the product that should be refreshed globally
+     *
      * @return this Refresher instance
      */
     public Refresher add(Product product) {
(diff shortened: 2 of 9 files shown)


src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -89,6 +119,13 @@ public Refresher add(Owner owner) {
         return this;
     }
 
+    /**
+     * @return the owners that are to be refreshed.
+     */
+    public Map<String, Owner> getOwners() {
+        return this.owners;
+    }
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -100,13 +137,22 @@ public Refresher add(Owner owner) {
...
         products.add(product);
         return this;
     }
 
+    /**
+     * @return the set of {@link Product}s that should be refreshed.
+     */
+    public Set<Product> getProducts() {
+        return this.products;
+    }
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

src/test/java/org/candlepin/controller/refresher/RefreshWorkerTest.java
```diff
@@ -1229,48 +1217,7 @@ public void testExecuteRetriesOnDatabaseDeadlock() {
...
-        worker.addSubscriptions(sinfo);
-
-        // Impl note: this only works because of the call to mockTransactionalFunctionality done in setup
-        this.mockEntityManager.getTransaction().begin();
-        Throwable thrown = assertThrows(Exception.class, () -> worker.execute(owner));
-        assertEquals(exception, thrown);
-
-        verify(mockProductCurator, times(1)).getProductsByIds(eq(null), any(Collection.class));
-        verify(mockProductCurator, times(1)).create(Mockito.any(Product.class), anyBoolean());
+        assertThrows(LockAcquisitionException.class, () -> worker.execute(owner));
```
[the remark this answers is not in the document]
joshmalbrecht: Handled in `RefresherTest.testRunShouldRetryTransactionForCertainException`

src/test/java/org/candlepin/controller/refresher/RefreshWorkerTest.java
```diff
@@ -1194,15 +1189,11 @@ public void testExecuteRetriesOnConstraintViolation() {
         RefreshWorker worker = this.buildRefreshWorker();
         worker.addSubscriptions(sinfo);
 
-        RefreshResult result = worker.execute(owner);
-        assertNotNull(result);
-
-        verify(mockProductCurator, times(2)).getProductsByIds(eq(null), any(Collection.class));
-        verify(mockProductCurator, times(3)).create(Mockito.any(Product.class), anyBoolean());
+        assertThrows(ConstraintViolationException.class, () -> worker.execute(owner));
```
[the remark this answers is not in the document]
joshmalbrecht: Handled in `RefresherTest.testRunShouldRetryTransactionForCertainException`

src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -149,49 +195,55 @@ public void run() {
             subscriptions.addAll(subs);
         }
 
-        for (SubscriptionInfo subscription : subscriptions) {
-            // drop any subs for owners in our owners list. we'll get them with the full
-            // refreshPools call.
-            OwnerInfo so = subscription.getOwner();
+        for (Owner owner : this.owners.values()) {
+            Supplier<Owner> refreshTask = () -> {
+                for (SubscriptionInfo subscription : subscriptions) {
```
nikosmoum: I wouldn't mix these 2 blocks (the for loop over subscriptions, and the for loop over the owners), since the former (subscriptions) is a dead code branch. That was supposed to be doing Refreshes for specific Subscriptions/Products (instead of for a specific Owner), but it has never really been used.

In fact I would say, move everything that is before the original `for (Owner owner : this.owners.values()) {` part, into its own private method, call that method from the run() method, but comment it out with a comment that explains that if we ever want to do Product-specific refreshes, we can uncomment it. 

And only the Owner for loop should be part of the refreshTask

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsSpecTest.java
```diff
@@ -1962,6 +1963,36 @@ public void shouldDetectProductChangesAcrossMultipleOrgsSCACertificateRegenerati
         assertNotEquals(owner2SCALastUpdate2, owner2SCALastUpdate3);
     }
 
+    @Test
+    @OnlyInHosted
+    public void shouldRecoverFromConcurrentProductCreation() throws Exception {
```
nikosmoum: This test isn't reproducing the issue unfortunately. I've copied it into my local checkout (without copying the production changes), and run it 100 times with `@RepeatedTest(100)`, and it succeeded all 100 times. Unless there's a way to make it fail when the fix isn't applied, we should remove it

src/main/java/org/candlepin/controller/refresher/RefreshWorker.java
```diff
@@ -385,92 +376,79 @@ private void mapExistingContent(Collection<Content> content) {
...
+     * @throws LockAcquisitionException
+     *  if two owners create the same products or content in simultaneous transactions but in a specific
+     *  ordering
+     *
      * @return
      *  the result of this refresh operation
      */
     @SuppressWarnings("indentation")
     public RefreshResult execute(Owner owner) {
-        Supplier<RefreshResult> block = () -> {
```
nikosmoum, line 393: I wonder why refresh was split into 2 transactions in the first place. To save time/resources upon failure? Chris may know, but we can ask him when he's back from PTO

src/main/java/org/candlepin/controller/Refresher.java
joshmalbrecht: Yeah, this makes sense and is a great idea! I made the changes and removed the now unused unit tests in `RefresherTest.java`.

spec-tests/src/test/java/org/candlepin/spec/pools/RefreshPoolsSpecTest.java
joshmalbrecht: Yeah, this spec test won't recreate the issue and I should have made a comment on the PR that points this out. I added this spec test because I noticed that the concurrent product creation scenario wasn't being tested at the spec test level. Unfortunately, I don't see a way to trigger an exception in the problematic code area at the spec test level to recreate the actual issue. 

I removed the spec test.

src/main/java/org/candlepin/controller/refresher/RefreshWorker.java
joshmalbrecht: Yeah, I'm sure either, but it is a good idea to have Chris take a look!

src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -178,21 +250,6 @@ public void run() {
...
-                this.recalculatePoolQuantitiesForOwner(owner);
-                this.updateRefreshDate(owner);
-            }
-            catch (SubscriptionServiceException e) {
-                throw new RuntimeException(
-                    String.format("Unexpected subscription error for organization '%s'", owner.getKey()), e);
-            }
-            catch (ProductServiceException e) {
-                throw new RuntimeException("Unexpected product error in pool refresh", e);
-            }
```
nikosmoum, line 194: Shouldn't these also be present in the refreshTask definition?

src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -100,15 +137,50 @@ public Refresher add(Owner owner) {
...
+    }
+
     public void run() {
+        // This call to refreshProducts is commented out because it is currently a dead code branch.
+        // If we are to use product/subscription specific refreshes in the future, we can uncomment this
+        // method invocation.
+        // this.refreshProducts();
+
+        for (Owner owner : this.owners.values()) {
+            Supplier<Owner> refreshTask = () -> {
```
nikosmoum: The `Supplier` interface made sense when applied in the `RefreshWorker.execute()` which had to return a `RefreshResult`, but in this new context, it doesn't need to return anything. So using the `Runnable` interface makes more sense (you'll have to add a `execute(Runnable action)` method in `ExpectedExceptionRetryWrapper` for this to work)

src/main/java/org/candlepin/controller/refresher/RefreshWorker.java
Ceiu: Back in the day, the processing of product data for the versioned products on a per-org basis was a huge chunk of the refresh time. And with lots of refreshes happening in parallel, it's a very common spot for a collision to occur and require the whole transaction to be rolled back. The general idea was to try to chunk up the logic and rollback sooner such that on a retry the product updating section would be mostly no-ops, and we wouldn't have wasted any time doing the pool refresh portion of the operation.

However, these days I think that's less of a concern with performance improvements and the entity layering model change. One thing we may want to investigate is how frequently we still hit collision-based retries during refresh, but I feel they're relatively rare beyond the first handful of orgs.

If my prediction and memory hold true, then it's probably no longer necessary to have separate transactions for the logical chunks of the operation. At least, not in the way we're designing.
A better approach overall is something like we're doing with Arbiter in that we just sync product data, then sync access as separate operations. That eliminates this problem entirely. Probably too much work to port to CP at this point in its life.

src/main/java/org/candlepin/controller/Refresher.java
```diff
@@ -100,15 +137,50 @@ public Refresher add(Owner owner) {
...
+            };
 
+            // Retry this operation if we hit a unique constraint violation (two orgs creating the
+            // same products or content in simultaneous transactions), or we deadlock (same deal,
+            // but in a spicy entity order).
+            new ExpectedExceptionRetryWrapper()
+                .addException(ConstraintViolationException.class)
+                .addException(LockAcquisitionException.class)
+                .retries(CONSTRAINT_VIOLATION_RETRIES)
+                .execute(() -> new Transactional(this.poolCurator.getEntityManager())
```
Ceiu: Don't create the transactional object directly. You can use:

```
.execute(() -> this.poolCurator.transactional()
  .allowExistingTransactions()
  .execute(refreshTask));
```

src/main/java/org/candlepin/controller/Refresher.java
joshmalbrecht: That makes sense! I added the `execute(Runnable action)` and updated the RefresherTask to be a Runnable.

src/main/java/org/candlepin/controller/Refresher.java
joshmalbrecht: Ah good catch! Yes they should. Updated

src/main/java/org/candlepin/controller/Refresher.java
joshmalbrecht: Good catch! Updated

src/main/java/org/candlepin/controller/util/ExpectedExceptionRetryWrapper.java
```diff
@@ -138,6 +138,36 @@ public <O> O execute(Supplier<O> action) {
...
+     *
+     * @param action
+     *  the action to retry until successful completion
+     */
+    public void execute(Runnable action) {
+        int retries = 0;
+
+        while (true) {
+            try {
+                action.run();
```
Ceiu: For overloads, rather than copy the logic -- which then has to be maintained separately whenever a change occurs -- we should wrap the provided runnable in a supplier and pass that through to the current implementation. 

An example of this can be seen in the Transactional class. Here, it would look something like this:

```
Supplier wrapped = () -> {
  action.run();
  return null;
}

this.execute(wrapped);
```

It may be possible to fold it down into one statement, but declaring it like this helps avoid issues with type inference picking the wrong overload and/or requiring an explicit cast in the call.

Ceiu (changes requested):
One last thing from me

src/main/java/org/candlepin/controller/util/ExpectedExceptionRetryWrapper.java
joshmalbrecht: Great point, this makes the method much more maintainable! Updated

PR #5248, opened 2025-10-01
Update openapi yml

Tomboyo:
This MR updates the OpenAPI yaml.

- The getConsuemr API now documents the 410: GONE response for when a consumer was deleted.


Ceiu:
The changes themselves look fine enough, but the PR/commit message should be updated to follow our standards:

- The commits should be squashed into a single commit, referencing the source Jira task where applicable
- The commit verbiage should be in past tense rather than present/future tense

PR #5455, opened 2026-04-06
[PQC] CANDLEPIN-1172: Removed extraneous key usage from generated certs

Ceiu:
- Certs generated by the X509CertificateBuilder will no longer include key usage flags for key or data encipherment


Ceiu:
Creating this as a draft PR as a temporary holding measure to allow downstream stakeholders a chance to test their flows to verify this won't break anything. I'm 99.9% certain it won't, but we'll do the nice thing and give them a chance to test everything here.

Ceiu:
This needs to be re-opened against main now that the PQC feature branch has been merged

PR #5458, opened 2026-04-08
[PQC] CANDLEPIN-1175: Enforce RHEL crypto policy for primitive crypto operations

bakajstep:
RHEL crypto policy restrictions (jdk.certpath.disabledAlgorithms) are not enforced by Java during primitive operations or when using BouncyCastle. Added CryptoPolicyValidator that reads the JVM security property at startup and validates configured schemes against it, throwing an error if any scheme uses forbidden algorithms or key sizes. The same validation is applied during manifest import if the manifest contains a scheme.json.

### DIFF
--- a/src/main/java/org/candlepin/pki/CryptoPolicyValidator.java
+++ b/src/main/java/org/candlepin/pki/CryptoPolicyValidator.java
@@ -1,0 +1,372 @@
+/*
+ * Copyright (c) 2009 - 2026 Red Hat, Inc.
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.pki;
+
+import org.candlepin.config.ConfigurationException;
+import org.candlepin.sync.SchemeFile;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.ByteArrayInputStream;
+import java.security.Security;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.security.interfaces.DSAPublicKey;
+import java.security.interfaces.ECPublicKey;
+import java.security.interfaces.RSAPublicKey;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+
+
+/**
+ * Validates cryptographic schemes against the system's crypto policy as defined by the JVM security
+ * property {@code jdk.certpath.disabledAlgorithms}.
+ *
+ * <p>RHEL crypto policy restrictions are loaded into JVM security properties but are not enforced by
+ * Java during primitive cryptographic operations (key/cert generation) or when using BouncyCastle as
+ * the TLS provider. This class reads those restrictions and enforces them for Candlepin's configured
+ * schemes.
+ *
+ * <p>The parser handles the JDK disabled algorithms format:
+ * <ul>
+ *   <li>Plain algorithm name: {@code MD2} (fully disabled)</li>
+ *   <li>Key size constraint: {@code RSA keySize < 2048}</li>
+ *   <li>Conditional constraints ({@code jdkCA}, {@code denyAfter}, {@code usage}): ignored, as they
+ *       are certpath/TLS-context-specific and not relevant to primitive operations</li>
+ *   <li>{@code include <property>}: resolved recursively</li>
+ * </ul>
+ */
+@Singleton
+public class CryptoPolicyValidator {
+    private static final Logger log = LoggerFactory.getLogger(CryptoPolicyValidator.class);
+
+    static final String CERTPATH_DISABLED_ALGORITHMS_PROPERTY = "jdk.certpath.disabledAlgorithms";
+
+    private static final Pattern KEY_SIZE_PATTERN = Pattern.compile(
+        "keySize\\s*(<=|<|==|!=|>=|>)\\s*(\\d+)", Pattern.CASE_INSENSITIVE);
+
+    private final List<DisabledAlgorithm> disabledAlgorithms;
+
+    @Inject
+    public CryptoPolicyValidator() {
+        this(Security.getProperty(CERTPATH_DISABLED_ALGORITHMS_PROPERTY));
+    }
+
+    /**
+     * Constructs a validator from a raw disabled algorithms string (the format used by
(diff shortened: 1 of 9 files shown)


src/main/java/org/candlepin/pki/CryptoPolicyValidator.java
```diff
@@ -0,0 +1,372 @@
...
+
+    /**
+     * Represents a key size constraint from the crypto policy (e.g. "keySize < 2048").
+     *
+     * @param operator
+     *  the comparison operator as a string
+     * @param value
+     *  the key size threshold in bits
+     */
+    record KeySizeConstraint(String operator, int value) {
```
Ceiu, line 359: If we're going to approach something like a proper parser+tokenizer, we shouldn't be doing the evaluation on every check. This should be, at minimum, an enum.

src/main/java/org/candlepin/pki/CryptoPolicyValidator.java
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        return false;
+    }
+
+    /**
+     * Attempts to extract the key size (in bits) from a Base64-encoded X.509 certificate. Returns null
+     * if the certificate cannot be parsed or the key type is unrecognized.
+     */
+    private static Integer extractKeySizeFromCertificate(String base64Certificate) {
```
Ceiu, line 302: nack to this

This logic heavily codes us against specific key implementations and is extraneous for what we need. Also, it won't work for certs using ML-DSA key pairs since those aren't represented below.

src/test/java/org/candlepin/sync/ImporterTest.java
```diff
@@ -243,8 +198,13 @@ private void updateReleaseVersion(String version, String release) throws URISynt
     }
 
     private Importer buildImporter() {
+        return buildImporter(this.cryptoPolicyValidator);
+    }
+
+    private Importer buildImporter(CryptoPolicyValidator cryptoPolicyValidator) {
         return new Importer(this.mockConsumerTypeCurator, this.mockRulesImporter, this.mockOwnerCurator,
             this.mockIdentityCertCurator, this.refresherFactory, this.cryptoManager,
+            cryptoPolicyValidator,
```
Ceiu, line 207: Validation should be left to the CryptoManager as one of its interfaces. The backend logic can (and probably should) be compartmentalized to a validator class, but that should be transparent to the callers for exactly this reason: it's another dependency for a concern that's already owned by the CryptoManager.

This will also mean other changes to it, such as deciding how we want to deal with validation. E.g. do we really need to check if the scheme from a manifest can be used right before we use them? Probably not, but right now the validation step would logically include that. So we'll have to think about how to cleanly split those concerns apart without conflating the API.

src/main/java/org/candlepin/pki/CryptoPolicyValidator.java
```diff
@@ -0,0 +1,372 @@
...
+ *   <li>Conditional constraints ({@code jdkCA}, {@code denyAfter}, {@code usage}): ignored, as they
+ *       are certpath/TLS-context-specific and not relevant to primitive operations</li>
+ *   <li>{@code include <property>}: resolved recursively</li>
+ * </ul>
+ */
+@Singleton
+public class CryptoPolicyValidator {
+    private static final Logger log = LoggerFactory.getLogger(CryptoPolicyValidator.class);
+
+    static final String CERTPATH_DISABLED_ALGORITHMS_PROPERTY = "jdk.certpath.disabledAlgorithms";
```
Ceiu, line 66: This should be declared as public or private. If we're leaving it package private for testing, that's a violation of our testing standards.

src/main/java/org/candlepin/pki/CryptoPolicyValidator.java
```diff
@@ -0,0 +1,372 @@
...
+     * Validates algorithms from a manifest's {@link SchemeFile} against the system crypto policy.
+     * The key size is extracted from the embedded X.509 certificate if present.
+     *
+     * @param schemeFile
+     *  the scheme file from a manifest to validate
+     *
+     * @throws ConfigurationException
+     *  if the scheme file's algorithms violate the crypto policy
+     */
+    public void validateSchemeFile(SchemeFile schemeFile) {
```
Ceiu, line 135: This is hyper-specific to the manifest flow, and contributes to the abstractions-for-the-sake-of-abstractions problem we have throughout the codebase. Validation should be generic enough that this level isn't concerned with where the scheme originated, so such a method should not be needed.

src/main/java/org/candlepin/pki/CryptoPolicyValidator.java
```diff
@@ -0,0 +1,372 @@
...
+
+        Integer keySize = extractKeySizeFromCertificate(schemeFile.certificate());
+
+        List<String> violations = checkAlgorithms(
+            schemeFile.signatureAlgorithm(),
+            schemeFile.keyAlgorithm(),
+            keySize);
+
+        if (!violations.isEmpty()) {
+            throw new ConfigurationException(String.format(
```
Ceiu, line 148: This is a moot point given the other comments, but a scheme file originating from a manifest would not be a configuration issue -- it'd be a custom issue for the user to solve (the manifest cannot be loaded because the scheme it was built with is not supported by this Candlepin instance).

src/main/java/org/candlepin/pki/CryptoPolicyValidator.java
```diff
@@ -0,0 +1,372 @@
...
+
+        return violations;
+    }
+
+    /**
+     * Returns the parsed list of disabled algorithm entries. Visible for testing.
+     *
+     * @return an unmodifiable list of disabled algorithm entries
+     */
+    List<DisabledAlgorithm> getDisabledAlgorithms() {
```
Ceiu, line 209: No modifier

src/main/java/org/candlepin/pki/impl/bc/BouncyCastleCryptoManager.java
```diff
@@ -107,6 +108,10 @@ public BouncyCastleCryptoManager(Configuration config, BouncyCastleProvider secu
         Stream.concat(this.schemes.stream(), Stream.of(this.defaultScheme))
             .forEach(this::validateScheme);
 
+        // Validate schemes against the system crypto policy
+        Stream.concat(this.schemes.stream(), Stream.of(this.defaultScheme))
```
Ceiu, line 112: This should not be a separate validation step from the existing internal validation logic. I imagine this may change as we expose this/these method(s), but given the current implementation, this should not be a separate loop through the schemes.

Ceiu (changes requested):
There's a lot to unpack here. The primary problem with this approach is that it adds a completely separate validator that's independent of the crypto manager and its centralization of all the crypto and scheme stuff, while still being used by said crypto manager.

While I'm not a huge fan of that approach, it's not necessarily wrong on its own. However, if we were to do that, then *all* validation should move over to it rather than having some of it in the crypto manager and some in the new validator.

Additionally, the parsing logic also doesn't seem to pick a direction. It kind of parses+tokenizes, but stops short of actually committing to such an approach; leaving us in this "we've parsed, but we're still doing string testing on every check" space which is the worst of both worlds.

Out of curiosity, how much of this PR is AI generated? There are a lot of things in here that aren't your typical style, and the comments are written and formatted in a way that you don't normally write.

Ceiu:
Also, before I forget: the commit message doesn't at all follow our conventions. Something's amiss here.

Ceiu:
After some discussions, we're converting this to a draft while we wait for additional information to ensure we're implementing the right thing.

nikosmoum:
I've reached out and received feedback, and it turns out 'candlepin respects the underlying RHEL crypto policy' is about the algorithms allowed during the TLS connections to/from candlepin/tomcat itself. The crypto policy of the system Candlepin runs on should not affect the cert/key generation for clients, because client's crypro policies may differ from it.

PR #5459, opened 2026-04-08
[PQC] Renamed the ConsumerClient.exportCertificates utility function

Ceiu:
- Renamed the ConsumerClient.exportCertificates to exportCertificatePayloads to better indicate its actual behavior
- Updated several call sites to use the new method name


joshmalbrecht (approved):
Looks good!

PR #5460, opened 2026-04-08
[PQC] CANDLEPIN-1161: Added crypto capabilities support to cloud reg flows

Ceiu:
- The CloudRegistrationDTO now accepts optionally specifying crypto capabilities during token retrieval to influence certificate generation during anonymous consumer flows


Ceiu:
This is a tad more jank than I'd like, so if we want to reject these changes and rethink how anon consumers can get certs, I'm fine with that so long as we are permitted the time to do so.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
```diff
@@ -87,9 +92,52 @@
...
+
     private static final String STANDARD_TOKEN_TYPE = "CP-Cloud-Registration";
     private static final String ANON_TOKEN_TYPE = "CP-Anonymous-Cloud-Registration";
 
+
+    private static Stream<Arguments> capabilitiesSource() {
+        return CryptoCapabilities.getSupportedCapabilities()
+            .stream()
+            .map(Arguments::of);
+    }
```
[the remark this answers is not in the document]
Ceiu: CryptoCapabilities.getSupportedCapabilities() already includes null in its output.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
```diff
@@ -117,6 +165,68 @@ public void shouldAllowRegistrationWithValidToken() {
...
+
+    @ParameterizedTest(name = "[{index}] CryptographicCapabilitiesDTO")
+    @MethodSource("capabilitiesSource")
+    public void shouldNotAutomaticallyPropagateCapabilitiesFromV1Token(
+        CryptographicCapabilitiesDTO capabilities) {
+
+        // Test verifies that while the token endpoint does accept capabilities, those capabilities are not
+        // propagated to a registration request automatically. The user must provide them, again, during the
+        // normal registration or they will be omitted.
```
nikosmoum, line 166: we don't need this test, since what it is asserting is an existing design choice (that the state of an anonymous consumer does not get propagated into the normal consumer record once they properly register, and the client is supposed to provide all the data again anyway), but we could leave it to serve as documentation I suppose for someone trying to understand how autoreg works?

Ceiu:
Converting this to a draft for a bit to give us time to think out the implementation. Will convert back for re-review and merge, or close depending on how we feel about this implementation.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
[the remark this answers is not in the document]
Ceiu: fixed

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
```diff
@@ -261,6 +415,45 @@ public void shouldReceiveAnonTokenForV2AuthWithNonSyncedOwnerInCandlepin() {
             .returns(ANON_TOKEN_TYPE, CloudAuthenticationResultDTO::getTokenType);
     }
 
+    @ParameterizedTest(name = "[{index}] CryptographicCapabilitiesDTO")
+    @MethodSource("capabilitiesSource")
+    public void shouldAllowV2AnonymousTokenGenerationWithCryptoCapabilities(
```
nikosmoum, line 387: this test plus the `shouldAllowV2AnonymousTokenGenerationWithCryptoCapabilities` seem redundant, since the next one (`shouldExportCertificatesWithAnonymousTokenAccordingToProvidedCapabilities`) is already exercising all this

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -121,6 +123,26 @@ public Response cloudAuthorize(CloudRegistrationDTO cloudRegistrationDTO, Intege
...
+        try {
+            // Impl note: we won't have a real anon cloud consumer until much later, but we can still use a
+            // fake/temp instance to do our storage checks all the same. It's a bit of wasted effort, but it's
+            // better than firing off jobs and hitting our adapters (and other backing services) just to throw
+            // an exception.
+            AnonymousCloudConsumer temp = new AnonymousCloudConsumer();
+
+            if (cryptoCapabilities != null) {
+                temp.setSupportedKeyAlgorithmOids(cryptoCapabilities.getKeyAlgorithms());
+                temp.setSupportedSignatureAlgorithmOids(cryptoCapabilities.getSignatureAlgorithms());
```
nikosmoum, line 138: as we talked about offline, this validation could be extracted instead of using entity setters

nikosmoum (changes requested):

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
Ceiu: Yeah, that's why I added these tests: to make it clear that this specific behavior is known and expected. I'm trying to get in the habit of covering all of our behavior, implied or explicit.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
Ceiu: To an extent, but what I'm trying to separate each requirement with a specific test, even if a later test re-tests something as a requirement to get to where it needs. Perhaps there are better testing structures for that (nested tests, perhaps?), but that's the goal. E.g.:

- verify that providing crypto capabilities does not in any way impact token generation
- verify that the certificate export operation respects the crypto capabilities attached to an anonymous consumer (however we got there)

In my mind those are very different concerns, and I would like those requirements to survive refactors or anything of the such. For instance, if we come up with new ways to create anonymous consumers, the second test would survive a refactor without losing coverage of the first.

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
Ceiu: I'll poke around with some ideas in this space, like seeing if we can leverage the validator framework we use in other areas.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
nikosmoum: > the second test would survive a refactor

fair enough

Ceiu:
This needs to be re-opened against main now that the PQC feature branch has been merged

PR #5463, opened 2026-04-10
[M] CANDLEPIN-1161: Added crypto capabilities support to cloud reg flows

Ceiu:
- The CloudRegistrationDTO now accepts optionally specifying crypto capabilities during token retrieval to influence certificate generation during anonymous consumer flows

### DIFF
--- a/spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
+++ b/spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
@@ -15,6 +15,7 @@
 package org.candlepin.spec;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.candlepin.spec.bootstrap.assertions.CertificateAssert.assertThatCert;
 import static org.candlepin.spec.bootstrap.assertions.JobStatusAssert.assertThatJob;
 import static org.candlepin.spec.bootstrap.assertions.StatusCodeAssertions.assertBadRequest;
 import static org.candlepin.spec.bootstrap.assertions.StatusCodeAssertions.assertNotFound;
@@ -30,9 +31,11 @@ import org.candlepin.dto.api.client.v1.AsyncJobStatusDTO;
 import org.candlepin.dto.api.client.v1.CertificateDTO;
 import org.candlepin.dto.api.client.v1.ClaimantOwner;
 import org.candlepin.dto.api.client.v1.CloudAuthenticationResultDTO;
+import org.candlepin.dto.api.client.v1.CloudRegistrationDTO;
 import org.candlepin.dto.api.client.v1.ConsumerDTO;
 import org.candlepin.dto.api.client.v1.ConsumerDTOArrayElement;
 import org.candlepin.dto.api.client.v1.ContentDTO;
+import org.candlepin.dto.api.client.v1.CryptographicCapabilitiesDTO;
 import org.candlepin.dto.api.client.v1.OwnerDTO;
 import org.candlepin.dto.api.client.v1.ProductDTO;
 import org.candlepin.dto.api.client.v1.UserDTO;
@@ -47,8 +50,10 @@ import org.candlepin.spec.bootstrap.client.api.CloudRegistrationClient;
 import org.candlepin.spec.bootstrap.client.api.OwnerClient;
 import org.candlepin.spec.bootstrap.client.request.Request;
 import org.candlepin.spec.bootstrap.client.request.Response;
+import org.candlepin.spec.bootstrap.data.builder.CloudRegistrations;
 import org.candlepin.spec.bootstrap.data.builder.Consumers;
 import org.candlepin.spec.bootstrap.data.builder.Contents;
+import org.candlepin.spec.bootstrap.data.builder.CryptoCapabilities;
 import org.candlepin.spec.bootstrap.data.builder.Owners;
 import org.candlepin.spec.bootstrap.data.builder.Pools;
 import org.candlepin.spec.bootstrap.data.builder.Products;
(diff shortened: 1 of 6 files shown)


src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -121,6 +123,26 @@ public Response cloudAuthorize(CloudRegistrationDTO cloudRegistrationDTO, Intege
...
 
+        // If the user has provided crypto capabilities, verify that they won't cause storage problems later
+        // to avoid doing a lot of unnecessary work before throwing our BRE
+        CryptographicCapabilitiesDTO cryptoCapabilities = cloudRegistrationDTO.getCryptographicCapabilities();
+        try {
+            // Impl note: we won't have a real anon cloud consumer until much later, but we can still use a
+            // fake/temp instance to do our storage checks all the same. It's a bit of wasted effort, but it's
+            // better than firing off jobs and hitting our adapters (and other backing services) just to throw
+            // an exception.
+            AnonymousCloudConsumer temp = new AnonymousCloudConsumer();
```
Ceiu, line 134: From the last PR: Investigate moving this to the validator framework or otherwise abstracting out the validation logic instead of relying on the model object for this behavior.

Ceiu: Looking into this a bit, the hard part here is that the model object does its check after converting the data into its internal storage format. We could indeed duplicate that logic out into a validation check, but doing so complicates things in a couple ways:

- if the model object uses the validator, the validator either becomes aware of the conversion logic, duplicating it and throwing out the result, or it only checks the final converted data length. In the case of the latter, it would mean the conversion logic needs to be done at any call site where the validator would be used.
- if the model object does not use a validator, then we still have to ensure that the logic is similar enough that both the model object's conversion and validation loosely matches the resource input validation conversion and logic. Basically this is the anti-DRY approach.

The other obnoxious thing here is that our validator framework doesn't actually *exist*. We have several validators that all validate different things with completely different APIs and intended uses. The generalized validators that fact and config validation uses aren't actually used at the resource level. 

Generalizing validation to clean up this specific branch could be a task in its own right, and I wonder how quickly it'll fractal out of control.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
```diff
@@ -87,9 +92,19 @@
...
+    private static final int CLOUDREG_V2 = 2;
+
     private static final String STANDARD_TOKEN_TYPE = "CP-Cloud-Registration";
     private static final String ANON_TOKEN_TYPE = "CP-Anonymous-Cloud-Registration";
 
+
+    private static Stream<Arguments> capabilitiesSource() {
+        return CryptoCapabilities.getSupportedCapabilities()
+            .stream()
+            .map(Arguments::of);
```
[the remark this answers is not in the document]
Ceiu: As with every other time this comment is made by sourcery: capabilitiesSource includes null. :/

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
nikosmoum: Based on offline discussions: we're going to leave this implementation as is for now, and will create a task in the backlog to come back to it.

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -154,7 +176,8 @@ else if (version == 2) {
 
                 validateCloudAuthenticationResult(authResult);
 
-                CloudAuthenticationResultDTO resultDTO = processAdapterAuthResult(principal, authResult);
+                CloudAuthenticationResultDTO resultDTO = this.processAdapterAuthResult(principal,
+                    cryptoCapabilities, authResult);
```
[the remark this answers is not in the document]
nikosmoum: @Ceiu I think we want them to be immutable once set, right? a cloud vm with a temporary/anonymous cert switching capabilities is an extreme edge case

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
```diff
@@ -121,6 +123,26 @@ public Response cloudAuthorize(CloudRegistrationDTO cloudRegistrationDTO, Intege
...
+        CryptographicCapabilitiesDTO cryptoCapabilities = cloudRegistrationDTO.getCryptographicCapabilities();
+        try {
+            // Impl note: we won't have a real anon cloud consumer until much later, but we can still use a
+            // fake/temp instance to do our storage checks all the same. It's a bit of wasted effort, but it's
+            // better than firing off jobs and hitting our adapters (and other backing services) just to throw
+            // an exception.
+            AnonymousCloudConsumer temp = new AnonymousCloudConsumer();
+
+            if (cryptoCapabilities != null) {
+                temp.setSupportedKeyAlgorithmOids(cryptoCapabilities.getKeyAlgorithms());
```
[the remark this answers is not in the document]
nikosmoum: We decided not to do this for now, since we have no clean way of doing it

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
```diff
@@ -87,9 +92,19 @@
 @OnlyWithCapability("cloud_registration")
 class CloudRegistrationSpecTest {
 
+    private static final int CLOUDREG_V1 = 1;
+    private static final int CLOUDREG_V2 = 2;
```
nikosmoum: Do we still need these if we have helper methods in `CloudRegistrationClient.java` to perform either?

src/main/java/org/candlepin/resource/CloudRegistrationResource.java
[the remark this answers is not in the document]
Ceiu: Mutability isn't a concern here in either case, IMO. For our flows, they're currently effectively immutable; but if they could somehow change that through some new API in the future, the way the rest of the code works, it'd just cause the next generation of certs to use the right scheme.

What the bot is calling out here is weird. The method annotated is only ever called once, and only in the creation flow. Probably just identifying a general pattern.

spec-tests/src/test/java/org/candlepin/spec/CloudRegistrationSpecTest.java
Ceiu: Probably, but maybe not here. I've made the declarations in CloudRegistrationClient public, added documentation, removed these declarations, and updated the couple areas that use them to use the constants from CloudRegistrationClient.

nikosmoum (approved):

PR #5467, opened 2026-04-13
[M] CANDLEPIN-1191: Compile with Java 25

nikosmoum:
- Gradle updated to use Java 25 for compiling & running
- GH Actions updated to use Java 25 for compiling & running
- vagrant / ansible setup updated to install and use Java 25 for compiling and running


joshmalbrecht (approved):
Looks good!

PR #5475, opened 2026-04-16
[M] CANDLEPIN-1195: Added catch for EntityNotFoundException

joshmalbrecht:
- Added a catch block for ConsumerCurator.getConsumer() so that concurrently deleted entities that are eagerly fetched for Consumer does not cause a 500 error.

This change fixes the following spec test that intermittently fails:
`ConsumerResourceSpecTest.shouldReturn404Or410WhenConcurrentRegisterIsDeletedByAnotherRequest`

### Testing

You can add the `@RepeatedTest(100)` to the problematic `ConsumerResourceSpecTest.shouldReturn404Or410WhenConcurrentRegisterIsDeletedByAnotherRequest` spec test and compare before and after the changes.

### DIFF
--- a/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -59,6 +59,7 @@ import javax.inject.Inject;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 import javax.persistence.EntityManager;
+import javax.persistence.EntityNotFoundException;
 import javax.persistence.LockModeType;
 import javax.persistence.NoResultException;
 import javax.persistence.PersistenceException;
@@ -541,6 +542,10 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
             return em.createQuery(query).getSingleResult();
         }
         catch (NoResultException e) {
+            return null;
+        }
+        catch (EntityNotFoundException e) {
+            // Handle concurrently deleted entities that are eagerly fetched
             return null;
         }
     }
--- a/src/main/java/org/candlepin/model/ConsumerCurator.java
+++ b/src/main/java/org/candlepin/model/ConsumerCurator.java
@@ -59,6 +59,7 @@ import javax.inject.Inject;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 import javax.persistence.EntityManager;
+import javax.persistence.EntityNotFoundException;
 import javax.persistence.LockModeType;
 import javax.persistence.NoResultException;
 import javax.persistence.PersistenceException;
@@ -541,6 +542,11 @@ public class ConsumerCurator extends AbstractHibernateCurator<Consumer> {
             return em.createQuery(query).getSingleResult();
         }
         catch (NoResultException e) {
+            return null;
+        }
+        catch (EntityNotFoundException e) {
+            // Concurrent deletions of eagerly fetched relationships can cause this error.
+            log.error("Unable to retrieve consumer", e);
             return null;
         }
     }
--- a/src/test/java/org/candlepin/model/ConsumerCuratorTest.java
+++ b/src/test/java/org/candlepin/model/ConsumerCuratorTest.java
@@ -22,6 +22,9 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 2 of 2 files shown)


src/test/java/org/candlepin/model/ConsumerCuratorTest.java
```diff
@@ -332,6 +338,22 @@ public void testGetConsumerInputPartitioning() {
...
+        doReturn(spyEntityManager).when(spyCurator).getEntityManager();
+
+        QueryImplementor<Consumer> mockQuery = Mockito.mock(QueryImplementor.class);
+        doReturn(mockQuery).when(spyEntityManager).createQuery(any(CriteriaQuery.class));
+        doThrow(EntityNotFoundException.class).when(mockQuery).getSingleResult();
+
+        Consumer actual = spyCurator.getConsumer("uuid");
+
+        assertNull(actual);
+    }
```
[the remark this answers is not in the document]
joshmalbrecht: Updated

src/test/java/org/candlepin/model/ConsumerCuratorTest.java
```diff
@@ -332,6 +338,22 @@ public void testGetConsumerInputPartitioning() {
         }
     }
 
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testGetConsumerReturnsNullWhenEntityNotFoundExceptionIsThrown() {
```
bakajstep, line 343: Why is all the mocking here? This is curator, so it has `DatabaseTestFixture` so it could be only this:
```
@Test
public void testGetConsumerReturnsNullWhenEntityNotFoundExceptionIsThrownV2() {
    Consumer consumer = consumerCurator.getConsumer("unknown-uuid");

    assertNull(consumer);
}
```

bakajstep (commented):
i have one question otherwise LGTM

src/test/java/org/candlepin/model/ConsumerCuratorTest.java
joshmalbrecht: The problem is that `TypedQuery.getSingleResult()` would throw a `NoResultException` instead of the `EntityNotFoundException`.

Ceiu: We need a test that checks not the not-found path, but "is kind of found, but base entity no longer exists" -- which is what triggers EntityNotFoundException rather than NoResultException.

Mocking isn't ideal here, but it makes sense given the specific type of exception we're targeting. We'd need to prime Hibernate with partial state -- a consumer object that internally thinks it has some children, but that don't actually exist. I don't know how possible that'd be without mocks or reflection.

src/main/java/org/candlepin/model/ConsumerCurator.java
```diff
@@ -543,6 +544,10 @@ public Consumer getConsumer(String uuid) {
         catch (NoResultException e) {
             return null;
         }
+        catch (EntityNotFoundException e) {
```
Ceiu, line 547: Rare agreement with sonar. We can combine these two branches since it's effectively the same thing for us here.

```
catch (EntityNotFoundException | NoResultException e) {
   ...
}
```

Ceiu (changes requested):
The major question I have here is whether or not this is the right answer. EntityNotFound is kind of a weird exception in that it means we have a murky state within Hibernate with partial data. Are we sure that we want to hide that erroneous state?

src/test/java/org/candlepin/model/ConsumerCuratorTest.java
bakajstep: ohh ok I misinterpreted the EntityNotFoundException

src/main/java/org/candlepin/model/ConsumerCurator.java
Ceiu: Alternatively, as Sourcery suggests, keep both branches, but add warn/debug/error logging in the EntityNotFound branch since it's exceptional and weird.

joshmalbrecht:
> The major question I have here is whether or not this is the right answer. EntityNotFound is kind of a weird exception in that it means we have a murky state within Hibernate with partial data. Are we sure that we want to hide that erroneous state?

@Ceiu  Yeah, that's a good point, and I'm not so sure that returning null from the curator and hiding/logging the error is the best answer as well. Since the exception is more broad than this specific edge case, we would be covering up a much larger problem.

What are your thoughts on tossing out this specific edge case from the problematic test by adding a condition that checks for a 500 status code and checks for a specific line in the error message? 

Something like the following:

```java
                try {
                    userClient.consumers().deleteConsumer(consumer.getUuid());
                }
                catch (ApiException e) {
                    if (e.getCode() == 410 || e.getCode() == 404) {
                        expectedExceptions.add(e);
                    }
                    else if (e.getCode() == 500 &&
                        e.getMessage().contains("Unable to find org.candlepin.model.IdentityCertificate")) {

                        // While concurrently deleting and retrieving a Consumer, it is possible that we run
                        // into a EntityNotFoundException because the IdentityCertificate is eagerly loaded.
                        // Since this error does not pertain to the objective of this test, we are throwing
                        // out this exception.
                    }
                    else {
                        unexpectedExceptions.add(e);
                    }
                }
```

Ceiu:
Hrmm... I'm not sure I like that, either, since we're intentionally surfacing 500s to the user. Maybe keeping the structure you have now with ENF having its own catch branch and then doing a `log.error` so we can see how frequently it's happening in the logs. It's mostly hidden in terms of how it functions, but we'll still find out if it's a common enough problem to hit this.

Given how the test is operating, this specific instance for it seems to be expected or even intended to some degree. That feels a bit off, but I'm not sure the archeology necessary to find a better answer would give us anything new.

src/main/java/org/candlepin/model/ConsumerCurator.java
joshmalbrecht: Kept both branches, and added an error log to the `EntityNotFound` branch.

Ceiu (approved):

PR #5481, opened 2026-04-20
[M] Add CLAUDE.md, deploy-container script & overhaul documentation

nikosmoum:
- Added a deploy-container script equivalent to the existing deploy
  script, but for containers instead.
- Added a CLAUDE.md and .claude/settings.json.
- CLAUDE.md has no links to other documents to keep context small,
  but other documents can have links to CLAUDE.md.
- Made CONTRIBUTING.md more up to date.
- Migrated the versioned container instructions from CONTRIBUTING.md
  to containers/README.md.
- checkstyle changes: updated import group list, disallowed
  wildcards in static imports and disallowed uppercase package
  names.


nikosmoum:
I realized there's significant duplication between CLAUDE.md and CONTRIBUTING.md. I'll need to find a solution for that before we merge this.

CLAUDE.md
```diff
@@ -0,0 +1,117 @@
...
+### Language
+
+- Follow the instructions in [CONTRIBUTING.md#language](CONTRIBUTING.md#language) for language norms.
+
+### Git Commits
+
+- Follow the commit message format in [CONTRIBUTING.md#git-commits](CONTRIBUTING.md#git-commits).
+- Do not use 'Co-authored-by' trailer for AI-generated code. Use 'Assisted-by'.
+
+### Code Style
```
Ceiu: Should we convert that code style document we wrote to markdown and include it here instead?

CLAUDE.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+When validating incremental changes, run unit and spec tests scoped to the affected package or resource class.
+- Unit tests: JUnit 5, Mockito, AssertJ. Source in `src/test/java/`.
+- Spec tests: integration tests using an auto-generated OpenAPI client. Source in `spec-tests/src/test/java/`.
+  - Require a running Candlepin instance (default: `localhost:8443`) — see [Running Locally](#running-locally-dev-container).
+  - When writing spec tests, use [spec-tests/README.md](spec-tests/README.md)
+- Checkstyle config: `config/checkstyle/checkstyle.xml`. Zero warnings/errors tolerance.
+
+
```
joshmalbrecht: Nit: Extra new line

CONTRIBUTING.md
```diff
@@ -57,11 +52,9 @@ following terms in the code, messages, or comments. Note the alternatives we wil
 ##### Rules for Git Commit Message Style
 - Separate subject from the body with a blank line
 - Use past tense verbs
-- Wrap the lines at 72 characters
-- Use the body to explain what changed
-  - In the few cases it makes sense, please include reason the change was needed, or why it was implemented in a certain way
-- Prefix the subject to include the Jira card number if it exists (CANDLEPIN-1234:)
-- Prefix the subject to include the Bugzilla number if it exists (12345678:)
+- Wrap the lines at 70 characters
```
joshmalbrecht: Are we intentionally lowering the line length from 72 to 70?

joshmalbrecht (changes requested):
In general, this looks like a great starting place to me! I'm sure we will adjust things as we learn more about how to use Claude with our project.

I just had a question about the git commit line length change.

CLAUDE.md
```diff
@@ -0,0 +1,117 @@
+# CLAUDE.md
```
bakajstep, line 1: If we have `.claude` folder, should we also move `CLAUDE.md` to that folder so we have it in one place?

CLAUDE.md
```diff
@@ -0,0 +1,117 @@
...
+
+This file provides guidance to Claude Code when working with code in this repository.
+
+Claude and other AI tools/agents MUST read [CONTRIBUTING.md](CONTRIBUTING.md) before writing any code because
```
bakajstep: If we want to tell the agent to load another file from Claude, it should be referenced like you would do in CLI through `@` https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md

CLAUDE.md
```diff
@@ -0,0 +1,117 @@
...
+./gradlew spec                                              # All spec (integration) tests
+./gradlew spec --tests org.candlepin.spec.StatusSpec*       # Single spec test
+
+# Lint
+./gradlew checkstyle                                        # Checkstyle on all code
+```
+
+When validating incremental changes, run unit and spec tests scoped to the affected package or resource class.
+- Unit tests: JUnit 5, Mockito, AssertJ. Source in `src/test/java/`.
+- Spec tests: integration tests using an auto-generated OpenAPI client. Source in `spec-tests/src/test/java/`.
```
bakajstep: maybe i miss that but tell it that you need to firstly start the candlepin and then it can run spec tests

bakajstep (changes requested):

CLAUDE.md
nikosmoum: The standard is to have CLAUDE.md at the root of the repo. `.claude` is for settings, skills, agents, etc.

CLAUDE.md
nikosmoum: it's right on the next line, indented as a sub-bullet: `Require a running Candlepin instance`

CLAUDE.md
nikosmoum: I'm trying to minimize the amount of links from CLAUDE.md to other files, but you're right. The links that need to be there I will update to use `@`

CLAUDE.md
bakajstep: yes you are right sorry i missread it with global claude.md

CLAUDE.md
bakajstep: Yeah, I read that, but from what I've gathered, it's better to tell him, for example, that Candlepin needs to be running and give him instructions on how to start it(`Require a running Candlepin instance - ./bin/deploy -gaH`). We can leave like this and see if it can connect the dots.

CLAUDE.md
bakajstep: yes that is a good idea there is research that it perform better with smaller begging context

nikosmoum: I removed all links to other files

CLAUDE.md
nikosmoum: removed

CLAUDE.md
nikosmoum: added

CONTRIBUTING.md
nikosmoum: reverted

CONTRIBUTING.md
```diff
@@ -91,182 +80,40 @@ Other workflows include:
...
-Docker is not strictly necessary: it is used to run the MariaDB and PostgreSQL tests which are not enabled by default.
-
-### IDE Config and Code Style
-TODO
-#### Eclipse Setup
-TODO
-#### IDEA Setup
-TODO
+### Run Active Development Container
+Assuming podman & docker-compose are installed, the following script can be used to deploy
```
joshmalbrecht: We should update `docker-compose` to `podman-compose` because `deploy-container` uses podman-compose.

joshmalbrecht (changes requested):
Looks good! I just noticed a small change

CONTRIBUTING.md
nikosmoum: I know it's confusing, but `podman compose != podman-compose` :smile: 
The `deploy-container` script is currently running the former, which is a `podman` command, and the `compose` subcommand could be backed by either of: `podman-compose` or `docker-compose` packages.

I mentioned `docker-compose` because that's what I'm using (I don't use `podman-compose` because it's undermaintained, does not have feature parity with `docker-compose` and not officially supported by RH). But I will update this to clarify that you can use either

nikosmoum: updated

joshmalbrecht: Ahh yeah, that makes sense! I didn't even notice the missing '-' haha

joshmalbrecht (approved):

bakajstep (approved):

PR #5489, opened 2026-04-23
[M] CANDLEPIN-1193: Fixed ML-DSA signature verification failure with BC 1.84

bakajstep:
- SchemeFile.toScheme() was using CertificateFactory.getInstance("X.509") without specifying a security provider. On JDK 25, this defaults to the SUN provider which creates JDK-native ML-DSA public keys. These keys trigger a type mismatch bug in BC 1.84's SignatureSpi.verifyInit() where pqc.crypto.util.PublicKeyFactory returns the old MLDSAPublicKeyParameters type but the code casts to the new crypto.params.MLDSAPublicKeyParameters, causing a ClassCastException that silently fails signature verification.
- Using BC's provider explicitly ensures certificates produce BCMLDSAPublicKey objects, which take the direct code path and bypass the buggy branch.

### DIFF
--- a/src/main/java/org/candlepin/sync/Importer.java
+++ b/src/main/java/org/candlepin/sync/Importer.java
@@ -39,6 +39,7 @@ import org.candlepin.model.ImportUpstreamConsumer;
 import org.candlepin.model.Owner;
 import org.candlepin.model.OwnerCurator;
 import org.candlepin.model.UpstreamConsumer;
+import org.candlepin.pki.CertificateReader;
 import org.candlepin.pki.CryptoManager;
 import org.candlepin.pki.Scheme;
 import org.candlepin.service.SubscriptionServiceAdapter;
@@ -58,6 +59,7 @@ import org.xnap.commons.i18n.I18n;
 
 import tools.jackson.databind.ObjectMapper;
 
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
@@ -69,6 +71,7 @@ import java.io.Reader;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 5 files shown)


Ceiu (changes requested):
nack to this

Adding the security provider as input forces all areas of code using it to be aware of which security provider they're using, or having to fetch one from somewhere. In the vast majority of cases, it should be backend wiring, either coming from DI or whatever is generating the class in question. The CryptoManager call used to populate this parameter is a stop-gap for truly exceptional cases, which I don't yet believe this is. 

In this case, the SchemeFile itself is acting as a sort of self-factory, which is going to cause some problems and require updating the logic accordingly. We already have its logic mostly encapsulated in CertificateReader, which can probably be used in place of the security provider if we don't want to make major changes here.

src/main/java/org/candlepin/sync/Importer.java
```diff
@@ -501,7 +501,8 @@ private Optional<Scheme> loadSchemeFromManifest(File exportDir)
             SchemeFile schemeFile = this.mapper
                 .readValue(zipFile.getInputStream(schemeEntry), SchemeFile.class);
 
-            return Optional.of(SchemeFile.toScheme(schemeFile));
+            return Optional.of(SchemeFile.toScheme(schemeFile,
+                this.cryptoManager.getSecurityProvider()));
```
joshmalbrecht: I'm in favor of getting rid of the `SchemeFile.toScheme` method and using the `CertificateReader` to read the X509 cert and then using the Scheme.Builder to build the scheme in this method.

joshmalbrecht (changes requested):

src/main/java/org/candlepin/sync/Importer.java
bakajstep: updated

Ceiu (approved):

Ceiu:
Let's update the commit message to only have the second bullet. The first is fairly extraneous and doesn't really help with understanding the change.

**Edit:** The second is kinda full of extra info as well. Really we just need to explain that the .toScheme method was removed, and cert parsing is handled by the CertificateReader now to ensure that we're using the same security provider and infrastructure everywhere.

joshmalbrecht (approved):

PR #5492, opened 2026-04-27
[PoC - Do not merge] Use FFM OpenSSL listener instead of tomcat-native+apr

nikosmoum:
- To remove our dependency on tomcat-native (which is only available on EPEL repo), this changes tomcat's TLS configuration to use the OpenSSLLifecycleListener, which is tomcat's FFM bridge to openssl.
- This requires a dependency on openssl-devel instead (which is available on the RHEL AppStream repo).
- This requires tomcat version 9.0.111 or newer to work (or 10.1.48 and newer), due to a tomcat FFM bug on 9.0.110 or earlier that caused the server to reject client certificates signed by its own CA.
- Removed 'sslProtocol' property (it caused empty replies with FFM).
- Removed the 'LegacySSLContextEditor' from update-server-xml.py.

### DIFF
--- a/.github/containers/cs9.Containerfile
+++ b/.github/containers/cs9.Containerfile
@@ -10,7 +10,7 @@ RUN dnf -y --setopt install_weak_deps=False update && \
     dnf clean all
 
 # Prepare Tomcat
-ARG TOMCAT_VERSION=9.0.110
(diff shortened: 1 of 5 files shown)


nikosmoum:
Closing this. The full change will be done as a separate task

PR #5512, opened 2026-05-06
[M] CANDLEPIN-1211: Added debugging support for CP containers

bakajstep:
- replaced grep -oP with sed
- converted WAR_ARGS/CONF_ARGS to bash arrays

You can test that by running:
`./bin/deployment/deploy-container -Hag`
and then running remote debug against it.

### DIFF
--- a/.github/containers/cs9.Containerfile
+++ b/.github/containers/cs9.Containerfile
@@ -46,6 +46,7 @@ RUN update-crypto-policies --set DEFAULT:PQ
 ENV JAVA_HOME=/usr/lib/jvm/jre-25-openjdk
 ENV JRE_HOME=/usr/lib/jvm/jre-25-openjdk
 ENV CATALINA_OPTS=-Djavax.net.ssl.trustStore=$JAVA_HOME/lib/security/cacerts
+ENV CATALINA_OPTS="$CATALINA_OPTS -agentlib:jdwp=transport=dt_socket,server=y,address=*:8000,suspend=n"
 
 # Tomcat Setup
 COPY --from=builder /opt/tomcat/ /opt/tomcat/
--- a/.github/containers/mariadb.docker-compose.yml
+++ b/.github/containers/mariadb.docker-compose.yml
@@ -27,6 +27,7 @@ services:
       - WAR_FILE=${WAR_FILE}
     ports:
       - "8443:8443"
+      - "8000:8000"
     healthcheck:
       test: curl --fail -k https://localhost:8443/candlepin/status
       timeout: 5s
--- a/.github/containers/postgres.docker-compose.yml
+++ b/.github/containers/postgres.docker-compose.yml
@@ -26,6 +26,7 @@ services:
       - WAR_FILE=${WAR_FILE}
     ports:
       - "8443:8443"
+      - "8000:8000"
     healthcheck:
       test: curl --fail -k https://localhost:8443/candlepin/status
       timeout: 5s
--- a/bin/deployment/deploy-container
+++ b/bin/deployment/deploy-container
@@ -51,7 +51,7 @@ fi
 CONTAINERFILE="${REPO_ROOT}/.github/containers/cs9.Containerfile"
 
 # Extract the Tomcat version from the Containerfile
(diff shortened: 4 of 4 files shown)


.github/containers/mariadb.docker-compose.yml
```diff
@@ -27,6 +27,7 @@ services:
       - WAR_FILE=${WAR_FILE}
     ports:
       - "8443:8443"
+      - "8000:8000"
```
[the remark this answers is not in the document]
bakajstep: this is ok this is only for github ci and for local development

.github/containers/postgres.docker-compose.yml
```diff
@@ -26,6 +26,7 @@ services:
       - WAR_FILE=${WAR_FILE}
     ports:
       - "8443:8443"
+      - "8000:8000"
     healthcheck:
```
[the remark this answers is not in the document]
bakajstep: this is ok this is only for github ci and for local development

.github/containers/cs9.Containerfile
```diff
@@ -46,6 +46,7 @@ RUN update-crypto-policies --set DEFAULT:PQ
 ENV JAVA_HOME=/usr/lib/jvm/jre-25-openjdk
 ENV JRE_HOME=/usr/lib/jvm/jre-25-openjdk
 ENV CATALINA_OPTS=-Djavax.net.ssl.trustStore=$JAVA_HOME/lib/security/cacerts
+ENV CATALINA_OPTS="$CATALINA_OPTS -Xdebug -Xrunjdwp:transport=dt_socket,address=*:8000,server=y,suspend=n"
```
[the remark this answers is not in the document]
bakajstep: fixed

bin/deployment/deploy-container
```diff
@@ -71,27 +71,27 @@ fi
...
 if [ ${#TEST_EXTS[*]} -gt 0 ]; then
     EXT_STR=$(IFS=, ; echo "${TEST_EXTS[*]}")
-    WAR_ARGS="-Ptest_extensions=${EXT_STR}"
+    WAR_ARGS+=("-Ptest_extensions=${EXT_STR}")
 fi
 
 # Build the WAR
 echo "Building Candlepin WAR..."
-"${REPO_ROOT}/gradlew" -p "$REPO_ROOT" war "$DATABASE_SERVER" "$WAR_ARGS"
+"${REPO_ROOT}/gradlew" -p "$REPO_ROOT" war "$DATABASE_SERVER" ${WAR_ARGS[@]+"${WAR_ARGS[@]}"}
```
Ceiu, line 94: What is the intent behind the quoted expansion here (and further down on line 103)?

Ceiu (changes requested):

bin/deployment/deploy-container
bakajstep: Because we have set `set -u` -> treat unset variables as errors and if you would have just `"${WAR_ARGS[@]}"`  and the array would be empty it will throw error. This is way how to safely expand empty array without throwing error.

also lern that becasue initially i had there only `"${WAR_ARGS[@]}"`

Ceiu: In what context are you seeing this error?

I had assumed this was the case, but I can't manually trigger the issue since we don't have any unset values in this context. At least, not on the newer versions of Bash that exists on my system and in our vagrant boxes.

Ceiu: For clarity here, while I rejected the PR as "requests changes", I don't necessarily want changes here. I only chose that to ensure this topic is discussed before merging.
I want to understand why the "simpler" approach of not doing the expansion is causing errors for you but not me. If that ultimately means we don't make changes, sure -- especially in cases like this where the expansion isn't harmful and the perf delta doesn't matter.

bakajstep: no worries i get that i would be also currios when it would be working on my pc. Problem is mac is shipped with bash in version 3.2 and this issue was fixed in bash version 4.4.
[bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=529627)

This is the command:
`./bin/deployment/deploy-container -ga`

Ceiu (approved):

